PR #865, opened 2018-12-14
repair queue masterDB support

cam-a:


pkg/datarepair/checker/config.go
```diff
@@ -45,11 +47,32 @@ func (c Config) initialize(ctx context.Context) (Checker, error) {
...
+	driver, _, err := utils.SplitDBURL(c.QueueAddress)
 	if err != nil {
-		return nil, Error.Wrap(err)
+		return nil, err
+	}
+
+	var repairQueue *queue.Queue
+	switch driver {
+	case "postgres":
+		postgresDB, ok := ctx.Value("masterdb").(interface {
```
egonelbre: Always use master database. The master database decides which repair queue implementation to use.

pkg/datarepair/checker/config.go
```diff
@@ -45,11 +47,32 @@ func (c Config) initialize(ctx context.Context) (Checker, error) {
...
+			RepairQueueDB() storage.Queue
+		})
+		if !ok {
+			return nil, errs.New("unable to get master db instance")
+		}
+		repairQueue = queue.NewQueue(postgresDB.RepairQueueDB())
```
egonelbre: It should return a compatible queue already, so the `queue.NewQueue` shouldn't be necessary here.

storage/testqueue/queue.go
```diff
@@ -22,15 +23,15 @@ func New() *Queue {
 }
 
 //Enqueue add a FIFO element
-func (q *Queue) Enqueue(value storage.Value) error {
+func (q *Queue) Enqueue(_ context.Context, value storage.Value) error {
```
egonelbre: These shouldn't require changes.

storage/redis/client_queue.go
```diff
@@ -32,7 +34,7 @@ func (client *Queue) Close() error {
 }
 
 //Enqueue add a FIFO element, for the storage.Queue interface
-func (client *Queue) Enqueue(value storage.Value) error {
+func (client *Queue) Enqueue(_ context.Context, value storage.Value) error {
```
egonelbre: These shouldn't require changes.

storage/common.go
```diff
@@ -73,11 +74,11 @@ type KeyValueStore interface {
 //Queue is an interface describing queue stores like redis
 type Queue interface {
 	//Enqueue add a FIFO element
-	Enqueue(Value) error
+	Enqueue(context.Context, Value) error
```
egonelbre: These shouldn't really require changes.

satellite/satellitedb/repairqueue.go
```diff
@@ -0,0 +1,74 @@
...
+
+	_, err = r.db.Create_Injuredsegment(
+		ctx,
+		dbx.Injuredsegment_Info(val),
+	)
+	return err
+}
+
+func (r *repairQueueDB) Dequeue(ctx context.Context) (pb.InjuredSegment, error) {
+	res, err := r.db.First_Injuredsegment(ctx)
```
egonelbre: Fetching and deleting the first should happen in a transaction.

pkg/datarepair/queue/queue_test.go
```diff
@@ -12,60 +12,85 @@ import (
...
+			dqSeg, err := q.Dequeue(ctx)
+			assert.NoError(t, err)
+			assert.True(t, proto.Equal(addSegs[i], &dqSeg))
+		}
+	})
 }
 
 func TestParallel(t *testing.T) {
-	queue := NewQueue(testqueue.New())
+	ctx := testcontext.New(t)
```
egonelbre: run this with masterdb as well

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -162,4 +162,18 @@ read limitoffset (
 )
 
 update overlay_cache_node ( where overlay_cache_node.key = ? )
-delete overlay_cache_node ( where overlay_cache_node.key = ? )
\ No newline at end of file
+delete overlay_cache_node ( where overlay_cache_node.key = ? )
+model injuredsegment (
+	key info
```
egonelbre: I don't think we need a primary key on this table.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -162,4 +162,18 @@ read limitoffset (
 )
 
 update overlay_cache_node ( where overlay_cache_node.key = ? )
-delete overlay_cache_node ( where overlay_cache_node.key = ? )
\ No newline at end of file
+delete overlay_cache_node ( where overlay_cache_node.key = ? )
+model injuredsegment (
```
egonelbre: separate more clearly from other models

satellite/satellitedb/dbx/satellitedb.dbx
cam-a: I think it returns an error with no key. From the dbx readme under *Models*, "key is required and declares the primary key for the model"

egonelbre: ok, nevermind then

cmd/satellite/main.go
```diff
@@ -83,7 +81,7 @@ var (
 		Database string `help:"satellite database connection string" default:"sqlite3://$CONFDIR/master.db"`
 	}
 	qdiagCfg struct {
-		DatabaseURL string `help:"the database connection string to use" default:"redis://127.0.0.1:6378?db=1&password=abc123"`
+		DatabaseURL string `help:"the database connection string to use" default:"sqlite3://$CONFDIR/master.db"`
```
mniewrzal: Can you change name and message to be consistent with other flags in this file?

pkg/datarepair/checker/config.go
```diff
@@ -15,13 +15,13 @@ import (
...
-	"storj.io/storj/storage/redis"
 )
 
 // Config contains configurable values for checker
 type Config struct {
-	QueueAddress string        `help:"data checker queue address" default:"redis://127.0.0.1:6378?db=1&password=abc123"`
-	Interval     time.Duration `help:"how frequently checker should audit segments" default:"30s"`
+	QueueAddress     string        `help:"data checker queue address" default:"sqlite3://$CONFDIR/repairqueue.db"`
+	Interval         time.Duration `help:"how frequently checker should audit segments" default:"30s"`
+	IrreparabledbURL string        `help:"the database connection string to use" default:"sqlite3://$CONFDIR/irreparable.db"`
```
mniewrzal: I think its unused.

pkg/datarepair/checker/config.go
```diff
@@ -38,19 +38,22 @@ func (c Config) initialize(ctx context.Context) (Checker, error) {
...
 		Irreparable() irreparable.DB
 	})
 	if !ok {
 		return nil, Error.New("unable to get master db instance")
 	}
-	o := overlay.LoadServerFromContext(ctx)
-	redisQ, err := redis.NewQueueFrom(c.QueueAddress)
-	if err != nil {
-		return nil, Error.Wrap(err)
+	qdb, ok := ctx.Value("masterdb").(interface {
```
mniewrzal: This shouldn't be separate. Take a look at https://github.com/storj/storj/blob/master/pkg/overlay/config.go#L71

pkg/datarepair/queue/queue.go
```diff
@@ -13,9 +15,9 @@ import (
 
 // RepairQueue is the interface for the data repair queue
 type RepairQueue interface {
-	Enqueue(qi *pb.InjuredSegment) error
-	Dequeue() (pb.InjuredSegment, error)
-	Peekqueue(limit int) ([]pb.InjuredSegment, error)
+	Enqueue(ctx context.Context, qi *pb.InjuredSegment) error
```
mniewrzal: Is using context for all this methods is necessary?

satellite/satellitedb/database.go
```diff
@@ -65,10 +66,10 @@ func (db *DB) OverlayCache() storage.KeyValueStore {
 	return newOverlaycache(db.db)
 }
 
-// // RepairQueueDB is a getter for RepairQueueDB repository
-// func (db *DB) RepairQueueDB() queue.DB {
-// 	return &repairQueueDB{db: db.db}
-// }
+// RepairQueueDB is a getter for RepairQueueDB repository
+func (db *DB) RepairQueueDB() queue.RepairQueue {
```
mniewrzal: I think `DB` suffix is not necessary here.

satellite/satellitedb/repairqueue.go
```diff
@@ -0,0 +1,80 @@
...
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/utils"
+	dbx "storj.io/storj/satellite/satellitedb/dbx"
+)
+
+type repairQueueDB struct {
+	db *dbx.DB
+}
+
+func (r *repairQueueDB) Enqueue(ctx context.Context, seg *pb.InjuredSegment) error {
```
mniewrzal: To avoid passing context to method take a look here https://github.com/storj/storj/blob/queue-dbx/satellite/satellitedb/overlaycache.go#L18

satellite/satellitedb/repairqueue.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return pb.InjuredSegment{}, Error.Wrap(err)
+	}
+
+	res, err := tx.First_Injuredsegment(ctx)
+	if err != nil {
+		return pb.InjuredSegment{}, Error.Wrap(utils.CombineErrors(err, tx.Rollback()))
+	}
+	if res == nil {
+		return pb.InjuredSegment{}, Error.Wrap(utils.CombineErrors(Error.New("Empty queue"), tx.Rollback()))
```
mniewrzal: You should use `storage.ErrEmptyQueue` here. Similar to redis implementation https://github.com/storj/storj/blob/queue-dbx/storage/redis/client_queue.go#L48

satellite/satellitedb/repairqueue.go
```diff
@@ -0,0 +1,80 @@
...
+	seg := &pb.InjuredSegment{}
+	err = proto.Unmarshal(res.Info, seg)
+	if err != nil {
+		return pb.InjuredSegment{}, Error.Wrap(utils.CombineErrors(err, tx.Rollback()))
+	}
+	return *seg, Error.Wrap(tx.Commit())
+}
+
+func (r *repairQueueDB) Peekqueue(ctx context.Context, limit int) ([]pb.InjuredSegment, error) {
+	rows, err := r.db.Limited_Injuredsegment(ctx, limit, 0)
```
mniewrzal: Limit should be normalized with something like this:
	if limit <= 0 || limit > storage.LookupLimit {
		limit = storage.LookupLimit
	}

pkg/datarepair/queue/queue.go
egonelbre: Yes, usually should be passing the `ctx` around.

pkg/datarepair/checker/config.go
cam-a: Should I remove QueueAddress as well?

mniewrzal:
I wanted to say that everything except parallel test looks ok for me :+1: I'm not sure if it will be easy/possible to overcome a problem with sqlite and parallel access. Maybe we should limit that tests somehow. I don't know.

PR #866, opened 2018-12-14
RevocationDB fixes

bryanchriswhite:


pkg/provider/identity.go
```diff
@@ -82,21 +82,23 @@ type ServerConfig struct {
 type ServerOptions struct {
 	Config   ServerConfig
 	Ident    *FullIdentity
+	RevDB    *peertls.RevocationDB
```
egonelbre: This looks very weird to close "Options".

bryanchriswhite: do you have an alternate name or flow in mind?

pkg/provider/identity.go
```diff
@@ -294,11 +296,14 @@ func (ic IdentityConfig) Run(ctx context.Context, interceptor grpc.UnaryServerIn
 	if err != nil {
 		return err
 	}
+	defer func() { _ = opts.RevDB.Close() }()
```
egonelbre: Check the error.

pkg/provider/identity.go
egonelbre: not at the moment

pkg/provider/identity.go
bryanchriswhite: What should we do with an error in a defer? Or should it not be in a defer?

egonelbre: There's no really nice way of handling them at the moment.

One way would be to use something like:

```
defer func() { err = utils.CombineErrors(err, opts.RevDB.Close()) }
```

The other would be to manually close everywhere.

bryanchriswhite: I don't understand how your defer example would do anything useful and I'm pretty sure we want to use a defer for this

egonelbre: Take a look at https://play.golang.org/p/0qh7EuHW7Ut. It relies on the behavior that you can change the returned `err` variable.

PR #873, opened 2018-12-17
Proper NodeType Handling

jenlij:
https://storjlabs.atlassian.net/browse/V3-922
https://storjlabs.atlassian.net/browse/V3-910
https://storjlabs.atlassian.net/browse/V3-1009


pkg/pb/node.proto
```diff
@@ -36,6 +36,9 @@ message Node {
 enum NodeType {
     ADMIN = 0;
     STORAGE = 1;
+    INVALID = 2;
```
egonelbre: Invalid should be `0`, otherwise you won't see where it's unassigned.

pkg/kademlia/kademlia.go
```diff
@@ -260,6 +262,7 @@ func GetIntroNode(addr string) (*pb.Node, error) {
 			Transport: defaultTransport,
 			Address:   addr,
 		},
+		Type: pb.NodeType_ADMIN,
```
jenlij: This is the bootstrap node, I could change the nodetype name to reflect this

egonelbre: This probably should be `pb.NodeType_Unknown` or similar.

pkg/kademlia/peer_discovery.go
```diff
@@ -83,13 +83,17 @@ func (lookup *peerDiscovery) Run(ctx context.Context) (target *pb.Node, err erro
...
 				}
 				lookup.cond.L.Unlock()
-
-				neighbors, err := lookup.client.Lookup(ctx, *next, pb.Node{Id: lookup.target})
+				var nodeType pb.NodeType
+				if target != nil {
+					nodeType = target.Type
+					nodeType.PanicOnInvalid("Peer Discovery Run")
+				}
+				next.Type.PanicOnInvalid("next")
```
jenlij: @egonelbre it's panicking here during tests but I can't figure out why next.Type is nill

pkg/audit/verifier.go
```diff
@@ -61,7 +61,7 @@ func NewVerifier(transport transport.Client, overlay overlay.Client, id provider
 func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize, pieceNumber int,
 	id psclient.PieceID, pieceSize int64, fromNode *pb.Node, authorization *pb.SignedMessage) (s share, err error) {
 	defer mon.Task()(&ctx)(&err)
-
+	fromNode.Type.PanicOnInvalid("audit getShare")
```
jtolds: i would definitely vote against panics. also, in what scenario would a node even have an invalid type? is INVALID just if a node type is unset?

egonelbre: The panics are in place to find all the places where we haven't properly set the NodeType. They will be removed before final merge.

Yes, Invalid is to ensure that we don't leave node type empty.

jtolds: oh nice i just came back to reply to my own comment after i saw the todo in PanicOnInvalid. good work, carry on

pkg/pb/node.proto
```diff
@@ -34,8 +34,12 @@ message Node {
 
 // NodeType is an enum of possible node types
 enum NodeType {
-    ADMIN = 0;
-    STORAGE = 1;
+    INVALID = 0;
+    ADMIN = 1;
```
jtolds: ADMIN is the old name for SATELLITE. we don't need both.

do we ever need UNKNOWN? isn't UNKNOWN just anything above 4? I'd rather not have a specific UNKNOWN type.

pkg/provider/provider.go
```diff
@@ -31,17 +32,18 @@ type Responsibility interface {
...
 	grpc     *grpc.Server
 	next     []Responsibility
 	identity *FullIdentity
+	nodeType pb.NodeType
 }
 
 // NewProvider creates a Provider out of an Identity, a net.Listener, a UnaryInterceptorProvider and
 // a set of responsibilities.
-func NewProvider(opts *ServerOptions, lis net.Listener, interceptor grpc.UnaryServerInterceptor,
+func NewProvider(opts *ServerOptions, lis net.Listener, interceptor grpc.UnaryServerInterceptor, nodeType pb.NodeType,
```
jtolds: we can actually get away without changing the provider at all!

```
// StorageNodeConfig is a Config that implements provider.Responsibility as
// a storage node
type StorageNodeConfig Config

// Run implements provider.Responsibility
func (c StorageNodeConfig) Run(ctx context.Context, server *provider.Provider) error {
	return Config(c).Run(ctx, server, pb.NodeType_STORAGE)
}

// SatelliteConfig is a Config that implements provider.Responsibility as
// a satellite
type SatelliteConfig Config

// Run implements provider.Responsibility
func (c SatelliteConfig) Run(ctx context.Context, server *provider.Provider) error {
	return Config(c).Run(ctx, server, pb.NodeType_SATELLITE)
}

// Run does not implement provider.Responsibility. Please use a specific
// SatelliteConfig or StorageNodeConfig
func (c Config) Run(ctx context.Context, server *provider.Provider,
	nodeType pb.NodeType) (err error) {
```

then all you have to do is update commands to use kademlia.StorageNodeConfig or kademlia.SatelliteConfig instead of kademlia.Config, and everything in pkg/provider can be left alone

pkg/kademlia/kademlia.go
```diff
@@ -260,6 +260,9 @@ func GetIntroNode(addr string) (*pb.Node, error) {
 			Transport: defaultTransport,
 			Address:   addr,
 		},
+		// TODO: nodetype is an assumption for now, but we shouldn't need to know
+		// or care for bootstrapping
+		Type: pb.NodeType_SATELLITE,
```
egonelbre: We can use `pb.NodeType_UNKNOWN` for this.

pkg/kademlia/kademlia_test.go
```diff
@@ -99,7 +99,7 @@ func TestPeerDiscovery(t *testing.T) {
 	targetServer, _, targetID, targetAddress := startTestNodeServer(ctx)
 	defer targetServer.Stop()
 
-	bootstrapNodes := []pb.Node{{Id: bootID.ID, Address: &pb.NodeAddress{Address: bootAddress}}}
+	bootstrapNodes := []pb.Node{{Id: bootID.ID, Address: &pb.NodeAddress{Address: bootAddress}, Type: pb.NodeType_STORAGE}}
```
egonelbre: We can use `pb.NodeType_UNKNOWN` for this.

pkg/pb/node.proto
```diff
@@ -34,8 +34,10 @@ message Node {
 
 // NodeType is an enum of possible node types
 enum NodeType {
```
egonelbre: Let's have:
```
INVALID // to detect misconfigured peers
UNKNOWN // for connecting to things that we don't know what they are
SATELLITE
STORAGE
UPLINK
```

pkg/pb/utils.go
```diff
@@ -66,3 +67,11 @@ func CopyNode(src *Node) (dst *Node) {
 
 	return &node
 }
+
+// PanicOnInvalid panics if NodeType is invalid
+func (nt NodeType) PanicOnInvalid(from string) {
+	// TODO: Remove all references
+	if nt == NodeType_INVALID {
+		panic("INVALID NODE TYPE: " + from)
```
egonelbre: Let's maybe initially keep it logging (instead of panic) to find new issues. And before release let's remove these.

pkg/pb/node.proto
egonelbre: Yes, Unknown would be for connecting when we don't know what the other side supports. e.g. the bootstrap node.

pkg/provider/provider.go
egonelbre: This is nice. I forgot about this approach.

pkg/kademlia/kademlia.go
egonelbre: sure

pkg/pb/utils.go
egonelbre: Sure, let's use that for now.

pkg/kademlia/peer_discovery.go
```diff
@@ -83,13 +84,18 @@ func (lookup *peerDiscovery) Run(ctx context.Context) (target *pb.Node, err erro
...
 				}
 				lookup.cond.L.Unlock()
-
-				neighbors, err := lookup.client.Lookup(ctx, *next, pb.Node{Id: lookup.target})
+				var nodeType pb.NodeType
+				if target != nil {
+					nodeType = target.Type
+					nodeType.PanicOnInvalid("Peer Discovery Run")
+				}
+				fmt.Println(next)
```
egonelbre: Remove fmt.Println from here.

pkg/kademlia/routing.go
```diff
@@ -49,6 +49,9 @@ type RoutingTable struct {
 
 // NewRoutingTable returns a newly configured instance of a RoutingTable
 func NewRoutingTable(logger *zap.Logger, localNode pb.Node, kdb, ndb storage.KeyValueStore) (*RoutingTable, error) {
+
```
egonelbre: remove empty line

pkg/pb/utils.go
```diff
@@ -66,3 +71,11 @@ func CopyNode(src *Node) (dst *Node) {
 
 	return &node
 }
+
+// PanicOnInvalid panics if NodeType is invalid
+func (nt NodeType) PanicOnInvalid(from string) {
```
egonelbre: Let's rename this to `DPanic`

egonelbre: also update comment

PR #876, opened 2018-12-17
Payments API protobuf definition

mniewrzal:
https://storjlabs.atlassian.net/browse/V3-723


pkg/pb/payments.proto
```diff
@@ -0,0 +1,49 @@
...
+option go_package = "pb";
+
+// The service definition for the Payments API
+service Payments {
+    // Pay creates a payment to a single storage node
+    rpc Pay(PaymentRequest) returns (PaymentResponse);
+    // Calculate determines the outstanding balance for a given storage node
+    rpc Calculate(CalculateRequest) returns (CalculateResponse);
+    // AdjustPrices sets the prices paid by a satellite for data at rest and bandwidth
+    rpc AdjustPrices(AdjustPricesRequest) returns (AdjustPricesResponse);
```
egonelbre: I wonder, should this be part of the API or should that be changed by some other way?

mniewrzal: hmm I think it depends if this can be triggered automatically somewhere or it needs to be triggered on demand.

PR #877, opened 2018-12-17
Overlay Cache master db merge

mniewrzal:
Moves Overlay Cache db into master db.

https://storjlabs.atlassian.net/browse/V3-850


pkg/overlay/cache_test.go
```diff
@@ -100,40 +98,9 @@ func TestCache_Redis(t *testing.T) {
...
-	}
-	defer ctx.Check(client.Close)
-
-	testCache(ctx, t, client, sdb)
+	satellitedbtest.Run(t, func(t *testing.T, db *satellitedb.DB) {
+		testCache(ctx, t, db.OverlayCache(), sdb)
+	})
 }
 
 func TestCache_Store(t *testing.T) {
```
egonelbre: Shouldn't this be part of `satellitedb.NewInMemory` already?

egonelbre: Ah, it's using `teststore.New` here, probably not necessary to have it.

satellite/satellitedb/database.go
```diff
@@ -59,10 +60,10 @@ func (db *DB) StatDB() statdb.DB {
 	return &statDB{db: db.db}
 }
 
-// // OverlayCacheDB is a getter for OverlayCacheDB repository
-// func (db *DB) OverlayCacheDB() overlay.DB {
-// 	return &overlayCacheDB{db: db.db}
-// }
+// OverlayCache is a getter for overlay cache repository
+func (db *DB) OverlayCache() storage.KeyValueStore {
```
egonelbre: We want a better interface here than `storage.KeyValueStore`, but it's fine for now.

satellite/satellitedb/overlaycache.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// ReverseList lists all keys in revers order
+func (o *overlaycache) ReverseList(start storage.Key, limit int) (storage.Keys, error) {
+	return nil, nil
+}
+
+// Iterate iterates over items based on opts
+func (o *overlaycache) Iterate(opts storage.IterateOptions, fn func(storage.Iterator) error) error {
+	return nil
```
egonelbre: `return nil, errors.New("not implemented")`

satellite/satellitedb/overlaycache.go
```diff
@@ -0,0 +1,154 @@
...
+			keys = append(keys, decoded)
+		}
+	}
+
+	return keys, nil
+}
+
+// ReverseList lists all keys in revers order
+func (o *overlaycache) ReverseList(start storage.Key, limit int) (storage.Keys, error) {
+	return nil, nil
```
egonelbre: `return nil, errors.New("not implemented")`

satellite/satellitedb/overlaycache.go
```diff
@@ -0,0 +1,154 @@
...
+		for _, row := range rows {
+			decoded, err := hex.DecodeString(row.Key)
+			if err != nil {
+				return []storage.Key{}, err
+			}
+			keys = append(keys, decoded)
+		}
+	} else {
+		prefixValue := hex.EncodeToString(start)
+		rows, err := o.db.Query("SELECT key FROM overlay_cache_nodes n WHERE n.key LIKE $1 LIMIT $2", prefixValue+"%", limit)
```
egonelbre: Like is not the correct here, it should check for greater than rather than prefix.

egonelbre: Also let's avoid using raw queries, they break with different backends.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -137,3 +137,31 @@ read one (
...
   where  node.id = ?
 )
+
+//--- overlaycache ---//
+
+model overlay_cache_node (
+	key key
+	unique key
+
+	field key text
```
egonelbre: this should be a blob

satellite/satellitedb/overlaycache.go
mniewrzal: ah yes, I looked at wrong param inside redis `Iterate` method

satellite/satellitedb/overlaycache.go
```diff
@@ -0,0 +1,117 @@
...
+	return values, nil
+}
+
+func (o *overlaycache) Delete(key storage.Key) error {
+	_, err := o.db.Delete_OverlayCacheNode_By_Key(o.ctx, dbx.OverlayCacheNode_Key(key))
+	return err
+}
+
+func (o *overlaycache) List(start storage.Key, limit int) (keys storage.Keys, err error) {
+	rows, err := o.db.Limited_OverlayCacheNode_By_Key_GreaterOrEqual(o.ctx, dbx.OverlayCacheNode_Key(start), limit, 0)
```
egonelbre: limit should default to `storage.LookupLimit`

```
	if limit <= 0 || limit > LookupLimit {
		limit = LookupLimit
	}
```

PR #882, opened 2018-12-17
[Experiment] Use docker in Travis

egonelbre:


.travis.yml
```diff
@@ -73,6 +73,16 @@ matrix:
...
     
+    ### docker build test ###
+    - env: MODE=docker
+      services:
+        - docker
+      before_script:
+      before_install:
+      install:
+      script:
+        - make all-in-one
```
brimstone: This need something that actually passes or fails. `all-in-one` just starts the environment with a satellite, a storagenode, and a gateway.

egonelbre: I assumed it tests the setup?

brimstone: `all-in-one` does not test anything in a way I define test. It does stand up a local environment with production grade images.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN /app/satellite setup
+ENTRYPOINT ["/app/satellite", "run"]
+
+# Storage Node
+FROM alpine as storj-storagenode
+ENV SATELLITE_ADDR=
+EXPOSE 7776/udp 7777
+WORKDIR /app
+COPY --from=build-env /app/storagenode /app/storagenode
+RUN /app/storagenode setup
```
brimstone: So, every storagenode has the same identity?

egonelbre: Ah, right...

egonelbre:
Closing because fixed in https://github.com/storj/storj/pull/900

PR #886, opened 2018-12-17
Fix master

brimstone:


brimstone:
I think gospace has picked up a bad version. How do we flush gospace?

egonelbre:
@brimstone new hash is the correct one, it needs to use Go 1.11.4, it changed how has is calculated

brimstone:
That seems to be correct. Not everything was updated to 1.11.4 yet. This PR isn't needed.

PR #887, opened 2018-12-17
Refactor statdb

mobyvb:
https://storjlabs.atlassian.net/browse/V3-837

* Update statdb args/return values to minimize structs
* Simplify statdb.Update() to update all stats instead of an arbitrary subset determined by flags
* Remove CreateIfNotExists logic from statdb.Update()
* Simplify audit code structure

### DIFF
--- a/pkg/audit/reporter.go
+++ b/pkg/audit/reporter.go
@@ -13,13 +13,20 @@ import (
 )
 
 type reporter interface {
-	RecordAudits(ctx context.Context, failedNodes []*statdb.UpdateRequest) (err error)
+	RecordAudits(ctx context.Context, req *RecordAuditsInfo) (failed *RecordAuditsInfo, err error)
 }
 
 // Reporter records audit reports in statdb and implements the reporter interface
 type Reporter struct {
 	statdb     statdb.DB
 	maxRetries int
+}
+
+// RecordAuditsInfo is a struct containing arguments/return values for RecordAudits()
+type RecordAuditsInfo struct {
+	SuccessNodeIDs storj.NodeIDList
+	FailNodeIDs    storj.NodeIDList
+	OfflineNodeIDs storj.NodeIDList
 }
 
 // NewReporter instantiates a reporter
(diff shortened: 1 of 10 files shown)


satellite/satellitedb/statdb.go
```diff
@@ -31,7 +31,8 @@ type statDB struct {
 }
 
 // Create a db entry for the provided storagenode
-func (s *statDB) Create(ctx context.Context, createReq *statdb.CreateRequest) (resp *statdb.CreateResponse, err error) {
+func (s *statDB) Create(ctx context.Context, nodeID storj.NodeID,
+	startingStats *pb.NodeStats) (stats *pb.NodeStats, err error) {
```
egonelbre: nit: single line

pkg/audit/reporter.go
```diff
@@ -22,6 +22,13 @@ type Reporter struct {
 	maxRetries int
 }
 
+// RecordAuditsInfo is a struct containing arguments/return values for RecordAudits()
+type RecordAuditsInfo struct {
```
egonelbre: Let's also use a separate type for `NodeStats` and add a converter if necessary.

pkg/statdb/statdb_test.go
```diff
@@ -17,9 +17,9 @@ import (
 	"storj.io/storj/satellite/satellitedb/satellitedbtest"
 )
 
-func getRatio(s, t int) (success, total int64, ratio float64) {
+func getRatio(s, t int64) (success, total int64, ratio float64) {
```
egonelbre: nit: let's use longer names instead of `s`, `t`...

pkg/statdb/statdb_test.go
```diff
@@ -45,34 +50,28 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
...
 		assert.NoError(t, err)
-		s := resp.Stats
 		assert.EqualValues(t, auditSuccessRatio, s.AuditSuccessRatio)
 		assert.EqualValues(t, uptimeRatio, s.UptimeRatio)
 
-		getReq := &statdb.GetRequest{
-			Node: nodeID,
-		}
-		getResp, err := sdb.Get(ctx, getReq)
+		s, err = sdb.Get(ctx, nodeID)
```
egonelbre: nit: let's use longer names instead of `s`..., e.g. `stats`

pkg/statdb/statdb_test.go
```diff
@@ -125,131 +117,111 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
...
-		auditSuccessCount, auditCount, auditSuccessRatio := getRatio(4, 10)
-		uptimeSuccessCount, uptimeCount, uptimeRatio := getRatio(8, 25)
+		auditSuccessCount, auditCount, auditSuccessRatio := getRatio(currAuditSuccess, currAuditCount)
+		uptimeSuccessCount, uptimeCount, uptimeRatio := getRatio(currUptimeSuccess, currUptimeCount)
 
-		getReq := &statdb.GetRequest{
-			Node: nodeID,
-		}
-		getResp, err := sdb.Get(ctx, getReq)
+		s, err := sdb.Get(ctx, nodeID)
```
egonelbre: nit: let's use longer names instead of `s`..., e.g. `stats`

pkg/statdb/statdb_test.go
```diff
@@ -125,131 +117,111 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
...
-		auditSuccessCount, auditCount, auditSuccessRatio := getRatio(5, 11)
-		uptimeSuccessCount, uptimeCount, uptimeRatio := getRatio(8, 26)
+		auditSuccessCount, auditCount, auditSuccessRatio := getRatio(currAuditSuccess, currAuditCount)
+		uptimeSuccessCount, uptimeCount, uptimeRatio := getRatio(currUptimeSuccess, currUptimeCount)
 
-		getReq := &statdb.GetRequest{
-			Node: nodeID,
-		}
-		getResp, err := sdb.Get(ctx, getReq)
+		s, err := sdb.Get(ctx, nodeID)
```
egonelbre: nit: let's use longer names instead of `s`..., e.g. `stats`

pkg/statdb/statdb_test.go
```diff
@@ -269,13 +241,9 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
 			UptimeCount:        uptimeCount1,
 			UptimeRatio:        uptimeRatio1,
 		}
-		createReq := &statdb.CreateRequest{
-			Node:  nodeID1,
-			Stats: nodeStats,
-		}
-		resp, err := sdb.Create(ctx, createReq)
+
+		s, err := sdb.Create(ctx, nodeID1, nodeStats)
```
egonelbre: nit: let's use longer names instead of `s`..., e.g. `stats`

pkg/audit/reporter.go
coyle: curious as to the reason why

egonelbre: Separating the layers. protobuf is an implementation detail not the core data-structures of the system. Using protobuf in the underlying layers makes codes somewhat harder to test and read.

When you look at the layers there are:

```
database <- ... <- service <- endpoint <- transport
```

protobuf, grpc etc. are all about transport, not the protocol semantics. By depending directly on protobuf this separation less clearer and hence the whole structure less clearer.

PR #889, opened 2018-12-17
Node Lifecycle Hooks

dylanlott:


pkg/kademlia/routing.go
```diff
@@ -227,3 +229,21 @@ func (rt *RoutingTable) GetBucketTimestamp(bIDBytes []byte, bucket dht.Bucket) (
...
+	err2 := rt.ConnectionFailed(node)
+	if err2 != nil {
+		zap.L().Debug(fmt.Sprintf("error removing node from routing table: %+v : %+v", err, err2))
+	}
+	return
+}
+
+// ConnSuccess implements the Transport success function
+func (rt RoutingTable) ConnSuccess(node *pb.Node) {
+	err := rt.ConnectionSuccess(node)
```
jenlij: What happened to the AddNode method here?

pkg/discovery/lifecycle.go
```diff
@@ -0,0 +1,46 @@
...
+// GracefulDisconnect is called when a node alerts the network they're
+// going offline for a short period of time with intent to come back
+func (d *Discovery) GracefulDisconnect(id storj.NodeID) {
+}
+
+// ConnFailure implements the Transport failure function
+func (d *Discovery) ConnFailure(node *pb.Node, err error) {
+}
+
+// ConnSuccess implements the Transport success function
```
wthorp: //ConnSuccess implements the transport.*Observer* success function
right?  transport tells me very little compared to observer.  There are number of places where this comment appears.

dylanlott: I'll update this comment to be more detailed, but you're correct. It implements the transport.Observer `ConnSuccess` function.

pkg/discovery/lifecycle.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+
+	"go.uber.org/zap"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+// ConnectionSuccess is called when a node is connected to
+func (d *Discovery) ConnectionSuccess(address string, id storj.NodeID) {
+	err := d.cache.Put(context.Background(), id, pb.Node{
```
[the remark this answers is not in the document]
dylanlott: I'll look into this more, I actually don't know if it's possible or worth it

pkg/discovery/lifecycle.go
```diff
@@ -0,0 +1,46 @@
...
+		Address: &pb.NodeAddress{
+			Address: address,
+		},
+		Id: id,
+	})
+	if err != nil {
+		zap.S().Error("error adding node to cache", zap.Error(err))
```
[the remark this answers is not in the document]
dylanlott: Probably should

pkg/overlay/client.go
```diff
@@ -51,7 +51,7 @@ type Options struct {
 
 // NewOverlayClient returns a new intialized Overlay Client
 func NewOverlayClient(identity *provider.FullIdentity, address string) (Client, error) {
-	tc := transport.NewClient(identity)
+	tc := transport.NewClient(identity, &Cache{}) // add overlay to transport client as observer
```
egonelbre: Why is it adding an empty cache to the client?

pkg/discovery/lifecycle.go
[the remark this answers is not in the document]
mobyvb: As far as I know we should pretty much never use `context.Background()` if it can be avoided. It also looks like you are just passing in `ctx` in `ConnectionFailure`. If it doesn't result in issues, I say don't use `context.Background()`

pkg/overlay/cache.go
```diff
@@ -127,3 +128,16 @@ func (o *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node) err
...
 }
+
+// ConnFailure implements the Transport Observer `ConnFailure` function
+func (o *Cache) ConnFailure(ctx context.Context, node *pb.Node, err error) {
+	// noop until we figure out what to do on ConnFailure
+}
+
+// ConnSuccess implements the Transport Observer `ConnSuccess` function
+func (o *Cache) ConnSuccess(ctx context.Context, node *pb.Node) {
+	err := o.Put(ctx, node.Id, *node)
```
mobyvb: why do you call `cache.Put()` here if you are also calling it from `discovery.ConnectionSuccess()`?

pkg/discovery/lifecycle.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// ConnectionFailure gets called when a node fails to be connected to
+func (d *Discovery) ConnectionFailure(ctx context.Context, id storj.NodeID) {
+	err := d.cache.Put(ctx, id, pb.Node{})
```
mobyvb: When are these (`ConnectionFailure`/`ConnectionSuccess`) called as opposed to the observer versions of these functions?

dylanlott:
@mobyvb ConnectionSuccess doesn't currently get called, I'm going to remove that code to reduce confusion.

dylanlott:
@egonelbre We pass an empty Cache to the Transport Client because it just needs a struct that will fulfill the Observer interface that we defined in Transport.

PR #894, opened 2018-12-18
Implement memory and file pipe

egonelbre:


internal/sync2/filepipe.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+type filePipeReader struct{ pipe *filepipe }
+type filePipeWriter struct{ pipe *filepipe }
+
+// Close implements io.Reader Close
+func (reader filePipeReader) Close() error { return reader.CloseWithError(io.EOF) }
+
+// Close implements io.Writer Close
+func (writer filePipeWriter) Close() error { return writer.CloseWithError(io.EOF) }
```
kaloyan-raev: Shouldn't we close the writer with `io.ErrClosedPipe` instead of `io.EOF`? The latter is supposed to be returned only be Read operations.

internal/sync2/filepipe.go
```diff
@@ -0,0 +1,275 @@
...
+	// has the reader finished?
+	if pipe.readerDone {
+		pipe.mu.Unlock()
+		return 0, pipe.readerErr
+	}
+
+	// have we closed already
+	if pipe.writerDone {
+		pipe.mu.Unlock()
+		return 0, os.ErrClosed
```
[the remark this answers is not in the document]
egonelbre: Yes it is.

internal/sync2/filepipe.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// check how much do they want to write
+	canWrite := pipe.limit - pipe.write
+
+	// no more room to write
+	if canWrite == 0 {
+		pipe.mu.Unlock()
+		return 0, io.EOF
```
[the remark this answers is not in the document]
egonelbre: In this location it has hit the write limit, so `io.EOF` seems appropriate.

internal/sync2/filepipe.go
egonelbre: Yes

internal/sync2/pipe.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		pipe.limit = blockSize
+		pipe.nodata.L = &pipe.mu
+	}
+
+	return multipipe, nil
+}
+
+// NewMultiPipeMemory returns a new MultiPipe that is created in tempdir
+// if tempdir == "" the fill will be created it into os.TempDir
```
kaloyan-raev: This comment needs to be corrected. There is no tempdir required for the memory pipe.

internal/sync2/pipe.go
```diff
@@ -0,0 +1,280 @@
...
+	if err == nil && done {
+		err = io.EOF
+	}
+	return readAmount, err
+}
+
+// MultiPipe is a multipipe backed by a single file
+type MultiPipe struct {
+	pipes    []pipe
+	refcount int64
```
mniewrzal: What's the purpose of this field? I'm not sure if I'm missing something.

mniewrzal: In general I know what its doing but I'm not sure about its meaning in `MultiPipie`. Its just set to `blockCount` value and used to create `offsetFile`.

egonelbre: The meaning in `MultiPipe` is the number of pipes that are still open (starts with the number of pipes specified). The meaning in `pipe` is the number of half-pipes that are open (starts with 2).

Maybe renaming the `blockCount` to `pipeCount` and `blockSize` to `pipeSize` would make it clearer?

kaloyan-raev: All pipes in the multi pipe share the same underlying resource (i.e. file). This is why the refcount is also shared between all the pipes.

internal/sync2/pipe.go
```diff
@@ -0,0 +1,280 @@
...
+		closeErr := tempfile.Close()
+		if closeErr != nil {
+			return nil, fmt.Errorf("%v/%v", err, closeErr)
+		}
+		return nil, err
+	}
+
+	multipipe := &MultiPipe{
+		pipes:    make([]pipe, blockCount),
+		refcount: blockCount,
```
kaloyan-raev: @egonelbre Shouldn't this be `2 * blockCount`, i.e. 2 halves for each pipe?

internal/sync2/pipe.go
egonelbre: Also maybe rename `refcount` to `open` or `openPipes`.

internal/sync2/pipe.go
egonelbre: No, because the refcount counts open pipes. So `pipe` has it's own refcount for tracking either side of the pipe. i.e. you need two closes in `pipe` for `multipipe.refcount` to be decreased.

internal/sync2/pipe.go
mniewrzal: Ok, thanks for clarifications. Yes, using `pipies` in names should make it clearer.

internal/sync2/pipe.go
kaloyan-raev: Oh, yes. I am confusing the different refcounts.

PR #895, opened 2018-12-18
Fix overlay commands after master db merge

mniewrzal:
Changes:
* Uses master db for overlay commands
* `list` command uses tab writer for displaying results
* overlay cache db new method (select with limit) for `List` method and `start` param equal `nil`


cmd/overlay/main.go
```diff
@@ -19,7 +22,6 @@ import (
 
 var (
 	// Error is the error class for overlays
```
egonelbre: remove comment as well

PR #897, opened 2018-12-18
Validate farmers wallet and email configuration

mniewrzal:
Adds farmer wallet and mail validation during storage node startup.

image

https://storjlabs.atlassian.net/browse/V3-994


kaloyan-raev:
Perhaps it would be worth printing the email and wallet addresses as INFO logs.

PR #898, opened 2018-12-18
KBucket clean up. V3-923

aligeti:
https://storjlabs.atlassian.net/browse/V3-923


pkg/kademlia/routing.go
```diff
@@ -91,36 +91,19 @@ func (rt *RoutingTable) CacheSize() int {
 
 // GetBucket retrieves the corresponding kbucket from node id
 // Note: id doesn't need to be stored at time of search
-func (rt *RoutingTable) GetBucket(id storj.NodeID) (bucket dht.Bucket, ok bool) {
+func (rt *RoutingTable) GetBucket(id storj.NodeID) ([]*pb.Node, bool) {
```
coyle: Why change this to return a `[]*pb.Node` instead of a Bucket type?

aligeti: the ticket is to clean up the bucket interface as it is not used

coyle: No, the bucket interface is still around and being used. The implementation at pkg/kademlia/bucket.go was not being used

coyle: https://github.com/storj/storj/pull/898/files#diff-10d55b8cd7e4b1ce8e2c34ca4f5d2ea9L49

coyle: I'm ok with getting rid of the interface but it's still in the code and `GetBucket` seems like it should return a bucket type. Maybe change the name to be `GetNodes` ?

pkg/kademlia/routing.go
```diff
@@ -89,38 +89,21 @@ func (rt *RoutingTable) CacheSize() int {
 	return rt.rcBucketSize
 }
 
-// GetBucket retrieves the corresponding kbucket from node id
+// GetNodes retrieves the corresponding kbucket from node id
```
jenlij: maybe update the comment here to say "retrieves nodes within the same kbucket as the given node id"?

pkg/kademlia/routing_test.go
```diff
@@ -47,6 +47,10 @@ func TestGetBucket(t *testing.T) {
 	assert.True(t, ok)
 	assert.NoError(t, err)
 
+	type KBucket struct {
```
jenlij: do we still need this struct?

pkg/kademlia/routing.go
aligeti: yes. Made the change

pkg/kademlia/routing_test.go
aligeti: Done..

PR #899, opened 2018-12-18
Use fixed logger for Kademlia and make not being able to connect in discovery debug statement.

egonelbre:


pkg/kademlia/kademlia_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
```
coyle: this is awesome!

PR #900, opened 2018-12-18
Test all-in-one

brimstone:
This test production grade images.


egonelbre:
In principle not against this, but worried this will push the potential merge time at 30min after a commit. (due to travis job queueing)

Other things:
1. prebuild packages as a separate stage in dockerfiles - should cut a bunch off from the total time. (e.g. `go build -race storj.io/storj/pkg/...` before running the specific build for a binary)
2. build with race checker

docker-compose.yaml
```diff
@@ -32,16 +33,18 @@ services:
...
     - postgres
   storagenode:
-    image: storjlabs/storagenode:${VERSION}
+    image: storjlabs/storagenode:${VERSION:-latest}
     environment:
     - SATELLITE_ADDR=satellite:7777
-    - STORJ_LOG_LEVEL=debug
     - STORJ_IDENTITY_SERVER_ADDRESS=storagenode:7777
+    - STORJ_KADEMLIA_FARMER_EMAIL=hello@storj.io
+    - STORJ_KADEMLIA_FARMER_WALLET=0x0000000000000000000000000000000000000000
```
egonelbre: These should now be:
```
STORJ_KADEMLIA_OPERATOR_EMAIL
STORJ_KADEMLIA_OPERATOR_WALLET
```

brimstone:
I'm afraid this is going to continue to fail until the satellite stops returning itself in the list of possible nodes.

PR #904, opened 2018-12-18
Process for updating storage node stats in kademlia

aleitner:
https://storjlabs.atlassian.net/browse/V3-702


pkg/kademlia/routing.go
```diff
@@ -166,6 +166,13 @@ func (rt *RoutingTable) ConnectionSuccess(node *pb.Node) error {
 		return nil
 	}
 
+	// Add self to routing table
```
egonelbre: This should be correctly setup already at routing table startup.

aleitner: yes My PR is still wip

aleitner: The purpose of this was because yourself could be removed from your own routing table

egonelbre: But it shouldn't modify `rt.self` in the process. `self` seems to be used throughout this package to mean "this is who I am"...

pkg/piecestore/psserver/config.go
```diff
@@ -0,0 +1,87 @@
...
+	"log"
+	"path/filepath"
+
+	"go.uber.org/zap"
+	"golang.org/x/net/context"
+	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+
+	"storj.io/storj/pkg/kademlia"
+	"storj.io/storj/pkg/pb"
+	as "storj.io/storj/pkg/piecestore/psserver/agreementsender"
```
egonelbre: Don't use name aliases.

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,80 @@
...
+// Error is a standard error class for this package.
+var (
+	Error                = errs.Class("kademlia bucket refresher error")
+	defaultCheckInterval = flag.Duration("piecestore.kbucket-refresher.check-interval", time.Hour, "number of seconds to sleep between updating the kademlia bucket")
+)
+
+// refreshService contains the information needed to run the bucket refresher service
+type refreshService struct {
+	logger *zap.Logger
+	rt     dht.RoutingTable
```
egonelbre: Use `*kademlia.RoutingTable` directly.

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,80 @@
...
+func newService(logger *zap.Logger, rt dht.RoutingTable, server *Server) *refreshService {
+	return &refreshService{
+		logger: logger,
+		rt:     rt,
+		server: server,
+	}
+}
+
+// Run runs the bucket refresher service
+func (service *refreshService) Run(ctx context.Context) (err error) {
```
egonelbre: Is this logic only storagenode specific or does satellite need to implement this as well, or does it implement it differently?

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Error is a standard error class for this package.
+var (
+	Error                = errs.Class("kademlia bucket refresher error")
+	defaultCheckInterval = flag.Duration("piecestore.kbucket-refresher.check-interval", time.Hour, "number of seconds to sleep between updating the kademlia bucket")
+)
+
+// refreshService contains the information needed to run the bucket refresher service
+type refreshService struct {
+	logger *zap.Logger
```
egonelbre: `log`

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,80 @@
...
+	"go.uber.org/zap"
+
+	"storj.io/storj/pkg/dht"
+	"storj.io/storj/pkg/pb"
+)
+
+// Error is a standard error class for this package.
+var (
+	Error                = errs.Class("kademlia bucket refresher error")
+	defaultCheckInterval = flag.Duration("piecestore.kbucket-refresher.check-interval", time.Hour, "number of seconds to sleep between updating the kademlia bucket")
```
egonelbre: This shouldn't be a global flag.

pkg/kademlia/routing.go
egonelbre: I.e. implement `(rt *RoutingTable) UpdateSelf(node *pb.Node)` or similar rather than doing it through `ConnectionSuccess`

jenlij: yes self is added in NewRoutingTable. If for some reason something tries to remove it, we should add that as a restriction in removeNode

pkg/piecestore/psserver/bucket_refresher.go
egonelbre: In this case the `dht.DHT` should eventually go away. But as a general guideline for using interfaces:

1. when there are multiple implementations (e.g. `storage.KeyValueStore`)
2. when there's an import cycle that cannot be broken (e.g. masterdb things)
3. when it makes sense to be compatible to extension (e.g. `io.Reader`, `io.Writer` in API-s)

All interfaces create an indirection which makes code harder to understand. At the moment `dht.DHT` is creating an indirection... so it's not critical to fix, but no reason to keep using it either.

pkg/piecestore/psserver/bucket_refresher.go
egonelbre: The log lines read a little bit nicer with `log`, e.g. `service.log.Debug(...` vs. `service.logger.Debug(...`

Yeah both are used, but prefer `log`.

pkg/piecestore/psserver/bucket_refresher.go
jenlij: gotcha that makes sense ty

pkg/piecestore/psserver/bucket_refresher.go
aleitner: done

pkg/piecestore/psserver/bucket_refresher.go
aleitner: done

pkg/piecestore/psserver/bucket_refresher.go
aleitner: done

pkg/piecestore/psserver/config.go
aleitner: done

pkg/kademlia/routing.go
aleitner: created the function. Need to write a test now

pkg/kademlia/routing_test.go
```diff
@@ -161,6 +161,43 @@ func TestConnectionSuccess(t *testing.T) {
 	}
 }
 
+func TestUpdateSelf(t *testing.T) {
+	id := teststorj.NodeIDFromString("AA")
+	rt, cleanup := createRoutingTable(t, id)
+	defer cleanup()
+	address := &pb.NodeAddress{Address: "a"}
+	node := &pb.Node{Id: id, Address: address}
```
jenlij: Let's add field `Type: pb.NodeType_STORAGE` to pb.Node{}. I'm working on making sure all references to pb.Node have a Type

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,78 @@
...
+
+	self := service.rt.Local()
+
+	self.Restrictions = &pb.NodeRestrictions{
+		FreeBandwidth: stats.AvailableBandwidth,
+		FreeDisk:      stats.AvailableSpace,
+	}
+
+	// Update the routing table with latest restrictions
+	// TODO (aleitner): Do we want to change the name of ConnectionSuccess?
```
jenlij: Maybe ConnectionSuccess should be named something like UpdateOrAddNode

pkg/piecestore/psserver/config.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	krt, ok := rt.(*kademlia.RoutingTable)
+	if !ok {
+		return ServerError.New("Could not convert dht.RoutingTable to *kademlia.RoutingTable")
+	}
+
+	refreshProcess := newService(zap.L(), c.KBucketRefreshInterval, krt, s)
+
+	go func() {
```
jenlij: Why do we need this in a go routine?

pkg/piecestore/psserver/config.go
```diff
@@ -0,0 +1,94 @@
...
+		if err := refreshProcess.Run(ctx); err != nil {
+			cancel()
+		}
+	}()
+
+	defer func() {
```
jenlij: What does this do?

pkg/piecestore/psserver/bucket_refresher.go
egonelbre: I think the TODO comment can be removed now.

pkg/kademlia/routing.go
```diff
@@ -140,6 +140,27 @@ func (rt *RoutingTable) FindNear(id storj.NodeID, limit int) (nodes []*pb.Node,
 	return nodes, nil
 }
 
+// UpdateSelf updates a node on the routing table
+func (rt *RoutingTable) UpdateSelf(node *pb.Node) error {
+	if node.Id != rt.Local().Id {
+		return RoutingErr.New("self does not have a matching node id")
+	}
+
+	rt.mutex.Lock()
```
coyle: curious, why not just keep the mutex ?

pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := service.rt.UpdateSelf(&self); err != nil {
```
coyle: So, i'm wondering if it's best to not pass in the self to be updated but instead the items that need to be updated ?

There's the case that 
A gets `self` updates field Z
B gets `self' updates field G 
and they both call `UpdateSelf`
there is no deterministic way to know what the value of self is.

pkg/kademlia/routing_test.go
aleitner: done

pkg/piecestore/psserver/bucket_refresher.go
aleitner: done

pkg/piecestore/psserver/config.go
aleitner: The refresh process has a for loop that runs forever

pkg/piecestore/psserver/config.go
aleitner: shutdown the storage node cleanly

pkg/kademlia/routing.go
aleitner: I think I'm missing something because the mutex lock is there

PR #905, opened 2018-12-18
adds remove function for cache

dylanlott:


pkg/overlay/cache.go
```diff
@@ -128,3 +131,18 @@ func (o *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node) err
 
 	return o.DB.Put(nodeID.Bytes(), data)
 }
+
+// Delete will remove the node from the cache. Used when a node hard disconnects or fails
+// to pass a PING multiple times.
+func (o *Cache) Delete(ctx context.Context, id storj.NodeID) error {
+	if id.IsZero() {
+		return ErrDelete
```
mniewrzal: Shouldn't this be `ErrEmptyNode`? Like in Get method.

pkg/overlay/cache_test.go
```diff
@@ -98,6 +98,13 @@ func testCache(ctx context.Context, t *testing.T, store storage.KeyValueStore, s
 			assert.Error(t, err)
 		}
 	}
+
+	{ // Delete
+		err := cache.Delete(ctx, valid1ID)
+		assert.NoError(t, err)
+		err = cache.Delete(ctx, storj.NodeID{})
+		assert.Error(t, err)
```
mniewrzal: Maybe also testing invalid node id?

pkg/overlay/cache.go
```diff
@@ -21,6 +21,9 @@ const (
 	OverlayBucket = "overlay"
 )
 
+// ErrDelete is returned when there was an error deleting something
+var ErrDelete = errs.New("error deleting key:")
```
mniewrzal: Why `:` here?

pkg/overlay/cache.go
```diff
@@ -128,3 +131,18 @@ func (o *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node) err
 
 	return o.DB.Put(nodeID.Bytes(), data)
 }
+
+// Delete will remove the node from the cache. Used when a node hard disconnects or fails
+// to pass a PING multiple times.
+func (o *Cache) Delete(ctx context.Context, id storj.NodeID) error {
```
egonelbre: Why would we want to remove things from cache? I mean we can flag it as problematic or something instead.

pkg/overlay/cache.go
```diff
@@ -128,3 +131,18 @@ func (o *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node) err
...
+	if id.IsZero() {
+		return ErrDelete
+	}
+
+	err := o.DB.Delete(id.Bytes())
+	if err != nil {
+		return ErrDelete
```
egonelbre: no need for the separate error class

pkg/overlay/cache.go
coyle: It's assumed that things in the cache are good to go. If we have a bad node in the cache it should be removed.

egonelbre: Hard disconnect doesn't mean it's a bad node. E.g. losing network for half an hour would mean that the whole cache could be wiped.

coyle: and then the cache would rebuild itself the next run. The assumption is everything in the cache was contacted successfully within the last hour. It's not a bad node if it's removed, it's a node we didn't successfully speak with.

dylanlott: To add on to this: There are definitely cases where we want to remove a node from the cache, but also we aren't going to remove them after a single failure to communicate. I think we had talked in the past about removing nodes after X failed attempts, or X amount of time being unresponsive. This is just adding the method to the cache for that removal.

pkg/overlay/cache.go
dylanlott: Fixed

pkg/overlay/cache_test.go
dylanlott: Added another test

pkg/overlay/cache.go
dylanlott: Fixed

pkg/overlay/cache_test.go
```diff
@@ -98,6 +98,20 @@ func testCache(ctx context.Context, t *testing.T, store storage.KeyValueStore, s
 			assert.Error(t, err)
 		}
 	}
+
+	{ // Delete
+		// Test standard delete
+		err := cache.Delete(ctx, valid1ID)
```
mobyvb: nit - also test that `cache.Get` returns an error for `valid1ID` after this assertion

mobyvb: and maybe that it does not return an error before this assertion

dylanlott: Test added 👍

PR #906, opened 2018-12-18
CSR service

bryanchriswhite:
_description forthcoming_


cmd/certficates/main.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	authExportCmd = &cobra.Command{
+		Use:   "export [<email>, ...]",
+		Short: "Export authorization(s) from CSR authorization DB to a CSV file (or stdout)",
+		RunE:  cmdExportAuth,
+	}
+
+	setupCfg struct {
+		Overwrite bool `default:"false"`
```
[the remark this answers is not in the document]
bryanchriswhite: also, fix `--overwrite` to delete the old db file

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,270 @@
...
+// NewAuthorization creates a new, unclaimed authorization with a random token value
+func NewAuthorization() (*Authorization, error) {
+	var token Token
+	i, err := rand.Read(token[:])
+	if err != nil {
+		return nil, ErrAuthorization.Wrap(err)
+	}
+	if i != tokenLength {
+		return nil, ErrNotEnoughRandom
+	}
```
bryanchriswhite: is this a legit concern?

cmd/certficates/main.go
```diff
@@ -0,0 +1,342 @@
...
+	runCmd = &cobra.Command{
+		Use:   "run",
+		Short: "Run a certificate signing server",
+		RunE:  cmdRun,
+	}
+
+	authCmd = &cobra.Command{
+		Use:   "auth",
+		Short: "CSR authorization management",
+	}
```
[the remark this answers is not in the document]
coyle: maybe but I would be ok letting it be for now and moving later.

cmd/certficates/main.go
```diff
@@ -0,0 +1,341 @@
...
+		emails = strings.Split(string(list), authInfoCfg.Delimiter)
+	}
+
+	var emailErrs, printErrs utils.ErrorGroup
+	w := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+	if _, err := fmt.Fprintln(w, "Email\tClaimed\tAvail.\t"); err != nil {
+		return err
+	}
+
+	for _, email := range emails {
```
coyle: maybe pull this out into its own function ? this is a really large function right now

pkg/certificates/certificates_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}()
+
+	var authErrs utils.ErrorGroup
+	for i := 0; i < 5; i++ {
+		_, err := authDB.Create("user%s@example.com", 1)
+		if err != nil {
+			authErrs.Add(err)
+		}
+	}
+	if err = authErrs.Finish(); !assert.NoError(t, err) {
```
mobyvb: Why `!assert.NoError(t, err)` instead of `assert.Error(t, err)`?

mobyvb: I think you should split this into two separate lines instead of doing multiple different things in the if statement

pkg/certificates/certificates_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Fatal(err)
+	}
+	defer func() {
+		_ = authDB.Close()
+		ctx.Cleanup()
+	}()
+
+	var authErrs utils.ErrorGroup
+	for i := 0; i < 5; i++ {
+		_, err := authDB.Create("user%s@example.com", 1)
```
mobyvb: what does having the "%s" in the email do?

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,286 @@
...
+// certificate signature.
+type AuthorizationDB struct {
+	DB storage.KeyValueStore
+}
+
+// Authorizations is a slice of authorizations for convenient de/serialization
+// and grouping.
+type Authorizations []*Authorization
+
+// Authorization represents a single-use authorization token and it's status
```
mobyvb: nit typo - `it's` -> `its`

pkg/certificates/certificates.go
mobyvb: From https://golang.org/pkg/crypto/rand/:
"Read is a helper function that calls Reader.Read using io.ReadFull. On return, n == len(b) if and only if err == nil. "

In other words, it does not seem like this is a concern since `i==tokenLength` if there is no error

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,286 @@
...
+	defer mon.Task()(&ctx)(&err)
+
+	srv := NewServer(zap.L())
+	pb.RegisterCertificatesServer(server.GRPC(), srv)
+
+	return server.Run(ctx)
+}
+
+// Sign signs a valid certificate signing request's cert.
+func (c CertificateSigner) Sign(ctx context.Context, req *pb.SigningRequest) (*pb.SigningResponse, error) {
```
mobyvb: Should I assume `Sign()` will be updated in another PR?

pkg/certificates/certificates_test.go
bryanchriswhite: This is supposed to be formatted with `i`.. will fix

pkg/certificates/certificates.go
bryanchriswhite: Correct

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,286 @@
...
+
+	return &Authorization{
+		Token: token,
+	}, nil
+}
+
+// NewAuthDB creates or opens the authorization database specified by the config
+func (c CertSignerConfig) NewAuthDB() (*AuthorizationDB, error) {
+	// TODO: refactor db selection logic?
+	driver, source, err := utils.SplitDBURL(c.AuthorizationDBURL)
```
aligeti: Can the database be using the masterdb as in the satellite/satellitedb interface? Like what we did for irreparabledb, statdb, bwagreements ..

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,286 @@
...
+	return ErrAuthorizationDB.Wrap(a.DB.Close())
+}
+
+// Create creates a new authorization and adds it to the authorization database.
+func (a *AuthorizationDB) Create(email string, count int) (Authorizations, error) {
+	if count < 1 {
+		return nil, ErrAuthorizationCount
+	}
+
+	existingAuths, err := a.Get(email)
```
aligeti: Should we check for email is not empty before trying to get ?  just thinking out loud, will there be an entry in db with an empty email address?

bryanchriswhite: good point

pkg/certificates/certificates.go
bryanchriswhite: As we discussed on slack, copied here for posterity:

Yes it can but i'm not familiar with that part of the code yet, so in the interest of expediency I'd rather hold off on that for now. I'm also not sure how this will be deployed with respect to a "master database server".

AFAIK, this is only something we're doing for the alpha network. We're really not interested in being in the business of certificate authorities and cert-signing. There may be use cases for this service with our platform in other specific use cases but not in our general, long-term use case.

pkg/certificates/certificates.go
bryanchriswhite: nice - thanks for checking... now I just feel lazy :p

pkg/certificates/certificates_test.go
bryanchriswhite: Agreed to split that line. To answer your question: I'm simultaneously making an assertion that there is no error while adding behavior in the event that there is. If I were to do `assert.Error(t, err)` instead, the assertion would always fail if the test were otherwise passing.
```
=== RUN   TestAuthorizationDB_Emails
--- FAIL: TestAuthorizationDB_Emails (0.15s)
    certificates_test.go:260: 
        	Error Trace:	certificates_test.go:260
        	Error:      	An error is expected but got nil.
        	Test:       	TestAuthorizationDB_Emails
FAIL
```

pkg/certificates/certificates.go
bryanchriswhite: +1

PR #908, opened 2018-12-19
Add the ability for Kademlia to do a refresh for each bucket

wthorp:
Find all bucket that we have not contacted nodes on in the last hour
find a random ID in each of those bucket spaces
Do a lookup on that ID
This will allow us to find nodes in the same neighborhood as that node
These nodes need to be put into the overlay routing table or the overlay cache (this is done everytime we do a lookup for a node automatically)


internal/testplanet/node.go
```diff
@@ -171,7 +171,7 @@ func (node *Node) initOverlay(planet *Planet) error {
 		return err
 	}
 
-	kad, err := kademlia.NewKademliaWithRoutingTable(node.Log.Named("kademlia"), node.Info, planet.nodeInfos, node.Identity, 5, routing)
+	kad, err := kademlia.NewKademliaWithRoutingTable(context.Background(), node.Log.Named("kademlia"), node.Info, planet.nodeInfos, node.Identity, 5, routing)
```
egonelbre: Constructors shouldn't take context as argument they make staging difficult. I.e. add a separate `func` e.g. `kad.StartRefresher(ctx)`, or `kad.Start(ctx)`.

wthorp: fixed

pkg/kademlia/kademlia.go
```diff
@@ -100,26 +102,41 @@ func NewKademliaWithRoutingTable(log *zap.Logger, self pb.Node, bootstrapNodes [
...
 		return nil, BootstrapErr.Wrap(err)
 	}
-
 	k.nodeClient = nc
-
 	return k, nil
 }
 
+//StartRefresh occasionally refreshes stale kad buckets
+func (k *Kademlia) StartRefresh(ctx context.Context) {
```
egonelbre: Move this closer to `refresh(...)`, it makes it somewhat easier to read.

pkg/kademlia/kademlia.go
```diff
@@ -272,7 +278,6 @@ func (k *Kademlia) ListenAndServe() error {
 		return err
 	}
 	defer grpcServer.Stop()
```
egonelbre: There's no reason this should be deferred.

pkg/kademlia/kademlia.go
```diff
@@ -301,6 +305,66 @@ func GetIntroNode(addr string) (*pb.Node, error) {
...
+	bIDs, err := k.routingTable.GetBucketIds()
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	now := time.Now()
+	startID := bucketID{}
+	for _, bID := range bIDs {
+		ts, tErr := k.routingTable.GetBucketTimestamp(bID)
+		if tErr != nil {
+			err = errs.Combine(tErr)
```
egonelbre: This doesn't actually do anything too useful. Maybe use `errs.Group` instead, so:

```
var errors errs.Group

for ... {
    ts, terr := ...
    if terr != nil {
       errors.Add(terr)
       continue
    }
...

return errors.Err()
```

pkg/kademlia/kademlia.go
wthorp: Agreed, although my question is now "why is kademlia.ListenAndServe() never called in our codebase?"

egonelbre: Ah, because it shoudn't be called... i.e. the listener should be supplied from outside always. So feel free to remove the whole func.

pkg/kademlia/kademlia_test.go
```diff
@@ -450,6 +488,39 @@ func newTestServer(nn []*pb.Node) (*grpc.Server, *mockNodesServer) {
 	return grpcServer, mn
 }
 
+// TestRandomIds makes sure finds a random node ID is within a range (start..end]
+func TestRandomIds(t *testing.T) {
+	for x := 0; x < 1000; x++ {
```
mobyvb: how long does it take to run this test 1000 times? I wonder if there is a better way to test this

pkg/kademlia/kademlia.go
```diff
@@ -301,6 +262,86 @@ func GetIntroNode(addr string) (*pb.Node, error) {
...
+	}
+	return Error.Wrap(errors.Err())
+}
+
+// randomIDInRange finds a random node ID with a range (start..end]
+func randomIDInRange(start, end bucketID) (storj.NodeID, error) {
+	randID := storj.NodeID{}
+	divergedHigh := false
+	divergedLow := false
+	for x := 0; x < 32; x++ {
```
mobyvb: 32 is a magic number here. Maybe replace it with `len(bucketID)`

pkg/kademlia/kademlia_test.go
mobyvb: One thought I had would be selecting a predefined `start` and `end` rather than generating them randomly. And I'd select them so that there are only like 5-10 possibilities for random IDs between them. That way you could iterate fewer times and test more exhaustively.

It still isn't optimal but I am not sure how to exhaustively test something like this :/

pkg/kademlia/kademlia.go
```diff
@@ -222,60 +186,58 @@ func (k *Kademlia) Ping(ctx context.Context, node pb.Node) (pb.Node, error) {
...
+		concurrency: k.alpha, retries: defaultRetries, bootstrap: isBootstrap, bootstrapNodes: k.bootstrapNodes,
 	})
-
 	target, err := lookup.Run(ctx)
 	if err != nil {
 		return pb.Node{}, err
 	}
+	bucket, err := k.routingTable.getKBucketID(ID)
+	if err != nil {
+		k.log.Warn("Error getting getKBucketID in kad lookup")
```
mobyvb: why is this error not returned? (1)

pkg/kademlia/kademlia.go
```diff
@@ -222,60 +186,58 @@ func (k *Kademlia) Ping(ctx context.Context, node pb.Node) (pb.Node, error) {
...
+	} else {
+		err = k.routingTable.SetBucketTimestamp(bucket[:], time.Now())
+		if err != nil {
+			k.log.Warn("Error updating bucket timestamp in kad lookup")
```
mobyvb: why is this error not returned? (2)

pkg/kademlia/kademlia_test.go
wthorp: it take a fraction of a second.  it'd be prohibitive to test exhaustively, but I've run it several million times now.  its true that this is not a unit test, its a more of a fuzz / property-based test.

pkg/kademlia/kademlia.go
wthorp: I guess the question is, what benefit would we get by propagating it up?  It's never supposed to happen, but the worst case effect is that we do extra work ever hour.  We definitely wouldn't want things to stop working because of this.  But we would want to know about it.  DPanic() might be a better choice?

pkg/kademlia/kademlia.go
wthorp: same as above.

pkg/kademlia/kademlia.go
mobyvb: No it's fine, and that makes sense. I was just wondering.

PR #909, opened 2018-12-19
Improve logic for cutting the long tail during upload

kaloyan-raev:
Fixes https://storjlabs.atlassian.net/browse/V3-441

The infectious library now allows encoding pieces independently: https://github.com/vivint/infectious/pull/9. Thus `EncodeReader` will now buffer the unencoded segment data to a temp file. Each of the pieces readers will lookup stripes from the temp file to encode the respective erasure share.

This allows us to measure the time for reaching the repair threshold without having the piece readers to synchronize on the erasure encoding. Then we launch a 1.5x timer for reaching the success threshold. Slow nodes are cut if either the success threshold is reached or the timer expires, whichever comes first.


internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"sync"
+)
+
+// MemoryBuffer is a synchronized variable-sized memory buffer. Read will
+// block until new data is added or the buffer is closed. If the buffer is
+// closed, Read will continue reading as long as there is data in the buffer.
+type MemoryBuffer struct {
+	buf    *bytes.Buffer
+	cond   *sync.Cond
```
egonelbre: no need for this to be a pointer

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+	"bytes"
+	"io"
+	"sync"
+)
+
+// MemoryBuffer is a synchronized variable-sized memory buffer. Read will
+// block until new data is added or the buffer is closed. If the buffer is
+// closed, Read will continue reading as long as there is data in the buffer.
+type MemoryBuffer struct {
+	buf    *bytes.Buffer
```
egonelbre: no need for this to be a pointer

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+	cond   *sync.Cond
+	closed bool
```
egonelbre: and sync.Mutex to this struct to keep it close to cond

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return b.buf.Read(p)
+}
+
+// Write writes p to the buffer.
+func (b *MemoryBuffer) Write(p []byte) (n int, err error) {
+	b.cond.L.Lock()
+	defer b.cond.L.Unlock()
+	defer b.cond.Signal()
```
egonelbre: Use `Broadcast` to signal all readers... although it shouldn't matter in this case.

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+// Close closes the buffer.
+func (b *MemoryBuffer) Close() error {
+	return b.CloseWithError(nil)
+}
+
+// CloseWithError closes the buffer with error.
+func (b *MemoryBuffer) CloseWithError(err error) error {
+	b.cond.L.Lock()
+	defer b.cond.L.Unlock()
+	defer b.cond.Signal()
```
egonelbre: Use `Broadcast` to signal all readers... although it shouldn't matter in this case.

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"bytes"
+	"io"
+	"sync"
+)
+
+// MemoryBuffer is a synchronized variable-sized memory buffer. Read will
+// block until new data is added or the buffer is closed. If the buffer is
+// closed, Read will continue reading as long as there is data in the buffer.
+type MemoryBuffer struct {
```
egonelbre: Let's call it `MemoryPipe`... also you can actually use `filepipe` to implement it, since it relies on `ReadWriteCloserAt`. So if you implement a `type memoryWriterAt []byte` with `ReadAt`, `WriteAt` and `Close` then you can use it.

internal/sync2/membuf.go
```diff
@@ -0,0 +1,77 @@
...
+	return b.buf.Write(p)
+}
+
+// Close closes the buffer.
+func (b *MemoryBuffer) Close() error {
+	return b.CloseWithError(nil)
+}
+
+// CloseWithError closes the buffer with error.
+func (b *MemoryBuffer) CloseWithError(err error) error {
```
egonelbre: For `CloseWithError` I think the reader and writer should use different error. e.g. `writer.CloseWithError(err)` and then `writer.Close()` <-- should return `os.ErrClosed` not the error it was used to close.

internal/sync2/membuf.go
kaloyan-raev: Migrated to sync2.MultiPipe.

internal/sync2/copy_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"storj.io/storj/internal/memory"
+	"storj.io/storj/internal/sync2"
+)
+
+func TestCopy(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	r := io.LimitReader(rand.Reader, 32*memory.KB.Int64())
+	w := &bytes.Buffer{}
```
egonelbre: nit, use `ioutil.Discard` for writer

internal/sync2/copy_test.go
```diff
@@ -0,0 +1,43 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, n, 32*memory.KB.Int64())
+}
+
+func TestCopy_Cancel(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	r := io.LimitReader(rand.Reader, 32*memory.KB.Int64())
+	w := &bytes.Buffer{}
```
egonelbre: nit, use `ioutil.Discard` for writer

internal/sync2/io.go
```diff
@@ -75,7 +84,10 @@ func (file offsetFile) WriteAt(data []byte, at int64) (amount int, err error) {
 // Close implements io.Closer methods
 func (file offsetFile) Close() error {
 	if atomic.AddInt64(file.open, -1) == 0 {
-		return file.Close()
+		return utils.CombineErrors(
```
egonelbre: `errs.Combine`

internal/sync2/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	"sync/atomic"
+)
+
+type tee struct {
+	buffer ReadAtWriteAtCloser
+	open   *int64
+
+	mu      sync.Mutex
+	nodata  sync.Cond
+	newdata sync.Cond
```
egonelbre: maybe `noreader`, `needreader`?

internal/sync2/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	for tee.write > tee.maxRequired {
+		// are all readers already closed?
+		if atomic.LoadInt64(tee.open) <= 1 {
+			tee.mu.Unlock()
+			return 0, io.ErrClosedPipe
+		}
+		// wait until new data is required by any reader
+		tee.newdata.Wait()
```
egonelbre: I think this might be better if it first wrote to the buffer and then waited until a new request came in. This way there's less ping-pong between reader/writer.

pkg/eestream/rs_test.go
```diff
@@ -479,14 +420,14 @@ func testRSProblematic(t *testing.T, tt testCase, i int, fn problematicReadClose
...
-func readAll(readers []io.Reader) ([][]byte, error) {
+func readAll(readers []io.ReadCloser) ([][]byte, error) {
 	pieces := make([][]byte, len(readers))
 	errs := make(chan error, len(readers))
 	for i := range readers {
 		go func(i int) {
 			var err error
 			pieces[i], err = ioutil.ReadAll(readers[i])
-			errs <- err
+			errs <- utils.CombineErrors(err, readers[i].Close())
```
egonelbre: `errs.Combine`

internal/sync2/tee.go
```diff
@@ -0,0 +1,172 @@
...
+
+type tee struct {
+	buffer ReadAtWriteAtCloser
+	open   *int64
+
+	mu      sync.Mutex
+	nodata  sync.Cond
+	newdata sync.Cond
+
+	maxRequired int64
```
egonelbre: maybe `readtail`, `readLimit` or similar?

kaloyan-raev: It stores the maximum byte number requested by any of the readers so far. The writer checks it to see if it is necessary to add more data to the buffer.

So maybe a name like `readHead` instead of `readTail`. Or `maxReadHead` to denote that it is about the read head that is furthest ahead.

pkg/eestream/rs_test.go
kaloyan-raev: oh, yes. forgot it is already in the library.

internal/sync2/tee.go
kaloyan-raev: I don't see the difference with relation to the ping-pong. But changing it as you suggested would make the progress bar a little bit less realistic.

internal/sync2/tee.go
kaloyan-raev: `noreader` sounds good.

internal/sync2/tee.go
egonelbre: `maxRead`?

kaloyan-raev: Deal!

kaloyan-raev:
Part of this PR was extracted to #1022.

pkg/eestream/encode.go
```diff
@@ -98,260 +104,111 @@ func (rs *RedundancyStrategy) OptimalThreshold() int {
...
-		readers = append(readers, er.eps[i])
-	}
-	chanSize := mbm / (rs.TotalCount() * rs.ErasureShareSize())
-	if chanSize < 1 {
-		chanSize = 1
-	}
-	for i := 0; i < rs.TotalCount(); i++ {
-		er.eps[i].ch = make(chan block, chanSize)
+		rs:     rs,
+		pieces: make(map[int](*encodedPiece), rs.TotalCount()),
```
egonelbre: parenthesis can be removed

pkg/metainfo/kvmetainfo/buckets_test.go
```diff
@@ -351,17 +351,17 @@ func newDB(planet *testplanet.Planet) (*DB, error) {
 		return nil, err
 	}
 
-	rs, err := eestream.NewRedundancyStrategy(eestream.NewRSScheme(fc, int(1*memory.KB)), 3, 4)
+	rs, err := eestream.NewRedundancyStrategy(eestream.NewRSScheme(fc, 1*memory.KB.Int()), 3, 4)
```
egonelbre: Shouldn't all the `memory.KB` etc. now be `memory.KiB`?

kaloyan-raev: Yes, but let's fix them in another PR.

PR #913, opened 2018-12-20
add overwrite flag

littleskunk:


cmd/storagenode/main.go
```diff
@@ -112,6 +113,18 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
+	if !setupCfg.Overwrite && !valid {
+		fmt.Printf("storagenode configuration already exists (%v). rerun with --overwrite\n", setupDir)
+		return nil
+	} else if setupCfg.Overwrite && err == nil {
+		fmt.Println("overwriting existing satellite config")
+		err = os.RemoveAll(setupDir)
+		if err != nil {
+			return err
+		}
+	}
```
mobyvb: Check if err != nil

cmd/captplanet/setup.go
```diff
@@ -54,10 +54,9 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	valid, err := fpath.IsValidSetupDir(setupDir)
+	valid, _ := fpath.IsValidSetupDir(setupDir)
```
mniewrzal: Why are you removing `err`? Its used in `} else if setupCfg.Overwrite && err == nil {`

cmd/satellite/main.go
```diff
@@ -147,10 +147,9 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	valid, err := fpath.IsValidSetupDir(setupDir)
+	valid, _ := fpath.IsValidSetupDir(setupDir)
```
mniewrzal: Why are you removing `err`? Its used in `} else if setupCfg.Overwrite && err == nil {`

littleskunk:
Looks like we have another problem here. At the moment we whipe out a complete directory even if it doesn't contain a valid config.

mniewrzal:
Yes but I think it should be resolved as a separate issue.

littleskunk:
I have created a new issue for that: https://storjlabs.atlassian.net/browse/V3-1013

Now this PR should be ready for the final review.

PR #914, opened 2018-12-20
Cleanup overlay methods and names.

egonelbre:
* `overlay.NewOverlayCache` -> `overlay.NewCache`
* `overlay.NewOverlayClient` -> `overlay.NewClient`
* remove kademlia dependency


cmd/overlay/main.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2018 Storj Labs, Inc.
+// Copyright (C) 2018 Storj Labs, Incache.
```
kaloyan-raev: Something sneaked in here.

egonelbre: fixed

PR #915, opened 2018-12-20
Fix CA Timeout and show progress to the user

brimstone:
I know it's not really "progress", but keeping track of the highest difficulty seen so far gives that illusion to the user.

### DIFF
--- a/pkg/identity/certificate_authority.go
+++ b/pkg/identity/certificate_authority.go
@@ -11,6 +11,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"io/ioutil"
+	"log"
 
 	"github.com/zeebo/errs"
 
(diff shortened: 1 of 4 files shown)


pkg/identity/certificate_authority.go
```diff
@@ -178,16 +179,20 @@ func (pc PeerCAConfig) Load() (*PeerCertificateAuthority, error) {
...
 		opts.Concurrency = 1
 	}
 	ctx, cancel := context.WithCancel(ctx)
 
+	log.Printf("Generating a certificate matching a difficulty of %d\n", opts.Difficulty)
 	eC := make(chan error)
 	caC := make(chan FullCertificateAuthority, 1)
 	for i := 0; i < int(opts.Concurrency); i++ {
-		go newCAWorker(ctx, opts.Difficulty, opts.ParentCert, opts.ParentKey, caC, eC)
+		go newCAWorker(ctx, i, &highscore, opts.Difficulty, opts.ParentCert, opts.ParentKey, caC, eC)
```
egonelbre: Why is this highscore necessary?

pkg/identity/utils.go
```diff
@@ -12,8 +12,10 @@ import (
 	"encoding/asn1"
 	"encoding/pem"
 	"io/ioutil"
+	"log"
```
egonelbre: Pass in `zap.Logger` rather than relying on the global logger.

pkg/identity/certificate_authority.go
brimstone: It's to give the illusion of progress. It's pass in like this so each worker can know what the current "highscore" is and display the new one to the user.

pkg/identity/utils.go
brimstone: How? Logging is super weird in this project and I haven't figured it out yet. Is there a good example somewhere?

pkg/identity/certificate_authority.go
```diff
@@ -178,16 +179,20 @@ func (pc PeerCAConfig) Load() (*PeerCertificateAuthority, error) {
...
 	}
 
 	select {
```
egonelbre: Can we just use a for loop with delays and print `.`?

brimstone: That's not nearly as fun, and I've already figured out how to do it this way. ;)

egonelbre:
```
fmt.Print("generating certificate: ")
defer fmt.Println()
for {
	select {
	case <-ticker.C:
		fmt.Print(".")
	case ca := <-caC:
		return &ca, nil
	case err := <-eC:
		return nil, err
	case <-ctx.Done():
		return nil, ctx.Err()
	}
}
```

PR #916, opened 2018-12-20
delete and add projectmember api now accepts list of users

rikysya:
- addProjectMember api modified to accept list of users
- deleteProjectMember api modified to accept list of users

### DIFF
--- a/pkg/satellite/satelliteweb/satelliteql/mutation.go
+++ b/pkg/satellite/satelliteweb/satelliteql/mutation.go
@@ -7,6 +7,7 @@ import (
 	"github.com/graphql-go/graphql"
 	"github.com/skyrings/skyring-common/tools/uuid"
 	"github.com/zeebo/errs"
+
 	"storj.io/storj/pkg/satellite"
 	"storj.io/storj/pkg/utils"
 )
@@ -291,7 +292,7 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
 						userIDs = append(userIDs, id)
 					}
 
-					err := utils.CombineErrors(pErr, userErr.Err())
+					err := errs.Combine(pErr, userErr.Err())
 					if err != nil {
 						return nil, err
 					}
@@ -339,7 +340,7 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
 						userIDs = append(userIDs, id)
 					}
 
-					err := utils.CombineErrors(pErr, userErr.Err())
+					err := errs.Combine(pErr, userErr.Err())
 					if err != nil {
 						return nil, err
 					}
--- a/pkg/satellite/satelliteweb/satelliteql/mutation.go
+++ b/pkg/satellite/satelliteweb/satelliteql/mutation.go
@@ -269,24 +269,43 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
 						Type: graphql.NewNonNull(graphql.String),
 					},
 					fieldUserID: &graphql.ArgumentConfig{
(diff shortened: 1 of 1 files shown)


pkg/satellite/satelliteweb/satelliteql/mutation.go
```diff
@@ -269,24 +269,43 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
...
 					pID, _ := p.Args[fieldProjectID].(string)
-					uID, _ := p.Args[fieldUserID].(string)
+					uID, _ := p.Args[fieldUserID].([]interface{})
 
 					projectID, pErr := uuid.Parse(pID)
-					userID, uErr := uuid.Parse(uID)
 
-					err := utils.CombineErrors(pErr, uErr)
+					var userIDs []*uuid.UUID
+					var userErr error
```
egonelbre: use `errs.Group`

pkg/satellite/satelliteweb/satelliteql/mutation.go
```diff
@@ -269,24 +269,43 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
...
+						userIDs = append(userIDs, id)
+					}
+
+					err := utils.CombineErrors(pErr, userErr)
 					if err != nil {
 						return nil, err
 					}
 
-					err = service.AddProjectMember(p.Context, *projectID, *userID)
+					var addMemberErr error
```
egonelbre: use errs.Group

pkg/satellite/satelliteweb/satelliteql/mutation.go
```diff
@@ -269,24 +269,43 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
...
+					uID, _ := p.Args[fieldUserID].([]interface{})
 
 					projectID, pErr := uuid.Parse(pID)
-					userID, uErr := uuid.Parse(uID)
 
-					err := utils.CombineErrors(pErr, uErr)
+					var userIDs []*uuid.UUID
+					var userErr error
+
+					for _, userID := range uID {
```
egonelbre: In case of an error should it skip the return value or return null in that place? Not sure what the default way to handle is in graphql.

rikysya: Well in case of several uncorrelated errors or some half-completed process it would be descriptive to return result and error. But the problem is that graphql-go library doesn't parse the result if error is returned, so i guess i am forced to return nil if any error occurs

pkg/satellite/satelliteweb/satelliteql/mutation.go
```diff
@@ -269,24 +269,44 @@ func rootMutation(service *satellite.Service, types Types) *graphql.Object {
...
+						id, err := uuid.Parse(userID.(string))
+						if err != nil {
+							userErr.Add(err)
+							continue
+						}
+
+						userIDs = append(userIDs, id)
+					}
+
+					err := utils.CombineErrors(pErr, userErr.Err())
```
egonelbre: errs package also has `errs.Combine` now.

pkg/satellite/satelliteweb/satelliteql/mutation.go
```diff
@@ -6,7 +6,7 @@ package satelliteql
 import (
 	"github.com/graphql-go/graphql"
 	"github.com/skyrings/skyring-common/tools/uuid"
-
+	"github.com/zeebo/errs"
 	"storj.io/storj/pkg/satellite"
```
egonelbre: empty line between storj.io and external packages.

PR #920, opened 2018-12-20
Node lifecycle hooks

dylanlott:
Round 2 because the last PR got into a funky state.


internal/testplanet/node.go
```diff
@@ -175,12 +175,11 @@ func (node *Node) initOverlay(planet *Planet) error {
...
 		return utils.CombineErrors(err, routing.Close())
 	}
-	node.Kademlia = kad
 
+	node.Kademlia = kad
 	node.StatDB = node.Database.StatDB()
-
 	node.Overlay = overlay.NewCache(teststore.New(), node.StatDB)
-	node.Discovery = discovery.NewDiscovery(node.Overlay, node.Kademlia, node.StatDB)
+	node.Discovery = discovery.NewDiscovery(zap.L(), node.Overlay, node.Kademlia, node.StatDB)
```
wthorp: zap.L() -> node.Log.
Rule of thumb is that only the configs.go files should reference the global logger

pkg/kademlia/routing.go
```diff
@@ -202,3 +205,19 @@ func (rt *RoutingTable) GetBucketTimestamp(bIDBytes []byte) (time.Time, error) {
...
+	if err2 != nil {
+		zap.L().Debug(fmt.Sprintf("error with ConnFailure hook  %+v : %+v", err, err2))
+	}
+}
+
+// ConnSuccess implements the Transport success function
+func (rt RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
+	err := rt.ConnectionSuccess(node)
+	if err != nil {
+		zap.L().Debug("connection success error:", zap.Error(err))
```
wthorp: add logger to struct, call here

pkg/kademlia/routing.go
```diff
@@ -202,3 +205,19 @@ func (rt *RoutingTable) GetBucketTimestamp(bIDBytes []byte) (time.Time, error) {
 func (rt *RoutingTable) iterate(opts storage.IterateOptions, f func(it storage.Iterator) error) error {
 	return rt.nodeBucketDB.Iterate(opts, f)
 }
+
+// ConnFailure implements the Transport failure function
+func (rt RoutingTable) ConnFailure(ctx context.Context, node *pb.Node, err error) {
+	err2 := rt.ConnectionFailed(node)
+	if err2 != nil {
+		zap.L().Debug(fmt.Sprintf("error with ConnFailure hook  %+v : %+v", err, err2))
```
wthorp: add logger to struct, call here

pkg/overlay/cache.go
```diff
@@ -130,3 +131,16 @@ func (cache *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node)
...
+// ConnFailure implements the Transport Observer `ConnFailure` function
+func (cache *Cache) ConnFailure(ctx context.Context, node *pb.Node, err error) {
+	// noop until we figure out what to do on ConnFailure
+}
+
+// ConnSuccess implements the Transport Observer `ConnSuccess` function
+func (cache *Cache) ConnSuccess(ctx context.Context, node *pb.Node) {
+	err := cache.Put(ctx, node.Id, *node)
+	if err != nil {
+		zap.L().Debug("error putting node o cache:", zap.Error(err))
```
wthorp: I should say "add logger to struct, call here" but this is an interesting case.  You might also want to log some extra information about the context here, as knowing the class that created the error doesn't tell you a lot about where the error came from.

internal/testplanet/node.go
jenlij: i second that

pkg/discovery/lifecycle.go
```diff
@@ -0,0 +1,24 @@
...
+	"storj.io/storj/pkg/storj"
+)
+
+// GracefulDisconnect is called when a node alerts the network they're
+// going offline for a short period of time with intent to come back
+func (d *Discovery) GracefulDisconnect(id storj.NodeID) {
+}
+
+// ConnFailure implements the Transport Observer interface `ConnFailure` function
+func (d *Discovery) ConnFailure(ctx context.Context, node *pb.Node, err error) {
```
jenlij: You could probably still call these ConnectionFailure and ConnectionSuccess since they're in a different package

pkg/kademlia/routing.go
```diff
@@ -202,3 +205,19 @@ func (rt *RoutingTable) GetBucketTimestamp(bIDBytes []byte) (time.Time, error) {
 func (rt *RoutingTable) iterate(opts storage.IterateOptions, f func(it storage.Iterator) error) error {
 	return rt.nodeBucketDB.Iterate(opts, f)
 }
+
+// ConnFailure implements the Transport failure function
+func (rt RoutingTable) ConnFailure(ctx context.Context, node *pb.Node, err error) {
```
jenlij: Let's keep the receivers consistent and change (rt RoutingTable) to (rt *RoutingTable)

pkg/kademlia/routing.go
jenlij: Here too: Let's keep the receivers consistent and change (rt RoutingTable) to (rt *RoutingTable)

pkg/overlay/cache.go
```diff
@@ -130,3 +131,16 @@ func (cache *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node)
 
 	return cache.db.Put(nodeID.Bytes(), data)
 }
+
+// ConnFailure implements the Transport Observer `ConnFailure` function
+func (cache *Cache) ConnFailure(ctx context.Context, node *pb.Node, err error) {
+	// noop until we figure out what to do on ConnFailure
```
jenlij: maybe add TODO keyword in case we grep on that again in the future

pkg/transport/transport.go
```diff
@@ -24,6 +24,13 @@ var (
 	timeout = 20 * time.Second
 )
 
+// Observer implements the ConnSuccess and ConnFailure methods
+// for Discovery and other services to use
+type Observer interface {
+	ConnSuccess(ctx context.Context, node *pb.Node)
```
jenlij: again here you could be explicit and make it ConnectionSuccess - but these are nits so nw

pkg/overlay/cache.go
mobyvb: after `cache.Put`, could you add the following?
```
_, err := cache.statdb.UpdateUptime(ctx, node.Id, true)
if err != nil {
	// log statdb error
}
```

pkg/overlay/cache.go
mobyvb: Could you put the following code in here?

```
_, err := cache.statdb.UpdateUptime(ctx, node.Id, false)
if err != nil {
	// log err
}
```

pkg/overlay/client.go
```diff
@@ -51,7 +51,7 @@ type Options struct {
 
 // NewClient returns a new intialized Overlay Client
 func NewClient(identity *provider.FullIdentity, address string) (Client, error) {
-	tc := transport.NewClient(identity)
+	tc := transport.NewClient(identity, &Cache{}) // add overlay to transport client as observer
```
mobyvb: there is not a cache created elsewhere that we would want to use? If not, this is fine.

internal/testplanet/node.go
egonelbre: `node.Log.Named("discovery")`

pkg/discovery/config.go
```diff
@@ -51,7 +51,7 @@ func (c Config) Run(ctx context.Context, server *provider.Provider) (err error)
 	if !ok {
 		return Error.New("unable to get master db instance")
 	}
-	discovery := NewDiscovery(ol, kad, stat.StatDB())
+	discovery := NewDiscovery(zap.L(), ol, kad, stat.StatDB())
```
egonelbre: `zap.L().Named("discovery")`

pkg/kademlia/routing.go
dylanlott: That's going to change it to a pointer receiver, which will change the function signature of the Observer interface. I'm not sure I agree it should be changed.

pkg/kademlia/routing.go
dylanlott: Same point as above.

pkg/kademlia/routing.go
egonelbre: It wouldn't change the observer interface.

pkg/overlay/cache.go
dylanlott: You got it 👍

pkg/overlay/client.go
dylanlott: We dont' have access to a cache created elsewhere at this level, without getting import cycle errors. Besides, it only needs access to that function.

pkg/discovery/lifecycle.go
jenlij: Well If we can change the observer interface names, but just a suggestion not vital

pkg/kademlia/routing.go
dylanlott: Covered in PR #922

pkg/discovery/lifecycle.go
mobyvb: Oh yeah, I agree with changing Conn -> Connection across the board

pkg/overlay/cache.go
```diff
@@ -148,3 +149,23 @@ func (cache *Cache) Delete(ctx context.Context, id storj.NodeID) error {
...
+
+// ConnSuccess implements the Transport Observer `ConnSuccess` function
+func (cache *Cache) ConnSuccess(ctx context.Context, node *pb.Node) {
+	err := cache.Put(ctx, node.Id, *node)
+	if err != nil {
+		zap.L().Debug("error putting node o cache:", zap.Error(err))
+	}
+	_, err = cache.statDB.UpdateUptime(ctx, node.Id, true)
+	if err != nil {
+		zap.L().Debug("error updating statdDB with node connection info", zap.Error(err))
```
mobyvb: change to "error updating uptime for node in statDB"

pkg/overlay/cache.go
```diff
@@ -148,3 +149,23 @@ func (cache *Cache) Delete(ctx context.Context, id storj.NodeID) error {
 
 	return nil
 }
+
+// ConnFailure implements the Transport Observer `ConnFailure` function
+func (cache *Cache) ConnFailure(ctx context.Context, node *pb.Node, err error) {
+	// TODO: noop until we figure out what to do on ConnFailure
+	// Kademlia paper specifies 5 unsuccessful PINGs before removing the node
+	// from our routing table, but this is the cache so maybe we want to treat
+	// it differently.
```
mobyvb: We need to update uptime in statdb when we fail to connect to a node. Same as the other one, but with `false` instead of `true`

```
_, err := cache.statdb.UpdateUptime(ctx, node.Id, false)
if err != nil {
	// log statdb error
}
```

dylanlott: Added.

PR #921, opened 2018-12-20
adds delete functions and tests for cache

dylanlott:


pkg/overlay/cache.go
```diff
@@ -130,3 +133,18 @@ func (cache *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node)
 
 	return cache.db.Put(nodeID.Bytes(), data)
 }
+
+// Delete will remove the node from the cache. Used when a node hard disconnects or fails
```
wthorp: I think it's interesting that our business logic on this might change depending on the robustness of our transport layer.  This is what the kademlia paper says, BTW:

When a contact fails to respond to 5 RPCs in a row, it is considered stale. 
If a k-bucket is not full or its replacement cache is empty, Kademlia merely 
flags stale contacts rather than remove them. This ensures, among other things, 
that if a node's own network connection goes down temporarily, the node won't 
completely void all of its k-buckets. 

Would we have to re-bootstrap if our node's network connection went down?

dylanlott: Currently, we have no "retry" logic, as far as failed connection attempts. That's probably something we're going to need to store on the node. Might need to update the cache and node models to reflect a "retried" count.

dylanlott: We don't remove any nodes from the cache yet for failure to connect / reply. This is just a step in that direction, though.

PR #926, opened 2018-12-21
statreceiver nitpicking

egonelbre:
* Moved luacfg closer to statreceiver
* Converted example binary to an test example to ensure it's being run.
* Bunch of naming changes
* Few better error messages
* Fix default application path


cmd/statreceiver/copy.go
```diff
@@ -8,48 +8,48 @@ import (
...
+func NewPacketCopier(dest ...PacketDest) *PacketCopier {
+	return &PacketCopier{dest: dest}
 }
 
 // Packet implements the PacketDest interface
 func (p *PacketCopier) Packet(data []byte, ts time.Time) (ferr error) {
-	var errs utils.ErrorGroup
-	for _, d := range p.d {
-		errs.Add(d.Packet(data, ts))
+	var errs errs.Group
```
mobyvb: should this variable be named something besides `errs` since `errs` is the name of the zeebo package? I was confused at first when I saw `errs.Add`/`errs.Err` below

cmd/statreceiver/copy.go
```diff
@@ -8,48 +8,48 @@ import (
...
+	return &MetricCopier{dest: dest}
 }
 
 // Metric implements the MetricDest interface
 func (m *MetricCopier) Metric(application, instance string,
 	key []byte, val float64, ts time.Time) (ferr error) {
-	var errs utils.ErrorGroup
-	for _, d := range m.d {
-		errs.Add(d.Metric(application, instance, key, val, ts))
+	var errs errs.Group
```
mobyvb: variable name shouldn't be `errs`

egonelbre:
@jtolds it could be, but I'm not yet sure how and where yet, so not sure how to exactly organize it. At the moment it's mostly tied with statreceiver so organized it together with it. internal would also work, but yeah not sure. I also was thinking whether statreceiver should be in this repository at all, but yeah. I think it keeps core package cleaner at this moment.

PR #927, opened 2018-12-21
Add sleep for satellite startup

egonelbre:
https://storjlabs.atlassian.net/browse/V3-861

### DIFF
--- a/cmd/captplanet/run.go
+++ b/cmd/captplanet/run.go
@@ -146,6 +146,9 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		)
 	}()
 
+	// hack-fix t oensure that satellite gets up and running before starting storage nodes
+	time.Sleep(2 * time.Second)
+
 	// start the storagenodes
 	for i, v := range runCfg.StorageNodes {
 		go func(i int, v StorageNode) {


cmd/captplanet/run.go
```diff
@@ -146,6 +146,9 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		)
 	}()
 
+	// hack-fix t oensure that satellite gets up and running before starting storage nodes
+	time.Sleep(2 * time.Second)
```
mniewrzal: Wouldn't be safer to just remove goroutine for satellite startup?

egonelbre: the go-routine stops only when the satellite has finished, not earlier

mniewrzal: I mean if satellite will be starting without separate goroutine no need for sleep because starting storage nodes will happen when satellite will start. I was thinking about something like this:
```
	// start satellite
	_, _ = fmt.Printf("Starting satellite on %s\n",
		runCfg.Satellite.Identity.Server.Address)

	if runCfg.Satellite.Audit.SatelliteAddr == "" {
		runCfg.Satellite.Audit.SatelliteAddr = runCfg.Satellite.Identity.Server.Address
	}

	if runCfg.Satellite.Web.SatelliteAddr == "" {
		runCfg.Satellite.Web.SatelliteAddr = runCfg.Satellite.Identity.Server.Address
	}

	database, err := satellitedb.New(runCfg.Satellite.Database)
	if err != nil {
		errch <- errs.New("Error starting master database on satellite: %+v", err)
		return
	}

	err = database.CreateTables()
	if err != nil {
		errch <- errs.New("Error creating tables for master database on satellite: %+v", err)
		return
	}

	//nolint ignoring context rules to not create cyclic dependency, will be removed later
	ctx = context.WithValue(ctx, "masterdb", database)

	// Run satellite
	runCfg.Satellite.Identity.Run(ctx,
		grpcauth.NewAPIKeyInterceptor(),
		runCfg.Satellite.Kademlia,
		runCfg.Satellite.Audit,
		runCfg.Satellite.Overlay,
		runCfg.Satellite.Discovery,
		runCfg.Satellite.PointerDB,
		runCfg.Satellite.Checker,
		runCfg.Satellite.Repairer,
		runCfg.Satellite.BwAgreement,
		runCfg.Satellite.Web,
		runCfg.Satellite.Tally,
		runCfg.Satellite.Rollup,

		// NB(dylan): Inspector is only used for local development and testing.
		// It should not be added to the Satellite startup
		runCfg.Satellite.Inspector,
	)

	// start the storagenodes
	for i, v := range runCfg.StorageNodes {
		go func(i int, v StorageNode) {
```

egonelbre: `runCfg.Satellite.Identity.Run(ctx,` <-- this line blocks until satellite finishes

egonelbre: a better approach would be to poll satellite until it's accepting connections, but since we want to start using the actual binaries for development to discover more bugs, we can just use the hack-fix for the time being

mniewrzal: ah yes, my bad. Thanks for explanation.

PR #932, opened 2018-12-21
Implement mutex around satellitedb

egonelbre:
Implement mutex around satellite db to workaround sqlite "database is locked" issues.


pkg/datarepair/queue/queue_test.go
```diff
@@ -63,79 +69,91 @@ func TestSequential(t *testing.T) {
...
+			expected := dequeued.LostPieces[0]
+			assert.True(t, proto.Equal(addSegs[expected], &dequeued))
 		}
 	})
 }
 
 func TestParallel(t *testing.T) {
-	ctx := testcontext.New(t)
-	defer ctx.Cleanup()
+	t.Skip("logic is broken on database side")
```
dylanlott: nit: might be a good idea to add  TODO here so that it gets caught by a `grep` next time we're looking for cards / issues to make.

satellite/satellitedb/mutex.go
```diff
@@ -0,0 +1,267 @@
...
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/statdb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+// Mutex wraps DB in a mutex for non-concurrent databases
+type Mutex struct {
+	mu sync.Mutex
+	db *DB
```
dylanlott: Nit: consider naming it something different than `db` for less confusion later on

satellite/satellitedb/mutex.go
```diff
@@ -0,0 +1,267 @@
...
+	db.mu.Lock()
+	return db.mu.Unlock
+}
+
+// BandwidthAgreement is a getter for bandwidth agreement repository
+func (db *Mutex) BandwidthAgreement() bwagreement.DB {
+	return &muBandwidthAgreement{mu: db, db: db.db.BandwidthAgreement()}
+}
+
+// // PointerDB is a getter for PointerDB repository
```
dylanlott: Is there a reason this is commented out? Should we remove this?

egonelbre: PointerDB isn't yet implemented for master database.

pkg/datarepair/queue/queue_test.go
egonelbre: There's already an issue for it https://storjlabs.atlassian.net/browse/V3-1016

satellite/satellitedb/mutex.go
egonelbre: Any suggestions?

dylanlott:
Something specific to each struct might be good. Like “bwdb” for example. Just something to break up the db monotony.   
  

  
  

  
  
>

egonelbre:
> Something specific to each struct might be good. Like “bwdb” for example. Just something to break up the db monotony.

Ah, in that sense. Hmm, ideally this code would be generated rather than written manually. I didn't find a code-generator for this purpose, yet. The reason for the monotony is that it was easier to batch write them.

I really don't have anything against writing them as such. But, I don't think it will help against monotony, because the whole code unfortunately is all just wrapper code with a single func. It wouldn't make finding the specific func easier (because of the number of them), separating them would add more noise... but alas, it's just one brain interpretation of the situation.

But yeah, if you think changing them would help you work with the code easier then sure I'll make the change.

satellite/satellitedb/mutex.go
egonelbre: fixed

satellite/satellitedb/mutex.go
```diff
@@ -0,0 +1,267 @@
...
+
+// ReverseList lists all keys in revers order
+func (db *muOverlayCache) ReverseList(start storage.Key, limit int) (storage.Keys, error) {
+	defer db.mu.locked()()
+	return db.overlay.ReverseList(start, limit)
+}
+
+// Iterate iterates over items based on opts
+func (db *muOverlayCache) Iterate(opts storage.IterateOptions, fn func(storage.Iterator) error) error {
+	return errors.New("not implemented")
```
mniewrzal: Any reason why this is different from `ReverseList` or `Close`?

egonelbre: It has a function closure. Although I probably could still wrap it inside the mutex.

mniewrzal: No need to complicate this. I think it can stay as it is.

PR #933, opened 2018-12-21
adds to overlay config to allow requests including new nodes

navillasa:
Adds to the NodeSelectionConfig in the overlay pkg:
- `NewNodeAuditThreshold`: the number of audits a node must have to not be considered a New Node
- `NewNodePercentage`: the percentage of new nodes allowed per request
- `MatchingNodeRatio`: the ratio of requested nodes to matching nodes for a given list of restrictions

The overlay server's `FindStorageNodes` now gets reputable nodes and new nodes based on config settings.


pkg/overlay/server_test.go
```diff
@@ -65,3 +65,29 @@ func TestServer(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer ctx.Check(planet.Shutdown)
+
+	planet.Start(ctx)
+	// we wait a second for all the nodes to complete bootstrapping off the satellite
+	time.Sleep(2 * time.Second)
+
+	satellite := planet.Satellites[0]
+	server := overlay.NewServer(satellite.Log.Named("overlay"), satellite.Overlay, &pb.NodeStats{}, 0, 0.5)
```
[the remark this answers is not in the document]
egonelbre: Hmm, I think the issue here is more of that specific method should belong to the database not the endpoint nor service. This makes it somewhat more harder to test. Also I started wondering whether `newNodePercentage` should be a query argument or the overlay argument?

Should the user be able to specify the limit?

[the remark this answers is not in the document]
navillasa: Which method should belong to the database? NewServer? I think once there's a master database a lot of this code will be obsolete and yes `newNodePercentage` will be treated more as a query. For now I think it's okay though.

Do you think it would be better for the user to set the new node percentage instead of the overlay server config? If so, I can change it.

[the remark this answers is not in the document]
egonelbre: I mean the listing function. It takes a lot of filtering parameters and most of them don't have to do with the overlay logic itself, but rather selecting thing from the database. But, yeah it's OK for now.

As for the user specifying. I'm not sure maybe ask in slack what other think.

pkg/overlay/config.go
```diff
@@ -46,6 +46,10 @@ type NodeSelectionConfig struct {
 	UptimeCount       int64   `help:"the number of times a node's uptime has been checked" default:"0"`
 	AuditSuccessRatio float64 `help:"a node's ratio of successful audits" default:"0"`
 	AuditCount        int64   `help:"the number of times a node has been audited" default:"0"`
+
+	NewNodeAuditThreshold int64   `help:"the number of audits a node must have to not be considered a New Node" default:"0"`
```
mobyvb: shouldn't the default be at least 1? Otherwise a brand new node will not be considered a new node

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			requestedNodeAmt:      2,
+			reputableNodes:        3,
+			expectedResultLength:  2,
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:                  "case: zero reputable nodes found, only new nodes",
+			requestedNodeAmt:      2,
+			reputableNodes:        0,
+			expectedResultLength:  2,
```
mobyvb: shouldn't `expectedResultLength` be 0 since `newNodePercentage` is 0 for this case?

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:              "case: fewer than required new nodes",
+			requestedNodeAmt:  2,
+			reputableNodes:    3,
+			newNodePercentage: 0.5,
+			// this gives extra reputable instead
+			expectedResultLength:  3,
```
mobyvb: how is `requestedNodeAmt` 2 but `expectedResultLength` is 3?

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			expectedResultLength:  1,
+			newNodeAuditThreshold: 1,
+		},
+	} {
+		server := overlay.NewServer(satellite.Log.Named("overlay"), satellite.Overlay,
+			&pb.NodeStats{}, 2, tt.newNodeAuditThreshold, tt.newNodePercentage)
+
+		for i := 0; i <= tt.reputableNodes; i++ {
+			err := satellite.Overlay.Put(ctx, planet.StorageNodes[i].ID(), pb.Node{
+				Reputation: &pb.NodeStats{AuditCount: 1},
```
mobyvb: Here, I think you should say `Reputation: &pb.NodeStats{AuditCount: tt.newNodeAuditThreshold}` rather than explicitly putting `1` in

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			// this gives extra reputable instead
+			expectedResultLength:  3,
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:                  "case: more than required new nodes",
+			requestedNodeAmt:      2,
+			reputableNodes:        2,
+			newNodePercentage:     0.5,
+			expectedResultLength:  3,
```
mobyvb: should `expectedResultLength` be 2?

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			// todo(nat): fix nodes length issue
+			name:                  "case: zero new nodes found, only reputable nodes",
+			requestedNodeAmt:      3,
+			reputableNodes:        3,
+			newNodePercentage:     0.5,
+			expectedResultLength:  4,
```
mobyvb: `expectedResultLength` > `requestedNodeAmt`

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:                  "case: exactly the required amount of new and reputable nodes returned",
+			requestedNodeAmt:      1,
+			reputableNodes:        1,
+			newNodePercentage:     1,
+			expectedResultLength:  2,
```
mobyvb: `expectedResultLength` > `requestedNodeAmt`

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:              "case: low percentage of new nodes",
+			requestedNodeAmt:  3,
+			reputableNodes:    1,
+			newNodePercentage: 0.01,
+			// todo(nat): expect this result to be 1
+			expectedResultLength:  3,
```
mobyvb: 1 reputable node, 3 new nodes, low new node percentage -> I feel like `expectedResultLength` should not be larger than 2 under these conditions

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:                  "case: high percentage of new nodes",
+			requestedNodeAmt:      1,
+			reputableNodes:        1,
+			newNodePercentage:     3,
```
mobyvb: Should we allow `newNodePercentage` to be greater than 1?

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			expectedResultLength:  4,
```
mobyvb: `expectedResultLength` > `requestedNodeAmt`

pkg/overlay/server_test.go
```diff
@@ -65,3 +67,142 @@ func TestServer(t *testing.T) {
...
+			newNodeAuditThreshold: 1,
+		},
+		{
+			name:                  "case: 0% new nodes requested",
```
mobyvb: this case should be covered by the first few, but I guess it can't hurt to have an explicit case defined for it

navillasa: yeah I was considering taking this out too because redundant but I guess it's no issue to keep it

pkg/overlay/config.go
navillasa: good point!

pkg/overlay/server_test.go
navillasa: yeah sorry for not clarifying this sooner (it's kind of confusing) but after talking to JT, we decided the functionality should be like:

`result` =   `requestedNodeAmt` + _requestedNodeAmt * newNodePercentage_

where the part in italics is the amount of new nodes included in the result.

pkg/overlay/server_test.go
navillasa: yeahh since I was planning on coming back to this, I was being sneaky by putting 2 as expectedResultLength but it should've been 0. but that would fail, and I'm not sure if it's the test or the filtering code so I'll check it out.

pkg/overlay/server_test.go
navillasa: okay yeah this is another weird one that I just made the expected result length 3 to appease the test passing, but I think it should just be 1 and an error. but that wasn't happening so need to investigate.

pkg/overlay/server_test.go
navillasa: that's a good point even with the new way of calculating the node results. because then it's like 3x the requested nodes in addition to the requested amount, which could get really big.

pkg/overlay/server_test.go
mobyvb: ohhhhhhh

navillasa: yeah I think this means I should also rename `maxNodes`

navillasa:
replaced by this pr https://github.com/storj/storj/pull/1123

PR #936, opened 2018-12-26
Satellite api keys api

rikysya:


pkg/satellite/apikeys.go
```diff
@@ -24,15 +30,47 @@ type APIKeys interface {
...
+// APIKey describing api key model in the database
 type APIKey struct {
 	ID uuid.UUID `json:"id"`
 
 	// Fk on project
 	ProjectID uuid.UUID `json:"projectId"`
 
-	Key  []byte `json:"key"`
-	Name string `json:"name"`
+	Key  MockKey `json:"key"`
```
egonelbre: I wonder whether the Key should be part of this struct. In many places it's being overwritten by zeros. Maybe rename this as `APIKeyInfo` and then all the places that need that Key it is returned outside of the struct.

This should avoid accidentally leaking the Key and might remove some code.

pkg/satellite/apikeys.go
```diff
@@ -4,35 +4,72 @@
...
 	// Fk on project
 	ProjectID uuid.UUID `json:"projectId"`
 
-	Key  []byte `json:"key"`
 	Name string `json:"name"`
 
 	CreatedAt time.Time `json:"createdAt"`
 }
+
+// APIKey is a mock api key type
```
egonelbre: Not quite sure why is a `mock api key`

PR #937, opened 2018-12-26
[V3-911] fixed registration mandatory fields validation

NikolaiYurchenko:


web/satellite/src/utils/validation.ts
```diff
@@ -1,8 +1,14 @@
...
 // See LICENSE for copying information.
 
-export function validateEmail(email: string) : boolean {
+export function validateEmail(email: string): boolean {
     const rgx = /.*@.*\..*$/;
 
     return rgx.test(email);
 }
+
+export function validatePassword(password: string): boolean {
```
egonelbre: Wait, so no unicode passwords or special symbols?

Also is it validating or is it checking the strength?

NikolaiYurchenko: For now it was my task in jira to validate this field
There is link https://storjlabs.atlassian.net/secure/RapidBoard.jspa?rapidView=12&projectKey=V3&modal=detail&selectedIssue=V3-911&quickFilter=18

egonelbre: I think this would fail with `абвгдежзийклм①②③④⑤` so it's not really checking for alphabet nor numbers properly.

NikolaiYurchenko: Was tested, didn`t fail

NikolaiYurchenko: Yes, it should return false indeed. For now we accept only latin alphabet letters and numbers. We decided to do so by following best practice from other services (like Google for example). We only missed common punctuation symbols but we can update it in one of the next iterations. We don't think that kyrillic, arabic alphabet , hierogliphs or other not usual symbols need to be applied here. Just take a look on other services. Just my opinion. Thank you for your response for this PR

egonelbre: As an example, this is the output from my password generator with (roughly) strongest settings `¸:¦!iÈÒ+Ê^Dü.å¢Ú@÷þÍÇôÒÞÊaÑÖI¬àæÜ;FcSãV3`, using something excluding high ansi `_%X'N#ZM~^v7L<i^6H7g` (this should still work in GMail).

I'm not sure what the general usability for Asian countries are with regards to unicode symbols?

Troy Hunt recommends supporting them: https://www.troyhunt.com/passwords-evolved-authentication-guidance-for-the-modern-era/

PR #940, opened 2018-12-27
V3-976 Create db query for filtering team mebers

crawter:


pkg/satellite/satellitedb/projectmembers.go
```diff
@@ -30,23 +31,69 @@ func (pm *projectMembers) GetByMemberID(ctx context.Context, memberID uuid.UUID)
...
+
+	var projectMembers []satellite.ProjectMember
+	searchSubQuery := "'%" + pagination.Search + "%'"
+
+	rows, err := pm.db.Query(pm.db.Rebind(`
+		SELECT pm.*
+			FROM project_members pm
+				INNER JOIN users u ON pm.member_id = u.id
+					WHERE pm.project_id = ?
+					AND ( u.email LIKE `+searchSubQuery+` OR
```
egonelbre: Use query parameters. This seems to have an sql injection attack possible.

pkg/satellite/satellitedb/projectmembers.go
```diff
@@ -30,23 +31,69 @@ func (pm *projectMembers) GetByMemberID(ctx context.Context, memberID uuid.UUID)
...
+
+	rows, err := pm.db.Query(pm.db.Rebind(`
+		SELECT pm.*
+			FROM project_members pm
+				INNER JOIN users u ON pm.member_id = u.id
+					WHERE pm.project_id = ?
+					AND ( u.email LIKE `+searchSubQuery+` OR 
+						  u.first_name LIKE `+searchSubQuery+` OR 
+						  u.last_name LIKE `+searchSubQuery+` )
+						ORDER BY u.`+getOrder(pagination.Order)+` ASC
```
egonelbre: Use query parameters rather than adding strings.

pkg/satellite/satellitedb/projectmembers_test.go
```diff
@@ -69,30 +69,51 @@ func TestProjectMembersRepository(t *testing.T) {
...
 		assert.Equal(t, 2, len(members))
 
-		members, err = projectMembers.GetByProjectID(ctx, createdProjects[0].ID, 1, 1)
+		members, err = projectMembers.GetByProjectID(ctx, createdProjects[0].ID, satellite.Pagination{Limit: 6, Offset: 0, Search: "son", Order: 123})
+		assert.Nil(t, err)
+		assert.NoError(t, err)
+		assert.NotNil(t, members)
+		assert.Equal(t, 5, len(members))
+
+		members, err = projectMembers.GetByProjectID(ctx, createdProjects[0].ID, satellite.Pagination{Limit: 6, Offset: 3, Search: "son", Order: 2})
```
egonelbre: Test for injections.

pkg/satellite/service.go
```diff
@@ -395,22 +395,18 @@ func (s *Service) DeleteProjectMembers(ctx context.Context, projectID uuid.UUID,
...
+func (s *Service) GetProjectMembers(ctx context.Context, projectID uuid.UUID, pagination Pagination) (pm []ProjectMember, err error) {
 	defer mon.Task()(&ctx)(&err)
 	_, err = GetAuth(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	if limit < 0 || offset < 0 {
-		return nil, errs.New("invalid pagination argument")
+	if pagination.Limit > maxLimit {
```
egonelbre: What happens when limit is negative or not supplied?

crawter: @egonelbre this check is moved to repository

egonelbre: When the client requests without limit parameter doesn't it default to `0` making it not return any results?

crawter: yes. Should we return an error?

egonelbre: If `limit` is optional, then we should default to a reasonable value. If it's not optional then return an error.

pkg/satellite/satellitedb/projectmembers_test.go
```diff
@@ -163,3 +218,20 @@ func prepareUsersAndProjects(ctx context.Context, t *testing.T, users satellite.
 
 	return usersList, projectList
 }
+
+func TestSanitizedOrderColumnName(t *testing.T) {
```
egonelbre: This really doesn't test anything. Maybe do this test with calling GetByProjectID.

PR #941, opened 2018-12-27
repair storagenode diag command

littleskunk:


cmd/storagenode/main.go
```diff
@@ -165,7 +165,7 @@ func cmdDiag(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	// open the sql db
-	dbpath := filepath.Join(diagDir, "piecestore.db")
+	dbpath := filepath.Join(diagDir, "storage/piecestore.db")
```
egonelbre: Although change this to `filepath.Join(diagDir, "storage", "piecestore.db")`, that way it'll give the correct path for windows as well.

PR #943, opened 2018-12-28
Exclude satellite from storage node search

mniewrzal:
https://storjlabs.atlassian.net/browse/V3-1009


egonelbre:
This should be eventually fixed with https://github.com/storj/storj/pull/873. There shouldn't be a need for explictly removing self from the results.

mniewrzal:
Ok, thanks for info.

PR #944, opened 2018-12-28
Fix ignored setup arguments

mniewrzal:
https://storjlabs.atlassian.net/browse/V3-999


mniewrzal:
I'm still trying to figure out how to avoid saving flags specific only for setup.

mniewrzal:
Removing setup flags from config and additional cleaning (like https://storjlabs.atlassian.net/browse/V3-1040) will be done separately.

PR #948, opened 2018-12-29
Satellite Web store test coverage 

Barterio:
Added 100% coverage for these modules: 
* users
* projects
* projectMembers


web/satellite/src/utils/apolloManager.ts
```diff
@@ -10,6 +10,7 @@ import { getToken } from '@/utils/tokenManager';
 // Satellite url
 const satelliteUrl = new HttpLink({
 	uri: 'http://localhost:8081/api/graphql/v0',
+
```
crawter: ?

PR #950, opened 2019-01-01
CSR Service (part 2): cert signing rpc

bryanchriswhite:
+ implement certificate sign rpc method
+ implement certificate signer client/server
+ refactor `AuthorizationDB#Create`
+ refactor `NewTestIdentity`
+ add `AuthorizationDB#Claim`
+ add `Token#Equal`
+ fix `Authorizations#Marshal` when marshaling identities and certificates
+ tweak `Authorization#String` format
+ cert debugging improvements (jsondiff)
+ receive context arg in `NewTestIdentity`
+ misc. fixes


pkg/certificates/certificates.go
```diff
@@ -85,21 +97,49 @@ type Token struct {
...
+		AuthDB: authDB,
 	}
 
 	return &srv
 }
 
+// NewClient creates a new certificate signing grpc client
+func NewClient(ident *provider.FullIdentity, address string) (*Client, error) {
+	tc := transport.NewClient(ident)
+	conn, err := tc.DialAddress(context.Background(), address)
```
bryanchriswhite: should this context be passed in?

jtolds: yeah i totally spaced that the whole thing that inspired me to make sure wiki syncing was set up between github.com/storj/docs and github.com/storj/storj/wiki was to add to the code style guide that unless the case is exceptional, you should never create a new background context and always take one from somewhere else. if we need a context here, NewClient should take one as an argument.

jtolds: i think i was looking at another file though

pkg/certificates/certificates.go
```diff
@@ -274,6 +356,75 @@ func (a *AuthorizationDB) UserIDs() ([]string, error) {
...
 
+// Claim marks an authorization as claimed and records claim information
+func (a *AuthorizationDB) Claim(req *pb.SigningRequest, grpcPeer *peer.Peer, cert *x509.Certificate) error {
+	now := time.Now().Unix()
+	if !(now-MaxClaimDelaySeconds < req.Timestamp) ||
+		!(req.Timestamp < now+MaxClaimDelaySeconds) {
+		return ErrAuthorization.New("claim timestamp is outside of max delay window: %d", req.Timestamp)
+	}
+
+	ident, err := provider.PeerIdentityFromPeer(grpcPeer)
```
bryanchriswhite: do we want to check peer difficulty?

pkg/bwagreement/test/server_test.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap"
 
-	testidentity "storj.io/storj/internal/identity"
 	"storj.io/storj/internal/testcontext"
+	testidentity "storj.io/storj/internal/testidentity"
```
egonelbre: The alias can now be removed.

pkg/certificates/certificates_test.go
```diff
@@ -5,23 +5,49 @@ package certificates
...
+	"crypto/x509"
 	"encoding/gob"
 	"fmt"
+	"net"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
+	"storj.io/storj/pkg/identity"
```
egonelbre: organize the imports list

pkg/certificates/mocks/mock_client.go
```diff
@@ -0,0 +1,58 @@
...
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
+	grpc "google.golang.org/grpc"
+
+	pb "storj.io/storj/pkg/pb"
+)
+
+// MockCertificatesClient is a mock of CertificatesClient interface
+type MockCertificatesClient struct {
```
egonelbre: Is there a way this mock can be avoided and use an actual implementation instead? We have missed a bunch of bugs due to using mocks.

pkg/identity/certificate_authority.go
```diff
@@ -74,6 +75,32 @@ type FullCAConfig struct {
 	KeyPath  string `help:"path to the private key for this identity" default:"$CONFDIR/ca.key"`
 }
 
+// NewCA creates a new full identity with the given difficulty
+func NewCA(ctx context.Context, opts NewCAOptions) (
+	rv *FullCertificateAuthority, err error) {
```
egonelbre: remove line break

pkg/peertls/peertls_test/peertls_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2018 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-package peertls
+package peertls_test
```
egonelbre: The extra folder isn't necessary.

pkg/certificates/mocks/mock_client.go
bryanchriswhite: This mock is only used for testing the client. The server method has a test as well as an end to end test using an actual client and running grpc server.

pkg/identity/certificate_authority.go
bryanchriswhite: @jtolds

pkg/peertls/peertls_test/peertls_test.go
bryanchriswhite: You can't have multiple packages in the same folder,  what are you suggesting?

pkg/identity/certificate_authority.go
bryanchriswhite: why is this a named return value actually? any objections to replacing it with `_`?

pkg/certificates/mocks/mock_client.go
egonelbre: Sure, I just want to avoid adding new mock code. I like to see most of the mocks disappear from the code-base. So yeah, we can use the actual server for testing rather than the mock. For the error conditions we can use a erroring wrapper if necessary.

internal/debugging/certificates.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return err
+	}
+
+	c2JSON, err := c2.JSON()
+	if err != nil {
+		return err
+	}
+
+	diffType, diff := jsondiff.Compare(cJSON, c2JSON, &diffOpts)
```
egonelbre: Note there's also https://godoc.org/github.com/google/go-cmp/cmp#Diff already in our dependencies.

pkg/certificates/certificates_test.go
```diff
@@ -53,7 +78,8 @@ func TestAuthorizationDB_Create(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer func() {
-		_ = authDB.Close()
+		err := authDB.Close()
```
egonelbre: You can use `defer ctx.Check(authDB.Close)`

pkg/certificates/certificates_test.go
```diff
@@ -194,6 +221,225 @@ func TestAuthorizationDB_Get(t *testing.T) {
...
+	assert.Equal(t, grpcPeer.Addr.String(), claim.Addr)
+	assert.Equal(t, [][]byte{ident.CA.Raw}, claim.SignedChainBytes)
+	assert.Condition(t, func() bool {
+		return now-MaxClaimDelaySeconds < claim.Timestamp &&
+			claim.Timestamp < now+MaxClaimDelaySeconds
+	})
+}
+
+func TestAuthorizationDB_Claim_Invalid(t *testing.T) {
+	ctx := testcontext.New(t)
```
egonelbre: follow this immediately with `defer ctx.Cleanup()`

internal/debugging/certificates.go
bryanchriswhite: I think the formatting is nicer in jsondiff from what I can tell

pkg/certificates/certificates_test.go
bryanchriswhite: Nice

pkg/peertls/peertls_test/peertls_test.go
bryanchriswhite: Cool, didn't know that

internal/debugging/certificates.go
egonelbre: Sure, just trying to avoid pull in dependencies unless necessary.

pkg/certificates/certificates.go
```diff
@@ -286,6 +462,8 @@ func (a *Authorizations) Unmarshal(data []byte) error {
 // Marshal serializes a set of authorizations
 func (a Authorizations) Marshal() ([]byte, error) {
 	data := new(bytes.Buffer)
+	gob.Register(&ecdsa.PublicKey{})
```
wthorp: Should this be done in an init()?

pkg/certificates/certificates.go
```diff
@@ -186,18 +245,68 @@ func (c CertSignerConfig) NewAuthDB() (*AuthorizationDB, error) {
...
+	}
+
+	peerIdent, err := identity.PeerIdentityFromPeer(grpcPeer)
+	if err != nil {
+		return nil, err
+	}
+
+	signedPeerCA, err := c.Signer.Sign(peerIdent.CA)
+	if err != nil {
+		return nil, err
```
wthorp: Is Sign() a GRPC method?  Do we ever have any error messages that we want to hide from the client?

bryanchriswhite: It is; I'm pretty sure all error messages are intended for client consumption. We are actively preventing activation token `String` method from printing more than the first little bit of the token data for this reason. I'm not aware of any other sensitive info that might show up in error messages but you make a good point.

pkg/identity/certificate_authority.go
jtolds: my brain is broken. it is not my editor. https://media1.tenor.com/images/74afe6af7eefdda9fbb93e2475b7ad50/tenor.gif?itemid=5082944

PR #954, opened 2019-01-01
CSR automation in storage node setup

bryanchriswhite:
_(depends on #950)_

### DIFF
--- a/cmd/certificates/main.go
+++ b/cmd/certificates/main.go
@@ -38,9 +38,10 @@ var (
 	}
 
 	setupCmd = &cobra.Command{
-		Use:   "setup",
-		Short: "Setup a certificate signing server",
-		RunE:  cmdSetup,
+		Use:         "setup",
+		Short:       "Setup a certificate signing server",
+		RunE:        cmdSetup,
+		Annotations: map[string]string{"type": "setup"},
 	}
 
 	runCmd = &cobra.Command{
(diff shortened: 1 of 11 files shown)


bryanchriswhite:
I need to add integration tests for this but in the mean time this can be tested by:
```bash
go install ./cmd/{certificate,storagenode}
certificates setup
certificates auth create 1 user@example.com
certficates auth info user@example.com --show-tokens
# copy open token
certificates run --server.address 127.0.0.1:8888

# new terminal
storagenode setup --signer.address 127.0.0.1:8888 -- --signer.auth-token # paste token

# Check that the storage node's CA and identitiy cert chains contain 2 and 3 certs, respectively,
# and that the last cert matches the signer CA cert (default paths [linux]
# ~/.local/share/storj/storagenode/ca.cert,
# ~/.local/share/storj/storagenode/identity.cert, and
# ~/.local/share/storj/certificates/ca.cert respectively).
```

cmd/certificates/main.go
```diff
@@ -125,6 +128,11 @@ func cmdSetup(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	err = os.MkdirAll(setupDir, 0700)
```
coyle: nit -- https://golang.org/pkg/os/#FileMode

pkg/certificates/certificates.go
```diff
@@ -191,6 +191,90 @@ func ParseToken(tokenString string) (*Token, error) {
...
+	} else {
+		ident, err = identConfig.Create(ca)
+		if err != nil {
+			return err
+		}
+	}
+
+	signedChainBytes, err := c.Sign(ctx, ident)
+	if err != nil {
+		return fmt.Errorf("error occured while signing certificate: %s\n(identity files were still generated and saved, if you try again existnig files will be loaded)", err)
```
coyle: reason to not do errs.New()

PR #955, opened 2019-01-02
pkg/server: add both a public and private grpc server

jtolds:
private will be for the inspector.
also clean up grpc server creation, including cert verification and api key validation.


egonelbre:
I don't think we need two listeners in a single `server`. The appropriate `Peer` can start two different servers instead and register the endpoints accordingly.

jtolds:
yeah, i'm okay to replace this with the Peer stuff when it's ready. the sole point of this PR is to get #956 working, which i would like to have working asap

PR #956, opened 2019-01-02
pkg/inspector: split functionality to be per-service

jtolds:
now kad inspector features exist on every server that has kademlia running on the private rpc port. likewise, overlay and statdb.

this means that now i have what i really wanted all along: kad inspection features are now available on storage nodes


egonelbre:
I don't think we need two servers as part of `server` package. The appropriate `Peer` can start two different servers instead and register the endpoints accordingly.

jtolds:
> I don't think we need two listeners in a single server. The appropriate Peer can start two different servers instead and register the endpoints accordingly.

that's fine, once we have peers. i would like to be able to attach the inspector to any kademlia participant asap though. with that in mind, it's important that we don't expose debugging/internal details on the public port

pkg/auth/grpcauth/authentication.go
```diff
@@ -7,32 +7,45 @@ import (
...
-		if !ok || APIKey == "" {
-			return handler(ctx, req)
-		}
-		return handler(auth.WithAPIKey(ctx, []byte(APIKey)), req)
+func NewAPIKeyInterceptor() grpcutils.ServerInterceptor {
+	authfunc := grpc_auth.AuthFunc(
+		func(ctx context.Context) (context.Context, error) {
+			md, ok := metadata.FromIncomingContext(ctx)
+			APIKey := strings.Join(md["apikey"], "")
+			if !ok || APIKey == "" {
```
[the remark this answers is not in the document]
jtolds: oh, i meant to fix this actually. this was from the code that was there prior. i'll fix

pkg/auth/grpcauth/authentication.go
```diff
@@ -7,32 +7,45 @@ import (
...
-		err error) {
-
-		md, ok := metadata.FromIncomingContext(ctx)
-		APIKey := strings.Join(md["apikey"], "")
-		if !ok || APIKey == "" {
-			return handler(ctx, req)
-		}
-		return handler(auth.WithAPIKey(ctx, []byte(APIKey)), req)
+func NewAPIKeyInterceptor() grpcutils.ServerInterceptor {
+	authfunc := grpc_auth.AuthFunc(
```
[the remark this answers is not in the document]
jtolds: for packages? i was just matching the grpc middleware conventions: https://godoc.org/github.com/grpc-ecosystem/go-grpc-middleware

pkg/bwagreement/config.go
```diff
@@ -30,11 +30,14 @@ func (c Config) Run(ctx context.Context, server *provider.Provider) (err error)
 
 	zap.S().Debug("Starting Bandwidth Agreement Receiver...")
 
-	db, ok := ctx.Value("masterdb").(interface{ BandwidthAgreement() DB })
+	db, ok := ctx.Value("masterdb").(interface {
```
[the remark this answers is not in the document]
jtolds: okay

pkg/grpcutils/utils.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package grpcutils
+
+import (
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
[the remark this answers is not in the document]
jtolds: same as above

pkg/auth/grpcauth/authentication.go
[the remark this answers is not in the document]
coyle: huh ¯\_(ツ)_/¯ maybe I'm wrong then ... it just looked odd to me. Carry on

pkg/grpcutils/utils.go
egonelbre: The `grpc_middleware` was initially avoided because it's bringing in quite a lot of dependencies.

pkg/server/interceptors.go
```diff
@@ -12,43 +12,39 @@ import (
...
-	if err != nil {
-		// no zap errors for canceled or wrong file downloads
-		if storage.ErrKeyNotFound.Has(err) ||
-			status.Code(err) == codes.Canceled ||
-			status.Code(err) == codes.Unavailable ||
-			err == io.EOF {
+func defaultLogger() grpcutils.ServerInterceptor {
+	return grpcutils.ServerInterceptor{
+		Stream: func(srv interface{}, ss grpc.ServerStream,
+			info *grpc.StreamServerInfo, handler grpc.StreamHandler) (err error) {
```
egonelbre: remove line break

pkg/server/interceptors.go
```diff
@@ -12,43 +12,39 @@ import (
...
-func unaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{},
-	err error) {
-	resp, err = handler(ctx, req)
-	if err != nil {
-		// no zap errors for wrong file downloads
-		if status.Code(err) == codes.NotFound {
+		},
+		Unary: func(ctx context.Context, req interface{},
+			info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (
+			resp interface{}, err error) {
```
egonelbre: remove line break

pkg/server/grpc.go
```diff
@@ -0,0 +1,38 @@
...
+	"storj.io/storj/pkg/grpcutils"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls"
+)
+
+// SetupRPCs is a common place where both the public and private gRPC servers
+// are set up.
+func SetupRPCs(ident *identity.FullIdentity,
+	pcvs []peertls.PeerCertVerificationFunc) (public, private *grpc.Server,
+	err error) {
```
egonelbre: remove line breaks

pkg/grpcutils/utils.go
jtolds: oh, okay, i can take it out. the main thing i care about is that we keep stream and unary interceptors together. that's super important, regardless of if we use a chaining interceptor helper

pkg/auth/grpcauth/authentication.go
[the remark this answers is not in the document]
jtolds: just fyi, it appears all of your comments are on changes from #955, so i'll address them there

pkg/statdb/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+	sdb, ok := ctx.Value("masterdb").(interface {
+		StatDB() DB
+	})
+	if !ok {
+		return Error.New("unable to get master db instance")
+	}
+
+	pb.RegisterStatDBInspectorServer(server.PrivateRPC(),
+		NewInspector(sdb.StatDB()))
```
egonelbre: remove line break

pkg/grpcutils/utils.go
jtolds: wait, what dependencies? it didn't add more than itself to go.sum. is that because we already depend on its dependencies?

pkg/server/server.go
```diff
@@ -19,57 +22,79 @@ type Service interface {
...
+// NewServer creates a Server out of an Identity, public and private
+// gRPC handles and listeners, and a set of services. Closing a server will
+// stop the grpc servers and close the listeners.
+// A public handler is expected to be exposed to the world, whereas the private
+// handler is for inspectors and debug tools only.
+func NewServer(
+	identity *identity.FullIdentity,
+	publicSrv *grpc.Server, publicLis net.Listener,
+	privateSrv *grpc.Server, privateLis net.Listener,
+	services ...Service) *Server {
```
egonelbre: remove line breaks
also can this take instead a `identity *identity.FullIdentity, public, private Handle, services ...Service`

pkg/server/interceptors.go
```diff
@@ -12,43 +12,39 @@ import (
...
 
-func streamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) (err error) {
-	err = handler(srv, ss)
-	if err != nil {
-		// no zap errors for canceled or wrong file downloads
-		if storage.ErrKeyNotFound.Has(err) ||
-			status.Code(err) == codes.Canceled ||
-			status.Code(err) == codes.Unavailable ||
-			err == io.EOF {
+func defaultLogger() grpcutils.ServerInterceptor {
```
egonelbre: Pass in `zap.Logger`

pkg/server/grpc.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"storj.io/storj/pkg/auth/grpcauth"
+	"storj.io/storj/pkg/grpcutils"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls"
+)
+
+// SetupRPCs is a common place where both the public and private gRPC servers
+// are set up.
+func SetupRPCs(ident *identity.FullIdentity,
```
egonelbre: Pass in `zap.Logger`

internal/testplanet/node.go
```diff
@@ -52,43 +54,61 @@ func (planet *Planet) newNode(name string, nodeType pb.NodeType) (*Node, error)
...
+	// a per-node revocation db, or no revocation db?
+	// should it create a ca whitelist? if we do create a per-node revocation db,
+	// it should be added to node.Dependencies?
+	pcvs := server.PCVs(nil, nil, serverConfig.CertVerification.Extensions)
+
+	publicSrv, privateSrv, err := server.SetupRPCs(identity, pcvs)
 	if err != nil {
-		return nil, utils.CombineErrors(err, listener.Close())
+		return nil, utils.CombineErrors(err,
+			publicListener.Close(), privateListener.Close())
```
egonelbre: remove line break

internal/testplanet/node.go
```diff
@@ -52,43 +54,61 @@ func (planet *Planet) newNode(name string, nodeType pb.NodeType) (*Node, error)
...
 	}
 
+	node.Server = server.NewServer(
```
egonelbre: remove line break

jtolds:
wait wait, this change has PR #955 in it, and most of the review comments in this PR are actually about #955 
should i just abandon #955? what should i do for reviews when i have two order-dependent but otherwise separately reviewable changes?

pkg/utils/event.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package utils
+
+import "sync"
+
+// Event is a synchronization primitive that lets an arbitrary number
+// of concurrent waiters to wait for an event to happen 1 or more times.
+type Event struct {
```
egonelbre: We have `sync2` for synchronization primitives. Any reason you are not using `sync.Cond`?

This should definitely be a separate PR. And this needs tests.

pkg/server/server.go
```diff
@@ -19,57 +22,83 @@ type Service interface {
...
-	p.grpc.GracefulStop()
-	return nil
+	errch := make(chan error)
+	go func() {
+		errch <- Error.Wrap(p.public.Close())
+	}()
+	go func() {
+		errch <- Error.Wrap(p.private.Close())
+	}()
+	return errs.Combine(<-errch, <-errch)
```
egonelbre: This should stop waiting when one of them closes with an error.

pkg/server/server.go
```diff
@@ -19,57 +22,83 @@ type Service interface {
 	Run(ctx context.Context, server *Server) error
 }
 
+// Handle is a type that pairs a gRPC server to a listener
+type Handle struct {
+	Srv *grpc.Server
```
egonelbre: Server

pkg/server/server.go
```diff
@@ -19,57 +22,83 @@ type Service interface {
...
+	Lis net.Listener
```
egonelbre: Listener

pkg/server/server.go
```diff
@@ -19,57 +22,83 @@ type Service interface {
...
+	if err == grpc.ErrServerStopped {
+		return nil
+	}
+	return Error.Wrap(err)
+}
+
+// Close closes the gRPC server gracefully and shuts down the listener.
+func (h *Handle) Close() error {
+	h.Srv.GracefulStop()
+	_ = h.Lis.Close()
```
egonelbre: Don't ignore all errors. And I'm not sure whether this is necessary, GracefulStop should already close the listener.

pkg/server/config.go
```diff
@@ -8,50 +8,57 @@ import (
...
-	defer func() { _ = lis.Close() }()
+	defer func() { err = utils.CombineErrors(err, revdb.Close()) }()
 
-	opts, err := NewOptions(ident, sc)
+	publicLis, err := net.Listen("tcp", sc.PublicAddress)
 	if err != nil {
 		return err
 	}
-	defer func() { err = utils.CombineErrors(err, opts.RevDB.Close()) }()
+	defer func() { _ = publicLis.Close() }()
```
egonelbre: Don't ignore closing errors.

pkg/server/config.go
```diff
@@ -8,50 +8,57 @@ import (
...
 
-	s, err := NewServer(opts, lis, interceptor, services...)
+	privateLis, err := net.Listen("tcp", sc.PrivateAddress)
 	if err != nil {
 		return err
 	}
-	defer func() { _ = s.Close() }()
+	defer func() { _ = privateLis.Close() }()
```
egonelbre: Don't ignore closing errors.

pkg/certificates/certificates_test.go
```diff
@@ -45,6 +47,27 @@ var (
 	}
 )
 
+type noListener struct {
```
egonelbre: It seems it's getting messier than I expected.

pkg/utils/event.go
```diff
@@ -0,0 +1,44 @@
...
+	mtx         sync.Mutex
```
egonelbre: mu

pkg/utils/event.go
egonelbre:
```
type Event struct {
	mu     sync.Mutex
	unlock sync.Once
}

func NewEvent() {
	event := &Event{}
	event.mu.Lock()
	return event
}

func (event *Event) Wait() {
	event.mu.Lock()
	event.mu.Unlock()
}

func (event *Event) Fire() {
	event.unlock.Do(func(){
		event.mu.Unlock()
	})
}
```

```
// simplest but only allows close being called once
type Event struct {
	ch chan struct{}
}

func NewEvent() {
	return &Event{make(chan struct{})}
}

func (event *Event) Wait() { <-event.ch }
func (event *Event) Close() { close(event.ch) }
```

jtolds: yep, considered both of those options but the benefit of being able to use it like a sync.Mutex (zero value is initialized and ready to go) outweighed those two implementations for me. nonetheless, with the removal of the `noListener` struct, this also is unnecessary, and i removed it entirely

pkg/server/server.go
jtolds: why?

jtolds: i think it makes the most sense to wait for both here, but perhaps this is another use case for RunJointly (below)?

egonelbre: Actually, maybe it should up-to a specified timeout as well. The reason to return early is so that we don't hang to unclosed connections too long. AFAIR, `GracefulStop` waits for everyone to disconnect before returning.

pkg/utils/event.go
egonelbre: Sure, makes sense, just giving alternatives here that are bit easier to follow.

pkg/server/config.go
egonelbre: Let's add a comment why it's ignoring the error here then.

pkg/server/config.go
egonelbre: Let's add a comment why it's ignoring the error here then.

pkg/certificates/certificates_test.go
egonelbre: Because it added noListener and sync primitive Event (which would have also needed few tests) to support one test-case. The noListener had a stub address.

pkg/utils/utils.go
```diff
@@ -36,59 +38,22 @@ func SplitDBURL(s string) (string, string, error) {
...
-		}
-		return allErrors
-	}
-	return ""
-}
+var CombineErrors = errs.Combine
 
 // ErrorGroup contains a set of non-nil errors
-type ErrorGroup []error
+type ErrorGroup struct {
```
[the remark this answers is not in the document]
jtolds: yeah, absolutely

jtolds:
good catch on the api key error moby! i'll fix that

pkg/grpcutils/combine.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+)
+
+// CombineStreamServerInterceptors takes one or more grpc.StreamServerInterceptor and stacks them together into one.
+// Call order is left-to-right. The left-most interceptor is called first, with the right-most interceptor having the longest callstack.
+func CombineStreamServerInterceptors(interceptors ...grpc.StreamServerInterceptor) grpc.StreamServerInterceptor {
```
egonelbre: When you look at the previous implementation of the combining only supported two interceptors. The code was much easier to follow.

https://github.com/storj/storj/pull/956/files#diff-2c40b511d93b3aa7d4cbb2a52314f28fL46

jtolds: isn't that identical to lines 25-29 here? i thought i was using that code.

given the new structure, it made sense to make it work with n-ary interceptors, which i suppose i could do as a separate function, but that's the brief switch statement at the beginning

jtolds: i guess i should say, that link is for UnaryServerInterceptors, and this comment is on StreamServerInterceptors. I also shortened the UnaryServerInterceptor code by one call

egonelbre: But yeah, more of noting that this version is harder to read than the previous. Anyways, we can keep as is.

mniewrzal:
I found some problems with `inspect kad count`:

```
wywrzal@localhost:~/git/storj$ inspector kad count --address 127.0.0.1:44259
error processing request:: rpc error: code = Unimplemented desc = unknown service inspector.OverlayInspector
```
and
`inspect statdb getstats`
```
wywrzal@localhost:~/git/storj$ inspector statdb getstats 1tLq5KSc15BCdgYfD3Nu9v3hKAa2bico9LfZLUbgZvxiYmZ62M --address 127.0.0.1:44259
error processing request:: rpc error: code = Unimplemented desc = unknown service inspector.StatDBInspector
```


but rest of commands works ok :-)

```
wywrzal@localhost:~/git/storj$ inspector kad lookup 1tLq5KSc15BCdgYfD3Nu9v3hKAa2bico9LfZLUbgZvxiYmZ62M --address 127.0.0.1:44259
{
  "node": {
    "id": "1tLq5KSc15BCdgYfD3Nu9v3hKAa2bico9LfZLUbgZvxiYmZ62M",
    "address": {
      "address": "satellite.olio.lol:7777"
    },
    "type": "SATELLITE",
    "metadata": {
      "email": "jt@storj.io",
      "wallet": "0x0000000000000000000000000000000000000000"
    }
  }
}
```

jtolds:
Some of this PR has been merged, but other parts haven't. I'm leaving this open as a reminder that we have some outstanding security issues this PR fixes (Stream GRPC methods aren't checked for auth, for example)

egonelbre:
Closing since it's significantly out-of-date and will be worked in other issues as needed. The other issues can refer for this PR if they need particular snippets.

Issue #957, opened 2019-01-02
https://github.com/storj/storj/blob/master/pkg/kademlia/routing_helpers.go#L260 why append?


Fatfish90:
https://github.com/storj/storj/blob/master/pkg/kademlia/routing_helpers.go#L260 use append.
Why write like this:   xorArr[i] = id[i] ^ comparisonID[i]

Reference:   https://golang.org/src/crypto/cipher/xor.go

Fatfish90:
good, 👍.
The source code of append is: https://github.com/golang/go/blob/master/src/runtime/slice.go#L76
Actually the append func do a lot.

PR #958, opened 2019-01-02
pkg/kademlia: simplify code

egonelbre:
Each commit can be reviewed in isolation.

Fixes https://github.com/storj/storj/issues/957


pkg/kademlia/routing_helpers.go
```diff
@@ -23,10 +21,9 @@ import (
 func (rt *RoutingTable) addNode(node *pb.Node) (bool, error) {
 	rt.mutex.Lock()
 	defer rt.mutex.Unlock()
-	nodeIDBytes := node.Id.Bytes()
 
-	if bytes.Equal(nodeIDBytes, rt.self.Id.Bytes()) {
-		err := rt.createOrUpdateKBucket(rt.createFirstBucketID(), time.Now())
+	if node.Id == rt.self.Id {
```
coyle: why use `==` vs. `bytes.Equal` ?

PR #961, opened 2019-01-03
fix manipulated paychecks  V3-1023

aligeti:
https://storjlabs.atlassian.net/browse/V3-1023


pkg/bwagreement/test/common.go
```diff
@@ -84,3 +90,20 @@ func GenerateRenterBandwidthAllocation(pba *pb.PayerBandwidthAllocation, uplinkK
 		Data:      data,
 	}, nil
 }
+
+// get uplink's public key
+func getUplinkPubKey(uplinkKey crypto.PrivateKey) ([]byte, error) {
```
mobyvb: nit - rename to `getPublicKey` since this is a generic function. It can take in any `crypto.PrivateKey`, it does not need to be Uplink's

aligeti:
Will do. I wanted to fix the problem with minimal changes. If u approve it, I will open another ticket and do it.

PR #965, opened 2019-01-03
Differentiate GET and PUT operations in the tally service

wthorp:
https://storjlabs.atlassian.net/browse/V3-1036

Add the ability to differentiate GET and PUT operations (to a storage node) in the tally service
Differentiate between GET, PUT, GET repair, and GET Audit


pkg/accounting/tally/tally.go
```diff
@@ -153,18 +153,26 @@ func (t *tally) queryBW(ctx context.Context) error {
...
+	}
 	var latestBwa time.Time
 	for _, baRow := range bwAgreements {
 		rbad := &pb.RenterBandwidthAllocation_Data{}
 		if err := proto.Unmarshal(baRow.Agreement, rbad); err != nil {
 			t.logger.DPanic("Could not deserialize renter bwa in tally query")
 			continue
 		}
+		pbad := &pb.PayerBandwidthAllocation_Data{}
+		if err := proto.Unmarshal(rbad.GetPayerAllocation().GetData(), pbad); err != nil {
```
coyle: is it possible `rbad.GetPayerAllocation()` will return nil ?

wthorp: I'm not sure.  I think we shouldn't be storing agreements without payers.  I would hope we catch the kinds of things on communication edges, rather than checking continuously throughout our code.

PR #969, opened 2019-01-04
storj-sdk: tests

egonelbre:


cmd/storj-sdk/prefix.go
```diff
@@ -8,16 +8,26 @@ import (
...
-	len int
-	dst io.Writer
+	mu        sync.Mutex
+	prefixlen int
+	dst       io.Writer
+}
+
+const maxIDLength = 10
+
+func max(a, b int) int {
```
mniewrzal: Interesting that Go doesn't have `max` method for ints :)

cmd/storj-sdk/process.go
```diff
@@ -105,83 +112,162 @@ func (info *ProcessInfo) Env() []string {
...
+
+	// start the process
+	err = cmd.Start()
+	if err != nil {
+		return err
+	}
+
+	switch command {
+	case "setup":
+		// during setup we aren't starting the addresses, so we can release the depdendencies immediately
```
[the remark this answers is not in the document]
egonelbre: fixed

scripts/test-storj-sdk-aws.sh
```diff
@@ -0,0 +1,64 @@
...
+SRC_DIR=$TMPDIR/source
+DST_DIR=$(mktemp -d -t tmp.XXXXXXXXXX)
+mkdir -p $SRC_DIR $DST_DIR
+
+aws configure set aws_access_key_id     $GATEWAY_0_ACCESS_KEY
+aws configure set aws_secret_access_key $GATEWAY_0_SECRET_KEY
+aws configure set default.region        us-east-1
+
+head -c 1024 </dev/urandom > $SRC_DIR/small-upload-testfile      # create 1mb file of random bytes (inline)
+head -c 5120 </dev/urandom > $SRC_DIR/big-upload-testfile        # create 5mb file of random bytes (remote)
```
kaloyan-raev: Wouldn't it make sense to have the `big-upload-testfile` as 65mb, so it is larger than one segment?

PR #973, opened 2019-01-04
Added systemd script to service folder at the top level

phutchins:
This adds the systemd script for managing the storagenode as a service on linux hosts.


egonelbre:
Thinking whether these should be at the top-level. At the moment it seems that maybe managing them similarly how entrypoints and dockerfiles are handled seems more appropriate. So in `cmd/storagenode/storagenode.service`.

phutchins:
@egonelbre I think that would make sense if we had something that handled installing the service configs on the OS. Until then moving them deep into the file structure would make it more difficult for discovery.

How about we move the service folder into the example folder as thats a fairly common practice and will keep our top level clean. Once we automate the service installation process, if we decide to do that, we can move them along side their counterparts.

egonelbre:
@phutchins in example folder sounds nice as well. Yeah I was thinking of automatic installation as well.

egonelbre:
@phutchins that folder looks like `service/storagenode.service` and not `examples/systemd/storagenode.service` or `examples/service/storagenode.service`. Did you forget to commit/push?

phutchins:
@egonelbre ah yep! My push failed. Actually pushed this time. :)

PR #974, opened 2019-01-04
Integration test, travis quick-fix

bryanchriswhite:


.travis.yml
```diff
@@ -74,7 +74,7 @@ matrix:
         - make test-captplanet
 
     ### certificate signing tests ###
-    - env: MODE=integration
```
egonelbre: Let's merge these tests together with captplanet tests:
```
    ### Integration tests ###
    - env: MODE=integration
      services:
        - redis
      install:
        - source scripts/install-awscli.sh
        - go install -race storj.io/storj/cmd/{captplanet,storj-sdk,storagenode,satellite,uplink,gateway}
      script:
        - make test-captplanet
        - make test-certificate-signing
```

egonelbre: Also ensure that the `test-certificate-signing` is built using `-race`

bryanchriswhite: No objection here but do you care to explain why?

egonelbre: Time needing to wait before somebody can merge something. Every travis job eats up the resources meaning with multiple concurrent PR-s building we need to wait that much longer.

PR #979, opened 2019-01-05
prepare key generation for launch

jtolds:
this PR includes four commits:

 * first commit is a huge change - it changes id generation to be sha256 instead of sha3. keccak (sha3) is not only not asic-resistent, it's asic friendly, which is kind of not what you want for a PoW scheme. let's just use the tried and true thing.
 * second commit restructures the id generation code to prepare for the third commit
 * third commit is a tool that just generates keys until it's shut down, storing them indexed by their trailing zero count (the minimum difficulty they would pass)
 * fourth commit adds more difficulty tests

### DIFF
--- a/cmd/identity/keys.go
+++ b/cmd/identity/keys.go
@@ -1,0 +1,68 @@
+// Copyright (C) 2018 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package main
+
+import (
+	"crypto/ecdsa"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync/atomic"
+
+	"github.com/spf13/cobra"
+	"storj.io/storj/pkg/cfgstruct"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/process"
+	"storj.io/storj/pkg/storj"
+)
+
+var (
+	keysCmd = &cobra.Command{
+		Use:   "keys",
+		Short: "Manage keys",
+	}
+	keyGenerateCmd = &cobra.Command{
+		Use:   "generate",
+		Short: "generate lots of keys",
+		RunE:  cmdKeyGenerate,
+	}
+
+	keyCfg struct {
+		MinDifficulty int    `help:"minimum difficulty to output" default:"18"`
+		Concurrency   int    `help:"worker concurrency" default:"4"`
+		OutputDir     string `help:"output directory to place keys" default:"."`
+	}
+)
+
+func init() {
+	rootCmd.AddCommand(keysCmd)
+	keysCmd.AddCommand(keyGenerateCmd)
+	cfgstruct.Bind(keyGenerateCmd.Flags(), &keyCfg)
+}
+
+func cmdKeyGenerate(cmd *cobra.Command, args []string) (err error) {
+	ctx := process.Ctx(cmd)
+	err = os.MkdirAll(keyCfg.OutputDir, 0755)
+	if err != nil {
+		return err
+	}
+	counter := new(uint32)
+	return identity.GenerateKeys(ctx, uint16(keyCfg.MinDifficulty), keyCfg.Concurrency,
+		func(k *ecdsa.PrivateKey, id storj.NodeID) (done bool, err error) {
+			data, err := x509.MarshalECPrivateKey(k)
+			if err != nil {
+				return false, err
+			}
+			difficulty, err := id.Difficulty()
+			if err != nil {
+				return false, err
+			}
+			filename := fmt.Sprintf("gen-%02d-%d.key", difficulty, atomic.AddUint32(counter, 1))
+			fmt.Println("writing", filename)
+			err = ioutil.WriteFile(filepath.Join(keyCfg.OutputDir, filename), data, 0644)
+			return false, err
+		})
+}
--- a/cmd/identity/keys.go
+++ b/cmd/identity/keys.go
@@ -45,7 +45,7 @@ func init() {
 
 func cmdKeyGenerate(cmd *cobra.Command, args []string) (err error) {
 	ctx := process.Ctx(cmd)
-	err = os.MkdirAll(keyCfg.OutputDir, 0755)
+	err = os.MkdirAll(keyCfg.OutputDir, 0700)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


cmd/identity/keys.go
```diff
@@ -0,0 +1,68 @@
...
+			if err != nil {
+				return false, err
+			}
+			difficulty, err := id.Difficulty()
+			if err != nil {
+				return false, err
+			}
+			filename := fmt.Sprintf("gen-%02d-%d.key", difficulty, atomic.AddUint32(counter, 1))
+			fmt.Println("writing", filename)
+			err = ioutil.WriteFile(filepath.Join(keyCfg.OutputDir, filename), data, 0644)
```
bryanchriswhite: I assume we don't care about loose permissions here because these files are intended to be moved and renamed before use? 

Would it be worth tightening the permissions here to avoid the possibility of forgetting in the future when they'll actually be moved and used?

pkg/identity/generate.go
```diff
@@ -0,0 +1,73 @@
...
+			return k, id, nil
+		}
+	}
+	return k, id, storj.ErrNodeID.Wrap(err)
+}
+
+// GenerateKeys continues to generate keys until found returns done == false,
+// or the ctx is canceled.
+func GenerateKeys(ctx context.Context, minDifficulty uint16, concurrency int,
+	found func(*ecdsa.PrivateKey, storj.NodeID) (done bool, err error)) error {
```
bryanchriswhite: Would an options struct and/or a type alias for `found` make sense here? It's a pretty big function signature.

pkg/identity/identity.go
```diff
@@ -149,15 +150,24 @@ func PeerIdentityFromContext(ctx context.Context) (*PeerIdentity, error) {
...
+	// id = sha256(sha256(pkix(k)))
 	kb, err := x509.MarshalPKIXPublicKey(k)
 	if err != nil {
 		return storj.NodeID{}, storj.ErrNodeID.Wrap(err)
 	}
-	hash := make([]byte, len(storj.NodeID{}))
-	sha3.ShakeSum256(hash, kb)
-	return storj.NodeIDFromBytes(hash)
+	mid := sha256.Sum256(kb)
+	end := sha256.Sum256(mid[:])
```
bryanchriswhite: What's going on here?

pkg/storj/node.go
```diff
@@ -94,13 +94,15 @@ func (id NodeID) Less(b NodeID) bool {
...
-	for i := 1; i < idLen; i++ {
-		b := id[idLen-i]
+	var b byte
+	var zeroBits int
+	for i := 1; i <= idLen; i++ {
+		b = id[idLen-i]
 
 		if b != 0 {
-			zeroBits := bits.TrailingZeros16(uint16(b))
+			zeroBits = bits.TrailingZeros16(uint16(b))
```
bryanchriswhite: I see what you did there, nice

pkg/storj/node.go
```diff
@@ -94,13 +94,15 @@ func (id NodeID) Less(b NodeID) bool {
...
 			if zeroBits == 16 {
-				zeroBits = 0
+				panic("impossible codepath")
```
bryanchriswhite: This function signature was modified to add an error return on recent past in favor of a panic. Unless you plan on changing the signature back, why not return an error instead?

cmd/identity/keys.go
jtolds: oh, i didn't even think about this. 0644 is just habit. you're right, 0600 would be better

pkg/identity/generate.go
jtolds: i like the type alias idea. i'll do that

pkg/identity/identity.go
jtolds: just doin what satoshi did: https://bitcoin.stackexchange.com/questions/8443/where-is-double-hashing-performed-in-bitcoin. seems like it worked out okay. i can't talk myself out of being worried about hash length extension attacks myself

pkg/storj/node.go
jtolds: oh, good call!

cmd/identity/keys.go
```diff
@@ -0,0 +1,68 @@
...
+	cfgstruct.Bind(keyGenerateCmd.Flags(), &keyCfg)
+}
+
+func cmdKeyGenerate(cmd *cobra.Command, args []string) (err error) {
+	ctx := process.Ctx(cmd)
+	err = os.MkdirAll(keyCfg.OutputDir, 0755)
+	if err != nil {
+		return err
+	}
+	counter := new(uint32)
```
coyle: purely out of curiosity, why `counter := new(uint32)` vs. `var counter int32 = 0` ?

cmd/identity/keys.go
```diff
@@ -0,0 +1,68 @@
...
+				return false, err
+			}
+			difficulty, err := id.Difficulty()
+			if err != nil {
+				return false, err
+			}
+			filename := fmt.Sprintf("gen-%02d-%d.key", difficulty, atomic.AddUint32(counter, 1))
+			fmt.Println("writing", filename)
+			err = ioutil.WriteFile(filepath.Join(keyCfg.OutputDir, filename), data, 0644)
+			return false, err
```
coyle: `return true, err` ?

bryanchriswhite: According to the commit message, this command keeps generating until interrupted.

cmd/identity/keys.go
jtolds: i'm probably unreasonably worried about memory alignment issues:

https://golang.org/pkg/sync/atomic/#pkg-note-BUG says:
> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

this isn't 64 bit, so i don't really have a good reason here, but in general when using sync/atomic i've just been making it a habit

pkg/identity/generate.go
```diff
@@ -0,0 +1,73 @@
...
+				}
+				if done {
+					errchan <- nil
+					return
+				}
+			}
+		}()
+	}
+
+	return <-errchan
```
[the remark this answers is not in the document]
jtolds: no, cause all of the remaining errors will be context cancellation errors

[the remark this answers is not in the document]
egonelbre: add a note about ignoring the other errors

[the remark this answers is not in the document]
jtolds: kk

cmd/identity/keys.go
coyle: ahh ok what I get for not reading the commit messages thanks @bryanchriswhite

PR #981, opened 2019-01-07
kademlia routing table integration test suite

jtolds:
Over the weekend I noticed some weird behavior with Kademlia, so I started writing tests. 

This PR includes a new package called testrouting, which implements a very algorithmically slow but hopefully easy-to-keep-operationally-correct in-memory routing table. The routing table also supports writing out its current structure as a DOT graph for visualization. testrouting is primarily meant to help in coming up with generic routing table integration tests.

This PR also includes a new routing table integration test suite that runs against all current routing table implementations. Our existing routing table passes a lot of the tests, but not all of them, still debugging why. I have confirmed the tests should pass with the visualization graphs though.


jtolds:
ugh, just realized i have the tests assuming you should pull the oldest replacement cache entry first instead of the newest. i'll need to fix that

pkg/kademlia/routing_helpers_test.go
```diff
@@ -44,25 +55,29 @@ func newTestRoutingTable(localNode pb.Node) (*RoutingTable, error) {
...
 	if localNodeID == (storj.NodeID{}) {
 		localNodeID = teststorj.NodeIDFromString("AA")
 	}
 	localNode := pb.Node{Id: localNodeID}
 
-	rt, err := newTestRoutingTable(localNode)
+	rt, err := newTestRoutingTable(localNode, opts)
 	if err != nil {
-		t.Fatal(err)
+		panic(err)
```
coyle: what's the benefit of Fatal vs panic ?

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	rv := make([]*pb.Node, 0, limit)
+	for _, data := range nodes[:limit] {
+		rv = append(rv, data.node)
+	}
+	return rv, nil
+}
+
+// Local implements the dht.RoutingTable interface
+func (t *Table) Local() pb.Node {
```
coyle: In the interface definition we called `Local` what you are calling `Self` below. We also had it storing the Node and not just the ID.

pkg/kademlia/routing_helpers_test.go
jtolds: just that i don't have a *testing.T here anymore

egonelbre: When running tests in parallel then panic will end up in a random location between the test results.

pkg/kademlia/testrouting/testrouting.go
jtolds: right, so, this routing table doesn't have the node, because the routing table is actually going to be stale about self's address. i was saving this for another pr but i think we should change the interface of every routing table to not keep track of the self node, and only the self id.

jtolds:
> This looks really good. Just curious though, why re implement vs add tests to the existing implementation ?

well, so, this is adding tests to the existing implementation. the testrouting table was to help me understand what tests i was writing (it includes a visualizer). i don't think we should use testrouting in production (unless there are some significant changes i guess)

pkg/kademlia/testrouting/testrouting.go
jtolds: well, in the long term it will be the nat traversal system that will be responsible for keeping track of self's address. in the short term it's just the config system that knows the public address. planning for the public address changing is what i have in mind

coyle: well self is used for more than just address, we are using it right now for updating restrictions to be broadcast. Is there somewhere else we should keep track of metadata about a node for the network?

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,458 @@
...
+		NodeFromPrefix("50", "0"), NodeFromPrefix("58", "0"),
+		NodeFromPrefix("40", "0"), NodeFromPrefix("48", "0"),
+
+		// bucket 011 (same first two bits)
+		NodeFromPrefix("70", "0"), NodeFromPrefix("78", "0"),
+		NodeFromPrefix("60", "0"), NodeFromPrefix("68", "0"),
+
+		// bucket 00 (same first bit)
+		NodeFromPrefix("10", "0"),
+		NodeFromPrefix("00", "0"),
```
jenlij: Failing at NodeFromPrefix("00", "0") in TestSplits_Routing. Expects 1111111111111111111111111111111112m1s9K which represents an empty Node ID. Getting 1N96YXFr9Fc66JCJ7JdEzs2AxchSCHR2JptvQNysgkaDduN6LF instead.

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,458 @@
...
+
+		// bucket 1
+		NodeFromPrefix("c0", "0"),
+		NodeFromPrefix("f0", "0"),
+		NodeFromPrefix("88", "0"), // replacement cache
+		NodeFromPrefix("b8", "0"), // replacement cache
+	}, nodes)
+}
+
+func TestUnbalanced_Routing(t *testing.T)     { testUnbalanced(t, newRouting) }
```
jenlij: Getting 31 nodes rather than 32 expected

pkg/kademlia/routing_helpers_test.go
egonelbre: ah, wait I might be confusing it with data-race / fmt.Print

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,458 @@
...
+		Node(PadID("c0", "f"), "new-address:3")))
+
+	nodes, err = table.FindNear(PadID("c7139", "1"), 1)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(nodes))
+	require.Equal(t, PadID("c0", "f"), nodes[0].Id)
+	require.Equal(t, "new-address:3", nodes[0].Address.Address)
+}
+
+func TestUpdateCache_Routing(t *testing.T)     { testUpdateCache(t, newRouting) }
```
jenlij: Expected 3, got 2 items in cache

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,458 @@
...
+	require.NoError(t, err)
+
+	requireNodesEqual(t, []*pb.Node{
+		Node(PadID("00", "0"), "new-address:6"),
+		NodeFromPrefix("80", "0"),
+		NodeFromPrefix("c0", "0"),
+	}, nodes)
+}
+
+func TestFailureUnknownAddress_Routing(t *testing.T)     { testFailureUnknownAddress(t, newRouting) }
```
jenlij: Expected 3, got 2

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,458 @@
...
+	require.NoError(t, err)
+
+	requireNodesEqual(t, []*pb.Node{
+		Node(PadID("40", "0"), "address:2"),
+		NodeFromPrefix("80", "0"),
+		NodeFromPrefix("c0", "0"),
+	}, nodes)
+}
+
+func TestShrink_Routing(t *testing.T)     { testShrink(t, newRouting) }
```
jenlij: Error Trace:	routing_integration_test.go:435
        	            				routing_integration_test.go:415
        	Error:      	Received unexpected error:
        	            	routing table error: could not remove node routing table error: could not get k bucket routing table error: could not find k bucket
        	            		storj.io/storj/pkg/kademlia.(*RoutingTable).ConnectionFailed:211
        	            		storj.io/storj/pkg/kademlia.testShrink:435
        	            		storj.io/storj/pkg/kademlia.TestShrink_Routing:415
        	            		testing.tRunner:827
        	Test:       	TestShrink_Routing

pkg/kademlia/testrouting/testrouting.go
jtolds: i'm leaning towards yes. the main issue is that once we get to having automatic nat traversal, a node won't necessarily know what its public address even is. we should store metadata and node id and stuff somewhere, but having to store the address and potentially returning it in kademlia results has started to worry me

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,299 @@
...
+	"storj.io/storj/pkg/dht"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+type nodeData struct {
+	node      *pb.Node
+	ordering  int64
+	timestamp time.Time
```
egonelbre: Maybe, `lastConnected`, `updated` or `lastUpdated` or something similar?

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,299 @@
...
+	fails     int
+	inCache   bool
```
egonelbre: `cached`

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,299 @@
...
+	// to make sure we get the logic right, we're going to reconstruct the
+	// routing table binary tree data structure from first principles every time.
+	nodes := make([]*nodeData, 0, len(t.nodes))
+	for _, node := range t.nodes {
+		nodes = append(nodes, node)
+	}
+	var root bucket
+
+	// we'll replay the nodes in original placement order
+	sort.Sort(nodeDataOrderingSorter(nodes))
```
egonelbre: there's `sort.Slice`

pkg/kademlia/testrouting/viz.go
```diff
@@ -0,0 +1,66 @@
...
+package testrouting
+
+import (
+	"encoding/hex"
+	"fmt"
+	"io"
+)
+
+// Graph writes a DOT format visual graph description of the routing table to w
+func (t *Table) Graph(w io.Writer) error {
```
egonelbre: Serialize into `bytes.Buffer` then you don't have to worry about error handling and can use things like:

```
fmt.Fprintln(&buf, "digraph {\n  node [shape=box];\n")
defer fmt.Fprintln(&buf, "}")

...
```

egonelbre: alternatively a local func to hide error handling https://github.com/loov/layout/blob/master/format/dot/write.go#L10

pkg/kademlia/testrouting/viz.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = w.Write([]byte("}\n"))
+	return err
+}
+
+func (t *Table) graph(w io.Writer, b *bucket) error {
+	if b.split {
+		_, err := fmt.Fprintf(w, "b%s [label=\"%s\"];", b.prefix, b.prefix)
```
egonelbre: Instead of `\"%s\"` use `%q`.

pkg/kademlia/testrouting/viz.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		err = t.graph(w, b.dissimilar)
+		if err != nil {
+			return err
+		}
+		_, err = fmt.Fprintf(w, "b%s -> {b%s, b%s};",
+			b.prefix, b.similar.prefix, b.dissimilar.prefix)
+		return err
+	}
+	_, err := fmt.Fprintf(w, "b%s [label=\"%s\nrouting:\\l", b.prefix, b.prefix)
```
egonelbre: Use proper quoting.

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,299 @@
...
+		cell.fails = 0
+		// skip placement order and cache status
+		return nil
+	}
+
+	// add unconditionally (it might be going into a replacement cache)
+	t.nodes[node.Id] = &nodeData{
+		node:      node,
+		ordering:  t.counter,
+		timestamp: time.Now(),
```
egonelbre: Thinking: I wonder whether using logical time here might make more sense?

pkg/kademlia/routing_integration_test.go
jenlij: Either that, or the 0's ID

pkg/kademlia/routing_integration_test.go
jenlij: I think this error stems from the fact that we can't find the k bucket from the ID of all 0's.

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,606 @@
...
+func testSplits(t *testing.T, routingCtor routingCtor) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	table, close := routingCtor(PadID("55", "5"), 5, 2, 0)
+	defer close()
+
+	for _, prefix2 := range "08" {
+		for _, prefix1 := range "a69c23f1d7eb5408" {
+			require.NoError(t, table.ConnectionSuccess(
```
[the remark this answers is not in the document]
jtolds: unfortunately for this test, no. the order the nodes are added makes a significant impact to the structure of the table, as the rest of these tests verify.

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,606 @@
...
+func TestUnbalanced_Routing(t *testing.T)     { testUnbalanced(t, newRouting) }
+func TestUnbalanced_TestRouting(t *testing.T) { testUnbalanced(t, newTestRouting) }
+func testUnbalanced(t *testing.T, routingCtor routingCtor) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	table, close := routingCtor(PadID("ff", "f"), 5, 2, 0)
+	defer close()
+
+	for _, prefix1 := range "0123456789abcdef" {
```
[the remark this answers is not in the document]
jtolds: oh, that should be fine, though there is an Update test for that already

pkg/kademlia/testrouting/testrouting.go
jtolds: i keep forgetting about that!

pkg/kademlia/testrouting/viz.go
jtolds: kk

pkg/kademlia/testrouting/viz.go
jtolds: hmm, i wonder what the escape sequence language is in dot. it's probably not the same as go's but sure, this is strictly better

pkg/kademlia/testrouting/viz.go
jtolds: that's going to add another fmt.Sprintf call, so i can get the result of sprintf inside the quoted label

pkg/kademlia/testrouting/testrouting.go
jtolds: that's what the ordering field is

pkg/kademlia/routing_helpers_test.go
```diff
@@ -44,25 +55,29 @@ func newTestRoutingTable(localNode pb.Node) (*RoutingTable, error) {
 	return rt, nil
 }
 
-func createRoutingTable(t *testing.T, localNodeID storj.NodeID) (*RoutingTable, func()) {
+func createRoutingTableWith(localNodeID storj.NodeID, opts routingTableOpts) (*RoutingTable, func()) {
```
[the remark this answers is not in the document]
jtolds: fair

pkg/kademlia/routing_helpers_test.go
```diff
@@ -44,25 +55,29 @@ func newTestRoutingTable(localNode pb.Node) (*RoutingTable, error) {
...
 	if localNodeID == (storj.NodeID{}) {
 		localNodeID = teststorj.NodeIDFromString("AA")
```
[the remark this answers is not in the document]
jtolds: oh, yeah okay

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,606 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package kademlia
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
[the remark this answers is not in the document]
jtolds: i just learned about it too! i was poking around the testify documentation to try and figure out how to make my life easier

pkg/kademlia/testrouting/viz.go
jtolds: in retrospect i'm embarrassed by my increased likelihoods of sql injection style problems here, but in my defense, b.prefix is always alphanumeric

pkg/kademlia/routing_helpers_test.go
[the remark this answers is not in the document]
jtolds: changing this signature makes this PR touch way more files and lines. i'm going to hold off for another PR

pkg/kademlia/routing_helpers_test.go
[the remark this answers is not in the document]
jtolds: this also is turning into another largish separate pr.

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+	require.NoError(t, table.ConnectionFailed(NodeFromPrefix("c0", "0")))
+
+	nodes, err = table.FindNear(PadID("55", "55"), 9)
+	require.NoError(t, err)
+	requireNodesEqual(t, []*pb.Node{
+		NodeFromPrefix("d0", "0"),
+		NodeFromPrefix("e0", "0"),
+	}, nodes)
+}
```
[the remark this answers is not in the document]
jtolds: i want to do randomized testing for sure. let's do that in another pr

pkg/kademlia/testrouting/testrouting.go
jtolds: lastUpdated is good

pkg/kademlia/testrouting/testrouting.go
jtolds: i think that's less clear. it's not like the node is cached.

pkg/kademlia/testrouting/viz.go
jtolds: actually this is pretty tricky with the \\ls and multiple levels of printf and so on. because we're building a label and the prefix can't be anything but alphanumeric, i'm going to leave this as is with a comment

pkg/kademlia/testrouting/testrouting.go
jtolds: how do you feel about storing self in the routing table struct, but not in the routing table itself? the routing table interface could support updating and retrieving the self node, but it wouldn't get returned in node results or queries?

pkg/kademlia/routing_integration_test.go
jtolds: okay latest changes no longer use all zeros

pkg/kademlia/routing_integration_test.go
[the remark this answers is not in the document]
jtolds: okay to leave this for the specific Update test?

pkg/kademlia/testrouting/viz.go
egonelbre: I'm aware. The problem is that people copy-paste code. i.e. it's better to use `"%q"` or `strconv.Quote`... a comment is also fine, but it should be clear either that "it's properly quoting" or the value is sanitized, e.g. `sanitized(b.prefix)` or something.

It's more about good habits rather than this particular issue.

egonelbre: As for dot syntax, it's such that Go quotation should work:

```
_escaped_char : '\\' ( _unicode_char | '"' | '\\' ) ;
_char         : _unicode_char | _escaped_char ;
_string_lit   : '"' { _char } '"' ;
```

pkg/kademlia/testrouting/testrouting.go
coyle: Yea that's fine with me. Maybe Routing Table is an all together bad place? maybe just on the top level Kademlia struct ? We just need a `Self` somewhere to track data about a node.

egonelbre:
Closed because of https://github.com/storj/storj/pull/1270

PR #983, opened 2019-01-07
sync2: add Fence and Sleep

egonelbre:
Added fence and sleep for proper waiting between processes. Needed for storj-sdk.

### DIFF
--- a/internal/sync2/fence_test.go
+++ b/internal/sync2/fence_test.go
@@ -1,0 +1,45 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package sync2_test
+
+import (
+	"errors"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	"storj.io/storj/internal/sync2"
+)
+
+func TestFence(t *testing.T) {
+	var group errgroup.Group
+	var fence sync2.Fence
+	var done int32
+
+	for i := 0; i < 10; i++ {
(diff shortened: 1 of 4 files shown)


internal/sync2/sleep_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestSleep(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	start := time.Now()
+	if !sync2.Sleep(ctx, time.Second) {
+		t.Error("expected true as result")
+	}
+	if time.Since(start) < time.Second/2 {
```
kaloyan-raev: Why checking if the sleep took 500 ms instead of 1 sec? Aren't we expecting a full second of sleep?

egonelbre: Sleep is somewhat annoying that it can sleep less than a second as well. It has an +- error in terms of system tick granularity.

egonelbre: added a bit clarity to it

internal/sync2/fence_test.go
```diff
@@ -0,0 +1,45 @@
...
+		group.Go(func() error {
+			fence.Wait()
+			if atomic.LoadInt32(&done) == 0 {
+				return errors.New("fence not yet released")
+			}
+			return nil
+		})
+	}
+
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
egonelbre: made it 100 milliseconds

PR #984, opened 2019-01-07
Upgrade monkit-hw

egonelbre:


AlexandrTyshchenko:
@egonelbre Please upgrade monkit-hw to the latest version https://github.com/jtolds/monkit-hw

PR #985, opened 2019-01-07
Get satellites to send BW agreements to via KAD

wthorp:
https://storjlabs.atlassian.net/browse/V3-1035

Storage node should get info for satellites to send BW agreements to via KAD network instead of hard coded endpoint

Expected Behavior:

The Storage nodes are able to send the BW agreements they have to the correct satellites because they are getting the satellites IP/ port from the KAD network instead of hitting the hard-coded endpoint


wthorp:
After reviewing the code, I could not find a list of hard coded satellites associated with a storage node.  I did clean up agreeementsender.go somewhat to use transport and local logging.

mniewrzal:
@littleskunk From what I understand you saw hardcoded satellite ip in storage node config but I think you are referring to `piecestore.agreementsender.overlay-addr` which is a param for overlay client. BA agreements are sent to satellite after checking satellite id (from ba agreement) in overlay. IP is taken from overlay.

littleskunk:
Yes. My test would be a network with 2 satellites and 100 storage nodes. Now take the first satellite offline that is hardcoded in the storage node config. Is the storage node still able to deliver bwagreements to the second satellite? I don't think so. If that is correct we would have to deal with some additional questions. Can we make sure the primary satellite is always online? A single point of failure? Can we call that decentralized?

mniewrzal:
@littleskunk Yes, you are right. If service under `piecestore.agreementsender.overlay-addr` will go offline then storage node wont be able to send agreements to specific satellite. I think we should create another ticket because this problem most probably is beond this PR.

wthorp:
I've rewritten it to use Kademlia to find satellite addresses based on their node ids.  This assumes one global Kademlia network and N satellites each having their own overlay cache.  I'm curious if this is the intended design, or emergent architecture.

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -10,121 +10,90 @@ import (
...
 
-					// Delete from PSDB by signature
-					if err = as.DB.DeleteBandwidthAllocationBySignature(agreement.Signature); err != nil {
-						zap.S().Error(err)
-						return
-					}
-				}
-			}()
+func (as *AgreementSender) sendAgreementsToSatellite(ctx context.Context, satID storj.NodeID, agreements []*psdb.Agreement) {
+	as.log.Info("Sending agreements to satellite", zap.Int("number of agreements", len(agreements)), zap.String("sat node id", satID.String()))
```
littleskunk: Please change `sat node id` into `satellite id` or `satellite node id`

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -10,121 +10,96 @@ import (
...
-	zap.S().Info("AgreementSender is starting up")
-
-	type agreementGroup struct {
-		satellite  storj.NodeID
-		agreements []*psdb.Agreement
-	}
-
-	c := make(chan *agreementGroup, 1)
-
+// Run the afreement sender with a context to check for cancel
```
mniewrzal: typo `afreement`

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -10,121 +10,96 @@ import (
...
 
-					// Send agreement to satellite
-					r, err := client.BandwidthAgreements(ctx, msg)
-					if err != nil || r.GetStatus() != pb.AgreementsSummary_OK {
-						zap.S().Errorf("Failed to send agreement to satellite: %+v", err)
-						return
-					}
+func (as *AgreementSender) sendAgreementsToSatellite(ctx context.Context, satID storj.NodeID, agreements []*psdb.Agreement) {
+	as.log.Info("Sending agreements to satellite", zap.Int("number of agreements", len(agreements)), zap.String("sat node id", satID.String()))
+	// Get satellite ip from overlay by Lookup satellite
```
mniewrzal: comment needs adjustment

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -10,121 +10,96 @@ import (
...
+	}
+}
 
-					// Send agreement to satellite
-					r, err := client.BandwidthAgreements(ctx, msg)
-					if err != nil || r.GetStatus() != pb.AgreementsSummary_OK {
-						zap.S().Errorf("Failed to send agreement to satellite: %+v", err)
-						return
-					}
+func (as *AgreementSender) sendAgreementsToSatellite(ctx context.Context, satID storj.NodeID, agreements []*psdb.Agreement) {
```
mniewrzal: I'm wondering if wouldn't be more cleaner to add error to function response and use `ASError` to wrap things but not 100% sure.

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -10,121 +10,96 @@ import (
...
-// Initialize the Agreement Sender
-func Initialize(DB *psdb.DB, identity *provider.FullIdentity) (*AgreementSender, error) {
-	overlay, err := overlay.NewClient(identity, *defaultOverlayAddr)
-	if err != nil {
-		return nil, err
-	}
-
-	return &AgreementSender{DB: DB, identity: identity, overlay: overlay}, nil
+// New creates an Agreement Sender
+func New(log *zap.Logger, DB *psdb.DB, identity *provider.FullIdentity, kad *kademlia.Kademlia) *AgreementSender {
```
mniewrzal: It should be possible to use a general interface `dht.DHT` instead of Kademlia directly.

egonelbre: `dht.DHT` will be removed to simplify the code.

mniewrzal: oh, I saw it in recent JT code and though its current design :) thanks for info

pkg/piecestore/psserver/agreementsender/agreementsender.go
wthorp: I just tried refactoring to that, and it traded a few simple lines of repeated code for added complexity.  I decided to leave it for fear of introducing new bugs.

mniewrzal: :+1:

PR #988, opened 2019-01-08
WIP pdbclient shouldn't be remembering authorization

wthorp:
https://storjlabs.atlassian.net/browse/V3-868
Remove `authorization` field from inside pdbclient and figure out how the rest of code needs to be adjusted.


mniewrzal:
Main problem with `SignedMessage` is in store.go `Put` method. There is no direct call to any `pdb` methods because first we search for nodes to store, second we store data with ecclient and finally we are putting metadata into `pba`. We need `SignedMesseage` the second step.

https://github.com/storj/storj/blob/e4cb0d479a52261aec6f37a644d1d59667bbdf7d/pkg/storage/segments/store.go#L138

wthorp:
"Storage nodes will support three methods: get, put, and delete. Each method will
take a piece ID, a Satellite ID, a signature from the associated Satellite instance, and a
bandwidth allocation" - it sounds like we need to get everything signed by the satellite for a storage node to accept it?

jtolds:
> it sounds like we need to get everything signed by the satellite for a storage node to accept it?
yeah, the thinking was we could put that in the bandwidth allocation itself to cut down on public/private key crypto signature cost

PR #992, opened 2019-01-08
normalize login email

rikysya:
transform email to lower case before user creating, updating, authorization

### DIFF
--- a/pkg/satellite/validation.go
+++ b/pkg/satellite/validation.go
@@ -80,3 +80,23 @@ func validatePassword(pass string) error {
 
 	return errs.Combine()
 }
+
+// toLowerCase converts uppercase runes to lowercase equivalents
+// and returns resulting string
+func toLowerCase(s string) string {
+	if s == "" {
+		return s
+	}
+
+	var result []rune
+	for _, r := range s {
+		if unicode.IsUpper(r) {
+			result = append(result, unicode.SimpleFold(r))
+			continue
+		}
+
+		result = append(result, r)
+	}
+
+	return string(result)
+}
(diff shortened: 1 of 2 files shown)


pkg/satellite/validation.go
```diff
@@ -80,3 +80,23 @@ func validatePassword(pass string) error {
...
+
+	var result []rune
+	for _, r := range s {
+		if unicode.IsUpper(r) {
+			result = append(result, unicode.SimpleFold(r))
+			continue
+		}
+
+		result = append(result, r)
+	}
```
egonelbre:
```
    runes := []rune(s)
    for i, r := range runes { 
        runes[i] = unicode.SimpleFold(r)
    }
    return string(runes)
```

pkg/satellite/validation.go
```diff
@@ -80,3 +80,23 @@ func validatePassword(pass string) error {
 
 	return errs.Combine()
 }
+
+// toLowerCase converts uppercase runes to lowercase equivalents
+// and returns resulting string
+func toLowerCase(s string) string {
```
egonelbre: Rename to `normalizeEmail`. This is not for lower-casing the email.

wthorp: Unfortunately, email internationalization and unicode folding seems sufficiently bad as to warrant a clear description of the intent and possibly test cases.

The RFC 5321 says "The local-part of a mailbox MUST BE treated as case sensitive," but I believe assumes ASCII.  RFC 839 lowercases only ASCII.  Google removes periods and employs RFC 5233 subaddressing my.email+junk@gmail.com -- myemail@gmail.com

https://en.wikipedia.org/wiki/Email_address

Possibly see also "golang.org/x/text/cases"

PR #993, opened 2019-01-08
Enable import grouping checks

egonelbre:


scripts/check-imports.go
```diff
@@ -69,18 +73,36 @@ func main() {
...
 }
 
-func process(pkg *packages.Package) {
+func correctPackage(pkg *packages.Package) bool {
+	correct := true
 	for i, file := range pkg.Syntax {
-		checkImports(pkg.Fset, pkg.CompiledGoFiles[i], file)
+		path := pkg.CompiledGoFiles[i]
+		if !correctImports(pkg.Fset, path, file) {
+			if !isGenerated(path) { // ignore generated files
```
kaloyan-raev: Should we ignore generated files before doing the actual check for correct imports?

kaloyan-raev: OK

PR #997, opened 2019-01-08
Improve cert signing integration test+:

bryanchriswhite:
+ improve bash utils/test
+ add verify command
+ reorganize certificates commands
+ reorganize certficate configs
+ refactor certificate backup


pkg/certificates/config.go
```diff
@@ -0,0 +1,206 @@
...
+		err = errs.Combine(err, authDB.Close())
+	}()
+
+	signer, err := c.CA.Load()
+	if err != nil {
+		return err
+	}
+
+	srv := &CertificateSigner{
+		Log:           zap.L(),
```
mobyvb: does `Log` need to be public? In most other stuff I've seen it declared as lowercase.

scripts/test-certificate-signing.sh
```diff
@@ -3,10 +3,10 @@ set -o errexit
```
mobyvb: is travis currently running this test?

cmd/certificates/auth.go
```diff
@@ -0,0 +1,273 @@
...
+	authCmd.AddCommand(authCreateCmd)
+	cfgstruct.Bind(authCreateCmd.Flags(), &authCreateCfg, cfgstruct.ConfDir(defaultConfDir))
+	authCmd.AddCommand(authInfoCmd)
+	cfgstruct.Bind(authInfoCmd.Flags(), &authInfoCfg, cfgstruct.ConfDir(defaultConfDir))
+	authCmd.AddCommand(authExportCmd)
+	cfgstruct.Bind(authExportCmd.Flags(), &authExportCfg, cfgstruct.ConfDir(defaultConfDir))
+}
+
+func cmdCreateAuth(cmd *cobra.Command, args []string) error {
+	count, err := strconv.Atoi(args[0])
```
coyle: is there any concern args will ever be empty?

stefanbenten: He specified a minimum above in L34. That should prevent it from being empty

pkg/certificates/config.go
```diff
@@ -0,0 +1,206 @@
+// Copyright (C) 2018 Storj Labs, Inc.
```
mniewrzal: 2019

pkg/certificates/config.go
```diff
@@ -0,0 +1,206 @@
...
+		defer cancel()
+
+		ca, err = caConfig.Create(ctx)
+		if err != nil {
+			return err
+		}
+	}
+
+	identStatus := identConfig.Status()
+	if identStatus == identity.CertKey && !identConfig.Overwrite {
```
mniewrzal: nit - switch can be used here

cmd/certificates/setup.go
```diff
@@ -0,0 +1,91 @@
+// Copyright (C) 2018 Storj Labs, Inc.
```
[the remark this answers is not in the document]
mniewrzal: I believe that code created in 2018 needs to stay with date of creation.

cmd/certificates/verify.go
```diff
@@ -0,0 +1,145 @@
...
+		RunE: cmdVerify,
+	}
+
+	verifyCfg verifyConfig
+)
+
+type checkOpts struct {
+	verifyConfig
+	ca       *identity.FullCertificateAuthority
+	ident    *identity.FullIdentity
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not convinced that's necessary.. I've consistently used `ident` for variables; this is an non-exported type and I think the `ident` is a decent compromise between clarity and conciseness.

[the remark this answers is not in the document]
egonelbre: I would go with `identity` as well.

pkg/certificates/config.go
bryanchriswhite: How? The conditions use logical and in and negation. I don't see a way to reduce this to a switch.

mniewrzal: An example https://github.com/storj/storj/blob/8893884044bcbe636c093e64df2931a5fd6b87b5/pkg/piecestore/psserver/server.go#L270

egonelbre:
```
switch {
case identStatus == identity.CertKey && !identConfig.Overwrite:
...
case identStatus != identity.NoCertNoKey && !identConfig.Overwrite:
...
default:
...
}

File: cmd/certificates/setup.go [hunk 8]
```diff
@@ -0,0 +1,91 @@
...
+	setupCmd = &cobra.Command{
+		Use:         "setup",
+		Short:       "Setup a certificate signing server",
+		RunE:        cmdSetup,
+		Annotations: map[string]string{"type": "setup"},
+	}
+
+	setupCfg struct {
+		// NB: cert and key paths overridden in setup
+		CA identity.CASetupConfig
```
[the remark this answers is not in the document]
bryanchriswhite: Thanks for the info on `cfgstruct.BindSetup` and the `setup:"true"` tag, I was not aware of that. These fields are all used in `run` as well, the certificates service needs an identity as it is a grpc service.

PR #1003, opened 2019-01-09
Cache refresher

dylanlott:
* Adds cache refresher to Discovery 
* Updates context in testplanet tests to correctly pass node context


pkg/discovery/service.go
```diff
@@ -43,12 +44,17 @@ func NewDiscovery(logger *zap.Logger, ol *overlay.Cache, kad *kademlia.Kademlia,
...
 	nodes := d.kad.Seen()
 
-	for _, v := range nodes {
-		if err := d.cache.Put(ctx, v.Id, *v); err != nil {
-			return err
+	for _, node := range nodes {
+		if _, err := d.kad.Ping(ctx, *node); err != nil {
+			// fail ping refresh
+			d.log.Warn(fmt.Sprintf("unable to ping node %+v\n", node.Id))
+			_, err = d.statdb.UpdateUptime(ctx, node.Id, false)
```
wthorp: err needs checked for linting

pkg/discovery/service.go
```diff
@@ -43,12 +44,17 @@ func NewDiscovery(logger *zap.Logger, ol *overlay.Cache, kad *kademlia.Kademlia,
...
+		} else {
+			// succeed ping refresh
+			d.log.Info(fmt.Sprintf("successfully contacted node %s", node.Id))
+			_, err = d.statdb.UpdateUptime(ctx, node.Id, true)
```
wthorp: err needs checked for linting

pkg/discovery/config.go
```diff
@@ -55,16 +55,25 @@ func (c Config) Run(ctx context.Context, server *provider.Provider) (err error)
...
+	zap.L().Debug("Starting discovery")
 	ticker := time.NewTicker(c.RefreshInterval)
 	defer ticker.Stop()
 
 	go func() {
 		for {
 			select {
 			case <-ticker.C:
-				err := discovery.Refresh(ctx)
+				err := discovery.Bootstrap(ctx)
```
wthorp: Have we changed the meaning of bootstrap by calling it regularly?  Do we always call bootstrap now before a refresh?  Should we move the call to Bootstrap to within Refresh to avoid naming confusion?

internal/testplanet/planet.go
```diff
@@ -102,6 +102,7 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 
 	// init Satellites
 	for _, node := range planet.Satellites {
+		ctx := context.Background()
```
egonelbre: Nothing in New should require a context. If it does then the logic is in the wrong place.

See "Service Lifecycle Simplification".

internal/testplanet/planet.go
```diff
@@ -115,7 +116,7 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 		pb.RegisterPointerDBServer(node.Provider.GRPC(), pointerServer)
 		// bootstrap satellite kademlia node
 		go func(n *Node) {
-			if err := n.Kademlia.Bootstrap(context.Background()); err != nil {
+			if err := n.Kademlia.Bootstrap(ctx); err != nil {
```
egonelbre: Make it a `context.TODO()` instead.

internal/testplanet/planet.go
```diff
@@ -137,11 +138,17 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 			}))
 
 		go func(n *Node) {
+			if err := n.Discovery.Bootstrap(ctx); err != nil {
```
egonelbre: What's the difference between Bootstrap and Refresh? Why does it keep calling Bootstrap.

internal/testplanet/planet.go
```diff
@@ -150,9 +157,10 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 
 	// init storage nodes
 	for _, node := range planet.StorageNodes {
+		ctx := context.Background()
 		storageDir := filepath.Join(planet.directory, node.ID().String())
 
-		serverdb, err := psdb.OpenInMemory(context.Background(), storageDir)
+		serverdb, err := psdb.OpenInMemory(ctx, storageDir)
```
egonelbre: make it a `context.TODO()`

internal/testplanet/planet.go
```diff
@@ -167,11 +175,11 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 
 		node.Dependencies = append(node.Dependencies,
 			closerFunc(func() error {
-				return server.Stop(context.Background())
+				return server.Stop(ctx)
```
egonelbre: make it a `context.TODO()`

internal/testplanet/planet.go
```diff
@@ -167,11 +175,11 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
...
 			}))
 		// bootstrap all the kademlia nodes
 		go func(n *Node) {
-			if err := n.Kademlia.Bootstrap(context.Background()); err != nil {
+			if err := n.Kademlia.Bootstrap(ctx); err != nil {
```
egonelbre: make it a `context.TODO()`

Why does it need so much bootstrapping?

pkg/discovery/service.go
```diff
@@ -43,12 +44,17 @@ func NewDiscovery(logger *zap.Logger, ol *overlay.Cache, kad *kademlia.Kademlia,
...
-		if err := d.cache.Put(ctx, v.Id, *v); err != nil {
-			return err
+	for _, node := range nodes {
+		if _, err := d.kad.Ping(ctx, *node); err != nil {
+			// fail ping refresh
+			d.log.Warn(fmt.Sprintf("unable to ping node %+v\n", node.Id))
+			_, err = d.statdb.UpdateUptime(ctx, node.Id, false)
+		} else {
+			// succeed ping refresh
+			d.log.Info(fmt.Sprintf("successfully contacted node %s", node.Id))
```
egonelbre: This shouldn't be logged and definitely with `Info` level.

pkg/discovery/service.go
egonelbre: err need sto be checked.

pkg/discovery/service.go
```diff
@@ -57,14 +63,14 @@ func (d *Discovery) Refresh(ctx context.Context) error {
 
 // Bootstrap walks the initialized network and populates the cache
```
egonelbre: It doesn't look like it's walking the initialized cache.

pkg/discovery/config.go
dylanlott: I fixed this with that context bug fix I mentioned.

internal/testplanet/planet.go
dylanlott: This was due to a bug that I fixed. I had it pushed up like this to demonstrate the bug.

internal/testplanet/planet.go
dylanlott: It was due to a bug with Kademlia StartRefresh that is since fixed.

pkg/discovery/service.go
dylanlott: Fixed

internal/testplanet/planet.go
dylanlott: These tests need a consistent context to load the Cache and Kademlia from, or else the cache is an empty pointer and I get a goroutine error. I can hoist the Context creation up in to a higher function, but this feels like a problem with the use of Context.

dylanlott: That feels like it should be a different PR.

wthorp: All Egon is saying is 'move stuff with ctx to startup()'
I agree, there's already a loop in Startup() for Kademlia's StartRefresh() there.  At minimum, its wonky to have this into two different places.

internal/testplanet/planet.go
```diff
@@ -158,11 +159,17 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
 			}))
 
 		go func(n *Node) {
```
wthorp: I'd like to see this whole go routine and timer moved into Discovery for reusability, ala kad's StartRefresh()

internal/testplanet/planet.go
```diff
@@ -158,11 +159,17 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
...
+			if err := n.Discovery.Bootstrap(ctx); err != nil {
+				log.Error(err.Error())
+			}
 			// refresh the interval every 500ms
 			t := time.NewTicker(500 * time.Millisecond).C
```
jenlij: why 500ms?

dylanlott: I'm not sure, I actually didn't set that.

internal/testplanet/planet.go
dylanlott: Moved. 👍

internal/testplanet/planet.go
dylanlott: Okay, I moved all the context stuff to be passed down through args from the actual test setup.

pkg/discovery/service.go
```diff
@@ -39,32 +41,57 @@ func NewDiscovery(logger *zap.Logger, ol *overlay.Cache, kad *kademlia.Kademlia,
...
+		t := time.NewTicker(500 * time.Millisecond).C
+		for {
+			<-t
+			if err := d.Refresh(ctx); err != nil {
+				d.log.Error(err.Error())
+				errs = append(errs, err)
+			}
+		}
+	}()
+	return utils.CombineErrors(errs...)
```
wthorp: the for loop can never exit, an hence this never gets called... we're just making a potentially infinite list of errors in the meantime

PR #1005, opened 2019-01-09
Storage Node Peer

egonelbre:


storagenode/peer.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Config is all the configuration parameters for a Storage Node
+type Config struct {
+	Identity identity.Config
+
+	// TODO: switch to using server.Config when Identity has been removed from it
+	PublicAddress string `help:"public address to listen on" default:":7777"`
+	Kademlia      kademlia.Config
+	Piecestore    psserver.Config
```
mniewrzal: For sure `Piecestore`? currently we are using `Storage`. Or my understanding wrong and this won't be used with cmd later.

storagenode/peer.go
```diff
@@ -0,0 +1,211 @@
...
+
+	var err error
+
+	{ // setup listener and server
+		peer.Public.Listener, err = net.Listen("tcp", config.PublicAddress)
+		if err != nil {
+			return nil, errs.Combine(err, peer.Close())
+		}
+
+		publicConfig := provider.ServerConfig{Address: peer.Public.Listener.Addr().String()}
```
mniewrzal: Why using translation than direct `server.Config`?

egonelbre: Because I forgot to change it. Will fix.

storagenode/peer.go
egonelbre: Yeah, storage makes more sense

egonelbre: Fixed

storagenode/peer.go
egonelbre: fixed

PR #1009, opened 2019-01-09
api keys markup updated

NikolaiYurchenko:


web/satellite/src/components/apiKeys/CopyApiKeyPopup.vue
```diff
@@ -0,0 +1,258 @@
...
+        },
+        data: function () {
+            return {
+                imageSource: EMPTY_STATE_IMAGES.ADD_API_KEY,
+            };
+        },
+        methods: {
+            onCloseClick: function (): void {
+                // TODO: save popup states in store
+                this.$emit('onClose');
```
crawter: should be closed via action from app_state module

PR #1011, opened 2019-01-09
[V3-1031] added front end validation for project name

NikolaiYurchenko:


web/satellite/src/components/project/NewProjectPopup.vue
```diff
@@ -86,15 +82,17 @@ import { APP_STATE_ACTIONS, NOTIFICATION_ACTIONS, PROJETS_ACTIONS } from '@/util
                 this.$store.dispatch(APP_STATE_ACTIONS.TOGGLE_NEW_PROJ);
             },
             createProject: async function (): Promise<any> {
+                let projectName = this.$data.projectName.trim();
```
Barterio: const instead let

PR #1015, opened 2019-01-09
satellites reject expired BWAs

cam-a:
https://storjlabs.atlassian.net/browse/V3-1048


pkg/bwagreement/test/common.go
```diff
@@ -36,12 +36,20 @@ func GeneratePayerBandwidthAllocation(action pb.PayerBandwidthAllocation_Action,
 		return nil, err
 	}
 
+	// generate expiration date
+	var exp int64
+	if expired {
+		exp = time.Now().AddDate(0, 0, -5).Unix()
+	} else {
+		exp = time.Now().Add(time.Hour * 24 * 10).Unix()
```
jenlij: expires in 10 days?

satellite/satellitedb/bandwidthagreement.go
```diff
@@ -56,3 +57,7 @@ func (b *bandwidthagreement) GetAgreementsSince(ctx context.Context, since time.
 	}
 	return agreements, nil
 }
+
+func (b *bandwidthagreement) DeleteExpired(ctx context.Context) error {
+	return Error.New("DeleteExpired not implemented")
```
jenlij: add a TODO comment here in case we grep for that keyword

pkg/bwagreement/test/common.go
cam-a: Yeah, that's the test was using before I added the ability to generate an expired date

pkg/pointerdb/config.go
```diff
@@ -34,6 +34,7 @@ type Config struct {
 	MinRemoteSegmentSize int    `default:"1240" help:"minimum remote segment size"`
 	MaxInlineSegmentSize int    `default:"8000" help:"maximum inline segment size"`
 	Overlay              bool   `default:"true" help:"toggle flag if overlay is enabled"`
+	BWATTL               int    `default:"45"   help:"lifespan of bandwidth agreements in days"`
```
aligeti: rename to BwExpiration

aleitner: use time.Duration instead of int

aleitner: https://golang.org/pkg/time/#ParseDuration

PR #1013, opened 2019-01-09
Clean up Storage node setup

aleitner:
* Edit config on setup
* Default StorageAvailable: 1TiB (https://storjlabs.atlassian.net/browse/V3-1043)
* Default BandwidthAvailable: 500GiB (https://storjlabs.atlassian.net/browse/V3-1043)

*WARNING BACKWARDS INCOMPATIBLE*

PR #1018, opened 2019-01-10
pkg/cfgstruct: allow values to be configured with human readable sizes

jtolds:


jtolds:
merged into https://github.com/storj/storj/pull/1013

PR #1020, opened 2019-01-10
continue agreement sender loop on error

wthorp:
Jens Found a few issues ... one was my fault due to last minute refactoring

If the storagenode is unable to deliver one bwagreement it will stop sending any other bwagreement as well. Exit and try again one hour later with the same error message. I would expect skipping the bad bwagreement and go on with the next one.


mniewrzal:
Wont this result with sending invalid agreements over and over? maybe we need max number of retries?

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -93,13 +94,17 @@ func (as *AgreementSender) sendAgreementsToSatellite(ctx context.Context, satID
 		// Send agreement to satellite
 		r, err := client.BandwidthAgreements(ctx, msg)
 		if err != nil || r.GetStatus() != pb.AgreementsSummary_OK {
-			as.log.Error("Agreementsender failed to send agreement to satellite", zap.Error(err))
-			return
+			if bwagreement.BwAgreementError.Has(err) {
```
mniewrzal: Should we include package that is specific for satellite? @egonelbre

egonelbre: definitely not

PR #1021, opened 2019-01-10
Fix captplanet race

egonelbre:


cmd/captplanet/run.go
```diff
@@ -79,7 +79,7 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		}
 
 		//nolint ignoring context rules to not create cyclic dependency, will be removed later
-		ctx = context.WithValue(ctx, "masterdb", database)
+		ctx := context.WithValue(ctx, "masterdb", database)
```
wthorp: nit: maybe give ctx a new name here?

PR #1025, opened 2019-01-10
V3-978 Create missing unit tests for satellitedb

crawter:


pkg/satellite/satelliteauth/claims_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	claims := Claims{
+		ID:         *id,
+		Email:      "someEmail@ukr.net",
+		Expiration: time.Now(),
+	}
+
+	claimsBytes, err := claims.JSON()
+	assert.Nil(t, err)
+	assert.NoError(t, err)
```
mobyvb: Aren't `assert.Nil(t, err)` and `assert.NoError(t, err)` redundant?

crawter: @mobyvb yeah, it do the same, u're right

PR #1026, opened 2019-01-10
Divide uplink and gateway params set

mniewrzal:
With this PR:
* `gateway` command has own set of parameters and its config file is stored by default in `gateway` directory
* `uplink` command has own set of parameters and its config file is stored by default in `uplink` directory
* setup messages adjusted to each command

https://storjlabs.atlassian.net/browse/V3-1040


cmd/uplink/cmd/setup.go
```diff
@@ -20,13 +20,29 @@ import (
...
-	setupCmd = &cobra.Command{
+	uplinkSetupCmd = &cobra.Command{
 		Use:         "setup",
 		Short:       "Create an uplink config file",
-		RunE:        cmdSetup,
+		RunE:        cmdUplinkSetup,
 		Annotations: map[string]string{"type": "setup"},
 	}
-	setupCfg struct {
+	gatewaySetupCmd = &cobra.Command{
```
kaloyan-raev: How about moving all things related to gateway under `/cmd/gateway/cmd`?

This we would also avoid complicating var/type names to distinguish if they relate to the uplink or the gateway.

mniewrzal: Yea, good idea

egonelbre:
I was thinking about this and I think it would be better to completely separate `uplink` and `gateway`, such that `gateway` wouldn't depend on `uplink` commands (e.g. mount) and `uplink` wouldn't depend on minio.

mniewrzal:
Yes, I'm in the middle of this process.

cmd/gateway/Dockerfile
```diff
@@ -5,12 +5,12 @@ FROM storjlabs/golang as build-env
 
 # final stage
 FROM alpine
-ENV CONF_PATH=/root/.local/share/storj/uplink \
+ENV CONF_PATH=/root/.local/share/storj/gateway \
```
egonelbre: I suspect that these may have been intentionally using the same location to ensure that they use the same identity.

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+
+var (
+	// GWCmd represents the base gateway command when called without any subcommands
+	GWCmd = &cobra.Command{
+		Use:   "gateway",
+		Short: "The Storj client-side S3 gateway",
+	}
+	setupCmd = &cobra.Command{
+		Use:         "setup",
+		Short:       "Create an gateway config file",
```
egonelbre: `Create a gateway config file`

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+		Annotations: map[string]string{"type": "setup"},
+	}
+	runCmd = &cobra.Command{
+		Use:   "run",
+		Short: "Run the S3 gateway",
+		RunE:  cmdRun,
+	}
+
+	defaultConfDir     = fpath.ApplicationDir("storj", "gateway")
+	defaultIdentityDir = fpath.ApplicationDir("storj", "identity", "gateway")
```
egonelbre: Same here, I think it might have used uplink folders so that it's easier setup... but it can be left separate, either is fine.

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+
+	setupCfg GatewayFlags
+	runCfg   GatewayFlags
+
+	gwConfDir   string
```
egonelbre: use `confDir` to be consistent with other implementations

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+	cfgstruct.BindSetup(setupCmd.Flags(), &setupCfg, cfgstruct.ConfDir(defaultConfDir), cfgstruct.IdentityDir(defaultIdentityDir))
+}
+
+func cmdSetup(cmd *cobra.Command, args []string) (err error) {
+	setupDir, err := filepath.Abs(gwConfDir)
+	if err != nil {
+		return err
+	}
+
+	for _, flagname := range args {
```
egonelbre: You can add requirements to commands so you don't have to manually check the args list.

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+	"storj.io/storj/pkg/miniogw"
 	"storj.io/storj/pkg/process"
+	"storj.io/storj/pkg/storage/streams"
+	"storj.io/storj/pkg/storj"
+)
+
+// GatewayFlags configuration flags
+type GatewayFlags struct {
+	APIKey             string `default:"" help:"the api key to use for the satellite" setup:"true"`
+	GenerateMinioCerts bool   `default:"false" help:"generate sample TLS certs for Minio GW" setup:"true"`
```
egonelbre: Maybe make the flag `--generate-test-certs`.

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+	_, err = metainfo.ListBuckets(ctx, storj.BucketListOptions{Direction: storj.After})
+	if err != nil {
+		return fmt.Errorf("Failed to contact Satellite.\n"+
+			"Perhaps your configuration is invalid?\n%s", err)
+	}
+
+	return runCfg.Run(ctx)
+}
+
+func generateAWSKey() (key string, err error) {
```
egonelbre: Why is this called `AWSKey`?

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+//
+// Temporarily it also returns an instance of streams.Store until we improve
+// the metainfo and streas implementations.
+func (c *GatewayFlags) Metainfo(ctx context.Context) (storj.Metainfo, streams.Store, error) {
+	identity, err := c.Identity.Load()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return c.GetMetainfo(ctx, identity)
```
egonelbre: Here I got confused what `c` was. Use `flags`.

cmd/uplink/cmd/setup.go
```diff
@@ -4,48 +4,41 @@
...
 	"storj.io/storj/pkg/process"
-	"storj.io/storj/pkg/provider"
 )
 
+// UplinkFlags configuration flags
+type UplinkFlags struct {
+	APIKey        string `default:"" help:"the api key to use for the satellite" setup:"true"`
+	SatelliteAddr string `default:"localhost:7778" help:"the address to use for the satellite" setup:"true"`
+
+	miniogw.Config // TODO remove when GetMetainfo will be moved from minio
```
egonelbre: Let's duplicate the code and remove `minio` dependency from uplink. With some comment like "keep in sync with etc" on both sides. But we can also do that in a separate PR.

cmd/uplink/cmd/setup.go
```diff
@@ -4,48 +4,41 @@
...
-		RS     miniogw.RSConfig
-		Enc    miniogw.EncryptionConfig
-	}
+
+	setupCfg UplinkFlags
 
 	defaultConfDir     = fpath.ApplicationDir("storj", "uplink")
 	defaultIdentityDir = fpath.ApplicationDir("storj", "identity", "uplink")
 
 	cliConfDir  string
```
egonelbre: be consistent with other binary implementations

cmd/gateway/main.go
```diff
@@ -4,10 +4,212 @@
...
+	APIKey             string `default:"" help:"the api key to use for the satellite" setup:"true"`
+	GenerateMinioCerts bool   `default:"false" help:"generate sample TLS certs for Minio GW" setup:"true"`
+	SatelliteAddr      string `default:"localhost:7778" help:"the address to use for the satellite" setup:"true"`
+
+	miniogw.Config
+}
+
+var (
+	// GWCmd represents the base gateway command when called without any subcommands
+	GWCmd = &cobra.Command{
```
egonelbre: consistency with other implementations

cmd/gateway/main.go
mniewrzal: Copy paste, fixed :)

cmd/gateway/main.go
mniewrzal: done

cmd/gateway/main.go
mniewrzal: done

cmd/gateway/main.go
mniewrzal: done

cmd/gateway/main.go
mniewrzal: sounds ok, done

cmd/gateway/main.go
mniewrzal: done

cmd/uplink/cmd/setup.go
mniewrzal: lets do this in separate PR

cmd/gateway/main.go
mniewrzal: done

cmd/uplink/cmd/setup.go
mniewrzal: done

PR #1027, opened 2019-01-10
do not try to dial nil node from audit verifier

mobyvb:
https://storjlabs.atlassian.net/browse/V3-1015


pkg/audit/verifier.go
```diff
@@ -60,6 +60,10 @@ func NewVerifier(transport transport.Client, overlay overlay.Client, id provider
 func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize, pieceNumber int,
 	id psclient.PieceID, pieceSize int64, fromNode *pb.Node, pba *pb.PayerBandwidthAllocation, authorization *pb.SignedMessage) (s share, err error) {
 	defer mon.Task()(&ctx)(&err)
+
+	if fromNode == nil {
```
wthorp: So overlay.BulkLookup() returns nulls because cache.GetAll() returns nulls.  Part of me wonders if that's a valuable feature in the first place.

mobyvb: I think it is useful so that there is a simple mapping from node IDs passed into those functions to the returned nodes. Unless we return a map rather than a list, I think having nils in the returned list will be necessary.

PR #1028, opened 2019-01-11
adds AddObserver function for use in tests and easier listener management

dylanlott:
Adds the AddObserver() method to transport client for adding observers after instantiation

### DIFF
--- a/pkg/transport/transport.go
+++ b/pkg/transport/transport.go
@@ -5,6 +5,7 @@ package transport
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/zeebo/errs"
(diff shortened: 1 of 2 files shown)


pkg/transport/transport.go
```diff
@@ -36,12 +37,15 @@ type Client interface {
...
 	Identity() *provider.FullIdentity
+	AddObserver(obs Observer)
+	Observers() []Observer
 }
 
 // Transport interface structure
 type Transport struct {
 	identity  *provider.FullIdentity
 	observers []Observer
+	sync.Mutex
```
egonelbre: Don't expose the mutex and order such that it can be understood what it is protecting.

PR #1030, opened 2019-01-11
Add arguments to lockedgen for using it in other places

egonelbre:


scripts/lockedgen/main.go
```diff
@@ -20,20 +23,35 @@ import (
 
 func main() {
```
mniewrzal: Can you add `// +build ignore` for this script?

egonelbre: Not sure why?

mniewrzal: just to avoid having one more executable `lockedgen` on path. Like other scripts in this folder.

egonelbre: anyways did the move and added tag

PR #1034, opened 2019-01-11
Satellite Peer

egonelbre:
Contains many fixes to tests and adjustments to make `satellite.Peer` work, unfortunately this resulted in a large PR. It's probably easier to review this as is than to split and then do them separately.

This also temporarily disables pointerdb auth and audit tests. These will be fixed / reenabled in upcoming PR-s.


satellite/peer.go
```diff
@@ -0,0 +1,361 @@
+// Copyright (C) 2018 Storj Labs, Inc.
```
mniewrzal: 2019

PR #1036, opened 2019-01-11
Adding additional required values for satellite configuration

phutchins:


cmd/satellite/entrypoint
```diff
@@ -1,12 +1,32 @@
...
+	export STORJ_SERVER_ADDRESS="${IDENTITY_ADDR}"
+fi
+
+if [[ -n "${BOOTSTRAP_ADDR}" ]]; then
+	export STORJ_KADEMLIA_BOOTSTRAP_ADDR="${BOOTSTRAP_ADDR}"
 fi
 
 if [[ ! -f "${CONF_PATH}/config.yaml" ]]; then
-	if [[ -n "${CA_DIFFICULTY}" ]]; then
+	if [[ -n "${CA_DIFFICULTY:-}" ]]; then
```
egonelbre: Other entrypoints need this fix as well.

phutchins: Done.

cmd/satellite/entrypoint
```diff
@@ -1,12 +1,32 @@
...
+SETUP_PARAMS=""
+
 if [[ -n "${API_KEY}" ]]; then
 	export STORJ_POINTER_DB_AUTH_API_KEY="${API_KEY}"
+  export STORJ_AUDIT_API_KEY="${API_KEY}"
+  export STORJ_POINTER_DB_AUTH_API_KEY="${API_KEY}"
+  export STORJ_REPAIRER_API_KEY="${API_KEY}"
+fi
+
+if [[ -n "${OVERLAY_URL}" ]]; then
```
egonelbre: Don't all of the others need `:-}` as well?

https://travis-ci.org/storj/storj/jobs/478372277#L1789

phutchins: Intention was to only do those for the ones that had sane defaults. I went ahead and added to the rest of the conditionals and we can let the binary handle failure or warning.

PR #1040, opened 2019-01-11
Remove overwrite flags 

mobyvb:
* remove overwrite flags from uplink, satellite, storagenode, captplanet, bootstrap

https://storjlabs.atlassian.net/browse/V3-1013


cmd/captplanet/setup.go
```diff
@@ -42,6 +43,13 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	if !setupCfg.Overwrite && !valid {
 		return fmt.Errorf("captplanet configuration already exists (%v). Rerun with --overwrite", setupDir)
 	} else if setupCfg.Overwrite && err == nil {
+		opts := []string{"yes", "no"}
+		question := fmt.Sprintf("Are you sure you want to delete the following directory?\n%v", setupDir)
+		i := prompt.Choose(question, opts)
```
stefanbenten: Cant we choose the Confirm Option of the module here and get rid of the above declaration of Yes/No?
This is also breaking our tests as Travis isn't responding interactive.

stefanbenten:
@egonelbre Do you by chance know a way how we can get Travis to interactively respond to this prompt?

cmd/captplanet/setup.go
stefanbenten: should be as simple as this:
```
if ok := prompt.Confirm("Are you sure you want to delete the following directory?\n%v", setupDir); !ok {
    return
}
```

cmd/captplanet/setup.go
```diff
@@ -42,6 +43,13 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
+		if opts[i] == "no" {
+			return
```
stefanbenten: We could also add a proper error in the case of rejecting the prompt

PR #1041, opened 2019-01-11
Reorder graphite metric string for easier parsing

pgerbes1:


egonelbre:
redid the PR in https://github.com/storj/storj/pull/1362 so we can ignroe clabot issues

PR #1045, opened 2019-01-13
WIP: kad street surgery

wthorp:
Me trying to fix the issues exposed by https://github.com/storj/storj/pull/1003


wthorp:
testplanet now only bootstraps off of satellite, not all nodes
testplanet now only create overlays for satellites, not all nodes
testplanet now only bootstraps kademlia for storage nodes, not all nodes
Discovery creates new uptime records if they don't exist during Refresh()
Nodes update routing table on Ping()
testplanet now bootstraps kademlia for storage nodes after gRPC starts

egonelbre:
Whole testplanet setup is going to change, see the changes already made to storagenode.Peer and the pending PR for satellite.Peer... i.e. #1034

storagenode/peer.go
```diff
@@ -149,6 +149,13 @@ func (peer *Peer) Run(ctx context.Context) error {
 	defer cancel()
 
 	var group errgroup.Group
+	group.Go(func() error {
+		err := peer.Public.Server.Run(ctx)
```
wthorp: @egonelbre This is the only thing I think I found that might be of value... Kademlia has a 'Ping back' feature, so Kademlia's gRPC should always be started up before Kademlia bootstraps.

egonelbre: Well, since they are started in parallel the slight ordering change between the goroutines shouldn't change much.

PR #1046, opened 2019-01-14
add more bwagreement unit tests

littleskunk:
Storage node manipulates the bwagreement size and tries a few strategies to self sign it.

Storage node would like to take the satellite offline by submitting a corrupted PubKey.

Storage node sends an expired bwagreement.


pkg/bwagreement/test/server_test.go
```diff
@@ -21,18 +22,20 @@ import (
...
-		   Uplink would like to download a file from 2 storage nodes.
-		   Uplink requests a PayerBandwidthAllocation from the satellite. One serial number for all storage nodes.
-		   Uplink signes 2 RenterBandwidthAllocation for both storage node. */
-		satellitePubKey, satellitePrivKey, uplinkPrivKey := generateKeys(ctx, t)
-		server := bwagreement.NewServer(db.BandwidthAgreement(), zap.NewNop(), satellitePubKey)
+		testDatabase(ctx, t, db.BandwidthAgreement())
+	})
+}
+
+func testDatabase(ctx context.Context, t *testing.T, bwdb bwagreement.DB) {
```
thepaul: I like all these tests. These are great. But why not put them in separate functions and have testDatabase() call them all? The way this is, it takes a lot of careful review to be pretty sure that each step won't affect local variables or the server state in an important way.

egonelbre: Because of tparse, it's being quite noisy in Travis and don't want to add too much to it. However, yeah, once tparse shows information better, then it would be nice to have them separately.

thepaul: ok then

littleskunk:
One little last minute change. The unit test for expired bwagreements was using an bwagreement that expired 5 days ago. We can test that more accurate. I have changed it to 0 seconds (expires right now) and added 2 more test cases to also test 1s (still valid) and -23h55s (expired a day ago). That should cover all cases.

pkg/bwagreement/test/common.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 //GeneratePayerBandwidthAllocation creates a signed PayerBandwidthAllocation from a PayerBandwidthAllocation_Action
-func GeneratePayerBandwidthAllocation(action pb.PayerBandwidthAllocation_Action, satelliteKey crypto.PrivateKey, uplinkKey crypto.PrivateKey, expired bool) (*pb.PayerBandwidthAllocation, error) {
+func GeneratePayerBandwidthAllocation(action pb.PayerBandwidthAllocation_Action, satelliteKey crypto.PrivateKey, uplinkKey crypto.PrivateKey, expiration string) (*pb.PayerBandwidthAllocation, error) {
```
egonelbre: use here `time.Duration` as the argument

pkg/accounting/tally/tally_test.go
```diff
@@ -77,7 +77,7 @@ func TestQueryWithBw(t *testing.T) {
 
 func makeBWA(ctx context.Context, t *testing.T, bwDb bwagreement.DB, serialNum string, k *ecdsa.PrivateKey, action pb.PayerBandwidthAllocation_Action) {
 	//generate an agreement with the key
-	pba, err := test.GeneratePayerBandwidthAllocation(action, k, k, false)
+	pba, err := test.GeneratePayerBandwidthAllocation(action, k, k, "1h")
```
egonelbre: `time.Hour` instead of `"1h"`

PR #1047, opened 2019-01-14
Change overlay cache database interface

egonelbre:


pkg/audit/cursor_test.go
```diff
@@ -38,10 +38,17 @@ func TestAuditSegment(t *testing.T) {
...
+	defer tctx.Cleanup()
+
+	planet, err := testplanet.New(t, 1, 4, 1)
+	require.NoError(t, err)
+	defer tctx.Check(planet.Shutdown)
+
+	planet.Start(tctx)
+
+	// we wait a second for all the nodes to complete bootstrapping off the satellite
+	time.Sleep(2 * time.Second)
```
thepaul: Timed waits in tests always end up being fragile. Is there not any way to tell when all the nodes have completed bootstrapping?

satellite/satellitedb/overlaycache.go
```diff
@@ -6,122 +6,234 @@ package satellitedb
...
-			updateFields,
-		)
+		infos[i] = info
+	}
+	return infos, nil
+}
+
+// List lists nodes starting from cursor
+func (cache *overlaycache) List(ctx context.Context, cursor storj.NodeID, limit int) ([]*pb.Node, error) {
+	dbxInfos, err := cache.db.Limited_OverlayCacheNode_By_NodeId_GreaterOrEqual(ctx,
```
mniewrzal: I think we should cover invalid `limit` with min and max.

pkg/audit/cursor_test.go
egonelbre: Unfortunately, not yet. Yeah, I don't like it either, but there will be PR-s eventually to fix this.

thepaul: 👍

satellite/satellitedb/overlaycache.go
egonelbre: done

egonelbre: also created a separate task for reviewing the limit situation in the future

PR #1050, opened 2019-01-14
kad storagenode inspection and dashboard

dylanlott:
Creates a simple dashboard for SNOs to see their node's activity and verify that things are working correctly.


pkg/piecestore/psserver/server.go
```diff
@@ -214,6 +220,33 @@ func (s *Server) Stats(ctx context.Context, in *pb.StatsReq) (*pb.StatSummary, e
 	return &pb.StatSummary{UsedSpace: totalUsed, AvailableSpace: (s.totalAllocated - totalUsed), UsedBandwidth: totalUsedBandwidth, AvailableBandwidth: (s.totalBwAllocated - totalUsedBandwidth)}, nil
 }
 
+// Dashboard is a stream that sends data every `interval` seconds to the listener.
+func (s *Server) Dashboard(in *pb.DashboardReq, stream pb.PieceStoreRoutes_DashboardServer) (err error) {
+	ctx := stream.Context()
+	ticker := time.NewTicker(500 * time.Millisecond)
```
aleitner: let's make this 5-10 seconds?

dylanlott:
🅱️🅱️🅱️
  

  
  

  
  
>

pkg/piecestore/psserver/server.go
```diff
@@ -301,3 +334,31 @@ func getNamespacedPieceID(pieceID, namespace []byte) (string, error) {
...
+	if err != nil {
+		return &pb.DashboardStats{}, ServerError.Wrap(err)
+	}
+
+	// nodeId, err := rt.Local().Id)
+
+	return &pb.DashboardStats{
+		NodeId:          rt.Local().Id.String(),
+		NodeConnections: int64(len(nodes)),
+		Address:         "",
```
aleitner: Is this something we can get from kademlia?

pkg/piecestore/psserver/server.go
```diff
@@ -301,3 +334,31 @@ func getNamespacedPieceID(pieceID, namespace []byte) (string, error) {
...
+		Connection:      true,
```
aleitner: should probably figure out how to verify we are connected to something

dylanlott:
We could get the address from Kademlia I think but I’m not sure if it would be the correct tunneled address.   
  

  
For connection, I agree we need a better way to confirm they actually are connected. Maybe pinging the satellite would be good.   
  

  
  

  
  
>

cmd/storagenode/main.go
```diff
@@ -292,6 +304,58 @@ func cmdDiag(cmd *cobra.Command, args []string) (err error) {
...
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return err
+		}
+
+		clr()
+		color.Green("\n\033[1mStorage Node Dashboard Stats\033[1m\n")
```
egonelbre: Oh, this will definitely break on Windows. Use the appropriate stuff from `color` as necessary rather than directly using escape codes.

internal/testplanet/identities_table.go
```diff
@@ -6,154 +6,154 @@
 package testplanet
 
 var pregeneratedIdentities = NewIdentities(
```
egonelbre: no need to regenerate this

pkg/satellite/satellitedb/dbx/satellitedb.dbx.go
```diff
@@ -18,7 +18,6 @@ import (
 	"unicode"
 
 	"github.com/mattn/go-sqlite3"
-	"math/rand"
 )
```
egonelbre: These shouldn't need changes. Also update your dbx installation.

cmd/storagenode/main.go
```diff
@@ -315,6 +385,11 @@ func isOperatorWalletValid(wallet string) error {
 	return nil
 }
 
+// clr uses ANSI escape codes to clear the screen
+func clr() {
+	print("\033[H\033[2J")
```
egonelbre: for windows you need to invoke cls https://stackoverflow.com/questions/22891644/how-can-i-clear-the-terminal-screen-in-go

pkg/satellite/satellitedb/dbx/satellitedb.dbx.go
dylanlott: resolved

internal/testplanet/identities_table.go
dylanlott: resolved

dylanlott:
closing, solved by to #1072

PR #1051, opened 2019-01-14
Save only user-specific flags to storage node's config.yaml

kaloyan-raev:


pkg/process/exec_conf.go
```diff
@@ -128,10 +135,6 @@ func cleanup(cmd *cobra.Command) {
 		defer mon.TaskNamed("root")(&ctx)(&err)
 
 		vip := viper.New()
-		err = vip.BindPFlags(cmd.Flags())
```
thepaul: Why remove the binding of flags to viper?

pkg/process/exec_conf.go
```diff
@@ -54,7 +54,7 @@ var (
 
 // SaveConfig will save all flags with default values to outfilewith specific
 // values specified in 'overrides' overridden.
-func SaveConfig(flagset *pflag.FlagSet, outfile string, overrides map[string]interface{}) error {
+func SaveConfig(flagset *pflag.FlagSet, outfile string, overrides map[string]interface{}, saveAllDefaults bool) error {
```
egonelbre: Can we use a separate name for this, instead of boolean flag. e.g. `SaveConfigUser`, `SaveConfig`?

pkg/process/exec_conf.go
kaloyan-raev: This call makes the `Changed` field of every flag to be `true`. Which is problematic for the SaveConfig.

thepaul: But, then, doesn't this break the viper functionality? Are the flags still set somehow to the values in the config file that viper reads?

pkg/process/exec_conf.go
kaloyan-raev: Added `SaveConfigWithAllDefaults` function.

pkg/process/exec_conf.go
kaloyan-raev: You got a point. We reverted this line and found another solution for restoring the `Changed` field.

PR #1052, opened 2019-01-14
use Node ID for metrics ID

thepaul:
This change removes automatic metrics reporting for everything going
through process.Exec(), and re-adds metrics reporting for those commands
which are expected to be long-lived. Other commands (which may have been
intermittently sending metrics before this, if they ran unusually long)
will no longer send any metrics.

For commands where it makes sense, a node ID is used as the metrics ID.


pkg/process/metrics.go
```diff
@@ -28,11 +30,19 @@ var (
 		"application suffix")
 )
 
-func initMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (
+// InitMetrics: Initialize telemetry reporting. Makes a telemetry.Client and calls
```
coyle: linter is going to complain

pkg/process/metrics.go
```diff
@@ -48,3 +58,17 @@ func initMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (
 	go c.Run(ctx)
 	return nil
 }
+
+// InitMetrics: Initialize telemetry reporting, using the node ID corresponding
```
coyle: linter is going to complain

pkg/process/metrics.go
```diff
@@ -48,3 +58,17 @@ func initMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (
...
+// to the given certificate as the telemetry instance ID.
+func InitMetricsWithCertPath(ctx context.Context, r *monkit.Registry, certPath string) error {
+	var metricsID string
+	nodeID, err := identity.NodeIDFromCertPath(certPath)
+	if err != nil {
+		zap.S().Errorf("Could not read identity for telemetry setup: %v", err)
+		metricsID = ""
```
coyle: should we give it another option? seems odd to just leave it empty.

pkg/process/metrics.go
thepaul: siggggggghhhhhhhhhhhhh

pkg/process/metrics.go
thepaul: InitMetrics() will use the hw MAC address by default if you pass in an empty string, which is the way it worked before. Is it confusing to defer the work to InitMetrics() like this?

thepaul: to clarify: using the MAC address is the way it worked before, not the part about how passing in an empty string makes it do that.

coyle: Slightly, maybe a comment to explain what's happening downstream, just to prevent someone from pulling their hair out trying to figure out where the MAC address is getting set

pkg/process/metrics.go
```diff
@@ -28,11 +30,19 @@ var (
...
+func InitMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (
 	err error) {
 	if *metricCollector == "" || *metricInterval == 0 {
 		return Error.New("telemetry disabled")
 	}
+	if r == nil {
+		r = monkit.Default
+	}
+	if instanceID == "" {
+		instanceID = telemetry.DefaultInstanceID()
```
egonelbre: I'm not liking this from testing standpoint, but I guess we can manage for now.

PR #1054, opened 2019-01-15
Renaming and moving pkg/satellite to satellite/console

crawter:


satellite/console/consoleweb/api.go
```diff
@@ -10,12 +10,13 @@ import (
 	"net/http"
 	"strings"
 
+	"storj.io/storj/satellite/console"
```
egonelbre: group together with other storj.io packages

PR #1055, opened 2019-01-15
Consolidate identity management to identity cli commands

bryanchriswhite:
_(description forthcoming)_

### DIFF
--- a/cmd/identity/main.go
+++ b/cmd/identity/main.go
@@ -130,11 +130,21 @@ func cmdCSR(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	err = caConfig.SaveBackup(ca)
+	if err != nil {
+		return err
+	}
+
 	ca.Cert = signedChain[0]
 	ca.RestChain = signedChain[1:]
 	err = identity.FullCAConfig{
 		CertPath: caConfig.CertPath,
 	}.Save(ca)
+	if err != nil {
+		return err
+	}
+
+	err = identConfig.SaveBackup(ident)
 	if err != nil {
 		return err
 	}
--- a/cmd/identity/main.go
+++ b/cmd/identity/main.go
@@ -36,7 +36,6 @@ var (
 	csrCmd = &cobra.Command{
 		Use:   "csr <service>",
 		Short: "Send a certificate signing request for a service's CA certificate",
-		Long:  "TESTING123",
 		Args:  cobra.ExactArgs(1),
 		RunE:  cmdCSR,
 	}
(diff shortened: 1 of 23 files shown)


cmd/identity/main.go
```diff
@@ -7,6 +7,11 @@ import (
 	"crypto/x509/pkix"
 	"encoding/json"
 	"fmt"
+	"github.com/zeebo/errs"
+	"path/filepath"
+	"storj.io/storj/pkg/certificates"
```
coyle: nit sep

cmd/identity/main.go
```diff
@@ -21,13 +26,132 @@ var (
...
+		Use:   "new <service>[, ...]",
+		Short: "Create a new full identity for one or more services",
+		Args:  cobra.MinimumNArgs(1),
+		RunE:  cmdNewService,
+	}
+
+	csrCmd = &cobra.Command{
+		Use:   "csr <service>",
+		Short: "Send a certificate signing request for a service's CA certificate",
+		Long:  "TESTING123",
```
coyle: still need?

cmd/identity/main.go
```diff
@@ -21,13 +26,132 @@ var (
...
+		Args:  cobra.ExactArgs(1),
+		RunE:  cmdCSR,
+	}
+
+	config struct {
+		BaseDir        string `default:"$CONFDIR/identity" help:"Directory containing service subdirectories"`
+		Difficulty     uint64 `default:"15" help:"minimum difficulty for identity generation"`
+		Timeout        string `default:"5m" help:"timeout for CA generation; golang duration string (0 no timeout)"`
```
coyle: is this what we should remove ?

cmd/identity/main.go
```diff
@@ -21,13 +26,133 @@ var (
...
+	signedChain, err := identity.ParseCertChain(signedChainBytes)
+	if err != nil {
+		return nil
+	}
+
+	ca.Cert = signedChain[0]
+	ca.RestChain = signedChain[1:]
+	err = identity.FullCAConfig{
+		CertPath: caConfig.CertPath,
+	}.Save(ca)
```
bryanchriswhite: Save backups!

cmd/identity/main.go
```diff
@@ -21,13 +26,132 @@ var (
...
+		return errs.New("error occurred while signing certificate: %s\n(identity files were still generated and saved, if you try again existing files will be loaded)", err)
+	}
+
+	signedChain, err := identity.ParseCertChain(signedChainBytes)
+	if err != nil {
+		return nil
+	}
+
+	ca.Cert = signedChain[0]
+	ca.RestChain = signedChain[1:]
```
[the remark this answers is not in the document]
bryanchriswhite: it shouldn't have only 1 item because it should contain the cert that was sent to be signed and the cert that signed it; however, even if it did i don't think this would be an error.

PR #1056, opened 2019-01-15
rollup query

wthorp:
https://storjlabs.atlassian.net/browse/V3-720
The scope of this has changed somewhat from the JIRA ticket...
Select each NodeID in the raw table;  For each bucket daily, Total the tally for the node and save into the rollup table


pkg/accounting/common.go
```diff
@@ -18,4 +18,6 @@ const (
 	LastAtRestTally = "LastAtRestTally"
 	// LastBandwidthTally represents the accounting timestamp for the bandwidth allocation query
 	LastBandwidthTally = "LastBandwidthTally"
+	// LastAtReLastRollupstTally represents the accounting timestamp for rollup calculations
```
jenlij: is this name right "LastAtReLasRollupstTally"? Looks like something got dragged and dropped accidentally

pkg/accounting/rollup/rollup.go
```diff
@@ -5,11 +5,14 @@ package rollup
...
 	"time"
 
 	"go.uber.org/zap"
 
 	"storj.io/storj/pkg/accounting"
+	"storj.io/storj/pkg/storj"
+	dbx "storj.io/storj/satellite/satellitedb/dbx"
 )
 
 // Rollup is the service for totalling data on storage nodes for 1, 7, 30 day intervals
```
jenlij: update this comments :)

PR #1061, opened 2019-01-15
adds base 10 memory sizes

navillasa:


aleitner:
we need to update the tests

PR #1062, opened 2019-01-15
double-quote path variables in test-captplanet.sh

thepaul:


thepaul:
We're dropping captplanet so this is moot

PR #1070, opened 2019-01-16
Remove captplanet

egonelbre:


thepaul:
cmd/inspector/main.go also has a mention of captplanet, if you want to include that too.

And scripts/cert-gating.sh looks like it depends on captplanet, if that's a thing we use

egonelbre:
For `cert-gating.sh` I'm not quite sure how to fix that part, I'll let Bryan handle that.

PR #1072, opened 2019-01-16
storagenode dashboard is working and passing tests locally

dylanlott:


cmd/storagenode/main.go
```diff
@@ -119,12 +129,8 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 	} else {
 		zap.S().Info("Operator wallet: ", operatorConfig.Wallet)
 	}
-	ctx := process.Ctx(cmd)
-	if err := process.InitMetricsWithCertPath(ctx, nil, runCfg.Identity.CertPath); err != nil {
-		zap.S().Error("Failed to initialize telemetry batcher:", err)
-	}
```
thepaul: Wait why are you taking out metrics for storagenode?

cmd/storagenode/main.go
```diff
@@ -319,6 +390,42 @@ func isOperatorWalletValid(wallet string) error {
 	return nil
 }
 
+// clr clears the screen so it can be redrawn
+func clr() {
+	var clear = make(map[string]func())
+	clear["linux"] = func() {
```
thepaul: for the linux and darwin cases, you only need to printf "\033[H\033[J". that's the ANSI escape sequences for "move the cursor to the home position" and "clear the screen after the cursor".

thepaul: maybe Windows too, as I remember hearing that cmd.exe supports ANSI escape codes now.

cmd/storagenode/main.go
```diff
@@ -82,7 +92,6 @@ var (
 )
 
 const (
-	// default server address, only for storage node
```
mniewrzal: why you are removing this?

dylanlott:
Sure, but this works fine and it could be argued this is more understandable.
  

  
  

  
  
>

dylanlott:
I’m pretty sure this is a result of something being merged incorrectly, this shouldn’t be getting removed. I’ll fix this.   
  

  
  

  
  
>

cmd/storagenode/main.go
```diff
@@ -161,14 +167,15 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	overrides := map[string]interface{}{
-		"log.level": "info",
-	}
-	serverAddress := cmd.Flag("server.address")
-	if !serverAddress.Changed {
-		overrides[serverAddress.Name] = defaultServerAddr
+		"identity.cert-path":      setupCfg.Identity.CertPath,
```
mniewrzal: Why you reverted this block of code? Current version looks like this:
```
	overrides := map[string]interface{}{
		"log.level": "info",
	}
	serverAddress := cmd.Flag("server.address")
	if !serverAddress.Changed {
		overrides[serverAddress.Name] = defaultServerAddr
	}
```

dylanlott:
We only used this solution specifically because Windows doesn’t support it (as far as I was told and read) Egon can maybe lend more information.   
  

  
  

  
  
>

pkg/piecestore/psclient/liteclient.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (C) 2018 Storj Labs, Inc.
```
mniewrzal: 2019

thepaul:
It's usually worth it for long-running daemon programs to avoid forking. Plus I'd argue it's a lot easier to read one printf line plus a comment than it is to go over this whole thing with the multiple cases and the subprogram setup and the cleanup and the success check and so on.

pkg/piecestore/psclient/liteclient.go
```diff
@@ -0,0 +1,59 @@
...
+func NewLiteClient(ctx context.Context) (LiteClient, error) {
+	clientIdent, err := provider.NewFullIdentity(ctx, 12, 4)
+	if err != nil {
+		return nil, err
+	}
+
+	tc := transport.NewClient(clientIdent)
+	n := &pb.Node{
+		Address: &pb.NodeAddress{
+			Address:   ":7777",
```
mniewrzal: you hardcoded ip and port of storage node?

pkg/piecestore/psserver/config.go
```diff
@@ -28,7 +28,7 @@ var (
 
 // Config contains everything necessary for a server
 type Config struct {
-	Path                         string        `user:"true" help:"path to store data in" default:"$CONFDIR/storage"`
```
mniewrzal: why you reverted this change?

pkg/piecestore/psserver/server.go
```diff
@@ -120,18 +124,21 @@ func NewEndpoint(log *zap.Logger, config Config, storage *pstore.Storage, db *ps
...
 		totalAllocated:   allocatedDiskSpace,
 		totalBwAllocated: allocatedBandwidth,
 		verifier:         auth.NewSignedMessageVerifier(),
+		kad:              k,
 	}, nil
 }
 
 // New creates a Server with custom db
 func New(log *zap.Logger, storage *pstore.Storage, db *psdb.DB, config Config, pkey crypto.PrivateKey) *Server {
+	fmt.Printf("\n CONFIG %+v\n", config)
```
mniewrzal: remove Printf

thepaul:
(Having said that, this isn't that big of an issue for the dashboard and if it's working, then fine.)

cmd/storagenode/main.go
dylanlott: Added back in

cmd/storagenode/main.go
dylanlott: It wasn't being used anywhere.

cmd/storagenode/main.go
dylanlott: This wasn't meant to be changed, this PR has some weird merges in it. I'll fix this.

pkg/piecestore/psserver/config.go
dylanlott: result of weird merge - fixed

pkg/piecestore/psserver/server.go
dylanlott: removed

pkg/piecestore/psclient/liteclient.go
dylanlott: Updated to take an address argument, should be sufficient for now.

pkg/piecestore/psclient/liteclient.go
dylanlott: fixed

cmd/storagenode/main.go
thepaul: Interestingly, I took a look at the color library you're using, and it is actually _interpreting_ escape codes and translating them to native Windows calls as appropriate! If you write that sequence to `color.Output`, it should work on Windows.

But ok sorry I'll leave you alone :D

dylanlott: Oh, I actually just assumed that library wasn't going to work on Windows, but that's good to know. I'll get this PR merged, but I can submit a new PR later on to use the actual escape sequences later on.

thepaul: 👍

cmd/storagenode/main.go
```diff
@@ -301,6 +315,71 @@ func cmdDiag(cmd *cobra.Command, args []string) (err error) {
 	return err
 }
 
+func dashCmd(cmd *cobra.Command, args []string) (err error) {
+	ctx := context.Background()
+
+	// create new client on localhost:7777
+	lc, err := psclient.NewLiteClient(ctx, ":7777")
```
mniewrzal: currently default is 28967 :)

cmd/storagenode/main.go
```diff
@@ -319,6 +394,42 @@ func isOperatorWalletValid(wallet string) error {
 	return nil
 }
 
+// clr clears the screen so it can be redrawn
+func clr() {
+	var clear = make(map[string]func())
+	clear["linux"] = func() {
```
egonelbre: A regular switch here would have been sufficent

PR #1077, opened 2019-01-16
adds cli flags for dashboard server address config

dylanlott:


cmd/storagenode/main.go
```diff
@@ -89,6 +90,7 @@ var (
 	defaultConfDir string
 	defaultDiagDir string
 	confDir        string
+	serverAddress  string
```
mniewrzal: It should be cleaner with approach from different commands. like this:
https://github.com/storj/storj/blob/68ce08943f2ea0e7523ace7f4f76ea8131566368/cmd/satellite/main.go#L87

mniewrzal: wrong link https://github.com/storj/storj/blob/master/cmd/satellite/main.go#L88

PR #1078, opened 2019-01-16
Rename storj-sdk to storj-sim

egonelbre:


Makefile
```diff
@@ -65,12 +65,12 @@ proto: ## Rebuild protobuf files
...
-##@ SDK
+##@ Simulator
 
-.PHONY: install-sdk
-install-sdk: ## install storj-sdk
+.PHONY: install-sim
+install-sim: ## install storj-sim
 	@echo "Running ${@}"
-	@go install -race -v storj.io/storj/cmd/storj-sdk storj.io/storj/cmd/bootstrap storj.io/storj/cmd/satellite storj.io/storj/cmd/storagenode storj.io/storj/cmd/uplink storj.io/storj/cmd/gateway
+	@go install -race -v storj.io/storj/cmd/storj-sim storj.io/storj/cmd/bootstrap storj.io/storj/cmd/satellite storj.io/storj/cmd/storagenode storj.io/storj/cmd/uplink storj.io/storj/cmd/gateway
```
thepaul: (oops, commented in the wrong place before.) Did you mean to leave off `certificates` here, compared to what was removed in .travis.yml?

egonelbre: Yes, `storj-sim` is not using `certificates`, however it shouldn't have been removed from travis.

cmd/storj-sim/main.go
```diff
@@ -29,8 +29,8 @@ func main() {
 	var flags Flags
 
 	rootCmd := &cobra.Command{
-		Use:   "storj-sdk",
-		Short: "Storj SDK",
+		Use:   "storj-sim",
+		Short: "Storj Simulator",
```
thepaul: Maybe "Storj Network Simulator" would be a better full name?

PR #1080, opened 2019-01-16
Fixed unit tests for web satellite

Barterio:


web/satellite/tests/unit/store/projectMembers.spec.ts
```diff
@@ -102,7 +123,8 @@ describe('actions', () => {
 				id: '1'
 			}
 		};
-		addProjectMembersRequest = jest.fn().mockReturnValue({isSuccess: false});
+		jest.spyOn(api, 'addProjectMembersRequest').mockReturnValue({isSuccess: false});
```
crawter: will new Response<nil>() will be better for understanding instead of {isSuccess: false}?

PR #1084, opened 2019-01-17
Closing context doesn't stop storage node

aleitner:


pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -37,9 +37,7 @@ func newService(log *zap.Logger, interval time.Duration, rt *kademlia.RoutingTab
 }
 
 // Run runs the bucket refresher service
-func (service *refreshService) Run(ctx context.Context) (err error) {
-	defer mon.Task()(&ctx)(&err)
```
navillasa: why take out mon.Task?

pkg/server/config.go
```diff
@@ -52,7 +52,11 @@ func (sc Config) Run(ctx context.Context,
 	if err != nil {
 		return err
 	}
-	defer func() { _ = s.Close() }()
+
+	go func() {
+		<-ctx.Done()
+		_ = s.Close()
```
navillasa: should this error be checked?

pkg/piecestore/psserver/bucket_refresher.go
navillasa: oh since the new version doesn't return an error is it not necessary?

pkg/server/config.go
aleitner: nope

pkg/piecestore/psserver/bucket_refresher.go
aleitner: exactly

pkg/server/config.go
navillasa: why not lol

PR #1089, opened 2019-01-17
Try accessing bootstrap node for verification

mniewrzal:


pkg/kademlia/kademlia.go
```diff
@@ -193,6 +193,14 @@ func (k *Kademlia) Bootstrap(ctx context.Context) error {
 	if len(k.bootstrapNodes) == 0 {
 		return BootstrapErr.New("no bootstrap nodes provided")
 	}
+
+	for _, node := range k.bootstrapNodes {
+		_, err := k.nodeClient.Ping(ctx, node)
+		if err != nil {
+			return err
```
thepaul: is it really an error if we fail to ping _any_ of the bootstrap nodes? how about returning an error only if we fail to ping _all_ of them

mniewrzal: sounds good

bootstrap/peer.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// servers
+	Public struct {
+		Listener net.Listener
+		Server   *server.Server
+	}
+
+	// services and endpoints
+	// TODO: similar grouping to satellite.Peer
+	RoutingTable      *kademlia.RoutingTable
```
egonelbre: Here you can use structs similarly to how satellite does:
https://github.com/storj/storj/pull/1119/files#diff-e18cb5f53ca9336df150764f9fc70167R105 (ignore kdb/ndb thing there).

internal/testplanet/planet.go
```diff
@@ -116,34 +122,46 @@ func NewWithLogger(log *zap.Logger, satelliteCount, storageNodeCount, uplinkCoun
...
-			err := peer.Run(ctx)
-			if err == grpc.ErrServerStopped {
-				err = nil
-			}
-			if err != nil {
-				// TODO: better error handling
-				panic(err)
-			}
-		}(peer)
+		if peer.Local().Type == pb.NodeType_BOOTSTRAP {
```
egonelbre: It should be fine to start everything in a single loop since they are started in a goroutine. This change doesn't really help that much.

internal/testplanet/planet.go
```diff
@@ -350,6 +366,67 @@ func (planet *Planet) newStorageNodes(count int) ([]*storagenode.Peer, error) {
...
+	// if err != nil {
+	// 	return nil, err
+	// }
+
+	planet.databases = append(planet.databases, db)
+
+	config := bootstrap.Config{
+		Server: server.Config{
+			Address:            "127.0.0.1:0",
+			RevocationDBURL:    "bolt://" + filepath.Join(planet.directory, "revocation.db"),
```
egonelbre: This should be `filepath.Join(dbdir, "revocation.db")`. If it's copy-paste, then the bug is elsewhere as well.

pkg/kademlia/kademlia.go
```diff
@@ -188,14 +189,27 @@ func (k *Kademlia) Bootstrap(ctx context.Context) error {
 	if len(k.bootstrapNodes) == 0 {
 		return BootstrapErr.New("no bootstrap nodes provided")
 	}
+
+	var errs utils.ErrorGroup
```
egonelbre: use `errs.Group`

pkg/kademlia/kademlia.go
```diff
@@ -188,14 +189,27 @@ func (k *Kademlia) Bootstrap(ctx context.Context) error {
...
+	for _, node := range k.bootstrapNodes {
+		_, err := k.dialer.Ping(ctx, node)
+		if err != nil {
+			errs.Add(err)
+		}
+	}
+	err := errs.Finish()
```
egonelbre: I think we should continue if we have at least one successful ping.

internal/testplanet/planet.go
kaloyan-raev: Yes, it was a copy-paste. I will fix it in `newStorageNodes` too.

PR #1091, opened 2019-01-18
Set Identity Folder to Read Only after Identity Signature

stefanbenten:
To protect the identity against overwriting or deletion, set it to read only after successful signing from the CA

### DIFF
--- a/pkg/identity/identity.go
+++ b/pkg/identity/identity.go
@@ -13,7 +13,6 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/identity/identity.go
```diff
@@ -285,10 +285,11 @@ func (ic Config) Save(fi *FullIdentity) error {
 		return writeErr
 	}
 
-	return utils.CombineErrors(
-		writeChainDataErr,
-		writeKeyDataErr,
-	)
+	dataErr := utils.CombineErrors(writeChainDataErr, writeKeyDataErr)
```
egonelbre: This can be left as is.

stefanbenten:
This is not needed any longer 👍

PR #1092, opened 2019-01-18
Kademlia Dialer

egonelbre:


pkg/kademlia/pool.go
```diff
@@ -0,0 +1,209 @@
...
+	limit  sync2.Semaphore
+	mu     sync.Mutex
+	closed bool
+	recent []*Conn
+}
+
+// Conn represents a cached kademlia cache.
+type Conn struct {
+	pool     *Pool
+	refcount int32 // only modify when holding pool.mu
```
jtolds: oh whoa uh, okay

jtolds: i feel like the right thing here is to have conn not have the refcount, but have Pool have a mapping of Conns to refcounts internally

pkg/kademlia/pool.go
```diff
@@ -0,0 +1,209 @@
...
+
+	for i, conn := range pool.recent {
+		// TODO: verify also other properties
+		if conn.addr == target.GetAddress().Address {
+			conn.acquireLocked()
+
+			// this is a simplest way to implement recently used cache
+			// avoiding random connections pushing important connections off
+			k := i / 2
+			pool.recent[k], pool.recent[i] = pool.recent[i], pool.recent[k]
```
jtolds: neat

pkg/kademlia/dialer.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer dialer.limit.Unlock()
+
+	conn, err := dialer.dial(ctx, ask)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := conn.client.Query(ctx, &pb.QueryRequest{
+		Limit:    20,
```
jtolds: not part of this PR but this should absolutely be passed in and not hardcoded. this should be the kademlia k value, routing table depth, etc

pkg/kademlia/dialer.go
```diff
@@ -0,0 +1,101 @@
...
+		Sender:   &self,
+		Target:   &find,
+		Pingback: true,
```
jtolds: not part of this PR but pingback should not always be true. in fact, it should only be true during routine kademlia specific bucket refreshes

pkg/kademlia/dialer.go
egonelbre: yeah, noticed this as well, I'll add a TODO

pkg/kademlia/dialer.go
egonelbre: I'll add a TODO

pkg/kademlia/dialer.go
```diff
@@ -0,0 +1,101 @@
...
+	})
+	if err != nil {
+		return nil, errs.Combine(err, conn.disconnect())
+	}
+
+	return resp.Response, errs.Combine(err, conn.disconnect())
```
thepaul: err is guaranteed to be nil here. still call errs.Combine?

PR #1093, opened 2019-01-18
certificates config and --certs-dir fixes

bryanchriswhite:

PR #1094, opened 2019-01-18
Coyle/cert inspector

coyle:
Loads identity instead of creating one


thepaul:
Oh wait this looks like it conflicts heavily with #1093. Are we working at cross-purposes?

thepaul:
I've been educated. They don't conflict and all is well in Storj world.

PR #1096, opened 2019-01-18
pkg/node: fix connection logging

jtolds:


pkg/node/server.go
```diff
@@ -36,19 +36,19 @@ func (server *Server) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Quer
...
-
-		err = rt.ConnectionSuccess(req.Sender)
-		if err != nil {
-			server.log.Error("could not respond to connection success", zap.Error(err))
+		} else {
+			err = rt.ConnectionSuccess(req.Sender)
+			if err != nil {
+				server.log.Error("could not respond to connection success", zap.Error(err))
+			} else {
+				server.log.Sugar().Infof("Successfully connected with %s", req.Sender.Address.Address)
```
egonelbre: do we want to log this all the time?

jtolds: at least for now it's nice. maybe we move it to debug later

coyle: maybe debug?

PR #1101, opened 2019-01-18
Coyle/dockerfile

coyle:


cmd/storagenode/alpha/dashboard.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+./storagenode dashboard --config-dir /app/config --identity-dir /app/identity
```
kaloyan-raev: For now we should also have this argument: `--external-address=${ADDRESS}`. It is necessary for the storage node to request the bootstrap node to ping it with the proper external address.

coyle: Completely forgot about this! I'll add it in thanks @kaloyan-raev !

cmd/storagenode/alpha/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+set -euo pipefail
+
+if [[ ! -f "config/config.yaml" ]]; then
+	./storagenode setup --config-dir config --identity-dir identity
+fi
+
+RUN_PARAMS="${RUN_PARAMS:-} --config-dir config"
+RUN_PARAMS="${RUN_PARAMS:-} --identity-dir identity"
+
+RUN_PARAMS="${RUN_PARAMS:-} --kademlia.bootstrap-addr=bootstrap.storj.io:8888"
```
egonelbre: It's not using `STORJ_KADEMLIA_BOOTSTRAP_ADDR` env variable

cmd/storagenode/alpha/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+RUN_PARAMS="${RUN_PARAMS:-} --metrics.app-suffix=-alpha"
```
egonelbre: It's not using `STORJ_METRICS_APP_SUFFIX` env variable

kaloyan-raev: The only flags we should pass to `storagenode run` should be just `--config-dir` and `identity-dir`. All the rest should be configured via the standard Corba env vars, i.e. `STORJ_KADEMLIA_EXTERNAL_ADDRESS` instead of just `ADDRESS`.

cmd/storagenode/alpha/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+RUN_PARAMS="${RUN_PARAMS:-} --metrics.interval=30m"
+RUN_PARAMS="${RUN_PARAMS:-} --server.use-peer-ca-whitelist=true"
+RUN_PARAMS="${RUN_PARAMS:-} --kademlia.external-address=${ADDRESS}"
+RUN_PARAMS="${RUN_PARAMS:-} --kademlia.operator.email=${EMAIL}"
+RUN_PARAMS="${RUN_PARAMS:-} --kademlia.operator.wallet=${WALLET}"
+RUN_PARAMS="${RUN_PARAMS:-} --storage.allocated-bandwidth=${BANDWIDTH}"
+RUN_PARAMS="${RUN_PARAMS:-} --storage.allocated-disk-space=${STORAGE}"
+RUN_PARAMS="${RUN_PARAMS:-} --storage.whitelisted-satellite-i-ds=\"12EayRS2V1kEsWESU9QMRseFhdxYxKicsiFmxrsLZHeLUtdps3S\", \"1Vrf9xmmHw6KaVFMcfR2YPt8YpVVoQZGTUJyjYc6CajeYrAqrB\""
```
jtolds: there shouldn't be these escaped quotes?

PR #1107, opened 2019-01-19
Use storagenode.Peer for storagenode

egonelbre:


pkg/piecestore/psserver/bucket_refresher.go
```diff
@@ -19,16 +19,17 @@ var (
 	Error = errs.Class("kademlia bucket refresher error")
 )
 
-// refreshService contains the information needed to run the bucket refresher service
-type refreshService struct {
+// Monitor contains the information needed to run the bucket refresher service
```
coyle: I like the name change!

storagenode/peer.go
```diff
@@ -107,22 +113,29 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config) (*P
...
+
 		kdb, ndb := peer.DB.RoutingTable()
 		peer.RoutingTable, err = kademlia.NewRoutingTable(peer.Log.Named("routing"), self, kdb, ndb)
 		if err != nil {
 			return nil, errs.Combine(err, peer.Close())
 		}
 
 		// TODO: reduce number of arguments
-		peer.Kademlia, err = kademlia.NewWith(peer.Log.Named("kademlia"), self, nil, peer.Identity, config.Alpha, peer.RoutingTable)
+		peer.Kademlia, err = kademlia.NewWith(peer.Log.Named("kademlia"), self, []pb.Node{*in}, peer.Identity, config.Alpha, peer.RoutingTable)
```
coyle: why do we want a node now ?

egonelbre: Because initially it was only used in testplanet which sets the bootstrap nodes via `SetBootstrapNodes`, because at the creation time, we don't have the necessary info yet.

However storagenode binary doesn't specify it separately. I've considered always specify it outside, but it probably makes more sense to pass it in this way.

kaloyan-raev:
Useful INFO logs like:
```
2019-01-22T22:30:43.939+0200	INFO	Operator email: kaloyan@storj.io
2019-01-22T22:30:43.939+0200	INFO	Operator wallet: 0x1234567890123456789012345678901234567890
2019-01-22T22:30:43.962+0200	INFO	Node 198QTGZ5okQAefpy1jYmY2caEcvfghmhBnLcB8dtv8Wuhr3vmL started on :28967
```
have disappeared with this PR. Can we have them back?

littleskunk:
I am unable to start my storage node. The email address should be optional. With this PR it will be required.

pkg/kademlia/config.go
```diff
@@ -41,6 +43,40 @@ type OperatorConfig struct {
...
+func (c OperatorConfig) Verify(log *zap.Logger) error {
+	if err := isOperatorEmailValid(log, c.Email); err != nil {
+		return err
+	}
+
+	if err := isOperatorWalletValid(log, c.Wallet); err != nil {
+		return err
+	}
+
+	log.Sugar().Info("Operator email: ", c.Email)
```
kaloyan-raev: Can we show this only if the email is actually set? Currently I see this:
```
2019-01-23T11:45:17.682+0200	WARN	Operator email address isn't specified.
2019-01-23T11:45:17.682+0200	INFO	Operator email: 
```
Perhaps, move the logging into the `isOperatorEmailValid` function.

egonelbre: Although the config `Verify` part is one of the thing I'm unsure about... there might be a generally better way to handle this.

PR #1113, opened 2019-01-22
Change default difficulty to 30

egonelbre:


cmd/storj-sim/network.go
```diff
@@ -271,6 +271,8 @@ func identitySetup(network *Processes) (*Processes, error) {
 		identity.Arguments = Arguments{
 			"setup": {
 				"--identity-dir", process.Directory,
+				"--concurrency", "1",
```
wthorp: Why 1?

egonelbre: Since we are likely to find with a cert with one core, no need to have contention between the processes.

PR #1114, opened 2019-01-22
Semantic Versioning and Binary updates

aleitner:
TODO: Vet the go-github-selfupdate package dependencies


cmd/storagenode/main.go
```diff
@@ -15,9 +16,11 @@ import (
 	"sort"
 	"text/tabwriter"
 
+	"github.com/Masterminds/semver"
 	"github.com/fatih/color"
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/rhysd/go-github-selfupdate/selfupdate"
```
egonelbre: The package and it's dependencies need to be reviewed.

cmd/storagenode/main.go
```diff
@@ -15,9 +16,11 @@ import (
 	"sort"
 	"text/tabwriter"
 
+	"github.com/Masterminds/semver"
```
egonelbre: The package and it's dependencies need to be reviewed.

cmd/storagenode/main.go
egonelbre: Also one of the dependencies has an issue reported with McAfee anti-virus https://github.com/inconshreveable/go-update/issues/37

PR #1116, opened 2019-01-22
storagenode approvedSatelliteIDs

cam-a:
storagenode only accepts data from satellites it has approved

### DIFF
--- a/pkg/piecestore/psserver/config.go
+++ b/pkg/piecestore/psserver/config.go
@@ -18,6 +18,8 @@ var (
 // Config contains everything necessary for a server
 type Config struct {
 	Path                         string        `help:"path to store data in" default:"$CONFDIR/storage"`
+	ApprovedSatelliteIDs         string        `help:"a comma-separated list of approved satellite node ids" default:""`
+	SatelliteIDRestriction       bool          `help:"if true, restrict allowing puts to the approved satellite id list" default:"false"`
 	AllocatedDiskSpace           memory.Size   `user:"true" help:"total allocated disk space in bytes" default:"1TB"`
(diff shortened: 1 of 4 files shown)


pkg/piecestore/psserver/config.go
```diff
@@ -18,6 +18,8 @@ var (
 // Config contains everything necessary for a server
 type Config struct {
 	Path                         string        `help:"path to store data in" default:"$CONFDIR/storage"`
+	ApprovedSatelliteIDs         string        `help:"a comma-separated list of approved satellite node ids" default:""`
```
aleitner: renamed approved to Whitelisted

pkg/piecestore/psserver/readerwriter.go
```diff
@@ -114,3 +122,13 @@ func (s *StreamReader) Read(b []byte) (int, error) {
 
 	return n, nil
 }
+
+// approved returns true if a node ID exists in a list of approved node IDs
+func approved(id storj.NodeID, list []storj.NodeID) bool {
```
jenlij: maybe move this to the server file and use a server receiver so you can use server.whitelist rather than having to pass in a list of nodeids ie `func (s *Server) approved(id storj.NodeID) bool {...}`

pkg/piecestore/psserver/config.go
```diff
@@ -18,6 +18,8 @@ var (
...
+	SatelliteIDRestriction       bool          `help:"if true, restrict allowing puts to the approved satellite id list" default:"false"`
```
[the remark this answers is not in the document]
cam-a: lol I see what you mean

PR #1117, opened 2019-01-23
Make Bandwidth Agreements Secure / Trustable

wthorp:
Goals of this PR:
Fix Bandwidth agreements so that a storage node can trust them.
Right now only the satellite can trust them.


pkg/bwagreement/server.go
```diff
@@ -12,10 +12,45 @@ import (
...
+	//UpPrivKey wraps errors related to uplink private keys
+	UpPrivKey = errs.Class("Processing uplink private key")
+	//Renter wraps errors related to renter bandwidth allocations
+	Renter = errs.Class("Renter agreement")
+	//Payer wraps errors related to payer bandwidth allocations
+	Payer = errs.Class("Payer agreement")
+	//Missing indicates missing or empty information
+	Missing = errs.Class("Required field is empty")
+	//Ecdsa indicates a key was not an ECDSA key
+	Ecdsa = errs.New("Key is not ecdsa key")
```
kaloyan-raev: This should be all caps as it is an acronym.

pkg/bwagreement/test/server_test.go
```diff
@@ -6,19 +6,26 @@ package test
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
 
 	"storj.io/storj/internal/testcontext"
 	"storj.io/storj/internal/testidentity"
-	"storj.io/storj/internal/teststorj"
 	"storj.io/storj/pkg/bwagreement"
+	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/pb"
+	e "storj.io/storj/pkg/pb"
```
egonelbre: Duplicate import

pkg/piecestore/psserver/server_test.go
```diff
@@ -260,7 +252,7 @@ func TestRetrieve(t *testing.T) {
 					}),
 				}
 
-				s, err := cryptopasta.Sign(ba.Data, TS.k.(*ecdsa.PrivateKey))
+				s, err := cryptopasta.Sign(ba.Data, upID.Key.(*ecdsa.PrivateKey))
 				assert.NoError(err)
 				ba.Signature = s
```
littleskunk: `ba.PubKey` is missing. How can this test case work if it send invalid bwagreements? Maybe the storage node is not validating them?

wthorp: given the information you've presented, I'd say the storage node is not validating them.  now I need to figure out why.

littleskunk: The complete testset was disabled. I enabled them and was able to fix them.

pkg/piecestore/psserver/server_test.go
```diff
@@ -649,95 +591,45 @@ func newTestServerStruct(t *testing.T, ids []storj.NodeID) (*Server, func()) {
...
-			t.Fatal(serr)
-		}
-		// TODO:fix this error check
-		_ = os.RemoveAll(tmp)
-		// if err := os.RemoveAll(tmp); err != nil {
-		// 	t.Fatal(err)
-		// }
+	//init ps server grpc
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	assert.NoError(t, err)
```
jtolds: not a big deal but you probably want require.NoError for this type of code style. assert.NoError marks the test as failed but lets the test keep going. require.NoError halts the test.

jtolds: i do see that this affects most of this test suite

wthorp: Ah!  I didn't know about that one

pkg/pb/BandwidthUtils.go
```diff
@@ -0,0 +1,63 @@
...
+	} else if sm.GetSignature() == nil {
+		return false, Missing.New("message signature")
+	} else if sm.GetCerts() == nil {
+		return false, Missing.New("message certificates")
+	}
+	return true, nil
+}
+
+//Unpack helps get things out of a RenterBandwidthAllocation
+func (rba *RenterBandwidthAllocation) Unpack() (*RenterBandwidthAllocation_Data, *PayerBandwidthAllocation, *PayerBandwidthAllocation_Data, error) {
```
jtolds: should unpack check the signatures just so it's less easy to forget to do so in the future?

wthorp: In cases like audit, the signatures have already been validated.

PR #1119, opened 2019-01-23
Satellite Peer

egonelbre:
Ready for review, but still needs fixes for shutting down (race between kademlia refreshes and database closing).


internal/testplanet/planet.go
```diff
@@ -267,7 +279,19 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				MaxBufferMem:  4 * memory.MB,
 				APIKey:        "",
 			},
-			// TODO: Audit    audit.Config
+			Audit: audit.Config{
+				MaxRetriesStatDB: 0,
+				Interval:         30 * time.Second,
+			},
+			Tally: tally.Config{
+				Interval: 30 * time.Second,
```
jenlij: can this be reconfigured somewhere?

egonelbre: This is the testplanet config, but at the moment no. For the satellite binary it takes the argument from flags.

For testplanet I have two ideas with regards to disabling/adjusting configuration:
1. hooks for config (e.g. you could specify callbacks to change the config before creating the peer)
2. flags for disabling specific subsystems (e.g. you want to manually play with manually and you don't want to run it)

satellite/peer.go
```diff
@@ -346,6 +420,11 @@ func (peer *Peer) Close() error {
 		errlist.Add(peer.Kademlia.RoutingTable.SelfClose())
 	}
 
+	// if peer.Kademlia.ndb != nil || peer.Kademlia.kdb != nil {
```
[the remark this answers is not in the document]
egonelbre: Yes, we don't actually want to leak routing table. However, closing the underlying database causes a race with kademlia, since closing kademlia doesn't wait for all lookups to be returned... there probably are other subsystems that need to wait in Close to finish all pending jobs.

i.e. not pretty, but needs to be fixed.

PR #1120, opened 2019-01-23
init

rikysya:


satellite/console/consoleweb/server.go
```diff
@@ -0,0 +1,139 @@
...
+	applicationJSON    = "application/json"
+	applicationGraphql = "application/graphql"
+)
+
+// Error is satellite console error type
+var Error = errs.Class("satellite console error")
+
+// Server represents console web server
+type Server struct {
+	logger *zap.Logger
```
egonelbre: use `log` as name, the logging lines look nicer that way

satellite/console/consoleweb/server.go
```diff
@@ -0,0 +1,139 @@
...
+
+	config   Config
```
egonelbre: Move `type Config struct` above `type Server struct`, it makes it easier to see what the config arguments are for this server.

satellite/peer.go
```diff
@@ -155,6 +160,11 @@ type Peer struct {
 	}
 
 	// TODO: add console
```
egonelbre: remove todo

PR #1121, opened 2019-01-23
quick log cleanup for storage node

aleitner:


pkg/piecestore/psserver/server.go
```diff
@@ -145,9 +145,9 @@ func (s *Server) Stop(ctx context.Context) error {
 	)
 }
 
-// Piece -- Send meta data about a stored by by Id
+// Piece -- Send meta data about a piece stored by Id
 func (s *Server) Piece(ctx context.Context, in *pb.PieceId) (*pb.PieceSummary, error) {
-	s.log.Debug("Getting Meta", zap.String("Piece ID", in.GetId()))
+	s.log.Info("Getting Meta", zap.String("Piece ID", in.GetId()))
```
egonelbre: Try to keep one `Info` entry per call instead of multiple. So the entry-point could be `Debug` and when it exists with an error/successfully it would be `Info`.

aleitner: done

pkg/piecestore/psserver/server.go
```diff
@@ -199,14 +199,25 @@ func (s *Server) Piece(ctx context.Context, in *pb.PieceId) (*pb.PieceSummary, e
...
 // Stats will return statistics about the Server
 func (s *Server) Stats(ctx context.Context, in *pb.StatsReq) (*pb.StatSummary, error) {
 	s.log.Debug("Getting Stats...")
 
+	statsSummary, err := s.retrieveStats()
+	if err != nil {
+		return nil, err
+	}
+
+	s.log.Info("Succesfully retrieved Stats...")
```
jtolds: sp: Successfully

PR #1122, opened 2019-01-23
Making sure payments works

jenlij:


pkg/accounting/tally/tally.go
```diff
@@ -5,6 +5,7 @@ package tally
 
 import (
 	"context"
+	"fmt"
```
mobyvb: unused import?

PR #1123, opened 2019-01-23
allows requests to overlay to include new nodes

navillasa:
Supersedes https://github.com/storj/storj/pull/933

Adds to the NodeSelectionConfig in the overlay pkg:

- NewNodeAuditThreshold: the number of audits a node must have to not be considered a New Node
- NewNodePercentage: the percentage of new nodes allowed per request
- MatchingNodeRatio: the ratio of requested nodes to matching nodes for a given list of restrictions

The overlay server's FindStorageNodes now gets reputable nodes and new nodes based on config settings.

===============

Individually, all of the tests pass except for the last one, but they fail when run all together. Not sure what's going on there.

Also the last test case inconsistently passes/fails and I'm not sure why. If I set:
``` 
			requestedNodeAmt:      x,
			reputableNodes:        x,
			newNodePercentage:     0,
			expectedResultLength:  x,
			newNodeAuditThreshold: 1,
			totalNodes:            x,
```
Then the case will pass the majority of times, but sometimes it will fail because the actual length of the result nodes will be x-1 instead of x. I'm not sure if it's a problem with the test environment (where a node of type satellite is slipping in there somehow) or a problem in the code, so please check it out if you get a chance!


satellite/peer.go
```diff
@@ -247,14 +247,14 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
...
-		ns := &pb.NodeStats{
+		minStats := &pb.NodeStats{
 			UptimeCount:       config.Node.UptimeCount,
 			UptimeRatio:       config.Node.UptimeRatio,
 			AuditSuccessRatio: config.Node.AuditSuccessRatio,
 			AuditCount:        config.Node.AuditCount,
 		}
 
-		peer.Overlay.Endpoint = overlay.NewServer(peer.Log.Named("overlay:endpoint"), peer.Overlay.Service, ns)
+		peer.Overlay.Endpoint = overlay.NewServer(peer.Log.Named("overlay:endpoint"), peer.Overlay.Service, minStats, 0, 1, 0)
```
egonelbre: Overlay endpoint should take a struct, not this many arguments.

pkg/overlay/server_test.go
```diff
@@ -71,3 +75,171 @@ func TestServer(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		planet.Start(ctx)
+		// we wait a second for all the nodes to complete bootstrapping off the satellite
+		time.Sleep(2 * time.Second)
+
+		satellite := planet.Satellites[0]
+
+		server := overlay.NewServer(satellite.Log.Named("overlay"), satellite.Overlay.Service,
```
egonelbre: If we make server have the defaults and service the the configuration we don't need to recreate testplanet for every test.

pkg/overlay/server.go
```diff
@@ -23,19 +23,26 @@ var ServerError = errs.Class("Server Error")
...
-	cache     *Cache
-	metrics   *monkit.Registry
-	nodeStats *pb.NodeStats
+	log                   *zap.Logger
+	cache                 *Cache
+	metrics               *monkit.Registry
+	minStats              *pb.NodeStats
+	matchingNodeRatio     float64
+	newNodeAuditThreshold int64
+	newNodePercentage     float64
```
egonelbre: Let's make a proper struct for all of the arguments.

pkg/overlay/server.go
```diff
@@ -74,52 +81,76 @@ func (server *Server) FindStorageNodes(ctx context.Context, req *pb.FindStorageN
 	defer mon.Task()(&ctx)(&err)
 
 	opts := req.GetOpts()
-	maxNodes := req.GetMaxNodes()
-	if maxNodes <= 0 {
-		maxNodes = opts.GetAmount()
+	// todo(nat): change name from MaxNodes to minRequestedNodes or similar
```
egonelbre: Let's make this logic part of cache (later renamed to Service), it makes  it easier to test as well.

pkg/overlay/server.go
```diff
@@ -74,52 +81,76 @@ func (server *Server) FindStorageNodes(ctx context.Context, req *pb.FindStorageN
...
-				usedAddrs[addr] = true
-			}
-		}
-		if len(resultNodes) <= 0 {
+		reputableNodes, excluded, usedAddrs = server.filterNodes(ctx, reputableNodes, requiredReputableNodes, usedAddrs, excluded)
+
+		allReputableNodes = append(allReputableNodes, reputableNodes...)
+
+		// this means we've exhausted the overlay cache looking for reputable nodes or we have enough
+		if reputableStartID == (storj.NodeID{}) || int64(len(allReputableNodes)) >= requiredReputableNodes {
```
egonelbre: nodeid has method `IsEmpty`.

pkg/overlay/server.go
```diff
@@ -128,49 +159,151 @@ func (server *Server) FindStorageNodes(ctx context.Context, req *pb.FindStorageN
...
+	var resultNodes []*pb.Node
+
 	for _, v := range nodes {
 		if v == nil {
 			continue
 		}
 
-		nextStart = v.Id
 		if v.Type != pb.NodeType_STORAGE {
+			server.log.Debug("not storage node = " + v.Id.String() + " was " + v.Type.String())
```
egonelbre: remove the debug line

pkg/overlay/server.go
```diff
@@ -128,49 +159,151 @@ func (server *Server) FindStorageNodes(ctx context.Context, req *pb.FindStorageN
...
 		}
 
 		restrictions := v.GetRestrictions()
 		reputation := v.GetReputation()
 
+		if reputation.GetAuditCount() < server.newNodeAuditThreshold {
+			continue
+		}
+
 		if restrictions.GetFreeBandwidth() < minRestrictions.GetFreeBandwidth() ||
```
egonelbre: let's create a separate method for checking this, rather than having a large if statement.

pkg/overlay/server.go
navillasa: I see `IsZero` not `IsEmpty`, is that doing the same thing?

navillasa:
refactored here https://github.com/storj/storj/pull/1136

PR #1135, opened 2019-01-24
payments accesses database directly

cam-a:
refactor payments away from grpc client/server architecture for querying database


pkg/accounting/payments/payments.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"storj.io/storj/pkg/accounting"
+	"storj.io/storj/pkg/utils"
+	"storj.io/storj/satellite/satellitedb"
+)
+
+// GenerateCSV generates a payment report for all nodes for a given period
+func GenerateCSV(ctx context.Context, cfgDir string, dbPath string, id string, start time.Time, end time.Time) (string, error) {
+
+	//TODO: make sure  we can get config-dir of storj-sim
```
jenlij: we can remove this TODO if done

cmd/storj-sim/network.go
```diff
@@ -258,6 +261,48 @@ func newNetwork(flags *Flags) (*Processes, error) {
 	return processes, nil
 }
 
+// networkPayments generates a satellite's payment report for a given period
+func networkPayments(flags *Flags, args []string) error {
```
egonelbre: This change shouldn't be needed, you can run payments binary against the network rather than specifying a separate command.

cmd/payments/payments.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	layout := "2006-01-02"
+	filename := id + "--" + start.Format(layout) + "--" + end.Format(layout) + ".csv"
+	path := filepath.Join(pDir, filename)
+	file, err := os.Create(path)
+	if err != nil {
+		return "", err
+	}
+	defer utils.LogClose(file)
```
egonelbre: Do not use LogClose for new code.

cmd/payments/payments.go
```diff
@@ -0,0 +1,112 @@
...
+func generateCSV(ctx context.Context, cfgDir string, dbPath string, id string, start time.Time, end time.Time) (string, error) {
+
+	db, err := satellitedb.New(dbPath)
+	if err != nil {
+		return "", errs.New("error connecting to master database on satellite: %+v", err)
+	}
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			fmt.Printf("error closing connection to master database on satellite: %+v\n", err)
```
egonelbre: either consistently log, or return... don't do both

cmd/payments/payments.go
cam-a: Is there a new we are closing and checking errors?

egonelbre: file.Close and errs.Combine or wrap depending on the situation

cmd/satellite/main.go
```diff
@@ -68,6 +75,11 @@ var (
 		Database   string `help:"satellite database connection string" default:"sqlite3://$CONFDIR/master.db"`
 		QListLimit int    `help:"maximum segments that can be requested" default:"1000"`
 	}
+	paymentsCfg struct {
```
egonelbre: Use `runCfg` for it, less likely to get defaults mismatched

cmd/satellite/main.go
```diff
@@ -101,15 +113,17 @@ func init() {
 	rootCmd.AddCommand(setupCmd)
 	rootCmd.AddCommand(diagCmd)
 	rootCmd.AddCommand(qdiagCmd)
+	rootCmd.AddCommand(paymentsCmd)
```
egonelbre: let's create a top-level command `report` or `reports`, we'll likely have more of them not just on `payments`.

cmd/satellite/main.go
```diff
@@ -101,15 +113,17 @@ func init() {
...
 	cfgstruct.BindSetup(setupCmd.Flags(), &setupCfg, cfgstruct.ConfDir(defaultConfDir), cfgstruct.IdentityDir(defaultIdentityDir))
 	cfgstruct.Bind(diagCmd.Flags(), &diagCfg, cfgstruct.ConfDir(defaultConfDir), cfgstruct.IdentityDir(defaultIdentityDir))
 	cfgstruct.Bind(qdiagCmd.Flags(), &qdiagCfg, cfgstruct.ConfDir(defaultConfDir), cfgstruct.IdentityDir(defaultIdentityDir))
+	cfgstruct.Bind(paymentsCmd.Flags(), &paymentsCfg, cfgstruct.ConfDir(defaultConfDir), cfgstruct.IdentityDir(defaultIdentityDir))
 }
 
 func cmdRun(cmd *cobra.Command, args []string) (err error) {
 	log := zap.L()
-
+	fmt.Println(runCfg.Database)
```
egonelbre: I think this shouldn't be here.

cmd/satellite/payments.go
```diff
@@ -0,0 +1,109 @@
...
+
+	db, err := satellitedb.New(dbPath)
+	if err != nil {
+		return "", errs.New("error connecting to master database on satellite: %+v", err)
+	}
+	defer func() {
+		err = errs.Combine(err, db.Close())
+	}()
+
+	pDir := filepath.Join(cfgDir, "payments")
```
egonelbre: output to stdout instead of figuring out a magic path

egonelbre: alternatively take a path where to output it (not directory)

cmd/satellite/main.go
```diff
@@ -68,12 +79,16 @@ var (
...
+	paymentsCfg struct {
+		Database string `help:"satellite database connection string" default:"sqlite3://$CONFDIR/master.db"`
+	}
 
 	defaultConfDir = fpath.ApplicationDir("storj", "satellite")
 	// TODO: this path should be defined somewhere else
 	defaultIdentityDir = fpath.ApplicationDir("storj", "identity", "satellite")
 	confDir            string
 	identityDir        string
+	out                string
```
egonelbre: This should be part of the config rather than a global variable.

cmd/satellite/main.go
```diff
@@ -264,6 +284,39 @@ func cmdQDiag(cmd *cobra.Command, args []string) (err error) {
...
+	output := os.Stdout
+	if out != "" {
+		file, err := os.Create(out)
+		if err != nil {
+			return err
+		}
+		output = file
+	}
+	defer func() {
+		err = errs.Combine(err, output.Close())
```
egonelbre: This should be specific to file only, we don't want to close stdout, it would cause problems.

cmd/satellite/payments.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/pkg/accounting"
+	"storj.io/storj/satellite/satellitedb"
+)
+
+// generateCSV generates a payment report for all nodes for a given period
+func generateCSV(ctx context.Context, start time.Time, end time.Time, output *os.File) error {
```
egonelbre: use `output io.Writer`

cmd/satellite/main.go
```diff
@@ -57,6 +58,16 @@ var (
 		Short: "Repair Queue Diagnostic Tool support",
 		RunE:  cmdQDiag,
 	}
+	reportsCmd = &cobra.Command{
+		Use:   "reports",
+		Short: "Generate a report",
+	}
+	paymentsCmd = &cobra.Command{
```
egonelbre: Add requirement for the argument count

PR #1136, opened 2019-01-25
node selection refactor

navillasa:
- refactors FindStorageNodes, which now calls overlay cache db function FilterNodes
- adds FilterNodes function to overlay cache DB interface
- consolidates overlay’s NodeSelectionConfig
- adds TestNewNodeFiltering, which tests that the correct amount of nodes are returned given certain parameters such as NewNodeAuditThreshold and NewNodePercentage
  - NewNodeAuditThreshold: the number of audits a node must have to not be considered a New Node
  - NewNodePercentage: the ratio of new nodes requested from overlay when requesting reputable nodes
- changes proto FindStorageNodesRequest MaxNodes field to MinNodes


satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +25,159 @@ type overlaycache struct {
...
+func (cache *overlaycache) findReputableNodesQuery(ctx context.Context, req *filterReputableRequest) (*sql.Rows, error) {
+	var auditCount int64
+	if req.minReputation.AuditCount > req.newNodeAuditThreshold {
+		auditCount = req.minReputation.AuditCount
+	} else {
+		auditCount = req.newNodeAuditThreshold
+	}
+
+	// todo: add filtering for free bandwidth and free disk space and not in excluded
+	rows, err := cache.db.Query(cache.db.Rebind(`SELECT nodes.id,
```
mobyvb: replace all the arguments here with "?" and create an `args []interface` variable that you append all the arguments to. Then, call 
```
cache.db.Query(cache.db.Rebind(`SELECT .... `), args...)
```
Example at https://github.com/storj/storj/blob/master/satellite/satellitedb/statdb.go#L139-L143

Do the same with the other query.

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,251 @@ type overlaycache struct {
...
+	FROM overlay_cache_nodes
+	WHERE overlay_cache_nodes.audit_count > ?
+	AND overlay_cache_nodes.audit_count > ?
+	AND overlay_cache_nodes.audit_success_ratio > ?
+	AND overlay_cache_nodes.uptime_count > ?
+	AND overlay_cache_nodes.audit_uptime_ratio > ?
+	AND overlay_cache_nodes.free_bandwidth > ?
+	AND overlay_cache_nodes.free_disk > ?
+	AND overlay_cache_nodes.node_type == ?
+	LIMIT ?
```
thepaul: This wouldn't affect performance much, but you could increase readability some by omitting the table name from everything (because it's the default while querying from one table). Something like

```
SELECT node_id, node_type, address, free_bandwidth, free_disk, audit_success_ratio,
	audit_uptime_ratio, audit_count, audit_success_count, uptime_count, uptime_success_count
FROM overlay_cache_nodes
WHERE audit_count > ?
	AND audit_count > ?
	AND audit_success_ratio > ?
	AND uptime_count > ?
	AND audit_uptime_ratio > ?
	AND free_bandwidth > ?
	AND free_disk > ?
	AND node_type = ?
LIMIT ?
```

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,251 @@ type overlaycache struct {
...
+	overlay_cache_nodes.uptime_success_count
+	FROM overlay_cache_nodes
+	WHERE overlay_cache_nodes.audit_count > ?
+	AND overlay_cache_nodes.audit_count > ?
+	AND overlay_cache_nodes.audit_success_ratio > ?
+	AND overlay_cache_nodes.uptime_count > ?
+	AND overlay_cache_nodes.audit_uptime_ratio > ?
+	AND overlay_cache_nodes.free_bandwidth > ?
+	AND overlay_cache_nodes.free_disk > ?
+	AND overlay_cache_nodes.node_id NOT IN (?`+strings.Repeat(", ?", len(req.excluded)-1)+`)
```
thepaul: If we do still need to support excluded, this might be better sent as an array. You can send array arguments directly to pg using `pq.Array()`. IIRC this would look like

```
AND node_id NOT IN ?::text[]
```
and then you'd bind `pq.Array(req.excluded)` for the corresponding argument.

satellite/satellitedb/overlaycache.go
navillasa: oh awesome, I didn't know about this thanks!

satellite/satellitedb/overlaycache.go
navillasa: cool, thank you!

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,243 @@ type overlaycache struct {
 	db *dbx.DB
 }
 
+type filterReputableRequest struct {
```
mobyvb: I feel like you could use the same struct for both of these, and just don't use `minReputation` for your `filterNewNodes` function. That way you don't need two separate structs and you can keep everything else the same pretty uch

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,243 @@ type overlaycache struct {
...
+func (cache *overlaycache) filterReputableNodes(ctx context.Context, req *filterReputableRequest) (reputableNodes []*pb.Node, err error) {
+	rows, err := cache.findReputableNodesQuery(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = utils.CombineErrors(err, rows.Close())
+	}()
+
+	for rows.Next() {
```
mobyvb: this section as far as I can tell is identical for both `filterReputableNodes` and `filterNewNodes`, so maybe it's worth pulling out into a helper function like `sqlRowsToDbxNode` or something.

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,235 @@ type overlaycache struct {
...
+	WHERE audit_count >= ?
+	AND audit_count >= ?
+	AND audit_success_ratio > ?
+	AND uptime_count > ?
+	AND audit_uptime_ratio > ?
+	AND free_bandwidth > ?
+	AND free_disk > ?
+	AND node_id NOT IN ?::text[]
+	AND node_type == ?
+	LIMIT ?
```
thepaul: It looks like these two query cases can probably be merged now (since an empty excluded list shouldn't cause a problem anymore). Does that seem right?

satellite/satellitedb/overlaycache.go
```diff
@@ -22,6 +24,235 @@ type overlaycache struct {
...
+		node_type, address, free_bandwidth, free_disk, audit_success_ratio,
+		audit_uptime_ratio, audit_count, audit_success_count, uptime_count,
+		uptime_success_count
+		FROM overlay_cache_nodes
+		WHERE audit_count < ?
+		AND free_bandwidth > ?
+		AND free_disk > ?
+		AND node_id NOT IN ?::text[]
+		AND node_type == ?
+		LIMIT ?
```
thepaul: and these

satellite/satellitedb/overlaycache.go
navillasa: yeah that makes sense!

pkg/overlay/server.go
```diff
@@ -69,118 +66,48 @@ func (server *Server) BulkLookup(ctx context.Context, reqs *pb.LookupRequests) (
...
-
-		if restrictions.GetFreeBandwidth() < minRestrictions.GetFreeBandwidth() ||
-			restrictions.GetFreeDisk() < minRestrictions.GetFreeDisk() ||
-			reputation.GetUptimeRatio() < minReputation.GetUptimeRatio() ||
-			reputation.GetUptimeCount() < minReputation.GetUptimeCount() ||
-			reputation.GetAuditSuccessRatio() < minReputation.GetAuditSuccessRatio() ||
-			reputation.GetAuditCount() < minReputation.GetAuditCount() ||
-			contains(excluded, v.Id) {
-			continue
+		_, ok := status.FromError(err)
```
jtolds: ok isn't saying the status is okay. ok is saying that it was an error from this package. you need to actually check the real return value

jtolds: er, maybe i don't understand what you're trying to do

navillasa: I was also thinking of wrapping the resource exhausted error in a zeebo error, but wasn't sure if that was the right way to go about it since `Error` is already the default error class for overlay errors. so I wasn't sure if I should make like a child error... 😬

navillasa: oh actually I'll just do this, I think this works:
```
	foundNodes, err := server.cache.db.FilterNodes(ctx, filterNodesReq)
	if err != nil {
		stat, ok := status.FromError(err)
		if ok && stat.Code() == codes.ResourceExhausted {
			return &pb.FindStorageNodesResponse{
				Nodes: foundNodes,
			}, err
		}
		return nil, Error.Wrap(err)
	}
```

PR #1137, opened 2019-01-25
Kademlia flags cleanup

egonelbre:


pkg/piecestore/psclient/client.go
```diff
@@ -31,17 +30,9 @@ var ClientError = errs.Class("piecestore client error")
 var (
 	defaultBandwidthMsgSize = 32 * memory.KB
 	maxBandwidthMsgSize     = 64 * memory.KB
+	// TODO: make these configurable flags, removed global flag for now
```
jtolds: what's the reason to avoid these global flags? for these two values, we definitely don't know enough to know if we've landed on the right choices here yet. what harm does the init function cause?

pkg/piecestore/psserver/config.go
```diff
@@ -17,11 +17,16 @@ var (
...
+	WhitelistedSatelliteIDs string `help:"a comma-separated list of approved satellite node ids" default:""`
+	SatelliteIDRestriction  bool   `help:"if true, only allow data from approved satellites" default:"false"`
+
+	AllocatedDiskSpace memory.Size `user:"true" help:"total allocated disk space in bytes" default:"1TB"`
+	AllocatedBandwidth memory.Size `user:"true" help:"total allocated bandwidth in bytes" default:"500GiB"`
+
 	KBucketRefreshInterval       time.Duration `help:"how frequently Kademlia bucket should be refreshed with node stats" default:"1h0m0s"`
 	AgreementSenderCheckInterval time.Duration `help:"duration between agreement checks" default:"1h0m0s"`
+
+	// TTLCheckInterval time.Duration `help:"how frequently piece TTL is checked" default:"1h0m0s"`
```
jtolds: i'm a pretty big believer in not checking in commented out code. https://blog.kentcdodds.com/please-don-t-commit-commented-out-code-53d0b5b26d5f

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -63,7 +60,7 @@ func Open(ctx context.Context, storage *pstore.Storage, DBPath string) (db *DB,
 	db = &DB{
 		DB:      sqlite,
 		storage: storage,
-		check:   time.NewTicker(*defaultCheckInterval),
+		check:   time.NewTicker(time.Hour), // TODO: make configurable
```
jtolds: this was configurable? i don't understand what is motivating the removal of the global flags. what do we gain by not having the global flags?

jtolds:
i had an idea - if the goal is to reduce UX clutter when someone runs `--help`, perhaps we could leave these additional flags in, but just list them as hidden in spf13/pflag? either pkg/process or spf13/pflag itself deliberately pulls in everything defined in the stdlib flag package, but we could just make sure everything defined as a stdlib flag is just hidden in our spf13/pflag-using pkg/process system?

egonelbre:
There are few reasons, why I'd like to remove them:
* reduce flags clutter, of course hiding them is also an option, which we may need to do at some point anyways
* make it more obvious where the values are coming from, other than main...
* avoid accidental adding of flags things that don't need it, for example some piecestore flags have ended up in both uplink or satellite (cannot recall exactly which one it was), similarly go tests end up exposing the same flags
* make it easier to change defaults (i.e. similar to this problem https://github.com/storj/storj/blob/master/pkg/miniogw/config.go#L108); as an example https://github.com/jtolds/monkit-hw/blob/master/oom.go#L13 ends up in most places, however it's wrong for Windows... the only way to modify such things is to modify the dependent package... however some programs probably don't need metrics in the first place.

pkg/piecestore/psclient/client.go
egonelbre: Anyways, removed changes in piecestore and will make a proper PR to fix them.

pkg/piecestore/psserver/config.go
egonelbre: Anyways, removed changes in piecestore and will make a proper PR to fix them.

pkg/piecestore/psserver/psdb/psdb.go
egonelbre: Anyways, removed changes in piecestore and will make a proper PR to fix them.

jtolds:
yeah the monkit flag is kind of bad, i agree. i think my position is flags are better than completely hardcoded behavior (maybe that's not always true eg monkit on windows), but the config structs are even better

PR #1138, opened 2019-01-25
Separate identity from server config

egonelbre:
This separates Identity flags from Server. If we have multiple servers per peer, then it doesn't make sense to have identity multiple times.

Note this changes flags:
```
 --server.identity.key-path
 --server.identity.cert-path
```

Into:
```
 --identity.key-path
 --identity.cert-path
```

With the exception of certificates server.


phutchins:
Doesn't look like this will break anything on the deployment side.

PR #1139, opened 2019-01-25
Kademlia race

egonelbre:


pkg/kademlia/kademlia.go
```diff
@@ -185,12 +183,22 @@ func (k *Kademlia) SetBootstrapNodes(nodes []pb.Node) { k.bootstrapNodes = nodes
 // Bootstrap contacts one of a set of pre defined trusted nodes on the network and
 // begins populating the local Kademlia node
 func (k *Kademlia) Bootstrap(ctx context.Context) error {
+	defer k.bootstrapFinished.Release()
+
+	k.lookups.Add(1)
+	defer k.lookups.Done()
+
 	if len(k.bootstrapNodes) == 0 {
-		return BootstrapErr.New("no bootstrap nodes provided")
```
coyle: do we not care about not having a bootstrap node ?

pkg/kademlia/kademlia.go
```diff
@@ -299,20 +321,12 @@ func GetIntroNode(addr string) (*pb.Node, error) {
...
-	}()
-}
-
 // RunRefresh occasionally refreshes stale kad buckets
 func (k *Kademlia) RunRefresh(ctx context.Context) error {
+	k.lookups.Add(1)
+	defer k.lookups.Done()
+
 	ticker := time.NewTicker(5 * time.Minute)
-	time.Sleep(time.Duration(rand.Intn(300)) * time.Second) //stagger
```
coyle: Do you know what the point of this was originally ?

egonelbre: I think it might have been to avoid starting refresh during bootstrap.

pkg/kademlia/kademlia.go
egonelbre: I'm not sure what to do about this. I mean we do for non-bootstrap peers, but don't care for bootstrap peer. I'm not sure whether there's a better way to handle this. I mean it could also check for the node type.

pkg/kademlia/kademlia.go
```diff
@@ -104,28 +106,24 @@ func NewKademliaWithRoutingTable(log *zap.Logger, self pb.Node, bootstrapNodes [
...
 		dialer: NewDialer(log.Named("dialer"), transport.NewClient(identity, rt)),
 	}
+
 	return k, nil
 }
 
 // Close closes all kademlia connections and prevents new ones from being created.
 func (k *Kademlia) Close() error {
-	// Cancel the bootstrap context
-	ptr := atomic.LoadPointer(&k.bootstrapCancel)
```
dylanlott: Getting rid of this will be nice 👍

PR #1142, opened 2019-01-25
Improve identity cli ux:

bryanchriswhite:
+ add messaging
+ fix `--identity-dir` flag
+ function naming cleanup


pkg/identity/certificate_authority.go
```diff
@@ -95,21 +96,33 @@ func NewCA(ctx context.Context, opts NewCAOptions) (_ *FullCertificateAuthority,
...
-	log.Printf("Generating a certificate matching a difficulty of %d\n", opts.Difficulty)
+	fmt.Printf("Generating key with a minimum a difficulty of %d...\n", opts.Difficulty)
+	logStatus := func() {
+		count := atomic.LoadUint32(i)
+		hs := atomic.LoadUint32(highscore)
+		fmt.Printf("\rGenerated %d keys; best difficulty so far: %d", count, hs)
+	}
 	err = GenerateKeys(ctx, minimumLoggableDifficulty, int(opts.Concurrency),
 		func(k *ecdsa.PrivateKey, id storj.NodeID) (done bool, err error) {
+			count := atomic.LoadUint32(i)
```
bryanchriswhite: Is this awful for performance?

egonelbre: Use `count := atomic.AddUint32(i, 1)` and it's ~10ns per call.... i.e. the modulo operator can be slower.

egonelbre: * it's lower when there is no contention

bryanchriswhite: ha - that makes sense

bryanchriswhite:
@jtolds yea - i was gonna do that but wanted to make sure i got this far at least by today's test run... will do in another PR.

PR #1143, opened 2019-01-25
Add bucket to project relationship on satellite

rikysya:


cmd/storj-sim/network.go
```diff
@@ -271,6 +292,217 @@ func newNetwork(flags *Flags) (*Processes, error) {
 	return processes, nil
 }
 
+type graphqlResponse struct {
```
egonelbre: let's put this logic in a separate file

cmd/storj-sim/network.go
```diff
@@ -205,6 +211,21 @@ func newNetwork(flags *Flags) (*Processes, error) {
 			}
 
 			// TODO: maybe all the config flags should be exposed for all processes?
+			apiKey := vip.GetString("client.api-key")
+			if apiKey == "" {
+				parts := strings.Split(satellite.Address, ":")
```
egonelbre: Use https://golang.org/pkg/net/#SplitHostPort, using `strings.Split` doesn't work for IPv6 properly...

cmd/storj-sim/network.go
```diff
@@ -205,6 +211,21 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
+				consoleAddress := fmt.Sprintf("http://%s:%d/api/graphql/v0", parts[0], consolePort)
```
egonelbre: there can be multiple satellites and consoles running concurrently on the same machine

egonelbre: also users can later change both the satellite and console address in the config

cmd/storj-sim/console.go
```diff
@@ -0,0 +1,227 @@
...
+	client := http.Client{}
+
+	var (
+		createUserFormat    = "mutation {createUser(input:{email:\"%s\",password:\"%s\",firstName:\"%s\",lastName:\"%s\"})}"
+		createProjectFormat = "mutation {createProject(input:{name:\"%s\",description:\"%s\"}){id}}"
+		createAPIKeyFormat  = "mutation {createAPIKey(projectID:\"%s\",name:\"%s\"){key}}"
+		tokenFormat         = "query {token(email:\"%s\",password:\"%s\"){token}}"
+	)
+
+	alice := console.CreateUser{
```
egonelbre: Not quite sure why these types are necessary here? I would probably hardcode the full queries everywhere. This is a temporary solution.

rikysya: Just to make code more clear, i guess

cmd/storj-sim/network.go
```diff
@@ -206,6 +206,9 @@ func newNetwork(flags *Flags) (*Processes, error) {
 			}
 
 			// TODO: maybe all the config flags should be exposed for all processes?
+			// check if gateway config has an api key, if it's not
```
egonelbre: add space above so it's clear that it's a different message

satellite/console/consoleweb/consoleql/query_test.go
```diff
@@ -197,21 +201,26 @@ func TestGraphqlQuery(t *testing.T) {
...
-		//_, err = service.ActivateAccount(ctx, activationToken1)
-		//if err != nil {
-		//	t.Fatal(err)
-		//}
-		//user1.Email = "muu1@email.com"
+		t.Run("Activation", func(t *testing.T) {
+			t.Skip("skip it until we will have activation flow ready")
+
+			//TODO(yar): skip it until we will have activation flow ready
+			//activationToken1, err := service.GenerateActivationToken(
```
egonelbre: uncomment code

cmd/storj-sim/console.go
```diff
@@ -0,0 +1,227 @@
...
+	defer func() {
+		err = resp.Body.Close()
+	}()
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return graphqlResponse{}, err
+	}
+
+	fmt.Println(string(b))
```
egonelbre: also remove the debug code

cmd/storj-sim/network.go
```diff
@@ -205,6 +206,28 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
+			// create example project with key and add it to the config
+			// so that gateway can have access to the satellite
+			apiKey := vip.GetString("client.api-key")
+			if apiKey == "" {
+				consoleAddress := fmt.Sprintf(
+					"http://%s/api/graphql/v0",
+					net.JoinHostPort(host, strconv.Itoa(consolePort+i)))
```
egonelbre: Lets's use vip.GetString to get the actual address of satellite.

PR #1144, opened 2019-01-25
Storagenode online confirmation ping

dylanlott:
Lays out a simple online / offline check against the bootstrap node that appears on the dashboard


cmd/storagenode/main.go
```diff
@@ -387,7 +399,7 @@ func dashCmd(cmd *cobra.Command, args []string) (err error) {
 
 		fmt.Fprintf(color.Output, "Node ID: %s\n", color.YellowString(data.GetNodeId()))
 
-		if data.GetConnection() {
+		if online {
```
wthorp: Could we call this on a timer instead, in case of a temporary outage?

dylanlott: I'm gonna put it on a ticker 👍

cmd/inspector/main.go
```diff
@@ -131,7 +131,7 @@ type Inspector struct {
 // and the overlay cache
 func NewInspector(address, path string) (*Inspector, error) {
 	ctx := context.Background()
-	identity, err := identity.Config{
+	id, err := identity.Config{
```
egonelbre: nit: `id` in the codebase usually refers to `NodeID`, not the identity.

dylanlott:
I can update it to “fid” or something. Identity just felt wrong because it’s also a package name.   
  

  
  

  
  
>

egonelbre:
@dylanlott there are places `ident` is used, although I personally prefer and often use `identity`.

PR #1147, opened 2019-01-25
fixing entrypoint script looking for invalid flag

oCroso:


egonelbre:
Other entrypoints need this fix and it can be removed from docker-compose as well.

PR #1151, opened 2019-01-26
Implement WorkGroup

egonelbre:


pkg/kademlia/kademlia.go
```diff
@@ -185,7 +186,9 @@ func (k *Kademlia) SetBootstrapNodes(nodes []pb.Node) { k.bootstrapNodes = nodes
 func (k *Kademlia) Bootstrap(ctx context.Context) error {
 	defer k.bootstrapFinished.Release()
 
-	k.lookups.Add(1)
+	if !k.lookups.Start() {
+		return context.Canceled
```
wthorp: The context.Canceled error test is "context canceled" ... "context" seems incorrect / unintuitive here

egonelbre: I could create a custom error, but then peers would need to start expecting more different kinds of cancellations.

wthorp: makes sense

PR #1153, opened 2019-01-28
Piecestore client global flags

egonelbre:


mniewrzal:
As far I understand, after this change we will not have access to those params with command line, right?

egonelbre:
@mniewrzal ah, wait this is WIP forgot to add label

egonelbre:
Closing because of #1422

PR #1156, opened 2019-01-28
Added Pagination and SearchOptions reset on successfull adding new user to project.

Barterio:


web/satellite/src/components/team/AddUserPopup.vue
```diff
@@ -132,6 +132,13 @@ import { validateEmail } from '@/utils/validation';
                 }
 
                 this.$store.dispatch(NOTIFICATION_ACTIONS.SUCCESS, 'Members successfully added to project!');
+				this.$store.dispatch(PM_ACTIONS.SET_SEARCH_QUERY, '');
```
NikolaiYurchenko: Search query is not cleared in input. This will be not clear for user when input still filled with search query but we get not filtered members in result. Consider fetch members without clearing query

PR #1157, opened 2019-01-28
Refactors kademlia inspector to use FindNear

dylanlott:
* Removes GetBucket 
* Removes GetBuckets
* Updates identity path name


pkg/kademlia/inspector.go
```diff
@@ -30,7 +30,7 @@ func NewInspector(kad dht.DHT, identity *identity.FullIdentity) *Inspector {
 // CountNodes returns the number of nodes in the routing table
 func (srv *Inspector) CountNodes(ctx context.Context, req *pb.CountNodesRequest) (*pb.CountNodesResponse, error) {
 	// TODO: this is definitely the wrong way to get this
-	kadNodes, err := srv.dht.GetNodes(ctx, srv.identity.ID, 0)
+	kadNodes, err := srv.dht.FindNear(ctx, srv.identity.ID, 0)
```
jtolds: this 0 should probably not be 0

pkg/kademlia/kademlia_test.go
```diff
@@ -218,116 +218,6 @@ func TestRefresh(t *testing.T) {
 	assert.True(t, ts1.Equal(ts2))
 }
 
-func TestGetNodes(t *testing.T) {
```
jtolds: oh shoot, sorry, we should keep this, we just need to rename to FindNear

dylanlott: Updated 👍

pkg/kademlia/inspector.go
dylanlott: Updated to like, a billion.

cmd/inspector/main.go
```diff
@@ -131,9 +120,10 @@ type Inspector struct {
 // and the overlay cache
 func NewInspector(address, path string) (*Inspector, error) {
 	ctx := context.Background()
+
 	id, err := identity.Config{
-		CertPath: fmt.Sprintf("%s/ca.cert", path),
```
egonelbre: Read these values from the config, don't rely on hardcoded values, because these can be overridden.

dylanlott: This was only touched because it was named incorrectly, it had been changed to look for `identity.cert` and `identity.key`. I propose we leave this here and fix it elsewhere or in another PR if we want to change it.

PR #1159, opened 2019-01-28
closes db in satellite run cmd

jenlij:


cmd/satellite/main.go
```diff
@@ -137,8 +142,8 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 
 	runError := peer.Run(ctx)
 	closeError := peer.Close()
-
-	return errs.Combine(runError, closeError, db.Close())
+	err = errs.Combine(err, runError, closeError)
+	return err
```
egonelbre: `return errs.Combine(runError, closeError)`

jenlij: This will capture any error from the defer as well?

wthorp: defers run after the return, so the defer will capture this error.

jenlij: thanks!

PR #1160, opened 2019-01-28
testing DBX time hooks

wthorp:
added time hooks so we can test rollup


pkg/accounting/db.go
```diff
@@ -56,4 +56,6 @@ type DB interface {
 	SaveRollup(ctx context.Context, latestTally time.Time, isNew bool, stats RollupStats) error
 	// QueryPaymentInfo queries StatDB, Accounting Rollup on nodeID
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
+	//SetTimeHook provides a way to alter the timestamps sent to the database, for testing
+	SetTimeHook(timeHook func() time.Time)
```
egonelbre: This seems like a dangerous func to have, cannot we just add the time as an argument to the functions that need it?

wthorp: I guess I agree.  When I started, I assumed that DBX used database time, so I really wanted to use its autoinsert feature with times.  Given these hooks exist, there's almost no reason not to pass in the timestamp.

PR #1161, opened 2019-01-28
trying to fix a db race in tests

wthorp:
I'm not really sure how these races popped up, but I ran this a half dozen times, maybe I've fixed it.
go test storj.io/storj/pkg/piecestore/psserver -race -count 50


egonelbre:
I cannot reasonably see how this would fix test issues with psserver. `ioutil.TempDir` should always return a random directory so adding a separate id, shouldn't change anything.

PR #1162, opened 2019-01-29
storj-sim: read in config values for gateway

egonelbre:


cmd/storj-sim/network.go
```diff
@@ -169,21 +170,18 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
-				"ACCESS_KEY=" + accessKey,
-				"SECRET_KEY=" + secretKey,
-			},
+			Extra:      []string{},
 		})
 
 		// gateway must wait for the corresponding satellite to start up
 		process.WaitForStart(satellite)
 
+		accessKey, secretKey := randomKey(), randomKey()
```
mniewrzal: This won't be generated by `setup` command itself?

egonelbre: ah, right, we can now get rid of this part

PR #1170, opened 2019-01-29
V3-1091 Extend Users table with IsActive functionality

crawter:
In this pr i added functionality to confirm your registration.

During registration we create and insert user in db without email.
Email is now nullable.

Also this method generates new token, in which he will place ID, of created user, email, from input parameters, and expire date. 

To finish up registration and activate account - user should pass his activation token to ActivateAccount method, which will parse the token and update users email ( if needed )

### DIFF
--- a/satellite/console/consoleweb/consoleql/mutation.go
+++ b/satellite/console/consoleweb/consoleql/mutation.go
@@ -107,9 +107,9 @@ func rootMutation(service *console.Service, types Types) *graphql.Object {
 					},
 				},
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
-					input, _ := p.Args[InputArg].(string)
+					activationToken, _ := p.Args[InputArg].(string)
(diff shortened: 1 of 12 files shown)


satellite/console/consoleweb/consoleql/mutation.go
```diff
@@ -97,6 +99,24 @@ func rootMutation(service *console.Service, types Types) *graphql.Object {
...
+				Type: graphql.String,
+				Args: graphql.FieldConfigArgument{
+					InputArg: &graphql.ArgumentConfig{
+						Type: graphql.String,
+					},
+				},
+				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					input, _ := p.Args[InputArg].(string)
+
+					token, err := service.ActivateAccount(p.Context, input)
```
egonelbre: unclear what is `input` here

PR #1171, opened 2019-01-29
Cache refresher

dylanlott:
* Adds refresh capability to cache
* Adds Paginate function to cache interface
* Updates config handling for discovery service


pkg/overlay/cache.go
```diff
@@ -69,6 +71,16 @@ func (cache *Cache) Inspect(ctx context.Context) (storage.Keys, error) {
 	return nil, errors.New("not implemented")
 }
 
+// List returns a list of nodes from the cache DB
+func (cache *Cache) List(ctx context.Context, cursor storj.NodeID, limit int) ([]*pb.Node, error) {
+	return cache.db.List(ctx, cursor, limit)
+}
+
+// Paginate returns a list of `limit` nodes starting from `start` offset.
+func (cache *Cache) Paginate(ctx context.Context, limit int, offset int64) ([]*pb.Node, bool, error) {
```
egonelbre: Usually the order is "offset, limit"

pkg/overlay/cache.go
```diff
@@ -41,6 +41,8 @@ type DB interface {
 	GetAll(ctx context.Context, nodeIDs storj.NodeIDList) ([]*pb.Node, error)
 	// List lists nodes starting from cursor
 	List(ctx context.Context, cursor storj.NodeID, limit int) ([]*pb.Node, error)
+	// Paginate will page through the database nodes
+	Paginate(ctx context.Context, start int, limit int64) ([]*pb.Node, bool, error)
```
egonelbre: DB order is different from below.

pkg/overlay/cache.go
dylanlott: I could change this no problem. I just followed the function signature of the generated dbx functions.

pkg/overlay/cache.go
dylanlott: Good catch, Will update this.

pkg/discovery/service.go
```diff
@@ -22,11 +22,15 @@ var (
 
 	// Error is a general error class of this package
 	Error = errs.Class("discovery error")
+
+	// Pagination
+	offset = 0
```
jtolds: wait a minute, why is this global?

pkg/discovery/service.go
```diff
@@ -91,14 +95,47 @@ func (discovery *Discovery) Run(ctx context.Context) error {
...
+	list, more, err := discovery.cache.Paginate(ctx, discovery.config.RefreshLimit, int64(offset))
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	// more means there are more rows to page through in the cache
+	if more == false {
+		offset = 0
+	} else {
+		offset = offset + len(list)
```
jtolds: this is racy, since offset is global. should offset be something like discovery.refreshOffset?

pkg/discovery/service.go
```diff
@@ -91,14 +95,47 @@ func (discovery *Discovery) Run(ctx context.Context) error {
...
+	}
+
+	for _, node := range list {
+		ping, err := discovery.kad.Ping(ctx, *node)
+		if err != nil {
+			discovery.log.Info("could not pinging node")
+			_, err := discovery.statdb.UpdateUptime(ctx, ping.Id, false)
+			if err != nil {
+				discovery.log.Error("error updating node uptime in statdb")
```
jtolds: should we pull the node if it fails to respond to a ping?

dylanlott: I thought about this, and decided for a naive implementation to worry about it later / put it into a different PR.

pkg/discovery/service.go
aligeti: I am under the impression that there is one instance of overlay cache is running per satellite. And the refresh is a periodically ran every 'refresh interval'. Is my understanding right above the single instance of overlay cache (and refresh) per satellite?

dylanlott: We wanted it to persist it's place between loops of the refresh function so that each interval isn't biting off a huge chunk of the db. I am putting this on the discovery struct instead of making it global, though.

pkg/discovery/service.go
dylanlott: Putting it on the struct 👍

pkg/discovery/service.go
```diff
@@ -24,39 +24,49 @@ var (
...
-	kad    *kademlia.Kademlia
-	statdb statdb.DB
-
-	refreshInterval time.Duration
+	log     *zap.Logger
+	cache   *overlay.Cache
+	kad     *kademlia.Kademlia
+	statdb  statdb.DB
+	config  Config
+	refresh Refresh
```
jtolds: i know this probably feels like nitpicking but why is this another struct? why not just do `refreshOffset int64` in line here?

pkg/discovery/service.go
```diff
@@ -91,14 +101,47 @@ func (discovery *Discovery) Run(ctx context.Context) error {
...
+	if more == false {
+		discovery.refresh.offset = 0
+	} else {
+		discovery.refresh.offset = discovery.refresh.offset + int64(len(list))
+	}
+
+	for _, node := range list {
+		ping, err := discovery.kad.Ping(ctx, *node)
+		if err != nil {
+			discovery.log.Info("could not pinging node")
```
navillasa: maybe should be "could not ping node"?

PR #1179, opened 2019-01-30
Updated text fields through all project

Barterio:


internal/testplanet/planet.go
```diff
@@ -304,7 +304,7 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				Interval: 120 * time.Second,
 			},
 			Console: consoleweb.Config{
-				Address: "127.0.0.1:0",
+				Address: "127.0.0.1:8081",
```
egonelbre: There can be multiple consoles and satellites, so this must not be hardcoded. Also this is likely to clash with a developer-s host:port (see https://github.com/storj/storj/wiki/Testing#automatically-choose-ports).

Barterio: Yes, I shouldn't commit this, Forgot to remove

web/satellite/src/components/dashboard/DashboardArea.vue
```diff
@@ -19,6 +20,7 @@
             data: function () {
                 return {
                     emptyImage: EMPTY_STATE_IMAGES.PROJECT,
+                    additionalText:'bla <b>bla</b> bla',
```
NikolaiYurchenko: Test data should be cleared

PR #1180, opened 2019-01-30
Readable sizes in dashboard

aleitner:
Make dashboard in an easily readable format


aleitner:
Also gonna try using tab writer

cmd/storagenode/main.go
```diff
@@ -376,38 +378,60 @@ func dashCmd(cmd *cobra.Command, args []string) (err error) {
 			return err
 		}
 
-		clr()
+		clearScreen()
```
egonelbre: also let's move dashboard command into a separate file `dashboard.go`

PR #1183, opened 2019-01-30
Upload/download for testplanet Uplink

mniewrzal:


internal/testplanet/node_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testplanet"
+)
+
+func TestUploadDownload(t *testing.T) {
+	tctx := testcontext.New(t)
```
egonelbre: use `ctx` instead of `tctx`

internal/testplanet/node.go
```diff
@@ -4,25 +4,40 @@
...
+	"storj.io/storj/pkg/storage/segments"
+	"storj.io/storj/pkg/storage/streams"
 	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/stream"
 	"storj.io/storj/pkg/transport"
+	"storj.io/storj/satellite"
 )
 
 // Node is a general purpose
 type Node struct {
```
egonelbre: Let's rename `Node` into `Uplink` to be clearer, it's not being used for other things anymore

internal/testplanet/planet.go
```diff
@@ -68,12 +68,12 @@ type Planet struct {
 
 	peers     []Peer
 	databases []io.Closer
-	nodes     []*Node
+	nodes     []*Uplink
```
egonelbre: uplinks

internal/testplanet/uplink.go
```diff
@@ -0,0 +1,261 @@
...
+	"storj.io/storj/satellite"
+)
+
+// Uplink is a general purpose
+type Uplink struct {
+	Log              *zap.Logger
+	Info             pb.Node
+	Identity         *identity.FullIdentity
+	Transport        transport.Client
+	storageNodeCount int
```
egonelbre: Wondering whether we should take this as an argument or make it public for testing purposes.

internal/testplanet/uplink.go
```diff
@@ -0,0 +1,261 @@
...
+func (planet *Planet) newUplink(name string) (*Uplink, error) {
+	identity, err := planet.NewIdentity()
+	if err != nil {
+		return nil, err
+	}
+
+	uplink := &Uplink{
+		Log:              planet.log.Named(name),
+		Identity:         identity,
+		storageNodeCount: len(planet.StorageNodes),
```
egonelbre: Pass in the argument for storage node count rather than looking it up from planet.

internal/testplanet/uplink.go
egonelbre: Maybe let's change this to be TotalShares

mniewrzal: then it can be mistaken with RedundancyScheme.TotalShares

mniewrzal: For me `storageNodeCount` is a little bit more descriptive and I think we are not setting TotalShares == storageNodeCount. More like 8/10 or 95/100

internal/testplanet/uplink.go
```diff
@@ -0,0 +1,261 @@
...
+	readOnlyStream, err := metainfo.GetObjectStream(ctx, bucket, path)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	download := stream.NewDownload(ctx, readOnlyStream, streams)
+	defer func() { err = errs.Combine(err, download.Close()) }()
+
+	buffer := bytes.NewBuffer([]byte{})
+	_, err = io.Copy(buffer, download)
```
kaloyan-raev: You can use `ioutil.ReadAll` here.

internal/testplanet/uplink.go
```diff
@@ -0,0 +1,261 @@
...
+func (uplink *Uplink) getMetainfo(satellite *satellite.Peer) (db storj.Metainfo, ss streams.Store, err error) {
+	encScheme := uplink.getEncryptionScheme()
+	redScheme := uplink.getRedundancyScheme()
+
+	// redundancy settings
+	minThreshold := int(redScheme.RequiredShares)
+	repairThreshold := int(redScheme.RepairShares)
+	successThreshold := int(redScheme.OptimalShares)
+	maxThreshold := int(redScheme.TotalShares)
+	erasureShareSize := 1 * memory.KB
```
kaloyan-raev: Use `KiB` and `MiB` instead of `KB` and `MB`.

internal/testplanet/uplink.go
```diff
@@ -0,0 +1,261 @@
...
+		RequiredShares: int16(1 * uplink.StorageNodeCount / 5),
+		RepairShares:   int16(2 * uplink.StorageNodeCount / 5),
+		OptimalShares:  int16(3 * uplink.StorageNodeCount / 5),
+		TotalShares:    int16(4 * uplink.StorageNodeCount / 5),
+	}
+}
+
+func (uplink *Uplink) getEncryptionScheme() storj.EncryptionScheme {
+	return storj.EncryptionScheme{
+		Cipher:    storj.Cipher(1),
```
kaloyan-raev: Use the `storj.AESGCM` constant instead of `1`.

internal/testplanet/uplink_test.go
```diff
@@ -0,0 +1,40 @@
...
+	defer ctx.Cleanup()
+
+	planet, err := testplanet.New(t, 1, 10, 1)
+	require.NoError(t, err)
+	defer ctx.Check(planet.Shutdown)
+
+	planet.Start(ctx)
+	time.Sleep(2 * time.Second)
+
+	expectedData := make([]byte, 1024*1024*5)
```
kaloyan-raev: Use `5 * memory.MiB` instead of `1024*1024*5`

internal/testplanet/uplink.go
mniewrzal: Now I have error message that `EncryptionBlockSize must be a multiple of ErasureShareSize * RS MinThreshold`

kaloyan-raev: Which means that we are using `KB` instead of `KiB` somewhere else. We did such a mess with these units.

mniewrzal: ok, I found it. it was in my code.

PR #1185, opened 2019-01-30
deletes nodes from cache if unresponsive

dylanlott:


pkg/discovery/service.go
```diff
@@ -128,14 +128,17 @@ func (discovery *Discovery) Refresh(ctx context.Context) error {
 			if err != nil {
 				discovery.log.Error("error updating node uptime in statdb")
 			}
+			err = discovery.cache.Delete(ctx, node.Id)
+			if err != nil {
+				discovery.log.Error("error deleting unresponsive node from cache:", zap.Error(err))
```
egonelbre: no need to duplicate the "error" message everywhere, it's already in the log level

pkg/discovery/service.go
```diff
@@ -123,22 +123,25 @@ func (discovery *Discovery) Refresh(ctx context.Context) error {
...
 		ping, err := discovery.kad.Ping(ctx, *node)
 		if err != nil {
-			discovery.log.Info("could not pinging node")
+			discovery.log.Info("could not ping node")
 			_, err := discovery.statdb.UpdateUptime(ctx, ping.Id, false)
 			if err != nil {
-				discovery.log.Error("error updating node uptime in statdb")
+				discovery.log.Error("could not update node uptime in statdb")
+			}
+			err = discovery.cache.Delete(ctx, node.Id)
```
egonelbre: This seems extreme? Shouldn't we just remove it from kademlia routing table on failure? I mean we lose all history if we remove it from cache.

egonelbre: i.e. currently node selection is based on cache not stat db

dylanlott: It's what @jtolds said to do earlier today in stand up so I just went with it. I did consider creating a fallback or secondary cache that stored old nodes for X amount of time before we purged that and considered them a lost cause. For now, though, the Seen() cache on Kademlia acts like a sort of in-memory secondary cache that we can access if we want to try an unresponsive node again. I could add logic for that pretty easily. The secondary cache idea of mine was just larger in scope than we had talked about earlier today.

jtolds: so, i don't know if the combination of the overlay cache and statdb with the master database has made things more difficult, but the overlay cache should only consider nodes that have been found to be up in the last hour. if we don't successfully talk to a node in the last hour, it should be removed from the overlay cache. if we fail to talk to a node, it should be removed from the overlay cache. the overlay cache is a cache of what should be live in the overlay network (the dht) at any given time, with up to x amount of time of staleness accepted

egonelbre: Ah, I think it might be fine, since it will always get the latest info from statdb when it's readded to the cache. Of course, code is not clear.

PR #1186, opened 2019-01-30
node graveyard resurrection

dylanlott:
Resurrects nodes from the _graveyard_

Every X interval, the cache checks the Seen() map for nodes that it's talked to and tries to ping them again. If they've been removed from the cache in the last hour for unresponsiveness, this is their chance to be online again and get added back to the cache. If they still don't respond, they stay in Seen() and nothing happens. 

This PR address some comments in #1185

Also improves how we handle tickers in discovery/service.go


pkg/discovery/service.go
```diff
@@ -144,8 +160,39 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 	return nil
 }
 
+// graveyard attempts to ping all nodes in the Seen() map from Kademlia and adds them to the cache
+// if they respond. This is an attempt to resurrect nodes that may have gone offline in the last hour
```
jenlij: nit: keep the end of the sentence on the same line

pkg/discovery/service.go
```diff
@@ -144,8 +160,39 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
...
+// and were removed from the cache due to an unsuccessful response.
+func (discovery *Discovery) graveyard(ctx context.Context) error {
+	discovery.log.Debug("starting node graveyard resurrection")
+	seen := discovery.kad.Seen()
+	var errors utils.ErrorGroup
+
+	for _, n := range seen {
+		ping, err := discovery.kad.Ping(ctx, *n)
+		if err != nil {
+			discovery.log.Debug("could not ping node in graveyard check")
```
jenlij: maybe add the nodeID to the log?

pkg/discovery/service.go
```diff
@@ -144,8 +160,39 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
...
+			// we don't want to report the ping error to ErrorGroup because it's to be expected here.
+			continue
+		}
+
+		err = discovery.cache.Put(ctx, ping.Id, ping)
+		if err != nil {
+			discovery.log.Warn("could not update node uptime")
+			errors.Add(err)
```
jenlij: Is there a preference to use errorgroup rather than combineErrors?

dylanlott:
In other places I’ve been discouraged from adding the ID to the logs so 🤷‍♀️
  

  
  

  
  
>

jenlij:
Do we have some unit tests for this?

dylanlott:
Egon wants to move towards using errorgroup, since it’s easier to manage and under the hood it uses combine errors anyway.   
  

  
  

  
  
>

pkg/discovery/service.go
```diff
@@ -144,8 +160,39 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 	return nil
 }
 
+// graveyard attempts to ping all nodes in the Seen() map from Kademlia and adds them to the cache
+// if they respond. This is an attempt to resurrect nodes that may have gone offline in the last hour
+// and were removed from the cache due to an unsuccessful response.
+func (discovery *Discovery) graveyard(ctx context.Context) error {
+	discovery.log.Debug("starting node graveyard resurrection")
+	seen := discovery.kad.Seen()
```
aligeti: how is the graveyard discovery.kad.Seen() different from refresh's discovery.kad.Seen()? Are they looking at the same overlay cache?  graveyard()'s overlay cache(which holds non-pingable nodes) shouldn't it be different overlay cache from the refresh overlay cache?

dylanlott:
The refresh function needs to have the seen logic removed, I’ll do that   
  

  
  

  
  
>

pkg/discovery/service.go
dylanlott: I've been discouraged from adding node ID's to logs, so I didn't add it here.

pkg/discovery/service.go
dylanlott: @egonelbre has been pushing for using ErrGroup over CombineErrors. It uses CombineErrors under the hood, anyway.

pkg/discovery/service.go
```diff
@@ -71,24 +74,37 @@ func (discovery *Discovery) Close() error { return nil }
...
-
-		err = discovery.discover(ctx)
-		if err != nil {
-			discovery.log.Error("Error with cache discovery", zap.Error(err))
-		}
-
 		select {
-		case <-ticker.C: // redo
+		case <-refresh.C:
+			discovery.log.Info("REFRESH STARTING")
```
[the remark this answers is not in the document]
egonelbre: Yeah, remove the messages, these will start triggering every second for tests and aren't particularly useful.

pkg/discovery/service.go
```diff
@@ -71,24 +74,37 @@ func (discovery *Discovery) Close() error { return nil }
...
+				discovery.log.Error("Error with cache refresh: ", zap.Error(err))
+			}
+		case <-discover.C:
+			discovery.log.Info("DISCOVERY STARTING")
+			err := discovery.discover(ctx)
+			if err != nil {
+				discovery.log.Error("Error with cache discovery: ", zap.Error(err))
+			}
+		case <-graveyard.C:
+			discovery.log.Info("GRAVEYARD RESURRECTION STARTING")
```
egonelbre: Remove the message.

pkg/discovery/service.go
```diff
@@ -71,24 +74,37 @@ func (discovery *Discovery) Close() error { return nil }
...
 		select {
-		case <-ticker.C: // redo
+		case <-refresh.C:
+			discovery.log.Info("REFRESH STARTING")
+			err := discovery.refresh(ctx)
+			if err != nil {
+				discovery.log.Error("Error with cache refresh: ", zap.Error(err))
+			}
+		case <-discover.C:
+			discovery.log.Info("DISCOVERY STARTING")
```
egonelbre: Remove the messages.

pkg/discovery/service.go
```diff
@@ -71,24 +74,37 @@ func (discovery *Discovery) Close() error { return nil }
...
+				discovery.log.Error("Error with cache discovery: ", zap.Error(err))
+			}
+		case <-graveyard.C:
+			discovery.log.Info("GRAVEYARD RESURRECTION STARTING")
+			err := discovery.graveyard(ctx)
+			if err != nil {
+				discovery.log.Error("graveyard resurrection failed")
+			}
 		case <-ctx.Done():
+			discovery.log.Info("DISCOVERY DONE FIRED")
```
egonelbre: Remove message.

pkg/discovery/service.go
```diff
@@ -144,8 +160,39 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 	return nil
 }
 
+// graveyard attempts to ping all nodes in the Seen() map from Kademlia and adds them to the cache
+// if they respond. This is an attempt to resurrect nodes that may have gone offline in the last hour
+// and were removed from the cache due to an unsuccessful response.
+func (discovery *Discovery) graveyard(ctx context.Context) error {
+	discovery.log.Debug("starting node graveyard resurrection")
```
egonelbre: Remove message.

egonelbre:
Is there a test we could do that checks for the intended behavior?

PR #1187, opened 2019-01-31
storagenode dashboard: use base10 for all values

egonelbre:


kaloyan-raev:
What if I configured my available bandwidth in `TiB` and my available storage space in `TB`? In the dashboard, I would expect to see the Available and Used Bandwidth in `*iB` and the Available and Used Storage in `*B`.

egonelbre:
I really wouldn't expect that behavior either. I would expect all values be in either `?iB` or `?B` regardless of how they were configured.

PR #1192, opened 2019-01-31
[V3-1071] Enhanced project members search to full name.

Barterio:


satellite/satellitedb/projectmembers.go
```diff
@@ -46,8 +47,8 @@ func (pm *projectMembers) GetByProjectID(ctx context.Context, projectID uuid.UUI
 				INNER JOIN users u ON pm.member_id = u.id
 					WHERE pm.project_id = ?
 					AND ( u.email LIKE ? OR 
-						  u.first_name LIKE ? OR 
-						  u.last_name LIKE ? )
+						  u.first_name || u.last_name LIKE ? OR
```
egonelbre: Was there a reason we are not just using name instead of first and last name? (i.e. https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/)

crawter: @egonelbre
It just an additional check. Last name is not required but i'm pretty sure that most part of users will have some. 
Maybe it will make search a bit more comfortable.

rikysya: @egonelbre 
I guess it was a requirement

jtolds: where did the requirement come from? my preference is that the database stores "full name" and "nick name". i hadn't seen egon's link before, but it's good. to support internationalization we shouldn't assume first name and last name.

PR #1173, opened 2019-01-30
Clearer code for node selection

egonelbre:

PR #1195, opened 2019-01-31
Select nodes randomly

mobyvb:
https://storjlabs.atlassian.net/browse/V3-1128


mobyvb:
@jtolds according to this: https://stackoverflow.com/a/1209946
But I want to see if this change can be incorporated into #1173 instead of with this PR

PR #1198, opened 2019-01-31
go test fail due to consoleql UTC issues

wthorp:
Issue:
go test fails time comparison fails when using postgres backend

Steps to recreate issue:
[install postgres]
create teststorj database
run: STORJ_POSTGRES_TEST=postgres://postgres@localhost/teststorj?sslmode=disable go test ./...

Note that I'm not entirely sure if this solution fixes the problem or just fixes the test.


kaloyan-raev:
I don't reproduce this issue locally even after 100 attempts. Does it happen only on Travis? How do we know that the tests are really fixed? Do we have an example failure log?

wthorp:
It still happens on master if you have your local timezone set to a non-GMT zone.

Test:       	TestGraphqlQuery/Postgres/MyProjects_query
Error:      	Not equal:
expected: time.Time{wall:0x2af1f908, ext:63685050490, loc:(*time.Location)(0xc0003fb2c0)}
actual  : time.Time{wall:0x2af1f908, ext:63685050490, loc:(*time.Location)(0x4f0a0a0)}

kaloyan-raev:
But these two aren't really equal:
```
expected: time.Time{wall:0x2af1f908, ext:63685050490, loc:(*time.Location)(0xc0003fb2c0)}
actual : time.Time{wall:0x2af1f908, ext:63685050490, loc:(*time.Location)(0x4f0a0a0)}
```
It's the same wall time, but in different time zones. So there are different by a few hours. Unless I am reading it the wrong way.

wthorp:
https://golang.org/pkg/time/#Time.Equal ignores time zones, direct equality does not.

PR #1201, opened 2019-01-31
Updates transport dialopts to use WithBlock()

dylanlott:
Transport wasn't throwing an error when a node couldn't be connected to. WithBlock() forces the grpc connection to wait, allowing it to return an error so that transport alertFail() could be called correctly.


pkg/transport/transport.go
```diff
@@ -68,7 +68,7 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 		return nil, Error.Wrap(err)
 	}
 
-	options := append([]grpc.DialOption{dialOpt}, opts...)
+	options := append([]grpc.DialOption{dialOpt, grpc.WithBlock()}, opts...)
```
egonelbre: Dial is called below as well. Also there were places where grpc.WithBlock is used explicitly, so those can be removed.

dylanlott: Removed other WithBlock() instances in the codebase, and updated the DialAddress function to use the WithBlock() opts as well.

PR #1203, opened 2019-02-01
Cleanup repository a bit

egonelbre:
1. Move tools from `cmd` to `tools`
2. Remove some old code
3. Move code to examples


cmd/pdbinspect/main.go
```diff
@@ -1,211 +0,0 @@
-// Copyright (C) 2019 Storj Labs, Inc.
```
mobyvb: Why is pdbinspect being removed? Is its functionality somewhere else?

egonelbre: This seems outdated and will get outdated with more time. Better not to worry about this when refactoring.

pkg/process/metrics.go
```diff
@@ -28,8 +28,7 @@ var (
 
 // InitMetrics initializes telemetry reporting. Makes a telemetry.Client and calls
 // its Run() method in a goroutine.
-func InitMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (
-	err error) {
+func InitMetrics(ctx context.Context, r *monkit.Registry, instanceID string) (err error) {
```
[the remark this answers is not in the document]
egonelbre: I just made it single line no other changes.

egonelbre:
@jtolds I'll revert metric-receiver and grpc-debug.

PR #1204, opened 2019-02-01
Enable ipv6 tests

egonelbre:


cmd/storj-sim/network.go
```diff
@@ -301,3 +317,15 @@ func identitySetup(network *Processes) (*Processes, error) {
 
 	return processes, nil
 }
+
+func vipReadString(into *string, dir, flagName string) error {
```
wthorp: vipReadString isn't a very clear name.  Not that viperReadString is much better.

egonelbre: It's because we need the setting that is store in the config file for a particular entry.

For example let's say person runs `storj-sim network setup` then changes the gateway address. Now `storj-sim network test` needs to expose that server address in `GATEWAY_0_ADDR`... so this means we need to read that from the config file with that hack.

wthorp: I'm still missing something.  Doesn't storj-sim pick up on the config file changes @
https://github.com/storj/storj/blob/331f0cc8457d1156199653b58761330c7eaab7fa/cmd/storj-sim/process.go#L42 ??

why can't we just...
storj-sim --host "::1" network setup
# overwrite gateway server.address
storj-sim network test

egonelbre: Because we need to expose the gateway/0 `server.address` as an environment variable `GATEWAY_0_ADDR` to the [test script](https://github.com/storj/storj/blob/master/scripts/test-sim-aws.sh#L39), which is invoked with `storj-sim network test bash ./test-sim-aws.sh`.

wthorp: ooooooooh.  thanks for being patient

egonelbre: also name fixed

cmd/storj-sim/network.go
```diff
@@ -165,10 +168,15 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
+
+		process.ExecBefore["run"] = func(process *Process) error {
+			return readConfigString(&process.Address, process.Directory, "server.address")
+		}
 	}
 
 	// Create gateways for each satellite
-	for i, satellite := range satellites {
+	for i := range satellites {
+		satellite := satellites[i]
```
kaloyan-raev: A habit to change :)

PR #1208, opened 2019-02-01
Fix EOF error on audits (storage node)

kaloyan-raev:
`io.CopyN` may return `io.EOF` even after reading all necessary data. The reason is that `io.CopyN` was told to copy more bytes than available.

This PR also improves the debug logs to show the namespaced piece id, so it is easier to diagnose problems on the storage node.


pkg/piecestore/psserver/retrieve.go
```diff
@@ -168,18 +168,18 @@ func (s *Server) retrieveData(ctx context.Context, stream pb.PieceStoreRoutes_Re
 
 		used += nextMessageSize
 		n, err := io.CopyN(writer, storeFile, nextMessageSize)
+		if err != nil && err != io.EOF {
```
egonelbre: Doesn't this now end up in infinite loop when file returns io.EOF and we haven't reached the length?

egonelbre: I think on error we should still increase allocationTracking. I suspect it will cause problems otherwise.

kaloyan-raev: Hmm... I don't know.

I also read in the `io.CopyN` that 

> On return, written == n if and only if err == nil

So, it's seems I am fixing the wrong thing.

aleitner: Move the EOF check here to  where we check the err from `s.retrieveData`

kaloyan-raev:
The latest commit takes another approach. Instead of checking for EOF, io.Copy is invoked with the correct number of bytes to copy. This way EOF won't be returned if everything necessary is copied.

pkg/piecestore/psserver/retrieve.go
kaloyan-raev: I pushed another commit that makes sure we pass the correct number of bytes to CopyN. This way the EOF check is not necessary.

kaloyan-raev:
@aleitner @coyle @egonelbre Please review again

PR #1210, opened 2019-02-01
disable account activation

rikysya:
Disable account activation until email service is ready


rikysya:
Already implemented in https://github.com/storj/storj/pull/1143

PR #1215, opened 2019-02-01
identity improvements:

bryanchriswhite:
+ identity refactor
+ add `identity revocations`
+ add `identity certificate-authority revoke-peer`

```
$ identity --help
Identity management

Usage:
  /home/user/go/bin/identity [command]

Available Commands:
  revocations           Print revocation information from a revocation database
...
```

`revocations`
---
```
$ identity revocations --help
Print revocation information from a revocation database

Usage:
  /home/user/go/bin/identity revocations [service] [flags]

Flags:
  -h, --help                      help for revocations
      --revocation-dburl string   url for revocation database (e.g. bolt://some.db OR redis://127.0.0.1:6378?db=2&password=abc123) (default "bolt:///home/user/.local/share/storj/revocations.db")

Global Flags:
      --config-dir string           service config directory (default "/home/user/.local/share/storj")

```

`certificate-authority revoke-peer`
---
```
$ identity certificate-authority revoke-peer --help
Revoke a peer identity's CA certificate and add to local revocation database

Usage:
  /home/user/go/bin/identity certificate-authority revoke-peer [service] [revoked cert path] [flags]

Flags:
      --ca.cert-path string        path to the certificate chain for this identity (default "/home/user/.local/share/storj/identity/ca.cert")
      --ca.key-path string         path to the private key for this identity (default "/home/user/.local/share/storj/identity/ca.key")
  -h, --help                       help for revoke-peer
      --peer-ca.cert-path string   path to the certificate chain for this identity (default "/home/user/.local/share/storj/identity/ca.cert")
      --revocation-dburl string

Global Flags:
      --config-dir string           service config directory (default "/home/user/.local/share/storj")
```


cmd/certificates/auth.go
```diff
@@ -49,6 +50,23 @@ var (
 )
 
 func init() {
+	// NB: init functions are executed in lexicographical order of filename
+	confDirParam := cfgstruct.FindConfigDirParam()
```
egonelbre: Can we do all of the setup in main instead, that way it's clearer for order.

bryanchriswhite: this stuff has to be here otherwise it won't be defined in time for other `init` functions

egonelbre: I mean if we remove all init functions and do everything in `func main()` then there shouldn't be problems.

bryanchriswhite: ah - misunderstood... sure!

cmd/certificates/sign.go
```diff
@@ -5,59 +5,57 @@ package main
...
-	err = signCfg.CA.SaveBackup(ca)
+	err = signCfg.SigneeCACfg.SaveBackup(ca)
 	if err != nil {
 		return err
 	}
-	err = signCfg.Identity.SaveBackup(ident)
+	err = signCfg.SigneeIdentCfg.SaveBackup(ident)
 	if err != nil {
-		return err
+		log.Printf("unable to save backup of identity: %s\n", err.Error())
```
mobyvb: why is the error on line 54 returned but the error here is just logged out? Don't we want the same behavior for each `SaveBackup` error?

cmd/certificates/sign.go
```diff
@@ -5,59 +5,57 @@ package main
...
+	ca, err := signCfg.SigneeCACfg.Load()
 	if err != nil {
 		return err
 	}
 
-	ident, err := signCfg.Identity.Load()
+	ident, err := signCfg.SigneeIdentCfg.Load()
 	if err != nil {
-		return err
+		log.Printf("unable to load identity: %s\n", err.Error())
```
mobyvb: similar question - why are we logging the error here and not on line 36?

cmd/certificates/sign.go
bryanchriswhite: the thought was to allow for the signing of a ca without an identity having to be present. You could always generated an identity after signing.the CA.

cmd/certificates/sign.go
bryanchriswhite: (same)

cmd/certificates/sign.go
```diff
@@ -5,59 +5,57 @@ package main
...
 	}
 
 	signer, err := signCfg.Signer.Load()
 	if err != nil {
+		fmt.Println("two")
```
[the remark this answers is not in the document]
bryanchriswhite: :+1:

pkg/identity/extensions_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/internal/testpeertls"
+	"storj.io/storj/pkg/peertls"
+)
+
+type extensionHandlerMock struct {
+	mock.Mock
```
egonelbre: try using a field count and return nil instead of using `mock.Mock`, the code should end up nicer

cmd/certificates/sign.go
```diff
@@ -5,43 +5,39 @@ package main
 
 import (
 	"crypto/x509"
+	"log"
```
egonelbre: Use zap.Logger instead, mixing multiple logging approaches usually ends up looking bad on the console.

cmd/certificates/sign.go
egonelbre: If we don't have a SigneeIdentCfg specified can we avoid logging at all? The message might look confusing.

pkg/certificates/config.go
```diff
@@ -85,7 +85,7 @@ func (c CertServerConfig) NewAuthDB() (*AuthorizationDB, error) {
 }
 
 // Run implements the responsibility interface, starting a certificate signing server.
-func (c CertServerConfig) Run(ctx context.Context, server *server.Server) (err error) {
+func (c CertServerConfig) Run(ctx context.Context, srv *server.Server) (err error) {
```
egonelbre: Let's avoid cleanups in other packages that you are not directly touching.

pkg/identity/extensions_test.go
```diff
@@ -0,0 +1,261 @@
...
+			revokedLeafChain,
+			nil,
+			nil,
+			nil,
+		},
+		{
+			"certificate revocation - serial revocations",
+			peertls.TLSExtConfig{Revocation: true},
+			1,
+			func() []*x509.Certificate {
```
egonelbre: Using func closures inside tests usually indicates that it should be a separate test. These make hard to follow what's actually happening in the test and what's being tested and in which order.

egonelbre: Also asserts here will end up in the wrong subtest.

egonelbre: https://github.com/storj/storj/wiki/Testing#table-tests

pkg/identity/extensions_test.go
```diff
@@ -0,0 +1,261 @@
...
+	whitelistSignedKeys, whitelistSignedChain, err := testpeertls.NewCertChain(3)
+	assert.NoError(t, err)
+
+	err = peertls.AddSignedCertExt(whitelistSignedKeys[0], whitelistSignedChain[0])
+	assert.NoError(t, err)
+
+	_, unrelatedChain, err := testpeertls.NewCertChain(1)
+	assert.NoError(t, err)
+
+	tmp, err := ioutil.TempDir("", "TestParseExtensions")
```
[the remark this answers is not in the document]
egonelbre: Ah, didn't notice that this was a move. If it's a move then we can merge this one and do the fixes in a separate PR. That way changes are easier to see.

pkg/utils/utils.go
```diff
@@ -67,28 +69,18 @@ func (errs combinedError) Error() string {
 }
 
 // ErrorGroup contains a set of non-nil errors
-type ErrorGroup []error
+type ErrorGroup errs.Group
```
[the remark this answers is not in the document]
bryanchriswhite: Makes sense.. I'll avoid this in the future. The reason I did this was because I encountered some old code that was using `ErrGroup` and thought it would be nice to make it consistent everywhere that was still using it.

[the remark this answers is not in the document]
egonelbre: I agree, but the fix for that would be to start replacing `utils.ErrorGroup` package by package and finally remove `utils.ErrorGroup` as the last step.

cmd/certificates/sign.go
bryanchriswhite: :+1:

pkg/utils/utils.go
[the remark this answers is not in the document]
bryanchriswhite: yea, I sort of saw this as an intermediary step to keep the implementations the same.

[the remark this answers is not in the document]
bryanchriswhite: should I roll it back or is it fine to keep it?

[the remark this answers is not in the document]
egonelbre: If it doesn't cause test failures, then it's fine.

pkg/certificates/config.go
bryanchriswhite: going forward I'll keep an eye out to avoid that :+1:

pkg/utils/utils.go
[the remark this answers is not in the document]
bryanchriswhite: :+1:

PR #1216, opened 2019-02-01
add certificates claims+

bryanchriswhite:
Usage
---
```
$ certificates claims --help
Print claim information

Usage:
  /home/user/go/bin/certificates claims [flags]

Flags:
      --authorization-dburl string   url to the certificate signing authorization database (default "bolt:///home/user/.local/share/storj/cert-signing/authorizations.db")
      --raw                          if true, the raw data structures will be printed
...

Global Flags:
      --config-dir string           main directory for certificates configuration (default "/home/user/.local/share/storj/cert-signing")
...
```

Example
---
```
identity create test --difficulty 0
identity create certificates --difficulty 0
certificates setup --server.address 127.0.0.1:8888
certificates auth create 1 user@example.com
token=$(certificates auth export --all | cut -d"," -f2)
certificates run &
certs_id=$?
identity authorize test $token

certificates claim # <----
# [
#         {  
#                 "UserID": "user@example.com",
#                 "Token": "user@example.com:1JCzrVAv9MdV26ob7JLewLLXm3fYJ5grniN57bVtZYkxgMY4vsXrrYXG86MXPdtVeTGS1DfKLuPsuNBzhgQgqAFtAG4W1m",
#                 "Claim": {
#                         "Addr": "127.0.0.1:42640",
#                         "Timestamp": "2019-02-01 22:05:26 +0100 CET",
#                         "NodeID": "1263QtYjg2njUT7eEzei5snxm4buhwAUinDijRT6vCLGgJeSZAp"
#                 }          
#         }          
# ]

kill $certs_id
```


cmd/certificates/claims.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type printableAuth struct {
+	UserID string
+	Token  string
+	Claim  *pClaim
+}
+type pClaim struct {
+	Addr      string
+	Timestamp string
```
bryanchriswhite: Should just say `Time`

cmd/certificates/claims.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func init() {
+	rootCmd.AddCommand(claimsCmd)
+
+	cfgstruct.Bind(claimsCmd.Flags(), &claimsCfg, cfgstruct.ConfDir(defaultConfDir))
+}
+
+func cmdClaims(cmd *cobra.Command, args []string) error {
+	authDB, err := claimsCfg.NewAuthDB()
```
bryanchriswhite: Close `authDB`

cmd/certificates/claims.go
bryanchriswhite: Yes :p

pkg/certificates/certificates.go
```diff
@@ -305,7 +305,24 @@ func (a *AuthorizationDB) UserIDs() ([]string, error) {
 	return keys.Strings(), nil
 }
 
-// Claim marks an authorization as claimed and records claim information
+// List returns all authorizations in the database.
+func (a *AuthorizationDB) List() (auths Authorizations, _ error) {
```
navillasa: why the blank identifier for this error? just curious

cmd/certificates/claims.go
```diff
@@ -0,0 +1,102 @@
...
+	cfgstruct.Bind(claimsCmd.Flags(), &claimsCfg, cfgstruct.ConfDir(defaultConfDir))
+}
+
+func cmdClaims(cmd *cobra.Command, args []string) (err error) {
+	authDB, err := claimsCfg.NewAuthDB()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = authDB.Close()
```
egonelbre: Use `err = errs.Combine(err, authDB.Close()`, otherwise it will override the previous error return values.

cmd/certificates/claims.go
```diff
@@ -0,0 +1,102 @@
...
+	Token  string
+	Claim  *pClaim
+}
+type pClaim struct {
+	Addr   string
+	Time   string
+	NodeID string
+}
+
+func toPrintableAuth(a *certificates.Authorization) *printableAuth {
```
egonelbre: maybe instead of `a` `auth` would be better

cmd/certificates/claims.go
```diff
@@ -0,0 +1,102 @@
...
+	fmt.Println(string(jsonBytes))
+	return err
+}
+
+type printableAuth struct {
+	UserID string
+	Token  string
+	Claim  *pClaim
+}
+type pClaim struct {
```
egonelbre: maybe instead of `pClaim` use `printableClaim`

pkg/certificates/certificates.go
egonelbre: It's hard to follow code with short codes when jumping between packages https://github.com/storj/storj/wiki/Style#variable-naming. Maybe `a` -> `authdb`.

bryanchriswhite: @navillasa I wanted `auths` to be named so that I don't have to declare it in the method body but it's all or none with named return vars and since `err` is always to the left of `:=`, naming it would allocate an unused variable I think.

cmd/certificates/claims.go
bryanchriswhite: :+1:

cmd/certificates/claims.go
bryanchriswhite: :+1:

cmd/certificates/claims.go
bryanchriswhite: I think I planned on this but got sidetracked and never came back to it. Thanks for the reminder :p

pkg/certificates/certificates.go
bryanchriswhite: :+1:

PR #1218, opened 2019-02-03
don't put nil nodes in GetResponse

thepaul:
because it causes GRPC errors while marshaling (see Alexey's tracebacks
in rocketchat #dev from a couple hours before this commit).

i'm not at all sure this is the right fix, because it looks like kaloyan
_wanted_ to allow nil nodes specifically in a90aa1858, but maybe this is
a step.


egonelbre:
Let's add a test for this scenario, we have had quite a few failures due to missing nodes.

thepaul:
Nice. Thanks kaloyan

kaloyan-raev:
@egonelbre Added the test case.

PR #1219, opened 2019-02-03
better batch-generation

bryanchriswhite:
+ Generate complete identities instead of just keys
+ Optionally output as tarballs
+ Control permissions of output files/dirs

Usage
---
```
# directory output
identity batch-generate --output-dir ./v3-identities --min-difficulty 30

# tar output
identity batch-generate --output-dir ./v3-identities --min-difficulty 30 --tar
```
### Directory output
```
$ tree ./v3-identites
./v3-identities
├── gen-30-1
│   └── 1GzGMx3wy9
│       ├── ca.cert
│       ├── ca.key
│       ├── identity.cert
│       └── identity.key
└── gen-31-2
    └── 1bhMnXovR6
        ├── ca.cert
        ├── ca.key
        ├── identity.cert
        └── identity.key
```
### Tar output
```
$ tree ./v3-identites
./v3-identites
├── gen-30-1.tar
├── gen-30-3.tar
├── gen-31-2.tar
└── gen-32-4.tar

$ tar -tf ./v3-identities/gen-30-1.tar 
1k99cCKJoY/ca.cert
1k99cCKJoY/identity.cert
1k99cCKJoY/ca.key
1k99cCKJoY/identity.key
```


cmd/identity/keys.go
```diff
@@ -49,17 +68,142 @@ func cmdKeyGenerate(cmd *cobra.Command, args []string) (err error) {
...
+				tarPath: filepath.Join(keyCfg.OutputDir, genName),
+				key:     k,
+				id:      id,
+
+				tarData: new(bytes.Buffer),
+
+				errs: new(errs.Group),
+			}
+
+			go dID.build().writeTar(keyCfg.OutputDir).done()
```
bryanchriswhite: Not sure if this is a good idea... it's late and I was (am) tired.

Q's:
1. Separate go routine?
1. Chaining shenanigans?
1. Error shenanigans (and checking the error)?

bryanchriswhite:
@stefanbenten good call on the ux. We should indeed print some output like we do in other identity generation commands.

cmd/identity/keys.go
```diff
@@ -4,22 +4,41 @@
 package main
 
 import (
+	"archive/tar"
```
egonelbre: Any reason you aren't just writing them to a folder instead of tarring?

cmd/identity/keys.go
egonelbre: Yeah, no need for the goroutine.

cmd/identity/keys.go
egonelbre: Anyone using it, can later compact them as needed -- also that would make the code much simpler as well.

bryanchriswhite: It would definitely make the code simpler; honestly, I hadn't considered it. I think it's cleaner to have one file per identity though and makes it less likely that you make a mistake when transporting or using them.

egonelbre: I think anyone using batch-generate should have sufficient knowledge to understand that the whole directory is important.

cmd/identity/batch.go
```diff
@@ -51,6 +44,8 @@ var (
 		MinDifficulty int    `help:"minimum difficulty to output" default:"30"`
 		Concurrency   int    `help:"worker concurrency" default:"4"`
 		OutputDir     string `help:"output directory to place keys" default:"."`
+		Permissions   uint   `default:"0700" help:"permissions to apply to output directories and/or files"`
+		Tar           bool   ` default:"false" help:"f true, identities will be output as tarballs instead of directories"`
```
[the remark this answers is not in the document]
bryanchriswhite: I take your point; however:

1. It's done
1. I would argue that there is value in having the option to have tarball output (without having to script it additionally)
1. The difference as far as mainenance goes keeping this:
```
func (d *deferredIdentity) writeDir(idPathPart string) error {
	idDir := filepath.Join(d.path, idPathPart)
	if err := os.MkdirAll(idDir, os.FileMode(keyCfg.Permissions)); err != nil {
		return errs.Wrap(err)
	}

	for name, data := range d.files {
		if err := ioutil.WriteFile(filepath.Join(idDir, name), data, os.FileMode(keyCfg.Permissions)); err != nil {
			return errs.Wrap(err)
		}
	}
	return nil
}
```
in sync with this:
```
func (d *deferredIdentity) writeTar(idPathPart string) error {
	tarData := new(bytes.Buffer)
	tw := tar.NewWriter(tarData)

	for name, data := range d.files {
		err := writeToTar(tw, filepath.Join(idPathPart, name), data)
		if err != nil {
			return errs.Wrap(err)
		}
	}

	if err := tw.Close(); err != nil {
		return errs.Wrap(err)
	}

	err := ioutil.WriteFile(d.path+".tar", tarData.Bytes(), os.FileMode(keyCfg.Permissions))
	if err != nil {
		return errs.Wrap(err)
	}
	return nil
}
```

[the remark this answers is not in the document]
egonelbre: For comparison this is what it could look like without tar (ignore typos): https://gist.github.com/egonelbre/0c5662cd12c87e908761cd6364b06a59... roughly 2x less code, and clearer in what's happening.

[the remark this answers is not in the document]
stefanbenten: Creating the tar is in my opinion the smartest way when generating a bunch of identities.
Makes moving them an easier approach and it makes sure, that the necessary files stick together.
I could also think of using the tar as input for our identities in general.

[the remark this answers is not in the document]
egonelbre: Seeing from the current ed25519 changes, it's very likely to change. Although, not much.

"it exists" is not a good argument for keeping something, especially when usability is also in question. I do understand it's not a lot of code to maintain, but I would like to ensure that people don't add features because "we can", but rather that it solves a problem.

There are few follow-up questions about `tar` like, how do we handle support on Windows, since it doesn't support it by default? The permissions for the files aren't the same as `.Save` func-s set, why the differences?

i.e. I would like to avoid these things by doing the simple thing. Create the subfolder and save as usual.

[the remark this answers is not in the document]
egonelbre: I don't agree, but let's do only `tar` then, but also adjust the code to add default permissions that we elsewhere.

cmd/identity/batch.go
```diff
@@ -0,0 +1,206 @@
...
+	if err := peertls.WriteChain(identBytes, ident.Leaf, ident.CA); err != nil {
+		return errs.Wrap(err)
+	}
+
+	for name, chain := range map[string][]*x509.Certificate{
+		"ca.cert":       {ca.Cert},
+		"identity.cert": {ident.Leaf, ident.CA},
+	} {
+		chainBuf := new(bytes.Buffer)
+		if err := peertls.WriteChain(chainBuf, chain...); err != nil {
```
egonelbre:
```
err := errs.Combine(
  writeToTar(tw, "ca.cert", 0600, peertls.ChainBytes(ca.Cert)),
  writeToTar(tw, "identity.cert", 0600, peertls.ChainBytes(ident.Leaf, ident.CA)),
...
)
```

egonelbre: (same with `peertls.KeyBytes`)

cmd/identity/batch.go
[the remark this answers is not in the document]
coyle: My two cents here. I completely get the concern to leave our code base as simple as possible and eliminate extraneous features. I do think this feature is a nice user experience though. We could have another script to do this but that is more code and it's in another spot. I prefer having the code here and making it a simple flag for the user.

bryanchriswhite:
> Do we have tests for these ?

no

cmd/identity/batch.go
```diff
@@ -0,0 +1,280 @@
...
+	return nil
+}
+
+func updateStatus(counts map[uint16]*uint32, textWidget *text.Text) error {
+	textBuf := new(bytes.Buffer)
+	w := tabwriter.NewWriter(textBuf, 0, 2, 2, ' ', 0)
+	if _, err := fmt.Fprintf(w, "Difficulty\tCount\n"); err != nil {
+		return err
+	}
+	for diff, count := range counts {
```
bryanchriswhite: Sort by descending difficulty

cmd/identity/batch.go
```diff
@@ -0,0 +1,280 @@
...
+		container.Border(draw.LineStyleLight),
+		container.BorderTitle(" batch identity generation | press \"q\" to quit "),
+		container.PlaceWidget(textWidget),
+	)
+
+	//ctrl, err := termdash.NewController(t, c)
+	//if err != nil {
+	//	return err
+	//}
+	go func() {
```
bryanchriswhite: Add `--no-ui` or something for scripting compatibility?

cmd/identity/batch.go
```diff
@@ -0,0 +1,206 @@
...
+				id:   id,
+			}
+
+			if err := dID.build(); err != nil {
+				return false, err
+			}
+
+			idPathPart := id.String()[:10]
+			if keyCfg.Tar {
+				if err := dID.writeTar(idPathPart); err != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: The path above is the path to the tar or top-level directory, this path is the directory inside the tar or a sub-directory

cmd/identity/batch.go
```diff
@@ -0,0 +1,280 @@
...
+	"crypto/ecdsa"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync/atomic"
+	"text/tabwriter"
+
+	"github.com/mum4k/termdash"
```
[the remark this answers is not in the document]
bryanchriswhite: Again, I disagree.

Firstly, it's not. "a lot of code"; presently, being generous, it's 30 lines of termdash-specific code.

I think it's totally valid to want feedback on the progress of the command for the same reasons we added feedback to the other identity commands. Instead of doing something similar to what was done in [dashboard.go](https://github.com/storj/storj/blob/master/cmd/storagenode/dashboard.go), which is at least the same amount of code for just the setup. I think it makes sense to use something that does all the platform specific stuff for us; Additionally, termdash is testing and maintaining that code so we don't have to.

Based on conversations I've had with multiple people, we're probably going to want to do wizards for configuration/setup and/or make our dashboard(s) interactive in the future. This seemed like the perfect opportunity to get familiar with a library like termdash which can facilitate both those things with the additional benefits that I already mentioned. 

I'm not married to termdash in particular, there are other options, some even using the same Cross-Platform core ([termbox-go](https://github.com/nsf/termbox-go/blob/master/README.md)). I just chose termdash because it seemed like a good place to start.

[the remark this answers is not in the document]
egonelbre: termdash is 28000LOC of code, termbox is +4000LOC.

cmd/identity/batch.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/mum4k/termdash/container"
+	"github.com/mum4k/termdash/draw"
+	"github.com/mum4k/termdash/terminal/termbox"
+	"github.com/mum4k/termdash/terminalapi"
+	"github.com/mum4k/termdash/widgets/text"
+	"github.com/prometheus/common/log"
```
egonelbre: Was this an autocomplete thing?

bryanchriswhite: Yes, will fix

cmd/identity/batch.go
```diff
@@ -0,0 +1,200 @@
...
+	if err = ioutil.WriteFile(path+".tar", tarData.Bytes(), 0600); err != nil {
+		return errs.Wrap(err)
+	}
+	return nil
+}
+
+func writeToTar(tw *tar.Writer, name string, data []byte) error {
+	err := tw.WriteHeader(&tar.Header{
+		Name: name,
+		Mode: 0600,
```
egonelbre: I think certs are supposed to be 0644 based on  https://github.com/storj/storj/blob/master/pkg/identity/identity_test.go#L126

PR #1220, opened 2019-02-04
Stop storagenode on ctrl+c

mniewrzal:


cmd/storagenode/main.go
```diff
@@ -169,8 +171,15 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
...
+	var closeError error
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-c
+		closeError = peer.Close()
+	}()
+
 	runError := peer.Run(ctx)
-	closeError := peer.Close()
```
egonelbre: This fails to close peer when there's no signal.

mniewrzal: ok, maybe one of the pieces is not canceling itself. I will take a look.

pkg/server/server.go
```diff
@@ -82,5 +82,13 @@ func (p *Server) Run(ctx context.Context) (err error) {
 		return next.Run(ctx, p)
 	}
 
+	go func() {
+		select {
+		case <-ctx.Done():
+			// error is always nil
+			_ = p.Close()
```
egonelbre: Use errgroup, it's not necessarily nil in the future:

```
ctx, cancel := context.WithCancel(ctx)
defer cancel()

var group errgroup.Group
group.Go(func() error {
    <-ctx.Done()
    return p.Close()
})
group.Go(func() error {
    err := p.grpc.Serve(p.lis)
    cancel()
    return err
})

return group.Wait()
```

scripts/check-travis-tidy.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 set -ueo pipefail
```
egonelbre: this needs to be removed as well

PR #1221, opened 2019-02-04
testplanet: add capability to stop peers

egonelbre:
This is necessary for testing storage nodes being offline.

### DIFF
--- a/internal/testplanet/planet_test.go
+++ b/internal/testplanet/planet_test.go
@@ -7,27 +7,23 @@ import (
 	"context"
 	"strconv"
 	"testing"
+	"time"
 
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testcontext"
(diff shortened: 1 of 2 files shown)


internal/testplanet/planet_test.go
```diff
@@ -41,29 +37,26 @@ func TestBasic(t *testing.T) {
...
 	_, err = nodeClient.Ping(context.Background(), planet.StorageNodes[1].Local())
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
+	err = planet.StopPeer(planet.StorageNodes[0])
+	require.NoError(t, err)
+
+	time.Sleep(time.Second)
```
mniewrzal: Why `Sleep` at the end?

egonelbre: To wait for some potential failures.

PR #1224, opened 2019-02-04
Rollup tests

cam-a:
For generating payment reports, we query the accounting rollup table. We want to make sure that this service functions as expected. Once we refactor QueryPaymentInfo to sum data totals across rollups, we'll need to change what we expect the length of rows to be on line 89

### DIFF
--- a/pkg/accounting/rollup/rollup_test.go
+++ b/pkg/accounting/rollup/rollup_test.go
@@ -105,7 +105,7 @@ func TestQuery(t *testing.T) {
 	}
 }
 
-// contains
+// contains checks for and removes items if found
 func contains(entry storj.NodeID, list []*storj.NodeID) bool {
 	for i, n := range list {
 		if n == nil {
--- a/pkg/accounting/rollup/rollup_test.go
+++ b/pkg/accounting/rollup/rollup_test.go
@@ -34,10 +34,10 @@ func TestQuery(t *testing.T) {
 			nodes:  10,
 		},
 		{
-			days:   3,
-			atRest: float64(5000),
-			bw:     []int64{1000, 2000, 3000, 4000},
-			nodes:  10,
+			days:   5,
+			atRest: float64(20000),
+			bw:     []int64{4000, 8000, 12000, 16000},
+			nodes:  20,
 		},
 	}
 
--- a/pkg/accounting/rollup/rollup_test.go
+++ b/pkg/accounting/rollup/rollup_test.go
@@ -15,87 +15,145 @@ import (
 )
 
 func TestQueryOneDay(t *testing.T) {
-	ctx := testcontext.New(t)
-	defer ctx.Cleanup()
+	tests := []struct {
+		expected int
+		rest     float64
+		bw       []int64
+	}{
+		{
+			expected: 0,
+			rest:     float64(5000),
+			bw:       []int64{1000, 2000, 3000, 4000},
+		},
+	}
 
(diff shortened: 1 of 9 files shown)


pkg/accounting/rollup/rollup_test.go
```diff
@@ -8,79 +8,124 @@ import (
...
 
-func TestQueryOneDay(t *testing.T) {
-	// TODO: use testplanet
-
-	ctx, r, db, nodeData, cleanup := createRollup(t)
-	defer cleanup()
+func TestQuery(t *testing.T) {
+	tests := []struct {
+		days  int
+		rest  float64
```
jenlij: maybe rename to atRest, as i read "rest" as a sleep or something

pkg/accounting/rollup/rollup_test.go
```diff
@@ -8,79 +8,105 @@ import (
...
 
-	now := time.Now().UTC()
-	later := now.Add(time.Hour * 24)
+	for _, tt := range tests {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
 
-	err := db.Accounting().SaveAtRestRaw(ctx, now, nodeData)
-	assert.NoError(t, err)
+		planet, err := testplanet.New(t, 1, tt.nodes, 0)
```
jenlij: not sure if it will be a problem to create new test planets each loop

pkg/accounting/rollup/rollup_test.go
```diff
@@ -8,79 +8,105 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"go.uber.org/zap"
 
 	"storj.io/storj/internal/testcontext"
-	"storj.io/storj/internal/teststorj"
-	"storj.io/storj/pkg/accounting/rollup"
+	"storj.io/storj/internal/testplanet"
```
wthorp: nice.

pkg/accounting/rollup/rollup_test.go
```diff
@@ -8,79 +8,105 @@ import (
...
-	err := db.Accounting().SaveAtRestRaw(ctx, now, nodeData)
-	assert.NoError(t, err)
+		planet, err := testplanet.New(t, 1, tt.nodes, 0)
+		assert.NoError(t, err)
+		defer ctx.Check(planet.Shutdown)
 
-	// test should return error because we delete latest day's rollup
-	err = r.Query(ctx)
-	assert.NoError(t, err)
+		planet.Start(ctx)
```
egonelbre: Btw. we now also have `testplanet.Run` which tests against both sqlite and postgres.

pkg/accounting/rollup/rollup_test.go
```diff
@@ -8,79 +8,122 @@ import (
...
 
-	ctx, _, db, nodeData, cleanup := createRollup(t)
-	defer cleanup()
+		saveAndQuery(t, ctx, flags, planet)
+	})
+}
 
-	now := time.Now().UTC()
-	then := now.Add(time.Hour * -24)
+func TestThreeDays(t *testing.T) {
```
egonelbre: I think a single test across 3 days would be sufficient, not sure whether we gain something from having more. Alternatively it could be a single test and testplanet where in each loop iteration adds one additional day.

pkg/accounting/rollup/rollup_test.go
wthorp: this really would be nice to have

PR #1225, opened 2019-02-04
General satellite protobuf

aleitner:
Create an api endpoint on satellite for checking file health


thepaul:
Might be confusing to call this interface "Satellite", as there are several other interfaces that are also required to compose a satellite. Maybe HealthInspector, and group it with the other inspector interfaces?

dylanlott:
Yeah, I have to agree with thepaul about naming. I think something like filehealth.proto would be a better name for this since we have so many things named satellite alreayd.

aleitner:
JT and I talked and we thought it would be best to make a general satellite protobuf for a bunch of different things in the future. Not just checking file health

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,28 @@
...
+syntax = "proto3";
+option go_package = "pb";
+
+package satellite;
+
+service Satellite {
+    rpc Health(FileFealthRequest) returns (FileHealthResponse) {}
+}
+  
+message FileFealthRequest {
```
mniewrzal: Fealth - Health

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,28 @@
...
+    bytes file_id = 1;
```
mniewrzal: What do you mean by file_id? I think we are not using `file` in names around satellite.

aleitner:
dang y'all attacking a PR that hasn't even been written yet :,(

mniewrzal:
You should be happy to have feedback already :-)

satellite/peer.go
```diff
@@ -195,6 +195,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 			return nil, errs.Combine(err, peer.Close())
 		}
 
+		peer.Public.Endpoint = endpoints.NewServer(peer.Log.Named("satellite_endpoints"), config.SatelliteServer)
```
egonelbre: Use the same naming schema as above, without underscores.

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,31 @@
...
+// See LICENSE for copying information.
+
+syntax = "proto3";
+option go_package = "pb";
+
+package satellite;
+
+import "gogo.proto";
+
+service Satellite {
```
egonelbre: Satellite is a composite service, so calling it `Satellite`  makes it confusing. Adding a general service makes it a gravity center, which can easily become a kitchen-sink - you can see the similar effect, because we now also needed `endpoints` package.

Also, I don't think a single service is necessary since protobuf should be able to connect to a composite of interfaces... not sure whether there's a capability out-of-the-box, but based on the RPC, it seems plausible.

The file health is about talking to pointerdb / metainfo. So it should probably be in one of those. Let's call this endpoint `Metainfo`, because we we'll need one soon anyways and seems more suitable.

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,31 @@
...
+    rpc Health(FileHealthRequest) returns (FileHealthResponse) {}
+}
+  
+message FileHealthRequest {
```
egonelbre: We do not manage files, but instead `Objects`, so `ObjectHealthRequest`.

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,31 @@
...
+    message SegmentInfo{
+        int64 good_nodes = 1; // How many pieces are currently on good nodes
+        int64 bad_nodes = 2; // How many pieces are currently on bad nodes
+        int64 offline_nodes = 3; // How many pieces are currently on offline nodes
+        int64 below_recover = 4; // How many pieces are needed to recover
+        int64 below_repair = 5; // How many pieces are needed for being above the repair threshold
+        int64 below_success = 6; // How many pieces are needed to be at the success threshold
+    }
+
+    repeated SegmentInfo segments = 1;
```
egonelbre: Do we care about per-segment info? I think total over the full file seems sufficient.

pkg/pb/satellite.proto
```diff
@@ -0,0 +1,31 @@
...
+package satellite;
+
+import "gogo.proto";
+
+service Satellite {
+    rpc Health(FileHealthRequest) returns (FileHealthResponse) {}
+}
+  
+message FileHealthRequest {
+    string path = 1;
```
egonelbre: This needs to be bytes. Paths may contain invalid utf-8.

egonelbre: Also separate the bucket and path, preemptively, since we'll do that split everywhere.

satellite/client/client.go
```diff
@@ -0,0 +1,63 @@
...
+
+// SatelliteClient maintains variables required for talking to basic satellite endpoints
+type SatelliteClient struct {
+	log       *zap.Logger
+	transport transport.Client
+	kad       *kademlia.Kademlia
+}
+
+// New creates an Satellite Client
+func New(log *zap.Logger, identity *identity.FullIdentity, kad *kademlia.Kademlia) *SatelliteClient {
```
egonelbre: Pass in `transport.Client` not `identity.FullIdentity`, most places should use that approach. Kademlia seems unnecessary, since we communicate with the satellite address directly.

satellite/client/client.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	// Create client from satellite ip
+	conn, err := sc.transport.DialNode(ctx, &satellite)
+	if err != nil {
+		return nil, ClientError.Wrap(err)
+	}
+
+	client := pb.NewSatelliteClient(conn)
+	defer func() {
+		err := conn.Close()
```
egonelbre: Return the error don't log this.

satellite/client/client.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package client
```
egonelbre: General package names are problematic since they can easily start colliding with each other.

satellite/endpoints/config.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package endpoints
```
egonelbre: catch-all `packages endpoints` tend to become kitchen sinks. also see https://rakyll.org/style-packages/

satellite/peer.go
```diff
@@ -78,7 +79,8 @@ type Config struct {
 	Identity identity.Config
 
 	// TODO: switch to using server.Config when Identity has been removed from it
-	Server server.Config
+	Server          server.Config
+	SatelliteServer endpoints.Config
```
egonelbre: This is confusing when you have flags for satellite like "satellite --server.address xxx --satellite-server.something yyy`

pkg/pb/satellite.proto
aleitner: Changed to Metainfo

aleitner: I renamed everything called `metainfo` in `peer.go` to `pointerdb` as a result to reduce confusion

satellite/peer.go
aleitner: done

pkg/pb/satellite.proto
aleitner: done

pkg/pb/satellite.proto
aleitner: I think we can do more with per-segment info. Also if we just list everything together we won't know if we have bad segments or not

pkg/pb/satellite.proto
aleitner: done

aleitner: done

satellite/client/client.go
aleitner: done

satellite/client/client.go
aleitner: I'm not sure on how I should structure the package and name them

satellite/endpoints/config.go
aleitner: See above comment

pkg/pointerdb/metainfo/server/config.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package metainfoserver
```
egonelbre: Move this into `satellite/metainfo`

pkg/pointerdb/metainfo/server/server.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"storj.io/storj/pkg/pb"
+)
+
+var (
+	mon = monkit.Package()
+)
+
+// Server implements the network state RPC service
+type Server struct {
```
egonelbre: `Endpoint`

pkg/pointerdb/metainfo/server/server.go
```diff
@@ -0,0 +1,48 @@
...
+	logger *zap.Logger
+	config Config
+}
+
+// NewServer creates instance of Server
+func NewServer(logger *zap.Logger, config Config) *Server {
```
egonelbre: `logger` -> `log`

satellite/peer.go
```diff
@@ -134,11 +138,12 @@ type Peer struct {
 		Inspector *statdb.Inspector
 	}
 
-	Metainfo struct {
+	Pointerdb struct {
```
egonelbre: Let's not change these for now, we'll clean up these later instead.

satellite/peer.go
```diff
@@ -134,11 +138,12 @@ type Peer struct {
...
 		Database   storage.KeyValueStore // TODO: move into pointerDB
 		Allocation *pointerdb.AllocationSigner
 		Service    *pointerdb.Service
 		Endpoint   *pointerdb.Server
+		Metainfo   *metainfoserver.Server
```
egonelbre: Maybe call it `NewEndpoint`

pkg/pointerdb/metainfo/client/client.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package metainfoclient
```
egonelbre: Move this into `uplink/metainfo`

pkg/pointerdb/metainfo/client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	"storj.io/storj/pkg/transport"
+)
+
+var (
+	// ClientError wraps errors returned from client package
+	ClientError = errs.Class("metainfo client error")
+)
+
+// SatelliteClient maintains variables required for talking to basic satellite endpoints
+type SatelliteClient struct {
```
egonelbre: `Client` would be sufficient

pkg/pointerdb/metainfo/client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	satellite *pb.Node
+}
+
+// New creates an Satellite Client
+func New(log *zap.Logger, transport transport.Client, satellite *pb.Node) *SatelliteClient {
+	return &SatelliteClient{log: log, transport: transport, satellite: satellite}
+}
+
+// Stat will return the health of a specific path
+func (sc *SatelliteClient) Stat(ctx context.Context, path []byte, bucket []byte) (*pb.ObjectHealthResponse, error) {
```
egonelbre: `sc` -> `client`

pkg/pointerdb/metainfo/client/client.go
```diff
@@ -0,0 +1,55 @@
...
+// SatelliteClient maintains variables required for talking to basic satellite endpoints
+type SatelliteClient struct {
+	log       *zap.Logger
+	transport transport.Client
+	satellite *pb.Node
+}
+
+// New creates an Satellite Client
+func New(log *zap.Logger, transport transport.Client, satellite *pb.Node) *SatelliteClient {
+	return &SatelliteClient{log: log, transport: transport, satellite: satellite}
```
egonelbre: Uplink doesn't know the node info, only the address. The signature should probably be:

```
func Dial(ctx context.Context, log *zap.Logger, transport transport.Client, satelliteAddress string) *Client {
```

pkg/pointerdb/metainfo/server/config.go
aleitner: done

pkg/pointerdb/metainfo/server/server.go
aleitner: done

satellite/peer.go
aleitner: done

pkg/pointerdb/metainfo/server/server.go
aleitner: done

satellite/peer.go
aleitner: done

pkg/pointerdb/metainfo/client/client.go
aleitner: done

pkg/pointerdb/metainfo/client/client.go
aleitner: done

pkg/pointerdb/metainfo/client/client.go
aleitner: done

pkg/pointerdb/metainfo/client/client.go
aleitner: done

satellite/client/client.go
aleitner: done

cmd/uplink/cmd/stat.go
```diff
@@ -0,0 +1,63 @@
...
+
+	dst, err := fpath.New(args[0])
+	if err != nil {
+		return err
+	}
+
+	if dst.IsLocal() {
+		return fmt.Errorf("No bucket specified, use format sj://bucket/")
+	}
+
```
cam-a: Need to pass the encrypted path. Try something like this. Maybe there's a better way though?

```
metainfo, _, err := cfg.Metainfo(ctx)
bucketInfo, err := metainfo.GetBucket(ctx, obj.Bucket())

key := new(storj.Key)
copy(key[:], cfg.Enc.Key)

fullpath := obj.Bucket() + "/" + obj.Path()
encryptedPath, err := streams.EncryptAfterBucket(fullpath, bucketInfo.PathCipher, key)
```

cmd/uplink/cmd/stat.go
```diff
@@ -0,0 +1,63 @@
...
+	identity, err := cfg.Identity.Load()
+	if err != nil {
+		return err
+	}
+
+	metainfoClient := New(transport.NewClient(identity), setupCfg.SatelliteAddr)
+	
+	// Stat will return the health of a specific path
+	resp, err := metainfo.Stat(ctx, src.Bucket(), src.Path()) {
```
cam-a: dst.Bucket(), dst.Path()

PR #1226, opened 2019-02-04
storage node cert cache

wthorp:
Groundwork for storage nodes to have a cache of satellite certificates
This PR does not cover any move towards using the cert cache for validation


pkg/kademlia/dialer.go
```diff
@@ -86,6 +88,27 @@ func (dialer *Dialer) Ping(ctx context.Context, target pb.Node) (bool, error) {
 	return err == nil, errs.Combine(err, conn.disconnect())
 }
 
+// GetPeerID pings target.
+func (dialer *Dialer) GetPeerID(ctx context.Context, target pb.Node) (pID *identity.PeerIdentity, err error) {
```
egonelbre: I think this should have a name to indicate that it might connect to another system. `FetchPeerIdentity` or something? `ID` usually refers to the `NodeID` not the full/peer `Identity`

pkg/kademlia/kademlia.go
```diff
@@ -188,6 +188,19 @@ func (k *Kademlia) WaitForBootstrap() {
...
+func (k *Kademlia) GetPeerIdentity(ctx context.Context, nodeID storj.NodeID) (*identity.PeerIdentity, error) {
+	if !k.lookups.Start() {
+		return nil, context.Canceled
+	}
+	defer k.lookups.Done()
+	node, err := k.FindNode(ctx, nodeID)
+	if err != nil {
+		return nil, err
+	}
+	return k.dialer.GetPeerID(ctx, node)
```
egonelbre: Why is this doing an additional dial when it could've gotten the information from `FindNode`?

pkg/piecestore/psserver/readerwriter.go
```diff
@@ -69,7 +69,7 @@ func NewStreamReader(s *Server, stream pb.PieceStoreRoutes_StoreServer, bandwidt
 		}
 		// if whitelist does not contain PBA satellite ID, reject storage request
 		if len(s.whitelist) != 0 {
-			if !s.approved(pba.SatelliteId) {
+			if !s.isOnWhitelist(pba.SatelliteId) {
```
egonelbre: `isWhitelisted`?

pkg/kademlia/kademlia.go
wthorp: Lookup would involve changing functions in Kademlia, PeerDiscover, and Dialer.  I wasn't entirely sure if lookup connected even if the nodeid was already in the routing table.  This was way easier. :)

PR #1229, opened 2019-02-05
 Ensure checker tests run against postgres

egonelbre:


pkg/datarepair/checker/checker_test.go
```diff
@@ -19,162 +18,90 @@ import (
...
+}
 
-		injuredSegment, err := repairQueue.Dequeue(tctx)
-		assert.NoError(b, err)
+func TestOfflineNodes(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 0,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		time.Sleep(2 * time.Second)
+		time.Sleep(2 * time.Second)
```
mniewrzal: Second sleep is intentional?

egonelbre: nope

PR #1232, opened 2019-02-05
Ensure overlay tests run against postgres

egonelbre:


pkg/overlay/server_test.go
```diff
@@ -19,211 +19,202 @@ import (
...
-	{ // BulkLookup
-		result, err := server.BulkLookup(ctx, &pb.LookupRequests{
-			LookupRequest: []*pb.LookupRequest{
-				{NodeId: planet.StorageNodes[0].ID()},
-				{NodeId: planet.StorageNodes[1].ID()},
-				{NodeId: planet.StorageNodes[2].ID()},
-			},
-		})
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 8, UplinkCount: 1,
```
thepaul: This one changed from StorageNodeCount 4 to 8. Since all the others kept the same parameters, it's worth asking if this was intended?

egonelbre: Unintentional

egonelbre: fixed

PR #1233, opened 2019-02-05
Remove node package and simplify DHT interface

egonelbre:
1. Remove `node.ConnectionPool` because it was hardly used and was leaking connections. A connection pool needs a better design.
2. Simplified DHT interface to make refactoring easier.
3. Moved `node.Server` into `kademlia.Endpoint`, because that is specifically for kademlia rather than a general "node" interface.


pkg/kademlia/dialer_test.go
```diff
@@ -0,0 +1,138 @@
...
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 3,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		time.Sleep(2 * time.Second)
+		expectedKademliaEntries := 1 + len(planet.Satellites) + len(planet.StorageNodes)
+
+		// TODO: also use satellites
+		peers := planet.StorageNodes
+
+		{ // Ping
```
mniewrzal: Maybe separate methods for those 3 groups would be cleaner? Or better comment what first and second lookup is doing.

egonelbre: fixed

pkg/kademlia/dialer_test.go
```diff
@@ -0,0 +1,138 @@
...
+		{ // Ping: storage node pings all other storage nodes
+			self := planet.StorageNodes[0]
+
+			dialer := kademlia.NewDialer(zaptest.NewLogger(t), self.Transport)
+			defer ctx.Check(dialer.Close)
+
+			var group errgroup.Group
+
+			for i := range peers {
+				peer := peers[i]
```
kaloyan-raev: How about the less error-prone:
```
for _, peer := range peers {
	peer := peer
```

egonelbre: fixed

pkg/kademlia/dialer_test.go
```diff
@@ -0,0 +1,138 @@
...
+				{},    // empty target
+				{255}, // non-empty
+			}
+
+			var group errgroup.Group
+
+			for i := range targets {
+				target := targets[i]
+				for i := range peers {
+					peer := peers[i]
```
kaloyan-raev: Here too for both `peer` and `target`.

egonelbre: ah, missed

PR #1237, opened 2019-02-05
pass UTC times to db so that sqlite3 understands

thepaul:
this fixes the pkg/accounting/tally tests, at least for me.

(credit to Bill Thorp)


thepaul:
You are truly terrible for not knowing by heart all of the bugs in the go sqlite3 db driver. Let this be a lesson.

Oh ok I'll add that other one

PR #1240, opened 2019-02-05
restore previous folder layout in travis

egonelbre:


scripts/setup-gopath.sh
```diff
@@ -6,7 +6,7 @@ export PATH=~/bin:$PATH
 if [[ "${GO111MODULE}" == on ]]; then
     export GOMODULES=auto
     mkdir -p $GOPATH/src/storj.io/storj
-    cp ${TRAVIS_BUILD_DIR} $GOPATH/src/storj.io/storj
+    cp -r ${TRAVIS_BUILD_DIR} $GOPATH/src/storj.io/storj
```
thepaul: Might I suggest `-a` instead here, just to avoid some potential future issues?

PR #1242, opened 2019-02-06
SHA256Hash() and TBSBytes() needn't return error

thepaul:
according to Go docs, the error cases should never happen. change them to panics, just in case, but no need to return error and check it every single time.

NOTE: this change has two commits; the first actually corresponds to #1241. This change is meant to build on top of that one. I don't think Github provides any nice way to do that, so please bear with the awkwardness for now.

### DIFF
--- a/pkg/peertls/extensions.go
+++ b/pkg/peertls/extensions.go
@@ -247,17 +247,11 @@ func (r Revocation) Verify(signingCert *x509.Certificate) error {
 // TBSBytes (ToBeSigned) returns the hash of the revoked certificate hash and
 // the timestamp (i.e. hash(hash(cert bytes) + timestamp)).
 func (r *Revocation) TBSBytes() []byte {
-	var toHash bytes.Buffer
-	_, err := toHash.Write(r.CertHash)
-	if err != nil {
-		// bytes.Buffer#Write() is documented as never returning an error
-		panic(err)
-	}
+	var tsBytes [binary.MaxVarintLen64]byte
+	binary.PutVarint(tsBytes[:], r.Timestamp)
+	toHash := append(append([]byte{}, r.CertHash...), tsBytes[:]...)
 
-	// NB: append timestamp to revoked cert bytes
-	binary.PutVarint(toHash.Bytes(), r.Timestamp)
-
-	return pkcrypto.SHA256Hash(toHash.Bytes())
+	return pkcrypto.SHA256Hash(toHash)
 }
 
 // Sign generates a signature using the passed key and attaches it to the revocation.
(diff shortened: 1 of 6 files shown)


pkg/pkcrypto/hashing.go
```diff
@@ -0,0 +1,18 @@
...
+
+package pkcrypto
+
+import (
+	"crypto"
+)
+
+// SHA256Hash calculates the SHA256 hash of the input data
+func SHA256Hash(data []byte) []byte {
+	hash := crypto.SHA256.New()
```
egonelbre: You can use `return sha256.Sum256(data)[:]`

pkg/peertls/extensions.go
```diff
@@ -235,46 +234,40 @@ func (e ExtensionHandlers) VerifyFunc() PeerCertVerificationFunc {
...
 		return err
 	}
 	return nil
 }
 
 // TBSBytes (ToBeSigned) returns the hash of the revoked certificate hash and
 // the timestamp (i.e. hash(hash(cert bytes) + timestamp)).
-func (r *Revocation) TBSBytes() ([]byte, error) {
-	toHash := new(bytes.Buffer)
+func (r *Revocation) TBSBytes() []byte {
```
egonelbre: This code in this func looks wrong.

`bytes.Buffer` seems unnecessary, we can use `[]byte` directly and avoid the panic.

`binary.PutVarint` writes the content to the beginning of byte array and does not append.

bryanchriswhite: dang.. I can't believe I did this :man_facepalming:

pkg/peertls/extensions.go
```diff
@@ -235,46 +234,40 @@ func (e ExtensionHandlers) VerifyFunc() PeerCertVerificationFunc {
...
-		return err
-	}
-
-	r.Signature, err = signHashOf(key, data)
+	data := r.TBSBytes()
+	sig, err := pkcrypto.SignHashOf(key, data)
 	if err != nil {
 		return err
 	}
+	r.Signature = sig
```
bryanchriswhite: is it bad form to assign directly like it was doing before? (`r.Signature, err := ...`)

egonelbre: One is more defensive about the changing of structures. When something returns invalid data and an error, then this can prevent some bugs. But assuming well behaved data to be returned then it shouldn't matter. I generally prefer the shorter form.

thepaul: I do tend to prefer a more defensive approach. Want me to change it to the shorter?

egonelbre: I have no strong feeling about this.

PR #1246, opened 2019-02-06
 Add checker irreparable test

mniewrzal:


satellite/satellitedb/irreparabledb.go
```diff
@@ -41,6 +41,7 @@ func (db *irreparableDB) IncrementRepairAttempts(ctx context.Context, segmentInf
 		dbxInfo.RepairAttemptCount++
 		updateFields := dbx.Irreparabledb_Update_Fields{}
 		updateFields.RepairAttemptCount = dbx.Irreparabledb_RepairAttemptCount(dbxInfo.RepairAttemptCount)
+		updateFields.SegDamagedUnixSec = dbx.Irreparabledb_SegDamagedUnixSec(segmentInfo.RepairUnixSec)
```
kaloyan-raev: How does this change relate to the test case?

pkg/datarepair/checker/checker_test.go
```diff
@@ -104,3 +105,77 @@ func TestOfflineNodes(t *testing.T) {
...
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		time.Sleep(2 * time.Second)
+
+		const numberOfNodes = 10
+		pieces := make([]*pb.RemotePiece, 0, numberOfNodes)
+		// use online nodes
+		for i, storagenode := range planet.StorageNodes {
+			pieces = append(pieces, &pb.RemotePiece{
+				PieceNum: int32(i),
+				NodeId:   storagenode.Identity.ID,
```
kaloyan-raev: You can use the shorter `storagenode.ID()`.

satellite/satellitedb/irreparabledb.go
mniewrzal: I asked @aligeti and he said that repair attempt time should always be updated with attempt count. Turns out that it wasn't updated.

pkg/datarepair/checker/checker_test.go
```diff
@@ -104,3 +105,77 @@ func TestOfflineNodes(t *testing.T) {
...
+			},
+		}
+
+		// put test pointer to db
+		pointerdb := planet.Satellites[0].Metainfo.Service
+		err := pointerdb.Put(pointer.Remote.PieceId, pointer)
+		assert.NoError(t, err)
+
+		checker := planet.Satellites[0].Repair.Checker
+		err = checker.IdentifyInjuredSegments(ctx)
```
egonelbre: Note directly invoking those service methods may cause races in the future.

mniewrzal: Yes, I know but I think we don't have easy way to disable services triggered by timers. I will add comment about it.

PR #1248, opened 2019-02-06
Fix discovery busy-spin

egonelbre:


pkg/discovery/service.go
```diff
@@ -129,6 +127,10 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 	}
 
 	for _, node := range list {
+		if ctx.Err() != nil {
```
mniewrzal: Just an idea, maybe `Ping` and `UpdateUptime` should check if context was canceled internally?

egonelbre: I mean, it should, but I have no idea why doesn't.

egonelbre: Ah, found it finally.

egonelbre: However, it's still nice to fail early, since we might have a bunch of things in the queue.

mniewrzal: ah yes you are right, loop is continuing is case of error.

egonelbre: https://github.com/storj/storj/pull/1249 should fix the ping not being canceled issue

PR #1249, opened 2019-02-06
Proper planet shutdown

egonelbre:
Previously Planet was not waiting for all peers to stop before shutting down individual services causing flaky behavior.

Together with fixing context shutdown also found places where context is not being propagated.


pkg/transport/transport.go
```diff
@@ -94,7 +97,10 @@ func (transport *Transport) DialAddress(ctx context.Context, address string, opt
 	}
 
 	options := append([]grpc.DialOption{dialOpt, grpc.WithBlock()}, opts...)
-	conn, err = grpc.Dial(address, options...)
+	conn, err = grpc.DialContext(ctx, address, options...)
+	if err == context.Canceled {
```
mniewrzal: Can't we assume that if err != nil then conn == nil and remove this if statement?

satellite/console/consoleweb/server.go
```diff
@@ -129,7 +130,17 @@ func (s *Server) Run(ctx context.Context) error {
...
+	group.Go(func() error {
+		<-ctx.Done()
+		return s.server.Shutdown(nil)
+	})
+
+	group.Go(func() error {
+		return s.server.Serve(s.listener)
+	})
+
+	return group.Wait()
```
mniewrzal: Maybe similar code from pkg/server/server.go:100 should be adjusted to this form? it looks simpler and cleaner.

egonelbre: fixed

egonelbre: I just realized that it wouldn't close the other goroutine when only the serving returns.

PR #1250, opened 2019-02-06
 Update repairer tests

mniewrzal:
Won't work until https://storjlabs.atlassian.net/browse/V3-1154 will be fixed.


mniewrzal:
will bo done in #1494

PR #1251, opened 2019-02-06
Updating account activation flow

crawter:


internal/migrate/create_test.go
```diff
@@ -24,19 +24,19 @@ func TestCreate_Sqlite(t *testing.T) {
 	defer func() { assert.NoError(t, db.Close()) }()
 
 	// should create table
-	err = Create("example", &sqliteDB{db, "CREATE TABLE example_table (id text)"})
+	err = Create("example", &sqliteDB{db, "CREATE TABLE users (id text, email text, status int)"})
```
egonelbre: I don't think this file should change.

internal/migrate/create.go
```diff
@@ -47,6 +47,13 @@ func Create(identifier string, db DB) error {
 			return Error.Wrap(utils.CombineErrors(err, tx.Rollback()))
 		}
 
+		// creating multi column unique index with condition.
+		// status = 1 indicates that account is activated
+		_, err = tx.Exec(db.Rebind("CREATE UNIQUE INDEX console_unique_index_email_status ON users(email, status) WHERE (status != 0)"))
```
egonelbre: `internal/migrate` is shared by all different databases in the system, so this is definitely not the correct place where to do this.

satellite/console/users_test.go
```diff
@@ -48,21 +48,13 @@ func TestUserRepository(t *testing.T) {
...
 
-		t.Run("Can't insert user with same email twice", func(t *testing.T) {
-			user := &console.User{
-				FirstName:    name,
-				LastName:     lastName,
-				Email:        email,
-				PasswordHash: []byte(passValid),
-				CreatedAt:    time.Now(),
-			}
+			insertedUser.Status = 1
```
egonelbre: Use the named constant here.

satellite/console/users_test.go
```diff
@@ -93,6 +85,7 @@ func TestUserRepository(t *testing.T) {
 				FirstName:    newName,
 				LastName:     newLastName,
 				Email:        newEmail,
+				Status:       1,
```
egonelbre: Use the named constant here.

satellite/console/users.go
```diff
@@ -64,6 +64,18 @@ func (user *CreateUser) IsValid() error {
 	return errs.Combine()
 }
 
+// UserStatus - is used to indicate status of the users account
+type UserStatus int
+
+const (
+	// Inactive is a user status that he receives after registration
+	Inactive UserStatus = iota
```
egonelbre: Specify the numbers explicitly, otherwise people may accidentally rename/move these things and break things.

cmd/storj-sim/console.go
```diff
@@ -47,13 +52,47 @@ func graphqlDo(client *http.Client, request *http.Request, jsonResponse interfac
...
 }
 
+func generateActivationKey(userID uuid.UUID, email string, createdAt time.Time) (string, error) {
+	claims := consoleauth.Claims{
+		ID:         userID,
+		Email:      email,
+		Expiration: createdAt.Add(24 * time.Hour),
+	}
+
+	signer := &consoleauth.Hmac{Secret: []byte("my-suppa-secret-key")}
```
egonelbre: add a todo note about fixing this

PR #1254, opened 2019-02-06
Smtp mail sending library

rikysya:
-Create library which can send email messages via smtp protocol


pkg/mail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package mail
```
egonelbre: Let's move this into `internal/mail/gmail`.

pkg/mail/sender.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package mail
+
+import "net/smtp"
+
+// Sender encapsulates information about the mail sender
+type Sender struct {
+	Mail string
```
egonelbre: Is this an address or something else?

pkg/mail/server.go
```diff
@@ -0,0 +1,67 @@
...
+	"crypto/tls"
+	"net"
+	"net/smtp"
+
+	"github.com/zeebo/errs"
+)
+
+// SMTPServer is smtp server
+type SMTPServer struct {
+	Host, Port string
```
egonelbre: Using the full address directly is usually easier than keeping them split.

pkg/mail/server.go
```diff
@@ -0,0 +1,67 @@
...
+	Sender     Sender
+}
+
+// Address returns server's network address
+func (server *SMTPServer) Address() string {
+	return net.JoinHostPort(server.Host, server.Port)
+}
+
+// SendEmail sends email message to the given recipient
+func (server *SMTPServer) SendEmail(rcpt string, msg []byte) error {
```
egonelbre: We probably should take an `Message` type instead of `msg []byte` since things need to be formatted as RFC 822-style email, and it's probably easier to implement that with something like:

```
type Message struct {
	From    string
	To      string
	Subject string
	Id      string
	Date    time.Time

	ReceiptTo   []string
	ContentType string

	Body string
	Parts []Part
}

type Part struct {
	Type        string
	Encoding    string
	Disposition string
	Content     string
}
```

pkg/mail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/zeebo/errs"
+)
+
+// GmailAuth is XOAUTH2 implementation of smtp.Auth interface
+type GmailAuth struct {
```
egonelbre: Let's do `GmailAuth` as a separate PR, since this after adding message encoding will get bigger.

pkg/mail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+	Mail string
```
egonelbre: Is it `From`, `To`, authentication address etc.? This doesn't give enough information.

pkg/mail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Credentials represents OAUTH2 credentials
+type Credentials struct {
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+	TokenURI     string `json:"token_uri"`
+}
+
+// TokenStorage is a thread safe storage for OAUTH2 token and credentials
+type TokenStorage struct {
```
egonelbre: `TokenStore`

pkg/mail/server.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package mail
```
egonelbre: Let's move this into `internal/mail`.

pkg/mail/server.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"crypto/tls"
+	"net"
+	"net/smtp"
+
+	"github.com/zeebo/errs"
+)
+
+// SMTPServer is smtp server
+type SMTPServer struct {
```
egonelbre: Maybe `Client`/`Mailer`/`Proxy`/`Sender`, because it isn't actually a `Server` that listens for incoming messages.

internal/mail/message.go
```diff
@@ -0,0 +1,142 @@
...
+			fmt.Fprintf(&body, "Content-Disposition: %s;", msg.Disposition)
+		}
+		fmt.Fprintf(&body, "\r\n")
+		fmt.Fprintf(&body, msg.Body)
+	}
+
+	return tocrlf(body.Bytes()), nil
+}
+
+func formatAddress(address Address) string {
```
egonelbre: Use https://golang.org/pkg/net/mail/#Address for this.

internal/mail/message.go
```diff
@@ -0,0 +1,142 @@
...
+	ContentType string
+	Encoding    string
+	Disposition string
+
+	Body  string
+	Parts []Part
+}
+
+// Address is email address of a sender/recipient with name
+type Address struct {
```
egonelbre: https://golang.org/pkg/net/mail/#Address

internal/mail/message.go
```diff
@@ -0,0 +1,142 @@
...
+	// date and id are optional as they can be set by server itself
+	if !msg.Date.IsZero() {
+		fmt.Fprintf(&body, "Date: %v\r\n", msg.Date)
+	}
+	if msg.ID != "" {
+		fmt.Fprintf(&body, "Message-ID: <%v>\r\n", msg.ID)
+	}
+
+	// parse content type
+	mtype, _, err := mime.ParseMediaType(msg.ContentType)
```
egonelbre: Can we make a requirement that Body is always utf-8 plain text, maybe rename `Body` -> `PlainText` as well.

internal/mail/message.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch {
+	// multipart upload, body is treated as plain text content of the message
+	// to support readability
+	case strings.HasPrefix(mtype, "multipart"):
+		wr := multipart.NewWriter(&body)
```
egonelbre: multipart has funcs for writing the field values https://golang.org/pkg/mime/multipart/#Writer.WriteField

rikysya: Not sure that I am following you on this one, why do i need to write form field here?

egonelbre: Nevermind, my mistake. I assumed it did something else.

internal/mail/message.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Bytes builds message and returns result as bytes
+func (msg *Message) Bytes() ([]byte, error) {
+	// always returns nil on read and write, so most of the errors can be ignored
+	var body bytes.Buffer
+
+	// write headers
+	fmt.Fprintf(&body, "MIME-Version: 1.0\r\n")
+	fmt.Fprintf(&body, "Subject: %v\r\n", msg.Subject)
```
egonelbre: The subject needs to be properly escaped.

internal/mail/message.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Bytes builds message and returns result as bytes
+func (msg *Message) Bytes() []byte {
+	// always returns nil on read and write, so most of the errors can be ignored
+	var body bytes.Buffer
+
+	// write headers
+	fmt.Fprintf(&body, "MIME-Version: 1.0\r\n")
+	fmt.Fprintf(&body, "Subject: <%v>\r\n", msg.Subject)
```
egonelbre: What happens if subject is "\n\n\n\n\n\n"?

rikysya: it breaks everything

egonelbre: Use `mime.QEncoding.Encode(msg.Subject)` for the values... if there are multiple values you can pass in `mime.QEncoding.Encode("a", "b", "c")`.

internal/mail/message.go
```diff
@@ -81,12 +82,12 @@ func (msg *Message) Bytes() []byte {
...
 		for _, part := range msg.Parts {
-			header := textproto.MIMEHeader{"Content-Type": []string{part.Type}}
+			header := textproto.MIMEHeader{"Content-Type": []string{mime.QEncoding.Encode("utf-8", part.Type)}}
 			if part.Encoding != "" {
-				header["Content-Transfer-Encoding"] = []string{part.Encoding}
+				header["Content-Transfer-Encoding"] = []string{mime.QEncoding.Encode("utf-8", part.Encoding)}
 			}
 			if part.Disposition != "" {
-				header["Content-Disposition"] = []string{part.Disposition}
+				header["Content-Disposition"] = []string{mime.QEncoding.Encode("utg-8", part.Disposition)}
```
egonelbre: typo `utg-8`

PR #1255, opened 2019-02-06
Fixes Difficulty Display Difference

stefanbenten:
Found a small display difference during the batch generation.
It displayed the diff one higher as it really is.

### DIFF
--- a/cmd/identity/batch.go
+++ b/cmd/identity/batch.go
@@ -192,7 +192,7 @@ func renderStats(screen *cui.Screen, stats []uint32) error {
 		if count == 0 {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


cmd/identity/batch.go
```diff
@@ -192,7 +192,7 @@ func renderStats(screen *cui.Screen, stats []uint32) error {
 		if count == 0 {
 			continue
 		}
-		printf(w, "%d\t%d\n", difficulty, count)
+		printf(w, "%d\t%d\n", difficulty-1, count)
```
egonelbre: rewrite the loop above, that way you don't need the -1 for the loaduint either.

PR #1257, opened 2019-02-06
remove certificate from bandwidth agreement, use caches

wthorp:
This PR removed certificates from bandwidth agreements, and uses caches of certificates on the satellite and storage nodes instead.

It does not fully address the need to make agreement signing deterministic.


pkg/bwagreement/server.go
```diff
@@ -50,15 +50,15 @@ type DB interface {
 
 // Server is an implementation of the pb.BandwidthServer interface
 type Server struct {
-	db     DB
-	NodeID storj.NodeID
-	logger *zap.Logger
+	db       DB
+	identity *identity.FullIdentity
```
aligeti: all this file needs is just storj.NodeID and its public key. Passing FullIdentity, I am thinking we are passing way more information that needed. Shouldnt we just keep it to only needed data ?

wthorp:
Planned Storage Node refactor will encompass these changes

PR #1258, opened 2019-02-06
removes some SignedMessage use

navillasa:
Removes most instances of pb.SignedMessage (there's more to take out but they shouldn't hurt anyone as is).

There used to be places in psserver where a PieceID was hmac'd with the SatelliteID, which was gotten from a SignedMessage. This PR makes it so some functions access the SatelliteID from the Payer Bandwidth Allocation instead.

This requires passing a SatelliteID into psserver functions where they weren't before, so the following proto messages have been changed:

- PieceId - satellite_id field added
This is so the psserver.Piece function has access to the SatelliteID when it needs to get the namespaced pieceID.
This proto message should probably be renamed to PieceRequest, or a new PieceRequest message should be created so this isn't misnamed.

- PieceDelete - satellite_id field added
This is so the psserver.Delete function has access to the SatelliteID when receiving a request to Delete.


pkg/auth/signature_test.go
```diff
@@ -46,34 +46,6 @@ func TestSignedMessageVerifier(t *testing.T) {
 	identity, err := ca.NewIdentity()
 	assert.NoError(t, err)
 
-	signature, err := GenerateSignature(identity.ID.Bytes(), identity)
+	_, err = GenerateSignature(identity.ID.Bytes(), identity)
```
mniewrzal: I think `TestSignedMessageVerifier` can be removed completely.

pkg/pb/piecestore.proto
```diff
@@ -58,14 +58,14 @@ message PieceStore {
 
   RenterBandwidthAllocation bandwidth_allocation = 1;
   PieceData piece_data = 2;
-  SignedMessage authorization = 3;
+  PayerBandwidthAllocation payer_allocation = 3;
```
mniewrzal: `PayerBandwidthAllocation` should be inside `RenterBandwidthAllocation`. Most probably we don't need new field.

pkg/pb/piecestore.proto
```diff
@@ -84,7 +84,7 @@ message PieceRetrieval {
...
```
mniewrzal: `PayerBandwidthAllocation` should be inside `RenterBandwidthAllocation`. Most probably we don't need new field.

pkg/pb/piecestore.proto
egonelbre: How do we handle migration here. We shouldn't probably use the same tag number. Alternatively maybe make a completely new endpoint? @jtolds?

pkg/piecestore/psserver/server.go
```diff
@@ -178,15 +170,6 @@ func (s *Server) Piece(ctx context.Context, in *pb.PieceId) (*pb.PieceSummary, e
 		return nil, err
 	}
 
-	match, err := regexp.MatchString("^[A-Za-z0-9]{20,64}$", id)
```
egonelbre: why was this part removed?

pkg/pb/piecestore.proto
mniewrzal: maybe at the beginning, we can detect backward incompatible changes? I found something like this https://github.com/nilslice/protolock

egonelbre: sgtm

pkg/piecestore/psserver/server.go
navillasa: this was removed because all pieceIDs are passed into the `getNamespacedPieceID` function which will return a base58 encoded string, so I thought there wasn't a need for this check, but I can go ahead and add it back.

egonelbre: Ah, yeah... although I just realized there's a problem with this. It could have collisions. file-systems should be regarded as case-insensitive, whereas base58 is somewhat case sensitive. Yeah, let's keep it removed.

pkg/pb/piecestore.proto
navillasa: I know we talked already but in case anyone else is reviewing--
the Put function on the piecestore client side only has PBA (not RBA) as a parameter. It also doesn’t even put RBA on the &pb.PieceStore message to be sent. So when you call `rba := recv.GetRBA()` on the server side it’s going to be nil. This can be verified currently on master. https://github.com/storj/storj/blob/39a1f175c4ea76ff3dc837a88d582e10d9bd6fb1/pkg/piecestore/psclient/client.go#L120

pkg/pb/piecestore.proto
navillasa: I know we talked already but in case anyone else is reviewing--
the pieceRanger.Range function only has PBA on the struct, not RBA - so only PBA can be passed through as a message on the stream. Even though pieceRange later creates NewStreamReader which should add PBA to RBA, it doesn't seem to be working because calling `recv.GetBandwidthAllocation()` inside the Retrieve function on the server side yields nil (here and in master).
https://github.com/storj/storj/blob/877dd3deb0531025b43a481b702ee78b4ff7d0e0/pkg/piecestore/psclient/pieceranger.go#L51

mniewrzal:
I'm starting to think that we need to do this change in two steps. Mainly because we have some users, we don't want to break their nodes and we will have no option to sync satellite and storage nodes updates to avoid it:
First - storage node:
* mark `authorization` as deprecated in proto file
* remove `authorization` (SignedMessage) from psserver 
* don't remove `authorization` (SignedMessage) from psclient and in general from satellite

Second - satellite (after some time): 
* remove `authorization` (SignedMessage) from proto file
* remove all other occurrences of `authorization` (SignedMessage)

Of course, it depends if this is important right now? @jtolds @egonelbre

pkg/piecestore/psclient/client.go
```diff
@@ -117,15 +117,37 @@ func (ps *PieceStore) Meta(ctx context.Context, id PieceID) (*pb.PieceSummary, e
...
+func (ps *PieceStore) Put(ctx context.Context, id PieceID, data io.Reader, ttl time.Time, pba *pb.PayerBandwidthAllocation) error {
 	stream, err := ps.client.Store(ctx)
 	if err != nil {
 		return err
 	}
 
+	// Making a copy, otherwise there will be a data race
+	// when another goroutine tries to write the cached size
+	// of this instance at the same time.
+	pbaCopy := &pb.PayerBandwidthAllocation{
```
wthorp: There's probably a shorter way of expressing this... possibly *pbaCopy = *pba?

pkg/piecestore/psclient/pieceranger.go
```diff
@@ -63,10 +63,36 @@ func (r *pieceRanger) Range(ctx context.Context, offset, length int64) (io.ReadC
 		return ioutil.NopCloser(bytes.NewReader([]byte{})), nil
 	}
 
+	// Making a copy, otherwise there will be a data race
+	// when another goroutine tries to write the cached size
+	// of this instance at the same time.
+	pbaCopy := &pb.PayerBandwidthAllocation{
```
wthorp: There's probably a shorter way of expressing this... otherwise since we're doing it more than once we might want to make a .Clone() pointer receiver

pkg/piecestore/psclient/client.go
thepaul: There's a cached atomic variable contained in the structure that is getting written to by other goroutines, so Nat is avoiding reading the whole structure so that that one cached variable doesn't get read, tripping the race detector.

pkg/piecestore/psclient/pieceranger.go
thepaul: 👍

pkg/piecestore/psclient/client.go
navillasa: I think it definitely needs to be a deep copy, which ends up looking verbose, but I can make it a function to reduce repeated code.

pkg/piecestore/psclient/pieceranger.go
```diff
@@ -63,10 +63,36 @@ func (r *pieceRanger) Range(ctx context.Context, offset, length int64) (io.ReadC
...
+	pbaCopy.Certs = make([][]byte, len(r.pba.Certs))
+	copy(pbaCopy.Certs, r.pba.Certs)
+	pbaCopy.Signature = make([]byte, len(r.pba.Signature))
+	copy(pbaCopy.Signature, r.pba.Signature)
+
+	rba := &pb.RenterBandwidthAllocation{
+		PayerAllocation: *pbaCopy,
+		StorageNodeId:   r.c.remoteID,
+	}
+	fmt.Println(r.id.String())
```
mniewrzal: Remove printing

pkg/piecestore/psserver/retrieve.go
```diff
@@ -35,17 +35,19 @@ func (s *Server) Retrieve(stream pb.PieceStoreRoutes_RetrieveServer) (err error)
...
 		return RetrieveError.New("PieceStore message is nil")
 	}
 
-	id, err := getNamespacedPieceID([]byte(pd.GetId()), getNamespace(authorization))
+	rba := recv.GetBandwidthAllocation()
+	if rba == nil {
+		return RetrieveError.New("RenterBandwidthAllocation message is nil")
+	}
+
+	pba := rba.PayerAllocation
```
mniewrzal: Check `pba` for nil too

pkg/piecestore/psserver/store.go
```diff
@@ -51,7 +46,14 @@ func (s *Server) Store(reqStream pb.PieceStoreRoutes_StoreServer) (err error) {
 		return StoreError.New("piece ID not specified")
 	}
 
-	id, err := getNamespacedPieceID([]byte(pd.GetId()), getNamespace(authorization))
+	rba := recv.GetBandwidthAllocation()
+	if rba == nil {
+		return StoreError.New("RenterBandwidthAllocation message is nil")
+	}
+
+	pba := rba.PayerAllocation
```
mniewrzal: Check `pba` for nil too

pkg/piecestore/psclient/pieceranger.go
mniewrzal: I think there is `proto.Clone()` func but it needs to be checked if it work as expected

pkg/piecestore/psserver/retrieve.go
navillasa: pba can't be type nil for some reason. Should I check if it's just an empty instance like `pb.PayerBandwidthAllocation{}`?

mniewrzal: Ah you are right it cannot be nil. Checking if its empty will be ok.

PR #1259, opened 2019-02-07
fix TBSBytes func

bryanchriswhite:


bryanchriswhite:
This is a breaking change to revocations but I don't think that should be an issue - afaik nobody has used revocations yet.

pkg/peertls/extensions.go
```diff
@@ -247,17 +247,11 @@ func (r Revocation) Verify(signingCert *x509.Certificate) error {
...
-	_, err := toHash.Write(r.CertHash)
-	if err != nil {
-		// bytes.Buffer#Write() is documented as never returning an error
-		panic(err)
-	}
-
-	// NB: append timestamp to revoked cert bytes
-	binary.PutVarint(toHash.Bytes(), r.Timestamp)
+	var tsBytes [8]byte
+	binary.PutVarint(tsBytes[:], r.Timestamp)
```
egonelbre: Is there a reason why we are using varint, but not only the number of bytes it wrote? Also varint can be maximally 10 bytes.

Maybe use `binary.LittleEndian` or `binary.BigEndian` instead?

thepaul: with Varint, this code will still be good after the year 2157 😁

egonelbre: Unless we have people from the future using our system and they forget to change their time-zone to the past.

bryanchriswhite: that's on them honestly -- that just sounds like part of responsible time travel :p

bryanchriswhite: @egonelbre i fixed the length -- `binary.Write` has an error return value and the big and little endian  `ByteOrder` implementations only have `PutUintX` methods and it feels weird to cast the timestamp to a uint. What do you think about it as it is currently?

egonelbre: Casting it into uint64 should be fine as long as you cast it back with int64. That's how it's usually encoded. To avoid errors you can write as:

```
// big endian
v := uint64(r.timestamp)
b[0] = byte(v >> 56)
b[1] = byte(v >> 48)
b[2] = byte(v >> 40)
b[3] = byte(v >> 32)
b[4] = byte(v >> 24)
b[5] = byte(v >> 16)
b[6] = byte(v >> 8)
b[7] = byte(v)
```

PR #1260, opened 2019-02-07
Gmail auth implementation for SMTP

rikysya:
-Add gmail auth implementation and helper methods for SMTP

### DIFF
--- a/internal/mail/gmail/gmailauth.go
+++ b/internal/mail/gmail/gmailauth.go
@@ -32,7 +32,7 @@ func (auth *Auth) Start(server *smtp.ServerInfo) (proto string, toServer []byte,
 		return "", nil, err
 	}
 
-	format := fmt.Sprintf("user=%s\001auth=%s %s\001\001", auth.UserEmail, token.Type, token.AccessToken)
+	format := fmt.Sprintf("user=%s\x01auth=%s %s\x01\x01", auth.UserEmail, token.Type, token.AccessToken)
 	return "XOAUTH2", []byte(format), nil
 }
 
--- a/internal/mail/gmail/gmailauth.go
+++ b/internal/mail/gmail/gmailauth.go
@@ -18,15 +18,15 @@ import (
 	"github.com/zeebo/errs"
 )
 
-// GmailAuth is XOAUTH2 implementation of smtp.Auth interface
-type GmailAuth struct {
+// Auth is XOAUTH2 implementation of smtp.Auth interface for gmail
+type Auth struct {
 	UserEmail string
 
 	Storage *TokenStore
 }
 
 // Start returns proto and auth credentials for first auth msg
-func (auth *GmailAuth) Start(server *smtp.ServerInfo) (proto string, toServer []byte, err error) {
+func (auth *Auth) Start(server *smtp.ServerInfo) (proto string, toServer []byte, err error) {
 	token, err := auth.Storage.Token()
 	if err != nil {
 		return "", nil, err
@@ -37,7 +37,7 @@ func (auth *GmailAuth) Start(server *smtp.ServerInfo) (proto string, toServer []
 }
 
 // Next sends empty response to solve SASL challenge if response code is 334
-func (auth *GmailAuth) Next(fromServer []byte, more bool) (toServer []byte, err error) {
+func (auth *Auth) Next(fromServer []byte, more bool) (toServer []byte, err error) {
 	if more {
 		return make([]byte, 0), nil
 	}
--- a/internal/mail/gmail/gmailauth.go
+++ b/internal/mail/gmail/gmailauth.go
@@ -1,0 +1,174 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package gmail
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/smtp"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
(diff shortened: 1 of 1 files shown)


internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/zeebo/errs"
+)
+
+// GmailAuth is XOAUTH2 implementation of smtp.Auth interface
+type GmailAuth struct {
```
egonelbre: This can be now called just `Auth`

internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+func NewTokenStore(creds Credentials, token Token) *TokenStore {
+	return &TokenStore{
+		token: token,
+		creds: creds,
+	}
+}
+
+// Token retrieves token in a thread safe way and refreshes it if needed
+func (s *TokenStore) Token() (*Token, error) {
+	s.mu.Lock()
```
egonelbre: use `defer s.mu.Unlock()`, it ensures that locks get closed even in the presence of panics

internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Start returns proto and auth credentials for first auth msg
+func (auth *Auth) Start(server *smtp.ServerInfo) (proto string, toServer []byte, err error) {
+	token, err := auth.Storage.Token()
+	if err != nil {
+		return "", nil, err
+	}
+
+	format := fmt.Sprintf("user=%s\001auth=%s %s\001\001", auth.UserEmail, token.Type, token.AccessToken)
```
egonelbre: This escaping looks wrong. Also use hex values rather than octal, they should be a bit clearer to read.

internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+	return "XOAUTH2", []byte(format), nil
+}
+
+// Next sends empty response to solve SASL challenge if response code is 334
+func (auth *Auth) Next(fromServer []byte, more bool) (toServer []byte, err error) {
+	if more {
+		return make([]byte, 0), nil
+	}
+
+	return nil, nil
```
egonelbre: The whole func pretty much could be `return nil, nil`, the make thing seems superfluous... Also, it looks like it should contain something more?

internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,174 @@
...
+	return &Token{
+		AccessToken:  t.AccessToken,
+		RefreshToken: t.RefreshToken,
+		Type:         t.Type,
+		Expiry:       time.Now().Add(t.Expires * time.Second),
+	}, nil
+}
+
+// CredsFromJSON is a helper method for parsing OAUTH2 google app credentials from raw json data
+func CredsFromJSON(jsonData []byte) (*Credentials, error) {
```
egonelbre: This looks unused, and it's unclear where `jsonData` comes from.

internal/mail/gmail/gmailauth.go
rikysya: that's what SASL challenge requires. If you send wrong credentials you will receive encoded error with 334 code, to solve the challenge you need to send an empty response. Making bytes array explicitly is to prevent return on nil in golang SMTP library https://github.com/golang/go/blob/e1b49ad608c0b51acf75f8248e1167acbbcb5d30/src/net/smtp/smtp.go#L233, which will allow you to send next command even thought your authentication has failed.

internal/mail/gmail/gmailauth.go
rikysya: JSON data can come from a file, for example. I think that it is better to put responsibility for getting actual data(dealing with files or network) on code that call's this method.

internal/mail/gmail/gmailauth.go
rikysya: agree

internal/mail/gmail/gmailauth.go
egonelbre: I mean we use config flags throughout the system that can be passed in via environment flags or command line flags or config file. I'm not sure why is the extra way of configuring things needed?

rikysya: Ok, I get it now. So this method is for parsing credentials file that you download from google console. But if we want populate this info through config or flags, than yes it's unnecessary. I will remove this code then.

internal/mail/gmail/gmailauth.go
rikysya: I don't quite get why should we escape here. Token and it's type comes from the server it's not like you have to modify them. And if user will input bad data, then parsing will fail and you will receive authentication error. Or did I miss something?

egonelbre: fair enough :)

internal/mail/gmail/gmailauth.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	// handle google expires_in field value
+	var t struct {
+		AccessToken  string        `json:"access_token"`
+		RefreshToken string        `json:"refresh_token"`
+		Type         string        `json:"token_type"`
+		Expires      time.Duration `json:"expires_in"`
+	}
+	err = json.NewDecoder(bytes.NewReader(body)).Decode(&t)
```
egonelbre: Ah, you can bass here in `resp.Body` directly, you don't need to do the separate `ioutil.ReadAll`.

PR #1261, opened 2019-02-07
implement basic migration

egonelbre:
This implements a bare minimum database/file migration library.

It does not try to solve all the problems, but rather allow to have basic migration so that we can later figure out what needs we have with regards to good migration library.


internal/migrate/versions_test.go
```diff
@@ -0,0 +1,132 @@
...
+	require.NoError(t, err)
+
+	defer func() { assert.NoError(t, db.Close()) }()
+
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	testDB := &sqliteDB{DB: db}
+
+	err = ioutil.WriteFile(ctx.File("alpha.txt"), []byte("test"), 0644)
```
egonelbre: This test doesn't use this file.

internal/migrate/versions.go
```diff
@@ -84,13 +83,13 @@ func (migration *Migration) Run(log *zap.Logger, db DB) error {
 	}
 
 	if version < 0 {
-		log.Info("Latest Version", zap.Int("version", version))
+		log.Info("Migration: Latest Version", zap.Int("version", version))
```
egonelbre: remove the prefixes, they are not needed because the logger should already have the necessary prefix.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -92,8 +92,7 @@ func (db *DB) init() (err error) {
 		},
 	}
 
-	// TODO pass logger
-	return migration.Run(zap.NewNop(), migrate.NewSqliteDB(db.DB, ""))
+	return migration.Run(zap.L(), migrate.NewSqliteDB(db.DB, ""))
```
egonelbre: use `db.log.Named("migrate")`

mniewrzal: psdb.go doesn't have direct access to `log`

egonelbre: Ah, wait... do not change psdb in this PR... that should be a separate PR.

But, if psdb doesn't have access to log yet, then it should be passed down somehow.

internal/migrate/create.go
```diff
@@ -5,21 +5,71 @@ package migrate
...
 // DB is the minimal implementation that is needed by migration.
 type DB interface {
 	Begin() (*sql.Tx, error)
 	Schema() string
 	Rebind(string) string
 }
 
-// Error is the default migrate errs class
-var Error = errs.Class("migrate")
+type sqliteDB struct {
```
egonelbre: I don't think this should be exposed. The DB `interface{}` is based off of dbx...

pkg/piecestore/psserver/psdb/psdb.go
mniewrzal: ok, can be separte PR. Later we can pass logger down.

internal/migrate/create.go
mniewrzal: yeah but we need somehow rebind sql params for postgres script

egonelbre: It can be internal to the tests.

mniewrzal: ok, I assumed that at some point we will need postgres support. I will leave it internal.

egonelbre: dbx.DB has those methods already defined

internal/migrate/versions_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	multipleMigration(t, db, &sqliteDB{DB: db})
+}
+
+func TestMultipleMigrationPostgres(t *testing.T) {
+	if *testPostgres == "" {
+		t.Skipf("postgres flag missing, example:\n-postgres-test-db=%s", defaultPostgresConn)
+	}
+
+	db, err := sql.Open("postgres", *testPostgres)
```
egonelbre: This part would benefit using custom schema to avoid thrashing user database. i.e. https://github.com/storj/storj/blob/master/satellite/satellitedb/database.go#L66, but we can do that later

mniewrzal: I was thinking about reusing code from satellite but I didn't want to add to big overhead. I can reorganize this later.

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+	1. update a table schema
+	2. move files
+	3. update a table schema
+	4. update a table schema, which fails
+
+	In this case there's no easy way to rollback the moving of files.
+
+2. Undoing migrations.
+
+	Intentionally left out, because we do not gain that much from currently.
```
jtolds: good call. let's just make backups first

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+		if err := tx.Commit(); err != nil {
+			return Error.Wrap(err)
+		}
+	}
+
+	return nil
+}
+
+// createVersionTable creates a new version tabe
+func (migration *Migration) createVersionTable(log *zap.Logger, db DB) error {
```
wthorp: nit:  maybe name this ensureVersionTable?

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	_, err = tx.Exec(db.Rebind(`CREATE TABLE IF NOT EXISTS ` + migration.Table + ` (version int, commited_at text)`))
```
wthorp: why do we use multiple migration tables rather than one w/ an additional 'name' column?

egonelbre: I wasn't sure whether we want to hardcode the migration table -- single binary should probably still use a single table.

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// Run runs the migration steps
+func (migration *Migration) Run(log *zap.Logger, db DB) error {
+	err := migration.ValidTableName()
+	if err != nil {
+		return err
+	}
+
+	err = migration.createVersionTable(log, db)
```
thepaul: It might be worth acquiring an advisory lock here, in case it ever happens that two connections try to run migrations at the same time.

internal/migrate/versions.go
wthorp: why do we store commited_at as a string?

egonelbre: It's mainly for debugging later and STRING is supported on postgres and sqlite in the same way. Mainly avoids some complexity with handling different drivers.

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+
+	This probably should be done by the user of this library, when there's disk-space available.
+
+4. Figuring out what the exact executed steps are.
+*/
+
+// Migration describes a migration steps
+type Migration struct {
+	Table    string
+	OnCreate Action
```
wthorp: Is OnCreate substantially different than Steps[0]?  Should we simplify this more?

egonelbre: Yes, @mniewrzal will remove it.

mniewrzal: its already removed ;) just refresh github

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+	Steps    []*Step
+}
+
+// Step describes a single step in migration.
+type Step struct {
+	Description string
+	Version     int // Versions should start at 1
```
wthorp: if Migrations.Steps is an ordered list, do we really need Step.Version?

egonelbre: Yes, it should be ordered. I prefer to keep the version explicit, it's easier to find later the appropriate step.

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+
+// getLatestVersion finds the latest version table
+func (migration *Migration) getLatestVersion(log *zap.Logger, db DB) (int, error) {
+	tx, err := db.Begin()
+	if err != nil {
+		return -1, Error.Wrap(err)
+	}
+
+	var version sql.NullInt64
+	err = tx.QueryRow(db.Rebind(`SELECT MAX(version) FROM ` + migration.Table)).Scan(&version)
```
wthorp: why do we need to wrap a single SELECT in a transaction?

internal/migrate/versions.go
mniewrzal: we can add validation to be sure thats ordered.

internal/migrate/versions.go
egonelbre: Because that way we can keep the database interface requirements simpler.

internal/migrate/versions.go
egonelbre: https://golang.org/pkg/sort/#SliceIsSorted

internal/migrate/versions.go
```diff
@@ -0,0 +1,209 @@
...
+		if err != nil {
+			return Error.Wrap(errs.Combine(err, tx.Rollback()))
+		}
+
+		if err := tx.Commit(); err != nil {
+			return Error.Wrap(err)
+		}
+	}
+
+	for _, step := range migration.Steps {
```
wthorp: nit: this could just be a standard for x=, x<; x++ loop without the continue

egonelbre: With versioned migrations I've seen steps removed/optimized eventually, so would prefer with continue... it doesn't pose a performance concern either.

internal/migrate/versions.go
wthorp: I figured the Description handled finding the step later

mniewrzal: what do you mean?

internal/migrate/versions.go
wthorp: Some clarification (by naming or by code comments) around when to use a different 'Migration.Table' would be a good idea.  Different binaries might require the same set of database updates.  Different parts of an application might be entire atomic/unrelated.  The intent is not clear.

internal/migrate/versions.go
wthorp: "I prefer to keep the version explicit, it's easier to find later the appropriate step." ... If we see a failure during migrate, we could debug the error based on the friendly 'Description'.  I still don't understand the value of Version.  Its not a big deal though, it just seems like it will cause slightly more merge issues to have it.

mniewrzal: but how you want to distinguish which steps were applied with older bineries? let say your current db is in version 5 and binaries 10. how to avoid applying 1-5?

wthorp: You just use the index into the Migrate.Steps slice as the version that you store in the database.  The database still says you're at step 5, you just didn't have to explicitly specify Version=5 on the fifth step.  It's definitely nitpicking at this point.

PR #1263, opened 2019-02-07
Increase sleep time in overlay

egonelbre:
Overlay tests are flaky, increase time to 4sec.


egonelbre:
Closed in favor of #1275

PR #1264, opened 2019-02-07
Disable colors in dashboard by default. --color flag to enable them

kaloyan-raev:
Fixes https://storjlabs.atlassian.net/browse/V3-1146


aleitner:
Just ran on my terminal and it worked. That was much easier than I thought

cmd/storagenode/main.go
```diff
@@ -75,6 +75,7 @@ var (
 		Address         string `default:":28967" help:"address for dashboard service"`
 		ExternalAddress string `default:":28967" help:"address that your node is listening on if using a tunneling service"`
 		BootstrapAddr   string `default:"bootstrap.storj.io:8888" help:"address of server the storage node was bootstrapped against"`
+		Colors          bool   `default:"false" help:"use colors in dashboard interface"`
```
egonelbre: Usually it's called just `--color`

kaloyan-raev: Renamed.

PR #1265, opened 2019-02-07
Satellite Web Changed graphQL default port

Barterio:


web/satellite/src/utils/apolloManager.ts
```diff
@@ -9,7 +9,7 @@ import { getToken } from '@/utils/tokenManager';
 
 // Satellite url
 const satelliteUrl = new HttpLink({
-	uri: 'http://localhost:8081/api/graphql/v0',
+	uri: 'http://localhost:10100/api/graphql/v0',
```
egonelbre: Maybe let's not hardcode these and let the server set them while serving the page?

Barterio: Yes, I agree that we should parse Storj Config for default port mapping.
This is temporary solution to run web satellite account without changing source code.

PR #1266, opened 2019-02-07
Convert Payments to use SQL, for SUM() and Wallet

wthorp:
https://storjlabs.atlassian.net/browse/V3-1085

1. LEFT JOIN overlaycache on nodes.id = overlaycache.node_id to select wallet address
2. Sum by node id: 
"byte-hours:AtRest",
"bytes:BWRepair-GET",
"bytes:BWRepair-PUT",
"bytes:BWAudit",
"bytes:BWGet",
"bytes:BWPut"
3. Remove the date field (keep nodecreationdate though)

### DIFF
--- a/pkg/accounting/rollup/rollup_test.go
+++ b/pkg/accounting/rollup/rollup_test.go
@@ -30,7 +30,7 @@ func TestQuery(t *testing.T) {
 		timestamp := time.Now().UTC().AddDate(0, 0, -1*days)
 		start := timestamp
 
-		for i := 1; i <= days; i++ {
+		for i := 0; i <= days; i++ {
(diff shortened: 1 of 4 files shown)


pkg/accounting/rollup/rollup_test.go
```diff
@@ -50,21 +50,23 @@ func TestQuery(t *testing.T) {
 
 			rows, err := planet.Satellites[0].DB.Accounting().QueryPaymentInfo(ctx, start, end)
 			assert.NoError(t, err)
-			if i == 1 {
+			if i == 0 { // we need at least two days for rollup to work
 				assert.Equal(t, 0, len(rows))
-				return
+				continue
```
wthorp: this had been a bug!

cam-a: Nice catch

PR #1267, opened 2019-02-07
sync2.Cycle for controllable intervals

egonelbre:


internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+		return cycle.Run(ctx, fn)
+	})
+}
+
+// Run runs the specified function.
+func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error) error {
+	cycle.initialize()
+	defer close(cycle.quit)
+
+	currentInterval := cycle.interval
```
coyle: since you have a Setter on interval, should you also have a Getter ?

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+	cycle.ticker = time.NewTicker(currentInterval)
+	if err := fn(ctx); err != nil {
```
coyle: There's the possibility that this function take longer to complete than interval?

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+			if err := fn(ctx); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+// sendControl sends a control message
+func (cycle *Cycle) sendControl(message interface{}) {
+	cycle.initialize()
```
coyle: why call this here?

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+	select {
```
coyle: so this either sends a message or quits ?

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+	case cycle.control <- message:
+	case <-cycle.quit:
+	}
+}
+
+// Stop stops the cycle permanently
+func (cycle *Cycle) Stop() {
```
coyle: ok never mind this answers that

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+	cycle.sendControl(nil)
+}
+
+// ChangeInterval allows to change the ticker interval after it has started.
+func (cycle *Cycle) ChangeInterval(interval time.Duration) {
+	cycle.sendControl(interval)
```
coyle: the other control interfaces are structs, would it be cleaner to also make this a struct?

internal/sync2/cycle.go
egonelbre: Probably isn't needed anywhere, so no need to implement.

internal/sync2/cycle.go
egonelbre: Yeah, hmm... however this is inline more how the loop in general works, since it starts to tick immediately when running the func.

internal/sync2/cycle.go
egonelbre: because for the control functions to work properly they need to be initialized.

internal/sync2/cycle.go
egonelbre: This relies on closing behavior on channels. `<-quit` blocks until it's closed.

internal/sync2/cycle.go
egonelbre: sure can do that

egonelbre: fixed

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+
+// TriggerWait ensures that the loop is done at least once and waits for completion.
+// If it's currently running it waits for the previous to complete and then runs.
+func (cycle *Cycle) TriggerWait() {
+	done := make(chan struct{})
+	cycle.sendControl(cycleTrigger{done})
+	select {
+	case <-done:
+	case <-cycle.stop:
+	}
```
thepaul: `close(done)`? I know normally it's best for writers to close channels, but it's hard in this case to know if the writer will ever even know about it, and once this function exits for either reason, we do know nothing else should want to write to it.

internal/sync2/cycle.go
thepaul: Is it ever correct for this to be called before `Run()`, though?

egonelbre: Because in some cases Run will be called inside a goroutine, there's an inherent race between something outside calling the control method and Run. Even the test case has this issue. 

The problem isn't calling before Run... but rather that they are called concurrently.

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Start runs the specified function with an errgroup
+func (cycle *Cycle) Start(ctx context.Context, group *errgroup.Group, fn func(ctx context.Context) error) {
+	group.Go(func() error {
+		return cycle.Run(ctx, fn)
+	})
+}
+
+// Run runs the specified function.
```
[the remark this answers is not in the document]
egonelbre: Should be better now?

internal/sync2/cycle.go
egonelbre: added a close here

internal/sync2/cycle.go
```diff
@@ -0,0 +1,166 @@
...
+func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error) error {
+	cycle.initialize()
+	defer close(cycle.stop)
```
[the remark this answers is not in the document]
thepaul: I'd say it's acceptable to put the onus on the caller to coordinate so things are done in the right order, but this way should be fine too.

PR #1270, opened 2019-02-07
Routing table tests

jenlij:
This PR includes a new package called testrouting, which implements a very algorithmically slow but hopefully easy-to-keep-operationally-correct in-memory routing table. The routing table also supports writing out its current structure as a DOT graph for visualization. testrouting is primarily meant to help in coming up with generic routing table integration tests.

This PR also includes a new routing table integration test suite that runs against all current routing table implementations. Our existing routing table passes a lot of the tests, but not all of them, still debugging why. I have confirmed the tests should pass with the visualization graphs though.

Fresh PR on https://github.com/storj/storj/pull/981


pkg/dht/dht.go
```diff
@@ -19,6 +21,23 @@ type DHT interface {
 	Seen() []*pb.Node
 }
 
+// RoutingTable contains information on nodes we have locally
+type RoutingTable interface {
```
egonelbre: Not quite sure why is this being added back?

pkg/kademlia/testrouting/testrouting.go
```diff
@@ -0,0 +1,308 @@
...
+	var maxDepth int
+	t.walkLeaves(t.makeTree(), func(b *bucket) {
+		if b.depth > maxDepth {
+			maxDepth = b.depth
+		}
+	})
+	return maxDepth, nil
+}
+
+// GetNodes implements the dht.RoutingTable interface
```
egonelbre: X implements Y doesn't explain much in documentation. That could be an extra comment, but the documentation should describe the thing it does.

pkg/kademlia/routing_integration_test.go
```diff
@@ -0,0 +1,616 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package kademlia
```
egonelbre: Le'ts mark tests as `kademila_test`, such tests are less likely to fail

pkg/kademlia/replacement_cache_test.go
```diff
@@ -33,3 +36,25 @@ func TestAddToReplacementCache(t *testing.T) {
...
 }
+
+func TestRemoveFromReplacementCache(t *testing.T) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+	rt := createRoutingTableWith(storj.NodeID{244, 255}, routingTableOpts{cacheSize: 3})
+	defer ctx.Check(rt.Close)
+
+	kadBucketID2 := bucketID{127, 255}
+	node2 := teststorj.MockNode(string([]byte{100, 255}))
```
egonelbre: Use `node2 := storj.NodeID{100, 255}`

pkg/kademlia/routing_integration_helpers_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/pkg/dht"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+func ID(hexID string) (rv storj.NodeID) {
```
egonelbre: I do not think these make sense as a kademlia package interface, isolate these to tests.

pkg/kademlia/routing_integration_test.go
jenlij: this and moving dht.RoutingTable to a test file seem to create an import cycle. Any advice for that?

pkg/kademlia/replacement_cache_test.go
jenlij: i need the actual node here though

egonelbre: ah, sorry... mistakenly read it was only creating node id

PR #1272, opened 2019-02-07
Adds tests to discovery/graveyard

dylanlott:
Adds tests to discovery/service#searchGraveyard. 

We remove a node from the cache, then wait for the discovery service to find it again and test that it is entered back into the cache. Since it's online, it should be added back because 1) it responded and 2) it exists in the kademlia#Seen map.


pkg/discovery/service_test.go
```diff
@@ -28,3 +28,34 @@ func TestCache_Refresh(t *testing.T) {
 		}
 	})
 }
+
+func TestCache_Graveyard(t *testing.T) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	planet, err := testplanet.New(t, 1, 8, 0)
```
egonelbre: Test against both backends postgres and sqlite.

pkg/discovery/service_test.go
```diff
@@ -28,3 +28,34 @@ func TestCache_Refresh(t *testing.T) {
...
+	k := satellite.Kademlia.Service
+	k.WaitForBootstrap() // redundant, but leaving here to be clear
+
+	seen := k.Seen()
+	assert.NotNil(t, seen)
+
+	err = satellite.Overlay.Service.Delete(ctx, offline)
+	assert.NoError(t, err)
+
+	time.Sleep(5 * time.Second)
```
egonelbre: Note #1275 adds possibility to force a particular system to refresh, so you don't have to sleep anymore.

dylanlott: Ah, okay when this was written that hadn't been merged yet. I'll go update this to use the sync2 cycle stuff

pkg/discovery/service_test.go
dylanlott: Done

pkg/discovery/service_test.go
dylanlott: Updated 👍

PR #1275, opened 2019-02-08
Ensure everyone sees everyone else

egonelbre:


mniewrzal:
Just to summarize changes. Let me know if I'm right:
* connect each node with rest of nodes when testplanet is starting to avoid sleeps and flaky tests 
* use sync2.Cycle to unify refreshing around Kademlia

egonelbre:
Yes

PR #1277, opened 2019-02-08
Fix randomness in stopping nodes in TestGetObjectStream

kaloyan-raev:


pkg/metainfo/kvmetainfo/objects_test.go
```diff
@@ -153,8 +153,8 @@ func TestGetObjectStream(t *testing.T) {
 		// Stop randomly half of the storage nodes and remove them from satellite's overlay cache
 		perm := mathrand.Perm(len(planet.StorageNodes))
 		for i := 0; i < len(perm)/2; i++ {
```
egonelbre: `for _, i := range perm[:(len(perm)+1)/2 {`

kaloyan-raev: ok, this is less error prone. But without the `+1`, because this way it may stop more than half the nodes if total number of nodes is odd.

egonelbre: ah, right nevermind... For some reason I thought it was a good idea to round upwards... yes without `+1` is the correct one.

PR #1278, opened 2019-02-08
flaky psserver

egonelbre:


pkg/ranger/file.go
```diff
@@ -49,32 +51,35 @@ func (rr *fileRanger) Range(ctx context.Context, offset, length int64) (io.ReadC
...
 }
 
 // Close closes the underlying file.
-func (reader *FileReader) Close() error {
+func (reader *fileReader) Close() error {
 	return reader.file.Close()
 }
+
+func jitter() {
+	time.Sleep(time.Duration(rand.Intn(3)*100) * time.Microsecond * 0)
```
crawter: do we need to multiply duration on 0?

egonelbre: This is WIP and this is debug code.

PR #1279, opened 2019-02-08
 Disable Checker subsystem in tests

mniewrzal:


pkg/datarepair/checker/checker_test.go
```diff
@@ -99,18 +102,19 @@ func TestOfflineNodes(t *testing.T) {
...
 
 func TestIdentifyIrreparableSegments(t *testing.T) {
-	// TODO note satellite's: own sub-systems need to be disabled
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 3, UplinkCount: 0,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		checker := planet.Satellites[0].Repair.Checker
+		checker.Interval.Stop()
+
 		time.Sleep(2 * time.Second)
```
egonelbre: Remove sleeps as well.

mniewrzal: Ahh :)

pkg/datarepair/checker/checker.go
```diff
@@ -32,71 +34,63 @@ type Config struct {
...
 }
 
 // Run the checker loop
-func (c *checker) Run(ctx context.Context) (err error) {
+func (checker *Checker) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	for {
-		err = c.IdentifyInjuredSegments(ctx)
+	var group errgroup.Group
```
egonelbre: You don't need the group here. You can use `return checker.Interval.Run`. Group is only necessary when it's starting multiple cycles.

pkg/datarepair/checker/checker.go
```diff
@@ -32,71 +33,60 @@ type Config struct {
...
+type Checker struct {
 	statdb      statdb.DB
 	pointerdb   *pointerdb.Service
 	repairQueue queue.RepairQueue
 	overlay     pb.OverlayServer
 	irrdb       irreparable.DB
 	limit       int
 	logger      *zap.Logger
-	ticker      *time.Ticker
+	Interval    sync2.Cycle
```
egonelbre: Wondering whether there's a better name for this, like `Identify` or `Check`? But `Interval` is fine as well.

PR #1280, opened 2019-02-08
Refactor how bandwidth allocations are stored on storage nodes

aligeti:
https://storjlabs.atlassian.net/browse/V3-1173


pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -92,7 +92,7 @@ func (db *DB) init() (err error) {
 		return err
 	}
 
-	_, err = tx.Exec("CREATE TABLE IF NOT EXISTS `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB);")
+	_, err = tx.Exec("CREATE TABLE IF NOT EXISTS `bandwidth_agreements` (`satellite` BLOB, `uplink` BLOB, `serialnum` TEXT, `total` INT(10), `maxsize` INT(10), `createdunixsec` INT(10), `expirationunixsec` INT(10), `action` TEXT, `daystartdateunixsec` INT(10), `agreement` BLOB, `signature` BLOB);")
```
wthorp: 'daystartdateunixsec' etc are hard to read. Could we use underscores to separate words?  Googling indicates Postgres's identifier limit is 63 characters and Sqlite doesn't have a hard limit.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -243,7 +244,11 @@ func (db *DB) GetBandwidthAllocations() (map[storj.NodeID][]*Agreement, error) {
 		rbaBytes := []byte{}
 		agreement := &Agreement{}
 		var satellite []byte
-		err := rows.Scan(&satellite, &rbaBytes, &agreement.Signature)
+		var uplink []byte
+		var serialnum string
+		var total, maxsize, createdUnixSec, expirationUnixSec, daystartdateUnixSec int64
+		var action string
+		err := rows.Scan(&satellite, &uplink, &serialnum, &total, &maxsize, &createdUnixSec, &expirationUnixSec, &action, &daystartdateUnixSec, &rbaBytes, &agreement.Signature)
```
wthorp: Don't use 'SELECT * FROM' on line 232; specify only the fields you actually want to return.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -178,8 +173,14 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.RenterBandwidthAllocation) error {
 	// We begin extracting the satellite_id
 	// The satellite id can be used to sort the bandwidth agreements
 	// If the agreements are sorted we can send them in bulk streams to the satellite
-	_, err = db.DB.Exec(`INSERT INTO bandwidth_agreements (satellite, agreement, signature) VALUES (?, ?, ?)`,
-		rba.PayerAllocation.SatelliteId.Bytes(), rbaBytes, rba.GetSignature())
+	t := time.Now()
+	daystartdateunixsec := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location()).Unix()
```
wthorp: I know this was already here, but doesn't rounding off the time to a day seems a little weird?

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -12,13 +13,16 @@ import (
...
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testidentity"
+	"storj.io/storj/pkg/bwagreement/testbwagreement"
+	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/pb"
-	"storj.io/storj/pkg/storj"
 )
 
-const concurrency = 10
+const concurrency = 1
```
wthorp: I'm pretty sure we don't want to test with concurrency of one.  If changing this was the only way to make the test work, there's probably something wrong.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,93 @@ func (db *DB) Migration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				// some databases may have already this done, although the version may not match
+				Description: "Optimize storagenode tables",
+				Version:     1,
+				Action: migrate.Func(func(log *zap.Logger, db migrate.DB, tx *sql.Tx) error {
+					v1sql := migrate.SQL{
+						`ALTER TABLE bandwidth_agreements RENAME TO bandwidth_agreements_old`,
```
mniewrzal: No need to create new table. Just add columns to existing one. Something like this:
```
ALTER TABLE bandwidth_agreements ADD uplink BLOB;
ALTER TABLE bandwidth_agreements ADD serialnum TEXT NOT NULL DEFAULT "";
ALTER TABLE bandwidth_agreements ADD total INT(10);
ALTER TABLE bandwidth_agreements ADD maxsize INT(10);
...
```
Then iterate over `bandwidth_agreements` and just update new columns.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,93 @@ func (db *DB) Migration() *migrate.Migration {
...
+									daystartdateunixsec INT(10)
+								)`,
+						`INSERT INTO bandwidth_agreements (satellite, agreement, signature)
+								 SELECT satellite, agreement, signature FROM bandwidth_agreements_old`,
+					}
+					_ = v1sql.Run(log, db, tx)
+
+					// iterate through the table and fill
+					err := func() error {
+						rows, err := tx.Query(`SELECT * FROM  bandwidth_agreements ORDER BY satellite`)
```
mniewrzal: You don't need to select all columns, just need `agreement` - `SELECT agreement FROM ...`

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -233,7 +326,11 @@ func (db *DB) GetBandwidthAllocations() (map[storj.NodeID][]*Agreement, error) {
 		rbaBytes := []byte{}
 		agreement := &Agreement{}
 		var satellite []byte
-		err := rows.Scan(&satellite, &rbaBytes, &agreement.Signature)
+		var uplink []byte
+		var serialnum string
+		var total, maxsize, createdUnixSec, expirationUnixSec, daystartdateUnixSec int64
+		var action string
+		err := rows.Scan(&satellite, &rbaBytes, &agreement.Signature, &uplink, &serialnum, &total, &maxsize, &createdUnixSec, &expirationUnixSec, &action, &daystartdateUnixSec)
```
mniewrzal: You don't have to scan all columns. You can limit number of columns in ```rows, err := db.DB.Query(`SELECT * FROM bandwidth_agreements ORDER BY satellite`)``` above.

pkg/piecestore/psserver/psdb/testdata/sqlite.v0.sql
```diff
@@ -5,6 +5,5 @@ CREATE INDEX idx_ttl_expires ON ttl (expires);
 
 -- NEW DATA --
 
-INSERT INTO ttl VALUES(1,2,3,4);
```
mniewrzal: you shoudn't remove this. This value should be in DB before your migration. It shouldn't be in DB after your migration.

pkg/piecestore/psserver/psdb/testdata/sqlite.v1.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));
+CREATE TABLE `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB, `uplink` BLOB, `serialnum`	TEXT NOT NULL, `total` INT(10), `maxsize` INT(10), `createdunixsec`	INT(10), `expirationunixsec` INT(10), `action` TEXT, `daystartdateunixsec` INT(10));
+CREATE INDEX idx_ttl_expires ON ttl (expires);
+
+-- NEW DATA --
```
mniewrzal: `NEW DATA` is for inserting new data into tested db. Not needed in this case because we are not adding new table. In this file you need to describe DB state after migration.

pkg/piecestore/psserver/psdb/psdb.go
aligeti: done

pkg/piecestore/psserver/psdb/psdb.go
aligeti: done

pkg/piecestore/psserver/psdb/psdb.go
aligeti: Done

pkg/piecestore/psserver/psdb/testdata/sqlite.v0.sql
aligeti: done

pkg/piecestore/psserver/psdb/testdata/sqlite.v1.sql
aligeti: done

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -147,7 +147,18 @@ func TestHappyPath(t *testing.T) {
 		bandwidthAllocation("signed by test", nodeIDAB, 3),
 	}
 
+	type BWUSAGE struct {
```
wthorp: use more standardized naming here, either BWUsage or bwUsage

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -147,7 +147,18 @@ func TestHappyPath(t *testing.T) {
...
+		size    int64
+		timenow time.Time
+	}
+
+	bwtests := []BWUSAGE{
```
wthorp: This whole BWUSAGE thing looks pretty bogus.  I know you didn't write it though.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,84 @@ func (db *DB) Migration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				// some databases may have already this done, although the version may not match
+				Description: "Optimize storagenode tables",
```
mniewrzal: I would change description to something more meaningful like `Extending bandwidth_agreements table`

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,84 @@ func (db *DB) Migration() *migrate.Migration {
...
+						`ALTER TABLE bandwidth_agreements ADD COLUMN daystart_utc_sec INT(10)`,
+					}
+					err := v1sql.Run(log, db, tx)
+					if err != nil {
+						return err
+					}
+
+					// iterate through the table and fill
+					err = func() error {
+						rows, err := tx.Query(`SELECT agreement, signature FROM bandwidth_agreements ORDER BY satellite`)
```
mniewrzal: Is this order by necessary?

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,84 @@ func (db *DB) Migration() *migrate.Migration {
...
+							if err != nil {
+								return err
+							}
+						}
+						return rows.Err()
+					}()
+					if err != nil {
+						return err
+					}
+					_, err = tx.Exec(`
```
mniewrzal: unwrap lines

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -168,8 +246,14 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.RenterBandwidthAllocation) error {
 	// We begin extracting the satellite_id
 	// The satellite id can be used to sort the bandwidth agreements
 	// If the agreements are sorted we can send them in bulk streams to the satellite
-	_, err = db.DB.Exec(`INSERT INTO bandwidth_agreements (satellite, agreement, signature) VALUES (?, ?, ?)`,
-		rba.PayerAllocation.SatelliteId.Bytes(), rbaBytes, rba.GetSignature())
+	t := time.Now()
+	startofthedayunixsec := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location()).Unix()
```
mniewrzal: Shouldn't this line be the same like in migration?

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -233,7 +317,11 @@ func (db *DB) GetBandwidthAllocations() (map[storj.NodeID][]*Agreement, error) {
 		rbaBytes := []byte{}
 		agreement := &Agreement{}
 		var satellite []byte
-		err := rows.Scan(&satellite, &rbaBytes, &agreement.Signature)
+		var uplink []byte
+		var serialnum string
+		var total, max_size, createdUnixSec, expirationUnixSec, startofthedayUnixSec int64
+		var action string
+		err := rows.Scan(&satellite, &rbaBytes, &agreement.Signature, &uplink, &serialnum, &total, &max_size, &createdUnixSec, &expirationUnixSec, &action, &startofthedayUnixSec)
```
mniewrzal: Limit columns in SELECT to avoid many unused variables.

pkg/piecestore/psserver/psdb/psdb_test.go
aligeti: Done

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated the table description

pkg/piecestore/psserver/psdb/psdb.go
aligeti: Thnx. yes it needs to be changed to UTC location.

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: actually no. .UNIX() translates time with respect to UTC.

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,82 @@ func (db *DB) Migration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				// some databases may have already this done, although the version may not match
```
egonelbre: This comment is not valid.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,82 @@ func (db *DB) Migration() *migrate.Migration {
...
+				Description: "Extending bandwidth_agreements table and drop bwusagetbl",
+				Version:     1,
```
egonelbre: Maybe if it's 3 different versions, the code will be clearer... but it can also be a single func. Maybe we need something that applies a list of funcs or SQL-s combined?

We can solve this later.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,82 @@ func (db *DB) Migration() *migrate.Migration {
...
+				Action: migrate.Func(func(log *zap.Logger, db migrate.DB, tx *sql.Tx) error {
+					v1sql := migrate.SQL{
+						`ALTER TABLE bandwidth_agreements ADD COLUMN uplink BLOB`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN serial_num TEXT NOT NULL DEFAULT ""`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN total INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN max_size INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN created_utc_sec INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN expiration_utc_sec INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN action TEXT`,
```
egonelbre: Why is this as text instead of a number?

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -103,6 +103,82 @@ func (db *DB) Migration() *migrate.Migration {
...
 				},
 			},
+			{
+				// some databases may have already this done, although the version may not match
+				Description: "Extending bandwidth_agreements table and drop bwusagetbl",
+				Version:     1,
+				Action: migrate.Func(func(log *zap.Logger, db migrate.DB, tx *sql.Tx) error {
+					v1sql := migrate.SQL{
+						`ALTER TABLE bandwidth_agreements ADD COLUMN uplink BLOB`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN serial_num TEXT NOT NULL DEFAULT ""`,
```
egonelbre: this should be a blob

pkg/piecestore/psserver/psdb/psdb.go
aligeti: Thnx. Copy paste error :-). Updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: received as PUT/GET/GET_AUDIT/GET_REPAIR/PUT_REPAIR by the storage node from uplink/payer

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated to blob.

pkg/piecestore/psserver/psdb/psdb.go
mniewrzal: PayerBandwidthAllocation.Action is an in32 (BandwidthAction https://github.com/storj/storj/blob/c3d3f41d30f7eca3ecbe449985d67e817257dae4/pkg/pb/piecestore.pb.go#L28)

aligeti: Changed it to INT(10).

PR #1286, opened 2019-02-09
pkg/transport: require tls configuration for dialing

jtolds:
this PR:
 * separates TLS options from server options (because we need them for dialing too)
 * stops creating transports in multiple places
 * ensures that we actually check revocation, whitelists, certificate signing, etc, for all connections.

the names of packages are probably bad, but something like this PR is high priority


jtolds:
that's a good name. sadly, pkg/peertls/tlsopt depends on pkg/identity which depends on pkg/peertls, so i can't merge pkg/peertls/tlsopt with pkg/peertls without some deeper changes to break that import cycle

PR #1297, opened 2019-02-11
Repair checker is checking the same 1000 elements all the time

wthorp:
https://storjlabs.atlassian.net/browse/V3-925

Changes:
Modified test to recreate error
Removed the limit from Checker's IdentifyInjuredSegments call


pkg/datarepair/checker/checker.go
```diff
@@ -54,30 +53,35 @@ func NewChecker(pointerdb *pointerdb.Service, sdb statdb.DB, repairQueue queue.R
...
 		}
-		return nil
+		return err
 	})
 }
 
-// Close closes resources
-func (checker *Checker) Close() error { return nil }
+// Pause suspends the Checker loop
+func (checker *Checker) Pause() {
```
egonelbre: I think this won't be necessary anymore, since the loop itself has been exposed.

pkg/datarepair/checker/checker.go
```diff
@@ -86,11 +90,7 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 	err = checker.pointerdb.Iterate("", "", true, false,
 		func(it storage.Iterator) error {
 			var item storage.ListItem
-			lim := checker.limit
-			if lim <= 0 || lim > storage.LookupLimit {
-				lim = storage.LookupLimit
-			}
-			for ; lim > 0 && it.Next(&item); lim-- {
+			for it.Next(&item) {
```
egonelbre: Iterate may lock the whole database at the moment.

wthorp: @egonelbre I've been looking for where a DB might lock, but can't find it in our code.  Can you point me to this?

egonelbre: https://github.com/storj/storj/blob/master/satellite/satellitedb/locked.go

egonelbre: https://github.com/storj/storj/blob/master/storage/boltdb/client.go#L107

wthorp: I read it as View() is ok... `Bolt allows only one read-write transaction at a time but allows as many read-only transactions as you want at a time.` [https://github.com/boltdb/bolt#transactions]

jenlij:
is this also happening for tally and rollup in accounting?

pkg/datarepair/checker/checker.go
egonelbre: I may have misremembered this or it may has changed at some point.

wthorp: It seems that there may be some issues in Bolt when updating inside a iterator, but that doesn't seem to apply anywhere in our codebase at the moment

pkg/datarepair/checker/checker_test.go
```diff
@@ -21,57 +23,32 @@ func TestIdentifyInjuredSegments(t *testing.T) {
...
-		assert.Equal(t, len(expectedLostPieces), len(injuredSegment.LostPieces))
+		numValidNode := int32(len(planet.StorageNodes))
+		assert.Equal(t, "b", injuredSegment.Path)
+		assert.Equal(t, len(planet.StorageNodes), len(injuredSegment.LostPieces))
 		for _, lostPiece := range injuredSegment.LostPieces {
-			if !expectedLostPieces[lostPiece] {
-				t.Error("should be lost: ", lostPiece)
-			}
+			assert.True(t, lostPiece >= numValidNode, fmt.Sprintf("%d >= %d \n", lostPiece, numValidNode))
+			assert.True(t, lostPiece < numValidNode*2, fmt.Sprintf("%d < %d \n", lostPiece, numValidNode*2))
```
mniewrzal: Can you add comments (or error message) why such conditions? Its a little bit hard to understand with just numbers in fail info.

pkg/datarepair/checker/checker_test.go
```diff
@@ -177,3 +154,42 @@ func TestIdentifyIrreparableSegments(t *testing.T) {
 		assert.True(t, firstRepair < remoteSegmentInfo.RepairUnixSec)
 	})
 }
+
+func makePointer(t *testing.T, planet *testplanet.Planet, peiceID string, createLost bool) {
```
mniewrzal: `pieceID`

pkg/datarepair/checker/checker.go
```diff
@@ -40,9 +40,8 @@ type Checker struct {
 	repairQueue queue.RepairQueue
 	overlay     pb.OverlayServer
 	irrdb       irreparable.DB
-	limit       int
 	logger      *zap.Logger
-	Interval    sync2.Cycle
+	Loop        *sync2.Cycle
```
egonelbre: Any reason we are not keep the memory local?

storage/common.go
```diff
@@ -101,6 +101,20 @@ type Iterator interface {
 	Next(item *ListItem) bool
 }
 
+// BatchIterator iterates over a sequence of ListItems
```
egonelbre: Nothing seems to be using these

pkg/datarepair/checker/checker_test.go
wthorp: done

PR #1298, opened 2019-02-12
Implemented Notification for API key deletion result.

Barterio:


satellite/console/consoleweb/consoleql/mutation.go
```diff
@@ -329,32 +329,39 @@ func rootMutation(service *console.Service, types Types) *graphql.Object {
...
-					}
-
-					key, err := service.GetAPIKeyInfo(p.Context, *id)
-					if err != nil {
-						return nil, err
+					paramKeysID, _ := p.Args[FieldID].([]interface{})
+
+					var keyIds []uuid.UUID
+					var keys []console.APIKeyInfo
+					for _, id := range paramKeysID {
```
egonelbre: Should we return the error for all of the keys at the same time rather than one-by-one? Although here probably doesn't matter that much.

PR #1299, opened 2019-02-12
Storagenode migrations

mniewrzal:
This change:
* creates initial migration for psdb
* add test mechanism to validate migration to every version
* fix few small issues in versions.go and context.go


pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,23 @@ func OpenInMemory() (db *DB, err error) {
...
-	}
-
-	_, err = tx.Exec("CREATE TABLE IF NOT EXISTS `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB);")
-	if err != nil {
-		return err
+// Migration define piecestore DB migration
+func Migration() *migrate.Migration {
+	migration := &migrate.Migration{
+		Table: "sn_versions",
+		OnCreate: migrate.SQL{
```
egonelbre: Note, `OnCreate` is more related to upgrading from a version where we don't know the exact version yet. I probably chose the `OnCreate` name poorly. It's separate from all other steps, because it needs to figure out what we need to skip and do, when the version already exists.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,23 @@ func OpenInMemory() (db *DB, err error) {
...
+			"CREATE TABLE IF NOT EXISTS `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));",
+			"CREATE TABLE IF NOT EXISTS `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB);",
+			"CREATE INDEX IF NOT EXISTS idx_ttl_expires ON ttl (expires);",
+			"CREATE TABLE IF NOT EXISTS `bwusagetbl` (`size` INT(10), `daystartdate` INT(10), `dayenddate` INT(10));",
+			"PRAGMA journal_mode = WAL",
```
egonelbre: journal_mode should be an argument to sqlite, this probably doesn't work as well as I intended.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,23 @@ func OpenInMemory() (db *DB, err error) {
...
-
-	_, err = tx.Exec("CREATE TABLE IF NOT EXISTS `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB);")
-	if err != nil {
-		return err
+// Migration define piecestore DB migration
+func Migration() *migrate.Migration {
+	migration := &migrate.Migration{
+		Table: "sn_versions",
+		OnCreate: migrate.SQL{
+			"CREATE TABLE IF NOT EXISTS `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));",
```
egonelbre: These all should be `CREATE TABLE`, because we want exact behavior... it's easy to miss migration steps.

storagenode/storagenodedb/database.go
```diff
@@ -27,13 +28,14 @@ type Config struct {
 
 // DB contains access to different database tables
 type DB struct {
+	logger   *zap.Logger
```
egonelbre: use log as the name, that way the logging lines look nicer.

pkg/piecestore/psserver/psdb/psdb.go
mniewrzal: Do you think about moving inital table creation to `Steps`? Then we will start always from version 1.

pkg/piecestore/psserver/psdb/psdb.go
mniewrzal: what will be valid parameter? `file:somefile.db;journal_mode=WAL`?

storagenode/storagenodedb/database.go
mniewrzal: done

internal/migrate/versions.go
```diff
@@ -43,6 +43,7 @@ Scenarios it doesn't handle properly.
 type Migration struct {
 	Table string
 	Steps []*Step
+	Now   string // can be used to override version timestamp
```
egonelbre: `Now func() time.Time`

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,29 @@ func OpenInMemory() (db *DB, err error) {
...
+		Steps: []*migrate.Step{
+			{
+				Description: "initialize database",
+				Version:     0,
+				Action: migrate.SQL{
+					"CREATE TABLE `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));",
+					"CREATE TABLE `bandwidth_agreements` (`satellite` BLOB, `agreement` BLOB, `signature` BLOB);",
+					"CREATE INDEX idx_ttl_expires ON ttl (expires);",
+					"CREATE TABLE `bwusagetbl` (`size` INT(10), `daystartdate` INT(10), `dayenddate` INT(10));",
+					"PRAGMA journal_mode = WAL",
```
egonelbre: remove the pragma, that should be executed during database creation... cannot remember how though.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,29 @@ func OpenInMemory() (db *DB, err error) {
...
+// Migration define piecestore DB migration
+func Migration() *migrate.Migration {
+	migration := &migrate.Migration{
+		Table: "sn_versions",
+		Steps: []*migrate.Step{
+			{
+				Description: "initialize database",
+				Version:     0,
+				Action: migrate.SQL{
+					"CREATE TABLE `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));",
```
egonelbre: These statements should work already with an existing database because we already may have databases at step 0.
So use `CREATE TABLE IF NOT EXISTS`...

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -274,6 +276,52 @@ func TestBandwidthUsage(t *testing.T) {
...
+	latestVersion := 0
+	// TODO support missing intermediate versions
+	for _, step := range migration.Steps {
+		if latestVersion < step.Version {
+			latestVersion = step.Version
+		}
+	}
+
+	for version := 0; version <= latestVersion; version++ {
+		t.Run("migration-v"+strconv.Itoa(version), func(t *testing.T) {
```
egonelbre: One test should be about starting from the current storage node state, not from zero.

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -274,6 +276,52 @@ func TestBandwidthUsage(t *testing.T) {
...
+			dbpath := ctx.File("psdb.db")
+			db, err := psdb.Open(dbpath)
+			require.NoError(t, err)
+			defer ctx.Check(db.Close)
+
+			testedMigration := migration.TargetVersion(version)
+			err = testedMigration.Run(zap.NewNop(), db)
+			require.NoError(t, err)
+
+			out, err := exec.Command("sqlite3", dbpath, ".dump").Output()
```
egonelbre: Maybe there's a way to do this part with queries.

cmd/storagenode/main.go
```diff
@@ -151,7 +151,7 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Error("Failed to initialize telemetry batcher: ", err)
 	}
 
-	db, err := storagenodedb.New(databaseConfig(runCfg.Config))
+	db, err := storagenodedb.New(log, databaseConfig(runCfg.Config))
```
egonelbre: use `log.Named("db")`, it keeps clear the context

cmd/storagenode/main.go
```diff
@@ -245,7 +245,7 @@ func cmdDiag(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	db, err := storagenodedb.New(databaseConfig(runCfg.Config))
+	db, err := storagenodedb.New(zap.L(), databaseConfig(runCfg.Config))
```
egonelbre: use `log.Named("db")`, it keeps clear the context

internal/testplanet/planet.go
```diff
@@ -494,7 +494,7 @@ func (planet *Planet) newStorageNodes(count int) ([]*storagenode.Peer, error) {
 		if planet.config.Reconfigure.NewStorageNodeDB != nil {
 			db, err = planet.config.Reconfigure.NewStorageNodeDB(i)
 		} else {
-			db, err = storagenodedb.NewInMemory(storageDir)
+			db, err = storagenodedb.NewInMemory(log, storageDir)
```
egonelbre: use `log.Named("db")`, it keeps clear the context

pkg/piecestore/psserver/psdb/testdata/db.v0.sql
```diff
@@ -0,0 +1,9 @@
+PRAGMA foreign_keys=OFF;
+BEGIN TRANSACTION;
```
egonelbre: Ah, also forgot that we want each table to contain some data. Otherwise it will be easy to miss migrations that need to transform data.

pkg/piecestore/psserver/psdb/psdb_test.go
mniewrzal: For a moment was looking for that but most of the results give `.dump`. I will search once again before merging.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -48,16 +48,14 @@ func Open(DBPath string) (db *DB, err error) {
 		return nil, err
 	}
 
+	// TODO add "PRAGMA journal_mode = WAL" somehow
 	sqlite, err := sql.Open("sqlite3", fmt.Sprintf("file:%s", DBPath))
```
egonelbre: https://github.com/mattn/go-sqlite3#connection-string, maybe `?_journal=wal`

mniewrzal: that might work

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,28 @@ func OpenInMemory() (db *DB, err error) {
...
-		return err
+// Migration define piecestore DB migration
+func (db *DB) Migration() *migrate.Migration {
+	migration := &migrate.Migration{
+		Table: "versions",
+		Steps: []*migrate.Step{
+			{
+				Description: "Initial setup",
+				Version:     0,
+				Action: migrate.SQL{
```
aligeti: If there are more than two actions are there what is the preferred way to implement?  Should we do it as separate steps or two actions in action func()?

egonelbre: General rule can be thought of as 1 version per commit... of course sometimes it might be easier to separate them into different versions.

pkg/piecestore/psserver/psdb/migrate_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// snapshot represents clean DB state
+	snapshots.Add(&dbschema.Snapshot{
+		Version: -1,
+		Schema:  &dbschema.Schema{},
+		Script:  "",
+	})
+
+	// find all sql files
+	matches, err := filepath.Glob("testdata/db.*")
```
egonelbre: lets use `sqlite.*` to make clear what the sql applies to

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -72,50 +69,28 @@ func OpenInMemory() (db *DB, err error) {
...
+					"CREATE TABLE IF NOT EXISTS `ttl` (`id` BLOB UNIQUE, `created` INT(10), `expires` INT(10), `size` INT(10));",
```
egonelbre: Let's split columns to different lines, it makes clearer what columns exist.

pkg/piecestore/psserver/psdb/migrate_test.go
```diff
@@ -0,0 +1,133 @@
...
+		return ""
+	}
+	return tokens[1]
+}
+
+func TestMigrate(t *testing.T) {
+	snapshots, err := loadSnapshots()
+	require.NoError(t, err)
+
+	for _, base := range snapshots.List {
```
egonelbre: We can have two base versions here only.

1. empty database
2. version 0

Rest of the migrations/versions will be implicitly tested by migrations themselves.

mniewrzal: I'm not sure if following, version 0 is from sqlite.v0.sql, empty db is added in `loadSnapshots` as -1

egonelbre: It's more of adding a check for future, or in other words when someone adds version 2.

mniewrzal: ah analogy to https://github.com/storj/storj/blob/507fd817e4897a732d61b0e53edf38b1a45996b6/satellite/satellitedb/migrate_postgres_test.go#L102 ?

egonelbre: yes

mniewrzal: done

PR #1301, opened 2019-02-13
Satellite migrations

egonelbre:


satellite/satellitedb/migrate.go
```diff
@@ -0,0 +1,328 @@
...
+							if err := rows.Scan(&data); err != nil {
+								return ErrMigrate.Wrap(errs.Combine(err, rows.Close()))
+							}
+
+							var rba pb.RenterBandwidthAllocation
+							if err := proto.Unmarshal(data, &rba); err != nil {
+								return ErrMigrate.Wrap(errs.Combine(err, rows.Close()))
+							}
+
+							_, err := tx.Exec(`UPDATE bwagreements SET uplink_id = ?`, rba.PayerAllocation.UplinkId.Bytes())
```
jtolds: oh for real? data is actually just an uplink id? ha i didn't pick up on that. i thought data was the bwagreement or something

egonelbre: Well, it was bandwidth allocation. But now it’s just uplink id, not sure why it was changed that way though.

jtolds: oh oh i misread

satellite/satellitedb/migrate_postgres_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"go.uber.org/zap/zaptest"
+
+	"storj.io/storj/internal/dbutil/dbschema"
+	"storj.io/storj/internal/dbutil/pgutil"
+	"storj.io/storj/satellite/satellitedb"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+// VersionSchema defines a versioned schema
+type VersionSchema struct {
```
mniewrzal: This could be moved to `dbschema` package.

egonelbre: Will pull into a separate PR.

satellite/satellitedb/migrate_postgres_test.go
```diff
@@ -0,0 +1,189 @@
...
+func TestMigratePostgres(t *testing.T) {
+	if *satellitedbtest.TestPostgres == "" {
+		t.Skip("Postgres flag missing, example: -postgres-test-db=" + satellitedbtest.DefaultPostgresConn)
+	}
+
+	versions, err := loadVersions(*satellitedbtest.TestPostgres)
+	require.NoError(t, err)
+
+	for _, base := range versions {
+		base := base
```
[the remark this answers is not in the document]
egonelbre: removed

satellite/satellitedb/migrate_postgres_test.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		schema, err := pgutil.LoadSchemaFromSQL(connstr, string(data))
+		if err != nil {
+			return nil, err
+		}
+
+		schema.DropTable("versions")
```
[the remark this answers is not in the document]
egonelbre: The problem is that the sql scripts from dbx don't contain the version table and getting the different state dropped from database is more annoying than to use the existing schema that dbx produces.

mniewrzal:
Sorry, accident...

satellite/satellitedb/migrate_postgres_test.go
egonelbre: https://github.com/storj/storj/pull/1312

satellite/satellitedb/migrate_postgres_test.go
```diff
@@ -0,0 +1,172 @@
...
+	if *satellitedbtest.TestPostgres == "" {
+		t.Skip("Postgres flag missing, example: -postgres-test-db=" + satellitedbtest.DefaultPostgresConn)
+	}
+
+	snapshots, err := loadSnapshots(*satellitedbtest.TestPostgres)
+	require.NoError(t, err)
+
+	for _, base := range snapshots.List {
+		// versions 0 to 4 can be a starting point
+		if base.Version < 0 || 4 < base.Version {
```
mniewrzal: Maybe a constant for this? As far as I understand it will increase over time.

satellite/satellitedb/migrate.go
```diff
@@ -0,0 +1,330 @@
...
+	case "postgres":
+		migration := db.PostgresMigration()
+		return migration.Run(db.log.Named("migrate"), db.db)
+	default:
+		return migrate.Create("database", db.db)
+	}
+}
+
+// PostgresMigration returns steps needed for migrating postgres database.
+func (db *DB) PostgresMigration() *migrate.Migration {
```
mniewrzal: Is this using something from satellitedb.DB interface? Maybe can be just `satellitedb.PostgresMigration()`?

satellite/satellitedb/migrate.go
```diff
@@ -0,0 +1,330 @@
...
+	return &migrate.Migration{
+		Table: "versions",
+		Steps: []*migrate.Step{
```
wthorp: This file will grow very large.  Should we start a pattern where different steps are stored as individual files?

egonelbre: The problem is that not all the steps are SQL... also that would mean we need to start embedding the sql files anyways. As for largeness, I think satellites can afford to drop older migrations and just setup a newer ones without problems.

satellite/satellitedb/migrate.go
egonelbre: It's intentionally not exposed on the database interface. `satellite.Peer` shouldn't care how the database handles migrations internally.

satellite/satellitedb/migrate_postgres_test.go
egonelbre: No, it shouldn't... actually once we upgrade, having a single snapshot should be sufficient.

satellite/satellitedb/migrate.go
wthorp: I didn't mean storing the SQL as test files, simply:

return &migrate.Migration{
		Table: "versions",
		Steps: [
		  Step1,
		  Step2,
		  ...
		]
---
with separate files for:
Step1.go
Step2.go
...

egonelbre: Ah, sure, might make sense... we'll see how the whole thing evolves. `migrate_v1.go`, `migrate_v2to4.go` etc.

satellite/satellitedb/migrate_postgres_test.go
egonelbre: note I did add a constant.

but, the reason we don't need more base versions is because we either:
1. start from clean database
2. start from a database where we know the exact version

Currently the database can be in any of those states from 0 to 4.

satellite/satellitedb/migrate_postgres_test.go
```diff
@@ -0,0 +1,179 @@
...
+			connstr := pgutil.ConnstrWithSchema(*satellitedbtest.TestPostgres, schemaName)
+
+			// create a new satellitedb connection
+			db, err := satellitedb.New(log, connstr)
+			require.NoError(t, err)
+			defer func() { require.NoError(t, db.Close()) }()
+
+			// setup our own schema to avoid collisions
+			require.NoError(t, db.CreateSchema(schemaName))
+			//defer func() { require.NoError(t, db.DropSchema(schemaName)) }()
```
mniewrzal: Why its commented?

satellite/satellitedb/migrate.go
egonelbre: Why should anyone change an existing migration step or do you mean preventing that? Whether separate files or not, just seems weird putting a single statement in a separate file; but I'm not against the idea either. What do you exactly mean by applying step manually?

`migrate_v2to4`, nah that was just so we don't have a bunch of single line files... but can be separate for consistency.

satellite/satellitedb/migrate_postgres_test.go
egonelbre: ah, debug code, my mistake

egonelbre:
> I'm still extremely leery of using a custom migration manager with this style. Has anyone previously deployed software that used database migrations in this way? (If so, and it went well over a reasonably long period of time, I'll withdraw my concern.)

Yup, about a 35+ year old project uses similar approach - custom funcs + version numbers + snapshot before migration. However not sure how long they were using it... at least 10 years. However the largest tables there were only ~40GB and had to deal with a lot more complex data layout (combination of MySQL and plain-old btree and plain data files and scripts updating).

> There are a lot of hard-learned lessons in the engineering of good migration tools. I'm very worried about trusting extremely critical customer data to an unproven one.

As for this approach, it's not meant to be the final solution. Its main focus is to handle our immediate needs. I'm fine if we find a better, simpler or more robust solution in a month, and drop this one. I think this should handle our alpha/beta needs at least. If we do need something better, we should be able to switch it out as needed.

Of course, your concerns are definitely valid... I have many of my own... my main question is how do we migrate a database that is 1TB or larger without downtime? You can see some additional comments in https://github.com/storj/storj/blob/master/internal/migrate/versions.go#L19 that this approach ignores... I'm certain there are more.

satellite/satellitedb/migrate.go
thepaul: Yes, I meant preventing changes to existing steps.

Applying a step manually might look like `psql -1f migrations/20190214-add-and-index-customers-table.sql satellitedb`.

PR #1302, opened 2019-02-13
Add mail service to the satellite

rikysya:
- Added separate mail service to the satellite
- added mail config struct
- added mail configuration to testplanet
- renamed interna/gmail pkg to internal/oauth2

!Have pkg naming issues, not sure if pkg structure is correct


cmd/storj-sim/network.go
```diff
@@ -171,6 +171,8 @@ func newNetwork(flags *Flags) (*Processes, error) {
 
 				"--server.extensions.revocation=false",
 				"--server.use-peer-ca-whitelist=false",
+
+				// TODO(yar): add mail configuration
```
egonelbre: Maybe for testplanet and satellite we could make it auto-approve with some flag i.e. avoid sending an email out? Any other alternative (like setting up custom server and client etc.) seems very messy.

rikysya: It makes sense if we turn off account activation as well, otherwise satellite can be only used with test account we created for integration testing. Other option is, as we have two auth type supported, to use plain auth as default configuration to omit all oaaut2 info here.

satellite/peer.go
```diff
@@ -352,6 +363,49 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 		peer.Accounting.Rollup = rollup.New(peer.Log.Named("rollup"), peer.DB.Accounting(), config.Rollup.Interval)
 	}
 
+	{ // setup mail
+		// TODO(yar): test multiple satellites using same OAUTH credentials
+		mailConig := config.Mail
```
egonelbre: typo

satellite/peer.go
```diff
@@ -352,6 +363,49 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
...
+
+		from, err := mail.ParseAddress(mailConig.From)
```
egonelbre: handle the parsing of address inside either in auth or service, then it would avoid the `net/mail` import here

satellite/peer.go
```diff
@@ -352,6 +363,49 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
...
+			peer.Log.Named("mail:service"),
+			m.SMTPSender{
+				From:          *from,
+				Auth:          auth,
+				ServerAddress: mailConig.SMTPServerAddress,
+			},
+			mailConig.TemplatePath,
+		)
+
+		peer.Log.Debug(peer.Mail.Service.SendForgotPasswordEmail(context.Background(),
```
egonelbre: Is this debug code?

rikysya: yes

satellite/peer.go
```diff
@@ -16,6 +18,10 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 
+	m "storj.io/storj/internal/mail"
+	"storj.io/storj/internal/mail/oauth2"
+	m1 "storj.io/storj/satellite/mail"
```
egonelbre: Don't use single letter aliases. They span the whole file and understanding `m1` in arbitrary location is difficult to understand.

internal/testplanet/planet.go
```diff
@@ -453,12 +453,12 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 		}
 
 		config.Mail.SMTPServerAddress = "smtp.gmail.com:587"
```
egonelbre: This should be defined together with rest of config.

satellite/console/consoleweb/consoleql/query_test.go
```diff
@@ -12,26 +12,41 @@ import (
...
 	"storj.io/storj/pkg/auth"
 	"storj.io/storj/satellite"
 	"storj.io/storj/satellite/console"
 	"storj.io/storj/satellite/console/consoleauth"
 	"storj.io/storj/satellite/console/consoleweb/consoleql"
+	"storj.io/storj/satellite/mailservice"
 	"storj.io/storj/satellite/satellitedb/satellitedbtest"
 )
 
+type mockSender struct{}
```
egonelbre: This is not a mock. `mailDiscarder`, `discardSender`, `stubSender`.

satellite/mailservice/service.go
```diff
@@ -0,0 +1,155 @@
...
+	ClientSecret string `help:"oauth2 app's client secret" default:""`
+	TokenURI     string `help:"uri which is used when retrieving new access token" default:""`
+}
+
+var (
+	mon = monkit.Package()
+)
+
+// SMTPSender is
+type SMTPSender interface {
```
egonelbre: `Sender`

satellite/console/mail.go
```diff
@@ -0,0 +1,75 @@
...
+	InvitationSubject = ""
+	// ForgotPasswordSubject forgot password email subject
+	ForgotPasswordSubject = ""
+)
+
+// MailTemplate is implementation of satellite/mailservice.Template interface
+type MailTemplate struct {
+	to            mail.Address
+	subject       string
+	htmlPath      string
```
egonelbre: Mail templates shouldn't care about the full path of the templates. It makes constructing them more complicated. Let `mailservice` handle that concern.

satellite/mailservice/service.go
```diff
@@ -0,0 +1,155 @@
...
+	SendEmail(msg *post.Message) error
+	FromAddress() post.Address
+}
+
+// Template defines mailservice template for SendRendered method
+type Template interface {
+	To() []post.Address
+	Subject() string
+	HTMLPath() string
```
egonelbre: `Name()` would suffice.

satellite/mailservice/service.go
```diff
@@ -0,0 +1,155 @@
...
+	} else {
+		service.log.Info("successfully send message")
+	}
+
+	return err
+}
+
+// RenderHTML renders htmltemplate content of given Template and writes it to writer
+func RenderHTML(w io.Writer, tmpl Template) error {
+	template, err := htmltemplate.ParseFiles(tmpl.HTMLPath())
```
egonelbre: Let `mailservice` handle parsing, we don't want to reparse/compile the templates on every send.

internal/testplanet/planet.go
```diff
@@ -443,9 +444,25 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 			Rollup: rollup.Config{
 				Interval: 120 * time.Second,
 			},
+			Mail: mailservice.Config{
+				SMTPServerAddress: "smtp.gmail.com:587",
+				From:              "Labs <yaroslav-satellite-test@storj.io>",
+				Auth: mailservice.AuthConfig{
```
egonelbre: Can we flatten these a bit. Specifying config flags as `console.mail.auth.outh2.credentials.client-id=123` seems excessive.

Something like this should be sufficient:

```
Mail: mailservice.Config{
	OutgoingServer: "smtp.gmail.com:587",
	Sender:         "",
	AuthType:       "oauth2",
	ClientID:     "39350227628-vk0ek67d385scmgbvre8ijk1jsdui5kd.apps.googleusercontent.com",
	ClientSecret: "lOS7nLAwR14OjOt6HnpXfRnF",
},
```

Of course, particular subsystems might have different configs.

satellite/console/consolesim/mailsender.go
```diff
@@ -0,0 +1,84 @@
...
+// SendEmail tries to activate account from email
+func (sender *MailsSender) SendEmail(msg *post.Message) error {
+	if msg.Subject != console.ActivationSubject {
+		return nil
+	}
+
+	content := msg.Parts[0].Content
+	index := strings.Index(content, "Verify your account")
+
+	linkEndOffset := 0
```
egonelbre: You can use a regexp here.

satellite/console/service.go
```diff
@@ -41,11 +44,14 @@ type Service struct {
...
 
+	mail *mailservice.Service
+
+	templatePath string
 	passwordCost int
 }
 
 // NewService returns new instance of Service
-func NewService(log *zap.Logger, signer Signer, store DB, passwordCost int) (*Service, error) {
+func NewService(log *zap.Logger, signer Signer, store DB, mail *mailservice.Service, templatePath string, passwordCost int) (*Service, error) {
```
egonelbre: `console` probably shouldn't care about how templates are handled, keep that information inside mailservice.

satellite/console/service.go
```diff
@@ -93,12 +106,27 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser) (u *User, err
...
+			Address: u.Email,
+		},
+		ActivationSubject,
+		filepath.Join(s.templatePath, "Welcome"),
+	)
 
+	err = s.mail.SendRendered(
+		ctx,
+		&AccountActivationEmail{
+			MailTemplate:   tmpl,
```
egonelbre: Tie the template to a particular message rather than specifying it here.

satellite/mailservice/service.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Send is generalized method for sending custom email message
+func (service *Service) Send(ctx context.Context, msg *post.Message) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	return service.sender.SendEmail(msg)
+}
+
+// SendRendered renders content from htmltemplate and texttemplate templates then sends it
+func (service *Service) SendRendered(ctx context.Context, tmpl Template) (err error) {
```
egonelbre: Since everything else should be this func maybe call this `Send` and remove the other one?

satellite/mailservice/service.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// SMTPSender is
+type SMTPSender interface {
+	SendEmail(msg *post.Message) error
+	FromAddress() post.Address
+}
+
+// Template defines mailservice template for SendRendered method
+type Template interface {
```
egonelbre: This isn't a template really. This is the actual `Message`.

satellite/console/service.go
```diff
@@ -93,12 +106,27 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser) (u *User, err
...
+	if err != nil {
+		return u, err
+	}
+
+	tmpl := NewMailTemplate(
+		post.Address{
+			Name:    u.FirstName,
+			Address: u.Email,
+		},
+		ActivationSubject,
```
egonelbre: I think the subject should come from the template not from the code.

satellite/peer.go
```diff
@@ -360,21 +417,38 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 			return nil, errs.Combine(err, peer.Close())
 		}
 
-		peer.Console.Service, err = console.NewService(peer.Log.Named("console:service"),
-			// TODO: use satellite key
+		mailService := peer.Mail.Service
+		// fallback to use account activation mail sender
+		// when using storj-sim
+		if config.SimulateActivation {
```
egonelbre: The setup should be above. It's not really configuring console, but rather mail service.

satellite/console/mail.go
```diff
@@ -0,0 +1,75 @@
...
+	plainTextPath string
+}
+
+// NewMailTemplate creates new instance of MailTemplate
+func NewMailTemplate(to mail.Address, subject, prefix string) MailTemplate {
```
egonelbre: I think if you remove the `MailTemplate` type, make `to` an argument to `Send` and make `subject` a template concern then you can simplify all the templates to:

```
// ProjectInvitationEmail is mailservice template for project invitation email
type ProjectInvitationEmail struct {
	UserName    string
	ProjectName string
}

func (ProjectInvitationEmail) Template() string { return "project-invitation" }
```

internal/testplanet/planet.go
```diff
@@ -443,9 +444,25 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 			Rollup: rollup.Config{
 				Interval: 120 * time.Second,
 			},
+			Mail: mailservice.Config{
+				SMTPServerAddress: "smtp.gmail.com:587",
+				From:              "Labs <yaroslav-satellite-test@storj.io>",
```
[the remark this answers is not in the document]
egonelbre: `Mail: mailservice.Config{ AuthType: "simulated" }`

satellite/peer.go
rikysya: For me it's more like configuring console service to automatically activate account through mail service, but still console specific. This way you won't have to modify stimulative `MailSender` when there is any change in any of the services which uses `mailservice.Service`. If there were more services using `mailservice.Service`, then for storj-sim to work offline I would make global `mail.Config` option, for example to write all emails to file without connecting to any server. 
If I make console sim activation global, then `mailservice` will know about console implementation detail, which it shouldn't  in my opinion.

satellite/console/consolesim/mailsender.go
rikysya: Is it more efficient?

satellite/peer.go
egonelbre: Yeah, my mental model was a mail-sender that automatically clicks links.

satellite/mailservice/service.go
rikysya: But we need to  add user specific data, like activation token, and likelihood of every send to have unique content is very high

egonelbre: I mean those things don't affect parsing. The parsed templates are still the same, the template name and content is what changes.

satellite/console/consolesim/mailsender.go
egonelbre: Not really, but it's probably easier to read what's happening and less assumptions how the email is structured.

satellite/console/consolesim/mailsender.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// FromAddress return empty mail address
+func (sender *MailsSender) FromAddress() post.Address {
+	return post.Address{}
+}
+
+// SendEmail tries to activate account from email
+func (sender *MailsSender) SendEmail(msg *post.Message) error {
+	if msg.Subject != console.ActivationSubject {
```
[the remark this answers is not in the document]
rikysya: That will make it dependent of web api server which service doesn't actually require to work, or it's okay?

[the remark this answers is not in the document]
egonelbre: I mean for testing code I think that's fine. Except why would it be dependent on the web api server, it would only need to depend on `net/http`, as far as I understand.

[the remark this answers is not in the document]
rikysya: Oh, I mean that it requires web server to be started and to use `console.Service`, and that this dependence is indirect.(because service doesn't know who will use it). May be I am overthinking about it:)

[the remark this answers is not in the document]
egonelbre: So, I'm fine with depending on the web server being up and running. This means we end up testing more of our code. Of course, if there's some issue or having this approach complicates things significantly then sure... we can wire up the authorization service or database directly.

But, I understand the concern... if it doesn't work out, then it doesn't and we can use some other approach. For testing code the primary thing is about making it easy to use and ensure that changing a system doesn't require changing irrelevant tests.

go.sum
```diff
@@ -15,6 +15,10 @@ github.com/Sirupsen/logrus v1.0.6 h1:HCAGQRk48dRVPA5Y+Yh0qdCSTzPOyU1tBJ7Q9YzotII
 github.com/Sirupsen/logrus v1.0.6/go.mod h1:rmk17hk6i8ZSAJkSDa7nOxamrG+SP4P0mm+DAvExv4U=
 github.com/StackExchange/wmi v0.0.0-20180725035823-b12b22c5341f h1:5ZfJxyXo8KyX8DgGXC5B7ILL8y51fci/qYz2B4j8iLY=
 github.com/StackExchange/wmi v0.0.0-20180725035823-b12b22c5341f/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
+github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc h1:cAKDfWh5VpdgMhJosfJnn5/FoN2SRZ4p7fJNX58YPaU=
```
egonelbre: revert changes to go.sum

satellite/console/consoleweb/consoleql/mail.go
```diff
@@ -0,0 +1,60 @@
...
+	// ActivationSubject activation email subject
+	ActivationSubject = "Activate your email"
+	// InvitationSubject invitation email subject
+	InvitationSubject = ""
+	// ForgotPasswordSubject forgot password email subject
+	ForgotPasswordSubject = ""
+)
+
+// MailTemplate is implementation of satellite/mailservice.Template interface
+type MailTemplate struct {
```
egonelbre: If you make these methods which have hard-coded values then the code will look nicer.

```
func (AccountActivationEmail) Template() string { return "activation-email" }
```

internal/testplanet/planet.go
```diff
@@ -439,8 +440,17 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
+				AuthType:          "oauth2",
+				RefreshToken:      "1/EztXS-biwVfwGnhq0eZ0PHsjVNyWGApmoIfVqARKALg",
+				ClientID:          "39350227628-vk0ek67d385scmgbvre8ijk1jsdui5kd.apps.googleusercontent.com",
+				ClientSecret:      "lOS7nLAwR14OjOt6HnpXfRnF",
+				TokenURI:          "https://oauth2.googleapis.com/token",
+			},
 			Console: consoleweb.Config{
-				Address:      "127.0.0.1:0",
+				Address:      "127.0.0.1:10100",
```
egonelbre: This needs to be `:0`

satellite/console/consoleweb/consoleql/mail.go
```diff
@@ -0,0 +1,56 @@
...
+	ActivationPath = "activationPath"
+)
+
+// AccountActivationEmail is mailservice template with activation data
+type AccountActivationEmail struct {
+	ActivationLink string
+}
+
+// Template returns email template name
+func (tmpl *AccountActivationEmail) Template() string {
```
egonelbre: For such short things I usually make them single line. And since these don't need the receiver, then no need to define it.

```
// Template returns email template name
func (AccountActivationEmail) Template() string { return "welcome" }
```

satellite/console/consoleweb/server.go
```diff
@@ -103,13 +108,17 @@ func (s *Server) grapqlHandler(w http.ResponseWriter, req *http.Request) {
 		ctx = console.WithAuth(ctx, auth)
 	}
 
+	rootObject := make(map[string]interface{})
+	rootObject["origin"] = "http://" + s.config.Address + "/"
```
egonelbre: Note `s.config.Address` might be `127.0.0.1:0` here. Also since it's a public name you might also need to specify host (for the production) rather than the address.

satellite/mailservice/simulate/linkclicker.go
```diff
@@ -0,0 +1,64 @@
+package simulate
```
egonelbre: maybe `package mailclicker`

satellite/console/consoleweb/consoleql/mutation_test.go
```diff
@@ -61,6 +61,10 @@ func TestGrapqhlMutation(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		rootObject := make(map[string]interface{})
+		rootObject["origin"] = "http://doesntmatter.com/"
```
egonelbre: Use `example.com` that's the designated url for any demonstrative data.

satellite/mailservice/simulate/linkclicker.go
```diff
@@ -24,38 +24,20 @@ func (clicker *LinkClicker) FromAddress() post.Address {
 
 // SendEmail click all links from email html parts
 func (clicker *LinkClicker) SendEmail(msg *post.Message) error {
+	regx := regexp.MustCompile(`href="([^\s])+"`)
```
egonelbre:
```
rxLink := regexp.MustCompile(`<a .* href="(.*)">(.*)</a>`
```

Might suffice... maybe needs to avoid newlines. For these avoiding "packages" with "less correct" approaches, I usually add a comment like "// dirty way to find links without pulling in a dependency"... helps others orient why something "more correct wasn't used".

satellite/console/consoleweb/server.go
```diff
@@ -105,13 +109,18 @@ func (s *Server) grapqlHandler(w http.ResponseWriter, req *http.Request) {
 		ctx = console.WithAuth(ctx, auth)
 	}
 
+	rootObject := make(map[string]interface{})
+	//TODO: add public address to config for production
+	rootObject["origin"] = "http://" + s.listener.Addr().String() + "/"
```
stefanbenten: We should make https:// the default across the board in my opinion

satellite/mailservice/service.go
```diff
@@ -0,0 +1,121 @@
...
+	// TODO(yar): prepare plain text version
+	//text *texttemplate.Template
+}
+
+// New creates new service
+func New(log *zap.Logger, sender Sender, templatePath string) (*Service, error) {
+	var err error
+	service := &Service{log: log, sender: sender}
+
+	// TODO(yar): prepare plain text version
```
stefanbenten: Can we include this, instead of leaving it as TODO?

satellite/mailservice/service.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// SendRendered renders content from htmltemplate and texttemplate templates then sends it
+func (service *Service) SendRendered(ctx context.Context, to []post.Address, msg Message) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	var htmlBuffer bytes.Buffer
+	var textBuffer bytes.Buffer
+
+	// TODO(yar): prepare plain text version
```
stefanbenten: same

satellite/mailservice/service.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+		},
+	}
+
+	err = service.sender.SendEmail(m)
+	// log error
+	if err != nil {
+		service.log.Info("error from mail sender", zap.String("error", err.Error()))
+	} else {
+		service.log.Info("successfully send message")
```
stefanbenten: Include recipient for better track keeping

satellite/peer.go
```diff
@@ -363,30 +374,93 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 		peer.Accounting.Rollup = rollup.New(peer.Log.Named("rollup"), peer.DB.Accounting(), config.Rollup.Interval)
 	}
 
+	{ // setup mailservice
+		log.Debug("Setting up mail service")
+		// TODO(yar): test multiple satellites using same OAUTH credentials
```
stefanbenten: should be easy to simulate with storj-sim

web/satellite/src/utils/apolloManager.ts
```diff
@@ -9,7 +9,7 @@ import { getToken } from '@/utils/tokenManager';
 
 // Satellite url
 const satelliteUrl = new HttpLink({
-    uri: 'http://localhost:10100/api/graphql/v0',
+    uri: 'http://127.0.0.1:10100/api/graphql/v0',
```
stefanbenten: Already fixed in #1393

satellite/mailservice/simulate/linkclicker.go
rikysya: I experimented with this a lot and the one you provide picks all the code from first `a` start tag to last closing tag, if multiline is set, or miss cases when there is text or some other tag is inside `a` and single line set to true. Also the link could never include whitespace(or I hope so, it's usually escaped %20 or smth).

internal/post/sender.go
```diff
@@ -12,20 +12,25 @@ import (
...
+	return sender.From
+}
+
 // SendEmail sends email message to the given recipient
 func (sender *SMTPSender) SendEmail(msg *Message) error {
-	host, _, err := net.SplitHostPort(sender.ServerAddress)
-	if err != nil {
-		return err
-	}
+	host, _, _ := net.SplitHostPort(sender.ServerAddress)
```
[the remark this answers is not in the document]
rikysya: My model is that this stuff should be validate while creating the service, but each time you send the message, as it shouldn't change after initialization

satellite/console/consoleweb/server.go
rikysya: Yeah, that makes sense, but with publicdomain address, this is temporary

satellite/mailservice/service.go
rikysya: Will ad as soon as I have content

satellite/mailservice/service.go
rikysya: Good point

satellite/peer.go
rikysya: Haven't tested it yet

satellite/mailservice/simulate/linkclicker.go
egonelbre: Yeah, it needs to be set as non-greedy and non-multiline... oh-well let's use the one you have then.

internal/post/sender.go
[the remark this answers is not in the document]
stefanbenten: Are we checking it on startup somewhere?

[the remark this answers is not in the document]
egonelbre: add a comment why it's suppressing the error and a TODO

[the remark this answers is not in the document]
rikysya: satellite/peer.go line 390. Will move it to constructor though

PR #1304, opened 2019-02-13
pb.Equal fix

wthorp:
Fixing typo


pkg/pb/BandwidthUtils.go
```diff
@@ -32,7 +32,7 @@ func Equal(msg1, msg2 proto.Message) bool {
 	if err != nil {
 		return false
 	}
```
egonelbre: make the file name more conventional

PR #1306, opened 2019-02-13
Removed ReverseList from KeyValueStore interfaces

wthorp:
Cleanup / simplification in preparation for batched iterators work here:
https://storjlabs.atlassian.net/browse/V3-925

---

KeyValueStore.ReverseList() was only used by the Kademlia routing table buckets list, which is a bounded, _very_ short list that shouldn't need special treatment.

Most KeyValueStores don't support reverse iteration natively, so we had hacks in place instead.  This PR removes those hacks.

### DIFF
--- a/pkg/kademlia/routing_helpers.go
+++ b/pkg/kademlia/routing_helpers.go
@@ -299,18 +299,20 @@ func (rt *RoutingTable) getUnmarshaledNodesFromBucket(bID bucketID) ([]*pb.Node,
 
 // getKBucketRange: helper, returns the left and right endpoints of the range of node ids contained within the bucket
 func (rt *RoutingTable) getKBucketRange(bID bucketID) ([]bucketID, error) {
-	kadIDs, err := rt.kadBucketDB.ReverseList(bID[:], 2)
+	kadBucketIDs, err := rt.kadBucketDB.List(nil, 0)
 	if err != nil {
-		return nil, RoutingErr.New("could not reverse list k bucket ids %s", err)
+		return nil, RoutingErr.New("could not list all k bucket ids: %s", err)
 	}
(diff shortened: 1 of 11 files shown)


pkg/kademlia/routing_helpers.go
```diff
@@ -299,18 +299,20 @@ func (rt *RoutingTable) getUnmarshaledNodesFromBucket(bID bucketID) ([]*pb.Node,
 
 // getKBucketRange: helper, returns the left and right endpoints of the range of node ids contained within the bucket
 func (rt *RoutingTable) getKBucketRange(bID bucketID) ([]bucketID, error) {
-	kadIDs, err := rt.kadBucketDB.ReverseList(bID[:], 2)
+	kadBucketIDs, err := rt.kadBucketDB.List(nil, 0)
```
egonelbre: Maybe specify the limit, so if it does hit the limit it is visible. I know it's probably never going to happen, but nevertheless.

wthorp: the limit should be bit length (256) + bucket size (20), but I don't think we currently clean up our tree structure sufficiently to enforce that

storage/testsuite/test_list.go
```diff
@@ -58,25 +58,12 @@ func testList(t *testing.T, store storage.KeyValueStore) {
...
 
 	for _, test := range tests {
 		var keys storage.Keys
 		var err error
-		if !test.Reverse {
-			keys, err = store.List(test.First, test.Limit)
-		} else {
-			keys, err = store.ReverseList(test.First, test.Limit)
-		}
+		keys, err = store.List(test.First, test.Limit)
```
mniewrzal: You can remove also `Reverse  bool` from `type Test struct `

PR #1307, opened 2019-02-13
Mapping routing tables

bryanchriswhite:
_super wip_


Makefile
```diff
@@ -72,6 +72,12 @@ install-sim: ## install storj-sim
 	@echo "Running ${@}"
 	@go install -race -v storj.io/storj/cmd/storj-sim storj.io/storj/cmd/bootstrap storj.io/storj/cmd/satellite storj.io/storj/cmd/storagenode storj.io/storj/cmd/uplink storj.io/storj/cmd/gateway storj.io/storj/cmd/identity storj.io/storj/cmd/certificates
 
+##@ Utility
+.PHONY: map
+map: ## build a graphviz map of routing tables
```
mobyvb: `map-network` might be a little clearer

cmd/inspector/map.go
```diff
@@ -0,0 +1,156 @@
...
+	"storj.io/storj/pkg/storj"
+)
+
+type DOTGraph struct {
+	name     string
+	routeMap map[string][]string
+}
+
+var (
+	cmdMap = &cobra.Command{
```
mobyvb: same "map" -> "map network" suggestion applies for this file

Makefile
bryanchriswhite: :+1:

cmd/inspector/main.go
```diff
@@ -110,10 +110,11 @@ var (
 
 // Inspector gives access to kademlia and overlay cache
 type Inspector struct {
-	identity      *identity.FullIdentity
-	kadclient     pb.KadInspectorClient
-	overlayclient pb.OverlayInspectorClient
-	statdbclient  pb.StatDBInspectorClient
+	transportClient transport.Client
+	identity        *identity.FullIdentity
+	kadclient       pb.KadInspectorClient
```
navillasa: extremely nitty nit but `transportClient` is camelCased but `kadclient` and the other clients aren't

cmd/inspector/map-network.go
```diff
@@ -0,0 +1,172 @@
...
+	n := &network{
+		ctx:    ctx,
+		verify: verify,
+		seen:   map[string]struct{}{args[0]: {}},
+		addrs:  []string{args[0]},
+		out:    tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0),
+		err:    errLog,
+		graph:  dot2.NewDot("StorjNetwork"),
+	}
+	_, _ = fmt.Fprintf(n.out, "digraph StorjNetwork {\n")
```
navillasa: maybe handle error

cmd/inspector/map-network.go
```diff
@@ -0,0 +1,172 @@
...
+
+			n.walk(newAddr)
+		}
+	}
+}
+
+func buildDot(routes routes) string {
+	dot := dot2.NewDot("storj network")
+
+	for _, route := range routes {
```
navillasa: I'm not sure how big `routes` and `route.neighbors` will be, but I'm thinking it could be better to do a map than a nested loop unless `route.neighbors` is super short. but also idk could be fine!

cmd/inspector/main.go
bryanchriswhite: I didn't originally do the inspector struct so I'm reluctant to change an existing field name in this PR but I agree we should be consistent and be using camelCase.

bryanchriswhite:
Closing because the inspector service has been moved to a private grpc server and is therefore no longer a reliable way to map the network.

PR #1310, opened 2019-02-13
Add uplink tests

dylanlott:
This commit adds an end to end test of the uplink binary against the storj-sim network.

### DIFF
--- a/cmd/storj-sim/network.go
+++ b/cmd/storj-sim/network.go
@@ -257,6 +257,7 @@ func newNetwork(flags *Flags) (*Processes, error) {
 				}
 			}
 
+			fmt.Printf("\nAPI KEY %+v\n", vip.Get("client.api-key"))
 			accessKey := vip.GetString("minio.access-key")
 			secretKey := vip.GetString("minio.secret-key")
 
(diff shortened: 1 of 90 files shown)


cmd/storj-sim/network.go
```diff
@@ -171,6 +171,7 @@ func newNetwork(flags *Flags) (*Processes, error) {
 
 				"--server.extensions.revocation=false",
 				"--server.use-peer-ca-whitelist=false",
+				"--pointer-db.auth.api-key", "abc123",
```
egonelbre: this is not needed

scripts/test-uplink.sh
```diff
@@ -0,0 +1,60 @@
...
+cleanup(){
+    rm -rf "$TMPDIR"
+    echo "cleaned up test successfully"
+}
+
+trap cleanup EXIT
+
+BUCKET=bucket-123
+SRC_DIR=$TMPDIR/source
+DST_DIR=$TMPDIR/dst
```
egonelbre: Let's fix this thing in the test-aws script as well.

cmd/storj-sim/network.go
dylanlott: Removing it 👍

scripts/test-uplink.sh
dylanlott: Fixed

egonelbre: did you push?

PR #1311, opened 2019-02-14
Create dbutil package for sqlite

mniewrzal:
Changes are part of migration testing mechanism (satellite and storagenode). 

related ticket https://storjlabs.atlassian.net/browse/V3-1172


internal/dbutil/sqliteutil/query.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return schema, err
+	}
+
+	return schema, nil
+}
+
+var (
+	// matches UNIQUE (a,b)
+	rxUnique = regexp.MustCompile(`UNIQUE[\s]*\((.*?)\)`)
```
egonelbre: `[\s]` -> `\s`

internal/dbutil/sqliteutil/query.go
```diff
@@ -0,0 +1,191 @@
...
+		defer func() { err = errs.Combine(err, rows.Close()) }()
+
+		for rows.Next() {
+			var defName, defType, defSQL string
+			err := rows.Scan(&defName, &defType, &defSQL)
+			if err != nil {
+				return err
+			}
+			if defType == "table" {
+				tableDefinitions = append(tableDefinitions, &Definition{name: defName, sql: defSQL})
```
egonelbre: You can add the partial table and index directly to the schema, that way you don't have to have the intermediate Definition.

mniewrzal: in schema there is no place for `sql`

wthorp:
Is there a ticket that describes why this work is happening?

egonelbre:
`dbschema` stuff is for testing migration end-results... e.g. see https://github.com/storj/storj/pull/1301/files#diff-373ea6586bb8703504cb9dbe4479eb5cR24 as an example

mniewrzal:
its part of migration testing and for storage node migration ticket is here https://storjlabs.atlassian.net/browse/V3-1172

internal/dbutil/sqliteutil/db.go
```diff
@@ -0,0 +1,79 @@
...
+		Schema:  schema,
+		Data:    data,
+	}, err
+}
+
+// QueryData loads all data from tables
+func QueryData(db dbschema.Queryer, schema *dbschema.Schema) (*dbschema.Data, error) {
+	return dbschema.QueryData(db, schema, func(columnName string) string {
+		quoted := strconv.Quote(columnName)
+		// TODO is this the same like postgres 'quote_nullable'?
```
egonelbre: Yes, sqlite `.dump` was using `quote` as well.

internal/dbutil/sqliteutil/query.go
```diff
@@ -0,0 +1,186 @@
...
+				}
+			}
+		}
+		return err
+	}()
+	if err != nil {
+		return nil, err
+	}
+
+	// discover indexes
```
navillasa: it might be worth breaking the "find tables and indexes", "discover tables" and "discover indexes" code into separate helper functions for readability but this isn't critical.

internal/dbutil/sqliteutil/query.go
```diff
@@ -0,0 +1,186 @@
...
+					table.PrimaryKey = append(table.PrimaryKey, name)
+				}
+
+			}
+
+			matches := rxUnique.FindAllStringSubmatch(definition.sql, -1)
+			for _, match := range matches {
+				// TODO feel this can be done easier
+				var columns []string
+				for _, name := range strings.Split(match[1], ",") {
```
navillasa: just clarification question- what's at `match[0]`? I also feel like there might be more efficient way to do this but not sure what it is.

egonelbre: `match[0]` is the full regular expression match... `match[1]` is the first group match.

Performance doesn't matter that much here since 1. it's only for testing and it's iterating the table schema, which at most is ~10 tables.

PR #1312, opened 2019-02-14
dbutil add snapshot logic

egonelbre:


internal/dbutil/pgutil/data.go
```diff
@@ -0,0 +1,71 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/internal/dbutil/dbschema"
+)
+
+// QueryData loads all data from tables
+func QueryData(db Queryer, schema *dbschema.Schema) (*dbschema.Data, error) {
```
mniewrzal: hmm everything except `quote_nullable` look to be independent from `postgres`. Maybe it can moved to `dbutil`?

egonelbre: Hmm, I could move it into `dbschema with an optional `quoteColumn func(string)` ?

mniewrzal: yes, I think thats easiest solution.

egonelbre: done

internal/dbutil/pgutil/data.go
```diff
@@ -0,0 +1,18 @@
...
+	"strconv"
+
+	"storj.io/storj/internal/dbutil/dbschema"
+)
+
+// QueryData loads all data from tables
+func QueryData(db dbschema.Queryer, schema *dbschema.Schema) (*dbschema.Data, error) {
+	return dbschema.QueryData(db, schema, func(columnName string) string {
+		quoted := strconv.Quote(columnName)
+		return `quote_nullable(` + quoted + `) as ` + quoted
```
wthorp: I think strconv.Quote + quote_nullable will double-escape things

internal/dbutil/pgutil/db.go
```diff
@@ -0,0 +1,100 @@
...
+func (db *DB) Close() error {
+	return errs.Combine(
+		DropSchema(db.DB, db.Schema),
+		db.DB.Close(),
+	)
+}
+
+// LoadSchemaFromSQL inserts script into connstr and loads schema.
+func LoadSchemaFromSQL(connstr, script string) (_ *dbschema.Schema, err error) {
+	db, err := Open(connstr, "load-schema")
```
wthorp: does using the fixed 'load-schema' schema name here have data race implications?

internal/dbutil/dbschema/data.go
```diff
@@ -0,0 +1,107 @@
...
+			defer func() { err = errs.Combine(err, rows.Close()) }()
+
+			row := make(RowData, len(columnNames))
+			rowargs := make([]interface{}, len(columnNames))
+			for i := range row {
+				rowargs[i] = &row[i]
+			}
+
+			for rows.Next() {
+				err := rows.Scan(rowargs...)
```
wthorp: I need to brush up on my variadics.  Does this work, or should rowargs be []*interface{} instead?
"If an argument has type *interface{}, Scan copies the value provided by the underlying driver without conversion"

internal/dbutil/pgutil/db.go
egonelbre: no `Open` always randomizes the full schema name.

wthorp: ah yes

internal/dbutil/pgutil/data.go
egonelbre: `strconv.Quote` escapes the column name not the value. `quote_nullable` escapes the value retrieved from the database.

internal/dbutil/dbschema/data.go
egonelbre: I do know that the resulting value is a string, so I provide an string pointer as a place to store the data. If I didn't know the actual type, then yes, I would need to use `*interface{}` instead.

PR #1313, opened 2019-02-14
Reduce Log Messages on Level Info

stefanbenten:
The Uplink has too many Log Outputs on Level Info, that arent necessary.
Its also inconsistent across Upload, Download and Delete.


stefanbenten:
@kaloyan-raev Also if we start to cut the long tail, shouldnt the log output be suppressed? 
At that stage its not a error rather than a expected behaviour.

PR #1314, opened 2019-02-14
Fix invalid config key msg for dashbord

mniewrzal:
Fix or workaround (not sure) for:
image


pkg/process/exec_conf.go
```diff
@@ -179,19 +179,21 @@ func cleanup(cmd *cobra.Command) {
...
-			if cmd.Flags().Lookup(key) == nil {
-				// flag couldn't be found
-				brokenKeys = append(brokenKeys, key)
-			} else {
-				oldChanged := cmd.Flag(key).Changed
-				err := cmd.Flags().Set(key, vip.GetString(key))
-				if err != nil {
-					// flag couldn't be set
-					brokenVals = append(brokenVals, key)
+		if cmd.Annotations["type"] != "helper" {
```
egonelbre: Let's still keep the check for broken values. You can just skip the printing below, we can still collect the broken values.

mniewrzal: ok

pkg/process/exec_conf.go
```diff
@@ -214,12 +214,14 @@ func cleanup(cmd *cobra.Command) {
...
-			logger.Sugar().Infof("Invalid configuration file key: %s", key)
-		}
-		for _, key := range brokenVals {
-			logger.Sugar().Infof("Invalid configuration file value for key: %s", key)
+		if cmd.Annotations["type"] != "helper" {
+			// okay now that logging is working, inform about the broken keys
+			for _, key := range brokenKeys {
+				logger.Sugar().Infof("Invalid configuration file key: %s", key)
+			}
+			for _, key := range brokenVals {
```
egonelbre: I think we still want to print invalid values?

mniewrzal: yes, you are right

PR #1315, opened 2019-02-14
fix kad refresh bad bucket range

wthorp:
Per @littleskunk it looks like our kad refreshes aren't right.  I believe this should fix the issue:

"I have 5 buckets. I would expect a refresh on each bucket. On each bucket I would expect ping messages to the same group of nodeIDs. To me it looks like we don't care about the 5 buckets and just generate a global random ID and use that. It might be a totally different bucket."

2019-02-13T22:08:11.061Z        INFO    kademlia        refresh called
2019-02-13T22:08:11.061Z        INFO    kademlia        refresh {"random ID": "1AwdMCQQq5xBN1g28Twa2G1spQjZHs3qAbhvT9a5UG9xEZ5Dvm", "startID": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0], "BucketID": [63,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]}
2019-02-13T22:08:11.081Z        INFO    kademlia        refresh {"random ID": "1jzQHp8ThYcWWi3UwfEpicz3NLZmUjonCSdjdprWeMowGsfBQ6", "startID": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0], "BucketID": [127,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]}
2019-02-13T22:08:11.102Z        INFO    kademlia        refresh {"random ID": "1HPjR7FN9gohPWbtHfjUZ24KdFbjkUX4avzEkAvy7n7azi9YyT", "startID": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0], "BucketID": [223,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]}
2019-02-13T22:08:11.123Z        INFO    kademlia        refresh {"random ID": "12kZaw7KFB8nZa9DWhBgFG3s8BU6ozu1uhQuVxSsTemybfpbEUG", "startID": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0], "BucketID": [239,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]}
2019-02-13T22:08:11.144Z        INFO    kademlia        refresh {"random ID": "1DhMhnRUAJHdMN2NgGtRAgP16rpRUFRudb18NCvuBUHr99w6xe", "startID": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0], "BucketID": [255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]}


pkg/kademlia/kademlia.go
```diff
@@ -322,7 +322,9 @@ func (k *Kademlia) refresh(ctx context.Context, threshold time.Duration) error {
 		if tErr != nil {
 			errors.Add(tErr)
 		} else if now.After(ts.Add(threshold)) {
-			rID, _ := randomIDInRange(startID, keyToBucketID(bID))
+			endID := keyToBucketID(bID)
+			rID, _ := randomIDInRange(startID, endID)
+			startID = endID
```
egonelbre: this doesn't look right

so if we have buckets:

1. refresh
2. not stale
3. not stale
4. needs refresh

then the last refresh would refresh buckets 2-4, which might miss 4.

wthorp: doh!  fixed it

PR #1318, opened 2019-02-14
Replace list calls with iterates in kademlia routing table 

jenlij:


pkg/kademlia/routing_helpers.go
```diff
@@ -234,27 +223,16 @@ func (rt *RoutingTable) getNodeIDsWithinKBucket(bID bucketID) (storj.NodeIDList,
...
-	}
-	for _, v := range allNodeIDsBytes {
-		if (bytes.Compare(v, left[:]) > 0) && (bytes.Compare(v, right[:]) <= 0) {
-			nodeIDsBytes = append(nodeIDsBytes, v)
-			if len(nodeIDsBytes) == rt.bucketSize {
-				break
-			}
+	var ids []storj.NodeID
+
+	err = rt.iterateNodes(storj.NodeID{}, func(nodeID storj.NodeID, protoNode []byte) error {
```
egonelbre: is there a reason we are not supplying `left` as the starting point? Also doing iteration over the nodes twice seems wasteful.

pkg/kademlia/routing_helpers_test.go
```diff
@@ -349,24 +308,25 @@ func TestNodeIsWithinNearestK(t *testing.T) {
...
-			closest: true,
+			closest: false,
 		},
 		{testID: "E",
 			nodeID:  storj.NodeID{133, 255},
 			closest: false,
 		},
 	}
-
+	xors := []storj.NodeID{}
```
egonelbre: are the `xors` used somehow?

pkg/kademlia/kademlia_test.go
```diff
@@ -91,11 +91,11 @@ func TestPeerDiscovery(t *testing.T) {
 	defer cleanup()
 	// make new identity
 	bootServer, mockBootServer, bootID, bootAddress := startTestNodeServer(ctx)
```
egonelbre: You need to wait for all the servers to properly stop.

pkg/kademlia/routing_helpers.go
jenlij: that's fair, I'll condense it

pkg/kademlia/routing_helpers_test.go
jenlij: good catch, that's residual. will remove

pkg/kademlia/routing_helpers.go
```diff
@@ -155,51 +154,41 @@ func (rt *RoutingTable) createOrUpdateKBucket(bID bucketID, now time.Time) error
...
+		return bucketID{}, RoutingErr.Wrap(err)
 	}
-	return nodeIDs[rt.bucketSize]
+	return match, nil
 }
 
 // nodeIsWithinNearestK: helper, returns true if the node in question is within the nearest k from local node
 func (rt *RoutingTable) nodeIsWithinNearestK(nodeID storj.NodeID) (bool, error) {
-	nodeKeys, err := rt.nodeBucketDB.List(nil, 0)
+	closestNodes, err := rt.FindNear(rt.self.Id, rt.bucketSize)
```
wthorp: Can we just check to see if nodeID is closestNodes at this point?

pkg/kademlia/routing_helpers.go
```diff
@@ -155,51 +154,41 @@ func (rt *RoutingTable) createOrUpdateKBucket(bID bucketID, now time.Time) error
...
-	if len(nodeIDs) < rt.bucketSize+1 { //adding 1 since we're not including local node in closest k
-		return nodeIDs[len(nodeIDs)-1]
+		return bucketID{}, RoutingErr.Wrap(err)
 	}
-	return nodeIDs[rt.bucketSize]
+	return match, nil
 }
 
 // nodeIsWithinNearestK: helper, returns true if the node in question is within the nearest k from local node
 func (rt *RoutingTable) nodeIsWithinNearestK(nodeID storj.NodeID) (bool, error) {
```
wthorp: rename to something like 'wouldBeInNearestK' if that's more like what it does

pkg/kademlia/routing_helpers_test.go
```diff
@@ -349,24 +308,25 @@ func TestNodeIsWithinNearestK(t *testing.T) {
...
 			closest: true,
 		},
 		{testID: "D",
 			nodeID:  storj.NodeID{191, 255},
-			closest: true,
+			closest: false,
 		},
 		{testID: "E",
 			nodeID:  storj.NodeID{133, 255},
 			closest: false,
```
wthorp: if the bucketsize is 6 and we're adding 5 things, so are we ever not within K?

jenlij: Oh I wonder if that's new, I see it's 6 now, I thought it was 2 at one point. In that case, yeah this whole thing is wrong then.

wthorp:
Some of your linting troubles are from casting bucket ids to node ids.  Recently you (we?) put in a PR where we can changed BucketID from being a type definition to a type alias.  I'm really not sure which I like better.  https://yourbasic.org/golang/type-alias/

PR #1321, opened 2019-02-15
Add protolock to build

mniewrzal:
Adds `protolock` for checking protobuf backward incompatible changes.

https://github.com/nilslice/protolock


.travis.yml
```diff
@@ -57,12 +57,15 @@ matrix:
         - go get github.com/ckaznocha/protoc-gen-lint
         - go get golang.org/x/tools/go/packages
         - cp go.mod.backup go.mod
+        # install protolock
+        - go get -u github.com/nilslice/protolock/...
```
egonelbre: Move it upwards before restoring go.mod, and don't use `go get -u` it will try to reupdate everything that has already been cleanly installed.

.travis.yml
```diff
@@ -57,12 +57,15 @@ matrix:
...
       script:
         - go run ./scripts/check-copyright.go
         - go run ./scripts/check-imports.go ./...
         - go run ./scripts/protobuf.go --protoc=$HOME/protoc/bin/protoc lint
         - golangci-lint run
         - ./scripts/check-travis-tidy.sh
+        - protolock status
```
egonelbre: move it right after protobuf.go test

mniewrzal: I will try also to check status before `protoc` in `protobuf.go` script

mniewrzal: before generating

.travis.yml
egonelbre: and use the concerete path of `github.com/nilslice/protolock/cmd/protolock`, might avoid downloading some dependencies

.travis.yml
egonelbre: yeah, that would also work

scripts/protobuf.go
```diff
@@ -115,12 +116,26 @@ func install(deps ...string) error {
...
+	local, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	cmd.Dir = findProtolockDir(local)
+	out, err := cmd.CombinedOutput()
+	if len(out) > 0 {
+		fmt.Println(string(out))
+	}
+	if err != nil {
```
navillasa: maybe doesn't matter but why is the error check happening after the len(out) check?

scripts/protobuf.go
```diff
@@ -235,3 +250,12 @@ func listProtoFiles(root string) ([]string, error) {
 
 	return files, err
 }
+
+func findProtolockDir(dir string) string {
+	protolock := filepath.Join(dir, "proto.lock")
+	if _, err := os.Stat(protolock); err == nil {
+		return dir
+	}
```
navillasa: do you need to error check here or is it okay?

scripts/protobuf.go
egonelbre: Because we want to output the content why it failed, usually it contains useful error message what the actual error is.

PR #1323, opened 2019-02-18
handle server termination in kademlia tests

jenlij:


pkg/kademlia/kademlia_test.go
```diff
@@ -185,20 +174,27 @@ func testNode(t *testing.T, bn []pb.Node) (*Kademlia, *grpc.Server, func()) {
 	grpcServer := grpc.NewServer(identOpt)
 
 	pb.RegisterNodesServer(grpcServer, s)
-	go func() { assert.NoError(t, grpcServer.Serve(lis)) }()
+	ctx.Go(func() error {
```
coyle: nit -- 
```
ctx.Go(func() error {
if err := grpcServer.Serve(lis); err != nil && err == grpc.ErrServerStopped {
		return nil
  }
  return err
})
```

egonelbre: I'm not sure whether that's more readable or not.

PR #1326, opened 2019-02-18
restrict actions on PayerBandwidthAllocation

coyle:
https://storjlabs.atlassian.net/browse/V3-1076


pkg/pointerdb/allocation.go
```diff
@@ -64,3 +70,16 @@ func (allocation *AllocationSigner) PayerBandwidthAllocation(ctx context.Context
 	}
 	return pba, nil
 }
+
+func (allocation *AllocationSigner) restrictActions(peerID storj.NodeID, action pb.BandwidthAction) error {
+	switch action {
+	case pb.BandwidthAction_GET_REPAIR, pb.BandwidthAction_PUT_REPAIR, pb.BandwidthAction_GET_AUDIT:
```
mniewrzal: In general looks good. I'm just wondering if opposite logic wouldn't be more secure. If we will add another satellite action then by default it will become allowed. If we will specify here uplink actions then all not specified actions will be restricted by default.

egonelbre: Yup I agree, by default it should return always restricted for all action types. Then you add exceptions when to allow.

coyle: done

PR #1327, opened 2019-02-19
V3-1152 Node bootstrap web backend

crawter:
https://storjlabs.atlassian.net/browse/V3-1152


cmd/storj-sim/network.go
```diff
@@ -111,13 +111,14 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
-		storageNodePort = 11000
-		consolePort     = 10100
+		configDir        = flags.Directory
+		host             = flags.Host
+		gatewayPort      = 9000
+		bootstrapPort    = 9999
+		satellitePort    = 10000
+		storageNodePort  = 11000
+		consolePort      = 10100
+		bootstrapWebPort = 10010
```
egonelbre: We need a better strategy for picking these, since they seem really disjointed. But it's fine for now.

bootstrap/bootstrapweb/service.go
```diff
@@ -0,0 +1,53 @@
...
+	if kademlia == nil {
+		return nil, errs.New("kademlia can't be nil")
+	}
+
+	return &Service{log: log, kademlia: kademlia}, nil
+}
+
+// IsNodeUp is a method for checking if node is up
+func (s *Service) IsNodeUp(ctx context.Context, nodeID pb.NodeID) (bool, error) {
+	nodes, err := s.kademlia.FindNear(ctx, nodeID, findNearLimit)
```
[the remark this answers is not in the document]
jtolds: FindNear *should* determine whether the node is accessible. An invariant we rely on is that kademlia.FindNear only returns accessible nodes. (this is different than routingtable.FindNear). If this is not what kademlia.FindNear is doing we need to fix FindNear. FindNear should be the most general solution to this problem.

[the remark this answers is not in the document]
egonelbre: `FindNear` is more of an internal function at the moment. `FindNode` is the one that finds a node on the network. I know it's confusing. We probably need to rename `FindNear` to be more explicit that it's only searching routing table.

internal/storjql/satellite/apikey.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-package consoleql
+package satellite
```
egonelbre: Wait, why everything was moved into internal?

bootstrap/peer.go
```diff
@@ -225,6 +225,14 @@ func (peer *Peer) Close() error {
 		}
 	}
 
+	if peer.BootstrapWeb.Endpoint != nil {
+		errlist.Add(peer.BootstrapWeb.Endpoint.Close())
+	} else {
+		if peer.BootstrapWeb.Endpoint != nil {
+			errlist.Add(peer.Public.Listener.Close())
```
egonelbre: Why is this closing the public listener when it's not registered to it?

crawter: doh, my bad. It seems that we have the same issue in satellite/peer Close method

if peer.Console.Endpoint != nil {
		errlist.Add(peer.Console.Endpoint.Close())
	} else {
		if peer.Console.Endpoint != nil {
			errlist.Add(peer.Public.Listener.Close())
		}
	}

i will remove redundant code from both methods

bootstrap/peer.go
```diff
@@ -68,6 +73,13 @@ type Peer struct {
 		Endpoint     *kademlia.Endpoint
 		Inspector    *kademlia.Inspector
 	}
+
+	// Web server with web UI
+	BootstrapWeb struct {
```
egonelbre: `Web` should be here sufficient.

cmd/storj-sim/network.go
egonelbre: Maybe let's use `9998` here, less likely to end up conflicting with satellite.

internal/storjql/schema.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"storj.io/storj/bootstrap/bootstrapweb"
+	"storj.io/storj/bootstrap/bootstrapweb/bootstrapserver/bootstrapql"
+	"storj.io/storj/satellite/console"
+	"storj.io/storj/satellite/console/consoleweb/consoleql"
+	"storj.io/storj/satellite/mailservice"
+)
+
+// creatingSchemaMutex locks graphql.NewSchema method because it's not thread-safe
+var creatingSchemaMutex sync.Mutex
```
egonelbre: I think only expose the mutex here, that way bootstrap and satellite won't end up dependent on each other. At the moment if you try to compile bootstrap it will end up compiling all satellite dependencies into the same binary.

satellite/peer.go
```diff
@@ -531,10 +531,6 @@ func (peer *Peer) Close() error {
 
 	if peer.Console.Endpoint != nil {
 		errlist.Add(peer.Console.Endpoint.Close())
-	} else {
-		if peer.Console.Endpoint != nil {
-			errlist.Add(peer.Public.Listener.Close())
```
egonelbre: It still needs to close the listener. I.e. `peer.Console.Listener`

bootstrap/peer.go
egonelbre: Note, it should close the appropriate listener nevertheless. `peer.Web.Listener`

satellite/console/consoleweb/consoleql/schema.go
```diff
@@ -1,25 +1,17 @@
...
 	"storj.io/storj/satellite/console"
 	"storj.io/storj/satellite/mailservice"
 )
 
-// creatingSchemaMutex locks graphql.NewSchema method because it's not thread-safe
-var creatingSchemaMutex sync.Mutex
-
-// CreateSchema creates both type
+// CreateSchema creates a schema for satellites console graphql api
 func CreateSchema(service *console.Service, mailService *mailservice.Service) (graphql.Schema, error) {
```
egonelbre: The locking still needs to happen inside CreateSchema-s since it's being called from tests as well. Meaning there's still a  potential race.

cmd/storj-sim/network.go
```diff
@@ -131,6 +132,9 @@ func newNetwork(flags *Flags) (*Processes, error) {
 	bootstrap.Arguments = withCommon(Arguments{
 		"setup": {
 			"--identity-dir", bootstrap.Directory,
+
+			"--bootstrap-web.address", net.JoinHostPort(host, strconv.Itoa(bootstrapWebPort)),
```
egonelbre: `--web.address`

bootstrap/peer.go
```diff
@@ -39,6 +42,8 @@ type Config struct {
 
 	Server   server.Config
 	Kademlia kademlia.Config
+
+	BootstrapWeb bootstrapserver.Config
```
egonelbre: `Web` is sufficient, since the full type name is `bootstrap.Config`

PR #1328, opened 2019-02-20
Adds DumpNodes method to Kademlia Inspector

dylanlott:


pkg/kademlia/routing.go
```diff
@@ -147,6 +147,39 @@ func (rt *RoutingTable) GetBucketIds() (storage.Keys, error) {
 	return kbuckets, nil
 }
 
+// DumpNodes loops returns all nodes in the nodeBucketDB and marshals them to &pb.Nodes
+func (rt *RoutingTable) DumpNodes() ([]*pb.Node, error) {
```
wthorp: Jen is working on replacing all the calls to List() w/ iterators [https://github.com/storj/storj/pull/1318].  I'd use routing.go's iterate() https://github.com/storj/storj/blob/36e4d8c1c28594cb9b30950db227cc7fd025886e/pkg/kademlia/routing.go#L262 here

pkg/kademlia/routing.go
```diff
@@ -149,6 +149,19 @@ func (rt *RoutingTable) GetBucketIds() (storage.Keys, error) {
 	return kbuckets, nil
 }
 
+// DumpNodes iterates through all nodes in the nodeBucketDB and marshals them to &pb.Nodes, then returns them
+func (rt *RoutingTable) DumpNodes() ([]*pb.Node, error) {
+	var nodes []*pb.Node
+	err := rt.iterate(storj.NodeID{}, func(newID storj.NodeID, protoNode []byte) error {
+		newNode := pb.Node{}
+		err := proto.Unmarshal(protoNode, &newNode)
+		nodes = append(nodes, &newNode)
```
mniewrzal: Should we append when err != nil?

pkg/kademlia/routing.go
```diff
@@ -149,6 +149,19 @@ func (rt *RoutingTable) GetBucketIds() (storage.Keys, error) {
...
+		return err
+	})
+
+	return nodes, Error.Wrap(err)
```
mniewrzal: I think its safer to return nil array when an error occurs.

pkg/overlay/inspector.go
```diff
@@ -30,3 +31,8 @@ func (srv *Inspector) CountNodes(ctx context.Context, req *pb.CountNodesRequest)
 		Count: int64(len(overlayKeys)),
 	}, nil
 }
+
+// DumpNodes returns all of the nodes in the overlay cachea
```
mniewrzal: typo - cachea

pkg/kademlia/routing.go
dylanlott: Good point. I'll rework this to use ErrorGroup handling.

pkg/kademlia/routing.go
dylanlott: Same as above, using ErrorGroup here for this.

pkg/overlay/inspector.go
dylanlott: Resolved

PR #1333, opened 2019-02-20
BWA aliases

wthorp:
Bandwidth agreement naming has always been confusing

This PR fixes some of the naming issues by aliasing the protocol buffer names.
Using aliases prevents changing the actual protocol buffers, which would cause conflicts.

This also sets up for possibly using the OrderLimit / Order types as composite structs to fix some of things Bill broke with deterministic serialization.


pkg/accounting/tally/tally_test.go
```diff
@@ -69,7 +69,7 @@ func sendGeneratedAgreements(ctx context.Context, t *testing.T, db satellite.DB,
 
 	agreements := make([]*psdb.Agreement, len(actions))
 	for i, action := range actions {
-		pba, err := testbwagreement.GeneratePayerBandwidthAllocation(action, satID, upID, time.Hour)
+		pba, err := testbwagreement.GenerateFundsOrder(action, satID, upID, time.Hour)
```
egonelbre: `OrderLimit` makes more sense, because we are not actually sending funds, but rather setting a transfer limit. Funds / payment calculation is entirely separate.

pkg/accounting/tally/tally_test.go
```diff
@@ -69,11 +69,11 @@ func sendGeneratedAgreements(ctx context.Context, t *testing.T, db satellite.DB,
...
 		require.NoError(t, err)
 		err = db.CertDB().SavePublicKey(ctx, pba.UplinkId, upID.Leaf.PublicKey)
 		assert.NoError(t, err)
-		rba, err := testbwagreement.GenerateRenterBandwidthAllocation(pba, snID.ID, upID, 1000)
+		rba, err := testbwagreement.GenerateFileOrder(pba, snID.ID, upID, 1000)
```
egonelbre: We are not transferring files. Hence just name `Order`.

wthorp:
Ah, I accidentally pushed something I didn't want yet, I must have reverted too far back

PR #1339, opened 2019-02-21
Try splitting tests into two

egonelbre:


egonelbre:
didn't work as intended

PR #1340, opened 2019-02-21
Bucket usage info rollup table creation and implementation

rikysya:
- Added bucket_usage rollup table to masterDB
- Interface for working with BucketUsage added to consoleDB
- Implementation of BucketUsage
- Initial unit tests


satellite/console/bucketusage.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/skyrings/skyring-common/tools/uuid"
+)
+
+// BucketUsages is bucket usage rollup store interface
+type BucketUsages interface {
+	Get(context.Context, uuid.UUID) (*BucketUsage, error)
```
egonelbre: Add ProjectID as an argument here and usage bucket name as the other key.

rikysya: More than one roll up can have the same projectID and bucketID, so it's not sufficient to get particular roll up, if I understood right what you meant. Also, buckets to project relationship is many to one, assuming we have unique bucket ID we have enough info to determine which project it belongs to. And does it really matter which project bucket belongs to on `accounting` pkg level? My model is that `accounting` collects all usage with respect to buckets, than `console` queries for all buckets related to current project, than asks for each buckets roll ups. Please, correct me if I am wrong, so I can get better understanding of what's going on. Thanks!

PR #1341, opened 2019-02-21
Remove everything from Node protobuf except ID and Address

kaloyan-raev:
Related to https://storjlabs.atlassian.net/browse/V3-1229

This is an attempt to clear routing table and overlay cache from everything but node id and address. Thus the `Node` protobuf has only these two fields left.

We need to define new protobufs for those cases where metadata, restrictions and reputation is necessary. This is mostly the places with the commented code.


bootstrap/peer.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	"storj.io/storj/bootstrap/bootstrapweb"
 	"storj.io/storj/bootstrap/bootstrapweb/bootstrapserver"
+	"storj.io/storj/pkg/dht"
```
egonelbre: we probably should be getting rid of dht package, it's mainly indirection that doesn't help with clarity.

jenlij: We do use the dht package to create a testing interface for the routing table...

go.mod
```diff
@@ -7,11 +7,8 @@ require (
 	github.com/go-ole/go-ole v1.2.1 // indirect
```
jenlij: Don't we usually not commit the changes to go.mod and go.sum?

internal/testplanet/uplink.go
```diff
@@ -134,7 +133,7 @@ func (uplink *Uplink) DialMetainfo(ctx context.Context, destination Peer, apikey
 func (uplink *Uplink) DialPiecestore(ctx context.Context, destination Peer) (*piecestore.Client, error) {
 	node := destination.Local()
 
-	conn, err := uplink.Transport.DialNode(ctx, &node)
+	conn, err := uplink.Transport.DialNode(ctx, &node.Node)
```
jenlij: nit: kinda weird sounding to use node.Node

pkg/audit/cursor.go
```diff
@@ -155,18 +155,14 @@ func (cursor *Cursor) createOrderLimits(ctx context.Context, pointer *pb.Pointer
...
 		}
 
-		if node != nil {
-			node.Type.DPanicOnInvalid("auditor order limits")
-		}
-
 		limits[piece.GetPieceNum()] = &pb.AddressedOrderLimit{
 			Limit:              orderLimit,
-			StorageNodeAddress: node.Address,
+			StorageNodeAddress: node.GetNode().GetAddress(),
```
jenlij: good

pkg/audit/reporter.go
```diff
@@ -84,11 +84,11 @@ func (reporter *Reporter) recordAuditFailStatus(ctx context.Context, failedAudit
 	failedIDs := storj.NodeIDList{}
 
 	for _, nodeID := range failedAuditNodeIDs {
-		_, err := reporter.statdb.Update(ctx, &statdb.UpdateRequest{
-			NodeID:       nodeID,
-			IsUp:         true,
-			AuditSuccess: false,
-		})
+		// _, err := reporter.statdb.Update(ctx, &statdb.UpdateRequest{
```
jenlij: why are these commented out?

pkg/audit/reporter.go
```diff
@@ -121,11 +121,11 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
 	failedIDs := storj.NodeIDList{}
 
 	for _, nodeID := range successNodeIDs {
-		_, err := reporter.statdb.Update(ctx, &statdb.UpdateRequest{
-			NodeID:       nodeID,
-			IsUp:         true,
-			AuditSuccess: true,
-		})
+		// _, err := reporter.statdb.Update(ctx, &statdb.UpdateRequest{
```
jenlij: here too

bootstrap/peer.go
egonelbre: Not really, I mean we do use it in tests however I don't know a good reason to keep it in `dht` and keep referring to it. Alternatively the `RoutingTable` interface would make more sense as an interface in kademlia package, rather than `dht`.

pkg/audit/verifier.go
```diff
@@ -17,8 +17,8 @@ import (
 	"storj.io/storj/internal/memory"
 	"storj.io/storj/pkg/auth/signing"
 	"storj.io/storj/pkg/identity"
-	"storj.io/storj/pkg/overlay"
 	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/statdb"
```
egonelbre: I was thinking about the whole change, and I think merging everything into overlay rather than `statdb` will be a much smaller PR and code will end up clearer. Overlay has much clearer and nicer name than `statdb`.

kaloyan-raev:
@jenlij This PR is not really ready for review. Github automatically requested all affected code owners for review.

pkg/audit/verifier.go
kaloyan-raev: Does it really matter if statdb would be merged to overlay or vice versa? There was an idea that the merged thing is called `nodedb`. I plan "statdb" to be renamed to whatever we decide as the last step.

kaloyan-raev: On second thought, I decided to follow your advice and start over in smaller steps. This PR became too big and quite messy. I have a new PR #1570 with the first step of just merging the statdb pkg to overlay pkg.

kaloyan-raev:
I am closing this PR has become quite big and messy and it is hard to merge the latest changes from master into it. I started by changing the Node protobufs first and then merging StatDB and Overlay, but it seems this was the harder approach. I am starting over the opposite way. First I will merge StatDB and Overlay and then I will change the Node protobufs. The first step is in PR #1570.

PR #492, opened 2018-10-17
[Experiment] TLS 1.3

egonelbre:
Do not merge. Just an experiment.

PR #1342, opened 2019-02-21
dial node/address with bad server cert

bryanchriswhite:
This PR fixes the indefinite hang in the case where the remote peer (server) has an invalid tls configuration. It also adds tests to `transport` for the cases where client/server tls configurations are in/valid.


bryanchriswhite:
I'm not sure how best to deal with the `FetchPeerIdentity` problem. Because of the use of a custom dialer, the grpc peer doesn't have `authInfo`. We receive it in the dialer function when calling `tlsConfig.ClientHandshake` but I'm not sure how best to persist it from there to the places where we currently need it; i.e. `FetchPeerIdentity`.

egonelbre:
I know this can be done via using custom credentials (see https://github.com/storj/storj/pull/492/files). Not sure whether there's an easier way.

pkg/transport/transport.go
```diff
@@ -64,9 +64,11 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 		return nil, Error.New("no address")
 	}
 
-	// add ID of node we are wanting to connect to
-	dialOpt := transport.tlsOpts.DialOption(node.Id)
-	options := append([]grpc.DialOption{dialOpt, grpc.WithBlock(), grpc.FailOnNonTempDialError(true)}, opts...)
+	options := append([]grpc.DialOption{
+		transport.tlsOpts.DialOption(ctx, node.Id),
+		grpc.WithBlock(),
+		grpc.FailOnNonTempDialError(true),
```
jtolds: should this still be here?

pkg/transport/transport.go
```diff
@@ -89,8 +91,11 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 func (transport *Transport) DialAddress(ctx context.Context, address string, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	dialOpt := transport.tlsOpts.DialOption(storj.NodeID{})
-	options := append([]grpc.DialOption{dialOpt, grpc.WithBlock(), grpc.FailOnNonTempDialError(true)}, opts...)
+	options := append([]grpc.DialOption{
+		transport.tlsOpts.DialOption(ctx, storj.NodeID{}),
+		grpc.WithBlock(),
+		grpc.FailOnNonTempDialError(true),
```
jtolds: same

pkg/transport/transport.go
bryanchriswhite: Yes

pkg/transport/transport.go
bryanchriswhite: We still need these for the same reasons as before. The change to transport credentials just ensures that we get a non-temporary error that will cause grpc to stop trying to reconnect.

pkg/peertls/tlsopts/tls.go
```diff
@@ -37,25 +45,46 @@ func (opts *Options) ServerOption() grpc.ServerOption {
 }
 
 // DialOption returns a grpc `DialOption` for making outgoing connections
-// to the node with this peer identity
-// id is an optional id of the node we are dialing
-func (opts *Options) DialOption(id storj.NodeID) grpc.DialOption {
+// to the node with this peer identity.
+// id is an optional id of the node we are dialing.
+func (opts *Options) DialOption(ctx context.Context, id storj.NodeID) grpc.DialOption {
```
bryanchriswhite: Don't need to receive `ctx`

pkg/peertls/tlsopts/tls.go
```diff
@@ -37,25 +45,46 @@ func (opts *Options) ServerOption() grpc.ServerOption {
...
 		),
 	}
+}
 
-	return grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig))
+// ClientHandshake returns a non-temporary error if an error is returned from
+// the underlying `ClientHandshake` call.
+func (tc *TransportCredentials) ClientHandshake(ctx context.Context, authority string, conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	tlsConn, authInfo, err := tc.TransportCredentials.ClientHandshake(ctx, authority, conn)
+	if err != nil {
```
bryanchriswhite: Maybe we should make sure this error is what we think it is to allow for proper functionality on the event of a temporary error (whatever that might be).

examples/grpc-debug/main.go
```diff
@@ -40,7 +40,7 @@ func main() {
 
 	dialOption := clientOptions.DialOption(storj.NodeID{})
 
-	conn, err := grpc.Dial(*targetAddr, dialOption)
+	conn, err := grpc.Dial(*targetAddr, dialOption, grpc.WithInsecure())
```
egonelbre: Add a note why it's fine to have `grpc.WithInsecure` here.

pkg/peertls/tlsopts/tls.go
```diff
@@ -15,8 +18,14 @@ import (
 	"storj.io/storj/pkg/storj"
 )
 
+// TransportCredentials wraps an embedded credentials.TransportCredentials.
+// It decorates the `ClientHandshake` method.
+type TransportCredentials struct {
```
egonelbre: Should this be in `pkg/transport` instead?

pkg/peertls/utils.go
```diff
@@ -22,6 +22,32 @@ import (
 	"storj.io/storj/pkg/pkcrypto"
 )
 
+// NonTemporaryError is an error with a `Temporary` method which always returns false.
+// It is intended for use with grpc.
+//
+// (see https://godoc.org/google.golang.org/grpc#WithDialer
+// and https://godoc.org/google.golang.org/grpc#FailOnNonTempDialError).
+type NonTemporaryError struct {
```
egonelbre: Same here, should this be in `pkg/transport`?

pkg/transport/transport_test.go
```diff
@@ -64,34 +86,140 @@ func TestDialNode(t *testing.T) {
...
+	defer ctx.Check(planet.Shutdown)
+
+	whitelistPath, err := planet.WriteWhitelist()
+	require.NoError(t, err)
+
+	planet.Start(ctx)
+
+	client := planet.StorageNodes[0].Transport
+	signedIdentities := testplanet.NewPregeneratedSignedIdentities()
+	ident, err := signedIdentities.NewIdentity()
```
egonelbre: `testplanet.PregeneratedSignedIdentity(0)`

pkg/transport/transport_test.go
```diff
@@ -27,11 +30,30 @@ func TestDialNode(t *testing.T) {
...
+	whitelistPath, err := planet.WriteWhitelist()
+	require.NoError(t, err)
+
 	planet.Start(ctx)
 
 	client := planet.StorageNodes[0].Transport
 
-	{ // DialNode with invalid targets
+	unsignedIdentities := testplanet.NewPregeneratedIdentities()
+	unsignedIdent, err := unsignedIdentities.NewIdentity()
```
egonelbre: `testplanet.PregeneratedIdentity(0)`

pkg/peertls/tlsopts/tls.go
```diff
@@ -15,8 +18,14 @@ import (
 	"storj.io/storj/pkg/storj"
 )
 
+// TransportCredentials wraps an embedded credentials.TransportCredentials.
+// It decorates the `ClientHandshake` method.
```
egonelbre: This comment could be better. Maybe `TransportCredentials wraps credential.TransportCredentials ensuring invalid certificates don't hang dialing.`

The information about embedding and decorating usually don't offer insight why something exists.

examples/grpc-debug/main.go
bryanchriswhite: This should be removed... good catch!

pkg/peertls/tlsopts/tls.go
bryanchriswhite: :+1:

pkg/transport/transport_test.go
bryanchriswhite: :+1:

pkg/transport/transport_test.go
bryanchriswhite: :+1:

bryanchriswhite: I will eventually stop doing this :p

pkg/peertls/utils.go
bryanchriswhite: :+1:

pkg/peertls/tlsopts/tls.go
bryanchriswhite: :+1:

pkg/peertls/utils.go
egonelbre: Can we implement this as 
```
type NonTemporaryError struct { error }
func (err NonTemporaryError) Temporary() bool { return false }
```

bryanchriswhite: yet another good idea

pkg/transport/transport.go
bryanchriswhite: The non-temporary error change has now been pushed into the peer certificate verification function wrapper.

pkg/transport/transport.go
jtolds: oh, sorry, for some reason i thought it was FailOnTempDialError

PR #1343, opened 2019-02-21
Add Storage Node IDs to Orderlimits (PBAs)

wthorp:
https://storjlabs.atlassian.net/browse/V3-1174

It sounds like this work will not be needed / long-term desirable based on changes coming to Orders and OrderLimits (to would prevent DDOS attacks, we shouldn't list all the storage nodes)


jenlij:
Should we try to get this merged anyway?

wthorp:
I'm leaning towards no

PR #1345, opened 2019-02-21
Adds dump-nodes and count capability to the overlay 

dylanlott:


pkg/overlay/cache.go
```diff
@@ -79,6 +79,16 @@ func (cache *Cache) Inspect(ctx context.Context) (storage.Keys, error) {
 	return nil, errors.New("not implemented")
 }
 
+// DumpNodes returns all of the nodes in the overlay cache for inspector use
+func (cache *Cache) DumpNodes(ctx context.Context) ([]*pb.Node, error) {
+	n, err := cache.db.List(ctx, storj.NodeID{}, 0)
```
kaloyan-raev: This will list only the first 1,000 nodes: https://github.com/storj/storj/blob/master/satellite/satellitedb/overlaycache.go#L134-L136

Is this the intention?

pkg/kademlia/routing.go
```diff
@@ -147,6 +147,39 @@ func (rt *RoutingTable) GetBucketIds() (storage.Keys, error) {
 	return kbuckets, nil
 }
 
+// DumpNodes loops returns all nodes in the nodeBucketDB and marshals them to &pb.Nodes
+func (rt *RoutingTable) DumpNodes() ([]*pb.Node, error) {
+	var nodes []*pb.Node
+
+	nodeKeys, err := rt.nodeBucketDB.List(nil, 0)
```
kaloyan-raev: Same here - the default list limit is 1,000 keys.

PR #1347, opened 2019-02-22
Storage node hash calculation on upload

mniewrzal:
Storage node is calculating hash of uploaded data and send it back to uplink with signature. 


pkg/piecestore/psserver/readerwriter.go
```diff
@@ -95,9 +100,18 @@ func (s *StreamReader) Read(b []byte) (int, error) {
...
 	n, err := s.src.Read(b)
 	s.sofar += int64(n)
+	if err != nil {
+		_, errHash := s.hash.Write(b[:n])
+		if errHash != nil {
+			return n, errs.Combine(err, errHash)
+		}
+		return n, err
+	}
+	_, err = s.hash.Write(b[:n])
```
kaloyan-raev: Can we simplify this change to something like this?
```go
_, errHash := s.hash.Write(b[:n])
err = errs.Combine(err, errHash)
```

mniewrzal: Yes, looks better

PR #1349, opened 2019-02-22
Support Data Science team with providing lost segment/piece info.

aligeti:
https://storjlabs.atlassian.net/browse/V3-1239


aligeti:
duplicate

PR #1351, opened 2019-02-22
Total file count through Monkit

aligeti:
https://storjlabs.atlassian.net/browse/V3-1238
https://storjlabs.atlassian.net/browse/V3-1239

### DIFF
--- a/pkg/accounting/tally/stats.go
+++ b/pkg/accounting/tally/stats.go
@@ -7,8 +7,8 @@ import (
 	"storj.io/storj/pkg/pb"
 )
 
-// Stats all stats fields
-type Stats struct {
+// stats all stats fields
+type stats struct {
 	Segments        int64
 	InlineSegments  int64
 	RemoteSegments  int64
@@ -24,7 +24,7 @@ type Stats struct {
 }
 
 // Combine aggregates all the stats
-func (s *Stats) Combine(o *Stats) {
+func (s *stats) Combine(o *stats) {
 	s.Segments += o.Segments
 	s.InlineSegments += o.InlineSegments
 	s.RemoteSegments += o.RemoteSegments
@@ -40,7 +40,7 @@ func (s *Stats) Combine(o *Stats) {
 }
 
 // AddSegment groups all the data based the passed pointer
-func (s *Stats) AddSegment(pointer *pb.Pointer, last bool) {
+func (s *stats) AddSegment(pointer *pb.Pointer, last bool) {
 	s.Segments++
 	switch pointer.GetType() {
 	case pb.Pointer_INLINE:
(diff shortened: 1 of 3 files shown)


pkg/datarepair/checker/checker.go
```diff
@@ -24,8 +25,9 @@ import (
 
 // Error is a standard error class for this package.
 var (
-	Error = errs.Class("checker error")
-	mon   = monkit.Package()
+	Error           = errs.Class("checker error")
+	mon             = monkit.Package()
+	totalFilesCount int64
```
egonelbre: This definitely shouldn't be a global variable.

pkg/accounting/tally/tally.go
```diff
@@ -101,15 +101,46 @@ func (t *Tally) calculateAtRestData(ctx context.Context) (latestTally time.Time,
...
+					bucketID := storj.JoinPaths(project, bucketName)
+
+					// paths are iterated in order, so everything in a bucket is
+					// iterated together. When a project or bucket changes,
+					// the previous bucket is completely finished.
+					if currentBucket != bucketID {
+						bucketCount++
+						if currentBucket != "" {
+							// report the previous bucket and add to the totals
+							currentBucketStats.Report("bucket.")
```
egonelbre: Do we want to submit each bucket as the same metric or separately?

pkg/accounting/tally/stats.go
```diff
@@ -0,0 +1,84 @@
...
+		case pb.Pointer_INLINE:
+			s.InlineFiles++
+		case pb.Pointer_REMOTE:
+			s.RemoteFiles++
+		}
+	}
+}
+
+// Report reports the stats thru monkit
+func (s *Stats) Report(prefix string) {
```
egonelbre: Add the dot inside here instead of supplying it from outside. Less likely to make a mistake.

pkg/accounting/tally/stats.go
```diff
@@ -0,0 +1,84 @@
...
+// See LICENSE for copying information.
+
+package tally
+
+import (
+	"storj.io/storj/pkg/pb"
+)
+
+//Stats all stats fields
+type Stats struct {
```
egonelbre: This probably doesn't need to be public.

pkg/accounting/tally/stats.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package tally
+
+import (
+	"storj.io/storj/pkg/pb"
+)
+
+//Stats all stats fields
```
egonelbre: nit: add a space after `//`

aligeti: done

pkg/accounting/tally/stats.go
aligeti: I thought about it.. but I am guessing going forward we might move this file out to Pkg,  just to handle more statistics. Should we have a separate pkg for statistics?

pkg/accounting/tally/tally.go
aligeti: Per my understanding we are currently not submitting any info based on bucket. we are aggregating the needed numbers (total segments, total files, total size of the files.... etc) while parsing thru pointerdb. Did this clarify the comment?

pkg/accounting/tally/stats.go
jtolds: no this is pretty tally specific. i agree it doesn't need to be public

aligeti: making it private

pkg/accounting/tally/tally.go
jtolds: oh definitely the same. basically, metric names should all be fixed by compile time. if metric names are dynamic like that the time series db will get overrun

pkg/accounting/tally/stats.go
egonelbre: Callers should use `Report("bucket")` instead of `Report("bucket.")`. Inside the func use `mon.IntVal(prefix + ".segments").Observe(s.Segments)`.

pkg/accounting/tally/tally.go
egonelbre: Is there a way to make monkit aware of the multiple buckets per call? So that it would collect summaries rather than the values? Currently it seems to be aggregating as `0:alpha, 0:beta, 0:gamma, 1:alpha, 1:beta, 1:gamma` rather than `0:[alpha, beta, gamma] 1:[alpha, beta, gamma]`. Not sure how much it makes a difference though.

PR #1354, opened 2019-02-25
Use SignedHash on client/uplink side

mniewrzal:
Changes:
* psclient receives storage node hash and compare it to own hash for verification
* uplink sends delete request when hashes don't match
* valid hashes are propagated up to segments.Store for future sending to satellite  

### DIFF
--- a/pkg/piecestore/psclient/readerwriter.go
+++ b/pkg/piecestore/psclient/readerwriter.go
@@ -77,7 +77,7 @@ func (s *StreamWriter) Close() error {
 
 	clientHash := s.hash.Sum(nil)
 	if bytes.Compare(reply.SignedHash.Hash, clientHash) != 0 {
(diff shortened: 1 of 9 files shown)


pkg/piecestore/psclient/readerwriter.go
```diff
@@ -14,12 +17,24 @@ import (
 	"storj.io/storj/pkg/utils"
 )
 
-// StreamWriter creates a StreamWriter for writing data to the piece store server
+// StreamWriter is astruct for writing data to the piece store server
```
egonelbre: typo... also maybe `StreamWriter handles uplink or satellite writing data to piecestore server.`... whether it's "a struct" is really not that important to be mentioned in documentation.

mniewrzal: done

PR #1355, opened 2019-02-25
Endpoint for local node info

kaloyan-raev:
Fixes https://storjlabs.atlassian.net/browse/V3-1240

Adds a new `Info` method to the Kademlia endpoint that returns the following local node info:
- ID
- Type
- Metadata (email and wallet)
- Restrictions (free storage and bandwidth)

The new endpoint is exposed as `inspector kad node-info` command too.


pkg/kademlia/dialer.go
```diff
@@ -109,6 +109,23 @@ func (dialer *Dialer) FetchPeerIdentity(ctx context.Context, target pb.Node) (pI
 	return identity.PeerIdentityFromPeer(p)
 }
 
+// Info connects to a node address and returns its node info.
+func (dialer *Dialer) Info(ctx context.Context, address *pb.NodeAddress) (*pb.InfoResponse, error) {
```
egonelbre: Just realizing that having `Dialer` with methods `FetchPeerIdentity` and `Info` seems weird, seems like it's being done this way because it's convenient rather than clear separation of concerns. `Dialer` was meant mainly for handling kademlia communication protocol, nothing else. We can think how to handle that later.

Also instead of `Info`, maybe use method name `FetchInfo` to make clearer that it's making a request.

pkg/kademlia/dialer.go
```diff
@@ -109,6 +109,23 @@ func (dialer *Dialer) FetchPeerIdentity(ctx context.Context, target pb.Node) (pI
...
+		return nil, context.Canceled
+	}
+	defer dialer.limit.Unlock()
+
+	conn, err := dialer.dialAddress(ctx, address)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := conn.client.Info(ctx, &pb.InfoRequest{})
```
egonelbre: Maybe `client.RequestInfo`? The GRPC methods seem to be verbs rather than nouns.

pkg/pb/overlay.proto
```diff
@@ -85,6 +86,16 @@ message QueryResponse {
 message PingRequest {};
 message PingResponse {};
 
+// TODO: add fields that validate who is requesting the info
+message InfoRequest {}
+
+message InfoResponse {
+    bytes id = 1 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
```
egonelbre: This shouldn't be here. We cannot trust the other side giving valid response. Use `PeerIdentityFromPeer` to get the identity.

pkg/pb/overlay.proto
```diff
@@ -85,6 +86,16 @@ message QueryResponse {
...
+    node.NodeType type = 2;
+    node.NodeMetadata metadata = 3;
+    node.NodeRestrictions restrictions = 4;
```
egonelbre: Is there a better name for this?

pkg/pb/overlay.proto
```diff
@@ -85,6 +86,16 @@ message QueryResponse {
 message PingRequest {};
 message PingResponse {};
 
+// TODO: add fields that validate who is requesting the info
+message InfoRequest {}
+
+message InfoResponse {
+    bytes id = 1 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
+    node.NodeType type = 2;
+    node.NodeMetadata metadata = 3;
```
egonelbre: Is there a better name for this? `NodeOperator` maybe?

pkg/pb/overlay.proto
kaloyan-raev: Maybe there is a better name, but `NodeRestrictions` is an already defined type from the past. It's just referenced here.

pkg/pb/overlay.proto
kaloyan-raev: Same here. The `NodeMetadata` is an already existing type. It is not defined in this PR.

pkg/pb/overlay.proto
egonelbre: I'm aware that these are already defined, it's more of we can improve the naming more easily right now, instead of later when it's integrated into elsewhere.

pkg/pb/overlay.proto
egonelbre: I'm aware that these are already defined, it's more of we can improve the naming more easily right now, instead of later when it's integrated into elsewhere.

pkg/pb/overlay.proto
egonelbre: Also completely fine to handle it later.

pkg/pb/overlay.proto
egonelbre: Also completely fine to handle it later.

pkg/pb/overlay.proto
kaloyan-raev: I don't understand. The ID is just one of the "info" properties returned by the node. So if the sender has just the address of the target node, it will receive the target's ID, type, metadata and restrictions. I don't understand how this relates with the peer identity.

egonelbre: Because the responder can write anything it wants to that field.

egonelbre: The only thing we can trust is the NodeID and certificate that it's using for communicating with us. Because, we can verify that it actually has the corresponding private key and certificate.

pkg/kademlia/dialer.go
kaloyan-raev: Changed in latest commit.

pkg/kademlia/dialer.go
kaloyan-raev: Changed in latest commit.

pkg/pb/overlay.proto
kaloyan-raev: OK. I see the issue. I will remove the ID from the response.

kaloyan-raev: Done.

pkg/pb/overlay.proto
kaloyan-raev: OK. I will define a new `NodeOperator` here and remove `NodeMetadata` in #1341.

pkg/pb/overlay.proto
kaloyan-raev: How about `NodeCapacity` instead of `NodeRestrictions`?

egonelbre: Yes, that's a really nice name.

pkg/kademlia/dialer.go
```diff
@@ -109,6 +109,23 @@ func (dialer *Dialer) FetchPeerIdentity(ctx context.Context, target pb.Node) (pI
...
+	defer dialer.limit.Unlock()
+
+	conn, err := dialer.dial(ctx, target)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := conn.client.RequestInfo(ctx, &pb.InfoRequest{})
+
+	return resp, errs.Combine(err, conn.disconnect())
```
egonelbre: Since we established connection we can still return the node id. So the method would be:

```
func (dialer *Dialer) FetchInfo(ctx context.Context, target pb.Node) (storj.NodeID, *pb.InfoResponse, error)

    p  := &peer.Peer{}
    pCall := grpc.Peer(p)
    resp, err = conn.client.RequestInfo(ctx, &pb.InfoRequest{}, pCall)
    id ... := identity.PeerIdentityFromPeer(p)
   return id, resp, ...
}
```

kaloyan-raev: OK. This sounds good. I will bring the Node ID back to the response.

pkg/pb/overlay.proto
kaloyan-raev: Done.

pkg/pb/overlay.proto
kaloyan-raev: Done.

pkg/pb/inspector.proto
```diff
@@ -110,6 +112,17 @@ message LookupNodeResponse {
   node.NodeMetadata meta = 2;
 }
 
+message NodeInfoRequest {
+  bytes id = 1 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
```
egonelbre: I'm confused. Why is this here?

egonelbre: ah, wait this is the kademlia thing.

pkg/kademlia/dialer.go
kaloyan-raev: Done.

pkg/pb/inspector.proto
kaloyan-raev: Initially it was only the address in the `NodeInfoRequest`. Then, when you asked to remove the ID from the response, I added the ID in the `NodeInfoRequest`, just for the sake for looking like other similar requests - `PingRequest` and `FetchPeerIdentity`.

I removed it again in the latest commit, because it does not makes sense to me. Having only the address is enough to complete the `NodeInfoRequest`. This may be helpful for other use cases where the caller knows only the address of the bootstrap node or the satellite and wants to find its ID using the `NodeInfoRequest`.

pkg/kademlia/dialer.go
```diff
@@ -109,6 +109,26 @@ func (dialer *Dialer) FetchPeerIdentity(ctx context.Context, target pb.Node) (pI
...
+
+	conn, err := dialer.dialAddress(ctx, address)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	p := &peer.Peer{}
+	pCall := grpc.Peer(p)
+	resp, err := conn.client.RequestInfo(ctx, &pb.InfoRequest{}, pCall)
+	id, idErr := identity.PeerIdentityFromPeer(p)
```
mniewrzal: Should we do this step if `RequestInfo` returns error?

kaloyan-raev: You are right. Fixed it in the latest commit.

PR #1090, opened 2019-01-18
Tests for identity and certificates

egonelbre:
Needs PR #1083

PR #1356, opened 2019-02-25
add testcontext.Compile

egonelbre:
Compiling binaries change from #1090


wthorp:
How is this going to be used?

egonelbre:
@wthorp you can see https://github.com/storj/storj/pull/1090/files#diff-c9ac549b9837d4a9aaa032e4ba4771aeR18 for the fuller example.

internal/testcontext/compile.go
```diff
@@ -0,0 +1,31 @@
...
+package testcontext
+
+import (
+	"os/exec"
+	"path"
+)
+
+// Compile compiles the specified package and returns the executable name.
+func (ctx *Context) Compile(pkg string) string {
+	ctx.test.Helper()
```
bryanchriswhite: What does this do?

internal/testcontext/compile.go
```diff
@@ -0,0 +1,31 @@
...
+
+	exe := ctx.File("build", path.Base(pkg)+".exe")
```
bryanchriswhite: Why a `.exe` extension? Does this not work on windows otherwise or something?

internal/testcontext/compile_norace.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+// +build !race
+
+package testcontext
+
+const raceEnabled = false
```
bryanchriswhite: How does this work: two constants with the same name and different values in the same package?

internal/testcontext/compile_norace.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+// +build !race
```
bryanchriswhite: What does this do?

internal/testcontext/compile.go
egonelbre: This marks the function as a test helper. Which means that it isn't listed when a test fails and instead points to the nearest non-test-helper.

This means when compiling fails it will point to the line calling `ctx.Compile` instead of `ctx.test.Error(string(out))` inside this function.

internal/testcontext/compile.go
egonelbre: Yup, Windows.

internal/testcontext/compile_norace.go
egonelbre: They have different build tags.

internal/testcontext/compile_norace.go
egonelbre: That is a build tag. 
* https://golang.org/pkg/go/build/#hdr-Build_Constraints
* https://dave.cheney.net/tag/build-tags

That `race` constraint is specifically for `-race`.

internal/testcontext/compile_norace.go
egonelbre: Just to clarify, the build tags in these files are exclusive. So one or the other will always be included, but never both at the same time. Hence it's fine to have the same name.

bryanchriswhite:
Cool - thanks for explaining!

PR #1359, opened 2019-02-25
Validate piece hash on satellite

mniewrzal:
The satellite receives pieces signed hashes in Pointer. If signed hash cannot be validated then piece is removed from Pointer and not saved in DB.


pkg/pb/pointerdb.proto
```diff
@@ -50,8 +50,10 @@ message RemoteSegment {
   // TODO: may want to use customtype and fixed-length byte slice
   string piece_id = 2;
   repeated RemotePiece remote_pieces = 3;
-
+  
   bytes merkle_root = 4; // root hash of the hashes of all of these pieces
+  
+  repeated piecestoreroutes.SignedHash remote_pieces_hashes = 5;
```
egonelbre: Wondering whether we can avoid storing the signed hashes in the database some how.

mniewrzal: Yeah, its something I'm concerned about too. I'm not sure but my understanding is that if we calculate merkle tree and save only merkle root in Pointer then we are able to use it to verify single piece hash.

kaloyan-raev: I have no idea if we can calculate merkle tree from the piece hashes. But if we keep with the Signed Hashes, they should be just another field of the `RemotePiece`, not a new array under `RemoteSegment`. I.e. Each `RemotePiece` has its `SignedHash` field.

mniewrzal: > they should be just another field of the `RemotePiece` 
Ok, makes totally make sense

My only concern with keeping all signed hashes is size of DB.

egonelbre: I don't think we need to keep the signature. As long as we have verified an incoming hash to be valid, it should be sufficient for storing.

mniewrzal: but next step will be using hashes in audi. Point 8 from Kaloyan `Prevent uplinks to corrupt storage nodes reputation` doc

kaloyan-raev: Makes sense. If validated once, the signature will be valid in future too. No sense to keep it. If not valid, we are not going to store the respective piece in the pointer at all.

kaloyan-raev: @mniewrzal Egon's point is that you only need to keep the hash in the Pointer, but not the signature.

mniewrzal: ah ok :) at some point I had version with just hashes and without storing signature and certs

mniewrzal: done, only hashes are stored

pkg/storage/segments/store.go
```diff
@@ -239,6 +243,7 @@ func makeRemotePointer(nodes []*pb.Node, rs eestream.RedundancyStrategy, pieceID
 		remotePieces = append(remotePieces, &pb.RemotePiece{
 			PieceNum: int32(i),
 			NodeId:   nodes[i].Id,
+			Hash:     hashes[i],
```
kaloyan-raev: Where does the uplink validates the hash before including it in the pointer? Or is it going to be a separate PR?

mniewrzal: It was done in previous PR - https://github.com/storj/storj/blob/81408a3c9e0637024d4c544252071b41492dc88a/pkg/piecestore/psclient/readerwriter.go#L74

pkg/pointerdb/pointerdb.go
```diff
@@ -100,6 +100,33 @@ func (s *Server) validateSegment(req *pb.PutRequest) error {
...
+		var remotePieces []*pb.RemotePiece
+		remote := pointer.Remote
+		for _, piece := range remote.RemotePieces {
+			err := auth.VerifyMsg(piece.Hash, piece.NodeId)
+			if err == nil {
+				piece.Hash.SetCerts(nil)
+				piece.Hash.SetSignature(nil)
+				remotePieces = append(remotePieces, piece)
+			} else {
+				s.logger.Warn("unable to verify piece hash: %v", zap.Error(err))
```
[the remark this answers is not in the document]
mniewrzal: TODO added, this is definitely another PR. It will be first case when satellite is using psclient to connect to storage node.

pkg/pointerdb/pointerdb.go
```diff
@@ -100,6 +100,33 @@ func (s *Server) validateSegment(req *pb.PutRequest) error {
...
 
+func (s *Server) filterValidPieces(pointer *pb.Pointer) error {
+	if pointer.Type == pb.Pointer_REMOTE {
+		var remotePieces []*pb.RemotePiece
+		remote := pointer.Remote
+		for _, piece := range remote.RemotePieces {
+			err := auth.VerifyMsg(piece.Hash, piece.NodeId)
+			if err == nil {
+				piece.Hash.SetCerts(nil)
+				piece.Hash.SetSignature(nil)
```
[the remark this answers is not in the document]
mniewrzal: done

pkg/storage/segments/store.go
kaloyan-raev: Cool. I missed it.

pkg/pointerdb/pointerdb.go
```diff
@@ -100,6 +100,35 @@ func (s *Server) validateSegment(req *pb.PutRequest) error {
...
 }
 
+func (s *Server) filterValidPieces(pointer *pb.Pointer) error {
+	if pointer.Type == pb.Pointer_REMOTE {
+		var remotePieces []*pb.RemotePiece
+		remote := pointer.Remote
+		for _, piece := range remote.RemotePieces {
+			err := auth.VerifyMsg(piece.Hash, piece.NodeId)
+			if err == nil {
+				// set to nil after verification to avoid storing in DB
```
egonelbre: Not liking that we are conflating the message and the way we store things... but I guess it's fine for now.

pkg/pointerdb/pointerdb_test.go
```diff
@@ -85,6 +97,55 @@ func TestServicePut(t *testing.T) {
...
+		assert.NoError(t, err)
+	}
+
+	// public key did not match expected signer
+	for i := numOfValidPieces; i < len(pieces); i++ {
+		identity, err := testidentity.NewTestIdentity(ctx)
+		assert.NoError(t, err)
+		pieces[i] = &pb.RemotePiece{
+			PieceNum: int32(i),
+			NodeId:   teststorj.NodeIDFromString(strconv.Itoa(i)),
```
egonelbre: `storj.NodeID{i}`

pkg/pointerdb/pointerdb_test.go
```diff
@@ -85,6 +97,55 @@ func TestServicePut(t *testing.T) {
...
+
+		_, err = rand.Read(pieces[i].Hash.Hash)
+		assert.NoError(t, err)
+		err = auth.SignMessage(pieces[i].Hash, *identity)
+		assert.NoError(t, err)
+	}
+
+	// public key did not match expected signer
+	for i := numOfValidPieces; i < len(pieces); i++ {
+		identity, err := testidentity.NewTestIdentity(ctx)
```
egonelbre: `testplanet.NewPregeneratedIdentities()` to avoid creating so many identities in a test.

PR #1361, opened 2019-02-25
some certificates cmd cleanup:

bryanchriswhite:
+ make claims flags consistent with other certificates sub-commands flags
+ move claim commands registration/binding into main
+ prevent overwrite fields on CA/Identity setup-configs from being
written to config.yaml
+ prevent overwtite flag on certificates config from being read from
config.yaml

### DIFF
--- a/pkg/certificates/config.go
+++ b/pkg/certificates/config.go
@@ -30,7 +30,7 @@ type CertClientConfig struct {
 
 // CertServerConfig is a config struct for use with a certificate signing service server
 type CertServerConfig struct {
-	Overwrite          bool   `default:"false" help:"if true, overwrites config AND authorization db is truncated"`
+	Overwrite          bool   `default:"false" help:"if true, overwrites config AND authorization db is truncated" setup:"true"`
 	AuthorizationDBURL string `default:"bolt://$CONFDIR/authorizations.db" help:"url to the certificate signing authorization database"`
 	MinDifficulty      uint   `default:"30" help:"minimum difficulty of the requester's identity required to claim an authorization"`
 	CA                 identity.FullCAConfig
--- a/pkg/identity/certificate_authority.go
+++ b/pkg/identity/certificate_authority.go
@@ -53,7 +53,7 @@ type CASetupConfig struct {
 	KeyPath        string `help:"path to the private key for this identity" default:"$IDENTITYDIR/ca.key"`
 	Difficulty     uint64 `help:"minimum difficulty for identity generation" default:"30"`
 	Timeout        string `help:"timeout for CA generation; golang duration string (0 no timeout)" default:"5m"`
-	Overwrite      bool   `help:"if true, existing CA certs AND keys will overwritten" default:"false"`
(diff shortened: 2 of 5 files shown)


pkg/identity/identity.go
```diff
@@ -58,7 +58,7 @@ type FullIdentity struct {
 type SetupConfig struct {
 	CertPath  string `help:"path to the certificate chain for this identity" default:"$IDENTITYDIR/identity.cert"`
 	KeyPath   string `help:"path to the private key for this identity" default:"$IDENTITYDIR/identity.key"`
-	Overwrite bool   `help:"if true, existing identity certs AND keys will overwritten for" default:"false"`
+	Overwrite bool   `help:"if true, existing identity certs AND keys will overwritten for" default:"false" setup:"true"`
```
egonelbre: Do we still want to support overwrite here?

pkg/identity/certificate_authority.go
```diff
@@ -53,7 +53,7 @@ type CASetupConfig struct {
 	KeyPath        string `help:"path to the private key for this identity" default:"$IDENTITYDIR/ca.key"`
 	Difficulty     uint64 `help:"minimum difficulty for identity generation" default:"30"`
 	Timeout        string `help:"timeout for CA generation; golang duration string (0 no timeout)" default:"5m"`
-	Overwrite      bool   `help:"if true, existing CA certs AND keys will overwritten" default:"false"`
+	Overwrite      bool   `help:"if true, existing CA certs AND keys will overwritten" default:"false" setup:"true"`
```
egonelbre: Do we still want to support overwrite here?

pkg/certificates/config.go
```diff
@@ -30,7 +30,7 @@ type CertClientConfig struct {
 
 // CertServerConfig is a config struct for use with a certificate signing service server
 type CertServerConfig struct {
-	Overwrite          bool   `default:"false" help:"if true, overwrites config AND authorization db is truncated"`
+	Overwrite          bool   `default:"false" help:"if true, overwrites config AND authorization db is truncated" setup:"true"`
```
egonelbre: Do we still want to support overwrite here?

pkg/identity/identity.go
bryanchriswhite: I think so but I don't have a strong opinion about it. I feel like it's best to make as minimal changes to these configs as possible until we have tests for the compiled binaries.

pkg/identity/certificate_authority.go
bryanchriswhite: ^

pkg/certificates/config.go
bryanchriswhite: ^

pkg/identity/identity.go
egonelbre: We removed overwrite from other tools. I think it's probably safer not to have them. Rather than have overwrite I would display a message: `X already exists, keeping old for safety. If you want to erase your X, please manually delete files "Y", "Z", "W"`.

PR #1364, opened 2019-02-26
Bwagreements not properly deleted

aligeti:
https://storjlabs.atlassian.net/browse/V3-1286

### DIFF
--- a/pkg/piecestore/psserver/agreementsender/agreementsender.go
+++ b/pkg/piecestore/psserver/agreementsender/agreementsender.go
@@ -102,7 +102,7 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID 
 			}
 		}
 		// Delete from PSDB by signature
-		if err = as.DB.DeleteBandwidthAllocationBySignature(agreement.Signature); err != nil {
+		if err = as.DB.DeleteBandwidthAllocationBySerialnum(rba.PayerAllocation.SerialNumber); err != nil {
 			as.log.Error("Agreementsender failed to delete bandwidth allocation", zap.Error(err))
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -102,7 +102,7 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
 			}
 		}
 		// Delete from PSDB by signature
-		if err = as.DB.DeleteBandwidthAllocationBySignature(agreement.Signature); err != nil {
+		if err = as.DB.DeleteBandwidthAllocationBySerialnum(rba.PayerAllocation.SerialNumber); err != nil {
```
wthorp: was there only one call to DeleteBandwidthAllocationBySignature?
if so, remove it as well.  you might check to see if there's anything else that's done by signature that could be leftover / wrong while you're at it

aligeti: updated

wthorp: Have we tested / analyzed that this actually solves the JIRA ticket, or are we just assuming it does?

aligeti: yes I have tested it. I made the agreement sender time to 5sec. Ie the agreement sender run every 5sec instead of 1hr. If everything works , the bwagreement-table should be empty without any entries, and saw the result as expected. The bandwidth_agrement table is empty. 

Also while agreements are being sent, I uploaded about 500MB file to see if it exits out.. or any errors happen. Expected out is nothing should happen and the bandwidth_agreement table should be empty ie all the agreeements are sent to satellite. Result was as expected.

PR #1365, opened 2019-02-26
uplink ttl

cam-a:
Add ability to specify ttl for uploads

https://storjlabs.atlassian.net/browse/V3-1189

### DIFF
--- a/cmd/uplink/cmd/cp.go
+++ b/cmd/uplink/cmd/cp.go
@@ -36,7 +36,7 @@ func init() {
 		RunE:  copyMain,
 	}, RootCmd)
 	progress = cpCmd.Flags().Bool("progress", true, "if true, show progress")
-	expires = cpCmd.Flags().String("expires", "", "optional expiration date of an object")
+	expires = cpCmd.Flags().String("expires", "", "optional expiration date of an object. Please use format (yyyy-mm-ddThh:mm:ssZhh:mm)")
 }
 
 // upload transfers src from local machine to s3 compatible object dst
--- a/cmd/uplink/cmd/cp.go
+++ b/cmd/uplink/cmd/cp.go
@@ -26,7 +26,7 @@ import (
 
 var (
 	progress *bool
-	ttl      *int
+	expires  *int
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


cmd/uplink/cmd/cp.go
```diff
@@ -34,6 +36,7 @@ func init() {
 		RunE:  copyMain,
 	}, RootCmd)
 	progress = cpCmd.Flags().Bool("progress", true, "if true, show progress")
+	ttl = cpCmd.Flags().Duration("ttl", 0, "number of days before file deletion. Default (0) is indefinite")
```
mniewrzal: Maybe we can name it `expiration` or something like that?

kaloyan-raev: or `days-to-live`, or `expires-in`.

cmd/uplink/cmd/cp.go
```diff
@@ -34,6 +36,7 @@ func init() {
 		RunE:  copyMain,
 	}, RootCmd)
 	progress = cpCmd.Flags().Bool("progress", true, "if true, show progress")
+	expires = cpCmd.Flags().Int("expires", 0, "number of days before file deletion. Default (0) is indefinite")
```
egonelbre: This probably should be specified as a `time.Duration` and such that you can specify `1yr`, `1day` etc.

cmd/uplink/cmd/cp.go
```diff
@@ -34,6 +36,7 @@ func init() {
 		RunE:  copyMain,
 	}, RootCmd)
 	progress = cpCmd.Flags().Bool("progress", true, "if true, show progress")
+	expires = cpCmd.Flags().String("expires", "", "optional expiration date of an object")
```
mniewrzal: can you add example date in a description?

cam-a:
Actually, I think I need to edit this still
From time package doc: 
"... when used with time.Parse they do not accept all the time formats permitted by the RFCs"

mniewrzal:
hmm if it supports major formats then I wouldn't worry at the moment. If main format like `2014-10-01T20:30:00Z` is working then its fine for me. You can merge as is and create jira ticket to improve situation later.

PR #1366, opened 2019-02-26
Adding dockerfile for running the web UI for Satellite

phutchins:


satellite/peer.go
```diff
@@ -181,6 +181,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 	var err error
 
 	{ // setup listener and server
+		fmt.Println("Starting listener and server")
```
kaloyan-raev: We have to remove these `fmt.Println`s before merging.

cmd/satellite/Dockerfile
```diff
@@ -2,15 +2,30 @@ ARG CGO_ENABLED=1
...
 EXPOSE 7777
+EXPOSE 10100
 WORKDIR /app
 VOLUME /root/.local/share/storj/satellite
+COPY /web/satellite /app
 COPY --from=build-env /app /app/satellite
-COPY cmd/satellite/entrypoint /entrypoint
+COPY /cmd/satellite/entrypoint /entrypoint
+COPY --from=satellite-ui /app/dist /app/dist
+COPY --from=satellite-ui /app/static /app/static
```
jtolds: nice, glad this worked

cmd/satellite/Dockerfile
```diff
@@ -2,15 +2,30 @@ ARG CGO_ENABLED=1
...
     CONF_PATH=/root/.local/share/storj/satellite \
-    BOOTSTRAP_ADDR=bootstrap.storj.io:8080
+    BOOTSTRAP_ADDR=bootstrap.storj.io:8080 \
+    STORJ_CONSOLE_STATIC_DIR=/app \
+    STORJ_CONSOLE_ADDRESS=0.0.0.0:10100
 EXPOSE 7777
+EXPOSE 10100
 WORKDIR /app
 VOLUME /root/.local/share/storj/satellite
+COPY /web/satellite /app
```
jtolds: wait, why is this needed?

cmd/storagenode/entrypoint
```diff
@@ -1,6 +1,14 @@
 #!/bin/sh
 set -euo pipefail
 
+if [[ ! -f "${IDENT_PATH}/identity.key" ]]; then
+  cd /tmp
+  wget https://storj-v3-alpha-builds.storage.googleapis.com/e82d336-go1.11/identity_linux_amd64.zip
+  unzip identity_linux_amd64.zip
+  mv identity_linux_amd64 identity
+  /tmp/identity create --difficulty 5 storagenode
```
jtolds: this concerns me for three reasons -

first, didn't we just copy the identity binary we compiled in?
secondly, this is way too low of a difficulty and identities shouldn't get generated automatically (ssh keys aren't).
third, if this works, this implies that we're not requiring signed certificates

cmd/satellite/Dockerfile
stefanbenten: you need the entire folder to have the frontend working
Thats what Jens and me got told from the Kiev Team

cmd/storagenode/entrypoint
stefanbenten: I think Bender is on it to do exactly that.

satellite/console/consoleweb/server.go
```diff
@@ -62,6 +63,8 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, list
 		listener: listener,
 	}
 
+	log.Printf("Starting Satellite UI...")
```
kaloyan-raev: You can use the `logger` that was passed to the constructor, i.e. `logger.Debug(...)`.

satellite/peer.go
```diff
@@ -181,6 +181,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 	var err error
 
 	{ // setup listener and server
+		zap.S().Debug("Starting listener and server")
```
kaloyan-raev: Sorry, I've just noticed the there is a logger `log` passed to the constructor. So the corrector logging should be done via `log.Debug(...)`.

cmd/storagenode/entrypoint
phutchins: Updated default difficulty to 30 with a way to configure it. This enables testing in dev environment until we take the time to add high difficulty certs and enforce signing.

satellite/console/consoleweb/server.go
phutchins: Updated.

kaloyan-raev: There is no `s` at this point. It should be `logger.Debug(...)` instead of `s.log.Debug(...)`.

satellite/console/consoleweb/server.go
```diff
@@ -6,6 +6,7 @@ package consoleweb
 import (
 	"context"
 	"encoding/json"
+	"log"
```
kaloyan-raev: This import has to be removed as it is not used.

satellite/peer.go
```diff
@@ -181,6 +181,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 	var err error
 
 	{ // setup listener and server
+		log.Debug("Starting listener and server")
```
egonelbre: Any particular reason these log lines were added? The following lines aren't actually starting them, they are just setting them up. Starting happens in a different place.

kaloyan-raev: @phutchins needed them to debug some issue and wanted to keep them for future debugging too.

stefanbenten: We encountered issues where the setup/startup hung due to incorrect paths in the config file.
As the satellite didnt print out at which step it was, it was very difficult to troubleshoot

PR #1368, opened 2019-02-26
Change payments to use statDB for wallet addresses rather than Overlay cache

dylanlott:


pkg/statdb/statdb.go
```diff
@@ -52,4 +52,12 @@ type NodeStats struct {
 	UptimeRatio        float64
 	UptimeSuccessCount int64
 	UptimeCount        int64
+	Wallet             string
+	Email              string
+}
+
+// Meta is a struct for organizing the meta data for a node operator
+type Meta struct {
```
kaloyan-raev: We have the `pb.NodeOperator` type with the same two fields. Is it really necessary to define this new type here?

pkg/statdb/statdb_test.go
```diff
@@ -50,7 +50,12 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
 			UptimeSuccessCount: currUptimeSuccess,
 		}
 
-		stats, err := sdb.Create(ctx, nodeID, nodeStats)
+		meta := &statdb.Meta{
+			Wallet: "",
+			Email:  "",
+		}
```
kaloyan-raev: `meta := &statdb.Meta{}` is an equivalent shorter notation.

cmd/satellite/payments.go
```diff
@@ -54,11 +54,14 @@ func generateCSV(ctx context.Context, start time.Time, end time.Time, output io.
 
 	for _, row := range rows {
 		nid := row.NodeID
-		wallet, err := db.OverlayCache().GetWalletAddress(ctx, nid)
+
+		// TODO: Get wallet address here from StatDB instead
```
kaloyan-raev: Is this TODO comment still valid?

dylanlott: Nope, this will get removed, good catch

pkg/statdb/statdb_test.go
dylanlott: I'll replace these

pkg/statdb/statdb.go
dylanlott: I didn't realize pb.NodeOperator had those fields on them. I can try to replace with pb.NodeOperator.

pkg/statdb/statdb.go
```diff
@@ -52,4 +52,12 @@ type NodeStats struct {
 	UptimeRatio        float64
 	UptimeSuccessCount int64
 	UptimeCount        int64
+	Wallet             string
+	Email              string
```
mobyvb: imo, it would be better to add a `Meta` struct here rather than putting Wallet/Email directly in the `NodeStats` struct. Then you'd access Wallet with `nodeStats.MetaInfo.Wallet` or something. But what you have works if you don't wanna do that

dylanlott: I ended up using pb.NodeOperator because it already has all the info we needed on it, I just didn't realize we already had that as an option.

pkg/statdb/statdb.go
```diff
@@ -52,4 +53,5 @@ type NodeStats struct {
 	UptimeRatio        float64
 	UptimeSuccessCount int64
 	UptimeCount        int64
+	Meta               pb.NodeOperator
```
kaloyan-raev: Can we have this field named `Operator` instead of `Meta`? "Meta" is too generic. When referenced in code `stats.Operator.Wallet` looks clearer than `stats.Meta.Wallet`.

pkg/statdb/inspector.go
```diff
@@ -41,6 +41,10 @@ func (srv *Inspector) CreateStats(ctx context.Context, req *pb.CreateStatsReques
 		AuditSuccessCount:  req.AuditSuccessCount,
 		UptimeCount:        req.UptimeCount,
 		UptimeSuccessCount: req.UptimeSuccessCount,
+		Meta: pb.NodeOperator{
+			Email:  "",
+			Wallet: "",
+		},
```
kaloyan-raev: This code change is unnecessary. It just initializes the `Meta` field with its default empty value, which will be done implicitly anyway.

pkg/statdb/statdb_test.go
```diff
@@ -115,6 +117,7 @@ func testDatabase(ctx context.Context, t *testing.T, sdb statdb.DB) {
 				AuditSuccessCount:  tt.auditSuccessCount,
 				UptimeCount:        tt.uptimeCount,
 				UptimeSuccessCount: tt.uptimeSuccessCount,
+				Meta:               pb.NodeOperator{},
```
kaloyan-raev: Same here - unnecessary initialization to the default empty value.

satellite/satellitedb/statdb.go
```diff
@@ -52,8 +58,12 @@ func (s *statDB) Create(ctx context.Context, nodeID storj.NodeID, startingStats
 		totalUptimeCount   int64
 		uptimeSuccessCount int64
 		uptimeRatio        float64
+		wallet             string
+		email              string
 	)
 
+	fmt.Printf("starting stats: %+v\n", startingStats)
```
kaloyan-raev: Don't forget to remove this or transform to debug log if still necessary.

kaloyan-raev:
There is something I still miss in this PR. OK, now the `generateCSV` takes the operator's wallet from statdb instead of from the overlay cache. But where does that wallet come from in first place to be added to the statdb? I'd expect to see some code calling the new node info endpoint that was added in #1355.

dylanlott:
@kaloyan-raev I intended this PR to be just the database changes and the setup necessary for the statDB to track wallet and email. In a separate PR I was going to update where the statDB adds a wallet for users it's tracking.

pkg/statdb/statdb_test.go
dylanlott: Done 👍

satellite/satellitedb/statdb.go
dylanlott: Removed 👍

pkg/statdb/statdb_test.go
kaloyan-raev: I still see `Operator:           pb.NodeOperator{},` in a few places.

kaloyan-raev:
@dylanlott Does this mean that if we merge this PR as is, `generateCSV` will not include any wallets, because there is no code to put them in statdb? If this is true, then merging this PR will make the master branch worse. Without this PR, the report includes at least the wallets of the nodes that are online, which is better then nothing.

egonelbre:
The problem is that we don't know whether the wallets are valid or not.

kaloyan-raev:
Yes. This is also true. My point is that I see little to no value in having another PR for adding the actual wallets in statdb.

kaloyan-raev:
Anyway, it looks like #1379 already removed the wallet from overlay cache.

pkg/discovery/service.go
```diff
@@ -158,6 +159,24 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 		if err != nil {
 			discovery.log.Error("could not put node into cache", zap.String("ID", ping.Id.String()), zap.Error(err))
 		}
+
+		// update email and wallet with correct info
+		_, info, err := discovery.kad.FetchInfo(ctx, &pb.NodeAddress{
+			Address: ping.GetAddress().String(),
+		})
```
kaloyan-raev: No need to create new `pb.NodeAddress`. You can use `ping.GetAddress()` directly, i.e.
```go
_, info, err := discovery.kad.FetchInfo(ctx, ping.GetAddress())
```

pkg/discovery/service.go
```diff
@@ -158,6 +159,24 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
...
+		if err != nil {
+			discovery.log.Warn("could not fetch node info", zap.String("ID", ping.GetAddress().String()))
+		}
+
+		_, err = discovery.statdb.UpdateOperator(ctx, ping.Id, &statdb.NodeStats{
```
kaloyan-raev: This call should happen only if `discovery.kad.FetchInfo` above was successful.

pkg/discovery/service.go
dylanlott: Fixed 👍

pkg/discovery/service.go
dylanlott: Fixed

pkg/statdb/statdb.go
```diff
@@ -26,6 +27,8 @@ type DB interface {
 	FindInvalidNodes(ctx context.Context, nodeIDs storj.NodeIDList, maxStats *NodeStats) (invalid storj.NodeIDList, err error)
 	// Update all parts of single storagenode's stats.
 	Update(ctx context.Context, request *UpdateRequest) (stats *NodeStats, err error)
+	// UpdateNodeStats takes a NodeStats and updates the applicable entries
```
kaloyan-raev: This comment should be updated to reflect the method.

pkg/statdb/statdb.go
```diff
@@ -26,6 +27,8 @@ type DB interface {
...
+	UpdateOperator(ctx context.Context, node storj.NodeID, updatedStats *NodeStats) (stats *NodeStats, err error)
```
kaloyan-raev: If we are updating just the operator fields here, we should have `updatedOperator *pb.NodeOperator` param instead of `updatedStats *NodeStats`.

kaloyan-raev: This is more about the param type instead of just the name.

satellite/satellitedb/testdata/postgres.v5.sql
```diff
@@ -0,0 +1,148 @@
...
+INSERT INTO "irreparabledbs" ("segmentpath", "segmentdetail", "pieces_lost_count", "seg_damaged_unix_sec", "repair_attempt_count") VALUES ('\x49616d5365676d656e746b6579696e666f30', '\x49616d5365676d656e7464657461696c696e666f30', 10, 1550159554, 10);
+INSERT INTO "injuredsegments" ("id", "info") VALUES (1, '\x0a0130120100');
+
+INSERT INTO "certrecords" VALUES (E'0Y0\\023\\006\\007*\\206H\\316=\\002\\001\\006\\010*\\206H\\316=\\003\\001\\007\\003B\\000\\004\\360\\267\\227\\377\\253u\\222\\337Y\\324C:GQ\\010\\277v\\010\\315D\\271\\333\\337.\\203\\023=C\\343\\014T%6\\027\\362?\\214\\326\\017U\\334\\000\\260\\224\\260J\\221\\304\\331F\\304\\221\\236zF,\\325\\326l\\215\\306\\365\\200\\022', E'L\\301|\\200\\247}F|1\\320\\232\\037n\\335\\241\\206\\244\\242\\207\\204.\\253\\357\\326\\352\\033Dt\\202`\\022\\325', '2019-02-14 08:07:31.335028+00');
```
egonelbre: Note here should be a few new data entries that contain the wallet and email such that the proceeding tests can verify whether wallet/email are properly handled.

PR #1371, opened 2019-02-27
Use sync2.Cycle in agreement sender

mniewrzal:
This change replaces `Ticker` with `sync2.Cycle` and introduce basic tests for sending agreements to satellite.


pkg/piecestore/psserver/agreementsender/agreementsender_test.go
```diff
@@ -0,0 +1,118 @@
...
+			assert.Equal(t, 1, len(allocPerSat))
+			putAllocation[node.ID()] = allocPerSat[0].Agreement.Total
+		}
+	}
+
+	for _, node := range planet.StorageNodes {
+		node.Agreements.Sender.Loop.Trigger()
+	}
+
+	time.Sleep(1 * time.Second)
```
egonelbre: Is this sleep necessary? There are quite a few methods on cycle that allows for it to be completed.

pkg/piecestore/psserver/agreementsender/agreementsender_test.go
```diff
@@ -0,0 +1,118 @@
...
+			// sum with previous upload PUTs
+			putAllocation[node.ID()] += allocPerSat[0].Agreement.Total
+		}
+	}
+
+	for _, node := range planet.StorageNodes {
+		node.Agreements.Sender.Loop.Trigger()
+	}
+
+	time.Sleep(1 * time.Second)
```
egonelbre: Same here, probably not necessary.

pkg/piecestore/psserver/agreementsender/agreementsender_test.go
mniewrzal: it fails locally without Sleep so most probably it will fail on travis

egonelbre: Use `TriggerWait`.

mniewrzal: ah I misunderstand earlier whats `TriggerWait` is doing. Now its working without sleep. Thanks

mniewrzal:
wont be applicable to new piecestore - #1422

PR #1372, opened 2019-02-27
change payment cmd name

JessicaGreben:
Rename `satellite report payments` command so that it is more clear that its reporting node usage data that will be used for payments.


cmd/satellite/main.go
```diff
@@ -60,9 +60,9 @@ var (
 		Short: "Generate a report",
 	}
 	paymentsCmd = &cobra.Command{
```
aleitner: maybe we should change variable names too?

JessicaGreben: yeah i was wondering that...if all the "payment" names should be updated to "usage" stuff.

PR #1374, opened 2019-02-27
Enhance the Storage Node to keep all BWA's for 90 days

aligeti:
https://storjlabs.atlassian.net/browse/V3-1296


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/alicebob/gopher-json v0.0.0-20180125190556-5a6b3ba71ee6 // indirect
 	github.com/alicebob/miniredis v0.0.0-20180911162847-3657542c8629
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da // indirect
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
```
egonelbre: why this dependency was added?

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -113,6 +113,7 @@ func (db *DB) Migration() *migrate.Migration {
 						`ALTER TABLE bandwidth_agreements ADD COLUMN total INT(10)`,
 						`ALTER TABLE bandwidth_agreements ADD COLUMN max_size INT(10)`,
 						`ALTER TABLE bandwidth_agreements ADD COLUMN created_utc_sec INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN status TEXT DEFAULT 'UNSENT'`,
```
mniewrzal: This probably should be `int` and we can have enum for all values.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -245,15 +248,46 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.Order) error {
...
+	}
+	return err
+}
+
+// UpdateBandwidthAllocationStatus update the bwa payout status
+func (db *DB) UpdateBandwidthAllocationStatus(serialnum string, status pb.AgreementsSummary_Status) (err error) {
+	defer db.locked()()
+
+	switch status {
+	case pb.AgreementsSummary_FAIL:
```
mniewrzal: Status `FAIL` is not equal to `REJECTED`. `FAIL` should be set when storage node tried sending agreement to satellite but for some reason it failed (failed connection, old satellite). Such `FAIL` should be sent once again with next iteration. Status `REJECTED` is a state when the satellite explicitly stated that this agreement is wrong (malformed signature, corrupted data, etc.) and such agreement shouldn't be sent once again.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -295,11 +329,11 @@ func (db *DB) GetBandwidthAllocationBySignature(signature []byte) ([]*pb.Order,
 	return agreements, nil
 }
 
-// GetBandwidthAllocations all bandwidth agreements and sorts by satellite
+// GetBandwidthAllocations all bandwidth agreements
 func (db *DB) GetBandwidthAllocations() (map[storj.NodeID][]*Agreement, error) {
 	defer db.locked()()
 
-	rows, err := db.DB.Query(`SELECT satellite, agreement FROM bandwidth_agreements`)
+	rows, err := db.DB.Query(`SELECT satellite, agreement FROM bandwidth_agreements WHERE status = "UNSENT" OR status = "RETRY"`)
```
mniewrzal: You don't have status `RETRY`

go.mod
aligeti: I am looking into it....no idea how it got in

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated with int values

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -245,15 +258,45 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.Order) error {
...
+	defer db.locked()()
+
+	switch status {
+	case pb.AgreementsSummary_OK:
+		//set update status to "SENT"
+		_, err = db.DB.Exec(`UPDATE bandwidth_agreements SET status = ? WHERE serial_num = ?`, bwaStatusSENT, serialnum)
+	default:
+		// received REJECT(due to bad signature or any data in rba changed) or
+		// FAIL(due to serialnum duplicates) from satellite, set the bwa status to "REJECT"
+		_, err = db.DB.Exec(`UPDATE bandwidth_agreements SET status = ? WHERE serial_num = ?`, bwaStatusREJECT, serialnum)
```
littleskunk: Rejected != Fail
The storage node should retry the failed bwagreements but don't retry the rejected bwagreement.

aligeti: updated ..  when 'fail' received from satellite, bwa status remains "unsent"

go.mod
mniewrzal: This is still not removed.

aligeti: I copied from master... not sure why is it getting added...fixing it right away

aligeti: I copied my go.mod and go.sum from the master and it shows as different even though the files are same...

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -88,22 +94,21 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
...
-			if err != nil || r.GetStatus() == pb.AgreementsSummary_FAIL {
-				as.log.Warn("Agreementsender failed to send agreement to satellite : will retry", zap.Error(err))
-				continue
-			} else if r.GetStatus() == pb.AgreementsSummary_REJECTED {
-				//todo: something better than a delete here?
-				as.log.Error("Agreementsender had agreement explicitly rejected by satellite : will delete", zap.Error(err))
+			// CASE 1: successful connection with sat established but either failed or reject recd
+			if sat.GetStatus() != pb.AgreementsSummary_OK {
+				// updates the bwa status to "REJECT"
+				_ = as.DB.UpdateBandwidthAllocationStatus(rba.PayerAllocation.SerialNumber, sat.GetStatus())
```
mniewrzal: You should log at least log error. Here and in next `_ = as.DB.UpdateBandwidthAllocationStatus`.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -245,15 +258,48 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.Order) error {
...
+	t := time.Now().Add(time.Hour * 24 * -90).Unix()
+	_, err := db.db.Exec(`DELETE FROM bandwidth_agreements WHERE created_utc_sec < ?`, t)
+	if err == sql.ErrNoRows {
+		err = nil
+	}
+	return err
+}
+
+// UpdateBandwidthAllocationStatus update the bwa payout status
+func (db *DB) UpdateBandwidthAllocationStatus(serialnum string, status pb.AgreementsSummary_Status) (err error) {
```
mniewrzal: DB shouldn't be aware of satellite statuses. As a parameter use `bwaStatus` (int) and do a switch in agreementsender.

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bmizerany/assert"
```
mniewrzal: We are using `"github.com/stretchr/testify/assert"`

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -88,22 +94,21 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
...
-			// Send agreement to satellite
-			r, err := client.BandwidthAgreements(ctx, &rba)
-			if err != nil || r.GetStatus() == pb.AgreementsSummary_FAIL {
-				as.log.Warn("Agreementsender failed to send agreement to satellite : will retry", zap.Error(err))
-				continue
-			} else if r.GetStatus() == pb.AgreementsSummary_REJECTED {
-				//todo: something better than a delete here?
-				as.log.Error("Agreementsender had agreement explicitly rejected by satellite : will delete", zap.Error(err))
+			// CASE 1: successful connection with sat established but either failed or reject recd
+			if sat.GetStatus() != pb.AgreementsSummary_OK {
```
cam-a: what about a switch statement instead? That way, we don't have to call updateBandwidthAllocation on FAIL if default status is UNSENT anyway

pkg/piecestore/psserver/agreementsender/agreementsender.go
aligeti: added log messages

pkg/piecestore/psserver/agreementsender/agreementsender.go
cam-a: or just continue on FAIL

pkg/piecestore/psserver/psdb/psdb_test.go
aligeti: thnx...updated

pkg/piecestore/psserver/agreementsender/agreementsender.go
cam-a: change `pb.AgreementsSummary_OK` to `pb.AgreementsSummary_REJECTED`

pkg/piecestore/psserver/psdb/psdb.go
aligeti: made bwaStatus type

pkg/piecestore/psserver/agreementsender/agreementsender.go
aligeti: updated with switch-case

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -88,22 +94,26 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
...
+			switch sat.GetStatus() {
+			case pb.AgreementsSummary_FAIL:
+				// CASE FAILED: connection with sat couldnt be established or connection
+				// is established but lost before receiving response from satellite
+				// no updates to the bwa status is done, so it remains "UNSENT"
+				as.log.Warn("Agreementsender lost connection", zap.Error(err))
+			default:
+				// CASE REJECTED: successful connection with sat established but either failed or reject recd
+				err = as.DB.UpdateBandwidthAllocationStatus(rba.PayerAllocation.SerialNumber, psdb.BwaStatusREJECT)
+				as.log.Warn("Agreementsender err", zap.Error(err))
```
mniewrzal: log only when err != nil

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -88,22 +94,26 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
...
 		if err != nil {
-			as.log.Warn("Agreementsender failed to deserialize agreement : will delete", zap.Error(err))
+			switch sat.GetStatus() {
+			case pb.AgreementsSummary_FAIL:
+				// CASE FAILED: connection with sat couldnt be established or connection
+				// is established but lost before receiving response from satellite
+				// no updates to the bwa status is done, so it remains "UNSENT"
+				as.log.Warn("Agreementsender lost connection", zap.Error(err))
+			default:
+				// CASE REJECTED: successful connection with sat established but either failed or reject recd
```
mniewrzal: `failed or reject recd`

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -30,6 +30,19 @@ var (
 	Error = errs.Class("psdb")
 )
 
+// BwaStatus keep tracks of the BWA payout status
+type BwaStatus int32
+
+const (
+	// BwaStatusUNSENT sets the BWA payout status to UNSENT
+	BwaStatusUNSENT BwaStatus = 0
```
mniewrzal: Use `iota` for this - https://github.com/storj/storj/blob/953d77e9af5179b001e74dc6976ef9e4af7aeb8b/internal/sync2/fence.go#L32

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -245,15 +261,48 @@ func (db *DB) WriteBandwidthAllocToDB(rba *pb.Order) error {
...
+		err = nil
+	}
+	return err
+}
+
+// UpdateBandwidthAllocationStatus update the bwa payout status
+func (db *DB) UpdateBandwidthAllocationStatus(serialnum string, status BwaStatus) (err error) {
+	defer db.locked()()
+
+	switch status {
```
mniewrzal: This switch is unnecessary. Just need ```_, err = db.db.Exec(`UPDATE bandwidth_agreements SET status = ? WHERE serial_num = ?`, status, serialnum)```

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -283,6 +284,38 @@ func TestHappyPath(t *testing.T) {
...
+		{serialnum: "serialnum_2", setStatus: psdb.BwaStatusREJECT, expStatus: psdb.BwaStatusREJECT},
+		{serialnum: "serialnum_3", setStatus: psdb.BwaStatusSENT, expStatus: psdb.BwaStatusSENT},
+	}
+	t.Run("UpdateBandwidthAllocationStatus", func(t *testing.T) {
+		for P := 0; P < concurrency; P++ {
+			t.Run("#"+strconv.Itoa(P), func(t *testing.T) {
+				t.Parallel()
+				for _, bw := range bwatests {
+					err := db.UpdateBandwidthAllocationStatus(bw.serialnum, bw.setStatus)
+					if err != nil {
```
mniewrzal: Use `require.NoError(t, err)`

pkg/piecestore/psserver/psdb/psdb_test.go
```diff
@@ -283,6 +284,38 @@ func TestHappyPath(t *testing.T) {
...
+						t.Fatal(err)
+					}
+					status, err := db.GetBwaStatusBySerialNum(bw.serialnum)
+					if err != nil {
```
mniewrzal: Use `require.NoError(t, err)`

pkg/piecestore/psserver/agreementsender/agreementsender.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb_test.go
aligeti: updated

pkg/piecestore/psserver/psdb/psdb_test.go
aligeti: updated

pkg/piecestore/psserver/agreementsender/agreementsender.go
aligeti: updated the comment spelling

jtolds:
I just read the ticket and perhaps this isn't the right place, but I am worried about performance issues of having old BWAs in the same table with the ones we care about. @egonelbre @mniewrzal @aleitner @aleitner do you think from a performance perspective it would be worth it to not add a new column but move old bwas to a different table? 

right now, when a file comes in, multiple RBWAs come in during the stream. i believe we keep the best in memory and then only write to sqlite at the end, but i don't think we respond to the client until we've written to disk. this means that the end-user uplink-side performance is visibly impacted the slower sqlite is at updating its tables and indexes. so we want as small of a table for BWAs as possible

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -113,6 +126,7 @@ func (db *DB) Migration() *migrate.Migration {
 						`ALTER TABLE bandwidth_agreements ADD COLUMN total INT(10)`,
 						`ALTER TABLE bandwidth_agreements ADD COLUMN max_size INT(10)`,
 						`ALTER TABLE bandwidth_agreements ADD COLUMN created_utc_sec INT(10)`,
+						`ALTER TABLE bandwidth_agreements ADD COLUMN status INT(10) DEFAULT 0`,
```
littleskunk: This DB migration will not be executed cause all storage nodes are running DB version 1 already. We need to move this to the next release which would be version 2.

mniewrzal: @aligeti You need to add next migration instead of modifying an existing one.

pkg/piecestore/psserver/psdb/testdata/sqlite.v1.sql
```diff
@@ -11,6 +11,7 @@ INSERT INTO bandwidth_agreements VALUES(
     2669568,
     0,
     1550666511,
+    0,
```
egonelbre: Older version files must not be changed.

egonelbre:
@jtolds, Yes only the best should be written to the disk, however there are several gotchas:
1. when receiving the first bandwidth allocation (order limit) must be written to the database immediately, this is needed so that we don't allow a single multiple uses for a single allocation, due to our current model
2. every new upload/download must verify to avoid reuse attacks... of course needs to verify signatures etc.
3. every next allocation must verify whether serial number hasn't been changed and that the signature is still valid
4. the best then can be finally written to database

The allocation must be kept in the database at least as long as the expiration date of the allocation... otherwise see reuse attacks.

jtolds:
@egonelbre okay, so, shoot, so your comment isn't even about the performance of the BWA table itself. i'm arguing old BWAs should go in a separate table to keep the primary BWA table fast and quick.

separately, we are not going to be able to beat s3 if we make the client wait for fsyncs on nodes. we should be able to benefit from the parallelism for redundancy. uplinks should not be slowed down by waiting for disk writes. 

specifically, the first allocation should not be written to disk immediately. for the security issue you describe, that should be tracked in memory and periodically batched to disk. ideally, the final allocation should also be batched to disk in the background. if we need a map of live serial numbers, that should be cached in ram.

it's super super important perf wise that sqlite transaction commits aren't in the hotpath of disk writes

egonelbre:
We need to check against all serial numbers that still have a valid expiration date. But, yes, that map could be held in memory.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -30,6 +30,19 @@ var (
 	Error = errs.Class("psdb")
 )
 
+// BwaStatus keep tracks of the BWA payout status
+type BwaStatus int32
+
+const (
+	// BwaStatusUNSENT sets the BWA status to UNSENT
+	BwaStatusUNSENT = iota
```
egonelbre: `AgreementStatusUnsent` etc.

pkg/piecestore/psserver/agreementsender/agreementsender.go
```diff
@@ -79,31 +85,37 @@ func (as *AgreementSender) SendAgreementsToSatellite(ctx context.Context, satID
...
+		sat, err := client.BandwidthAgreements(ctx, &rba)
 		if err != nil {
-			as.log.Warn("Agreementsender failed to deserialize agreement : will delete", zap.Error(err))
+			switch sat.GetStatus() {
+			case pb.AgreementsSummary_FAIL:
+				// CASE FAILED: connection with sat couldnt be established or connection
+				// is established but lost before receiving response from satellite
+				// no updates to the bwa status is done, so it remains "UNSENT"
+				as.log.Warn("Agreementsender lost connection", zap.Error(err))
+			default:
```
cam-a: I wonder if it's better to set default to FAIL rather than REJECTED

mniewrzal: hmm not sure here. if you feel strong about it I can change it.

cam-a: No strong feelings. I'm fine with merging as is

littleskunk:
Oh sorry. Everyone is waiting for my review. Let me run a test real quick.

littleskunk:
I have updated my storage node. DB migration was working. Looks good so far.

PR #1382, opened 2019-02-28
Cleanup errors handling around piece hash verification

mniewrzal:
Cleanup errors handling around piece hash verification.


pkg/piecestore/psclient/readerwriter.go
```diff
@@ -71,19 +71,23 @@ func (s *StreamWriter) Close() error {
 		return err
 	}
 
-	if err := auth.VerifyMsg(reply.SignedHash, s.signer.remoteID); err != nil {
+	s.storagenodeHash = reply.SignedHash
+
+	zap.S().Debugf("Stream close and recv summary: %v", reply)
```
egonelbre: Don't use global logger.

pkg/piecestore/psclient/client.go
```diff
@@ -163,13 +163,15 @@ func (ps *PieceStore) Put(ctx context.Context, id PieceID, data io.Reader, ttl t
...
-	if err == ErrHashDoesNotMatch {
-		return nil, errs.Combine(err, ps.Delete(ctx, id, rba.PayerAllocation.SatelliteId))
-	}
 	if err != nil && err != io.EOF {
 		return nil, ClientError.New("failure during closing writer: %v", err)
 	}
 
+	err = writer.Verify()
+	if err != nil {
+		return nil, errs.Combine(err, ps.Delete(ctx, id, rba.PayerAllocation.SatelliteId))
```
egonelbre: Once we have proper verification in place then this direct `Delete` will fail. Otherwise the uplink would be able to delete the remote piece directly and then claim that it never deleted, hence making storage node reputation worse.

pkg/piecestore/psclient/readerwriter.go
mniewrzal: Global loggers are unfortunately used widely around client/uplink code. To pass common log instance I would need to go and update things like segments.Store, ecclient.

egonelbre: Oh well, we'll fix that with other storage node fixes then.

pkg/piecestore/psclient/client.go
mniewrzal: Yes, you are right. At the moment I see three options:
* don't delete piece on storage node if hashes don't match
* send hash also from uplink to piecestore and storage node will be able to delete piece if hashes don't match
* send a specific message from uplink to let storage node know that should delete piece (inside upload loop) 
@kaloyan-raev

egonelbre: Not quite sure what to do about them.

Don't delete seems the easy option for now and let GC handle the rest.

> Send hash also from uplink to piecestore and storage node will be able to delete piece if hashes don't match. 

So this would be the first part of the handshake.

Uplink sends the final signed hash to storage node. Storage Node keeps it and responds with its own signed hash.

We could add an additional message for "HashMismatch" that is signed by uplink and storage node then must keep a database of them, such that it can respond to audits with the appropraite message. However this makes the whole communication much more complicated.

So adding an exhange of signed hashes seems the most appropriate.

kaloyan-raev: Oh, shoot! Yes, the uplink should not be able to delete pieces on storage nodes without first requesting a valid Order for this from the satellite.

I agree on this approach for now:
> Don't delete seems the easy option for now and let GC handle the rest.

Then this sounds promising as a next step:
> Uplink sends the final signed hash to storage node. Storage Node keeps it and responds with its own signed hash.

However, the storage node does not need to keep the signed hash from the uplink. It just needs to compare it with its own hash. If they match, the storage node will return its signed hash. Otherwise, it won't return the signed hash and the uplink won't be able to include that piece in the pointer. So the storage node should be safe.

Also I am not sure how would this addition round trip of signed hash exchanges would affect performance.

mniewrzal: Egon has also a different idea about why we should keep uplink hashes on storage node:
```
Thought: just now realized that maybe we don't have have to store the storage node signed piece hashes on the satellite side. As long as storage node has a signed hash by uplink. When auditing then satellite can ask for both.
```

For this PR I will just remove deleting from uplink.

kaloyan-raev:
@mniewrzal I think the PR title needs to be updated.

mniewrzal:
Done

PR #1384, opened 2019-02-28
Warn about permissions when creating identity

aleitner:


pkg/utils/utils.go
```diff
@@ -84,3 +86,31 @@ func discardNil(ch chan error) chan error {
...
+	if err != nil {
+		return false, err
+	}
+
+	if !info.IsDir() {
+		return false, errs.New("Path %s is not a directory", path)
+	}
+
+	// Check if the user bit is enabled in file permission
+	if info.Mode().Perm()&(1<<(uint(7))) == 0 {
```
egonelbre: maybe `if info.Mode.Perm() & 0200 == 0 {`?

pkg/utils/utils.go
```diff
@@ -84,3 +86,31 @@ func discardNil(ch chan error) chan error {
...
+		return false, errs.New("Write permission bit is not set on this file for user")
+	}
+
+	var stat syscall.Stat_t
```
egonelbre: All `syscall` things need to have appropriate build tags, because they are not cross-platform. Use os package instead (if possible), otherwise you need to write code for all other platforms.

aleitner: I made a function for windows but I wasn't sure how to do things like IFDEF in go

egonelbre: You need to create separate files for each of the different platforms. You can see the full list in https://golang.org/doc/install/source#environment. For posix compatible code you probably can use an appropriate build-tag.

egonelbre: Also before committing, it needs to be tested with Windows natively, Windows docker and Windows WSL.

aleitner: removed windows specific utils

pkg/utils/utils.go
aleitner: done

pkg/utils/utils.go
```diff
@@ -84,3 +85,22 @@ func discardNil(ch chan error) chan error {
...
+	if err != nil {
+		return false, err
+	}
+
+	if !info.IsDir() {
+		return false, errs.New("Path %s is not a directory", path)
+	}
+
+	// Check if the user bit is enabled in file permission
+	if info.Mode().Perm()&0200 == 0 {
```
wthorp: 0200 i magic number to me.  What does this mean?  What if my group has permissions but not me?

I have no idea if its correct, but what the did here is at least explicit: see https://stackoverflow.com/questions/28969455/golang-properly-instantiate-os-filemode

pkg/identity/utils.go
```diff
@@ -58,22 +58,36 @@ func writeFile(path string, dirmode, filemode os.FileMode, data []byte) error {
...
+		if os.IsNotExist(err) {
+			hasCert = false
+		} else {
+			return NoCertNoKey, err
+		}
+	}
 
 	_, err = os.Stat(keyPath)
-	hasKey := !os.IsNotExist(err)
+	hasKey := true
```
navillasa: nit - maybe this should just be set once below line 61?

cmd/certificates/setup.go
```diff
@@ -55,7 +55,11 @@ func cmdSetup(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if config.Identity.Status() != identity.CertKey {
+	status, err := config.Identity.Status()
+	if err != nil {
+		return err
+	}
+	if status != identity.CertKey {
```
navillasa: is there a way that identity.CertKey could be empty and still of that type? would that need to be checked here? (idk but just wondering. maybe that's something that could be handled somewhere else.)

cmd/identity/main.go
```diff
@@ -128,6 +137,11 @@ func cmdAuthorize(cmd *cobra.Command, args []string) error {
 	ctx := process.Ctx(cmd)
 
 	serviceDir := serviceDirectory(args[0])
+
+	if writable, err := utils.IsWritable(serviceDir); !writable && err != nil {
```
navillasa: if err != nil but it is writable, I think that err would still need to be returned

pkg/utils/utils.go
aleitner: Sure I can define a variable or I can just leave a nice little comment of what the number means on the line above!

cmd/identity/main.go
aleitner: Should've been `||` thanks

pkg/identity/utils.go
aleitner: done

stefanbenten:
> Got this problem during testing the authorize command:
> `identity --identity-dir="/Users/natalievillasana/notsudodir" authorize storagenode token`
> error occurred while signing certificate: rpc error: code = Unknown desc = authorization error: difficulty must be greater than: 29
> (identity files were still generated and saved, if you try again existing files will be loaded)

@navillasa That means your identity that you generated was below 30 difficulty. Did you generate it with the normal way `identity create storagenode` ?

navillasa:
> > Got this problem during testing the authorize command:
> > `identity --identity-dir="/Users/natalievillasana/notsudodir" authorize storagenode token`
> > error occurred while signing certificate: rpc error: code = Unknown desc = authorization error: difficulty must be greater than: 29
> > (identity files were still generated and saved, if you try again existing files will be loaded)
> 
> @navillasa That means your identity that you generated was below 30 difficulty. Did you generate it with the normal way `identity create storagenode` ?

It shouldn't matter because Alex's code should ensure that it errors with `permission denied` before getting to the point of checking id difficulty.

PR #1386, opened 2019-02-28
grpc: add a version number to all rpc method names

jtolds:
i don't know if this is a good idea, but the rationale is right now our rpc method names are weird. if you want to do a Ping request to a storage node over the DHT, the RPC method is currently "overlay.Nodes/Ping" which is weird, cause the overlay is not at all involved.

so we should change grpc package names anyway. we could get support for breaking-change supporting api versioning in the same move?


egonelbre:
google cloud recommends naming things as just `v1`, `v1alpha1` etc... (https://cloud.google.com/apis/design/versioning). Didn't find a recommendation about unstable apis, so used `v0_1` for now.

kaloyan-raev:
So with this change `overlay.Nodes/Ping` will become `v0_1.Nodes/Ping`?

egonelbre:
@kaloyan-raev Yes, although the `Nodes` name could be better... e.g. `Kademlia`.

kaloyan-raev:
I don't know... It's pinging a node, it's not pinging Kademlia.

egonelbre:
It's pinging the node and asking for kademlia to ping back. It's more about the kademlia protocol rather than anything else. But I do understand that view as well.

egonelbre:
Should we close this for now?

We do need to clear up the naming, however now I think we have a much clearer separation between things. We have "internal" and "external" API-s, which are probably nicer if they were covered by different requirements.

PR #1387, opened 2019-02-28
accounting: support null wallet addresses

jtolds:
problem is sometimes the wallet address is NULL and database/sql refuses to scan that into a `string` field


jenlij:
oops we should merge this asap it's breaking storage nodes

PR #1389, opened 2019-02-28
Identity versioning

bryanchriswhite:
### [Design doc](https://docs.google.com/document/d/1iEogpzIhRaX8t2v2CnDaI52T-2qurtXeCVgs_weRWFs) | [Jira ticket](https://storjlabs.atlassian.net/browse/V3-1188)

This PR adds the ability to version node identities. See the design doc for more info.


pkg/identity/certificate_authority_test.go
```diff
@@ -61,6 +71,9 @@ func TestFullCertificateAuthority_NewIdentity(t *testing.T) {
 }
 
 func TestFullCertificateAuthority_Sign(t *testing.T) {
+	// TODO: fix extension serialization
+	t.Skipf("certificate extension serialization fix required")
```
coyle: Is there a ticket for this ?

pkg/identity/identity_test.go
```diff
@@ -87,50 +92,129 @@ func TestFullIdentityFromPEM(t *testing.T) {
...
-		err := ic.Save(fi)
-		assert.NoError(t, err)
+		{ // test saving
+			err := identCfg.Save(ident)
+			assert.NoError(t, err)
 
-		certInfo, err := os.Stat(ic.CertPath)
-		assert.NoError(t, err)
+			certInfo, err := os.Stat(identCfg.CertPath)
+			assert.NoError(t, err)
```
coyle: we should probably use [require.NoError](https://godoc.org/github.com/stretchr/testify/require#NoError) instead of assert.NoError

pkg/storj/node.go
```diff
@@ -184,3 +230,16 @@ func (n NodeIDList) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
 
 // Less implements sort.Interface.Less()
 func (n NodeIDList) Less(i, j int) bool { return n[i].Less(n[j]) }
+
+func (id NodeID) versionByte() byte {
+	return id[NodeIDSize-1]
```
coyle: is it possible NodeIDSize <= 0 ? or larger than len(id)+1 ?

bryanchriswhite: the length of `id` is `NodeIDSize`

pkg/identity/certificate_authority_test.go
bryanchriswhite: nice catch - that's already done

pkg/identity/identity_test.go
bryanchriswhite: :+1:

bryanchriswhite: I updated several other places where we should be using `require` instead of `assert`. I think using `require` here might not be necessary, but I could be wrong. There are two tests here that are sharing a `*testing.T`.

pkg/identity/certificate_authority.go
```diff
@@ -49,6 +48,7 @@ type FullCertificateAuthority struct {
 
 // CASetupConfig is for creating a CA
 type CASetupConfig struct {
+	VersionNumber  uint   `default:"0" help:"which identity version to use (0 is latest)"`
```
mniewrzal: nit: can you move `default` to the end?

cmd/identity/batch.go
```diff
@@ -40,6 +40,8 @@ var (
 	}
 
 	keyCfg struct {
+		// TODO: where is this used and should it be conistent with "latest" alias?
+		VersionNumber uint   `default:"0" help:"version of identity (0 is latest)"`
```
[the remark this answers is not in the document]
bryanchriswhite: I agree that we should be consistent but I think it's better to have the default first because the help text is long and it makes it inconvenient to look up what the default is or if there even is one.

internal/testidentity/pregenerated_identities_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls"
+	"storj.io/storj/pkg/storj"
+)
+
+func TestPregeneratedIdentity(t *testing.T) {
+	IdentityVersionsTest(t, func(t *testing.T, version storj.IDVersion, ident *identity.FullIdentity) {
+		assert.Equal(t, version.Number, ident.ID.Version().Number)
```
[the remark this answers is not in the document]
bryanchriswhite: I think in this case it makes sense to continue checking other assertions if the version isn't equal. I would prefer to know about multiple failed assertions when running the test rather than having to find out one at a time as I fix a failing test. Because the version being wrong doesn't cause a panic or an unexpected error I think it's reasonable to allow the test to continue.

internal/testpeertls/certificates.go
```diff
@@ -8,22 +8,34 @@ import (
...
+		return nil, nil, err
+	}
+
 	for i := 0; i < length; i++ {
 		key, err := pkcrypto.GeneratePrivateKey()
 		if err != nil {
 			return nil, nil, err
 		}
-		keys = append(keys, key)
+		keys = append([]crypto.PrivateKey{key}, keys...)
```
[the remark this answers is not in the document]
bryanchriswhite: I changed the order to be consistent with the certificate chain. Previously the key indexes were backwards from the certificate indexes; i.e. keys[0] was the private key for the certificate at certs[1] (in the case of a 2 certificate chain).

internal/testpeertls/certificates_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func TestNewCertChain(t *testing.T) {
+	testidentity.CompleteIdentityVersionsTest(t, func(t *testing.T, version storj.IDVersion, ident *identity.FullIdentity) {
+		for length := 2; length < 4; length++ {
+			t.Logf("length: %d", length)
+			keys, chain, err := NewCertChain(length, version.Number)
+			require.NoError(t, err)
+
+			assert.Len(t, chain, length)
```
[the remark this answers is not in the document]
bryanchriswhite: (discussed above)

pkg/identity/certificate_authority.go
bryanchriswhite: (discussed above)

PR #1390, opened 2019-03-01
sqlfix

littleskunk:


egonelbre:
Fix already exists in github.com/storj/storj/pull/1380

PR #1394, opened 2019-03-01
Use DialNode for requesting node info

kaloyan-raev:
The Local Node Info endpoint introduced in #1355 was implemented over the Transport Client's `DialAddress` instead of `DialNode`. This made is less secure as `DialAddress` does not validate if the target node is really the one with the requested Node ID.

This PR fixes the issue by using the `DialNode` method.


pkg/kademlia/dialer.go
```diff
@@ -110,28 +110,20 @@ func (dialer *Dialer) FetchPeerIdentity(ctx context.Context, target pb.Node) (pI
 }
 
 // FetchInfo connects to a node address and returns its node info.
```
egonelbre: comment needs updating

PR #1396, opened 2019-03-03
Reduce Log Messages on Production Settings

stefanbenten:
This log messages produces Gigabytes of log, that arent super necessary right now


pkg/accounting/tally/tally.go
```diff
@@ -163,7 +163,7 @@ func (t *Tally) calculateAtRestData(ctx context.Context) (latestTally time.Time,
 				}
 				pieceSize := segmentSize / int64(minReq)
 				for _, piece := range pieces {
-					t.logger.Info("found piece on Node ID" + piece.NodeId.String())
+					t.logger.Debug("found piece on Node ID" + piece.NodeId.String())
```
egonelbre: I would remove this completely.

stefanbenten: Done

PR #1398, opened 2019-03-04
add private listener to grpc server

JessicaGreben:
We want to run a private GPRC server in addition to the public GRPC server so that certain services are available on localhost instead of on the internet. For example, inspectors, debug, and dashboard can run on localhost.

To accomplish that, this PR includes the following changes:
- `server.New` now creates a private GRPC server (in addition to the original public server)
- there is a new method to `DialAddressInsecure` which creates an insecure connection to a node without tls
- the inspector clients now use this new `DialAddressInsecure` function to make inspector grpc calls
- the dashboard functionality now uses this new `DialAddressInsecure` function to make grpc calls


storagenode/peer.go
```diff
@@ -259,5 +249,8 @@ func (peer *Peer) ID() storj.NodeID { return peer.Identity.ID }
 // Local returns the peer local node info.
 func (peer *Peer) Local() pb.Node { return peer.Kademlia.RoutingTable.Local() }
 
-// Addr returns the public address.
-func (peer *Peer) Addr() string { return peer.Public.Server.Addr().String() }
+// PublicAddr returns the public address.
+func (peer *Peer) PublicAddr() string { return peer.Server.PublicAddr().String() }
```
egonelbre: Your life might be easier if you don't change this method name just yet. A second PR renaming frequent methods can make PR review faster.

pkg/server/server.go
```diff
@@ -53,8 +65,17 @@ func New(opts *tlsopts.Options, lis net.Listener, interceptor grpc.UnaryServerIn
...
-// Addr returns the server's listener address
-func (p *Server) Addr() net.Addr { return p.lis.Addr() }
+// PublicAddr returns the server's listener address
+func (p *Server) PublicAddr() net.Addr { return p.publicListener.Addr() }
+
+// PrivateAddr returns the server's listener address
+func (p *Server) PrivateAddr() net.Addr { return p.privateListener.Addr() }
+
+// PublicListener returns the server's public listener
+func (p *Server) PublicListener() net.Listener { return p.publicListener }
```
egonelbre: I think you won't need to expose the listeners. Exposing the appropriate GRPC would be sufficient.

pkg/server/server.go
```diff
@@ -24,22 +24,34 @@ type Service interface {
 // Server represents a bundle of services defined by a specific ID.
 // Examples of servers are the satellite, the storagenode, and the uplink.
 type Server struct {
-	lis      net.Listener
-	grpc     *grpc.Server
-	next     []Service
-	identity *identity.FullIdentity
+	publicListener  net.Listener
+	privateListener net.Listener
+	grpc            *grpc.Server
```
egonelbre: You'll need to add public grpc / private grpc separately.

One nice way to organize these is to have:

```
public struct {
	listener net.Listener
	grpc     *grpc.Server
}
private struct {
	listener net.Listener
	grpc     *grpc.Server
}
```

pkg/server/server.go
```diff
@@ -24,22 +24,34 @@ type Service interface {
...
 	}
 
+	pubLis, err := net.Listen("tcp", pubAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	privLis, err := net.Listen("tcp", privAddr)
+	if err != nil {
+		return nil, err
```
egonelbre: `pubLis` is not closed properly when there's an error.

pkg/server/server.go
```diff
@@ -21,30 +22,62 @@ type Service interface {
...
+	privateListener, err := net.Listen("tcp", privateAddr)
+	if err != nil {
+		return nil, errs.Combine(err, privateListener.Close())
+	}
+	private := private{
+		listener: privateListener,
+		grpc: grpc.NewServer(
+			grpc.StreamInterceptor(streamInterceptor),
+			grpc.UnaryInterceptor(unaryInterceptor),
+			opts.ServerOption(),
```
jtolds: you probably don't need all of our security options for the private listener, since it is supposed to only listen on localhost anyway. there's a grpc insecure option to make it listen on plain http which could be good for local curl debugging

satellite/peer.go
```diff
@@ -417,8 +407,10 @@ func (peer *Peer) Run(ctx context.Context) error {
 	})
 	group.Go(func() error {
 		// TODO: move the message into Server instead
-		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.Public.Server.Addr().String())
-		return ignoreCancel(peer.Public.Server.Run(ctx))
+		// Don't change the format of this comment, it is used to figure out the node id.
+		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.Addr())
+		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.PrivateAddr())
```
jtolds: maybe this log message could distinguish private vs public

storagenode/peer.go
```diff
@@ -209,8 +198,10 @@ func (peer *Peer) Run(ctx context.Context) error {
 	})
 	group.Go(func() error {
 		// TODO: move the message into Server instead
-		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.Public.Server.Addr().String())
-		return ignoreCancel(peer.Public.Server.Run(ctx))
+		// Don't change the format of this comment, it is used to figure out the node id.
+		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.Addr())
+		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.PrivateAddr())
```
jtolds: same

pkg/server/server.go
JessicaGreben: good point. let me look into which options we need and remove the unnecessary ones

satellite/peer.go
JessicaGreben: good idea. I will add public vs private

pkg/transport/transport.go
```diff
@@ -116,6 +124,32 @@ func (transport *Transport) DialAddress(ctx context.Context, address string, opt
 	return conn, Error.Wrap(err)
 }
 
+// DialNodeInsecure returns an insecure grpc connection without tls to a node.
+//
+// Use this method for communication with localhost. For example, with the inspector or debugging services.
+// Otherwise in most cases DialNode should be used for communicating with nodes since it is secure.
+func (transport *Transport) DialNodeInsecure(ctx context.Context, node *pb.Node, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
```
jtolds: since this doesn't even  node node id checking, you could change this to be DialAddressInsecure for the same unencrypted effect

pkg/certificates/certificates_test.go
```diff
@@ -616,16 +616,12 @@ func TestCertificateSigner_Sign_E2E(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, serverIdent)
 
-	listener, err := net.Listen("tcp", "127.0.0.1:0")
-	require.NoError(t, err)
-	require.NotNil(t, listener)
-
-	serverConfig := server.Config{Address: listener.Addr().String()}
-	opts, err := tlsopts.NewOptions(serverIdent, serverConfig.Config)
+	sc := server.Config{Address: "127.0.0.1:0", PrivateAddress: "127.0.0.1:7778"}
```
egonelbre: Tests shouldn't use designated ports, they developers have all kinds of stuff running on their local system and it's likely to conflict with something. Use `127.0.0.1:0`.

cmd/storj-sim/network.go
```diff
@@ -112,14 +112,17 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
-		host             = flags.Host
-		gatewayPort      = 9000
-		bootstrapPort    = 9999
-		satellitePort    = 10000
-		storageNodePort  = 11000
-		consolePort      = 10100
-		bootstrapWebPort = 10010
+		configDir              = flags.Directory
+		host                   = flags.Host
+		gatewayPort            = 9000
```
egonelbre: I've been wondering whether there's a nice consistent way how to assign these ports. Currently they seem quite messy and disjointed. Maybe everything something like bootstrap ports start at `9900`, satellites start at `10XXY`, where XX is the satellite index and Y is a subsystem (gateway at `10XX9`), storage nodes at `11XXY`? Or something like that? No need to do in this PR, but just something to think about.

pkg/server/server.go
```diff
@@ -21,30 +22,58 @@ type Service interface {
...
 	unaryInterceptor := unaryInterceptor
 	if interceptor != nil {
 		unaryInterceptor = combineInterceptors(unaryInterceptor, interceptor)
 	}
 
-	return &Server{
-		lis: lis,
+	publicListener, err := net.Listen("tcp", publicAddr)
+	if err != nil {
+		return nil, errs.Combine(err, publicListener.Close())
```
egonelbre: If there's an error with listening it won't require closing the returned thing

pkg/server/server.go
```diff
@@ -21,30 +22,58 @@ type Service interface {
...
 		grpc: grpc.NewServer(
 			grpc.StreamInterceptor(streamInterceptor),
 			grpc.UnaryInterceptor(unaryInterceptor),
 			opts.ServerOption(),
 		),
+	}
+
+	privateListener, err := net.Listen("tcp", privateAddr)
+	if err != nil {
+		return nil, errs.Combine(err, privateListener.Close())
```
egonelbre: here you would need to close publicListener, not privateListener

bootstrap/peer.go
```diff
@@ -194,8 +184,11 @@ func (peer *Peer) Run(ctx context.Context) error {
 	})
 	group.Go(func() error {
 		// TODO: move the message into Server instead
-		peer.Log.Sugar().Infof("Node %s started on %s", peer.Identity.ID, peer.Public.Server.Addr().String())
-		return ignoreCancel(peer.Public.Server.Run(ctx))
+		// Don't change the format of this comment, it is used to figure out the node id.
+		peer.Log.Sugar().Infof("Node %s", peer.Identity.ID)
```
egonelbre: Does storj-sim still show node id-s in log? I suspect this might have broken due to the format change.

JessicaGreben: sorry, I tried to see how storj-sim was parsing and thought this syntax was fine.  Let me check again and I can revert the comment change.

egonelbre: storj-sim uses a really stupid approach. it chceks whether log line starts with "Node " and searches until " started". So if you add ` started` after the `%s` it should work.

JessicaGreben: ah ok, I will make that change.

PR #1403, opened 2019-03-04
[V3-1147] Ensure certificate validation happens properly

bryanchriswhite:
https://storjlabs.atlassian.net/browse/V3-1147


pkg/peertls/tlsopts/options.go
```diff
@@ -64,30 +71,58 @@ func (opts *Options) configure(c Config) (err error) {
...
 		if err != nil {
 			return err
 		}
-		pcvs = append(pcvs, identity.VerifyUnrevokedChainFunc(opts.RevDB))
+		opts.VerificationFuncs.Add(identity.VerifyUnrevokedChainFunc(opts.RevDB))
 	}
 
 	exts := peertls.ParseExtensions(c.Extensions, parseOpts)
-	pcvs = append(pcvs, exts.VerifyFunc())
+	opts.VerificationFuncs.Add(exts.VerifyFunc())
```
bryanchriswhite: We might want to propagate the client/server separation into the extension handling in the future but nothing at the moment requires this.

pkg/peertls/tlsopts/options.go
```diff
@@ -64,30 +71,58 @@ func (opts *Options) configure(c Config) (err error) {
...
-	for i, f := range pcvs {
+// Add adds verification functions so the client and server lists.
+func (vf *VerificationFuncs) Add(verificationFuncs ...peertls.PeerCertVerificationFunc) {
+	vf.ClientAdd(verificationFuncs...)
+	vf.ServerAdd(verificationFuncs...)
+}
+
+// ClientAdd adds verification functions so the client list.
+func (vf *VerificationFuncs) ClientAdd(verificationFuncs ...peertls.PeerCertVerificationFunc) {
+	verificationFuncs = removeNils(verificationFuncs)
```
navillasa: why do the nil funcs need to be removed? when would they usually be set to nil?

pkg/peertls/tlsopts/options_test.go
```diff
@@ -103,7 +106,48 @@ func TestNewOptions(t *testing.T) {
...
+	testCases := []struct {
+		name   string
+		identF identFunc
+	}{
+		{"unsigned client identity", testplanet.PregeneratedIdentity},
+		{"signed client identity", testplanet.PregeneratedSignedIdentity},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			ident, err := testCase.identF(0)
```
navillasa: maybe doesn't need to be tested here in this test necessarily, but should `ident` also be checked to make sure it has appropriate fields set?

pkg/peertls/tlsopts/options.go
bryanchriswhite: some of the factories for verification functions like the one for peer CA whitelist verification return nil instead of a function if the whitelist doesn't exist or is empty or something like that.

coyle: maybe add a comment in the code about why?

PR #1404, opened 2019-03-04
allocates 1kb for audit bandwidth agreements instead of 256kb

navillasa:


pkg/audit/verifier.go
```diff
@@ -67,7 +67,8 @@ func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize
 		return s, Error.New("no node returned from overlay for piece %s", id.String())
 	}
 	fromNode.Type.DPanicOnInvalid("audit getShare")
-	ps, err := psclient.NewPSClient(ctx, d.transport, fromNode, 0)
+	// TODO(nat): figure out how to make allocated size for audits 1024 without this division
+	ps, err := psclient.NewPSClient(ctx, d.transport, fromNode, shareSize/8)
```
kaloyan-raev: Could you explain in the comment above why dividing by 8, so it does not look like a magic number?

pkg/audit/verifier.go
```diff
@@ -67,7 +67,10 @@ func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize
 		return s, Error.New("no node returned from overlay for piece %s", id.String())
 	}
 	fromNode.Type.DPanicOnInvalid("audit getShare")
-	ps, err := psclient.NewPSClient(ctx, d.transport, fromNode, 0)
+	// TODO(nat): the reason for dividing by 8 is because later in psclient/readerwriter.go
+	// the bandwidthMsgSize is arbitrarily multiplied by 8 as a reasonable threshold
```
aleitner: arbitrarily!?

navillasa: @egonelbre 's words!

PR #1405, opened 2019-03-05
V3-1284 Node bootstrap web frontend

crawter:


web/bootstrap/src/views/Failure.vue
```diff
@@ -0,0 +1,162 @@
...
+	</div>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator';
+
+@Component({
+    mounted() {
+        setTimeout(() => {
+            (document as any).querySelector('.failure-container').classList.add('active');
```
Barterio: Is it possible to be null at any stage?

crawter: no, i won't return null

PR #1407, opened 2019-03-05
convert psserver dashboard into an inspector

JessicaGreben:
We want to run the storagenode dashboard behind a private GPRC server that is available on localhost instead of on the internet. To accomplish this, this PR creates a piecestore server inspector and moves the dashboard functionality out into this new inspector.

There is a separate PR to create the private GRPC server. Ref: https://github.com/storj/storj/pull/1398


pkg/piecestore/psserver/inspector.go
```diff
@@ -0,0 +1,119 @@
...
+		InternalAddress:  "",
+		ExternalAddress:  s.ps.kad.Local().Address.Address,
+		Connection:       true,
+		Uptime:           ptypes.DurationProto(time.Since(s.ps.startTime)),
+		Stats:            statsSummary,
+	}, nil
+}
+
+// Dashboard is a stream that sends data every `interval` seconds to the listener.
+func (s *Inspector) Dashboard(in *pb.DashboardReq, stream pb.PieceStoreRoutes_DashboardServer) (err error) {
```
egonelbre: Let's convert it into a RPC call here, the stream isn't giving us any benefits. Let the gui do repeated calls instead.

JessicaGreben: That sounds like a good idea to have the GUI responsible for repeat calls.  However,  to keep the scope of this case narrow, do you think we can add that to another ticket and implement that change separately?

egonelbre: Either is fine, but I think it might be faster to do it in this PR.

JessicaGreben: ok, i removed the stream.

PR #1410, opened 2019-03-06
Adds UpdateUptimeOrCreate to StatDB

dylanlott:
* Adds UpdateUptimeOrCreate to StatDB
* Checks node.IsZero() returns false before attempting to ping node
* Adds tests for StatDB UpdateUptimeOrCreate method


pkg/discovery/service.go
```diff
@@ -133,10 +133,14 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 			return ctx.Err()
 		}
 
+		if node.Id.IsZero() {
```
egonelbre: Why would there be a zero node in the cache?

dylanlott: This was a holdover from some previous testing, I'll remove it.

pkg/discovery/service.go
```diff
@@ -133,10 +133,14 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
...
+			continue
+		}
+
 		ping, err := discovery.kad.Ping(ctx, *node)
 		if err != nil {
 			discovery.log.Info("could not ping node", zap.String("ID", node.Id.String()), zap.Error(err))
-			_, err := discovery.statdb.UpdateUptime(ctx, node.Id, false)
+			_, err := discovery.statdb.UpdateUptimeOrCreate(ctx, node.Id, false)
```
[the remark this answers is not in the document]
dylanlott: I made this PR in response to Philip showing some errors in Satellite logs where we were trying to update uptime for nodes that didn't exist yet. I think I'm going to redo this PR to keep the interface the same for StatDB but do the heavy lifting for node uptime checks in overlay instead.

PR #1411, opened 2019-03-06
Add new piece id definition

egonelbre:


pkg/storj/pieceid.go
```diff
@@ -0,0 +1,109 @@
...
+// PieceID2 is the unique identifier for pieces
+type PieceID2 [32]byte
+
+// NewPieceID creates a piece ID
+func NewPieceID() PieceID2 {
+	var id PieceID2
+
+	_, err := rand.Read(id[:])
+	if err != nil {
+		panic(err)
```
kaloyan-raev: In other places in the code base we are returning error instead of panic if `rand.Rand` fails. I know such errors should not happen, but do we really want to crash the satellite if this ever happens?

egonelbre: If we have a broken random then the system has completely failed and pretty much should be set on fire.

pkg/storj/pieceid.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return id
+}
+
+// PieceIDFromString decodes a hex encoded piece ID string
+func PieceIDFromString(s string) (PieceID2, error) {
+	idBytes, err := hex.DecodeString(s)
```
kaloyan-raev: How about using using Base32 encoding instead of Hex/Base16? Base32 is still single case, but it generates shorter encoded strings and uses more letter than numbers. I find it weird to have the piece file names with so many numbers.

egonelbre: This is not the encoding that filestoring will use. This is just for printing.

kaloyan-raev: Nevertheless, when I am debugging my eyes will feel better with this:
```
JEQGY2LLMUQEEYLTMUZTEIDCMV2HIZLSEB2GQYLOEBUGK6BB
```
than this:
```
49206C696B652042617365333220626574746572207468616E2068657821
```

egonelbre: fixed

PR #1413, opened 2019-03-06
Minimal metainfo API

mniewrzal:
Initial draft for minimal `metainfo` API


satellite/metainfo/metainfo.go
```diff
@@ -0,0 +1,107 @@
...
+	mon          = monkit.Package()
+	segmentError = errs.Class("metainfo error")
+)
+
+// APIKeys is api keys store methods used by endpoint
+type APIKeys interface {
+	GetByKey(ctx context.Context, key console.APIKey) (*console.APIKeyInfo, error)
+}
+
+type Server struct {
```
egonelbre: Endpoint

satellite/metainfo/metainfo.go
```diff
@@ -0,0 +1,107 @@
...
+	logger *zap.Logger
```
egonelbre: log

mniewrzal: ah yeah, wrong copy-paste

pkg/storage/segments/store.go
```diff
@@ -418,3 +417,35 @@ func convertTime(ts *timestamp.Timestamp) time.Time {
 	}
 	return t
 }
+
+func split(path storj.Path) (projectID, bucket string, objectPath storj.Path, segmentIndex int64, err error) {
+	comp := storj.SplitPath(path)
+	if len(comp) < 3 {
```
mniewrzal: It is ok for accessing only bucket? Won't it be `l/bucket_name` without additional path? Also I'm not sure if `projectID` will be visible on client side.

mniewrzal:
merged into `hammer` branch

PR #1416, opened 2019-03-06
Latest containers for release branches

phutchins:
Adding conditional around release branches so that we namespace latest container


Makefile
```diff
@@ -194,16 +198,16 @@ deploy: ## Update Kubernetes deployments in staging (jenkins)
 push-images: ## Push Docker images to Docker Hub (jenkins)
 	docker tag storjlabs/satellite:${TAG} storjlabs/satellite:latest
 	docker push storjlabs/satellite:${TAG}
-	docker push storjlabs/satellite:latest
+	docker push storjlabs/satellite:${LATEST_TAG}
 	docker tag storjlabs/storagenode:${TAG} storjlabs/storagenode:latest
```
brimstone: Need to update the tag here too.

PR #1417, opened 2019-03-06
repairer uses pointerdb service rather than pointerdb client

jenlij:
https://storjlabs.atlassian.net/browse/V3-1311


pkg/datarepair/repairer/repairer.go
```diff
@@ -29,31 +44,42 @@ type Service struct {
 	repairer  SegmentRepairer
 	limiter   *sync2.Limiter
 	ticker    *time.Ticker
+	pdbServer *pointerdb.Server
```
egonelbre: Other servers/endpoints/services/database should not use other servers/endpoints. The Server Endpoint are specifically for communicating outside world, nothing else. The solution here would be to introduce a specific service that allows you to do iteration rather than putting that logic in the server.

pkg/datarepair/repairer/repairer.go
```diff
@@ -7,15 +7,30 @@ import (
...
 	"storj.io/storj/pkg/transport"
 	"storj.io/storj/storage"
 )
 
+// Config contains configurable values for repairer
+type Config struct {
+	MaxRepair     int           `help:"maximum segments that can be repaired concurrently" default:"100"`
+	Interval      time.Duration `help:"how frequently checker should audit segments" default:"3600s"`
+	OverlayAddr   string        `help:"Address to contact overlay server through"`
+	PointerDBAddr string        `help:"Address to contact pointerdb server through"`
```
egonelbre: This shouldn't be necessary anymore.

pkg/datarepair/repairer/repairer.go
```diff
@@ -7,15 +7,30 @@ import (
...
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/pkg/transport"
 	"storj.io/storj/storage"
 )
 
+// Config contains configurable values for repairer
+type Config struct {
+	MaxRepair     int           `help:"maximum segments that can be repaired concurrently" default:"100"`
+	Interval      time.Duration `help:"how frequently checker should audit segments" default:"3600s"`
+	OverlayAddr   string        `help:"Address to contact overlay server through"`
```
egonelbre: This shouldn't be necessary anymore.

mobyvb: It is used in `repairer.Run`: `oc, err = overlay.NewClientContext(ctx, service.transport, service.config.OverlayAddr)`
Should this be changed?

egonelbre: Both are internal services for satellite so it makes sense to call them directly rather than going through costly grpc.

mobyvb: Is the same true of ecclient?

mobyvb: Also, there is no overlay.Service as far as I know. Should I use overlay.Server?

egonelbre: `ecclient` is necessary because storage nodes aren't internal to the satellite.

Nothing should directly depend on a server nor endpoint (with the only exception being peer, which needs to wire them up). This is what got into the "APIKey" mess in the first place.

But you could use `overlay.Cache`

egonelbre: PS: you can do this in a separate PR, if it makes life easier.

mobyvb: I'll make a separate PR for it.

mobyvb:
https://github.com/storj/storj/pull/1475#issuecomment-472444272

egonelbre:
Merged as part of https://github.com/storj/storj/pull/1422

PR #1420, opened 2019-03-07
Adding fix for release branch latest tag pushing

phutchins:


Makefile
```diff
@@ -4,10 +4,12 @@ GOARCH ?= amd64
 COMPOSE_PROJECT_NAME := ${TAG}-$(shell git rev-parse --abbrev-ref HEAD)
 BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed "s!/!-!g")
 ifeq (${BRANCH},master)
-TAG    	:= $(shell git rev-parse --short HEAD)-go${GO_VERSION}
-LATEST_TAG := latest
+	TRACKED_BRANCH := true
```
brimstone: What is going on here?

Makefile
```diff
@@ -118,19 +120,19 @@ images: satellite-image storagenode-image uplink-image gateway-image ## Build ga
 
 .PHONY: gateway-image
 gateway-image: ## Build gateway Docker image
-	${DOCKER_BUILD} --pull=true -t storjlabs/gateway:${TAG}${CUSTOMTAG} -f cmd/gateway/Dockerfile .
+	${DOCKER_BUILD} --pull=true -t storjlabs/gateway:${TAG}${CUSTOMTAG} -t storjlabs/gateway:${LATEST_TAG} -f cmd/gateway/Dockerfile .
```
brimstone: Does that work?

Makefile
```diff
@@ -196,18 +198,24 @@ deploy: ## Update Kubernetes deployments in staging (jenkins)
 
 .PHONY: push-images
 push-images: ## Push Docker images to Docker Hub (jenkins)
-	docker tag storjlabs/satellite:${TAG} storjlabs/satellite:latest
+	docker tag storjlabs/satellite:${TAG}
```
brimstone: I know this doesn't work.

Makefile
```diff
@@ -4,10 +4,12 @@ GOARCH ?= amd64
...
 BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed "s!/!-!g")
 ifeq (${BRANCH},master)
-TAG    	:= $(shell git rev-parse --short HEAD)-go${GO_VERSION}
-LATEST_TAG := latest
+	TRACKED_BRANCH := true
+  TAG    	:= $(shell git rev-parse --short HEAD)-go${GO_VERSION}
+  LATEST_TAG := latest
 else
-TAG    	:= $(shell git rev-parse --short HEAD)-${BRANCH}-go${GO_VERSION}
+	TRACKED_BRANCH := true
```
stefanbenten: Isn't that Value set in either case now?

Makefile
phutchins: Adding a variable to allow us to tell if this is a tracked branch (i.e. master, or a release branch). This variable will be the gate for adding latest tags.

Makefile
```diff
@@ -196,18 +198,20 @@ deploy: ## Update Kubernetes deployments in staging (jenkins)
...
-	docker tag storjlabs/storagenode:${TAG} storjlabs/storagenode:latest
-	docker push storjlabs/storagenode:${TAG}
-	docker push storjlabs/storagenode:${LATEST_TAG}
-	docker tag storjlabs/uplink:${TAG} storjlabs/uplink:latest
-	docker push storjlabs/uplink:${TAG}
-	docker push storjlabs/uplink:${LATEST_TAG}
-	docker tag storjlabs/gateway:${TAG} storjlabs/gateway:latest
-	docker push storjlabs/gateway:${TAG}
-	docker push storjlabs/gateway:${LATEST_TAG}
+	docker push storjlabs/satellite:${TAG}${CUSTOMTAG}
```
brimstone: Should the satellite-ui image be pushed too?

Makefile
phutchins: Pushing two tags would but the additional tag wasn't needed as we conditionally tag it down below.

Makefile
phutchins: Correct. Fixed.

Makefile
phutchins: I don't think there is value in pushing it as it's primarily used for development purposes.

PR #1421, opened 2019-03-07
Satellite console new activation flow

crawter:


cmd/storj-sim/console.go
```diff
@@ -80,7 +80,7 @@ func generateActivationKey(userID uuid.UUID, email string, createdAt time.Time)
 	return token.String(), nil
 }
 
-func addExampleProjectWithKey(key *string, address string) error {
+func addExampleProjectWithKey(key *string, activationAddress, address string) error {
```
egonelbre: does storj-sim need to adjusted as well?

crawter: @egonelbre we need this changes for integration tests

PR #1426, opened 2019-03-08
psdb: index on serial number for bandwidth_agreements

littleskunk:
My storage node is getting more than 10,000 bwagreements per day. We want to store them for 90 days. That is a total of close to 1 million bwagreements at the end. On a faster storage nodes even more.

The index should improve the performance for most of the select and update queries.


mniewrzal:
@littleskunk in the title you mentioned serial number and signature but in the code you added only serial number.

egonelbre:
It really doesn't make to have an index on the signature.

mniewrzal:
ok, then just title needs to be changed

littleskunk:
Yes my first Idea was to add both but I noticed that the signature queries are not used anymore. We might delete these queries and use serial number only. For now I have not touched them.

PR #1427, opened 2019-03-08
Hammer: uplink/piecestore streaming

egonelbre:


uplink/piecestore/upload.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (client *Upload) Write(data []byte) (written int, _ error) {
+	// if we already encountered an error, keep returning it
+	if client.sendError != nil {
+		return 0, ErrProtocol.Wrap(client.sendError)
+	}
+
+	// hash the content so far
+	_, _ = client.hash.Write(data) // guaranteed not to return error
```
egonelbre: only update content when we have sent it?

PR #1428, opened 2019-03-08
encrypted private grpc

bryanchriswhite:
Is there any specific reason not to do this?

This would add encryption to the private GRPC server. This is different from how we use TLS on the public server in that it performs no "peer certificate verification" and doesn't require a client certificate. It just uses the existing server identity's certificate on the server side.

### DIFF
--- a/pkg/peertls/tlsopts/tls.go
+++ b/pkg/peertls/tlsopts/tls.go
@@ -22,6 +22,15 @@ func (opts *Options) ServerOption() grpc.ServerOption {
 	return grpc.Creds(credentials.NewTLS(tlsConfig))
 }
 
+// ServerNoClientAuthOption returns a grpc `ServerOption` for incoming connections
+// to the node with this full identity. As opposed to `otps.ServerOption`, does not
+// require the client to have its own tls certificate.
+func (opts *Options) ServerNoClientAuthOption() grpc.ServerOption {
+	config := &tls.Config{ClientAuth: tls.NoClientCert}
+	tlsConfig := opts.tlsConfig(true, config)
+	return grpc.Creds(credentials.NewTLS(tlsConfig))
+}
+
 // DialOption returns a grpc `DialOption` for making outgoing connections
 // to the node with this peer identity.
 func (opts *Options) DialOption(id storj.NodeID) (grpc.DialOption, error) {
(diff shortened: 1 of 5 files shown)


pkg/transport/insecure.go
```diff
@@ -5,19 +5,23 @@ package transport
...
+	"google.golang.org/grpc/credentials"
 )
 
-// DialAddressInsecure returns an insecure grpc connection without tls to a node.
+// DialAddressEncrypted returns a grpc connection with encryption to a node; no client auth/cert is used.
 //
 // Use this method for communication with localhost. For example, with the inspector or debugging services.
 // Otherwise in most cases DialNode should be used for communicating with nodes since it is secure.
-func DialAddressInsecure(ctx context.Context, address string, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
+func DialAddressEncrypted(ctx context.Context, address string, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
```
egonelbre: This needs some better name. `DialAddressClient`, `DialAddressInspector` or something like that... `Encrypted` probably isn't clear in what the intent of the func is.

bryanchriswhite: I get what you're saying but I'm having a hard time coming up with another name. `DialInspectorAddress` is currently the only thing it's being used for but it's not restricted to that usage. The most specific, non-usage related description is that it dials an encrypted connection.

bryanchriswhite: What about `DialAddressEncryptedInsecure`?

jtolds: DialAddress is already Encrypted. Maybe just DialAddressUnverified?

jtolds:
for a localhost-only debug port, this adds no security benefit and makes connection slightly more complicated, but i can also see arguments for "if it doesn't matter why not prefer secure." in addition grpc prefers ssl, so okay, i guess it doesn't matter.

pkg/transport/insecure.go
bryanchriswhite: Good point; I like that.

pkg/peertls/tlsopts/tls.go
```diff
@@ -22,6 +22,15 @@ func (opts *Options) ServerOption() grpc.ServerOption {
 	return grpc.Creds(credentials.NewTLS(tlsConfig))
 }
 
+// ServerNoClientAuthOption returns a grpc `ServerOption` for incoming connections
+// to the node with this full identity. As opposed to `otps.ServerOption`, does not
```
JessicaGreben: spelling err: `otps.ServerOption` -> `opts.ServerOption`

bryanchriswhite:
@jt, the thought was, why not more secure by default, and with the debug token bit that Matt was talking about this would allow a user to safely expose this externally if they wanted.

jtolds:
oh okay, exposing externally is nice

pkg/peertls/tlsopts/tls.go
bryanchriswhite: nice catch :D

bryanchriswhite:
This is waiting for some more work on the private grpc endpoint.

mniewrzal:
@bryanchriswhite do you plan to continue developing this PR?

bryanchriswhite:
> @bryanchriswhite do you plan to continue developing this PR?

Yes

PR #1429, opened 2019-03-08
Hammer: auth/signing package

egonelbre:
Implements signing and signature verification of Order, OrderLimit and PieceHash

### DIFF
--- a/pkg/auth/signing/encode.go
+++ b/pkg/auth/signing/encode.go
@@ -11,18 +11,21 @@ import (
 
 func EncodeOrderLimit(limit *pb.OrderLimit2) ([]byte, error) {
 	signature := limit.SatelliteSignature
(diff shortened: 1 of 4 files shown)


pkg/auth/signing/encode.go
```diff
@@ -0,0 +1,28 @@
...
+package signing
+
+import (
+	"github.com/gogo/protobuf/proto"
+
+	"storj.io/storj/pkg/pb"
+)
+
+func EncodeOrderLimit(limit *pb.OrderLimit2) ([]byte, error) {
+	signature := limit.SatelliteSignature
```
mniewrzal: Shouldn't this have also `limit.SatelliteSignature = nil`?

egonelbre: yes

PR #1431, opened 2019-03-08
Hammer: piecestore.Client and piecestore.Endpoint communication

egonelbre:


internal/testplanet/uplink.go
```diff
@@ -130,6 +132,20 @@ func (uplink *Uplink) DialPointerDB(destination Peer, apikey string) (pdbclient.
 	return pdbclient.NewClient(uplink.Transport, destination.Addr(), apikey)
 }
 
+// DialPointerDB dials destination with apikey and returns pointerdb Client
```
mniewrzal: copied comment

PR #1434, opened 2019-03-08
fix kademlia bootstrap and getting peer identity from context

bryanchriswhite:
This fixes a bug that causes a panic 7-9 minutes into a storagenode running in kubernetes staging.


pkg/identity/identity.go
```diff
@@ -181,6 +181,10 @@ func PeerIdentityFromCerts(leaf, ca *x509.Certificate, rest []*x509.Certificate)
 
 // PeerIdentityFromPeer loads a PeerIdentity from a peer connection
 func PeerIdentityFromPeer(peer *peer.Peer) (*PeerIdentity, error) {
+	if peer.AuthInfo == nil {
+		return nil, Error.New("peer AuthInfo is nil")
+	}
+
 	tlsInfo := peer.AuthInfo.(credentials.TLSInfo)
```
egonelbre: Let's also check for this conversion.

pkg/kademlia/kademlia.go
```diff
@@ -121,38 +121,34 @@ func (k *Kademlia) Bootstrap(ctx context.Context) error {
...
 		if ctx.Err() != nil {
-			errs.Add(ctx.Err())
-			return errs.Err()
+			errGroup.Add(ctx.Err())
+			return errGroup.Err()
 		}
 
 		p := &peer.Peer{}
 		pCall := grpc.Peer(p)
 		_, err := k.dialer.PingAddress(ctx, node.Address.Address, pCall)
```
egonelbre: Why not use `k.dialer.FetchPeerIdentity` here?

egonelbre: `identity, err := k.dialer.PingAddress(`, since it already does it internally, no need for the `pCall` thing.

pkg/identity/identity.go
bryanchriswhite: Do we actually need both, I forgot about checking the second return arg - wouldn't that be sufficient on it's own?

pkg/kademlia/kademlia.go
bryanchriswhite: I didn't quite follow, can you elaborate?

pkg/identity/identity.go
egonelbre: The problem with not checking the conversion is that it will panic otherwise. So if a user can figure out how to get a different auth into the system, then it would panic taking down the whole server. (e.g. some grpc change upstream)

pkg/kademlia/kademlia.go
egonelbre: `PingAddress` already returns the identity, no need to do the peer.Peer thing:

```
p := &peer.Peer{}
pCall := grpc.Peer(p)
```

egonelbre: Ah sorry, I meant `FetchPeerIdentity`

egonelbre: ah right, it doesn't support arbitrary address... maybe let `PingAddress` return the identity, so the options logic doesn't leak outside?

egonelbre:
Added "Do Not Merge", because needs to be verified on k8s.

brimstone:
Errors with this PR as of commit 3db32745:
```
Configuration saved to: /root/.local/share/storj/storagenode/config.yaml
2019-03-18T14:04:22.674Z        INFO    Configuration loaded from: /root/.local/share/storj/storagenode/config.yaml
2019-03-18T14:04:23.472Z        INFO    Operator email: ops@storj.io
2019-03-18T14:04:23.473Z        INFO    Operator wallet: 0x0000000000000000000000000000000000000000
2019-03-18T14:04:23.675Z        INFO    db.migration    No Version
2019-03-18T14:04:23.675Z        INFO    db.migration.0  Initial setup
2019-03-18T14:04:23.675Z        INFO    db.migration.1  Extending bandwidth_agreements table and drop bwusagetbl
2019-03-18T14:04:23.771Z        INFO    db.migration.2  Add status column for bandwidth_agreements
2019-03-18T14:04:23.772Z        INFO    db.migration.3  Add index on serial number for bandwidth_agreements
2019-03-18T14:04:23.871Z        DEBUG   Initialized telemetry batcher with id = "1..."
2019-03-18T14:04:23.875Z        INFO    db.info.migration       No Version
2019-03-18T14:04:23.875Z        INFO    db.info.migration.0     Initial setup
2019-03-18T14:04:24.072Z        INFO    piecestore      Remaining Bandwidth     {"bytes": 536870912000}
2019-03-18T14:04:24.072Z        WARN    piecestore      Disk space is less than requested. Allocating space     {"bytes": 126252167168}
2019-03-18T14:04:24.072Z        INFO    Node 1... started
2019-03-18T14:04:24.072Z        INFO    Public server started on 1...%!(EXTRA string=[::]:7777)
2019-03-18T14:04:24.072Z        INFO    Private server started on 1...%!(EXTRA string=127.0.0.1:7778)
2019-03-18T14:04:24.073Z        DEBUG   agreements      AgreementSender is running      {"duration": "1h0m0s"}
2019-03-18T14:04:24.073Z        INFO    piecestore:monitor      Remaining Bandwidth     {"bytes": 536870912000}
2019-03-18T14:04:24.074Z        WARN    piecestore:monitor      Disk space is less than requested. Allocating space     {"bytes": 126252167168}
transport error: connection error: desc = "transport: authentication handshake failed: read tcp 100.96.49.227:43330->35.225.62.220:7777: read: connection reset by peer"
2019-03-18T14:10:39.886Z        DEBUG   transport error: connection error: desc = "transport: authentication handshake failed: read tcp 100.x.x.x:43330->35.x.x.x:7777: read: connection reset by peer"
        storj.io/storj/pkg/transport.(*Transport).DialAddress:116
        storj.io/storj/pkg/kademlia.(*Dialer).dialAddress:161
        storj.io/storj/pkg/kademlia.(*Dialer).FetchPeerIdentityUnverified:119
        storj.io/storj/pkg/kademlia.(*Kademlia).Bootstrap:130
        storj.io/storj/storagenode.(*Peer).Run.func1:251
        golang.org/x/sync/errgroup.(*Group).Go.func1:57
```

The 35.x.x.x address is the satellite, also the bootstrap address for the storagenode.

egonelbre:
I'll get this merged since it does fix things. There's something else going on with k8s.

PR #1435, opened 2019-03-08
Removes concept of email from kademlia metadata

dylanlott:


pkg/kademlia/config.go
```diff
@@ -54,30 +54,17 @@ func (c Config) Verify(log *zap.Logger) error {
 
 // OperatorConfig defines properties related to storage node operator metadata
 type OperatorConfig struct {
```
egonelbre: The email must still be specified for answering info query for trusted satellite. Also removing this is the cause for build failures.

dylanlott: Then we can't remove emails from Kademlia until that is fixed.

internal/testplanet/planet.go
```diff
@@ -395,7 +395,6 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				Alpha:  5,
 				DBPath: storageDir, // TODO: replace with master db
 				Operator: kademlia.OperatorConfig{
-					Email:  prefix + "@example.com",
```
egonelbre: operator configs should stay as it was before

pkg/kademlia/config.go
```diff
@@ -54,8 +54,8 @@ func (c Config) Verify(log *zap.Logger) error {
...
-	Email  string `user:"true" help:"operator email address" default:""`
```
egonelbre: keep the order as before

pkg/kademlia/config.go
```diff
@@ -69,15 +69,6 @@ func (c OperatorConfig) Verify(log *zap.Logger) error {
 	return nil
 }
 
-func isOperatorEmailValid(log *zap.Logger, email string) error {
```
egonelbre: keep the order as it was before

PR #1436, opened 2019-03-08
framework for deleting expired Orders from Satellite

wthorp:
Adds "DeleteExpired" to satellite BwAgreement DB.  Since we don't know what we want to do, this method is currently deliberately unimplemented.  Removed bwagreement DBX code since its a hassle.

https://storjlabs.atlassian.net/browse/V3-1312

Changes in this PR:
Added DB Scanner and Valuer support to storj.NodeID
Removes manual conversions of bytes / NodeIDs from bwagreement DB code
Renamed CreateAgreement() to SaveOrder()
Added SaverOrder(), GetExpired() and DeleteExpired() SQL logic
Removed annoying DBX generated SQL logic


pkg/accounting/tally/tally.go
```diff
@@ -16,6 +17,7 @@ import (
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/pointerdb"
 	"storj.io/storj/pkg/storj"
+	satellitedb "storj.io/storj/satellite/satellitedb/dbx"
```
egonelbre: Rest of the system shouldn't directly use dbx. Expose what you need via interfaces.

satellite/satellitedb/dbx/satellitedb.dbx.go
```diff
@@ -2691,54 +2692,10 @@ func __sqlbundle_Render(dialect __sqlbundle_Dialect, sql __sqlbundle_SQL, ops ..
...
-					p++
-				}
-				result.WriteString(x[start:p])
-			}
-
-			return result.String()
-		}
-		wasSpace = justSpace
-	}
+var __sqlbundle_reSpace = regexp.MustCompile(`\s+`)
```
egonelbre: Update your dbx.

pkg/bwagreement/server.go
```diff
@@ -20,6 +20,7 @@ import (
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/pkcrypto"
 	"storj.io/storj/pkg/storj"
+	satellitedb "storj.io/storj/satellite/satellitedb/dbx"
```
egonelbre: The comment about tally applies here as well. "satellitedb" is a specific implementation that shouldn't be used, unless you actually need to initialize it. e.g. only specific cases are 1. main.go, 2. testplanet, 3. tests.

pkg/bwagreement/server.go
```diff
@@ -42,6 +42,17 @@ type UplinkStat struct {
 	TotalTransactions int
 }
 
+//SavedOrder allows use of dbx.Bwagreement without the namespace
+type SavedOrder struct {
+	Serialnum     string
+	StorageNodeID []byte
```
egonelbre: `storj.NodeID`

pkg/bwagreement/server.go
```diff
@@ -42,6 +42,17 @@ type UplinkStat struct {
...
+	UplinkID      []byte
```
egonelbre: `storj.NodeID`

pkg/bwagreement/server.go
```diff
@@ -42,6 +42,17 @@ type UplinkStat struct {
 	TotalTransactions int
 }
 
+//SavedOrder allows use of dbx.Bwagreement without the namespace
+type SavedOrder struct {
+	Serialnum     string
```
egonelbre: `SerialNumber []byte`

pkg/bwagreement/server.go
```diff
@@ -42,6 +42,17 @@ type UplinkStat struct {
 	TotalTransactions int
 }
 
+//SavedOrder allows use of dbx.Bwagreement without the namespace
+type SavedOrder struct {
```
egonelbre: You could also use `pb.Order`, although I do prefer database logic not be dependent on transport concerns.

pkg/bwagreement/server.go
wthorp: this is a string everywhere else in the codebase

pkg/accounting/tally/tally.go
```diff
@@ -85,6 +85,21 @@ func (t *Tally) Tally(ctx context.Context) error {
...
 			errBWA = errs.New("Saving for bandwidth failed : %v", err)
+		} else {
+			//remove expired records
+			now := time.Now()
+			_, err = t.bwAgreementDB.GetExpired(tallyEnd, now)
+			if err != nil {
+				return err
+			}
+			var expiredOrdersHaveBeenSaved bool
+			//todo: write files to disk or whatever we decide to do here
```
mobyvb: we should make sure there is a ticket in jira for this future work

pkg/storj/node.go
```diff
@@ -143,6 +144,19 @@ func (id NodeID) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + id.String() + `"`), nil
 }
 
+// Value set a NodeID to a database field
+func (id NodeID) Value() (driver.Value, error) {
```
mobyvb: might be worth adding a quick test or two for `Value()` and `Scan()`. They look good to me though.

satellite/satellitedb/bandwidthagreement.go
```diff
@@ -85,23 +79,41 @@ func (b *bandwidthagreement) GetTotals(ctx context.Context, from, to time.Time)
...
 			return totals, err
 		}
-		id, err := storj.NodeIDFromBytes(nodeID)
+		totals[nodeID] = data
+	}
+	return totals, nil
+}
+
+//GetExpired gets orders that are expired and were created before some time
+func (b *bandwidthagreement) GetExpired(before time.Time, expiredAt time.Time) (orders []bwagreement.SavedOrder, err error) {
```
mobyvb: are there tests for `GetExpired` and `DeleteExpired` anywhere?

pkg/accounting/tally/tally.go
wthorp: Created https://storjlabs.atlassian.net/browse/V3-1368

pkg/storj/node.go
wthorp: Added

pkg/storj/node.go
```diff
@@ -143,6 +144,19 @@ func (id NodeID) MarshalJSON() ([]byte, error) {
...
+	return id.Bytes(), nil
+}
+
+// Scan extracts a NodeID from a database field
+func (id *NodeID) Scan(src interface{}) (err error) {
+	b, _ := src.([]byte)
```
egonelbre: check the type cast result and return an appropriate error if it's a mismatch

wthorp: done

satellite/satellitedb/bandwidthagreement.go
wthorp: nope :D

wthorp:
> This wasn't changed in this PR, but do `GetTotals` and `GetUplinkStats` still need the context argument?

No.  I almost changed them, but I didn't want to pull in too many unrelated changes.

satellite/satellitedb/bandwidthagreement.go
cam-a: Do we need `before` in `GetExpired` and `DeleteExpired`? Is there a case where we don't want to delete something that is expired because it was not created before some time?

wthorp: Yes.  We need to make sure we ran the accounting tally before we delete expired agreements.

PR #1437, opened 2019-03-09
filestore changes for piece store 

egonelbre:
This ensures the filestore can be more easily integrated into piece store.


storage/filestore/dir.go
```diff
@@ -77,10 +79,19 @@ func (dir *Dir) DeleteTemporary(file *os.File) error {
...
-	hex := hex.EncodeToString(ref[:])
-	return filepath.Join(dir.blobdir(), hex[0:2], hex[2:])
+// blobToPath converts blob reference to a filepath
+func (dir *Dir) blobToPath(ref storage.BlobRef) string {
+	namespace := pathEncoding.EncodeToString(ref.Namespace)
+	key := pathEncoding.EncodeToString(ref.Key)
+	return filepath.Join(dir.blobdir(), namespace, key[:2], key[2:])
+}
+
+// blobToTrashPath converts blob reference to a filepath
```
aleitner: better comment to understand the difference between `blobToTrashPath ` and `blobToPath `

storage/filestore/blob.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"io"
+	"os"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/storage"
+)
+
+// blobWriter implements reader that is offset by blob header size
```
kaloyan-raev: The comment should start with `blobReader`

storage/filestore/blob.go
```diff
@@ -0,0 +1,59 @@
...
+	*os.File
+}
+
+func newBlobReader(file *os.File) *blobReader {
+	return &blobReader{file}
+}
+
+// Size returns how large is the blob.
+func (blob *blobReader) Size() int64 {
+	stat, _ := blob.Stat()
```
kaloyan-raev: Why isn't there any errcheck here?

storage/filestore/dir.go
```diff
@@ -18,10 +18,12 @@ import (
 )
 
 const (
-	blobPermission = 0600
-	dirPermission  = 0700
+	blobPermission = 0655
+	dirPermission  = 0755
 )
 
+var pathEncoding = base32.NewEncoding("abcdefghijklmnopqrstuvwxyz234567").WithPadding('1')
```
kaloyan-raev: Quite interesting encoding. I can understand that using the small letters instead of the caps improves readability. But why the "1" padding?

storage/filestore/blob.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Commit moves the file to the target location.
+func (blob *blobWriter) Commit() error {
+	err := blob.store.dir.Commit(blob.File, blob.ref)
+	return Error.Wrap(err)
+}
+
+// Size returns how much has been written so far.
+func (blob *blobWriter) Size() int64 {
+	p, _ := blob.Seek(0, io.SeekCurrent)
```
kaloyan-raev: errcheck?

storage/filestore/dir.go
```diff
@@ -18,10 +18,12 @@ import (
 )
 
 const (
-	blobPermission = 0600
-	dirPermission  = 0700
+	blobPermission = 0655
+	dirPermission  = 0755
```
kaloyan-raev: Why changing the permissions? What's the reason for other users to mess up with the blob storage on the file system?

storage/filestore/dir.go
egonelbre: because I had to pick something, I also could use `_`... or other valid file character

egonelbre: Also I could use no-padding and a prefix

egonelbre: fixed

PR #1439, opened 2019-03-11
framework to delete raw accounting tallies

wthorp:
Adds "DeleteRawBefore" to satellite Accounting DB. Since we don't know what we want to do, this method is currently deliberately unimplemented.

https://storjlabs.atlassian.net/browse/V3-1312

Changes in this PR:
Added DeleteRawBefore() SQL logic
Added WithTx() for cleaner transaction handling
Replaced old transaction handling with WithTx() in Accounting.go
Renamed rollup.Query() to rollup.RollupRaws()


pkg/accounting/rollup/rollup.go
```diff
@@ -109,5 +109,15 @@ func (r *Rollup) Query(ctx context.Context) error {
...
 		return nil
 	}
-	return Error.Wrap(r.db.SaveRollup(ctx, latestTally, rollupStats))
+	err = r.db.SaveRollup(ctx, latestTally, rollupStats)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	//removed the rolled-up raws
+	var rolledUpRawsHaveBeenSaved bool
+	//todo: write files to disk or whatever we decide to do here
```
mobyvb: same question as on the other PR: is there a jira ticket for this work?

wthorp: https://storjlabs.atlassian.net/browse/V3-1306

PR #1440, opened 2019-03-11
Hammer: implement piece storage

egonelbre:


storagenode/pieces/readwrite.go
```diff
@@ -0,0 +1,122 @@
...
+	w := &Writer{}
+	w.buf = *bufio.NewWriterSize(blob, bufferSize)
+	w.blob = blob
+	w.hash = pkcrypto.NewHash()
+	return w, nil
+}
+
+// Write writes data to the blob and calculates the hash.
+func (w *Writer) Write(data []byte) (int, error) {
+	_, err := w.hash.Write(data)
```
kaloyan-raev: Why are we adding the data to the hash first? Should we add only those bytes that were _actually_ written to buf _after_ writing them successfully?

storagenode/pieces/readwrite.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return 0, Error.Wrap(err)
+	}
+	n, err := w.buf.Write(data)
```
mniewrzal: Is it possible that hash will write more bytes than `buf`?

egonelbre: ah, right

mniewrzal: Kaloyan better addressed my concerns :)

storagenode/pieces/readwrite.go
egonelbre: yes, fixed

uplink/piecestore/upload.go
```diff
@@ -66,8 +66,10 @@ func (client *Upload) Write(data []byte) (written int, _ error) {
 		return 0, ErrProtocol.Wrap(client.sendError)
 	}
 
-	// hash the content so far
-	_, _ = client.hash.Write(data) // guaranteed not to return error
+	fullData := data
+	defer func() {
+		_, _ = client.hash.Write(fullData[:written]) // guaranteed not to return error
```
mniewrzal: You can use `pkcrypto.NewHash()` for this `client.hash` in constructor too I guess.

PR #1441, opened 2019-03-11
 Signing improvements

mniewrzal:
* OrderLimit signing/verifying on metainfo
* fix for getting peer identity from stream

### DIFF
--- a/pkg/auth/signing/peers.go
+++ b/pkg/auth/signing/peers.go
@@ -39,13 +39,6 @@ type PublicKey struct {
 	Key  crypto.PublicKey
 }
 
-func SigneeFromFullIdentity(identity *identity.FullIdentity) Signee {
-	return &PublicKey{
-		Self: identity.ID,
-		Key:  identity.Leaf.PublicKey,
-	}
-}
-
 func SigneeFromPeerIdentity(identity *identity.PeerIdentity) Signee {
 	return &PublicKey{
 		Self: identity.ID,
--- a/pkg/auth/signing/peers.go
+++ b/pkg/auth/signing/peers.go
@@ -39,6 +39,13 @@ type PublicKey struct {
 	Key  crypto.PublicKey
 }
 
+func SigneeFromFullIdentity(identity *identity.FullIdentity) Signee {
+	return &PublicKey{
+		Self: identity.ID,
+		Key:  identity.Leaf.PublicKey,
+	}
+}
+
 func SigneeFromPeerIdentity(identity *identity.PeerIdentity) Signee {
 	return &PublicKey{
 		Self: identity.ID,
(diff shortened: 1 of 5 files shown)


pkg/auth/signing/peers.go
```diff
@@ -39,6 +39,13 @@ type PublicKey struct {
 	Key  crypto.PublicKey
 }
 
+func SigneeFromFullIdentity(identity *identity.FullIdentity) Signee {
```
egonelbre: You can use `SignerFromFullIdentity` since it implements both `Signee` and `Signer`.

satellite/peer.go
```diff
@@ -331,6 +331,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 			peer.Metainfo.Allocation,
 			peer.Overlay.Service,
 			peer.DB.Console().APIKeys(),
+			peer.Identity,
```
egonelbre: Pass in the signer rather than the identity.

satellite/metainfo/metainfo.go
```diff
@@ -44,12 +45,18 @@ type Endpoint struct {
 	allocation           *pointerdb.AllocationSigner
 	cache                *overlay.Cache
 	apiKeys              APIKeys
+	identity             *identity.FullIdentity
```
egonelbre: no need to keep the identity here

pkg/auth/signing/peers.go
mniewrzal: thanks, I missed that

uplink/piecestore/download.go
```diff
@@ -36,14 +36,14 @@ type Download struct {
 }
 
 func (client *Client) Download(ctx context.Context, limit *pb.OrderLimit2, offset, size int64) (*Download, error) {
-	peer, err := identity.PeerIdentityFromContext(ctx)
+	stream, err := client.client.Download(ctx)
```
egonelbre: I wonder whether we can get peer identity from *grpc.ClientConn. I'm just trying to avoid sending the orderlimit unless we already have peer identity.

satellite/metainfo/metainfo.go
mniewrzal: argh, yes

uplink/piecestore/download.go
mniewrzal: but we are sending OrderLimit few lines later. We are getting stream, peer identity and then we are using stream

egonelbre: ah, yeah, nevermind

mniewrzal: maybe name `Upload` for both methods is confusing. We have `Upload` in `Upload`

egonelbre: Yeah, I think so, but using different names might be confusing as well. We can put into bikeshed if you wish.

PR #1444, opened 2019-03-11
ECClient uses new piecestore.Client

kaloyan-raev:


pkg/storage/ec/client.go
```diff
@@ -47,16 +50,25 @@ func NewClient(tc transport.Client, memoryLimit int) Client {
...
-func (ec *ecClient) newPSClient(ctx context.Context, n *pb.Node) (psclient.Client, error) {
+func (ec *ecClient) newPSClient(ctx context.Context, n *pb.Node) (*piecestore.Client, error) {
 	n.Type.DPanicOnInvalid("new ps client")
-	return ec.newPSClientFunc(ctx, ec.transport, n, 0)
+	conn, err := ec.transport.DialNode(ctx, n)
+	if err != nil {
+		return nil, err
+	}
+	return ec.newPSClientFunc(
+		zap.L().Named(n.Id.String()),
```
mniewrzal: This might be a very long prefix.

kaloyan-raev: This is what @egonelbre suggested.

mniewrzal: ok

PR #1446, opened 2019-03-11
hammer: new sqlite schema for storage node

egonelbre:


storagenode/storagenodedb/infodb.go
```diff
@@ -0,0 +1,163 @@
...
+						uplink_piece_hash BLOB    NOT NULL, -- serialized pb.PieceHash signed by uplink
+						uplink_cert_id    INTEGER NOT NULL,
+
+						FOREIGN KEY(uplink_cert_id) REFERENCES certificate(cert_id)
+					)`,
+					// primary key by satellite id and piece id
+					`CREATE UNIQUE INDEX pk_pieceinfo ON pieceinfo(satellite_id, piece_id)`,
+
+					// table for storing bandwidth usage
+					`CREATE TABLE bandwidth_usage (
```
stefanbenten: Is that table there for keeping track of the amount transfered for the Dashboard?
If so, are we resetting that table every month?

storagenode/storagenodedb/usedserials.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return ErrInfo.Wrap(err)
+}
+
+// SerialNumberFn is callback from IterateAll
+type SerialNumberFn func(satelliteID storj.NodeID, serialNumber []byte, expiration time.Time)
+
+// IterateAll iterates all serials.
+// Note, this will lock the database and should only be used during startup.
+func (db *usedSerials) IterateAll(ctx context.Context, fn SerialNumberFn) (err error) {
```
stefanbenten: What does this function provide?
Ensuring that the database is healthy?

storagenode/storagenodedb/infodb.go
egonelbre: at the moment there's no resetting code, because I'm unsure what the exact logic should be

storagenode/storagenodedb/usedserials.go
egonelbre: When we restart the storagenode we need to reload all the serial numbers into memory.

stefanbenten: Oh yeah, good idea.

storagenode/storagenodedb/infodb.go
stefanbenten: The usage monitor should reset every 1st of the month.
Thats how our bandwidth metric has been published to work.

egonelbre: we can query a particular month or rolling month whichever is necessary

egonelbre: with regards to resetting things we can add that later as it's not critical part of the system

stefanbenten: Sounds good to me! Nice work

storagenode/storagenodedb/infodb.go
```diff
@@ -0,0 +1,163 @@
...
+						expiration    TIMESTAMP NOT NULL
+					)`,
+					// primary key on satellite id and serial number
+					`CREATE UNIQUE INDEX pk_used_serial ON used_serial(satellite_id, serial_number)`,
+					// expiration index to allow fast deletion
+					`CREATE INDEX idx_used_serial ON used_serial(expiration)`,
+
+					// certificate table for storing uplink/satellite certificates
+					`CREATE TABLE certificate (
+						cert_id       INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
```
mniewrzal: Maybe you explained earlier but why we need `cert_id`? Won't node_id will unique?

egonelbre: Because a single node_id can have multiple leaf certificates.

PR #1448, opened 2019-03-11
Add ability to view irreparable segments on satellite

cam-a:
This PR 
- creates an IrreparableInspector API for requesting entries from the satellite's IrreparableDB
- adds an inspector subcommand, `irreparable`, which calls IrreparableInspector API for viewing segments
- adds a new method to IrreparableDB, `GetLimited`, for pagination of results


cmd/inspector/main.go
```diff
@@ -435,9 +444,67 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
 	return nil
 }
 
+func getSegments(cmd *cobra.Command, args []string) error {
+	if *limit <= int64(0) {
```
coyle: https://play.golang.org/p/nJ540hDdY8b

pkg/pb/inspector.proto
```diff
@@ -47,6 +47,33 @@ service PieceStoreInspector {
...
+  int64 last_repair_attempt = 4;
+  int64 repair_attempt_count = 5;
+}
+
+message SegmentGroup {
+  repeated IrreparableSegment segments = 1;  
+}
+
+message ListSegmentsResponse {
+  bytes data = 1;
```
wthorp: Why is this bytes instead of repeated IrreparableSegment?

cam-a: I thought it might be good to marshal the segments before sending them back

cam-a: I thought it might be good to marshal the segments before sending them back

wthorp: unless they need signing or some special treatment, marshaling is what protocol buffers do :D

pkg/pb/inspector.proto
```diff
@@ -47,6 +47,33 @@ service PieceStoreInspector {
...
+// ListSegments
+message ListSegmentsRequest {
+  int64 limit = 1;
+  int64 offset = 2;
+}
+
+message IrreparableSegment {
+  bytes encrypted_path = 1;
+  bytes segment_detail = 2;
+  int64 lost_pieces = 3;
```
wthorp: I'm a fan of adding [(gogoproto.nullable) = false]; to things that will always have a value

cmd/inspector/main.go
```diff
@@ -435,9 +445,49 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
...
+
+		for _, seg := range res.Segments {
+			//TODO: format results
+			fmt.Println(seg)
+		}
+
+		length = int64(len(res.Segments))
+		offset += length
+
+		fmt.Printf("%d results\n", length)
```
coyle: Do you still need this ?

cmd/inspector/main.go
```diff
@@ -435,9 +445,49 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
...
+	// query DB and paginate results
+	for length >= irreparableLimit {
+		res, err := i.irrdbclient.ListSegments(context.Background(), &pb.ListSegmentsRequest{Limit: irreparableLimit, Offset: offset})
+		if err != nil {
+			return ErrRequest.Wrap(err)
+		}
+
+		for _, seg := range res.Segments {
+			//TODO: format results
+			fmt.Println(seg)
```
coyle: Do you still need this?

pkg/datarepair/irreparable/irreparable_test.go
```diff
@@ -22,35 +23,84 @@ func TestIrreparable(t *testing.T) {
...
+
+		{ // GetLimited limit 3, offset 1
+			segs, err := irrdb.GetLimited(ctx, 3, 1)
+			assert.NoError(t, err)
+			assert.Equal(t, 2, len(segs))
+			assert.Equal(t, segments[1], segs[0])
+			assert.Equal(t, segments[2], segs[1])
+		}
+
+		// When limit or offset is negative, postgres returns an error, but SQLite does not
```
coyle: Does this need to be commented out ?

pkg/pb/inspector.proto
```diff
@@ -47,6 +47,29 @@ service PieceStoreInspector {
...
 }
 
+service IrreparableInspector {
+  // ListSegments returns damaged segments
+  rpc ListSegments(ListSegmentsRequest) returns (ListSegmentsResponse);
+}
+
+// ListSegments
+message ListSegmentsRequest {
+  int64 limit = 1;
```
coyle: we could probably get away with int32

pkg/pb/inspector.proto
```diff
@@ -47,6 +47,29 @@ service PieceStoreInspector {
...
+  int64 offset = 2;
+}
+
+message IrreparableSegment {
+  bytes encrypted_path = 1;
+  bytes segment_detail = 2;
+  int64 lost_pieces = 3;
```
coyle: int32 should be enough

cmd/inspector/main.go
```diff
@@ -435,9 +449,106 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
...
+	RepairAttemptCount   int64       `json:"repair_attempt_count"`
+	SegmentDetail        *pb.Pointer `json:"segment_detail"`
+}
+
+// sortSegments by the object they belong to
+func sortSegments(segments []*formattedSegment) map[string][]*formattedSegment {
+	objects := make(map[string][]*formattedSegment)
+	for _, seg := range segments {
+		path := storj.SplitPath(seg.EncryptedSegmentPath)
+		path = append(path[:1], path[2:]...)
```
mobyvb: I am pretty sure instead of 
```
path = append(path[:1], path[2:]...)
objPath := strings.Join(path, "/")
```
You should do
```
objPath := strings.Join(path[1:], "/")
```
Or maybe even
```
objPath := storj.JoinPath(path[1:])
```

This is because you just want to get rid of the very first part of the path ("l" or "s0" or "s1" etc..) rather than the bucket name.

mobyvb: jk you're right

cmd/inspector/main.go
```diff
@@ -435,9 +449,106 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
...
+		objects[objPath] = append(objects[objPath], seg)
+	}
+	return objects
+}
+
+// unpackSegmentDetail unmarshals a slice of segments' pointers for viewing
+func unpackSegmentDetail(packed []*pb.IrreparableSegment) (unpacked []*formattedSegment, err error) {
+	for _, seg := range packed {
+		p := &pb.Pointer{}
+		err = proto.Unmarshal(seg.SegmentDetail, p)
```
wthorp: This isn't your fault, but its a little bogus that EncryptedSegmentDetail is a *marshaled* pb.Pointer and that EncryptedSegmentPath is a search prefix / storage.Key.  Neither is actually encrypted.  There's likely no coherent reason that you have to manually unmarshal SegmentDetail.  I give you permission to rage refactor this.  :D :P

pkg/pb/inspector.proto
```diff
@@ -47,6 +47,29 @@ service PieceStoreInspector {
...
+  rpc ListSegments(ListSegmentsRequest) returns (ListSegmentsResponse);
+}
+
+// ListSegments
+message ListSegmentsRequest {
+  int32 limit = 1;
+  int32 offset = 2;
+}
+
+message IrreparableSegment {
```
wthorp: If IrreparableSegment is nearly an identical clone of RemoteSegmentInfo, I suggest refactoring to remove RemoteSegmentInfo.  While you're at it, life would probably be better if you renamed away from 'Encrypted' (I don't think they are), and also if you stored EncryptedSegmentDetail as a pb.Pointer directly so it automatically marshals.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -36,6 +36,8 @@ read one (
 	where  irreparabledb.segmentpath = ?
 )
 
+read limitoffset ( select irreparabledb )
```
wthorp: Don't you need an order by here? (didn't I make this comment already?)

cmd/inspector/main.go
```diff
@@ -435,9 +449,106 @@ func CreateCSVStats(cmd *cobra.Command, args []string) (err error) {
...
+	SegmentDetail        *pb.Pointer `json:"segment_detail"`
+}
+
+// sortSegments by the object they belong to
+func sortSegments(segments []*formattedSegment) map[string][]*formattedSegment {
+	objects := make(map[string][]*formattedSegment)
+	for _, seg := range segments {
+		path := storj.SplitPath(seg.EncryptedSegmentPath)
+		path = append(path[:1], path[2:]...)
+		objPath := strings.Join(path, "/")
```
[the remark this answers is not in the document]
thepaul: It looks like the results here are solely for presentation to the user by way of the `inspector` command-line tool, so `/` is probably more correct than anything else (maybe until we start supporting configurable path delimiters like S3 does)..

PR #1449, opened 2019-03-11
hammer: add definition for serial number

egonelbre:
not yet sure whether this is a good idea?


egonelbre:
Closed in favor of https://github.com/storj/storj/pull/1508

PR #1451, opened 2019-03-11
Add Application Name to PostgreSQL Connection String

stefanbenten:
This PR adds a proper Application Name to the PostgreSQL Connection String, if none is provided

### DIFF
--- a/internal/dbutil/pgutil/db.go
+++ b/internal/dbutil/pgutil/db.go
@@ -111,6 +111,6 @@ func CheckApplicationName(s string) (r string) {
 		}
 		return
 	}
-	//return source as is if ApplicationName is set
+	//return source as is if application_name is set
 	return s
 }
--- a/internal/dbutil/pgutil/db.go
+++ b/internal/dbutil/pgutil/db.go
@@ -106,10 +106,9 @@ func CheckApplicationName(s string) (r string) {
 		if !strings.Contains(s, "?") {
 			r = s + "?application_name=Satellite"
 			return
-		} else {
-			r = s + "&application_name=Satellite"
-			return
 		}
+		r = s + "&application_name=Satellite"
+		return
 	}
 	//return source as is if application_name is set
 	return s
--- a/internal/dbutil/pgutil/db.go
+++ b/internal/dbutil/pgutil/db.go
@@ -5,7 +5,6 @@ package pgutil
 
 import (
 	"database/sql"
-	"fmt"
 	"strings"
 
 	"github.com/zeebo/errs"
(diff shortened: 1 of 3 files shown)


stefanbenten:
@egonelbre Shoot, navigated into the wrong file...
Relocated.

satellite/satellitedb/database.go
```diff
@@ -41,7 +43,9 @@ func New(log *zap.Logger, databaseURL string) (satellite.DB, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	if !strings.Contains(source, "ApplicationName") {
+		source = fmt.Sprintf("%s?%s", source, "ApplicationName=Satellite")
```
egonelbre: the `source` may already contain query parameters and the driver can here be sqlite as well

internal/dbutil/pgutil/db.go
```diff
@@ -98,3 +100,17 @@ func QuerySnapshot(db dbschema.Queryer) (*dbschema.Snapshot, error) {
 		Data:    data,
 	}, err
 }
+
+//CheckApplicationName ensures that the Connection String contains an application name
+func CheckApplicationName(s string) (r string) {
+	if !strings.Contains(s, "ApplicationName") {
```
aleitner: shouldn't you be checking for `application_name` instead of `ApplicationName`?

stefanbenten: Done, good catch!

internal/dbutil/pgutil/db.go
```diff
@@ -98,3 +100,17 @@ func QuerySnapshot(db dbschema.Queryer) (*dbschema.Snapshot, error) {
...
+func CheckApplicationName(s string) (r string) {
+	if !strings.Contains(s, "application_name") {
+		if !strings.Contains(s, "?") {
+			r = fmt.Sprintf("%s?%s", s, "application_name=Satellite")
+		} else {
+			r = fmt.Sprintf("%s%s%s", s, "&", "application_name=Satellite")
+		}
+		return
+	}
+	//return source as is if ApplicationName is set
```
aleitner: Doesn't really need to be fixed but you could change this comment too

internal/dbutil/pgutil/db.go
```diff
@@ -98,3 +100,17 @@ func QuerySnapshot(db dbschema.Queryer) (*dbschema.Snapshot, error) {
...
 	}, err
 }
+
+//CheckApplicationName ensures that the Connection String contains an application name
+func CheckApplicationName(s string) (r string) {
+	if !strings.Contains(s, "application_name") {
+		if !strings.Contains(s, "?") {
+			r = fmt.Sprintf("%s?%s", s, "application_name=Satellite")
+		} else {
+			r = fmt.Sprintf("%s%s%s", s, "&", "application_name=Satellite")
```
wthorp: nit:
you could really just use string concatenation here and above.
you could also directly return here and above.

stefanbenten: Sure, let me change that real quick!

PR #847, opened 2018-12-12
proposal, wip: second draft of libuplink

jtolds:

PR #1452, opened 2019-03-11
Adds Libuplink

dylanlott:


lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+// etc), may be restricted to a single bucket, may be restricted to a prefix
+// within a bucket, or may even be restricted to a single file within a bucket.
+// NB(dylan): You need an Access to start a Session
+type Access struct {
+	Permissions Macaroon
+
+	// TODO: these should be per-bucket somehow maybe? oh man what a nightmare
+	// Could be done via []Bucket with struct that has each of these
+	// PathCipher       storj.Cipher // i.e. storj.AESGCM
+	// EncPathPrefix    storj.Path
```
thepaul: At least EncPathPrefix won't need to be per-bucket. :D

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+	// Key              storj.Key
+	// EncryptionScheme storj.EncryptionScheme
+
+	// Something like this?
+	// TODO(dylan): Shouldn't actually use string, this is just a placeholder
+	// until a more precise type is figured out - probably type Bucket
+	Buckets map[string]BucketOpts
```
thepaul: This map should probably be indexed by a pair of (ProjectID, BucketName). Or I suppose we could say "you have to instantiate a new Uplink object for every different ProjectID you're working with" and then this could safely be just `string` for bucket names.

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Uplink represents the main entrypoint to Storj V3. An Uplink connects to
+// a specific Satellite and caches connections and resources, allowing one to
+// create sessions delineated by specific access controls.
+type Uplink struct {
+	ID            *identity.FullIdentity
```
thepaul: Are you sure it's worth exporting all of these fields? Users will probably end up changing things when they're not supposed to...

lib/uplink/buckets.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// BucketOpts holds the cipher, path, key, and enc. scheme for each bucket since they
+// can be different for each
+type BucketOpts struct {
+	PathCipher       storj.Cipher
+	EncPathPrefix    storj.Path
```
thepaul: I don't think buckets have a top-level encrypted path prefix associated with them. I think when deriving an encrypted path you start with an hmac of the bucket's encryption key and the bucket's name.

lib/uplink/buckets.go
```diff
@@ -0,0 +1,62 @@
...
+	Key              storj.Key
+	EncryptionScheme storj.EncryptionScheme
+}
+
+// CreateBucketOptions holds the bucket opts
+type CreateBucketOptions struct {
+	PathCipher storj.Cipher
+	// this differs from storj.CreateBucket's choice of just using storj.Bucket
+	// by not having 2/3 unsettable fields.
```
thepaul: I might be way out of date, but I believe CreateBucket does need to know the root encryption key and encryption scheme, so it can create the toplevel bucket metainfo in pointerdb.

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// ParseAccess parses a serialized Access
+func ParseAccess(data []byte) (Access, error) {
+	panic("TODO")
+}
+
+// Serialize serializes an Access message
+func (a *Access) Serialize() ([]byte, error) {
+	panic("TODO")
```
thepaul: Is serializing an Access message a thing we want to do? Would the Access struct still include the encryption key and everything? I'm kind of thinking that the only extra thing that you might need to serialize (beyond the normal GRPC stuff) is the macaroons.

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+	ul "storj.io/storj/uplink"
+)
+
+// Config holds the configs for the Uplink
+type Config struct {
+	// MaxBufferMem controls upload performance and is system-specific
+	MaxBufferMem int
+
+	// These should only be relevant for new files; these values for existing
+	// files should come from the metainfo index. It's unlikely these will ever
```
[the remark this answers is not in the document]
dylanlott: I should have added a TODO But yes, I definitely plan to make these configurable via a passed-in configuration struct or something. I just haven't gotten to that part. I'll make a note in the code!

lib/uplink/buckets.go
thepaul: Or no, just the encryption scheme.

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Access is all of the access information an application needs to store and
+// retrieve data. Someone with a share may have no restrictions within a project
+// (can create buckets, list buckets, list files, upload files, delete files,
+// etc), may be restricted to a single bucket, may be restricted to a prefix
+// within a bucket, or may even be restricted to a single file within a bucket.
+// NB(dylan): You need an Access to start a Session
+type Access struct {
+	Permissions Macaroon
```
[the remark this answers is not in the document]
dylanlott: I attached Macaroons to the Access struct because Access structs can be per bucket, which will allow us to control permissions to a per bucket or per object resolution. This is basically another TODO here.

lib/uplink/main.go
dylanlott: That's not a bad idea, actually. I'll look into that. Making another TODO 👍

lib/uplink/main.go
dylanlott: Noted, I'll make them private, I wasn't even thinking about that part yet haha

lib/uplink/buckets.go
dylanlott: So I can just remove EncPathPrefix?

jtolds: they do if you're only granted access to a subfolder, but i agree bucketopts is not the struct for it.

lib/uplink/main.go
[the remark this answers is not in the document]
thepaul: Even if it's per-bucket, though, I mean, you still might need multiple macaroons. Maybe I have one for write access and a separate one for read access? Or maybe there are multiple subfolders, and I have a different macaroon for access to each

lib/uplink/main.go
dylanlott: This is a hold over from the libuplink-draft branch PR that I was shown for this ticket. I'm not actually sure if we want to do this or not yet.

lib/uplink/main.go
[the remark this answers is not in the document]
jtolds: we will probably need a place to specify defaults for new objects. this could be that place

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+	EncBlockSize  int
+	MaxInlineSize int
+	SegmentSize   int64
+	TLSConfig     tlsopts.Config
+}
+
+// Session represents a specific access session.
+type Session struct {
+	TransportClient transport.Client
+	Gateway         *miniogw.Gateway
```
jtolds: a gateway should be implemented using libuplink, not the other way around.

lib/uplink/main.go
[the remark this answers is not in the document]
jtolds: This is why there's a session. you can make a new session per access type. that work @thepaul?

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+	// until a more precise type is figured out - probably type Bucket
+	Buckets map[string]BucketOpts
+}
+
+// Uplink represents the main entrypoint to Storj V3. An Uplink connects to
+// a specific Satellite and caches connections and resources, allowing one to
+// create sessions delineated by specific access controls.
+type Uplink struct {
+	ID            *identity.FullIdentity
+	Session       *Session
```
jtolds: session should not be a field here. there may be more than one session per uplink

lib/uplink/buckets.go
thepaul: Checking..... yes, it looks like you can do that. You only need an encrypted path prefix if you're dealing with a "directory" (which is just a virtual thing as far as the object store is concerned, but would affect the encrypted path names of files created within it).

lib/uplink/main.go
jtolds: yeah, we need encryption details if you are only shared a subfolder

lib/uplink/main.go
```diff
@@ -0,0 +1,142 @@
...
+	return nil
+}
+
+// Access creates a new share, potentially further restricted from the Access used
+// to create this session.
+func (s *Session) Access(ctx context.Context, caveats ...Caveat) (Access, error) {
+	panic("TODO")
+}
+
+func getGateway(ctx context.Context, identity *identity.FullIdentity) (*miniogw.Gateway, error) {
```
jtolds: this package should have no dependency on miniogw at all. instead, the code in miniogw should be moved to this package and miniogw should use this package

lib/uplink/buckets.go
dylanlott: This part of the code is incomplete, so I'll add a TODO here for that as well.

lib/uplink/main.go
dylanlott: Ah, I think I might have misunderstood how this was supposed to fit into our ecosystem then. I can work on restructuring this then. Back to the drawing board.

lib/uplink/main.go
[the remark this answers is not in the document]
thepaul: That would make it technically possible to do all the things, but sounds like kind of a nightmare to work with for anything beyond trivially straightforward uses. Maybe I'm not envisioning the right usage scenarios.

lib/uplink/main.go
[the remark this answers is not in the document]
dylanlott: I think this is probably the best place to declare our defaults and then override them and pass them through on a per-Session basis.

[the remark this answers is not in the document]
thepaul: That makes sense. :+1:

lib/uplink/buckets.go
dylanlott: Okay, I added `EncConfig` struct to this so it can be made with different encryption

lib/uplink/main.go
thepaul: Oh, I get it, this is for communicating details of a share with another user, not for directly participating in the network. Cool then

lib/uplink/buckets.go
```diff
@@ -60,3 +63,84 @@ func (s *Session) ListBuckets(ctx context.Context, opts storj.BucketListOptions)
...
+		startAfter = list.Items[len(list.Items)-1].Name
+	}
+
+	return bucketItems, err
+}
+
+// MakeBucketWithLocation makes a bucket with a location and returns an error if there were any problems.
+func (client *Client) MakeBucketWithLocation(ctx context.Context, bucket string, location string) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	// TODO: This current strategy of calling bs.Get
```
thepaul: Yes, and a similar problem occurs with DeleteBucket: there is a race between checking if the bucket is empty and then deleting it. And from my trying to follow it down, it looks to me like DeleteBucket itself doesn't check for that sort of thing; it just says "ok" and removes the bucket data whether there are entries there or not. The "don't Delete non-empty buckets" and "don't Put buckets that already exist" rules should really be enforced by the pointerdb if we care about them at all. And then in that case, there would be no need to do these racy checks; you'd just call DeleteBucket/CreateBucket/etc and report whatever errors came back.

But fixing that would be way out of scope for this ticket, and as you say, this is the way it's already done in the gateway and uplink command-line tool, so this is good for now :)

lib/uplink/buckets.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// ListBuckets will list authorized buckets
+func (s *Session) ListBuckets(ctx context.Context, opts storj.BucketListOptions) (
+	storj.BucketList, error) {
+	return storj.BucketList{}, nil
+}
+
+// DeleteBucket deletes a bucket and returns an error if there was a problem
+func (client *Client) DeleteBucket(ctx context.Context, bucket string) (err error) {
```
jtolds: there should be very few methods on the *Client itself. this shouldn't be here. what macaroon is it using? Your client constructor takes identity *identity.FullIdentity and cfg Config. actions like this need to be on an object that was constructed with access control.

lib/uplink/buckets.go
```diff
@@ -0,0 +1,146 @@
...
+		return minio.BucketNotEmpty{Bucket: bucket}
+	}
+
+	err = client.metainfo.DeleteBucket(ctx, bucket)
+
+	return convertError(err, bucket, "")
+}
+
+// GetBucketInfo gets the bucket information and returns the info and an error type
+func (client *Client) GetBucketInfo(ctx context.Context, bucket string) (bucketInfo minio.BucketInfo, err error) {
```
jtolds: same

lib/uplink/buckets.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if err != nil {
+		return minio.BucketInfo{}, convertError(err, bucket, "")
+	}
+
+	return minio.BucketInfo{Name: info.Name, Created: info.Created}, nil
+}
+
+// ListBuckets returns a list of buckets in the store
+func (client *Client) ListBuckets(ctx context.Context) (bucketItems []minio.BucketInfo, err error) {
```
jtolds: same (and below)

lib/uplink/buckets.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		startAfter = list.Items[len(list.Items)-1].Name
+	}
+
+	return bucketItems, err
+}
+
+// MakeBucketWithLocation makes a bucket with a location and returns an error if there were any problems.
+func (client *Client) MakeBucketWithLocation(ctx context.Context, bucket string, location string) (err error) {
```
jtolds: MakeBucketWithLocation is an anachronism from miniogw. This should just be CreateBucket and shouldn't need a location

jtolds: also, shouldn't be on a *Client

lib/uplink/config.go
```diff
@@ -0,0 +1,149 @@
...
+// Config uplink configuration
+type Config struct {
+	Client ClientConfig
+	RS     RSConfig
+	Enc    EncryptionConfig
+	TLS    tlsopts.Config
+}
+
+// GetMetainfo returns an implementation of storj.Metainfo
+func (c Config) GetMetainfo(ctx context.Context, identity *identity.FullIdentity) (db storj.Metainfo, ss streams.Store, err error) {
```
jtolds: this shouldn't be exposed

lib/uplink/partial.go
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2018 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package uplink
+
+import "context"
+
+// NewPartialUpload starts a new partial upload and returns that partial
+// upload id
```
jtolds: let's skip on these for now

lib/uplink/access.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Access is all of the access information an application needs to store and
+// retrieve data. Someone with a share may have no restrictions within a project
+// (can create buckets, list buckets, list files, upload files, delete files,
+// etc), may be restricted to a single bucket, may be restricted to a prefix
+// within a bucket, or may even be restricted to a single file within a bucket.
+type Access struct {
```
jtolds: this comment is no longer correct. the old Access got renamed to be Permissions, so this comment belongs more with Permissions now. Access is more like Session

lib/uplink/access.go
```diff
@@ -0,0 +1,112 @@
...
+	Permissions Permissions
+	Uplink      *Uplink
+}
+
+// A Macaroon represents an access credential to certain resources
+type Macaroon interface {
```
jtolds: to be honest, let's rename this interface to be APIKey for now and get rid of the Restrict method and the Caveat type.

lib/uplink/access.go
```diff
@@ -0,0 +1,112 @@
...
+	Serialize() ([]byte, error)
+	Restrict(caveats ...Caveat) Macaroon
+}
+
+// Permissions are parsed by Uplink and return an Access struct
+type Permissions struct {
+	Macaroon Macaroon
```
jtolds: and call this APIKey

lib/uplink/access.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Caveat could be a read-only restriction, a time-bound
+// restriction, a bucket-specific restriction, a path-prefix restriction, a
+// full path restriction, etc.
+type Caveat interface {
+}
+
+// ParseAccess parses a serialized Access
+func ParseAccess(data []byte) (Access, error) {
```
jtolds: these two methods no longer make much sense. let's remove them

lib/uplink/access.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return b, nil
+}
+
+// GetBucket returns a Bucket with the given Encryption information
+func (a *Access) GetBucket(ctx context.Context, bucket string, encryption storj.EncryptionScheme) *Bucket {
+	opts := &BucketOpts{
+		PathCipher:       encryption.Cipher,
+		EncryptionScheme: a.Uplink.config.GetEncryptionScheme(),
```
jtolds: can we let people specify this per bucket?

lib/uplink/buckets.go
```diff
@@ -0,0 +1,30 @@
...
+package uplink
+
+import (
+	"storj.io/storj/pkg/storj"
+	ul "storj.io/storj/uplink"
+)
+
+// BucketOpts holds the cipher, path, key, and enc. scheme for each bucket since they
+// can be different for each
+type BucketOpts struct {
```
jtolds: can we call this Encryption or something?

lib/uplink/common.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2018 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package uplink
+
+import "storj.io/storj/pkg/storj"
+
+// Encryption specifies an individual bucket's encryption choices
+type Encryption struct {
```
navillasa: how is this different from BucketOpts? or why does it not have `EncryptionScheme` on it as a field? could they be the same struct?

lib/uplink/main.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2018 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package uplink
+
+import (
+	"context"
+
+	"storj.io/storj/pkg/identity"
+	ul "storj.io/storj/uplink"
```
navillasa: nit - I think I've seen preference to not do this shortened import name thing in the repo, unless you know otherwise.

lib/uplink/buckets.go
```diff
@@ -0,0 +1,30 @@
...
+	PathCipher       storj.Cipher
+	EncPathPrefix    storj.Path
+	Key              storj.Key
+	EncryptionScheme storj.EncryptionScheme
+}
+
+// Bucket is a struct that allows operations on a Bucket after a user providers Permissions
+type Bucket struct {
```
navillasa: should `storj.Bucket` be on this struct? I'm looking at this but not sure how current it is: (https://github.com/storj/storj/pull/847/files)

lib/uplink/objects.go
```diff
@@ -0,0 +1,76 @@
...
+	IsPrefix bool
+
+	Metadata map[string]string
+
+	Created  time.Time
+	Modified time.Time
+	Expires  time.Time
+
+	Size     int64
+	Checksum string
```
navillasa: should Checksum should be `[]byte`?

lib/uplink/objects.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// this differs from storj.Object by not having Version (yet), and not
+	// having a Stream embedded. I'm also not sold on splitting ContentType out
+	// from Metadata but shrugemoji.
+}
+
+// ObjectPutOpts controls options about uploading a new Object, if authorized.
+type ObjectPutOpts struct {
```
navillasa: I think there should also be an Object struct with just the `ObjectMeta` on it (I'm just looking at this though, might be wrong https://github.com/storj/storj/pull/847/files)

lib/uplink/objects.go
dylanlott: It should be! I just haven't changed it yet 👍 It will be in the next push though

lib/uplink/buckets.go
dylanlott: Fixed 👍

lib/uplink/main.go
dylanlott: I needed a way to acces /uplink but since this package is called `uplink`I had to do this weirdness.

navillasa: ahh ok!

lib/uplink/buckets.go
dylanlott: Updated

lib/uplink/common.go
dylanlott: It's been renamed to Encryption 👍

lib/uplink/access.go
dylanlott: Fixed 👍

lib/uplink/main.go
egonelbre: There is no conflict in importing `uplink`.

lib/uplink/buckets.go
```diff
@@ -0,0 +1,129 @@
...
+
+	reader := bytes.NewReader(data)
+	mutableStream, err := obj.CreateStream(ctx)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	upload := stream.NewUpload(ctx, mutableStream, streams)
+
+	_, err = io.Copy(upload, reader)
```
navillasa: might need `Error.Wrap` if it's not nil

lib/uplink/libuplink_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// Get these configs figured out cause they're throwing hella errors
+	config := getDefaultConfig()
+	config.Client.OverlayAddr = satellite.Addr()
+	config.Client.PointerDBAddr = satellite.Addr()
+	config.Client.APIKey = planet.Uplinks[0].APIKey[satellite.ID()]
+
+	config.RS.MinThreshold = 1 * len(planet.StorageNodes) / 5
+	config.RS.RepairThreshold = 2 * len(planet.StorageNodes) / 5
+	config.RS.SuccessThreshold = 3 * len(planet.StorageNodes) / 5
+	config.RS.MaxThreshold = 4 * len(planet.StorageNodes) / 5
```
navillasa: why divided by 5? just wondering

lib/uplink/objects.go
```diff
@@ -0,0 +1,84 @@
...
+
+	Encryption *Encryption
+}
+
+// ListObjectsField numbers the fields of list objects
+type ListObjectsField int
+
+const (
+	// ListObjectsMetaNone opts
+	ListObjectsMetaNone ListObjectsField = 0
```
aleitner: you can just set this equal to iota and then not have any`=` for the following constants

lib/uplink/libuplink_test.go
dylanlott: I emulated what Egon did in the testplanet uplink tests to get the numbers here.

PR #1455, opened 2019-03-12
hammer: EOF handling

egonelbre:


pkg/accounting/tally/tally.go
```diff
@@ -50,6 +50,8 @@ func New(logger *zap.Logger, accountingDB accounting.DB, bwAgreementDB bwagreeme
 
 // Run the Tally loop
 func (t *Tally) Run(ctx context.Context) (err error) {
+	return nil
```
navillasa: what does returning nil at the top do?

egonelbre: it's there to prevent tally from running, it's disabled for now because the tally/audit haven't been updated yet for the new protocol

navillasa: ohh okay

PR #1458, opened 2019-03-12
TLS extension handling overhaul

bryanchriswhite:
+ refactor and reorganize extension-related code
+ add more and better tests

#### This is a prerequisite for [identityV2](https://storjlabs.atlassian.net/browse/V3-1188?atlOrigin=eyJpIjoiMGY5ODE5MmEyZTc2NDVkNGI3NDZjZjhjYzI3ZGFkMjMiLCJwIjoiaiJ9)


pkg/peertls/tlsopts/options.go
```diff
@@ -45,45 +47,48 @@ func NewOptions(i *identity.FullIdentity, c Config) (*Options, error) {
...
 	}
 
 	return opts, nil
 }
 
+func (opts *Options) ExtensionOptions() peertls.ExtensionOptions {
+	return peertls.ExtensionOptions{
+		PeerCAWhitelist: opts.PeerCAWhitelist,
+		RevDB:           opts.RevDB,
+		PeerIDVersions:  opts.Config.Extensions.PeerIDVersions,
```
bryanchriswhite: Move to identity versioning PR

pkg/peertls/extensions/extensions_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testpeertls"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls/extensions"
+	"storj.io/storj/pkg/peertls/tlsopts"
+)
+
+func TestParseExtensions(t *testing.T) {
```
bryanchriswhite: The file that originally contained this test was moved... this test is bad but does cover some things and needs to be split up and cleaned up. While I did originally write it, it was a while ago - don't judge me. I made the minimal change to keep it working for now.

pkg/peertls/extensions/extensions.go
```diff
@@ -0,0 +1,211 @@
...
+
+func init() {
+	// NB: register all handlers defined in this file.
+	AllHandlers.Register(
+		CAWhitelistSignedLeafHandler,
+	)
+}
+
+// NewHandler builds a `Handler` pointer from an `ExtensionID` and a `handlerFactoryFunc`.
+func NewHandler(id *ExtensionID, handlerFactory HandlerFactoryFunc) *HandlerFactory {
```
mobyvb: NewHandler -> NewHandlerFactory

pkg/peertls/extensions/extensions.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// ErrVerifyCASignedLeaf is used when a signed leaf extension signature wasn't produced
+	// by any CA in the whitelist.
+	ErrVerifyCASignedLeaf = Error.New("leaf not signed by any CA in the whitelist")
+	// ErrUniqueExtensions is used when multiple extensions have the same Id
+	ErrUniqueExtensions = Error.New("extensions are not unique")
+)
+
+// ExtensionID is an alias to an `asn1.ObjectIdentifier`.
+type ExtensionID asn1.ObjectIdentifier
```
mobyvb: type ExtensionID = asn1.ObjectIdentifier

pkg/peertls/extensions/revocations_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testidentity"
+	"storj.io/storj/internal/testpeertls"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls"
+	"storj.io/storj/pkg/peertls/extensions"
+	"storj.io/storj/storage"
+)
+
+func TestRevocationCheckExtensionHandler(t *testing.T) {
```
mobyvb: Should not contain word extension

pkg/peertls/tlsopts/options.go
```diff
@@ -37,58 +41,110 @@ type VerificationFuncs struct {
...
+	for _, cert := range chain {
+		for _, ext := range cert.ExtraExtensions {
+			extensionMap[ext.Id.String()] = ext
+		}
+	}
+	return extensionMap
+}
+
+// ExtensionOptions converts options for use in extension handling.
+func (opts *Options) ExtensionOptions() *extensions.Options {
```
mobyvb: check whether this is necessary still

pkg/peertls/tlsopts/options.go
```diff
@@ -37,58 +41,110 @@ type VerificationFuncs struct {
...
+		opts.RevDB, err = identity.NewRevDB(opts.Config.RevocationDBURL)
 		if err != nil {
 			return err
 		}
-		opts.VerificationFuncs.Add(identity.VerifyUnrevokedChainFunc(opts.RevDB))
 	}
 
-	exts := peertls.ParseExtensions(c.Extensions, parseOpts)
-	opts.VerificationFuncs.Add(exts.VerifyFunc())
+	//opts.VerificationFuncs.Add(peertls.AllExtensionHandlers.VerifyFunc(opts.Options()))
```
mobyvb: remove comment

cmd/identity/main.go
```diff
@@ -218,8 +218,8 @@ func printExtensions(cert []byte, exts []pkix.Extension) error {
 	for _, e := range exts {
 		var data interface{}
 		switch e.Id.String() {
-		case peertls.ExtensionIDs[peertls.RevocationExtID].String():
-			var rev peertls.Revocation
+		case extensions.RevocationExtID.String():
```
egonelbre: Maybe use `e.Id.Equal(extensions.RevocationExtID)` instead.

internal/testidentity/revocation_db.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/peertls/extensions"
+	"storj.io/storj/storage"
+)
+
+// RevocationDBsTest runs the passed test function with each type of revocation database.
+func RevocationDBsTest(ctx *testcontext.Context, t *testing.T, test func(*testing.T, extensions.RevocationDB, storage.KeyValueStore)) {
```
egonelbre: Each root-subtest probably should use a separate testcontext otherwise the files in the same folder can end up conflicting.

pkg/identity/revocation_db.go
```diff
@@ -22,25 +22,73 @@ type RevocationDB struct {
 	DB storage.KeyValueStore
 }
 
+// NewRevDB returns a new revocation database given the URL
+func NewRevDB(revocationDBURL string) (*RevocationDB, error) {
```
egonelbre: `NewRevocationDB`

pkg/identity/revocation_db.go
```diff
@@ -22,25 +22,73 @@ type RevocationDB struct {
...
+		}
+	default:
+		return nil, extensions.ErrRevocationDB.New("database scheme not supported: %s", driver)
+	}
+
+	return db, nil
+}
+
+// NewRevocationDBBolt creates a bolt-backed RevocationDB
+func NewRevocationDBBolt(path string) (*RevocationDB, error) {
```
egonelbre: Do we need to expose the specific implementations?

internal/testidentity/revocation_db.go
```diff
@@ -0,0 +1,50 @@
...
+		require.NoError(t, err)
+
+		{
+			// Test using redis-backed revocation DB
+			dbURL := "redis://" + redisServer.Addr() + "?db=0"
+			redisRevDB, err := identity.NewRevDB(dbURL)
+			require.NoError(t, err)
+
+			test(t, redisRevDB, redisRevDB.DB)
+			ctx.Check(redisRevDB.Close)
```
egonelbre: Use `defer ctx.Check(db.Close)` after construction that way when the test panics it won't leave garbage around

internal/testidentity/revocation_db.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		redisServer.Close()
```
egonelbre: use `defer redisServer.Close()`, in case of panics/t.Fatal, this won't be run.

pkg/identity/revocation_db_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	testidentity.RevocationDBsTest(ctx, t, func(t *testing.T, revDB extensions.RevocationDB, db storage.KeyValueStore) {
+		// NB: key indices are reversed as compared to chain indices
+		keys, chain, err := testpeertls.NewCertChain(2)
+		require.NoError(t, err)
+
+		firstRevocation, err := extensions.NewRevocationExt(keys[0], chain[peertls.LeafIndex])
+		require.NoError(t, err)
+
+		time.Sleep(time.Second)
```
egonelbre: why is this sleep here?

pkg/peertls/extensions/extensions.go
```diff
@@ -0,0 +1,211 @@
...
+func NewHandler(id *ExtensionID, handlerFactory HandlerFactoryFunc) *HandlerFactory {
+	return &HandlerFactory{
+		id:      id,
+		factory: handlerFactory,
+	}
+}
+
+// AddSignedCertExt generates a signed certificate extension for a cert and attaches
+// it to that cert.
+func AddSignedCertExt(key crypto.PrivateKey, cert *x509.Certificate) error {
```
egonelbre: `extensions.AddSignedCertExt` is the full name, so maybe something like `extensions.AddSignedCert` would be nicer, because the `Ext` part is already part of the package name.

pkg/peertls/extensions/extensions.go
```diff
@@ -0,0 +1,211 @@
...
+
+// Options holds common options for use in handling extensions.
+type Options struct {
+	PeerCAWhitelist []*x509.Certificate
+	RevDB           RevocationDB
+}
+
+// HandlerFactories is a collection of `HandlerFactory`s for convenience.
+// Defines `Register` and `WithOptions` methods.
+type HandlerFactories []*HandlerFactory
```
egonelbre: These factories etc. make me feel like there is a nicer design around the corner, but I'm unsure what it exactly is.

pkg/identity/revocation_db.go
bryanchriswhite: what do you mean?

egonelbre: Ah, just whether `NewRevocationDBBolt` needs to be public or not?

bryanchriswhite: ah - good question.. i'll check.

internal/testpeertls/revocations.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	revocation = revokingCert.ExtraExtensions[0]
+	return append([]*x509.Certificate{revokingCert}, chain[peertls.CAIndex:]...), revocation, nil
+}
+
+// RevokeCA revokes the CA certificate in the passed chain and adds a revocation
+// extension to that certificate, recording this action.
+func RevokeCA(keys []crypto.PrivateKey, chain []*x509.Certificate) ([]*x509.Certificate, pkix.Extension, error) {
+	caCert := chain[peertls.CAIndex]
```
coyle: is it possible for **peertls.CAIndex** to be larger than the size of chain ?

pkg/identity/revocation_db.go
```diff
@@ -22,25 +22,73 @@ type RevocationDB struct {
...
 	}
 	if revBytes == nil {
 		return nil, nil
 	}
 
-	rev := new(peertls.Revocation)
+	rev := new(extensions.Revocation)
 	if err = rev.Unmarshal(revBytes); err != nil {
-		return rev, peertls.ErrRevocationDB.Wrap(err)
+		return rev, extensions.ErrRevocationDB.Wrap(err)
```
coyle: to make reviewing easier, it might be good to break name changes out into a separate PR

pkg/identity/revocation_db_test.go
coyle: I'm guessing based on the comment, it's because the timestamps need to be different so the time needs to change between creations. Is it possible instead of waiting until the time is actually different, can we just set the timestamp to be something else ?

pkg/peertls/extensions/extensions.go
coyle: @egonelbre Are you of the opinion that factory methods are an anti pattern ?

pkg/peertls/extensions/extensions_test.go
coyle: lol

pkg/identity/revocation_db.go
bryanchriswhite: Will be more careful in the future :+1:.

I would've done better git-foo to do it but there was package refactoring happening at the same time to avoid import cycles

internal/testpeertls/revocations.go
bryanchriswhite: Good point.. I'll address this in the next PR (#1554)

pkg/identity/revocation_db_test.go
bryanchriswhite: We _could_ pass the timestamp to `NewRevocationExt` but tests are the only place where we really want a time other than now.

pkg/peertls/tlsopts/options.go
bryanchriswhite: I think it is

pkg/identity/revocation_db_test.go
bryanchriswhite: @egonelbre: what @coyle said

pkg/peertls/extensions/extensions.go
bryanchriswhite: i'm interested in the answer as well

bryanchriswhite: maybe just more meaningful names would already be a good start

PR #1459, opened 2019-03-12
Extracts utility methods for properly calculating piece size

kaloyan-raev:


pkg/pb/metainfo.proto
```diff
@@ -84,7 +84,7 @@ message SegmentWriteRequest {
     bytes path = 2;
     int64 segment = 3;
     pointerdb.RedundancyScheme redundancy = 4;
-    int64 max_segment_size = 5;
+    int64 max_encrypted_segment_size = 5;
```
egonelbre: I wonder whether we should request with the max encrypted segment size or max encrypted piece size?

kaloyan-raev: I can do it either way. But in any case the Metainfo Service must still call `eestream.CalcPieceSize` on `DownloadSegment`, because we keep the encrypted segment size in the pointer, but not the piece size.

PR #1466, opened 2019-03-12
Repair queue isolation level fix

wthorp:
The ticket says "postgres / sqlite need retry logic for transactions".  The particular PR handles "figure out why we need to skip this test".  Other PRs may follow.

https://storjlabs.atlassian.net/browse/V3-1016

The SELECT/DELETE SQL transaction in repairqueue.go Dequeue() starts off with roughly 'a read lock'.  This means that several concurrent go routines read the same entry instead of different entries.  Later the deletes fail and the queue has rogue entries.

Changes in this PR:
Unskipped parallel repair queue test
Moved Repairqueue Dequeue() from DBX to SQL so I could fix isolation level
Implemented custom SQLite and Postgres Dequeue handlers for performance reasons
Added InjuredSegment convenience SQL Scanner / Valuer methods for easy marshaling
Added Context to some SQL calls to make them cancellable


satellite/satellitedb/repairqueue.go
```diff
@@ -31,38 +34,56 @@ func (r *repairQueue) Enqueue(ctx context.Context, seg *pb.InjuredSegment) error
...
-	} else if !deleted {
-		return pb.InjuredSegment{}, Error.Wrap(utils.CombineErrors(Error.New("Injured segment not deleted"), tx.Rollback()))
+		rows.Close()
+		return seg, err
 	}
-	if err := tx.Commit(); err != nil {
-		return pb.InjuredSegment{}, Error.Wrap(err)
+	rows.Close()
+	//delete
+	res, err := tx.Exec(r.db.Rebind(`DELETE FROM injuredsegments WHERE id = ?`), id)
```
thepaul: In the PostgreSQL case, we can actually do this whole function in one simpler query:

```go
func (r *repairQueue) Dequeue(ctx context.Context) (seg pb.InjuredSegment, err error) {
    err = r.db.DB.QueryRowContext(ctx, `
        DELETE FROM injuredsegments
            WHERE id = (
                SELECT id FROM injuredsegments
                    ORDER BY something
                    FOR UPDATE SKIP LOCKED
                    LIMIT 1
            )
            RETURNING info
    `).Scan(&seg)
    return seg, err // err will be sql.ErrNoRows when there were no rows left in the table
}
```

This has advantages besides being shorter: no explicit transaction is necessary, and the query optimizer will be able to see what you're doing and do it more efficiently!

Buuuuuuuut for the sqlite case, I'm pretty sure you won't be able to do `DELETE ... RETURNING`. So you'd still need to have a transaction, do a QueryRowContext to get both the id and the info, and do a delete statement and check the errors and RowsAffected. So maybe I haven't saved you any code. But at least it would work better in PG land...

egonelbre:
Note V3-1016 is to ensure that transactions (and all queries) work in multi-server setup. And there are multiple places where transactions are used, not just the queue. The `TestParallel` was just the obvious case. Also after fixing the transactions then `mutex` shouldn't be necessary on sqlite implementation.

PR #1470, opened 2019-03-13
Rework ECClient tests on testplanet

kaloyan-raev:


pkg/storage/ec/client_planet_test.go
```diff
@@ -0,0 +1,172 @@
...
+	dataSize     = 32 * memory.KiB
+	storageNodes = 4
+)
+
+func TestECClient(t *testing.T) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	planet, err := testplanet.New(t, 1, storageNodes, 1)
+	if !assert.NoError(t, err) {
```
mniewrzal: you can use `require.NoError(t, err)` to fail fast

kaloyan-raev: True. Missed this one.

PR #1472, opened 2019-03-13
hammer: Update segment.Store tests

mniewrzal:


pkg/storage/segments/store_test.go
```diff
@@ -484,6 +274,50 @@ func TestCalcNeededNodes(t *testing.T) {
...
+	require.NoError(t, err)
+
+	metainfo, err := planet.Uplinks[0].DialMetainfo(context.Background(), planet.Satellites[0], TestAPIKey)
+	require.NoError(t, err)
+
+	ec := ecclient.NewClient(planet.Uplinks[0].Transport, 0)
+	fc, err := infectious.NewFEC(4, 6)
+	require.NoError(t, err)
+
+	rs, err := eestream.NewRedundancyStrategy(eestream.NewRSScheme(fc, 1*memory.KB.Int()), 4, 6)
```
kaloyan-raev: Use `KiB` instead of `KB`.

kaloyan-raev: Any specific reason to set the `repairThreshold` to the `requiredCount` in the `NewRedundancyStrategy` constructor. If you don't care, just pass `0` for both `repairThreshold` and `optimalThreshold`.

pkg/storage/segments/store_test.go
```diff
@@ -1,464 +1,254 @@
...
-	mockPDB := mock_pointerdb.NewMockClient(ctrl)
-	rs := eestream.RedundancyStrategy{
-		ErasureScheme: mock_eestream.NewMockErasureScheme(ctrl),
-	}
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
 
-	ss := segmentStore{mockOC, mockEC, mockPDB, rs, 10}
-	assert.NotNil(t, ss)
+	planet, err := testplanet.New(t, 1, 10, 1)
```
kaloyan-raev: Do you really need 10 storage nodes for this test? If not reduce the number to the required for the tests. This will reduce the test time.

pkg/storage/segments/store_test.go
```diff
@@ -1,464 +1,196 @@
...
-
-		_, err := ss.Put(ctx, strings.NewReader(tt.readerContent), tt.expiration, func() (storj.Path, []byte, error) {
-			return tt.pathInput, tt.mdInput, nil
+		runTest(t, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet, segmentStore segments.Store) {
+			_, err := segmentStore.Put(ctx, bytes.NewReader(tt.content), tt.expiration, func() (storj.Path, []byte, error) {
+				return tt.path, tt.metadata, nil
+			})
+			require.NoError(t, err, tt.name)
+
+			_, _, err = segmentStore.Get(ctx, tt.path)
```
kaloyan-raev: Should we assert that the expected data can be read from the returned ranger? Also that the metadata is the expected one?

mniewrzal: I will add this

PR #1475, opened 2019-03-13
Add testplanet test for segment store repairer

mobyvb:
https://storjlabs.atlassian.net/browse/V3-1377


kaloyan-raev:
@mobyvb Team Hammer is changing the communication protocol with storage nodes. This leads to lots of changes to other components, including the repairer. Most of this PR is in conflict with our changes.

We are currently adapting the Repairer (works still in progress) to use the new protocol. Perhaps, it would be best to adapt this PR to the new repairer. All our changes are currently in the "hammer" branch.

PR #1479, opened 2019-03-13
add bwa_bucket table, add migrations

JessicaGreben:
This PR does the following changes:
1. Create a new table, `bucket_bandwidth_usage`, which creates a relationship between buckets and bwagreements.
2. When a `PayerBandwidthAllocation/pb.OrderLimit` is created, we now also create a `bucket_bandwidth_usage` record.
3. Add functionality into the Tally service to sum up the `bucket_bandwidth_usage` records for each bucket.
4. Add functionality in the Rollup service to create a `BucketRollup` in the `bucket_usage` table.

This PR is a WIP.  Still todo:
1. This PR will need to be reworked once Hammer branch is merged.
2. Still need to fully test these changes to make sure it behaves as expected.
3. Still need to write tests for this.
 
Open questions:
1. Currently creating a bucket_bandwidth_usage record in pointerdb/allocation.go in `PayerBandwidthAllocation` method.  Is this the right place for creating it?
2. What time should be set for RollupEndTime?


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -333,7 +333,7 @@ model bucket_usage (
     unique rollup_end_time bucket_id
 
     field  id                 blob
-    field  bucket_id          blob      //--TODO: add foreign key constraint--//
+    field  bucket_id          text      //--TODO: add foreign key constraint--//
```
egonelbre: this should be a blob, text implies valid utf8

satellite/satellitedb/migrate.go
```diff
@@ -341,6 +341,28 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				Description: "Change bucket_id type",
+				Version:     6,
```
egonelbre: Version must always be incremented with each step... but you can do multiple SQL statements in a `migrate.SQL`, it's a list of strings.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: the bucket id will be a string in the form "projectname/bucketname".  should it still be a blob?

egonelbre: Yes. I think we should have the project name in a separate field. Unless there's a reason we cannot keep them in the same field, I guess.

JessicaGreben: Currently there is no concept of a proper bucketID that is a blob. There is only this bucket name string.  I can add a separate field for the bucket_id string, I guess I will call it `bucket_name` instead, but for now that would be the bucket_id blob field will always be empty.  Does that sound right?

egonelbre: Note, the correct translation of `string` in Go for databases usually is `[]byte`, `blob` or `bytes` because Go strings are not guaranteed to be utf-8. There must be other checks in place for them to be considered utf-8.

satellite/satellitedb/dbx/satellitedb.dbx.postgres.sql
```diff
@@ -131,6 +131,15 @@ CREATE TABLE api_keys (
 	UNIQUE ( key ),
 	UNIQUE ( name, project_id )
 );
+CREATE TABLE bwagreement_buckets (
+	id bytea NOT NULL,
+	serialnum text NOT NULL REFERENCES bwagreements( serialnum ) ON DELETE CASCADE,
```
egonelbre: This would need to be `bytes`.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -390,3 +390,25 @@ read one (
...
 	where  certRecord.id = ?
 )
+
+//--- bwagreement_bucket ---//
+
+model bwagreement_bucket (
+	key id
+
+	field id              blob
+	field serialnum       bwagreement.serialnum cascade
```
egonelbre: This would need to be `bytes`, because of other reasons. With hammer it's going to be a bytes, it was previously encoded as a string, for unfortunate reasons. I'm not sure what the effects of cascade deletes will be with large tables.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: So keep the bucket_id blob field.  Add a project_name string field, and add a bucket_name string field.  Is that what you mean?

egonelbre: I would do:
```
field project_id blob
field bucket_id blob
```

JessicaGreben: oh ok i didn't know string could mean `blob` in database

egonelbre: blob is effectively a `[]byte` and go strings are `[]byte` with some sugar.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: I haven't made any changes to this yet.  We need to test the performance of this first.  I'm going to work with stefan to test it. Will post results once we have them.

egonelbre: ah, also... use `serial_number`, we need to fix the other eventually as well

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -390,3 +391,26 @@ read one (
 	select certRecord
 	where  certRecord.id = ?
 )
+
+//--- bwagreement_bucket ---//
+
+model bwagreement_bucket (
```
egonelbre: `bucket_bandwidth_usage`

satellite/satellitedb/migrate.go
```diff
@@ -341,6 +341,23 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				Description: "Add bwagreement_buckets table, alter bucket_usage columns",
+				Version:     7,
+				Action: migrate.SQL{
+					`ALTER TABLE bucket_usages ALTER COLUMN bucket_id TYPE BYTEA;`,
```
egonelbre: it was already `BYTEA`

JessicaGreben: oops, let me fix

PR #1486, opened 2019-03-14
hammer: more linting

egonelbre:


storagenode/trust/service.go
```diff
@@ -24,12 +25,14 @@ type Pool struct {
 	trustedSatellites  map[storj.NodeID]*satelliteInfoCache
 }
 
+// satellite info cache, caches information about a satellite.
```
kaloyan-raev: Should it be `satelliteInfoCache` to match the type name?

egonelbre: yes

PR #1487, opened 2019-03-14
hammer: Piecestore inspector/dashboard

mniewrzal:


storagenode/peer.go
```diff
@@ -99,9 +99,10 @@ type Peer struct {
 	}
 
 	Storage2 struct {
-		Trust    *trust.Pool
-		Store    *pieces.Store
-		Endpoint *piecestore.Endpoint
+		Trust     *trust.Pool
+		Store     *pieces.Store
+		Endpoint  *piecestore.Endpoint
+		Inspector *piecestore.Inspector
```
egonelbre: I guess we can remove the other reference.

mniewrzal: true

storagenode/piecestore/inspector.go
```diff
@@ -4,14 +4,121 @@
...
+	totalUsedSpace, err := inspector.pieceInfo.SpaceUsed(ctx)
+	if err != nil {
+		return nil, err
+	}
+	usage, err := inspector.usageDB.Summary(ctx, getBeginningOfMonth(), time.Now())
+	if err != nil {
+		return nil, err
+	}
+
+	totalUsedBandwidth := usage.Get + usage.GetAudit + usage.GetRepair + usage.Put + usage.PutRepair +
```
egonelbre: add it as a method on BandwidthUsage

storagenode/bandwidth/usage.go
```diff
@@ -52,3 +52,9 @@ func (usage *Usage) Include(action pb.Action, amount int64) {
 		usage.Unknown += amount
 	}
 }
+
+// TotalUsedBandwidth sums all type of bandwidth
+func (usage *Usage) TotalUsedBandwidth() int64 {
```
egonelbre: `Total` is sufficient, since it's full name is `bandwidth.Usage.Total()`.

storagenode/piecestore/inspector.go
```diff
@@ -4,14 +4,120 @@
...
+	return statsSummary, nil
+}
+
+func (inspector *Inspector) getDashboardData(ctx context.Context) (*pb.DashboardResponse, error) {
+	statsSummary, err := inspector.retrieveStats(ctx)
+	if err != nil {
+		return &pb.DashboardResponse{}, Error.Wrap(err)
+	}
+
+	nodes, err := inspector.kademlia.FindNear(ctx, storj.NodeID{}, 10000000)
```
kaloyan-raev: I wonder what would be the performance impact of asking Kademlia to find the nearest 10 million nodes in a large network?

egonelbre: I guess, not great.

kaloyan-raev: We have also `kademlia.DumpNodes` that returns all noes from the local routing table.

If the goal is just to count the number of nodes in the routing table, then perhaps it would be better to add `kademlia.NodeCount` method that will make this in a more optimal way, without wasting memory, but just increasing a counter.

egonelbre: `DumpNodes` is used for debugging kademlia. Just remembered that, `FindNear` looks up from local database and it doesn't actually query the network.

kaloyan-raev: But we just want to count the number of nodes in the routing table (i.e. the neighborhood size), shouldn't we have a more efficient way? Remember, the dashboard will call this every 3 seconds along with other things.

egonelbre: Ah, right... yeah. We can improve that later I guess.

kaloyan-raev: Sure. Let's just have a TODO comment for now.

mniewrzal: @kaloyan-raev can you add TODO?

egonelbre: added

PR #1490, opened 2019-03-14
hammer: remove psclient

egonelbre:


aleitner:
Not much to review lol

PR #1494, opened 2019-03-14
Add repairer tests

mobyvb:
https://storjlabs.atlassian.net/browse/V3-1377

do not merge until https://github.com/storj/storj/pull/1525 is merged


pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package segments_test
+
+import (
+	"crypto/rand"
+	"testing"
+	time "time"
```
egonelbre: this doesn't need an alias

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/pkg/overlay"
+	"storj.io/storj/pkg/pb"
+	ecclient "storj.io/storj/pkg/storage/ec"
+	"storj.io/storj/pkg/storage/segments"
+	"storj.io/storj/pkg/storj"
+)
+
+func TestSegmentStoreRepair(t *testing.T) {
+	numStorageNodes := 10
```
egonelbre: specify this in config and use `len(planet.StorageNodes)` elsewhere

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		// first, upload some remote data
+		uplink := planet.Uplinks[0]
+		satellite := planet.Satellites[0]
+
+		testData := make([]byte, 5*memory.MiB)
+		_, err := rand.Read(testData)
+		assert.NoError(t, err)
+
+		err = uplink.Upload(ctx, satellite, "test/bucket", "test/path", testData)
```
egonelbre: use `testbucket`

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		assert.NoError(t, err)
```
egonelbre: `require.NoError(t, err)`, since the test following it doesn't make sense if it fails.

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package segments_test
+
+import (
+	"crypto/rand"
```
egonelbre: Use `math/rand` in tests, it's faster

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		_, err := rand.Read(testData)
+		assert.NoError(t, err)
+
+		err = uplink.Upload(ctx, satellite, "test/bucket", "test/path", testData)
+		assert.NoError(t, err)
+
+		// get a remote segment from pointerdb
+		pdb := satellite.Metainfo.Service
+		listResponse, _, err := pdb.List("", "", "", true, 0, 0)
+		assert.NoError(t, err)
```
egonelbre: `require.NoError(t, err)`

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		var pointer *pb.Pointer
+		for _, v := range listResponse {
+			path = v.GetPath()
+			pointer, err = pdb.Get(path)
+			assert.NoError(t, err)
+			if pointer.GetType() == pb.Pointer_REMOTE {
+				break
+			}
+		}
+		// calculate how many storagenodes to kill
```
egonelbre: add a line break above here, it makes the code slightly clearer to read

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		nodesToKeepAlive := make(map[storj.NodeID]bool)
+		for i, piece := range remotePieces {
+			if i >= toKill {
+				nodesToKeepAlive[piece.NodeId] = true
+				continue
+			}
+			nodesToKill[piece.NodeId] = true
+			lostPieces = append(lostPieces, piece.GetPieceNum())
+		}
+		for _, node := range planet.StorageNodes {
```
egonelbre: not completely sure, but I think you could do here `for _, node := range planet.StorageNodes[:toKill] {` and later use `for _, node := range planet.StorageNodes[toKill:] {`.

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+			if nodesToKill[node.ID()] {
+				err = planet.StopPeer(node)
+				assert.NoError(t, err)
+			}
+		}
+
+		// repair segment
+		overlayDB := satellite.DB.OverlayCache()
+		statDB := satellite.DB.StatDB()
+		oc := overlay.NewCache(overlayDB, statDB)
```
egonelbre: Don't create a new cache, use the existing one on satellite.

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		as := satellite.Metainfo.Allocation
+		ec := ecclient.NewClient(satellite.Transport, 0)
+		repairer := segments.NewSegmentRepairer(pdb, as, oc, ec, satellite.Identity, &overlay.NodeSelectionConfig{}, time.Minute)
```
egonelbre: I think the `NodeSelectionConfig` should be properly selected.

mobyvb: What do you mean? Get it from the satellite or pass in nil?

pkg/storage/segments/repairer_test.go
mobyvb: I do not think this would work because `planet.StorageNodes` does not necessarily correlate to the storagenodes in the remote pieces from pointerdb

pkg/storage/segments/repairer_test.go
egonelbre: I think, get it from satellite.

mobyvb: It does not look like it is publicly accessible from the satellite Peer. In satellitePeer.New(), `nodeSelectionConfig` is created and passed into the `repairer.NewService()` as well as `overlay.NewServer()` and `metainfo.NewEndpoint()`. In all three of these, the selection preferences are private

egonelbre: Ah... hmm... we can leave as is for now, if it doesn't break tests. There are other PR-s that might end up cleaning it.

pkg/eestream/encode.go
```diff
@@ -267,8 +267,8 @@ func (er *EncodedRanger) Range(ctx context.Context, offset, length int64) ([]io.
 // CalcPieceSize calculates what would be the piece size of the encoded data
 // after erasure coding data with dataSize using the given ErasureScheme.
 func CalcPieceSize(dataSize int64, scheme ErasureScheme) int64 {
-	stripes := dataSize / int64(scheme.StripeSize())
-	if dataSize%int64(scheme.StripeSize()) != 0 {
+	stripes := (dataSize + uint32Size) / int64(scheme.StripeSize())
```
navillasa: why do you have to add the uint32Size? just curious

pkg/storage/segments/repairer_test.go
```diff
@@ -0,0 +1,122 @@
...
+		oc := satellite.Overlay.Service
+		as := satellite.Metainfo.Allocation
+		ec := ecclient.NewClient(satellite.Transport, 0)
+		repairer := segments.NewSegmentRepairer(pdb, as, oc, ec, satellite.Identity, &overlay.NodeSelectionConfig{}, time.Minute)
+		assert.NotNil(t, repairer)
+
+		err = repairer.Repair(ctx, path, lostPieces)
+		assert.NoError(t, err)
+
+		// kill nodes kept alive to ensure repair worked
```
navillasa: this part is cool

pkg/eestream/encode.go
mobyvb: this is actually from kaloyan's PR here https://github.com/storj/storj/pull/1525

internal/testplanet/uplink.go
```diff
@@ -193,6 +193,55 @@ func (uplink *Uplink) Upload(ctx context.Context, satellite *satellite.Peer, buc
...
+
+	metainfo, streams, err := config.GetMetainfo(ctx, uplink.Identity)
+	if err != nil {
+		return err
+	}
+
+	encScheme := config.GetEncryptionScheme()
+	redScheme := config.GetRedundancyScheme()
+
+	// create bucket if not exists
```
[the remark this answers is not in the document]
mobyvb: This upload function is pretty much identical to the other upload function in this file, except for the Reed Solomon config part at the top

PR #1495, opened 2019-03-14
pkg/process: make unexpected config values and env variables fatal errors

jtolds:
a bunch of commandline utilities will need to be fixed up to work with the same config as the main application, such as the dashboard


pkg/process/exec_conf.go
```diff
@@ -217,11 +232,11 @@ func cleanup(cmd *cobra.Command) {
 		// okay now that logging is working, inform about the broken keys
 		if cmd.Annotations["type"] != "helper" {
 			for _, key := range brokenKeys {
-				logger.Sugar().Infof("Invalid configuration file key: %s", key)
+				logger.Sugar().Fatalf("Invalid configuration file key: %s", key)
```
brimstone: Does this mean that environment variables will cause a warning "Invalid configuration _file_ key" ?

egonelbre: It will stop the program from starting completely.

brimstone: I'll be more blunt. Can this be changed to include "environment variable" in the error message about environment variables please?

jtolds: that will require more work but okay

jtolds: whoops, actually, it turns out it will be painfully obvious already. the only way this line happens is if an env var that starts with STORJ_ isn't expected. which means that when it's a config file mistake it will look like "debug.missing_key" and when it's an env var it will look like "STORJ_DEBUG_MISSING_KEY" but i'll take out the explicit mention of a file. sufficient?

jtolds:
still to do: don't pass storj-sim env vars down from storj-sim to subprocesses

pkg/process/exec_conf.go
brimstone: I think that's ok.

PR #1497, opened 2019-03-15
hammer: few tests for piecestore and pieces

egonelbre:


storagenode/piecestore/endpoint_test.go
```diff
@@ -4,31 +4,84 @@
...
+
+		download, err := planet.Uplinks[0].DownloadStream(ctx, planet.Satellites[0], "test/bucket", "test/path")
+		require.NoError(t, err)
+
+		pos, err := download.Seek(tt.offset, io.SeekStart)
+		require.NoError(t, err)
+		assert.Equal(t, pos, tt.offset)
+
+		data := make([]byte, tt.size)
+		n, err := io.ReadFull(download, data)
```
mniewrzal: does it make sense to make `data` larger than `tt.size` to check if will read only `tt.size`?

egonelbre: not really, because the `io.ReadFull` reads the tt.size. Since there's no ranger, there's no way to set the upper limit here.

PR #1498, opened 2019-03-15
hammer: set node restrictions in FindStorageNodesRequest

kaloyan-raev:


satellite/metainfo/metainfo.go
```diff
@@ -129,11 +129,21 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
...
+	}
+
+	maxPieceSize := eestream.CalcPieceSize(req.GetMaxEncryptedSegmentSize(), redundancy)
+
 	// TODO most probably needs more params
 	request := &pb.FindStorageNodesRequest{
 		Opts: &pb.OverlayOptions{
-			Amount:       int64(req.Redundancy.Total),
-			Restrictions: &pb.NodeRestrictions{},
+			Amount: int64(req.Redundancy.Total),
```
mniewrzal: Do you think we still need this TODO above?

kaloyan-raev: No idea. It was not me who put it :)
If you want, I can remove it.

mniewrzal: I mean, do you think we have all the necessary params? :)

kaloyan-raev: I think they are enough for now. I will remove the TODO comment.

PR #1499, opened 2019-03-15
hammer: storage node monitor

mniewrzal:
PR is a little bit larger than it could be because I refactored inspector to avoid dependency cycles. I can split it into 2 PR if necessary.


storagenode/monitor/monitor.go
```diff
@@ -0,0 +1,134 @@
...
+// Config defines parameters for storage node disk and bandwidth usage monitoring.
+type Config struct {
+	Interval time.Duration `help:"how frequently Kademlia bucket should be refreshed with node stats" default:"1h0m0s"`
+}
+
+// Service which monitors disk usage and updates kademlia network as necessary.
+type Service struct {
+	log                *zap.Logger
+	store              *pieces.Store
+	inspector          *inspector.Endpoint
```
egonelbre: Services and endpoints shouldn't depend on other endpoints. This can eventually end up causing issues with authentication. Depend on the `pieces.DB` instead or separate inspector into a endpoint and a service

storagenode/monitor/monitor.go
```diff
@@ -0,0 +1,134 @@
...
+	rt                 *kademlia.RoutingTable
```
egonelbre: `routingTable`

storagenode/inspector/inspector.go
```diff
@@ -1,15 +1,17 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-package piecestore
+package inspector
```
egonelbre: moving this into separate package is a good idea

storagenode/peer.go
```diff
@@ -101,7 +103,8 @@ type Peer struct {
 		Trust     *trust.Pool
 		Store     *pieces.Store
 		Endpoint  *piecestore.Endpoint
-		Inspector *piecestore.Inspector
+		Inspector *inspector.Endpoint
+		Monitor   *monitor.Service
```
egonelbre: remove old monitor

storagenode/monitor/monitor.go
mniewrzal: ok

mniewrzal:
@aleitner any feedback welcome :)

PR #1501, opened 2019-03-15
 Add Repair method to ECClient 

kaloyan-raev:
Fixes: https://storjlabs.atlassian.net/browse/V3-1154

Adds new `Repair` method to ECClient, to be used by the `SegmentRepairer`.

The difference with the existing ECClient.Put method is that the Repair method:
- Does not fail if the provided order limits are less than the repair threshold.
- Does not fail if the repaired pieces are less than the repair threshold.
- Launches the timer at the beginning for a configurable timeout (default: 1 minute).
- Continues until the timer expires or the success threshold is reached.

### DIFF
--- a/pkg/storage/ec/client.go
+++ b/pkg/storage/ec/client.go
@@ -31,6 +31,7 @@ var mon = monkit.Package()
 // Client defines an interface for storing erasure coded data to piece store nodes
 type Client interface {
 	Put(ctx context.Context, limits []*pb.AddressedOrderLimit, rs eestream.RedundancyStrategy, data io.Reader, expiration time.Time) (successfulNodes []*pb.Node, successfulHashes []*pb.PieceHash, err error)
+	Repair(ctx context.Context, limits []*pb.AddressedOrderLimit, rs eestream.RedundancyStrategy, data io.Reader, expiration time.Time, timeout time.Duration) (successfulNodes []*pb.Node, successfulHashes []*pb.PieceHash, err error)
 	Get(ctx context.Context, limits []*pb.AddressedOrderLimit, es eestream.ErasureScheme, size int64) (ranger.Ranger, error)
 	Delete(ctx context.Context, limits []*pb.AddressedOrderLimit) error
 }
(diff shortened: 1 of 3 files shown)


pkg/storage/ec/client.go
```diff
@@ -168,6 +170,100 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, r
...
+	// due to errors instead of slowness.
+	if timer != nil {
+		timer.Stop()
+	}
+
+	// TODO: clean up the partially uploaded segment's pieces
+	defer func() {
+		select {
+		case <-ctx.Done():
+			err = utils.CombineErrors(
```
egonelbre: `errs.Combine`

kaloyan-raev: Fixed

pkg/storage/segments/repairer.go
```diff
@@ -5,6 +5,7 @@ package segments
 
 import (
 	"context"
+	time "time"
```
egonelbre: this definitely doesn't need an alias

pkg/storage/ec/client.go
```diff
@@ -168,6 +169,100 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, r
...
+	infos := make(chan info, len(limits))
+
+	psCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	for i, addressedLimit := range limits {
+		go func(i int, addressedLimit *pb.AddressedOrderLimit) {
+			hash, err := ec.putPiece(psCtx, ctx, addressedLimit, readers[i], expiration)
+			infos <- info{i: i, err: err, hash: hash}
+		}(i, addressedLimit)
```
mobyvb: I get an error with my repairer tests here when `addressedLimit` is nil. The `limits` arg passed in has nil values from `repairer.Repair` (unless I am doing something improperly)

pkg/storage/segments/repairer.go
kaloyan-raev: This is something done by autoformatting for some reason.

kaloyan-raev:
Closed in favor of #1509

PR #1503, opened 2019-03-15
audit service won't hang on bad storage node conns

navillasa:
- creates new audit service config value: minBytesPerSecond. The default value is 128B.
- minBytesPerSecond is used to calculate the context deadline which is now used in the audit service's `getShare` function.
- the getShare function should now timeout according to the proportional context deadline, however the audit service should continue normally (aka report the bad nodes to StatDB)
- nodes that take too long to send data (thus triggering the timeout) will now be marked as failing the audit
- added a TestGetShareTimeout


pkg/transport/slowtransport.go
```diff
@@ -0,0 +1,62 @@
...
+	"google.golang.org/grpc"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+)
+
+// SlowTransport is a slow version of transport
+type SlowTransport struct {
+	client  Client
+	latency time.Duration
```
egonelbre: This should be `dialLatency`.

pkg/transport/slowtransport.go
```diff
@@ -0,0 +1,62 @@
...
+// DialNode dials a node with latency
+func (slowTransport *SlowTransport) DialNode(ctx context.Context, node *pb.Node, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-time.After(slowTransport.latency):
+		break
+	}
+
+	return slowTransport.client.DialNode(ctx, node, opts...)
```
egonelbre: I think we also want to implement slow read / write, not just the dialing part.

pkg/audit/service.go
```diff
@@ -30,38 +33,42 @@ type Service struct {
 	Verifier *Verifier
 	Reporter reporter
 
-	ticker *time.Ticker
+	Loop sync2.Cycle
 }
 
 // NewService instantiates a Service with access to a Cursor and Verifier
-func NewService(log *zap.Logger, sdb statdb.DB, interval time.Duration, maxRetries int, pointers *pointerdb.Service, allocation *pointerdb.AllocationSigner, transport transport.Client, overlay *overlay.Cache, identity *identity.FullIdentity) (service *Service, err error) {
+func NewService(log *zap.Logger, sdb statdb.DB, interval time.Duration, maxRetries int,
```
egonelbre: Let's use the `Config` as argument here, there are too many arguments already.

pkg/audit/verifier.go
```diff
@@ -72,14 +78,26 @@ func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize
...
+		bandwidthMsgSize = shareSize + 8 - remainder
+		adjustedBandwidthMsgSize = bandwidthMsgSize / 8
 	}
 
-	ps, err := psclient.NewPSClient(ctx, d.transport, fromNode, bandwidthMsgSize)
+	seconds := bandwidthMsgSize / d.minBytesPerSecond.Int()
+
+	allottedTime := time.Now().Local().Add(time.Second * time.Duration(seconds))
+
+	newCtx, cancel := context.WithDeadline(ctx, allottedTime)
```
egonelbre: You can use `context.WithTimeout` instead of `context.WithDeadline`.

pkg/audit/verifier.go
```diff
@@ -72,14 +78,26 @@ func (d *defaultDownloader) getShare(ctx context.Context, stripeIndex, shareSize
...
+	defer func() {
```
egonelbre: defer cancel()

pkg/audit/timeout_test.go
```diff
@@ -0,0 +1,70 @@
...
+			stripe, err = cursor.NextStripe(ctx)
+			if stripe != nil || err != nil {
+				break
+			}
+		}
+		require.NoError(t, err)
+		require.NotNil(t, stripe)
+
+		overlay := planet.Satellites[0].Overlay.Service
+		tc := planet.Satellites[0].Transport
```
egonelbre: Let's also stop one of the storage nodes containing data in this test.

pkg/transport/slowtransport.go
navillasa: we'll do this in another PR

littleskunk:
Tests are looking good. I modified my storage node to sleep for an hour. With this PR the audit job is not hanging and on top of that I get a failed audit penalty. Nice :)

pkg/audit/newverifier_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+		err := planet.Satellites[0].Audit.Service.Close()
+		assert.NoError(t, err)
+
+		uplink := planet.Uplinks[0]
+		testData := make([]byte, 5*memory.MiB)
+		_, err = rand.Read(testData)
+		assert.NoError(t, err)
+
+		err = uplink.Upload(ctx, planet.Satellites[0], "test/bucket", "test/path", testData)
```
mobyvb: `testbucket` instead of `test/bucket`

pkg/audit/newverifier_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit_test
+
+import (
+	"crypto/rand"
```
mobyvb: `math/rand` instead of `crypto/rand` (faster for tests)

pkg/audit/newverifier_test.go
```diff
@@ -0,0 +1,66 @@
...
+			stripe, err = cursor.NextStripe(ctx)
+			if stripe != nil || err != nil {
+				break
+			}
+		}
+		require.NoError(t, err)
+		require.NotNil(t, stripe)
+
+		transport := planet.Satellites[0].Transport
+
```
mobyvb: nit - get rid of whitespace here, and below minBytesPerSecond for readability

pkg/audit/newverifier_test.go
```diff
@@ -0,0 +1,66 @@
...
+		minBytesPerSecond := 128 * memory.B
+
+		verifier := audit.NewVerifier(zap.L(), transport, overlay, planet.Satellites[0].Identity, minBytesPerSecond)
+		require.NotNil(t, verifier)
+
+		err = planet.StopPeer(planet.StorageNodes[0])
```
mobyvb: why are you stopping these nodes in particular? How do they relate to the stripe?

pkg/audit/timeout_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit_test
+
+import (
+	"crypto/rand"
```
mobyvb: math/rand

pkg/audit/timeout_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		// This config value will create a very short timeframe allowed for receiving
+		// data from storage nodes. This will cause context to cancel and start
+		// downloading from new nodes.
+		minBytesPerSecond := 110 * memory.KB
+
+		verifier := audit.NewVerifier(zap.L(), slowClient, overlay, planet.Satellites[0].Identity, minBytesPerSecond)
+		require.NotNil(t, verifier)
+
+		err = planet.StopPeer(planet.StorageNodes[0])
```
mobyvb: Why stop these nodes  in particular?

pkg/audit/timeout_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		err := planet.Satellites[0].Audit.Service.Close()
+		assert.NoError(t, err)
+
+		uplink := planet.Uplinks[0]
+		testData := make([]byte, 5*memory.MiB)
+		_, err = rand.Read(testData)
+		assert.NoError(t, err)
+
+		err = uplink.Upload(ctx, planet.Satellites[0], "test/bucket", "test/path", testData)
```
mobyvb: "test/bucket" -> "testbucket"

pkg/audit/verifier.go
```diff
@@ -46,25 +50,34 @@ type defaultDownloader struct {
...
 }
 
 // getShare use piece store client to download shares from nodes
 func (d *defaultDownloader) getShare(ctx context.Context, limit *pb.AddressedOrderLimit, stripeIndex int64, shareSize int32, pieceNum int) (share Share, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	bandwidthMsgSize := shareSize
+
+	// determines number of seconds allotted for receiving data from a storage node
+	seconds := time.Second * time.Duration(bandwidthMsgSize/d.minBytesPerSecond.Int32())
```
mobyvb: should we be doing integer division for `bandwidthMsgSize/d.minBytesPerSecond`? It might be more precise if we do floating point division

pkg/audit/newverifier_test.go
navillasa: @egonelbre just said to stop some

pkg/audit/timeout_test.go
navillasa: thought it would be good to stop some "randomly" but guess I could do it actually "randomly." do you have any recommendations?

pkg/audit/verifier.go
egonelbre: The way to avoid precision errors is to ensure that multiplication happens before division.

```
time.Duration(int(time.Second) * bandwidthMsgSize/d.minBytesPerSecond.Int()))
```

navillasa: so just convert them both to floats then divide?

navillasa: okay cool

PR #1508, opened 2019-03-18
Add serial number type

egonelbre:


pkg/storj/serialnumber.go
```diff
@@ -0,0 +1,105 @@
...
+// Value set a SerialNumber to a database field
+func (id SerialNumber) Value() (driver.Value, error) {
+	return id.Bytes(), nil
+}
+
+// Scan extracts a SerialNumber from a database field
+func (id *SerialNumber) Scan(src interface{}) (err error) {
+	b, ok := src.([]byte)
+	if !ok {
+		return ErrNodeID.New("SerialNumber Scan expects []byte")
```
kaloyan-raev: Wrong error class?

PR #1509, opened 2019-03-18
Add Repair method to ECClient

kaloyan-raev:
Fixes: https://storjlabs.atlassian.net/browse/V3-1154

Adds new `Repair` method to ECClient, to be used by the `SegmentRepairer`.

The difference with the existing ECClient.Put method is that the Repair method:
- Does not fail if the provided order limits are less than the repair threshold.
- Does not fail if the repaired pieces are less than the repair threshold.
- Launches the timer at the beginning for a configurable timeout (default: 1 minute).
- Continues until the timer expires or the success threshold is reached.


pkg/storage/ec/client.go
```diff
@@ -168,12 +174,111 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, r
...
+		timer.Stop()
+	}
+
+	// TODO: clean up the partially uploaded segment's pieces
+	defer func() {
+		select {
+		case <-ctx.Done():
+			err = errs.Combine(
+				Error.New("repair cancelled"),
+				// ec.Delete(context.Background(), nodes, pieceID, pba.SatelliteId), //TODO
```
mniewrzal: Is this comment still valid? Maybe you can also wrap this into one line.

kaloyan-raev: Done.

PR #1513, opened 2019-03-18
WIP, Experimental : SQL without mutex with SetMaxOpenConns()

wthorp:
This code is an experiment seeking feedback.  Although the JIRA ticket says "postgres / sqlite need retry logic for transactions", I wasn't clear what the goal was and if this was a legitimate issue.

https://storjlabs.atlassian.net/browse/V3-1016

While there may be remaining unknown / unqualified issues with improper transaction handling, there are concrete issues with our code...

1) The need for mutexes associated with SQLite
2) "too many clients already" error with Postgres tests using high levels of parallelism
3) Multiple calls to sql.Open() [primarily in tests], resulting in multiple connection pools in the same code

Some of these issues can be addressed by using sql.db.SetMaxOpenConns()

1) A mutex can be avoided by limiting SQLite access to a single thread per process, and adding a 30 second busy_timeout for multiprocessing.  SQLite's `cache=shared` supposedly fixes multi-threading issues, but does not fix our issues (possibly due to multiple connection pools / multiprocessing), so it currently should not be used.
2) The default Go Postgres MaxOpenConns must approach the PostgreSQL's backend maximum number of connections.  Reducing it prevents multiprocessing issues in our tests.
 3) While conceptually multiple connection pools remain an issue that should be fixed, they're mitigated by reducing the size of the pools.


pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -358,54 +368,38 @@ func (db *DB) GetBandwidthAllocations() (map[storj.NodeID][]*Agreement, error) {
...
-			return nil, err
-		}
 		agreements[satelliteID] = append(agreements[satelliteID], agreement)
 	}
 	return agreements, nil
 }
 
 // GetBwaStatusBySerialNum get BWA status by serial num
 func (db *DB) GetBwaStatusBySerialNum(serialnum string) (status AgreementStatus, err error) {
-	defer db.locked()()
```
egonelbre: It wouldn't been easier to make Lock and Unlock noop initially.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -325,28 +342,21 @@ func (db *DB) GetBandwidthAllocationBySignature(signature []byte) ([]*pb.Order,
 		}
 	}()
 
-	agreements := []*pb.Order{}
+	orders := []*pb.Order{}
```
egonelbre: No need to change these. They will be rewritten anyways.

internal/dbutil/sqliteutil/db.go
```diff
@@ -14,7 +14,8 @@ import (
 
 // LoadSchemaFromSQL inserts script into connstr and loads schema.
 func LoadSchemaFromSQL(script string) (_ *dbschema.Schema, err error) {
-	db, err := sql.Open("sqlite3", ":memory:")
+	db, err := sql.Open("sqlite3", "file::memory:?mode=memory&_journal=WAL&_busy_timeout=30000")
```
egonelbre: Let's make the constant public in `sqliteutil` and use it throughout the rest of the code. Or alternatively, let's create `sqliteutil.OpenInMemory()`.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -10,10 +10,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/lib/pq"
```
egonelbre: This only has sqlite backend, so importing `pq` doesn't seem reasonable.

pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -10,10 +10,11 @@ import (
...
+	"github.com/mattn/go-sqlite3"
 	_ "github.com/mattn/go-sqlite3" // register sqlite to sql
 	"github.com/zeebo/errs"
 	"go.uber.org/zap"
```
egonelbre: Most of the code in `psdb` will be replaced by code in `storagenode/storagenodedb`.

egonelbre:
This could actually work quite nicely for sqlite.

pkg/piecestore/psserver/psdb/psdb.go
wthorp: yeah, but its already done :/

mniewrzal:
@wthorp do you plan to continue this development?

wthorp:
Not planning on doing anything with this.

PR #1514, opened 2019-03-18
file health Api

aligeti:
https://storjlabs.atlassian.net/browse/V3-629
https://storjlabs.atlassian.net/browse/V3-611


pkg/metainfo/kvmetainfo/metainfo.go
```diff
@@ -30,17 +31,19 @@ type DB struct {
...
 	segments segments.Store
 	pointers pdbclient.Client
+	overlay  overlay.Client
 
 	rootKey *storj.Key
 }
 
 // New creates a new metainfo database
-func New(buckets buckets.Store, streams streams.Store, segments segments.Store, pointers pdbclient.Client, rootKey *storj.Key) *DB {
+func New(buckets buckets.Store, streams streams.Store, segments segments.Store, pointers pdbclient.Client, overlay overlay.Client, rootKey *storj.Key) *DB {
```
egonelbre: It shouldn't need overlay.

pkg/storj/object.go
```diff
@@ -91,6 +93,9 @@ type Segment struct {
 	// Encryption
 	EncryptedKeyNonce Nonce
 	EncryptedKey      EncryptedPrivateKey
+	// Pieces building info
+	Needed int32
+	Online int32
```
egonelbre: These aren't properties of the segment, this is derived information and shouldn't be part of this.

pkg/storj/object.go
```diff
@@ -59,6 +59,8 @@ type Stream struct {
 	// FixedSegmentSize is the size of each segment,
 	// when all segments have the same size. It is -1 otherwise.
 	FixedSegmentSize int64
+	// SegmentList is segments info
+	SegmentList []Segment
```
egonelbre: These aren't properties of the Stream, this is derived information and shouldn't be part of this.

pkg/metainfo/kvmetainfo/metainfo.go
aligeti: done

pkg/storj/object.go
aligeti: removed

pkg/storj/object.go
aligeti: done

pkg/storj/object.go
```diff
@@ -97,4 +97,5 @@ type Segment struct {
 type Piece struct {
 	Number   byte
 	Location NodeID
+	Online   bool
```
egonelbre: This does not make sense. `storj` package describes the data model how it's stored and communicated not current status of things. This PR should not change `storj` package.

pkg/metainfo/kvmetainfo/stream.go
```diff
@@ -62,11 +64,16 @@ func (stream *readonlyStream) segment(ctx context.Context, index int64) (segment
 		segment.Size = stream.info.FixedSegmentSize
 		copy(segment.EncryptedKeyNonce[:], segmentMeta.KeyNonce)
 		segment.EncryptedKey = segmentMeta.EncryptedKey
+		components := storj.SplitPath(segmentPath)
+		pointer, limits, _ = stream.db.metainfo.ReadSegment(ctx, components[1], components[2], index)
```
egonelbre: This will create a bunch of orders that will never get handled. `ReadSegment` is about intending to read.

pkg/storj/object.go
aligeti: removed the changes.

pkg/metainfo/kvmetainfo/stream.go
aligeti: removed changes

egonelbre:
This has gone on very long and still many problems:

1. we must try to avoid creating load on the satellite, we should do as little DB requests as possible
1. we also should avoid transferring more information than needed, we can always go back and ask for more
1. is this for internal debugging and diagnosis or public? if it's internal only then we shouldn't make it part of the public endpoint.

Let's implement this:

```
    rpc ObjectHealth(ObjectHealthRequest) returns (ObjectHealthResponse) {}

message ObjectHealthRequest {
    bytes bucket = 1;
    bytes encrypted_path = 2;
}

message ObjectHealthResponse {
    repeated SegmentHealth segments;
}

message SegmentHealth {
    int32 index            = 1;

    int32 minimum_required = 2;
    int32 repair_threshold = 3;
    int32 total            = 4;

    int32 online = 5;
}
```

cmd/uplink/cmd/stat.go
```diff
@@ -0,0 +1,73 @@
...
+		Short: "stat a Storj object",
+		RunE:  statObject,
+	}, RootCmd)
+}
+
+// copyMain is the function executed when cpCmd is called
+func statObject(cmd *cobra.Command, args []string) (err error) {
+	ctx := process.Ctx(cmd)
+
+	if len(args) == 0 {
```
mniewrzal: you can use `cobra.MinimumNArgs`
```
	pingNodeCmd = &cobra.Command{
		Use:   "ping <node_id> <ip:port>",
		Short: "ping node at provided ID",
		Args:  cobra.MinimumNArgs(2),
		RunE:  PingNode,
	}
```

pkg/metainfo/kvmetainfo/metainfo.go
```diff
@@ -30,17 +31,19 @@ type DB struct {
 	streams  streams.Store
 	segments segments.Store
 	pointers pdbclient.Client
+	metainfo metainfo.Client
```
mniewrzal: you need to merge with master. `metainfo.Client` is already here.

pkg/metainfo/kvmetainfo/stream.go
```diff
@@ -37,7 +37,7 @@ func (stream *readonlyStream) SegmentsAt(ctx context.Context, byteOffset int64,
 	return stream.Segments(ctx, index, limit)
 }
 
-func (stream *readonlyStream) segment(ctx context.Context, index int64) (segment storj.Segment, err error) {
+func (stream *readonlyStream) Segment(ctx context.Context, index int64) (segment storj.Segment, count int64, err error) {
```
mniewrzal: I'm not very familiar with kvmetainfo but I don't think its good idea to mix info about segment with retrieving storage node info with `StorageNodeInfo`. Maybe a separate method that will use `segment()` and later iterate over `segment.Pieces` (which contains NodeID).

pkg/storj/metainfo.go
```diff
@@ -189,6 +189,8 @@ type ReadOnlyStream interface {
 	// SegmentsAt returns the segment that contains the byteOffset and following segments.
 	// Limit specifies how much to return at most.
 	SegmentsAt(ctx context.Context, byteOffset int64, limit int64) (infos []Segment, more bool, err error)
+	// Segment returns a particular segment at index, and current the storage node addresses.
+	Segment(ctx context.Context, index int64) (segment Segment, count int64, err error)
```
mniewrzal: Name `count` gives no info what this is.

satellite/metainfo/metainfo.go
```diff
@@ -120,6 +120,31 @@ func (endpoint *Endpoint) SegmentInfo(ctx context.Context, req *pb.SegmentInfoRe
 	return &pb.SegmentInfoResponse{Pointer: pointer}, nil
 }
 
+// StorageNodeInfo returns address of storage nodeID
+func (endpoint *Endpoint) StorageNodeInfo(ctx context.Context, req *pb.StorageNodeInfoRequest) (resp *pb.StorageNodeInfoResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	nodes, err := endpoint.cache.GetAll(ctx, req.NodeId)
```
mniewrzal: Who will be client of this method (we or users)? if users, you need to add authentication.

mniewrzal:
@aligeti is this work will be continued or we can close this PR?

PR #1515, opened 2019-03-18
tlsopts: fix helper

jtolds:


pkg/peertls/tlsopts/options.go
```diff
@@ -118,10 +118,11 @@ func (vf *VerificationFuncs) ServerAdd(verificationFuncs ...peertls.PeerCertVeri
 }
 
 func removeNils(verificationFuncs []peertls.PeerCertVerificationFunc) []peertls.PeerCertVerificationFunc {
-	for i, f := range verificationFuncs {
-		if f == nil {
+	for i := 0; i < len(verificationFuncs); i++ {
```
egonelbre: writing these using append usually ends up as clearer code:
```
result := verificationFuncs[:0]
for _, f := range verificationFuncs {
    if f != nil {
        result = append(result, f)
    }
}
return result
```

jtolds: oh that's neat

PR #1517, opened 2019-03-18
add ability to specify redundancy scheme in testplanet upload

cam-a:
add new function,`UploadWithConfig`, for setting optional redundancy scheme values. We need this in order to have enough free nodes in the testplanet instance to test repair

### DIFF
--- a/internal/testplanet/planet.go
+++ b/internal/testplanet/planet.go
@@ -51,7 +51,6 @@ import (
 	"storj.io/storj/satellite/satellitedb"
 	"storj.io/storj/storagenode"
 	"storj.io/storj/storagenode/storagenodedb"
-	"storj.io/storj/uplink"
 )
 
 // Peer represents one of StorageNode or Satellite
@@ -70,7 +69,6 @@ type Config struct {
 	StorageNodeCount int
 	UplinkCount      int
 
-	RS          *uplink.RSConfig
 	Identities  *Identities
 	Reconfigure Reconfigure
 }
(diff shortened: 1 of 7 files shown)


internal/testplanet/planet.go
```diff
@@ -69,6 +70,7 @@ type Config struct {
 	StorageNodeCount int
 	UplinkCount      int
 
+	RS          *uplink.RSConfig
```
egonelbre: This doesn't make sense as a testplanet config. RS is a configuration about how you upload a single object / bucket files, and not a global thing. This probably makes more sense as an argument or modifier to uplink.

The rational is that tests should upload with different configurations.

egonelbre: Maybe make the argument as an option to upload?

egonelbre: One approach would be to not make it part of the config, but instead just allow overriding the uplink config itself and defaulting as usual.

internal/testplanet/uplink.go
```diff
@@ -166,8 +166,15 @@ func (uplink *Uplink) DialOverlay(destination Peer) (overlay.Client, error) {
 }
 
 // Upload data to specific satellite
-func (uplink *Uplink) Upload(ctx context.Context, satellite *satellite.Peer, bucket string, path storj.Path, data []byte) error {
+func (uplink *Uplink) Upload(ctx context.Context, satellite *satellite.Peer, redundancy *uplink.RSConfig, bucket string, path storj.Path, data []byte) error {
```
egonelbre: Don't modify the existing func, create a new one. This makes other tests harder to read.

internal/testplanet/uplink.go
```diff
@@ -166,8 +166,15 @@ func (uplink *Uplink) DialOverlay(destination Peer) (overlay.Client, error) {
...
 	config := uplink.getConfig(satellite)
+	if redundancy != nil {
+		config.RS.MinThreshold = redundancy.MinThreshold
```
egonelbre: This changes the existing behavior of the uplink, which is unexpected, given that it's an argument.

internal/testplanet/planet.go
cam-a: Do we want to pass in the entire uplink config, or just the uplink RSConfig?

egonelbre: The problem is that the config is "per bucket". At the moment I would pass in minimal, since once we flesh out more of the uplink api, we can start removing the test code from here.

PR #1518, opened 2019-03-18
adds pingbackTimeout to kademlia endpoint

navillasa:
- Adds a timeout value (time.Duration) to the Kademlia config. It's added to the Kademlia struct so that Endpoint can access it and Query won't hang.
- Passes config to kad.NewService as argument because adding `timeout` as another argument would make it way too long when it's already long.


pkg/kademlia/config.go
```diff
@@ -29,6 +29,8 @@ type Config struct {
 
 	// TODO: reduce the number of flags here
 	Alpha int `help:"alpha is a system wide concurrency parameter" default:"5"`
+	// TODO: make this a shorter amount of time
+	QueryTimeout int32 `help:"query timeout is the allotted time in seconds for a query to complete" default:"60"`
```
dylanlott: Nit: We could maybe even reduce this more aggressively, and if the user needs to tweak it they can. But honestly 60s is fine given how long we know these can take.

mniewrzal:
@navillasa tests are failing because `QueryTimeout` is always zero for some reason and kademlia requests end right away with `context deadline exceeded`.

pkg/kademlia/config.go
mniewrzal: I think this should be `time.Duration`. It will be more generic.

internal/testplanet/planet.go
```diff
@@ -403,6 +403,7 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 					Email:  prefix + "@example.com",
 					Wallet: "0x" + strings.Repeat("00", 20),
 				},
+				QueryTimeout: time.Duration(60) * time.Second, // TODO: why didn't it default to 60s when not set here?
```
egonelbre: `60 * time.Second`, no need for the conversion.

All params in testplanet must be explicitly specified because it ignores all of the config tags.

navillasa: ah, good to know.

pkg/kademlia/dialer.go
```diff
@@ -58,11 +62,16 @@ func (dialer *Dialer) Lookup(ctx context.Context, self pb.Node, ask pb.Node, fin
...
 	resp, err := conn.client.Query(ctx, &pb.QueryRequest{
-		Limit:    20, // TODO: should not be hardcoded, but instead kademlia k value, routing table depth, etc
-		Sender:   &self,
-		Target:   &find,
-		Pingback: true, // should only be true during bucket refreshing
+		Limit:        20, // TODO: should not be hardcoded, but instead kademlia k value, routing table depth, etc
+		Sender:       &self,
+		Target:       &find,
+		Pingback:     true, // should only be true during bucket refreshing
+		QueryTimeout: durationSeconds,
```
egonelbre: Why is this a protobuf request param?

pkg/kademlia/endpoint.go
```diff
@@ -35,8 +36,14 @@ func NewEndpoint(log *zap.Logger, service *Kademlia, routingTable *RoutingTable)
 
 // Query is a node to node communication query
 func (endpoint *Endpoint) Query(ctx context.Context, req *pb.QueryRequest) (*pb.QueryResponse, error) {
+
+	duration := time.Duration(req.QueryTimeout.Seconds) * time.Second
```
egonelbre: `QueryTimeout` should be specified for `Endpoint` not the request.

pkg/kademlia/dialer.go
navillasa: I was thinking it could be sent from the client (inspector) but I could remove it.

pkg/kademlia/kademlia.go
```diff
@@ -57,13 +57,14 @@ type Kademlia struct {
 }
 
 // NewService returns a newly configured Kademlia instance
-func NewService(log *zap.Logger, self pb.Node, bootstrapNodes []pb.Node, transport transport.Client, alpha int, rt *RoutingTable) (*Kademlia, error) {
+func NewService(log *zap.Logger, self pb.Node, transport transport.Client, rt *RoutingTable,
+	config Config) (*Kademlia, error) {
```
egonelbre: unwrap

pkg/kademlia/endpoint.go
navillasa: oh okay

pkg/kademlia/dialer.go
egonelbre: So the problem is that anyone can call this method with any value... however we want the actual server to decide the value. Not the remote untrusted caller.

navillasa: okay that makes sense

internal/testplanet/planet.go
```diff
@@ -403,6 +403,7 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 					Email:  prefix + "@example.com",
 					Wallet: "0x" + strings.Repeat("00", 20),
 				},
+				QueryTimeout: 60 * time.Second,
```
egonelbre: I think this should be `pingbackTimeout`

pkg/kademlia/endpoint.go
```diff
@@ -35,8 +36,14 @@ func NewEndpoint(log *zap.Logger, service *Kademlia, routingTable *RoutingTable)
 
 // Query is a node to node communication query
 func (endpoint *Endpoint) Query(ctx context.Context, req *pb.QueryRequest) (*pb.QueryResponse, error) {
+
+	duration := time.Duration(endpoint.service.queryTimeout.Seconds()) * time.Second
+
+	timedCtx, cancel := context.WithTimeout(ctx, duration)
```
egonelbre: Create the timedCtx inside the if statement, we don't need it otherwise, also the conversion above is not necessary you can directly use `context.WithTimeout(ctx, endpoint.pingbackTimeout)`

pkg/kademlia/kademlia.go
```diff
@@ -57,14 +58,15 @@ type Kademlia struct {
...
 	k := &Kademlia{
 		log:              log,
-		alpha:            alpha,
+		alpha:            config.Alpha,
 		routingTable:     rt,
-		bootstrapNodes:   bootstrapNodes,
+		bootstrapNodes:   config.BootstrapNodes(),
 		dialer:           NewDialer(log.Named("dialer"), transport),
 		refreshThreshold: int64(time.Minute),
+		queryTimeout:     config.QueryTimeout,
```
egonelbre: This timeout seems more about the endpoint and less about the service.

internal/testplanet/planet.go
navillasa: how much smaller? also, do you know why the overlay tests are failing now? it seems unrelated.

pkg/kademlia/endpoint.go
navillasa: ah okay thanks

pkg/kademlia/kademlia.go
navillasa: how should NewEndpoint access this value then besides via the kademlia config?

egonelbre: You can pass the timeout as an argument to `NewEndpoint` flag rather than kademlia service.

internal/testplanet/planet.go
egonelbre: not sure how much smaller... I guess we can leave it at 60s for now until we figure out a better number. Not sure about the tests.

pkg/kademlia/kademlia_test.go
```diff
@@ -528,5 +528,18 @@ func newKademlia(log *zap.Logger, nodeType pb.NodeType, bootstrapNodes []pb.Node
 	}
 	transportClient := transport.NewClient(tlsOptions, rt)
 
-	return NewService(log, self, bootstrapNodes, transportClient, alpha, rt)
+	queryTimeout := 60 * time.Second
+
+	kadConfig := Config{
```
egonelbre: no need to create a separate variable for config and query timeout, you can write them inline

internal/testplanet/planet.go
navillasa: last time I checked (prior to hammer merge) it could take up to 2 minutes to receive a node-info response back. so there will be node-info potentially missed if we make this smaller.

pkg/kademlia/kademlia.go
navillasa: unless I add an Endpoint config?

pkg/kademlia/dialer.go
```diff
@@ -19,9 +20,10 @@ import (
 
 // Dialer is a kademlia dialer
 type Dialer struct {
-	log       *zap.Logger
-	transport transport.Client
-	limit     sync2.Semaphore
+	log          *zap.Logger
+	transport    transport.Client
+	limit        sync2.Semaphore
+	queryTimeout time.Duration
```
egonelbre: this is isn't needed

navillasa: oops oversight, thanks

pkg/kademlia/kademlia.go
navillasa: but I'm not sure how I would let testplanet have access to an Endpoint config.. should the endpoint config be passed to NewService?

navillasa: never mind, I see! I think I'll just keep it as an argument.

PR #1520, opened 2019-03-19
Remove PointerDB client

kaloyan-raev:
- Removes completely the `pkg/pointerdb/pdbclient` package
- kvmetainfo is adapted with minimal changes to use the metainfo.Client instead.
- `OverlayAddr` and `PointerDBAddr` in `ClientConfig` are replaced with a single `SatelliteAddr`.


uplink/config.go
```diff
@@ -50,10 +49,7 @@ type EncryptionConfig struct {
 // ClientConfig is a configuration struct for the uplink that controls how
 // to talk to the rest of the network.
 type ClientConfig struct {
-	// TODO(jt): these should probably be the same
-	OverlayAddr   string `help:"Address to contact overlay server through"`
-	PointerDBAddr string `help:"Address to contact pointerdb server through"`
-
+	SatelliteAddr string      `help:"Address to contact the satellite"`
```
egonelbre: There are multiple places that reference `client.overlay-addr` and `client.pointer-db-addr`.

kaloyan-raev: They should become `client.satellite-addr`, right?

egonelbre: yes, unless they were setup from satellite addr

kaloyan-raev: Done.

cmd/gateway/main.go
```diff
@@ -110,9 +110,8 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	overrides := map[string]interface{}{
-		"client.api-key":         setupCfg.APIKey,
-		"client.pointer-db-addr": setupCfg.SatelliteAddr,
-		"client.overlay-addr":    setupCfg.SatelliteAddr,
+		"client.api-key":        setupCfg.APIKey,
+		"client.satellite-addr": setupCfg.SatelliteAddr,
```
egonelbre: Maybe, this can be removed, since it's already specified as `setupcfg`?

cmd/uplink/cmd/setup.go
```diff
@@ -56,9 +56,8 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	overrides := map[string]interface{}{
-		"client.api-key":         setupCfg.APIKey,
-		"client.pointer-db-addr": setupCfg.SatelliteAddr,
-		"client.overlay-addr":    setupCfg.SatelliteAddr,
+		"client.api-key":        setupCfg.APIKey,
+		"client.satellite-addr": setupCfg.SatelliteAddr,
```
egonelbre: Maybe this can be removed

cmd/gateway/main.go
kaloyan-raev: How can I access `setupcfg` in `uplink/config.go`?

egonelbre: Don't all the `setupCfg` flags get passed into `SaveConfigWithAllDefaults` automatically?

egonelbre: Or maybe, if we remove the "setup" tags from apikey and satellite-addr, in GatewayFlags, then they get automatically propagated?

kaloyan-raev: I deleted `ApiKey` and `SatelliteAddr` from `UplinkFlags` and `GatewayFlags`. So these two configs are now available only in the `ClientConfig`. I also added a `noprefix` tag to them. When `noprefix = true` then the configs are bound to flags without the prefix, i.e. as `--api-key` and `--satellite-addr` instead of `--client.api-key` and `--client.satellite-addr`. See my latest commit.

PR #1521, opened 2019-03-19
remove overlay endpoint

egonelbre:
This PR removes overlay endpoint which ends up simplifying a lot of code.

Some responsibilities were moved around, so:
1. overlay cache is responsible for selecting storage nodes, it might make sense to split it into a separate service, since the logic for it will definitely get more complex, but it should suffice for now.
2. overlay cache has now a method for figuring out offline nodes


egonelbre:
Waiting for https://github.com/storj/storj/pull/1520, to reduce duplicated work.

pkg/datarepair/checker/checker.go
```diff
@@ -217,24 +206,26 @@ func (checker *Checker) invalidNodes(ctx context.Context, nodeIDs storj.NodeIDLi
...
 // combine the offline nodes with nodes marked invalid by statdb
-func combineOfflineWithInvalid(offlineNodes []int32, invalidNodes []int32) (missingPieces []int32) {
-	missingPieces = append(missingPieces, offlineNodes...)
+func combineOfflineWithInvalid(offlineNodes []int, invalidNodes []int) (missingPieces []int32) {
+	for _, offline := range offlineNodes {
+		missingPieces = append(missingPieces, int32(offline))
+	}
 
-	offlineMap := make(map[int32]bool)
+	offlineMap := make(map[int]bool)
```
navillasa: why does the int type change depending on if they're in the offlineMap or missingPieces array?

pkg/overlay/cache.go
```diff
@@ -57,15 +57,56 @@ type DB interface {
...
+type NewNodeCriteria struct {
+	FreeBandwidth int64
+	FreeDisk      int64
+
+	AuditThreshold int64
+
+	Excluded []storj.NodeID
+}
+
 // Cache is used to store overlay data in Redis
```
navillasa: is this still Redis specific?

pkg/overlay/cache.go
```diff
@@ -78,42 +119,63 @@ func (cache *Cache) Inspect(ctx context.Context) (storage.Keys, error) {
 }
 
 // List returns a list of nodes from the cache DB
-func (cache *Cache) List(ctx context.Context, cursor storj.NodeID, limit int) ([]*pb.Node, error) {
+func (cache *Cache) List(ctx context.Context, cursor storj.NodeID, limit int) (_ []*pb.Node, err error) {
```
navillasa: why the underscored return values?

pkg/overlay/cache.go
```diff
@@ -200,27 +266,35 @@ func (cache *Cache) Put(ctx context.Context, nodeID storj.NodeID, value pb.Node)
 
 // Delete will remove the node from the cache. Used when a node hard disconnects or fails
 // to pass a PING multiple times.
-func (cache *Cache) Delete(ctx context.Context, id storj.NodeID) error {
+func (cache *Cache) Delete(ctx context.Context, id storj.NodeID) (err error) {
 	if id.IsZero() {
 		return ErrEmptyNode
 	}
+	defer mon.Task()(&ctx)(&err)
```
navillasa: shouldn't this happen before the if statement?

pkg/overlay/cache.go
egonelbre: because there's no need to give them names :)

pkg/overlay/cache.go
egonelbre: the reason I'm excluding the obvious error case is because it probably isn't very informative in the output... but I can also move it upwards

pkg/overlay/cache.go
navillasa: ok lol

pkg/datarepair/checker/checker.go
egonelbre: internally it's easier to use `int`-s, but since the protobuf requires specific sizes, it needs to use int32 as the result

pkg/overlay/cache.go
navillasa: it just seems odd to exclude it but 🤷🏻‍♀️doesn't seem like a big deal

pkg/overlay/cache.go
egonelbre: fixed

pkg/overlay/cache.go
```diff
@@ -170,7 +234,9 @@ func (cache *Cache) GetAll(ctx context.Context, ids storj.NodeIDList) ([]*pb.Nod
 }
 
 // Put adds a nodeID to the redis cache with a binary representation of proto defined Node
```
navillasa: nit - another redis reference

pkg/overlay/cache.go
egonelbre: ok, made consistent.

pkg/overlay/cache.go
egonelbre: fixed

PR #1522, opened 2019-03-19
Move psserver tests

mniewrzal:
Old psserver tests are moved and adapted to new piecestore. Also, few minor issues fixed found during wiring up tests.


pkg/storage/ec/client_planet_test.go
```diff
@@ -153,7 +153,10 @@ func newAddressedOrderLimit(action pb.PieceAction, satellite *satellite.Peer, up
 	if err != nil {
 		return nil, err
 	}
-
+	orderExpiration, err := ptypes.TimestampProto(time.Now().Add(24 * time.Hour))
```
egonelbre: I suspect we want a larger expiration time than 1day?

mniewrzal: this is for test only

egonelbre: ah, right :)

pkg/pointerdb/allocation.go
```diff
@@ -103,7 +103,7 @@ func (allocation *AllocationSigner) OrderLimit(ctx context.Context, parameters O
 	}
 
 	// convert bwExpiration from days to seconds
-	orderExpiration, err := ptypes.TimestampProto(time.Unix(int64(allocation.bwExpiration*86400), 0))
+	orderExpiration, err := ptypes.TimestampProto(time.Now().Add(time.Duration(allocation.bwExpiration*24) * time.Hour))
```
egonelbre: `bwExpiration` probably should be `time.Duration` in the first place, but we can fix that later.

storagenode/piecestore/endpoint_test.go
```diff
@@ -33,7 +39,7 @@ func TestUploadAndPartialDownload(t *testing.T) {
 	_, err = rand.Read(expectedData)
 	require.NoError(t, err)
 
-	err = planet.Uplinks[0].Upload(ctx, planet.Satellites[0], "test/bucket", "test/path", expectedData)
+	err = planet.Uplinks[0].Upload(ctx, planet.Satellites[0], "testBucket", "test/path", expectedData)
```
egonelbre: use lowercase, so it matches kaloyan-s PR

storagenode/piecestore/endpoint_test.go
```diff
@@ -85,3 +91,303 @@ func TestUploadAndPartialDownload(t *testing.T) {
...
+			err:     "missing piece id",
+		},
+		{ // should successfully store data
+			pieceID: teststorj.PieceIDFromString("99999999999999999999"),
+			content: []byte("xyzwq"),
+			action:  pb.PieceAction_GET,
+			err:     "expected put or put repair action got GET",
+		},
+	} {
+		sum := sha256.Sum256(tt.content)
```
egonelbre: use `pkcrypto.SHA256Hash`

storagenode/piecestore/endpoint_test.go
```diff
@@ -85,3 +91,303 @@ func TestUploadAndPartialDownload(t *testing.T) {
...
+
+	var serialNumber storj.SerialNumber
+	_, _ = rand.Read(serialNumber[:])
+
+	orderLimit := GenerateOrderLimit(
+		t,
+		planet.Satellites[0].ID(),
+		planet.Uplinks[0].ID(),
+		planet.StorageNodes[0].ID(),
+		teststorj.PieceIDFromString("existing-piece-id"),
```
egonelbre: You can use `storj.PieceID{1,2,3,4,5}`... the `teststorj.PieceIDFromString`, isn't that great.

storagenode/piecestore/endpoint_test.go
```diff
@@ -85,3 +91,303 @@ func TestUploadAndPartialDownload(t *testing.T) {
...
+		100,
+	)
+	signer := signing.SignerFromFullIdentity(planet.Satellites[0].Identity)
+	orderLimit, err = signing.SignOrderLimit(signer, orderLimit)
+	require.NoError(t, err)
+
+	uploader, err := client.Upload(ctx, orderLimit)
+	require.NoError(t, err)
+
+	expectedData := []byte("for-download-test")
```
egonelbre: Try to use completely random data. It's more likely to break if we get something wrong with the data encoding and sending.

storagenode/piecestore/verification_test.go
```diff
@@ -0,0 +1,130 @@
...
+		pieceID         storj.PieceID
+		action          pb.PieceAction
+		serialNumber    storj.SerialNumber
+		pieceExpiration time.Duration
+		orderExpiration time.Duration
+		limit           int64
+		err             string
+	}{
+		{ // unapproved satellite id
+			satelliteID:     testplanet.MustPregeneratedIdentity(0),
```
egonelbre: Use `storj.NodeID{100}` or similar. Since all of them end up using the same pregenerated identities, accidentally it might still end up in the approved list.

storagenode/piecestore/verification_test.go
```diff
@@ -0,0 +1,130 @@
...
+			pieceID:         teststorj.PieceIDFromString("piece-id-1"),
```
egonelbre: `storj.PieceID{1}`

storagenode/piecestore/verification_test.go
```diff
@@ -0,0 +1,130 @@
...
+			action:          pb.PieceAction_PUT,
+			serialNumber:    storj.SerialNumber([16]byte{1}),
```
egonelbre: `storj.SerialNumber{1}`

mniewrzal: done

storagenode/piecestore/verification_test.go
mniewrzal: done

storagenode/piecestore/verification_test.go
mniewrzal: you mean use pregenerated identity with index 100 or create custom identity?

storagenode/piecestore/endpoint_test.go
mniewrzal: done

storagenode/piecestore/endpoint_test.go
mniewrzal: done

storagenode/piecestore/endpoint_test.go
mniewrzal: done

storagenode/piecestore/endpoint_test.go
mniewrzal: done

pkg/pointerdb/allocation.go
mniewrzal: let fix this later

storagenode/piecestore/verification_test.go
egonelbre: Ah, it's the full identity not the ID, make the field name `satellite`, to be clearer. But yeah, in that case you can use `planet.NewIdentity()`, or if there are multiple places where you want one, then you could make `planet.MustNewIdentity()`.

storagenode/pieces/store.go
```diff
@@ -32,7 +32,7 @@ type Info struct {
 
 	PieceID         storj.PieceID
 	PieceSize       int64
-	PieceExpiration time.Time
+	PieceExpiration *time.Time
```
kaloyan-raev: What's the benefit of changing this to a pointer? I don't see it checked on any place in this PR.

mniewrzal: First of all `PieceExpiration` is not saved to DB this way, on storage node we are receiving nil timestamp so I tough that converting this to empty `time.Time` makes no sense.

kaloyan-raev: OK.

PR #1525, opened 2019-03-19
Fix CalcEncryptedSize and CalcPieceSize

kaloyan-raev:
When encrypting and erasure coding data, the data is padded to be divisible by the encoded block size or stripe size, respectively. This padding adds 4 bytes at the end of the padded data with the number of
bytes that have been added (the padding bytes).

These 4 bytes were not taken into account byte the CalcEncryptedSize and CalcPieceSize function, so they produced the wrong result in some corner cases.


pkg/encryption/encryption.go
```diff
@@ -10,8 +10,12 @@ import (
 	"storj.io/storj/pkg/storj"
 )
 
-// AESGCMNonceSize is the size of an AES-GCM nonce
-const AESGCMNonceSize = 12
+const (
+	// AESGCMNonceSize is the size of an AES-GCM nonce
+	AESGCMNonceSize = 12
+
+	uint32Size = 4
```
mobyvb: maybe add a comment here explaining the same thing as in your PR description?

pkg/eestream/encode.go
```diff
@@ -267,8 +267,8 @@ func (er *EncodedRanger) Range(ctx context.Context, offset, length int64) ([]io.
 // CalcPieceSize calculates what would be the piece size of the encoded data
 // after erasure coding data with dataSize using the given ErasureScheme.
 func CalcPieceSize(dataSize int64, scheme ErasureScheme) int64 {
```
egonelbre: Let's add tests for this as well.

pkg/encryption/encryption.go
kaloyan-raev: Added a short comment

pkg/eestream/encode.go
kaloyan-raev: Done.

pkg/eestream/rs_test.go
```diff
@@ -6,6 +6,7 @@ package eestream
 import (
 	"bytes"
 	"context"
+	cryptorand "crypto/rand"
```
egonelbre: in tests you can use `math/rand`.

`rand.New(rand.NewSource(rand.Int63()))` <-- to get a reader from `math/rand`. Not sure whether there's an easier way.

pkg/eestream/rs_test.go
```diff
@@ -640,3 +644,37 @@ func BenchmarkReedSolomonErasureScheme(b *testing.B) {
...
+
+		fc, err := infectious.NewFEC(2, 4)
+		require.NoError(t, err, errTag)
+		es := NewRSScheme(fc, 1*memory.KiB.Int())
+		rs, err := NewRedundancyStrategy(es, 0, 0)
+		require.NoError(t, err, errTag)
+
+		calculatedSize := CalcPieceSize(int64(dataSize), es)
+
+		readers, err := EncodeReader(ctx, PadReader(ioutil.NopCloser(io.LimitReader(cryptorand.Reader, dataSize)), es.StripeSize()), rs)
```
egonelbre: Do we care that we have a random data being encrypted? We could just have `InfiniteReader{}` that's used here.

kaloyan-raev: Which package is `InfiniteReader` in?

egonelbre:
```
type InfiniteReader struct {}
func (*InfiniteReader) Read(data []byte) (int, error) { return len(data), nil }
```

kaloyan-raev: Oh. You meant I should define one...
Nevertheless, I think it is better with the random data instead of all-zeros byte slice, or any other pattern.

pkg/eestream/rs_test.go
kaloyan-raev: Good to know how to create a random reader from `math/rand`. Thanks!

PR #1527, opened 2019-03-19
Piecestore space and bandwidth allocation check

mniewrzal:
Piecestore allocation check for disk space and bandwidth. Free space/bandwidth is compared during each read/write operation.
* disk and bandwidth check for upload
* bandwidth check for download


aleitner:
oh whoops it's WIP

storagenode/piecestore/endpoint.go
```diff
@@ -61,10 +63,11 @@ type Endpoint struct {
 }
 
 // NewEndpoint creates a new piecestore endpoint.
-func NewEndpoint(log *zap.Logger, signer signing.Signer, trust *trust.Pool, store *pieces.Store, pieceinfo pieces.DB, orders orders.DB, usage bandwidth.DB, usedSerials UsedSerials, config Config) (*Endpoint, error) {
+func NewEndpoint(log *zap.Logger, signer signing.Signer, trust *trust.Pool, store *pieces.Store, pieceinfo pieces.DB, orders orders.DB, usage bandwidth.DB, usedSerials UsedSerials, config Config, oldConfig psserver.Config) (*Endpoint, error) {
```
egonelbre: Maybe pass in `monitor.Service` into the endpoint and use the `monitor.Service` to handle the space and bandwidth logic?

mniewrzal: yeah, now when I'm checking space and bandwidth for upload it makes more sense

storagenode/piecestore/verification.go
```diff
@@ -149,3 +149,27 @@ func (endpoint *Endpoint) IsExpired(expiration *timestamp.Timestamp) bool {
...
 	return expirationTime.Before(time.Now().Add(-endpoint.config.ExpirationGracePeriod))
 }
+
+// VerifyAvailableSpace verifies whether there is available disk space
+func (endpoint *Endpoint) VerifyAvailableSpace(ctx context.Context, limit *pb.OrderLimit2) error {
+	availableSpace, err := endpoint.monitor.AvailableSpace(ctx)
+	if err != nil {
+		return ErrInternal.Wrap(err)
+	}
+	if availableSpace < limit.Limit {
```
aleitner: If the order limit is significantly higher than the actual amount being downloaded/uploaded then the storage node could be penalized for rejecting even though it might have sufficient bandwidth or storage space

aleitner: previously we would verify available bandwidth/disk on every read/write. Why are we not doing that anymore?

egonelbre: there's a todo note somewhere about it, but yeah, we should verfiy and update it on each action

aleitner: this seems like the pr to put it back

mniewrzal: so be it

egonelbre: The logic should roughly be:

```
endpoint.monitor.AddBandwith(ctx, limit.Amount)
defer endpoint.monitor.AddBandwith(ctx, order.Amount - limit.Amount)
```

mniewrzal: I'm not sure if im following, we need to check space/bandwidth on each read/write. As @aleitner  wrote we cannot use OrderLimit.Limit because it can be much higher than real data size for download/upload.

egonelbre: requests usually come in with limits with the maximum they intend to read, which is either "size of the piece" or "a specific size in case of audits"... generally the difference shouldn't be that high... Of course we can do it incrementally as well.

storagenode/monitor/monitor.go
```diff
@@ -88,23 +88,24 @@ func (service *Service) Run(ctx context.Context) (err error) {
 
 	// check your hard drive is big enough
 	// first time setup as a piece node server
+	allocatedDiskSpace := service.allocatedDiskSpace
 	if totalUsed == 0 && freeDiskSpace < service.allocatedDiskSpace {
-		service.allocatedDiskSpace = freeDiskSpace
-		service.log.Warn("Disk space is less than requested. Allocating space", zap.Int64("bytes", service.allocatedDiskSpace))
+		allocatedDiskSpace = freeDiskSpace
```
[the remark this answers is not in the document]
mniewrzal: you are right, I'm not sure why I change this (it was month ago...). will revert it

internal/testplanet/planet.go
```diff
@@ -594,7 +594,7 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatelliteIDs []strin
 			},
 			Storage: psserver.Config{
 				Path:                   "", // TODO: this argument won't be needed with master storagenodedb
-				AllocatedDiskSpace:     memory.TB,
+				AllocatedDiskSpace:     1500 * memory.GB,
```
egonelbre: `GiB`

PR #1530, opened 2019-03-20
Test that network stalls don't cause indefinite hangs in uplink

thepaul:
See V3-1164.

Right now, this just tests whether an uplink correctly times out after some amount of time when one of the nodes it's talking to suddenly goes offline.

This tool is meant to be run under `storj-sim network test`.

I am _very_ open to suggestions as to the following:

* where should this live? I've put it in `scripts/` right now, but I'm not sure it fits there, as I don't know exactly what we intended `scripts/` to be. Tools? Test scripts? Both?
* should this actually be written to use the golang 'testing' module? It kind of seems like it should, because it is a test after all, but I don't know how that would work.
* should we add this to `scripts/test-sim.sh`?


cmd/storj-sim/process.go
```diff
@@ -81,6 +82,7 @@ type Info struct {
 	Directory  string
 	ID         string
 	Extra      []string
+	Pid        int
```
egonelbre: move this before extra

thepaul: Ok

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package main
```
egonelbre: add a high-level description what this test does

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
...
+	"syscall"
+	"time"
+
+	"github.com/zeebo/errs"
+)
+
+var (
+	numTries      = flag.Int("num-tries", 20, "number of tries to cause a hang")
+	bucketName    = flag.String("bucket", "bukkit", "name of bucket to use for test")
+	fileSize      = flag.Int64("file-size", 5*1024*1024, "size of test file to use")
```
egonelbre: you can use memory.Size here

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
...
+	deleteTimeout = flag.Duration("timeout", 60*time.Second, "how long to wait for a delete to succeed or time out")
+
+	tryAgain = errs.New("test needs to run again")
+)
+
+func makeRandomContentsFile(path string, size int64) (err error) {
+	inFile, err := os.Open("/dev/urandom")
```
egonelbre: use `math/rand`

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
...
+func (ur *uplinkRunner) run(ctx context.Context, stdout, stderr io.Writer, args ...string) error {
+	cmdArgs := []string{"--config-dir", ur.configDir, "--log.level", ur.logLevel}
+	cmdArgs = append(cmdArgs, args...)
+	cmd := exec.CommandContext(ctx, ur.execName, cmdArgs...)
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+	return cmd.Run()
+}
+
+func (ur *uplinkRunner) GetOutputCtx(ctx context.Context, args ...string) ([]byte, error) {
```
egonelbre: I think it should be sufficient to have one func called `Run(ctx context.Context, args ...string)`... and ignore the output where you don't need it. Having 4 things that roughly do the same thing is slightly confusing.

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (ur *uplinkRunner) Call(args ...string) error {
+	return ur.CallCtx(context.Background(), args...)
+}
+
+// skip the first four whitespace-delimited fields and keep the rest
+var lsOutputRegexp = regexp.MustCompile(`(?m)^\s*(?:\S+\s+){4}(.*)$`)
+
+func (ur *uplinkRunner) doesRemoteExist(remotePath string) (bool, error) {
+	pathParts := strings.Split(remotePath, "/")
```
egonelbre: `path.Split`

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
...
+	if len(pathParts) < 2 {
+		return false, errs.New("invalid remote path %q", remotePath)
+	}
+	bucketAndDir := strings.Join(pathParts[:len(pathParts)-1], "/")
```
egonelbre: `path.Join`

thepaul: path.Join calls path.Clean(), which will wipe out the `sj://` part of the string, won't it?

egonelbre: ah, right

scripts/test-network-stalls.go
thepaul: that would work fine here, but I'm hesitant to use a URL-type string with path.Split, just because of potential confusion (you couldn't correctly use path.Join to reassemble the parts).

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,248 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
```
egonelbre: add `// +build ignore`, because there are multiple scripts in the same folder in the scripts dir

egonelbre:
This also could be written as a proper go test. `testcontext` has a method `Compile` which allows to compile a single binary, however it would need tweaking to ensure you compile all of the binaries in the specified folder with the correct name, otherwise storj-sim would have trouble finding them. 

And yes `scripts` is currently really confused in what it wants to be when it grows up.

scripts/test-network-stalls.go
thepaul: I had thought this was the norm in Go (see strings.Split..., bytes.Trim..., regexp.Regexp's Find... methods, exec.Command/exec.CommandContext, and most of strconv)

But ok, sure, this isn't a general-purpose library :smile:

scripts/test-network-stalls.go
thepaul: oh neat

scripts/test-network-stalls.go
egonelbre: The reason for `Command` and `CommandContext` is because of backwards compatibility. `context` was added after `exec` package was designed.

scripts/test-network-stalls.go
thepaul: The thing that makes life hard with `// +build ignore` is that it makes VSCode ignore it too, and I don't get any of the IDE niceness when working on it. Do you know a workaround, or is that just something one has to deal with in Go land?

egonelbre: Yeah, I know :'(... probably the solution would be to have the scripts somehow more nicely grouped and separate folders. I usually remove the `+build ignore` while working on the file, and later add it back.

scripts/test-network-stalls.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/internal/memory"
+)
+
+var (
+	numTries      = flag.Int("num-tries", 20, "number of tries to cause a hang")
+	bucketName    = flag.String("bucket", "bukkit", "name of bucket to use for test")
+	fileSize      = flag.Int64("file-size", int64(5*memory.MiB), "size of test file to use")
```
egonelbre: ah, what I meant was to use something like:

```
var filesize memory.Size = 5*memory.MiB
func init() {
    flag.Var(&filesize "file-size", "size of test file to use")
}
```

I know it's slightly annoying to write that way and probably doesn't matter that much here. But this allows to specify the command line flag as `command -file-size 5MB`

thepaul: oh, excellent. i didn't realize that was possible. i like it

scripts/test-sim.sh
```diff
@@ -19,14 +19,14 @@ export STORJ_NETWORK_DIR=$TMP
...
-storj-sim -x network test bash $SCRIPTDIR/test-uplink.sh
+storj-sim -x network test bash "$SCRIPTDIR"/test-sim-aws.sh
+storj-sim -x network test bash "$SCRIPTDIR"/test-uplink.sh
 storj-sim -x network destroy
 
 # setup the network with ipv6
 storj-sim -x --host "::1" network setup
 # aws-cli doesn't support gateway with ipv6 address, so change it to use localhost
-find $STORJ_NETWORK_DIR/gateway -type f -name config.yaml -exec sed -i "s/server.address: \"\[::1\]/server.address: \"127.0.0.1/" {} +
+find "$STORJ_NETWORK_DIR"/gateway -type f -name config.yaml -exec sed -i 's/server.address: "[::1]/server.address: "127.0.0.1/' '{}' +
```
egonelbre: I think this change is the reason why tests are failing at the moment.

thepaul: Oh dang

scripts/test-aio.sh
```diff
@@ -44,13 +44,18 @@ echo "=> Secret Key: $secret_access_key"
...
-head -c 1024 </dev/urandom > "$TMP_DIR/small-upload-testfile" # create 1mb file of random bytes (inline)
-head -c 5120 </dev/urandom > "$TMP_DIR/big-upload-testfile"   # create 5mb file of random bytes (remote)
-head -c 5 </dev/urandom > "$TMP_DIR/multipart-upload-testfile"     # create 5kb file of random bytes (remote)
+random_bytes_file(){
+	size=$1
+	output=$2
+	dd if=/dev/urandom of="$output" count=1 bs="$size" >/dev/null 2>&1
+}
+
+random_bytes_file 1x1024x1024 "$TMP_DIR/small-upload-testfile"     # create 1mb file of random bytes (inline)
```
dylanlott: _nice_

PR #1532, opened 2019-03-20
Web auth pages redesign

NikolaiYurchenko:


web/satellite/src/views/Register.vue
```diff
@@ -105,81 +169,84 @@ import { createUserRequest } from '@/api/users';
...
                     this.$data.firstNameError = 'Invalid First Name';
-                    hasError = true;
+                    return false;
                 }
 
-                if (!validateEmail(email)) {
+                if (!validateEmail(this.$data.email.trim())) {
                     this.$data.emailError = 'Invalid Email';
-                    hasError = true;
+                    return false;
```
crawter: run linter

web/satellite/src/views/Register.vue
```diff
@@ -105,81 +169,84 @@ import { createUserRequest } from '@/api/users';
...
-            Button,
+            HeaderlessInput,
             RegistrationSuccessPopup
+        },
+        mounted(): void {
+            let token: any = this.$route.query.token ? this.$route.query.token.toString() : null;
+
+            if(token) {
+                this.$data.secret = token;
+            }
```
crawter: if(this.$route.query.token) {
                this.$data.secret = this.$route.query.token.toString();
            }

web/satellite/tests/unit/store/users.spec.ts
```diff
@@ -19,7 +19,8 @@ describe('mutations', () => {
             user: {
                 firstName: '',
                 lastName: '',
-                email: ''
+                email: '',
+                secret: '',
```
crawter: secret is not a part of a user

PR #1535, opened 2019-03-20
storage node order sending

egonelbre:
Implements storage node order sending. This is roughly the same implementation as agreement sender. It will be wired up once we have implemented it on satellite.

This doesn't contain the satellite side. It will be a separate PR.


storagenode/orders/db_test.go
```diff
@@ -78,8 +83,44 @@ func TestOrders(t *testing.T) {
 
 		unsent, err := ordersdb.ListUnsent(ctx, 100)
 		require.NoError(t, err)
-
 		require.Empty(t, cmp.Diff([]*orders.Info{info}, unsent, cmp.Comparer(pb.Equal)))
+
+		// list by group
+		unsentGrouped, err := ordersdb.ListUnsentBySatellite(ctx)
+		require.NoError(t, err)
+		require.Empty(t, cmp.Diff(map[storj.NodeID][]*orders.Info{
```
mniewrzal: it would be nice to format this or separate into different var because its super hard to read this

storagenode/storagenodedb/orders.go
```diff
@@ -113,3 +115,148 @@ func (db *ordersdb) ListUnsent(ctx context.Context, limit int) (_ []*orders.Info
...
+	}
+
+	return infos, ErrInfo.Wrap(rows.Err())
+}
+
+// Archive marks order as being handled.
+func (db *ordersdb) Archive(ctx context.Context, satellite storj.NodeID, serial storj.SerialNumber, status orders.Status) error {
+	defer db.locked()()
+
+	result, err := db.db.Exec(`
```
mniewrzal: maybe it's not a very important case here but shouldn't we start using prepared statements for often executed statements?

egonelbre: Maybe, let's make an issue about testing the perf difference and see whether it's worthwhile.

egonelbre: created an issue

mniewrzal: thanks

PR #1536, opened 2019-03-20
code review checklist

jtolds:
https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository


.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+
+
+
+## Code Review Checklist
```
egonelbre: Make this an actual check list. github supports:

- [ ] Alpha
- [ ] Beta
- [ ] Gamma

egonelbre: syntax `- [ ] text`

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
...
+ * Does the PR describe what changes are being made?
+ * Does the code follow the [Effective Go style guide](https://golang.org/doc/effective_go.html)?
```
egonelbre: link to our guide?

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
...
+ * Does the new code have enough documentation? (comments and or diagrams)
+ * Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
```
egonelbre: maybe a check something like:
```
* Would two similar additional PR-s make the code hard to maintain? (In case it modifies existing code)
```

.github/pull_request_template.md
egonelbre: unfortunately doesn't support two checkboxes side-by-side

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+
+
+
+## Code Review Checklist
+ * Does the PR describe what changes are being made?
+ * Does the code follow the [Effective Go style guide](https://golang.org/doc/effective_go.html)?
+ * Does the new code have enough documentation? (comments and or diagrams)
```
egonelbre: Also "Does any of the documentation need updating?"

.github/pull_request_template.md
navillasa: which guide?

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+
+
+
+## Code Review Checklist
+ * Does the PR describe what changes are being made?
```
JessicaGreben: I think this should say: "Does the PR describe what changes are being made and why?".  I think its really important we all understand **why** we are doing what we do.

.github/pull_request_template.md
JessicaGreben: yeah i was wondering the same, there is this storj style guide https://github.com/storj/docs/blob/master/code/Style.md and also for testing https://github.com/storj/docs/blob/master/code/Testing.md

.github/pull_request_template.md
jtolds: oh nice

.github/pull_request_template.md
jtolds: what do you mean?

.github/pull_request_template.md
jtolds: oh good call

.github/pull_request_template.md
jtolds: good call

.github/pull_request_template.md
JessicaGreben: "enough documentation" is a little vague.  I'm personally curious, where do we expect people to be adding documentation? Seems like there are different places to do so.  For example there are two different wikis: https://wiki.storj.io/ and https://github.com/storj/storj/wiki (which apparently is sourced from the storj/docs repo).

.github/pull_request_template.md
jtolds: i added "Can this PR be broken into smaller PRs?" which loses some fidelity from what i think you mean but should lead to the same outcome? or i misunderstood completely

.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
+
+
+## Code Review Checklist
+ - [ ] Does the PR describe what changes are being made?
+ - [ ] Does the PR describe why the changes are being made?
+ - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
+ - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
+ - [ ] Could the PR be broken into smaller PRs?
+ - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
+ - [ ] Does the new code have enough documentation? (comments and or diagrams)
```
JessicaGreben: "enough documentation" seems a little vague. I'm personally curious, where do we expect people to be adding documentation? Seems like there are different places to do so. For example there are two different wikis: https://wiki.storj.io/ and https://github.com/storj/storj/wiki (which apparently is sourced from the storj/docs repo).

jtolds: good question! i think in this case we mean https://godoc.org/storj.io/storj and https://github.com/storj/docs, yeah. i'll add links

.github/pull_request_template.md
```diff
@@ -2,7 +2,15 @@
...
+ - [ ] Does any documentation need updating?
+
+Thanks for submitting a PR!
+
+![](https://media.giphy.com/media/14nakW0jC4HA6k/giphy.gif)
```
JessicaGreben: this gif isn't rendering for me when i "view" this file 😢

.github/pull_request_template.md
navillasa: ^that first link doesn't work for me JT. but yeah agree, maybe the checklist should say "Does the new code have up-to-date documentation that answers 'how do I use it?' and 'what does it do?'"?

.github/pull_request_template.md
egonelbre: It's about trying to predict whether this PR should be followed up by a refactor to make the code clearer. Or the refactor should be done before the PR.

This is about trying to figure out whether the PR is adding complexity without trying to clarify the code structure, before the making the change is difficult. Not quite sure how to nicely word it.

.github/pull_request_template.md
jtolds: ugh, our meta tags are still broken on storj.io so godoc doesn't work. i'll pull the link until i fix that

.github/pull_request_template.md
jtolds: i thiiiiink it should work in the actual PR though. i think it's a bit more loose

PR #1537, opened 2019-03-20
Macaroon Library

Barterio:


pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,217 @@
...
+			return nil, 0, errors.New("varint error")
+		}
+
+		return data[n:], int(value), nil
+	}
+	if n == 0 {
+		return nil, 0, errors.New("varint error")
+	}
+	return nil, 0, errors.New("varint error")
+}
```
crawter: value, n := binary.Uvarint(data)

	if n <= 0 || value > 0x7fffffff {
		return nil, 0, errors.New("varint error")
	}

	return data[n:], int(value), nil

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,217 @@
...
+		fieldType: fieldSignature,
+		data:      m.tail,
+	})
+
+	return data
+}
+
+// serializePacket converts packet to binary
+func serializePacket(data []byte, p packet) []byte {
+
```
crawter: remove empty line?

pkg/macaroon/serialize.go
Barterio: Good point

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,133 @@
...
+	_, err = rand.Read(secret)
+	if err != nil {
+		return nil, err
+	}
+
+	return secret, nil
+}
+
+// NewNonce generates cryptographically random 32 bytes
+func NewNonce() (nonce []byte, err error) {
```
jtolds: since NewSecret() and NewNonce() are identical, maybe just call both NewRand or something? or call both NewSecret()?

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Macaroon is a struct that determine contextual caveats and authorization
+type Macaroon struct {
+	head    []byte
+	caveats []Caveat
+	tail    []byte
+}
+
+// Caveat is a struct that determine restriction
+type Caveat struct {
```
jtolds: so, you'll want a Caveat to have lots of things in it, potentially being defined as a protobuf. something like https://github.com/storj/storj/blob/0dc098cd652bb5e80f025f2755026c419a3a77ea/lib/uplink/apikey.go#L33

so then the caveat will need a serialization mechanism so that you can turn it into a []byte for signing.

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return nonce, nil
+}
+
+// AddFirstPartyCaveat creates signed macaroon with appended caveat
+func (m *Macaroon) AddFirstPartyCaveat(c Caveat) (macaroon *Macaroon, err error) {
+	macaroon = m.Copy()
+
+	macaroon.caveats = append(macaroon.caveats, c)
+	macaroon.tail = sign(macaroon.tail, []byte(c.Identifier))
```
jtolds: so, right, here, if there was

```
func (c *Caveat) Serialize() []byte { ... }
```
then this line would be
```
macaroon.tail = sign(macaroon.tail, c.Serialize())
```

pkg/macaroon/macaroon.go
rikysya: But doesn't caveat should contain only one restriction in order for the process of creating more restricting macaroons to work correctly? Also, the restriction part you are talking about is what ID of the caveat is, there is also VerificationID and optional Location. Those parts are described in the official documentation. VerificationID is needed when you have a third party macaroon, for example. Anyway on this layer we don't care what restrictions caveat contains in it's ID because it should not affect the general flow of authorizing and serializing of the macaroon.

pkg/macaroon/macaroon.go
rikysya: agree, we can just use another secret as nonce

pkg/macaroon/macaroon.go
jtolds: > But doesn't caveat should contain only one restriction in order for the process of creating more restricting macaroons to work correctly? 
a caveat does not need to only have one restriction, no

> Anyway on this layer we don't care what restrictions caveat contains in it's ID because it should not affect the general flow of authorizing and serializing of the macaroon.
okay, but then why have a caveat struct?

pkg/macaroon/macaroon_test.go
```diff
@@ -0,0 +1,82 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package macaroon_test
```
mobyvb: More test suggestions:
Verify that `AddFirstPartyCaveat` returns an identical macaroon with two separate calls adding the same caveat
Verify that it returns two different macaroons when `AddFirstPartyCaveat` is called on the same macaroon, but with two different caveats
Verify that it returns two different macaroons when `AddFirstPartyCaveat` is called on two different macaroons, but with the same caveat

mobyvb: Maybe not all of these cases are necessary, so implement at your own discretion :)

jtolds:
okay, after a lot of thought, i think i'm convinced that you're all correct, let's just have this library ignore caveat formatting specifically and just do byte slices for caveats.

jtolds:
instead, we can add the storj v3 specific stuff on top, like https://github.com/storj/storj/pull/1862

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,127 @@
+// Copyright (C) 2019 Storj Labs, Inc. // See LICENSE for copying information.
```
egonelbre: linebreak missing

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,127 @@
...
+// Head returns copy of macaroon head
+func (m *Macaroon) Head() (head []byte) {
+	if len(m.head) == 0 {
+		return nil
+	}
+	return append([]byte(nil), m.head...)
+}
+
+// CaveatLen returns the number of caveats this macaroon has
+func (m *Macaroon) CaveatLen() int {
```
egonelbre: maybe `CaveatCount`?

pkg/macaroon/macaroon_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	assert "github.com/stretchr/testify/require"
+	"storj.io/storj/pkg/macaroon"
+)
+
+func TestNilMacaroon(t *testing.T) {
+	mac, err := macaroon.NewUnrestricted(nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, mac)
+	data := mac.Serialize()
```
egonelbre: some whitespace between each different check would be nice :)

pkg/macaroon/macaroon_test.go
```diff
@@ -0,0 +1,83 @@
...
+package macaroon_test
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	"storj.io/storj/pkg/macaroon"
+)
+
+func TestNilMacaroon(t *testing.T) {
```
egonelbre: I would love to see a fuzzing test, because it's easy to get parsing wrong.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+	var buf [binary.MaxVarintLen32]byte
+	n := binary.PutUvarint(buf[:], uint64(x))
+
+	return append(data, buf[:n]...)
+}
+
+// ParseMacaroon converts binary to macaroon
+func ParseMacaroon(data []byte) (*Macaroon, error) {
+	// skip version
+	data = data[1:]
```
egonelbre: Check for version missing.
Data length is not checked.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if sig.fieldType != fieldSignature {
+		return nil, errors.New("unexpected field found instead of signature")
+	}
+	if len(sig.data) != 32 {
+		return nil, errors.New("signature has unexpected length")
+	}
+	mac.tail = make([]byte, 32)
+	copy(mac.tail[:], sig.data)
+	//return data, nil
```
egonelbre: leftover code

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+	// Parse Identity
+	// Parse caveats
+	// Parse tail
+	return &mac, nil
+}
+
+// parseSection returns data leftover and packet array
+func parseSection(data []byte) ([]byte, []packet, error) {
```
egonelbre: tip: Usually these parsers are much easier to write if you stuff the data into a byte array. 

```
type decoder struct {
    data []byte
    result *Macaroon
}
func (dec *decoder) FieldType() (fieldType, error) { ... }
func (dec *decoder) Int64() (fieldType, error) { ... }
```

But current approach is fine as well.

pkg/macaroon/macaroon_test.go
jtolds: one thought i had looking at this - instead of writing all of this varint serialization by hand, etc, this is... exactly what protobufs are good for. it probably departs from any standards, but we could just do

```
message Macaroon {
   bytes head = 1;
   repeated bytes caveat = 2;
   bytes tail = 3;
}
```
and all of the parsing/serialization code would be written?

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+		if len(section) > 0 && section[0].fieldType == fieldLocation {
+			section = section[1:]
+		}
+		if len(section) == 0 || section[0].fieldType != fieldIdentifier {
+			return nil, errors.New("no Identifier in caveat")
+		}
+		cav := append([]byte(nil), section[0].data...)
+		section = section[1:]
+		if len(section) == 0 {
+			// First party caveat.
```
egonelbre: Should this be a todo note or something? It's unclear what needs to be done in the future.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+			mac.caveats = append(mac.caveats, cav)
+			continue
+		}
+		if len(section) != 1 {
+			return nil, errors.New("extra fields found in caveat")
+		}
+		if section[0].fieldType != fieldVerificationID {
+			return nil, errors.New("invalid field found in caveat")
+		}
+		//cav.VerificationId = section[0].data
```
egonelbre: Should this be a todo note or something? It's unclear what needs to be done in the future.

pkg/macaroon/macaroon.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		// Error skipped because sha256 does not return error
+		panic(err)
+	}
+
+	return signer.Sum(nil)
+}
+
+// NewSecret generates cryptographically random 32 bytes
+func NewSecret() (secret []byte, err error) {
```
egonelbre: Is there a reason we are not using `storj.Key` here?

pkg/macaroon/macaroon_test.go
egonelbre: I would rather follow a standard over pb

jtolds: well, "standard". the paper didn't propose a wire format, but perhaps working with https://github.com/rescrv/libmacaroons is good?

egonelbre: Yeah, looks like there's no great standard https://news.ycombinator.com/item?id=16073756.

thepaul: https://github.com/rescrv/libmacaroons/blob/master/doc/format.txt is emerging as a standard that other libraries support.

pkg/macaroon/macaroon.go
ifraixedes: although `len` is commonly used for knowing the size of slice/array in Go, personally I see that `*Count` is more intuitively from the human point of view for this case.

I would rather use `CaveatsCount` in plural than in singular.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+// See LICENSE for copying information.
+
+package macaroon
+
+import (
+	"encoding/binary"
+	"errors"
+)
+
+type fieldType int
```
ifraixedes: Do we expect negative numbers?
If not, I personally use `uint`.

I personally always prefer to specify the size rather than defaulting to some that change depending of the underlying architecture, but that's another story.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+func serializePacket(data []byte, p packet) []byte {
+	data = appendVarint(data, int(p.fieldType))
+	data = appendVarint(data, len(p.data))
+	data = append(data, p.data...)
+
+	return data
+}
+
+func appendVarint(data []byte, x int) []byte {
+	var buf [binary.MaxVarintLen32]byte
```
ifraixedes: No need that amount of allocation, considering that `uint64` is used and per https://golang.org/ref/spec#Size_and_alignment_guarantees with 8 is enough.

pkg/macaroon/serialize.go
ifraixedes: Below, I've seen that used `-1` as an invalid value, so `uint` cannot be if that's not changed.
I would still consider to use `uint` and use `0` as an invalid value.

However, I haven't read the Macaroons spec, so I don't know if it must be in this way.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	p.data = data[0:packLen]
+
+	return data[packLen:], p, nil
+}
+
+func parseVarint(data []byte) ([]byte, int, error) {
+	value, n := binary.Uvarint(data)
+	if n <= 0 || value > 0x7fffffff {
```
ifraixedes: if I'm not mistaken, `0x7fffffff` is the maximum value that an int32 can hold.
Being that the case, use `Math.MaxInt32` because it's clearer for the reader.

pkg/macaroon/serialize.go
```diff
@@ -0,0 +1,205 @@
...
+func parseSection(data []byte) ([]byte, []packet, error) {
+	prevFieldType := fieldType(-1)
+	var packets []packet
+	for {
+		if len(data) == 0 {
+			return nil, nil, errors.New("section extends past end of buffer")
+		}
+		rest, p, err := parsePacket(data)
+		if err != nil {
+			return nil, nil, err
```
ifraixedes: I would add more context to this error.
For example, being a user of this library, I could get here an error with message `"out of bounds"` (currently returned by line 186 of this file). The function is unexported, but it's called by `ParseMacaroon` one (currently at least in the line 79 of this file).

As a user, I will call the exported function `ParseMacaroon` and if I take as example the lines commented above, I will get and error `"out of bounds"` which without digging into the implementation I see that it won't provide me that much information about what's going on. This error could be improved with a message that it's being composed to the upstream caller just adding the operation that the inner functions do, for example the end error message would be more useful if it ends up something like `"parse section out of bounds"`.

I'm not entirely sure how this should be applied to this code base, I know that we use https://github.com/zeebo/errs and some other internal ones, and considering those, could be using a `Class`, etc. I think that other reviewers have more insights in how to do it aligned with the current code base/conventions.

iglesiasbrandon:
@Barterio I am going to merge this and you can address some of this comments in a different PR.

PR #1538, opened 2019-03-20
kademlia ping tracking

egonelbre:


cmd/storagenode/dashboard.go
```diff
@@ -90,10 +84,20 @@ func printDashboard(data *pb.DashboardResponse, online bool) error {
...
+	if err != nil {
+		lastContacted = time.Time{}
+	}
+	lastQueried, err := ptypes.Timestamp(data.LastQueried)
+	if err == nil {
+		if lastQueried.After(lastContacted) {
+			lastContacted = lastQueried
+		}
+	}
+	if lastContacted.IsZero() {
```
aleitner: Should we make this red if `(now - lastContacted > X ms) || lastContacted < now-data.GetUptime()` or

jtolds: maybe the first one? the second one will never be false since it only keeps track of last contacted in memory since uptime started. what should X be for the first one?

aleitner: Since there are no periodic pings I would probably be worried after maybe 24 hours?

jtolds:
FYI: https://github.com/storj/storj/pull/1539 is merged to release-alpha6

cmd/storagenode/dashboard.go
jtolds: done

egonelbre:
Note, update dashboard tests in `storagenode/dashboard`

cmd/storagenode/dashboard.go
```diff
@@ -19,12 +19,13 @@ import (
 	"go.uber.org/zap"
 
 	"storj.io/storj/internal/memory"
-	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/process"
 	"storj.io/storj/pkg/transport"
 )
 
+const contactWindow = time.Hour * 4
```
mniewrzal: `contactWindow` looks to be quite long but I have no strong opinion about that. @stefanbenten or @littleskunk

mniewrzal: nevermind, its just about colors in dashboard.

mniewrzal:
one more thing `Last Contact 0s ago` might be confusing, it can be interpreted as no contact.

littleskunk:
The highest value in production was 30 seconds from a user that joined while we updated the network. He don't has any data for audits. This is more or less our idle kademlia traffic. -> A contactWindows of 10 minutes should notify the user early and still avoid false positives.

jtolds:
> The highest value in production was 30 seconds from a user that joined while we updated the network. He don't has any data for audits. This is more or less our idle kademlia traffic. -> A contactWindows of 10 minutes should notify the user early and still avoid false positives.

this is actually going to change - in production we don't need to be doing kademlia refreshes so often. these small time windows are only during the early stages of the network

PR #1543, opened 2019-03-21
fix path validation with sj:/bucket

dylanlott:
This PR updates the `fpath` file tests and adds some validation handling around URL schemes with the format `sj:/bucket/`. 

The package now correctly handles paths with this format and automatically formats them to `sj://bucket`

## Code Review Checklist

Thanks for submitting a PR!


pkg/utils/urls.go
```diff
@@ -0,0 +1,31 @@
...
+package utils
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// ValidateURL takes in a url and returns a boolean if it's valid
+func ValidateURL(dst string) bool {
```
egonelbre: All path handling is in `fpath` package.

`utils` package is usually the worst place where to put things because they end up as kitchen sinks.

pkg/utils/urls.go
```diff
@@ -0,0 +1,31 @@
...
+	parsed, err := url.Parse(dst)
```
egonelbre: our paths are not URL-s, see examples in `fpath`

pkg/utils/urls.go
```diff
@@ -0,0 +1,31 @@
...
+		fmt.Printf("error parsing URL: %+v", err)
+		return false
+	}
+
+	if parsed.Scheme == "" {
+		fmt.Printf("url must have a valid scheme %s", parsed.Scheme)
+		return false
+	}
+
+	if strings.Contains(dst, "///") {
```
egonelbre: `///` is completely valid part of an object path

pkg/utils/urls.go
dylanlott: I'll move this over to fpath then.

pkg/utils/urls.go
dylanlott: This was an assumption on my part, I'll update the tests and the code to reflect this.

pkg/utils/urls.go
dylanlott: I'll remove this check.

egonelbre:
ah note forgot that in https://github.com/storj/storj/blob/master/internal/fpath/path.go#L30 we already do, hopefully correct, validation on the paths.

stefanbenten:
@dylanlott I pushed a change to your PR here, which simplifies the validation drastically.
Thanks to our meet yesterday, i came up with just handling the single slash "as valid", as the url.Parse() is changing it anyway.
I just tested it with the storj-sim and it worked fine 👍 

Let me know what you think!

internal/fpath/path.go
```diff
@@ -37,36 +37,37 @@ func New(p string) (FPath, error) {
...
 	var u *url.URL
 	var err error
+	//Loops to ensure URL is formatted correctly and does not get malformed during url.Parse()
 	for {
 		u, err = url.Parse(p)
 		if err != nil {
 			return fp, fmt.Errorf("malformed URL: %v, use format sj://bucket/", err)
 		}
-
+		//no scheme means local path
```
egonelbre: nit: space after `//`

internal/fpath/path.go
```diff
@@ -37,36 +37,37 @@ func New(p string) (FPath, error) {
...
 		}
-
+		//u.host equals the bucket name, if existing url is valid
 		if u.Host != "" {
 			break
 		}
-
-		p = strings.Replace(p, ":///", "://", 1)
+		//remove additional / if url.Parse() corrects from sj:/bucket to sj:///bucket
+		p = strings.Replace(u.String(), ":///", "://", 1)
```
egonelbre: The whole looping seems weird :/

stefanbenten: Yeah kind of, but ensures that the weird behaviour of url.Parse is taken care of

internal/fpath/path.go
stefanbenten: Done

PR #1545, opened 2019-03-21
adds slow kad dialer tests, adds timeout interceptor to transport

navillasa:
## why changes
- Kademlia service methods don't have a timeout for single shot rpc calls and they need one. Adding context.WithTimeout everywhere isn't very convenient, so we're adding a unary interceptor as a dial option to transport dials, which will also create a context with timeout.
- Kad dialer also needs tests to confirm that it will timeout if a dial or request take too long.

## what changes
- adds tests for the kad dialer to make sure that methods times out when there's a slow connection (the tests use slowTransport)
- adds a maybe redundant timeout test for kademlia.PingNode (originally intended to be an endpoint.Query test, but realized that function doesn't return transport errors)
- adds a UnaryInterceptor to the transport client to ensure that unary rpc calls come with a timeout
- adds a transport.NewClientWithTimeout constructor in addition to NewClient (which now uses the package's default request timeout of 20 seconds)

## Code Review Checklist
 - [no] Could the PR be broken into smaller PRs?
 - [I think so.] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [I think so.] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [Not to my knowledge.] Does any documentation need updating?
 
Thanks for submitting a PR!

### DIFF
--- a/bootstrap/peer.go
+++ b/bootstrap/peer.go
@@ -96,7 +96,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config) (*P
 			return nil, errs.Combine(err, peer.Close())
 		}
 
-		peer.Transport = transport.NewClient(options, 20*time.Second)
+		peer.Transport = transport.NewClient(options)
 
 		peer.Server, err = server.New(options, sc.Address, sc.PrivateAddress, nil)
 		if err != nil {
--- a/bootstrap/peer.go
+++ b/bootstrap/peer.go
@@ -96,7 +96,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config) (*P
 			return nil, errs.Combine(err, peer.Close())
 		}
 
-		peer.Transport = transport.NewClient(options)
+		peer.Transport = transport.NewClient(options, 20*time.Second)
 
 		peer.Server, err = server.New(options, sc.Address, sc.PrivateAddress, nil)
 		if err != nil {
--- a/pkg/certificates/certificates_test.go
+++ b/pkg/certificates/certificates_test.go
@@ -736,7 +736,7 @@ func TestNewClient(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(ident, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClientWithTimeout(tlsOptions, 20*time.Millisecond)
+	clientTransport := transport.NewClient(tlsOptions)
 
 	t.Run("Basic", func(t *testing.T) {
 		client, err := NewClient(ctx, clientTransport, listener.Addr().String())
--- a/pkg/certificates/certificates_test.go
+++ b/pkg/certificates/certificates_test.go
@@ -655,7 +655,7 @@ func TestCertificateSigner_Sign_E2E(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(clientIdent, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClient(tlsOptions, 20*time.Millisecond)
+	clientTransport := transport.NewClientWithTimeout(tlsOptions, 20*time.Millisecond)
 
 	client, err := NewClient(ctx, clientTransport, service.Addr().String())
 	require.NoError(t, err)
@@ -736,7 +736,7 @@ func TestNewClient(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(ident, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClient(tlsOptions, 20*time.Millisecond)
+	clientTransport := transport.NewClientWithTimeout(tlsOptions, 20*time.Millisecond)
 
 	t.Run("Basic", func(t *testing.T) {
 		client, err := NewClient(ctx, clientTransport, listener.Addr().String())
--- a/pkg/certificates/certificates_test.go
+++ b/pkg/certificates/certificates_test.go
@@ -655,7 +655,7 @@ func TestCertificateSigner_Sign_E2E(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(clientIdent, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClient(tlsOptions)
+	clientTransport := transport.NewClient(tlsOptions, 20*time.Millisecond)
 
 	client, err := NewClient(ctx, clientTransport, service.Addr().String())
 	require.NoError(t, err)
@@ -736,7 +736,7 @@ func TestNewClient(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(ident, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClient(tlsOptions)
+	clientTransport := transport.NewClient(tlsOptions, 20*time.Millisecond)
 
 	t.Run("Basic", func(t *testing.T) {
 		client, err := NewClient(ctx, clientTransport, listener.Addr().String())
--- a/pkg/certificates/certificates_test.go
+++ b/pkg/certificates/certificates_test.go
@@ -655,7 +655,7 @@ func TestCertificateSigner_Sign_E2E(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(clientIdent, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClientWithTimeout(tlsOptions, 20*time.Millisecond)
+	clientTransport := transport.NewClient(tlsOptions)
 
 	client, err := NewClient(ctx, clientTransport, service.Addr().String())
 	require.NoError(t, err)
--- a/pkg/kademlia/dialer_test.go
+++ b/pkg/kademlia/dialer_test.go
@@ -160,7 +160,7 @@ func TestDialer(t *testing.T) {
 
 func TestSlowDialerHasTimeout(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 3,
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 0,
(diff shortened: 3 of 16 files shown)


pkg/kademlia/kademlia_planet_test.go
```diff
@@ -39,3 +47,52 @@ func TestRequestInfo(t *testing.T) {
 		require.Equal(t, node.Local().Restrictions.GetFreeBandwidth(), info.GetCapacity().GetFreeBandwidth())
 	})
 }
+
+func TestPingTimeout(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 0,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+
+		{
```
mniewrzal: Unnecessary code block.

pkg/transport/transport.go
```diff
@@ -30,15 +31,30 @@ type Client interface {
...
 // Transport interface structure
 type Transport struct {
-	tlsOpts   *tlsopts.Options
-	observers []Observer
+	tlsOpts        *tlsopts.Options
+	observers      []Observer
+	requestTimeout time.Duration
+}
+
+// InvokeTimeout enables timeouts for an overly long method call
```
mniewrzal: I think it will be for all request not only overlay

pkg/kademlia/dialer_test.go
```diff
@@ -152,6 +158,131 @@ func TestDialer(t *testing.T) {
...
+
+			dialer := kademlia.NewDialer(zaptest.NewLogger(t), slowClient)
+			defer ctx.Check(dialer.Close)
+
+			var group errgroup.Group
+			defer ctx.Check(group.Wait)
+
+			group.Go(func() error {
+				_, err := dialer.FetchPeerIdentity(ctx, planet.Satellites[0].Local())
+				require.NotNil(t, err)
```
mniewrzal: I believe that you don't have to `require.NotNil` if next you are using `require.Error`

bootstrap/peer.go
```diff
@@ -96,7 +96,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config) (*P
 			return nil, errs.Combine(err, peer.Close())
 		}
 
-		peer.Transport = transport.NewClient(options)
+		peer.Transport = transport.NewClient(options, 20*time.Second)
```
mniewrzal: this look like a candidate for option flag or at least on const with default value. Something like `transport.NewClient(options, transport.DefaultRequestTimout)`.

pkg/transport/transport.go
navillasa: I had "overly long method call" there to mean that the method runs for too long or "over" what it should (not "overlay") but I can take it out so there's no confusion.

pkg/kademlia/dialer_test.go
navillasa: oops, that's fair

pkg/transport/transport.go
```diff
@@ -30,15 +31,30 @@ type Client interface {
...
+type InvokeTimeout struct {
```
egonelbre: move this to a separate file

pkg/transport/transport.go
```diff
@@ -30,15 +31,30 @@ type Client interface {
...
+func (it InvokeTimeout) Intercept(ctx context.Context, method string, req interface{}, reply interface{},
+	cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	timedCtx, cancel := context.WithTimeout(ctx, it.Timeout)
+	defer cancel()
+	return invoker(timedCtx, method, req, reply, cc, opts...)
 }
 
 // NewClient returns a newly instantiated Transport Client
-func NewClient(tlsOpts *tlsopts.Options, obs ...Observer) Client {
+func NewClient(tlsOpts *tlsopts.Options, requestTimeout time.Duration, obs ...Observer) Client {
```
egonelbre: let's make a separate func `NewClientWithTimeout`, and make `NewClient` use the default defined in the package. Rest of the code will be clearer.

pkg/transport/transport.go
```diff
@@ -65,9 +81,10 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 		dialOption,
 		grpc.WithBlock(),
 		grpc.FailOnNonTempDialError(true),
+		grpc.WithUnaryInterceptor(InvokeTimeout{transport.requestTimeout}.Intercept),
 	}, opts...)
 
-	timedCtx, cancel := context.WithTimeout(ctx, connWaitTimeout)
+	timedCtx, cancel := context.WithTimeout(ctx, transport.requestTimeout)
```
egonelbre: connect timeout and request timeout are different things

bootstrap/peer.go
navillasa: okay, I'm fine with adding that. can @egonelbre confirm that's cool before I change it?

pkg/transport/transport.go
navillasa: ok sounds good. and should the default be on the transport struct or a const?

pkg/transport/transport.go
navillasa: okay so there should be different timeout values for each one? should they both be field values on `Transport`?

pkg/transport/transport.go
egonelbre: making the default a const will be probably clearer

navillasa: ok 👍

pkg/kademlia/kademlia_planet_test.go
navillasa: also @egonelbre - is this test redundant with the dialer tests I added? I wanted to cover the endpoint functions (like Query) but realized that no errors are returned so it's kind of hard to check that the context deadline was hit. So I went down a level, but the `kademlia.Ping` function basically just wraps `dialer.PingNode` which already has a timeout test.

pkg/transport/transport.go
egonelbre: yeah, but not sure whether they should have different values

navillasa: 😬I guess I'll just set their default values as `20*time.Second` for now.

pkg/transport/transport.go
mniewrzal: sorry, no need to change

pkg/transport/transport.go
navillasa: I changed `connWaitTimeout` to `defaultDialTimeout` - does that seem better?

pkg/transport/transport.go
```diff
@@ -30,15 +31,26 @@ type Client interface {
...
 type Transport struct {
-	tlsOpts   *tlsopts.Options
-	observers []Observer
+	tlsOpts        *tlsopts.Options
+	observers      []Observer
+	requestTimeout time.Duration
 }
 
-// NewClient returns a newly instantiated Transport Client
+// NewClient returns a transport client with a specified timeout for requests
```
mniewrzal: maybe `with a default timeouts for requests`

pkg/transport/common.go
```diff
@@ -15,5 +15,7 @@ var (
 	//Error is the errs class of standard Transport Client errors
 	Error = errs.Class("transport error")
 	// default time to wait for a connection to be established
-	connWaitTimeout = 20 * time.Second
+	defaultDialTimeout = 20 * time.Second
```
mniewrzal: better to use `const`

pkg/kademlia/dialer_test.go
```diff
@@ -152,6 +158,127 @@ func TestDialer(t *testing.T) {
 	})
 }
 
+func TestSlowDialerHasTimeout(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 3,
```
mniewrzal: Why 3 uplinks? I don't where there are used.

pkg/certificates/certificates_test.go
```diff
@@ -736,7 +736,7 @@ func TestNewClient(t *testing.T) {
 
 	tlsOptions, err := tlsopts.NewOptions(ident, tlsopts.Config{})
 	require.NoError(t, err)
-	clientTransport := transport.NewClient(tlsOptions)
+	clientTransport := transport.NewClientWithTimeout(tlsOptions, 20*time.Millisecond)
```
egonelbre: Is there a reason we are not using default values here?

pkg/peertls/tlsopts/options_test.go
```diff
@@ -142,7 +143,7 @@ func TestOptions_ServerOption_Peer_CA_Whitelist(t *testing.T) {
 			dialOption, err := opts.DialOption(target.Id)
 			require.NoError(t, err)
 
-			transportClient := transport.NewClient(opts)
+			transportClient := transport.NewClientWithTimeout(opts, 20*time.Millisecond)
```
egonelbre: Is there a reason we are not using default values here?

pkg/certificates/certificates_test.go
navillasa: not really, I just thought it shouldn't take longer than that amount of time but I guess we're not explicitly testing for that here. I'll take it out.

pkg/transport/transport.go
```diff
@@ -30,15 +31,26 @@ type Client interface {
...
 func NewClient(tlsOpts *tlsopts.Options, obs ...Observer) Client {
 	return &Transport{
```
egonelbre: usually here it's nicer to call the more complex func here with the arguments, it avoids bugs by forgetting to add fields

pkg/peertls/tlsopts/options_test.go
navillasa: same answer^

pkg/transport/transport.go
navillasa: oh like, return `NewClientWithTimeout` but pass in the default?

egonelbre: yes

PR #1549, opened 2019-03-22
change default to true to enforce sa whitelist

JessicaGreben:
In order for satellite whitelisting to be enabled and enforced on the storagenode, `SatelliteIDRestriction` config setting must be set to true, but previously the default setting was false.  This means that both `WhitelistedSatelliteIDs` and `SatelliteIDRestriction` would need to be specified in the config.yaml for storagenodes.  This is confusing. As an example, currently the `storjlabs/storagenode:alpha` docker image is misconfigured and only sets `WhitelistedSatelliteIDs` which does not enforce whitelisting like you might think. Because of this, it seems preferable if `SatelliteIDRestriction` is enabled by default reducing the risk it is mistakenly not configured in the future. This PR makes that change.

## Code Review Checklist

Thanks for submitting a PR!


jtolds:
perhaps this is a good new use case for @wthorp's devDefault setting?

PR #1551, opened 2019-03-22
database for tracking information without storing orders / agreements

egonelbre:
Adds database declarations for storing order limit and order information without storing the full information.

## Code Review Checklist


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,72 @@ read limitoffset (
...
+}
+
+model storagenode_storage_rollup (
+	key    bucket_id
+	unique bucket_id interval_start interval_seconds
+
+	field bucket_id        blob
+	field interval_start   utimestamp
+	field interval_seconds uint32
+	field remote           uint64
```
stefanbenten: What is this field for? Arent all segements regarding storagenodes remote?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,72 @@ read limitoffset (
     orderby desc bucket_usage.rollup_end_time
 )
 
+//--- used serials ---//
+
+model used_serial {
+	key    serial_number
+	unique serial_number storage_node_id
```
stefanbenten: Depending on our queries it makes sense to revert the order here.
I assume that we query/rollup based on the storagenodeID and not per serial

satellite/satellitedb/dbx/satellitedb.dbx
stefanbenten: Should this be the amount, to match the pattern of bucket_storage_rollup?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,72 @@ read limitoffset (
...
+
+// --- storage node roll up tables --- //
+
+model storagenode_bandwidth_rollup {
+	key    storagenode_id
+	unique storagenode_id interval_start interval_seconds action
+
+	field storagenode_id   blob
+	field interval_start   utimestamp
+	field interval_seconds uint32
```
stefanbenten: Do we need a seperate metric for the interval here? Wouldnt it make more sense to assume it takes until the next interval_start is inserted?

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: we always query <serial number, storage node id>

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: For bandwidth not really, if I'm querying let's say for a day, and a storagenode doesn't have any entries vs. it's difficult to understand whether the bandwidth should be accounted for or not.

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: yes, it's using `remote` to match the `bucket_storage_rollup`, but I would also be fine with `total` or `amount`. No strong opinion here.

satellite/satellitedb/dbx/satellitedb.dbx
stefanbenten: Yeah true, what i am focused on is to keep the indexes as small as possible. But we can change that at any time later

satellite/satellitedb/dbx/satellitedb.dbx
stefanbenten: Its fine, wasnt not entirely clear to me.

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: cardinality of interval_seconds should be really low so postgres should be able to optimize it really well

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: ah note, we do not roll up this table.... it's just here to 1. figure out bucket_id and 2. to avoid duplicates

stefanbenten: Oh yeah, than it should be fine.

satellite/satellitedb/dbx/satellitedb.dbx
stefanbenten: I trust you on that one 💯

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,72 @@ read limitoffset (
...
+model used_serial {
+	key    serial_number
+	unique serial_number storage_node_id
+
+	field serial_number   blob
+	field storage_node_id blob
+
+	field bucket_id  blob
+	field expires_at utimestamp
+}
```
jtolds: proposal: two tables. order_limit_info and used_serials

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,102 @@ read limitoffset (
     orderby desc bucket_usage.rollup_end_time
 )
 
+//--- used serials ---//
```
JessicaGreben: this comment is in the wrong place now

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,102 @@ read limitoffset (
...
+
+model serial_number (
+	key    id
+	unique serial_number
+
+	field id            serial
```
JessicaGreben: is this `id` the storage_node_id?

egonelbre: this is the serial number id, not sure how to name it better

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: not really, but the comment does need updating. both tables are necessary for tracking used serials and information

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: ohh that makes sense. do we also need the storage_node_id here?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -377,6 +377,102 @@ read limitoffset (
...
+delete serial_number (
+	where serial_number.expires_at <= ?
+)
+
+// for preventing duplicate serial numbers
+create used_serial ()
+
+// --- bucket roll up tables --- //
+
+model bucket_bandwidth_rollup (
```
JessicaGreben: now that we have the `bucket_bandwidth_rollup` and `bucket_storage_rollup`, do we still need `bucket_usage` table?

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: that's in the `used_serials` table

egonelbre: reason for not using `serial_number` itself as the key in `used_serials` is because the serial is 4bytes, but serial number itself is 16bytes. given 100 used entries per limit, we end up saving 1.2kb

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: Yeah, we can eventually remove them, but first we need to implement the new one.

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: hopefully the comment is now clearer

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: ok I think I understand.  The `serial_number` gets created when orderlimits are created for uplink so we store the bucket_id in that table.  And the `used_serial` gets created when the SN sends orders, so we store the store_node_id with that. 

Also that answers my question why store an additional `id` instead of `serial_num` itself, the 4bytes vs 16 bytes. thanks.

satellite/satellitedb/testdata/postgres.v9.sql
```diff
@@ -0,0 +1,232 @@
+-- AUTOGENERATED BY gopkg.in/spacemonkeygo/dbx.v1
```
JessicaGreben: it is confusing these say "autogenerated" when they aren't

egonelbre: techincally I copy pasted it from the schema that dbx generates :D, but I'll replace them

PR #1554, opened 2019-03-22
extension serialization

bryanchriswhite:
### [Jira ticket](https://storjlabs.atlassian.net/browse/V3-1441?atlOrigin=eyJpIjoiYjQ3ZjA0OGRkMzk1NDliYThjYTY5ZTQ1YjYxMzdhNTIiLCJwIjoiaiJ9)

Extensions are currently being serialized in a way that is incompatible with usage via grpc's [`credentials` ](https://godoc.org/google.golang.org/grpc/credentials#NewTLS) as well as certificate signing :man_facepalming:. This PR changes how we do certificate extension serialization (into the certificate) to be "correct" and compatible.

Currently, extensions are being persisted as separate PEM blocks, following the certificate to which they belong. When deserialized, they are being added to their respective certificate's `ExtraExtensions` field. This means that when a certificate is deserialized, the extension data is not available and therefore is incompatible with the way grpc does TLS handshakes (only the binary certificate data is sent to the remote peer).

Instead of "storing" extensions in certificates' `ExtraExtensions` field, this PR serializes the extensions into the `Raw` (and `RawTBSCertificate`) fields. This way, when the `x509.Certificate` struct is PEM-encoded for persistence on-disk, the extensions will be included in their respective certificate PEM block. 
_(see: [`x509.Certificate`](https://golang.org/pkg/crypto/x509/#Certificate) for context)_

When adding extensions the "correct" way, the certificate needs to be re-signed by it's parent (or self, e.g. CA certificate). This requires that the private key corresponding to that signing certificate be available when managing extensions; hence, `ManageablePeerIdentity`.

This PR also changes revocation extensions' values. Previously, the value included the hash of the revoked certificate. Revocations better correlate to keys rather than the entire certificate, so this replaces the hash of the whole certificate with the hash of just the public key.

## Code Review Checklist

Thanks for submitting a PR!


cmd/identity/certificate_authority.go
```diff
@@ -211,5 +211,5 @@ func cmdCAExtensions(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	return printExtensions(ca.Cert.Raw, ca.Cert.ExtraExtensions)
+	return printExtensions(ca.Cert.Raw, ca.Cert.Extensions)
```
navillasa: (clarifying not suggesting otherwise lol) why are we using `Extensions` and not `ExtraExtensions` here?

pkg/identity/identity.go
```diff
@@ -53,6 +55,14 @@ type FullIdentity struct {
 	Key crypto.PrivateKey
 }
 
+// ManageablePeerIdentity is a `PeerIdentity` and it's corresponding `FullCertificateAuthority`
```
navillasa: super mega nit - "it's" should be "its"

navillasa: also, what does the addition of a ManageablePeerIdentity do, and how did we get around making changes to IDs that require CA auth before its existence?

internal/testpeertls/revocations.go
```diff
@@ -8,50 +8,90 @@ import (
...
 // with a "revoking" certificate, which contains a revocation extension recording
 // this action.
-func RevokeLeaf(keys []crypto.PrivateKey, chain []*x509.Certificate) ([]*x509.Certificate, pkix.Extension, error) {
-	var revocation pkix.Extension
-	revokingKey, err := pkcrypto.GeneratePrivateKey()
-	if err != nil {
-		return nil, revocation, err
+func RevokeLeaf(caKey crypto.PrivateKey, chain []*x509.Certificate) ([]*x509.Certificate, pkix.Extension, error) {
+	if len(chain) < 2 {
+		return nil, pkix.Extension{}, errs.New("revoking leaf implies a CA exists; chain too short")
```
navillasa: nit - could maybe help to make a zeebo errs "revocations" `Error` class, and then make this `Error.New`? not critical though

pkg/peertls/extensions/extensions.go
```diff
@@ -127,8 +127,11 @@ func AddSignedCert(key crypto.PrivateKey, cert *x509.Certificate) error {
 	return nil
 }
 
-// AddExtension adds one or more extensions to a certificate
-func AddExtension(cert *x509.Certificate, exts ...pkix.Extension) (err error) {
+// AddExtraExtension adds one or more extensions to a certificate for serialization.
+// NB: this *does not* serialize or persist the extension into the certificates's
+// raw bytes. To add a persistent extension use `FullCertificateAuthority.AddExtension`
```
navillasa: when would a not persistent extension be used over a persistent one?

pkg/peertls/extensions/revocations.go
```diff
@@ -116,14 +103,13 @@ func revocationChecker(opts *Options) HandlerFunc {
...
 
-		caHash := pkcrypto.SHA256Hash(ca.Raw)
 		leafHash := pkcrypto.SHA256Hash(leaf.Raw)
 
 		// NB: we trust that anything that made it into the revocation DB is valid
 		//		(i.e. no need for further verification)
 		switch {
-		case bytes.Equal(lastRev.CertHash, caHash):
-			return ErrRevokedCert
+		case lastRev.Final:
```
navillasa: why does the fallthrough happen here?

pkg/peertls/peertls.go
```diff
@@ -132,26 +127,30 @@ func ChainBytes(chain ...*x509.Certificate) ([]byte, error) {
...
-	} else {
-		signingKey = key
-	}
-
+func NewCert(publicKey crypto.PublicKey, parentKey crypto.PrivateKey, template, parent *x509.Certificate) (*x509.Certificate, error) {
 	if parent == nil {
 		parent = template
 	}
 
+	// TODO: how to do this better?
```
navillasa: this seems fine?

pkg/peertls/tlsopts/tls_test.go
```diff
@@ -58,3 +66,97 @@ func TestVerifyIdentity_error(t *testing.T) {
 		})
 	}
 }
+
+func TestExtensionMap_HandleExtensions(t *testing.T) {
+	// TODO: rename and/or move test
```
navillasa: maybe this and the test below should be moved to revocation_db_test.go since it seems like they're testing if the revocations are working properly?

cmd/identity/batch.go
```diff
@@ -111,7 +111,7 @@ func saveIdentityTar(path string, key crypto.PrivateKey, id storj.NodeID) error
 		return err
 	}
 
-	caCert, err := peertls.NewCert(key, nil, ct, nil)
+	caCert, err := peertls.NewSelfSignedCert(key, ct)
```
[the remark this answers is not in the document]
bryanchriswhite: We definitely should - all (hopefully) the bad variable names are from old commits and I wanted to avoid adding more to the  PR. I'm not sure if it's worth making a ticket for but it might be a good idea to just do a PR with variable name improvements in a bunch of places to clean this sort of thing up.

cmd/identity/certificate_authority.go
```diff
@@ -139,7 +139,7 @@ func cmdRevokeCA(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	if err := extensions.AddRevocationExt(ca.Key, ca.Cert, ca.Cert); err != nil {
+	if err := ca.Revoke(); err != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: [`FullCertificateAuthority#Revoke`](https://github.com/storj/storj/blob/07b2165aefb8ba121cf23c3cbebfda7ee9880e97/pkg/identity/certificate_authority.go#L458) wraps `extensions.NewRevocationExt` and `FullCertificateAuthority#AddExtension`. `extensions.AddRevocationExt` was deleted.

cmd/identity/certificate_authority.go
bryanchriswhite: Basically, `Extensions` is used for reading extensions in a certificate; it's what's populated when deserializing a certificate. `ExtraExtensions` is used for writing extensions; it's not populated when parsing and is encoded into the certificate bytes when serializing (which is done with the `x509.CreateCertificate` function).

bryanchriswhite: see https://golang.org/pkg/crypto/x509/#Certificate
and https://golang.org/pkg/crypto/x509/#CreateCertificate

pkg/identity/identity.go
bryanchriswhite: I'm the worst with "it's" vs "its". `ManageablePeerIdentity` provides a convenient place to add methods to make changes to identities that require the presence of its (nailed it) CA private key. Previously we were only using the `ExtraExtensions` field on the certificate structs (which is wrong) which didn't require re-signing. Now that we're using `Extensions`, we need to re-sign the certificate to add an extension.

internal/testpeertls/revocations.go
bryanchriswhite: Good idea. I was being lazy because it was a test helper :p

pkg/peertls/extensions/extensions.go
bryanchriswhite: Probably never but in order for an extension to be serialized (i.e. in `Extensions` when deserialized), it first has to be added to `ExtraExtensions`.

Example:
```golang
originalCert := x509.CreateCertificate(...)
originalCert.ExtraExtensions = append(originalCert.ExtraExtensions, pkix.Extension{...})

// originalCert is passed as `template` so that its `ExtraExtensions` are read
extendedCertBytes, err := x509.CreateCertificate(rand.Reader, originalCert, ...)

// when parsed, `extendedCert.Extensions` will contain the extension and `extendedCert.ExtraExtensions` will be empty.
extendedCert, err = x509.ParseCertificate(extendedCertBytes)
```

pkg/peertls/extensions/revocations.go
bryanchriswhite: If the revocation is final then we consider any certificate chain with the same node ID as revoked. I'm not sure if this is necessarily the right way to do this. I think that we want all CA revocations to always be final and this implementation would allow for a user to create a non-final CA revocation. We should probably be keeping a hash of the public key of the certificate that's being revoked instead of a hash of the certificate (what we're currently doing).

cmd/identity/certificate_authority.go
```diff
@@ -182,7 +182,7 @@ func cmdRevokePeerCA(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	ext, err := extensions.NewRevocationExt(ca.Key, peerCA.Cert)
+	ext, err := extensions.NewRevocationExt(ca.Key, peerCA.Cert, true)
```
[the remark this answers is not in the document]
bryanchriswhite: I think this needs to be changed.. see the reply to the comment on the fallthrough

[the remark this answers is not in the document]
bryanchriswhite: Removed "final". Revocation hashes are based on certificate public keys now instead of the raw certificate.

internal/testpeertls/revocations.go
```diff
@@ -62,6 +102,6 @@ func NewRevokedLeafChain() ([]crypto.PrivateKey, []*x509.Certificate, pkix.Exten
 		return nil, nil, pkix.Extension{}, err
 	}
 
-	newChain, revocation, err := RevokeLeaf(keys, certs)
+	newChain, revocation, err := RevokeLeaf(keys[0], certs)
```
bryanchriswhite: maybe wrong inded? (maybe should be `peertls.CAIndex`)

pkg/identity/cert_authority_test.go
```diff
@@ -105,3 +112,71 @@ func BenchmarkNewCA(b *testing.B) {
...
+	err = ca.AddExtension(randExt)
+	assert.NoError(t, err)
+
+	assert.Len(t, ca.Cert.ExtraExtensions, 0)
+	assert.Len(t, ca.Cert.Extensions, len(oldCert.Extensions)+1)
+
+	assert.Equal(t, oldCert.SerialNumber, ca.Cert.SerialNumber)
+	assert.Equal(t, oldCert.IsCA, ca.Cert.IsCA)
+	assert.Equal(t, oldCert.PublicKey, ca.Cert.PublicKey)
+	assert.Equal(t, randExt, ca.Cert.Extensions[len(ca.Cert.Extensions)-1])
```
bryanchriswhite: could instead use an extension map

pkg/identity/identity_test.go
```diff
@@ -141,6 +147,75 @@ func TestVerifyPeer(t *testing.T) {
...
+	assert.Equal(t, oldLeaf.IsCA, manageableIdentity.Leaf.IsCA)
+	assert.Equal(t, oldLeaf.PublicKey, manageableIdentity.Leaf.PublicKey)
+	assert.Equal(t, randExt, manageableIdentity.Leaf.Extensions[len(manageableIdentity.Leaf.Extensions)-1])
+
+	assert.NotEqual(t, oldLeaf.Raw, manageableIdentity.Leaf.Raw)
+	assert.NotEqual(t, oldLeaf.RawTBSCertificate, manageableIdentity.Leaf.RawTBSCertificate)
+	assert.NotEqual(t, oldLeaf.Signature, manageableIdentity.Leaf.Signature)
+}
+
+func TestManageableIdentity_Revoke(t *testing.T) {
```
bryanchriswhite: should be `TestManageableFullIdentity`

pkg/peertls/extensions/revocations_test.go
```diff
@@ -37,68 +33,97 @@ func TestRevocationCheckHandler(t *testing.T) {
 			assert.NoError(t, err)
 		}
 
-		revokedLeafChain, leafRevocationExt, err := testpeertls.RevokeLeaf(keys, chain)
+		revokingChain, leafRevocationExt, err := testpeertls.RevokeLeaf(keys[0], chain)
```
bryanchriswhite: shouldn't this be `keys[peertls.CAIndex]`?

pkg/peertls/extensions/revocations_test.go
```diff
@@ -37,68 +33,97 @@ func TestRevocationCheckHandler(t *testing.T) {
...
 func TestRevocationUpdateHandler(t *testing.T) {
-	ctx := testcontext.New(t)
-	defer ctx.Cleanup()
-
 	testidentity.RevocationDBsTest(t, func(t *testing.T, revDB extensions.RevocationDB, _ storage.KeyValueStore) {
 		keys, chain, err := testpeertls.NewCertChain(2)
 		assert.NoError(t, err)
 
-		olderRevokedChain, olderRevocation, err := testpeertls.RevokeLeaf(keys, chain)
+		olderRevokedChain, olderRevocation, err := testpeertls.RevokeLeaf(keys[0], chain)
```
bryanchriswhite: should this be `keys[peertls.CAIndex]`?

pkg/peertls/extensions/revocations_test.go
```diff
@@ -37,68 +33,97 @@ func TestRevocationCheckHandler(t *testing.T) {
...
 		require.NoError(t, err)
 
 		time.Sleep(time.Second)
-		revokedLeafChain, newerRevocation, err := testpeertls.RevokeLeaf(keys, chain)
+		revokedLeafChain, newerRevocation, err := testpeertls.RevokeLeaf(keys[0], chain)
```
bryanchriswhite: ^

pkg/peertls/extensions/revocations_test.go
```diff
@@ -37,68 +33,97 @@ func TestRevocationCheckHandler(t *testing.T) {
...
 		require.NoError(t, err)
 
 		time.Sleep(time.Second)
-		newestRevokedChain, newestRevocation, err := testpeertls.RevokeLeaf(keys, revokedLeafChain)
+		newestRevokedChain, newestRevocation, err := testpeertls.RevokeLeaf(keys[0], revokedLeafChain)
```
bryanchriswhite: ^

internal/testpeertls/revocations.go
bryanchriswhite: nvm... this is all cleaned up in #1389

cmd/identity/revocations.go
```diff
@@ -53,12 +53,7 @@ func cmdRevocations(cmd *cobra.Command, args []string) error {
 
 	revErrs := new(errs.Group)
 	for _, rev := range revs {
-		// TODO: figure out why this doesn't base64 encode []byte fields!
-		//revJSON, err := json.MarshalIndent(rev.CertHash, "", "\t")
-		//if err != nil {
-		//	revErrs.Add(err)
-		//}
-		fmt.Printf("certificate hash: %s\n", base64.StdEncoding.EncodeToString(rev.CertHash))
+		fmt.Printf("certificate public key hash: %s\n", base64.StdEncoding.EncodeToString(rev.KeyHash))
```
coyle: Do we still want these print lines ?

internal/testpeertls/revocations.go
```diff
@@ -8,50 +8,88 @@ import (
...
-	if err != nil {
-		return nil, revocation, err
+func RevokeLeaf(caKey crypto.PrivateKey, chain []*x509.Certificate) ([]*x509.Certificate, pkix.Extension, error) {
+	if len(chain) < 2 {
+		return nil, pkix.Extension{}, extensions.Error.New("revoking leaf implies a CA exists; chain too short")
+	}
+	ca := &identity.FullCertificateAuthority{
+		Key:       caKey,
+		Cert:      chain[peertls.CAIndex],
+		RestChain: chain[:peertls.CAIndex+1],
```
coyle: Is it possible for the index to be out of bounds?

internal/testpeertls/revocations.go
```diff
@@ -8,50 +8,88 @@ import (
...
+		Key:       caKey,
+		RestChain: chain[peertls.CAIndex+1:],
+	}
+
+	if err = ca.Revoke(); err != nil {
+		return nil, pkix.Extension{}, err
+	}
+
+	var revocationExt *pkix.Extension
+	for _, ext := range ca.Cert.Extensions {
```
coyle: maybe pull this out into a sep function call?

internal/testpeertls/revocations.go
```diff
@@ -62,6 +100,6 @@ func NewRevokedLeafChain() ([]crypto.PrivateKey, []*x509.Certificate, pkix.Exten
 		return nil, nil, pkix.Extension{}, err
 	}
 
-	newChain, revocation, err := RevokeLeaf(keys, certs)
+	newChain, revocation, err := RevokeLeaf(keys[0], certs)
```
coyle: is it possible for keys to be nil or have a length of 0 ?

bryanchriswhite: no

internal/testpeertls/revocations.go
bryanchriswhite: the condition on :21 should prevent that

cmd/identity/revocations.go
bryanchriswhite: yes, this command prints revocation information,

internal/testpeertls/revocations.go
bryanchriswhite: good call - there's actually a function that does most of this already.

PR #1561, opened 2019-03-23
project usage limiting

JessicaGreben:
This PR makes the following changes:
- adds a query to accountingDB to retrieve the total GET bandwidth from`bucket_bandwidth_rollup` table
-  adds a query to accountingDB to retrieve the current inline and remote storage amount from`bucket_storage_rollup` table
- checks that project limits aren't exceed before the SA creates OrderLimits for the uplink
- adds a MaxAlphaUsage config flag to rollup

## Code Review Checklist

Thanks for submitting a PR!


cmd/uplink/cmd/cp.go
```diff
@@ -39,166 +35,6 @@ func init() {
 	expires = cpCmd.Flags().String("expires", "", "optional expiration date of an object. Please use format (yyyy-mm-ddThh:mm:ssZhh:mm)")
 }
 
-// upload transfers src from local machine to s3 compatible object dst
-func upload(ctx context.Context, src fpath.FPath, dst fpath.FPath, showProgress bool) (err error) {
```
egonelbre: Try to avoid moving and changing things in the same PR because it's unclear currently whether this func changed. (Of course sometimes it's unavoidable doing both.)

pkg/accounting/db.go
```diff
@@ -36,7 +37,35 @@ type Rollup struct {
 	AtRestTotal    float64
 }
 
-// DB stores information about bandwidth usage
+// BucketBWRollup mirrors dbx.BucketBandwidthRollup and holds the bucket bw rollup info
```
egonelbre: Avoid using shorthands they usually reduce legibility. If the name seems too long, this usually indicates that there's a nicer way to organize code.

satellite/satellitedb/accounting.go
```diff
@@ -20,6 +22,133 @@ type accountingDB struct {
...
+	if exceeded {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+const (
+	// GB in bytes
+	gbBytes = (1 << (10 * 3)) // 1073741824 bytes
```
egonelbre: we have `memory.GB`

satellite/satellitedb/accounting.go
```diff
@@ -20,6 +22,133 @@ type accountingDB struct {
...
+	// For the alpha release, no more than 25GB of storage on disk or bandwidth usage can be exceeded
+	alphaMaxUsage = uint64(25 * gbBytes)
```
egonelbre: This probably should be in a config flag somewhere, also unless you need a specific size for consts you can keep them unspecified, because they end up being easier used.

satellite/satellitedb/accounting.go
```diff
@@ -20,6 +22,133 @@ type accountingDB struct {
...
+)
+
+// exceedsAlphaBandwidthUsage returns true if more than 25GB of bandwidth usage has been used in the past time frame
+// TODO: remove this code once we no longer need usage limiting for alpha release
+// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+func (db *accountingDB) exceedsAlphaBandwidthUsage(ctx context.Context, projectID uuid.UUID, since time.Time) (bool, error) {
+	bwRollups, err := db.GetAllBucketBWByProjectAndInterval(ctx, projectID, since)
```
egonelbre: We probably should use a custom query for this, since iterating over all the entries is going to be significantly wasteful.

pkg/accounting/db.go
```diff
@@ -54,4 +83,6 @@ type DB interface {
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
 	// DeleteRawBefore deletes all raw tallies prior to some time
 	DeleteRawBefore(ctx context.Context, latestRollup time.Time) error
+	// ExceedsAlphaUsage checks if the usage limits are exceeded for a given project
+	ExceedsAlphaUsage(ctx context.Context, projectID uuid.UUID) (bool, error)
```
egonelbre: `ExceedsAlphaUsage` is more of a domain logic thing, so probably should be part of the accounting package. The DB is more of methods `ProjectBandwidthUsage(ctx, project, from, to)` and `ProjectStorageUsage(ctx, project, from, to)`.

satellite/satellitedb/accounting.go
```diff
@@ -8,9 +8,11 @@ import (
 	"database/sql"
 	"time"
 
+	"github.com/skyrings/skyring-common/tools/uuid"
 	"github.com/zeebo/errs"
 
 	"storj.io/storj/pkg/accounting"
+	acct "storj.io/storj/pkg/accounting"
```
egonelbre: Avoid using aliases, they give inconsistent names for packages and make harder to understand what is the shortened name.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -435,18 +438,33 @@ model bucket_bandwidth_rollup (
...
+	where  bucket_bandwidth_rollup.project_id = ?
+	where  bucket_bandwidth_rollup.interval_start >= ?
+)
+
 model bucket_storage_rollup (
 	key    bucket_id
 	unique bucket_id interval_start interval_seconds
 
-	field bucket_id        blob
+	field bucket_name       blob
```
egonelbre: Is there a way we can avoid having `bucket_name` and `project_id` here, they will end up using a lot of data.

cmd/uplink/cmd/cp.go
JessicaGreben: i reverted those changes that just moved things.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: I agree we shouldn't store all 3 fields, since `bucket_id` duplicates the other two fields. However it seems simpler to store `bucket_name` and `project_id` in this table and join those when we need `bucketID`, versus only storing the `bucket_id` and needing to parse out the projectID when we need it.  Thoughts?

jtolds: i agree with jessica - as long as we need to keep track of a project id and a bucket name, using two separate fields won't be a ton of extra overhead. as soon as we have real honest-to-goodness foreign keys for buckets then we can do a migration, just like we would have needed to do anwyays

satellite/metainfo/metainfo.go
```diff
@@ -122,11 +129,31 @@ func (endpoint *Endpoint) SegmentInfo(ctx context.Context, req *pb.SegmentInfoRe
...
 	if err != nil {
 		return nil, status.Errorf(codes.Unauthenticated, err.Error())
 	}
 
+	// Check if this projectID has exceeded alpha usage limits, i.e. 25GB of bandwidth or storage usage used in the past month
+	// TODO: remove this code once we no longer need usage limiting for alpha release
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+	projectID := keyInfo.ProjectID
+	const avgDaysInMonth = 30
+	from := time.Now().AddDate(0, -avgDaysInMonth, 0) // past 30 days
```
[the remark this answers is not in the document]
JessicaGreben: oops good catch. thanks

satellite/satellitedb/accounting.go
```diff
@@ -20,6 +23,40 @@ type accountingDB struct {
...
 }
 
+// ProjectBandwidthTotal returns the sum of GET bandwidth usage for a projectID for a time frame
+func (db *accountingDB) ProjectBandwidthTotal(ctx context.Context, projectID uuid.UUID, from time.Time) (uint64, error) {
+	var total uint64
+	var query = fmt.Sprintf(`
+		SELECT SUM (settled) as total
+		FROM bucket_bandwidth_rollup
+		WHERE project_id = ? AND interval_start >= ? AND action = %d
+		`, pb.BandwidthAction_GET,
```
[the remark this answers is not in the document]
JessicaGreben: yeah good point. im not sure why i did that.

internal/testplanet/planet.go
```diff
@@ -447,7 +447,8 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				Interval: 30 * time.Second,
 			},
 			Rollup: rollup.Config{
-				Interval: 120 * time.Second,
+				Interval:      120 * time.Second,
```
egonelbre: 2 * time.Minute

pkg/accounting/db.go
```diff
@@ -56,4 +58,12 @@ type DB interface {
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
 	// DeleteRawBefore deletes all raw tallies prior to some time
 	DeleteRawBefore(ctx context.Context, latestRollup time.Time) error
+	// CreateBucketStorageTally creates a record for BucketStorageTally in the accounting DB table
+	CreateBucketStorageTally(ctx context.Context, tally BucketStorageTally) error
+	// CreateBucketBandwidthRollup creates a record for BucketBandwidthRollup in the accounting DB table
+	CreateBucketBandwidthRollup(ctx context.Context, rollup BucketBandwidthRollup) error
```
egonelbre: I don't think we should have a separate `CreateBucketBandwidthRollup` the queries adding rollups almost always should be upserts. See changes in https://github.com/storj/storj/pull/1586/files#diff-4e3c8fc5bb52e5d050d2e5af986f4c9fR36

pkg/accounting/projectusage.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package accounting
+
+const (
+	// AvgDaysInMonth is how many days in a month
+	AvgDaysInMonth = 30
```
egonelbre: `AverageDaysInMonth`

satellite/metainfo/metainfo.go
```diff
@@ -124,6 +131,28 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
 		return nil, status.Errorf(codes.Unauthenticated, err.Error())
 	}
 
+	// Check if this projectID has exceeded alpha usage limits, i.e. 25GB of bandwidth or storage used in the past month
+	// TODO: remove this code once we no longer need usage limiting for alpha release
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+	projectID := keyInfo.ProjectID
+	from := time.Now().AddDate(0, 0, -accounting.AvgDaysInMonth) // past 30 days
+	bandwidthTotal, err := endpoint.accountingDB.ProjectBandwidthTotal(ctx, projectID, from)
```
egonelbre: Doing these queries for each segment might hit the database quite hard, but not sure whether there's an easy better approach.

satellite/metainfo/metainfo.go
```diff
@@ -40,22 +43,26 @@ type APIKeys interface {
...
-	orders    *orders.Service
-	cache     *overlay.Cache
-	apiKeys   APIKeys
+	log           *zap.Logger
+	pointerdb     *pointerdb.Service
+	orders        *orders.Service
+	cache         *overlay.Cache
+	apiKeys       APIKeys
+	accountingDB  accounting.DB
+	maxAlphaUsage int64
```
egonelbre: use `memory.Size` here

pkg/accounting/db.go
JessicaGreben: this will be removed once michal branch is merged in

JessicaGreben: once michal's branch is merged, i will use this method instead https://github.com/storj/storj/pull/1586/files#diff-53d71b1fb61033682939adbe999b0cbbR73

satellite/metainfo/metainfo.go
```diff
@@ -129,6 +136,28 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
...
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+	bucketID := createBucketID(keyInfo.ProjectID, req.Bucket)
+	from := time.Now().AddDate(0, 0, -accounting.AverageDaysInMonth) // past 30 days
+	bandwidthTotal, err := endpoint.accountingDB.ProjectBandwidthTotal(ctx, bucketID, from)
+	if err != nil {
+		endpoint.log.Error("retrieving ProjectBandwidthTotal")
+	}
+	inlineTotal, remoteTotal, err := endpoint.accountingDB.ProjectStorageTotals(ctx, keyInfo.ProjectID)
+	if err != nil {
+		endpoint.log.Error("retrieving ProjectStorageTotals")
```
mniewrzal: log error with `endpoint.log.Error("retrieving ProjectStorageTotals", zap.Error(err))`

satellite/metainfo/metainfo.go
```diff
@@ -129,6 +136,28 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
...
 	}
 
+	// Check if this projectID has exceeded alpha usage limits, i.e. 25GB of bandwidth or storage used in the past month
+	// TODO: remove this code once we no longer need usage limiting for alpha release
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+	bucketID := createBucketID(keyInfo.ProjectID, req.Bucket)
+	from := time.Now().AddDate(0, 0, -accounting.AverageDaysInMonth) // past 30 days
+	bandwidthTotal, err := endpoint.accountingDB.ProjectBandwidthTotal(ctx, bucketID, from)
+	if err != nil {
+		endpoint.log.Error("retrieving ProjectBandwidthTotal")
```
mniewrzal: log error with `endpoint.log.Error("retrieving ProjectBandwidthTotal", zap.Error(err))`

satellite/satellitedb/accounting.go
```diff
@@ -167,16 +235,19 @@ func (db *accountingDB) SaveBucketTallies(ctx context.Context, intervalStart tim
 	if len(bucketTallies) == 0 {
 		return Error.New("In SaveBucketTallies with empty bucketTallies")
 	}
+
 	for bucketID, info := range bucketTallies {
-		bID := dbx.BucketStorageTally_BucketId([]byte(bucketID))
+		bucketIDComponents := storj.SplitPath(bucketID)
+		bName := dbx.BucketStorageTally_BucketName([]byte(bucketIDComponents[0]))
```
mniewrzal: `bucketName`

satellite/satellitedb/accounting.go
```diff
@@ -167,16 +235,19 @@ func (db *accountingDB) SaveBucketTallies(ctx context.Context, intervalStart tim
...
+		pID := dbx.BucketStorageTally_ProjectId([]byte(bucketIDComponents[1]))
```
mniewrzal: `projectID`

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -410,15 +412,20 @@ model bucket_bandwidth_rollup (
...
 read scalar (
 	select bucket_bandwidth_rollup
-	where  bucket_bandwidth_rollup.bucket_id = ?
+	where  bucket_bandwidth_rollup.bucket_name = ?
+	where  bucket_bandwidth_rollup.project_id = ?
 	where  bucket_bandwidth_rollup.interval_start = ?
 	where  bucket_bandwidth_rollup.action = ?
 )
 
+create bucket_bandwidth_rollup ()
```
mniewrzal: I think this wont be used

satellite/satellitedb/orders.go
```diff
@@ -91,12 +102,17 @@ func (db *ordersDB) UpdateBucketBandwidthInline(ctx context.Context, bucketID []
 	now := time.Now()
 	intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
 
+	pathEl := bytes.Split(bucketID, []byte("/"))
```
mniewrzal: `pathElement`

pkg/accounting/db.go
mniewrzal: I think `CreateBucketBandwidthRollup` is not used anymore

satellite/satellitedb/accounting.go
```diff
@@ -20,6 +23,71 @@ type accountingDB struct {
 	db *dbx.DB
 }
 
+// ProjectBandwidthTotal returns the sum of GET bandwidth usage for a projectID for a time frame
+func (db *accountingDB) ProjectBandwidthTotal(ctx context.Context, bucketID []byte, from time.Time) (int64, error) {
```
mniewrzal: This method should have only `projectID` as parameter.

JessicaGreben: the test kept failing for this because of the way i was using projectID so i changed it to bucketID to parse it like how we do in `UpdateBucketBandwidthSettle`.  Could we leave this alone for now until we rethink passing the all the args as `bucketName, projectID`?

PR #1563, opened 2019-03-24
Move from Unique to Index on SatelliteDB

stefanbenten:
## Code Review Checklist

Thanks for submitting a PR!


satellite/satellitedb/migrate.go
```diff
@@ -337,7 +337,7 @@ func (db *DB) PostgresMigration() *migrate.Migration {
   						get_egress bigint NOT NULL,
   						audit_egress bigint NOT NULL,
   						PRIMARY KEY ( id ),
-  						UNIQUE ( rollup_end_time, bucket_id )
+						UNIQUE ( rollup_end_time, bucket_id )
```
mniewrzal: You shouldn't change old migrations.

satellite/satellitedb/migrate.go
```diff
@@ -407,18 +409,25 @@ func (db *DB) PostgresMigration() *migrate.Migration {
...
 						interval_start timestamp NOT NULL,
 						interval_seconds integer NOT NULL,
 						inline bigint NOT NULL,
 						remote bigint NOT NULL,
-						PRIMARY KEY ( bucket_id ),
-						UNIQUE ( bucket_id, interval_start, interval_seconds )
+						PRIMARY KEY ( bucket_id, interval_start )
 					)`,
+					`ALTER TABLE bucket_usages DROP CONSTRAINT bucket_usages_rollup_end_time_bucket_id_key`,
+					`CREATE UNIQUE INDEX IF NOT EXISTS bucket_id_rollup_end_time_index ON bucket_usages (
```
mniewrzal: better not use `IF NOT EXISTS` because we might miss some states that we didn't cover e.g. this index existed for some reason and we won't know about it during migration.

satellite/satellitedb/migrate.go
stefanbenten: Hm good point, why did that change.. I'll address that

satellite/satellitedb/migrate.go
stefanbenten: Done

PR #1564, opened 2019-03-24
Satellite orders receiving

mniewrzal:
This change adds satellite endpoint for receiving `OrderLimits` sent by storage node. 
Change includes:
* wire up orders sender in storage node (also in testplanet)
* saving serial number for `OrderLimit` in `serial_numbers` table
* satellite endpoint for receiving, verifying and storing `OrderLimit` and `Order` serial number
* initial implementation for `Orders` DB
* basic test for sending orders to satellite

https://storjlabs.atlassian.net/browse/V3-1433

If PR will be to big I can try split into two smaller.

## Code Review Checklist

Thanks for submitting a PR!


satellite/satellitedb/orders.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	tx, err := db.db.Open(ctx)
+	if err != nil {
+		return err
+	}
+
+	serialNumber := orderLimits[0].SerialNumber
+
+	_, err = tx.Create_SerialNumber(
```
JessicaGreben: this method accepts a list of `OrderLimit2`, but why is only 1 `serial_number` created? I thought we would want to list through the list of `orderLimits` and create a serial_number for each orderLimit in the list

mniewrzal: I'm not sure if stay as is at the end of changes. In this PR serial number is generated per uplink request. Main goal was to avoid storing one serial number for one storage node (up to 95 per request) and save DB space. Thats why we have two tables: `serial_number` and `used_serials` (serial_number|storagenode_id). Serial is added to `serial_number` and later when OrderLimit is received we are filling `used_serials`.

satellite/orders/endpoint.go
```diff
@@ -0,0 +1,169 @@
...
+				SerialNumber: orderLimit.SerialNumber,
+				Status:       pb.SettlementResponse_REJECTED,
+			})
+			if err != nil {
+				return formatError(err)
+			}
+		}
+
+		if err = endpoint.DB.SettleRemoteOrder(ctx, orderLimit, order); err != nil {
+			duplicateRequest := strings.Contains(err.Error(), "UNIQUE constraint failed") || strings.Contains(err.Error(), "violates unique constraint")
```
egonelbre: Does it get the same error text on non-english postgres?

storagenode/orders/sender.go
```diff
@@ -105,16 +106,18 @@ func (sender *Sender) Run(ctx context.Context) error {
 
 		if len(ordersBySatellite) > 0 {
 			var group errgroup.Group
-			ctx, cancel := context.WithTimeout(ctx, sender.config.Timeout)
-			defer cancel()
```
egonelbre: Why was this moved inside the Go func?

mniewrzal: To have timeout per satellite. The same way how you did it in `agreementsender`.

egonelbre: There's no need to create multiple contexts, because the timeout will be shared between them.

satellite/orders/endpoint.go
mniewrzal: Good question, didn't think about it. Its copied from bandwidth agreement `Settlement`. I check if there is more elegant way to do this.

storagenode/orders/sender.go
mniewrzal: Earlier sounds like a good idea :) I will change this.

PR #1568, opened 2019-03-25
libuplink changes for public usage

jtolds:
## Code Review Checklist


lib/uplink/apikey.go
```diff
@@ -0,0 +1,44 @@
...
+	return a.key
+}
+
+// ParseAPIKey parses an API Key
+func ParseAPIKey(val string) (APIKey, error) {
+	return APIKey{key: val}, nil
+}
+
+// Share returns a new API Key with the given caveat restrictions added
+func (a APIKey) Share(caveats ...Caveat) APIKey {
```
kaloyan-raev: I suggest removing anything related to sharing from the API for the Vanguard release. We have a separate milestone for Access Management, don't we? We can introduce the respective API with that release.

lib/uplink/bucket.go
```diff
@@ -0,0 +1,66 @@
...
+
+// OpenObject returns an Object handle, if authorized.
+func (b *Bucket) OpenObject(ctx context.Context, path storj.Path) (o *Object, err error) {
+	defer mon.Task()(&ctx)(&err)
+	panic("TODO")
+}
+
+// UploadOptions controls options about uploading a new Object, if authorized.
+type UploadOptions struct {
+	Metadata map[string]string
```
kaloyan-raev: We should have `ContentType string` field here too.

lib/uplink/object.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// ObjectMeta contains metadata about a specific Object
+type ObjectMeta struct {
+	Bucket   string
+	Path     storj.Path
+	IsPrefix bool
+
+	Metadata map[string]string
+
```
kaloyan-raev: `ContentType` field here too.

lib/uplink/bucket.go
```diff
@@ -0,0 +1,66 @@
...
+	Expires  time.Time
+
+	// EncryptionScheme determines the object's encryption scheme. If not set, uses the Uplink default
+	EncryptionScheme *storj.EncryptionScheme
```
kaloyan-raev: We have this in `BucketConfig.Volatile`: `DataCipher` and `EncryptionBlockSize`, no? So we should remove the `EncryptionScheme` from here.

lib/uplink/apikey.go
jtolds: k fair point, just need to revert 0dc098c then

lib/uplink/bucket.go
jtolds: any reason not to just use the Metadata map for Content-Type? other systems do that

lib/uplink/object.go
jtolds: same. seems like that should just be in the Metadata map

lib/uplink/bucket.go
jtolds: you're right

lib/uplink/apikey.go
thepaul: i'll do that

lib/uplink/bucket.go
kaloyan-raev: My understanding of the `Metadata` map is that it is for used-defined or app/client-defined metadata. ContentType should be first-class metadata.

lib/uplink/uplink.go
```diff
@@ -4,10 +4,84 @@
...
+	maxBucketMetaSize = 10 * memory.MiB
 )
+
+// Config represents configuration options for an Uplink
+type Config struct {
+	// These config values are likely to change semantics or go away
+	// entirely between releases. Be careful when using them!
+	Volatile struct {
+		TLS struct {
+			SkipPeerCAWhitelist bool
```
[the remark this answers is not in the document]
thepaul: Probably not. Wonder why the linter didn't pick up on these? I'll try to fill them in.

lib/uplink/bucket.go
jtolds: even in a scenario where Content Type is treated specially, it's still okay to have it in the same map. many s3 libraries treat object metadata this way, even if it's user defined vs system defined. this is mainly because user defined or system defined, it's one namespace. what would it mean to have a user-defined ContentType that's different from the system defined one?

https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

lib/uplink/bucket.go
thepaul: Have we decided for sure, then, that all objects in a bucket must use the same encryption scheme?

lib/uplink/bucket.go
thepaul: This struct is defined by us, so the `Metadata` map can have whatever scope or meaning we want to give it. This doesn't have to map exactly to the GRPC metadata map.

That said, it doesn't seem like it matters a whole lot one way or the other.

lib/uplink/bucket.go
jtolds: oh, i'm sorry, i read @kaloyan-raev's message wrong. i thought he was saying these should be Volatile fields in this struct, not remove them entirely. i think we aren't certain about how this works yet but we do probably want to let people change some details per object within a bucket.

we can either mark this field volatile here, or we can remove it (for now)

lib/uplink/bucket.go
jtolds: pff whatever this is the most important decisiokay you're right i will stop caring

lib/uplink/project.go
```diff
@@ -0,0 +1,202 @@
...
+			MinThreshold     int
+			RepairThreshold  int
+			SuccessThreshold int
+			MaxThreshold     int
+		}
+
+		MaxBufferMem     memory.Size
+		ErasureShareSize memory.Size
+		SegmentSize      memory.Size
+		MaxInlineSize    memory.Size
```
thepaul: This one surely doesn't belong here, does it? This seems like it should be a decision made by the satellite operator and no one else.

jtolds: oh, the satellite operator definitely decides whether to accept or reject, but this is the hint on the client side of what to do, since the satellite doesn't actually advertise it's setting anywhere. this is in volatile though, right?

thepaul: Okay. Yeah, it is, I'm just trying to sanity check

lib/uplink/bucket.go
kaloyan-raev: My main concern is that it is error-prone to leave this to clients. What should be the exact key for Content Type in the metadata map? `content-type`, `contenttype`, `Content-Type` or `ContentType`? These are all different keys. If two different apps choose two different keys for the Content Type, this breaks the interoperability on the same data.

lib/uplink/bucket.go
kaloyan-raev: The `EncryptionScheme` field here is duplicate to the `DataCipher` and `EncryptionBlockSize` in `BucketConfig.Volatile`.

We have to decide if we want to configure the encryption 1) per bucket (for new uploads only); or 2) per object. But we should not have both ways.

And, I believe, we should have it the same way for the RS config too. If we have encryption config per object, we should also have the RS config per object. It does not make sense to have encryption configurable per object, but RS only per bucket (for new uploads only).

thepaul: I'm not sure I agree. It seems both reasonable and useful to have a default per bucket and also allow overriding those defaults for individual objects. Is your concern about users being confused by that?

lib/uplink/bucket.go
thepaul: Ok, I think everybody is fine with that. I'll add a ContentType field here and in ObjectMeta.

lib/uplink/bucket.go
kaloyan-raev: If we want some config to have bucket's default and still be configurable per object, then the RS config is much more important than the EncryptionScheme.

It's more likely that users would want to increase the durability of specific critical files rather than encrypting different files with different ciphers.

If we want to put this flexibility in the API now then let's do it for both RS and Encryption.

lib/uplink/project.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// BucketConfig represents configuration options for a specific Bucket
+type BucketConfig struct {
+	EncryptionAccess EncryptionAccess
+
+	// These config values are likely to change semantics or go away
+	// entirely between releases. Be careful when using them!
+	Volatile struct {
+		DefaultRS struct {
```
kaloyan-raev: We have the `storj.RedundancyScheme` type that we can use instead of `DefaultRS` and `ErasureShareSize`.

lib/uplink/bucket.go
thepaul: I'm also not sure if that's really more likely, but I have no objection to adding that as an explicit option per object. Let's put both RS and encryption params in a Volatile struct for now?

kaloyan-raev: ok, deal :)

lib/uplink/uplink.go
```diff
@@ -4,10 +4,84 @@
...
+
+// Uplink represents the main entrypoint to Storj V3. An Uplink connects to
+// a specific Satellite and caches connections and resources, allowing one to
+// create sessions delineated by specific access controls.
+type Uplink struct {
+	tc transport.Client
+}
+
+// NewUplink creates a new Uplink
+func NewUplink(ctx context.Context, cfg Config) (*Uplink, error) {
```
dylanlott: How do we plan to handle configs here?

thepaul: good question. maybe the library should be able to read a config file, if asked, like something called `uplink.ReadConfigFromFile()`?

dylanlott: That would work with the cmd util. Do we still want per-bucket configs as a possibility?

thepaul: I don't know. This is difficult for me to judge without seeing several "real life" uses of the library. I want to avoid putting something in here prematurely that doesn't meet user needs in quite the right way, and then end up being required to maintain it for a long time.

It would probably be better to put config handling in the cmd/uplink code for now, and then if we see a need, move it into the library.

lib/uplink/bucket.go
```diff
@@ -0,0 +1,93 @@
...
+	// or go away entirely between releases. Be careful when using them!
+	Volatile struct {
+		// DataCipher determines the ciphersuite to use for the Object's
+		// data encryption. If not set, the Bucket's default will be
+		// used.
+		DataCipher Cipher
+		// EncryptionBlockSize determines the unit size at which
+		// encryption is performed. See BucketConfig.EncryptionBlockSize
+		// for more information.
+		EncryptionBlockSize memory.Size
```
kaloyan-raev: We should use the `storj.EncryptionScheme` struct instead of separate `DataCipher` and `EncryptionBlockSize` fields.

Also in `ObjectMeta` and `BucketConfig`.

thepaul: See the comment on 5fbea44:

> It seems to make sense for each of these to have a RedundancyScheme, a DataCipher, and an EncryptionBlockSize. We could use storj.EncryptionScheme for the latter two, but we want a Cipher type where we can distinguish an unset value (0) from storj.Unencrypted (0). We could just use a *storj.EncryptionScheme and have the nil pointer mean to use defaults, but we would still have the same problem with PathCipher. It doesn't make sense for PathCipher to use storj.EncryptionScheme because it doesn't have a BlockSize. We could use uplink.Cipher for PathCipher and storj.EncryptionScheme for DataCipher, but the mismatch is mentally jarring.

kaloyan-raev: Why don't we just change the existing `storj.Cipher` to have `UnsetCipher` zero value instead of defining a brand new `uplink.Cipher`? It's a backward incompatible change, but we are anyway resetting the data in the complete V3 network with the release we are trying to push right now. So it's the best time to make this change.

lib/uplink/bucket.go
```diff
@@ -0,0 +1,93 @@
...
+
+		// RSParameters determines the Reed-Solomon and/or Forward Error
+		// Correction encoding parameters to be used for this Object.
+		RSParameters storj.RedundancyScheme
```
kaloyan-raev: Shouldn't we have a more generic name for this field? Theoretically `storj.RedundancyScheme` allows other encoding algorithms than Reed Solomon. This is why we came with `RedundancyScheme` as a more generic name for this type.

lib/uplink/bucket.go
thepaul: I'd be entirely ok with that, but we'd have to coordinate it with everybody and everything that might touch that field, and it might take longer. What if we made new `storj.CipherSuite` and `storj.EncryptionParameters` types and use those, and transition other things to using those over time?

lib/uplink/bucket.go
thepaul: That sounds fine. I didn't do that because I wanted to limit my changes, but you're right, it makes more sense.

jtolds: this is in the volatile section, right? but sure

lib/uplink/uplink.go
```diff
@@ -4,10 +4,94 @@
...
+		return nil, err
+	}
+	tlsOpts, err := tlsopts.NewOptions(identity, tlsopts.Config{UsePeerCAWhitelist: !cfg.Volatile.TLS.SkipPeerCAWhitelist})
+	if err != nil {
+		return nil, err
+	}
+	tc := transport.NewClient(tlsOpts)
+
+	return &Uplink{
+		tc: tc,
```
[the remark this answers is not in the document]
thepaul: I think we do. I have a wip that adds that, but want to fix master first

lib/uplink/project.go
```diff
@@ -0,0 +1,204 @@
...
+	metainfo      metainfo.Client
+	project       *kvmetainfo.Project
+	maxInlineSize memory.Size
+}
+
+// BucketConfig holds information about a bucket's configuration. This is
+// filled in by the caller for use with CreateBucket(), or filled in by the
+// library as Bucket.Config when a bucket is returned from OpenBucket().
+type BucketConfig struct {
+	// PathCipher indicates which ciphersuite is to be used for path
```
kaloyan-raev: "ciphersuite" is not a single word, right?

thepaul: It's often used as a single word these days, but yeah, I guess we capitalize it as though it's two words, so I should fix that

thepaul:
Uhhh how about I'll approve it on your behalf

PR #1569, opened 2019-03-25
V3-1307 combine first and last name to full name

crawter:


satellite/console/consoleweb/consoleql/mutation.go
```diff
@@ -283,12 +288,17 @@ func rootMutation(service *console.Service, mailService *mailservice.Service, ty
...
 					for _, user := range users {
+						userName := user.ShortName
+						if user.ShortName == "" {
+							userName = user.FullName
+						}
+
 						err = mailService.SendRendered(
 							p.Context,
-							[]post.Address{{Address: user.Email, Name: fmt.Sprintf("%s %s", user.FirstName, user.LastName)}},
+							[]post.Address{{Address: user.Email, Name: fmt.Sprintf("%s %s", user.FullName, user.ShortName)}},
```
rikysya: What is the meaning of combining Fullname and Shortname?

jtolds: not all cultures have the concept of first name and last name. the only general thing that works well across all cultures is full name and preferred short name

satellite/satellitedb/projectmembers.go
```diff
@@ -48,8 +48,8 @@ func (pm *projectMembers) GetByProjectID(ctx context.Context, projectID uuid.UUI
 				INNER JOIN users u ON pm.member_id = u.id
 					WHERE pm.project_id = ?
 					AND ( u.email LIKE ? OR 
-						  u.first_name || u.last_name LIKE ? OR
-						  u.last_name || u.first_name LIKE ? )
+						  u.full_name LIKE ? OR
```
Barterio: Assume you have 
u.full_name = 'Bohdan Artemenko', 
search_query = 'artemenko'
Will this code return 'Bohdan Artemenko' user?

satellite/console/consoleweb/consoleql/mutation.go
rikysya: `Name: fmt.Sprintf("%s %s", user.FullName, user.ShortName)}}`
I mean for email recipient name it should be one of the above but not both combined

satellite/satellitedb/projectmembers.go
crawter: @Barterio no, we should change this query, we will create a bug in our backlog

PR #1573, opened 2019-03-25
pkg/process: indicate if an error is causing program death

jtolds:
Occasionally, the satellite dies in our production environment, and the last log message is an error we believe didn't kill the process. But maybe it did! We should be more clear about which errors were fatal.

Lack of test justification: just changing log output. Log output is not something we need to test.

## Code Review Checklist

Thanks for submitting a PR!


pkg/process/exec_conf.go
```diff
@@ -240,8 +240,8 @@ func cleanup(cmd *cobra.Command) {
 
 		err = internalRun(cmd, args)
 		if err != nil {
-			_, _ = fmt.Fprintf(os.Stderr, "%v\n", err)
-			logger.Sugar().Debugf("%+v", err)
+			_, _ = fmt.Fprintf(os.Stderr, "Fatal error: %v\n", err)
+			logger.Sugar().Debugf("Fatal error: %+v", err)
```
egonelbre: lower case

PR #1574, opened 2019-03-25
adds test for correct download with non-critical amount of nodes offline

navillasa:
## why changes
The repairer test is flakily passing and failing, so we decided to create `TestDownloadWithSomeNodesOnline` as a "sanity check" type test to preface the repair test.
We found that there was a bug in satellite/metainfo/metainfo.go where the whole download would fail, even if the required number of pieces were still online. This fail was happening because the overlaycache.Get call returned an error, which wasn't properly handled. 

## what changes
- Adds error handling to ensure that the download would only fail if less than the minimum required number of nodes were online.
- Adds `TestDownloadWithSomeNodesOffline` to internal/testplanet/uplink_test.go, which uploads some data, stops some nodes, then downloads to make sure data is the same.

## Code Review Checklist

Thanks for submitting a PR!

### DIFF
--- a/internal/testplanet/uplink_test.go
+++ b/internal/testplanet/uplink_test.go
@@ -106,6 +106,9 @@ func TestDownloadWithSomeNodesOffline(t *testing.T) {
 			if nodesToKill[node.ID()] {
 				err = planet.StopPeer(node)
 				assert.NoError(t, err)
+
+				err = satellite.Overlay.Service.Delete(ctx, node.ID())
+				assert.NoError(t, err)
 			}
 		}
 
--- a/internal/testplanet/uplink_test.go
+++ b/internal/testplanet/uplink_test.go
@@ -13,6 +13,9 @@ import (
 	"storj.io/storj/internal/memory"
 	"storj.io/storj/internal/testcontext"
 	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/uplink"
 )
 
 func TestUploadDownload(t *testing.T) {
@@ -36,4 +39,79 @@ func TestUploadDownload(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Equal(t, expectedData, data)
+}
+
+func TestDownloadWithSomeNodesOffline(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 10, UplinkCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		// first, upload some remote data
+		ul := planet.Uplinks[0]
+		satellite := planet.Satellites[0]
+
+		satellite.Repair.Checker.Loop.Stop()
+
+		testData := make([]byte, 1*memory.MiB)
+		_, err := rand.Read(testData)
+		assert.NoError(t, err)
+
+		err = ul.UploadWithConfig(ctx, satellite, &uplink.RSConfig{
+			MinThreshold:     2,
+			RepairThreshold:  3,
+			SuccessThreshold: 4,
+			MaxThreshold:     5,
+		}, "testbucket", "test/path", testData)
+		require.NoError(t, err)
+
+		// get a remote segment from pointerdb
+		pdb := satellite.Metainfo.Service
+		listResponse, _, err := pdb.List("", "", "", true, 0, 0)
+		require.NoError(t, err)
+
+		var path string
+		var pointer *pb.Pointer
+		for _, v := range listResponse {
+			path = v.GetPath()
+			pointer, err = pdb.Get(path)
+			assert.NoError(t, err)
+			if pointer.GetType() == pb.Pointer_REMOTE {
+				break
+			}
+		}
+
+		// calculate how many storagenodes to kill
+		numStorageNodes := len(planet.StorageNodes)
+		redundancy := pointer.GetRemote().GetRedundancy()
+		remotePieces := pointer.GetRemote().GetRemotePieces()
+		minReq := redundancy.GetMinReq()
+		numPieces := len(remotePieces)
+		toKill := numPieces - int(minReq)
+
+		// we should have enough storage nodes to repair on
+		assert.True(t, (numStorageNodes-toKill) >= numPieces)
+
+		// kill nodes and track lost pieces
+		var lostPieces []int32
+		nodesToKill := make(map[storj.NodeID]bool)
+		nodesToKeepAlive := make(map[storj.NodeID]bool)
+		for i, piece := range remotePieces {
+			if i >= toKill {
+				nodesToKeepAlive[piece.NodeId] = true
+				continue
+			}
+			nodesToKill[piece.NodeId] = true
+			lostPieces = append(lostPieces, piece.GetPieceNum())
+		}
+		for _, node := range planet.StorageNodes {
+			if nodesToKill[node.ID()] {
+				err = planet.StopPeer(node)
+				assert.NoError(t, err)
+			}
+		}
+
+		// we should be able to download data without any of the original nodes
+		newData, err := ul.Download(ctx, satellite, "testbucket", "test/path")
+		assert.NoError(t, err)
+		assert.Equal(t, newData, testData)
+	})
 }
--- a/internal/testplanet/uplink_test.go
+++ b/internal/testplanet/uplink_test.go
@@ -43,13 +43,12 @@ func TestUploadDownload(t *testing.T) {
 
 func TestDownloadWithSomeNodesOffline(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 10, UplinkCount: 1,
+		SatelliteCount: 1, StorageNodeCount: 5, UplinkCount: 1,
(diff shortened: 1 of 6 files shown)


internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,81 @@ func TestUploadDownload(t *testing.T) {
...
+		// calculate how many storagenodes to kill
+		numStorageNodes := len(planet.StorageNodes)
+		redundancy := pointer.GetRemote().GetRedundancy()
+		remotePieces := pointer.GetRemote().GetRemotePieces()
+		minReq := redundancy.GetMinReq()
+		numPieces := len(remotePieces)
+		toKill := numPieces - int(minReq)
+
+		// we should have enough storage nodes to repair on
+		assert.True(t, (numStorageNodes-toKill) >= numPieces)
```
mobyvb: we can get rid of this line

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,81 @@ func TestUploadDownload(t *testing.T) {
 
 	assert.Equal(t, expectedData, data)
 }
+
+func TestDownloadWithSomeNodesOffline(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 10, UplinkCount: 1,
```
mobyvb: we probably only need 5 nodes for this test

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,81 @@ func TestUploadDownload(t *testing.T) {
...
+		remotePieces := pointer.GetRemote().GetRemotePieces()
+		minReq := redundancy.GetMinReq()
+		numPieces := len(remotePieces)
+		toKill := numPieces - int(minReq)
+
+		// we should have enough storage nodes to repair on
+		assert.True(t, (numStorageNodes-toKill) >= numPieces)
+
+		// kill nodes and track lost pieces
+		var lostPieces []int32
```
mobyvb: we do not need to track lost pieces for this test

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,81 @@ func TestUploadDownload(t *testing.T) {
...
+		nodesToKill := make(map[storj.NodeID]bool)
+		nodesToKeepAlive := make(map[storj.NodeID]bool)
```
mobyvb: we do not need to keep track of nodesToKeepAlive for this test

satellite/metainfo/metainfo.go
```diff
@@ -325,7 +326,9 @@ func (endpoint *Endpoint) createOrderLimitsForSegment(ctx context.Context, point
 
 		node, err := endpoint.cache.Get(ctx, piece.NodeId)
 		if err != nil {
-			return nil, err
+			endpoint.log.Error("error getting node from overlay cache")
```
kaloyan-raev: We should log the actual error too, i.e. something like: 
```
endpoint.log.Error("error getting node from overlay cache", zap.Error(err))
```

navillasa: should we log the nodeID too or no?

satellite/metainfo/metainfo.go
```diff
@@ -336,8 +339,13 @@ func (endpoint *Endpoint) createOrderLimitsForSegment(ctx context.Context, point
 			Limit:              orderLimit,
 			StorageNodeAddress: node.Address,
 		})
+	}
 
+	if len(limits) < redundancy.RequiredCount() {
+		err = Error.New("not enough nodes available. Available: %d, Required: %d", len(limits), redundancy.RequiredCount())
```
kaloyan-raev: Maybe change the error message to `"not enough nodes available: got %d, required %d"`

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,72 @@ func TestUploadDownload(t *testing.T) {
...
+		SatelliteCount: 1, StorageNodeCount: 5, UplinkCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+
+		// first, upload some remote data
+		ul := planet.Uplinks[0]
+		satellite := planet.Satellites[0]
+
+		testData := make([]byte, 1*memory.MiB)
+		_, err := rand.Read(testData)
+		assert.NoError(t, err)
```
kaloyan-raev: This should be `require.NoError`.

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,72 @@ func TestUploadDownload(t *testing.T) {
...
+		pdb := satellite.Metainfo.Service
+		listResponse, _, err := pdb.List("", "", "", true, 0, 0)
+		require.NoError(t, err)
+
+		var path string
+		var pointer *pb.Pointer
+		for _, v := range listResponse {
+			path = v.GetPath()
+			pointer, err = pdb.Get(path)
+			assert.NoError(t, err)
```
kaloyan-raev: Use `require.NoError`. Otherwise, it will panic on the next line if there is an error.

satellite/metainfo/metainfo.go
kaloyan-raev: Yes

navillasa: ok we ended up doing this a level down inside satellite/satellitedb/overlaycache.go

satellite/satellitedb/overlaycache.go
```diff
@@ -116,7 +116,7 @@ func (cache *overlaycache) Get(ctx context.Context, id storj.NodeID) (*pb.Node,
 		dbx.OverlayCacheNode_NodeId(id.Bytes()),
 	)
 	if err == sql.ErrNoRows {
-		return nil, overlay.ErrNodeNotFound
+		return nil, errs.Combine(overlay.ErrNodeNotFound, errs.New("couldn't find nodeID: %s", id.String()))
```
kaloyan-raev: `errs.Combine` is for combining several different errors into one. Here we actually want to format a single error. So we should change `overlay.ErrNodeNotFound` to be an error class instead of an error. Then instead of using `errs.Combine`, we would call `overlay.ErrNodeNotFound.New(id.String())`.

See `overlay.ErrNotEnoughNodes` for reference.

navillasa: ah okay got it. fixed now!

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,72 @@ func TestUploadDownload(t *testing.T) {
...
+		listResponse, _, err := pdb.List("", "", "", true, 0, 0)
+		require.NoError(t, err)
+
+		var path string
+		var pointer *pb.Pointer
+		for _, v := range listResponse {
+			path = v.GetPath()
+			pointer, err = pdb.Get(path)
+			require.NoError(t, err)
+			if pointer.GetType() == pb.Pointer_REMOTE {
```
wthorp: What does this do?

internal/testplanet/uplink_test.go
```diff
@@ -37,3 +40,72 @@ func TestUploadDownload(t *testing.T) {
...
+
+		// calculate how many storagenodes to kill
+		redundancy := pointer.GetRemote().GetRedundancy()
+		remotePieces := pointer.GetRemote().GetRemotePieces()
+		minReq := redundancy.GetMinReq()
+		numPieces := len(remotePieces)
+		toKill := numPieces - int(minReq)
+
+		nodesToKill := make(map[storj.NodeID]bool)
+		for i, piece := range remotePieces {
```
wthorp: couldn't you just kill the first toKill nodes rather than making this map?

internal/testplanet/uplink_test.go
navillasa: This makes sure we're getting a pointer to a remote segment (vs. a pointer to an inline segment, which the pdb.Get call could also potentially return). We want a remote segment because that means its pieces will be on storage nodes, and we want to test the download after taking some of these nodes offline.

internal/testplanet/uplink_test.go
navillasa: No, because StopPeer needs to take a planet.Peer as an argument, which we can't access in this loop.

internal/testplanet/uplink_test.go
wthorp: but you're only going to break out of `for _, v := range listResponse {`

navillasa: right, so at that point we'll have `pointer` assigned to a remote pointer

internal/testplanet/uplink_test.go
wthorp: maybe this?
for i:=0; i<toKill; i++{
planet.StopPeer(planet.StorageNodes[i].Identity.ID)

wthorp: are planet.StorageNodes's peers?  EG:
for i:=0; i<toKill; i++{
err = planet.StopPeer(planet.StorageNodes[i])

navillasa: the problem with this is that not all storage nodes in the test get uploaded to, and for this test, we specifically want to kill storage nodes that are storing pieces (as we found from looking in the pointer).

internal/testplanet/uplink_test.go
wthorp: ah. somehow my brain could process all that because it fails if all pointers are inline.  not a bid deal.  LGTM.

PR #1576, opened 2019-03-26
Add Versioning Server

stefanbenten:
Ticket: https://storjlabs.atlassian.net/browse/V3-1218 

This PR add a versioning server (versioncontrol) and adds the necessary client side services which integrate into satellite, bootstrap and storagenode. It will ensure that the client software version is up to date and restrict startup for outdated versions.

## Code Review Checklist

Thanks for submitting a PR!


stefanbenten:
@jtolds I can change the flag system to take in multiple sets and return all of them in a list of Version.Info.
Is that fine?

jtolds:
using a flag for a list is fine. comma separated or something. the info struct has stuff that's really more useful to each program directly (e.g. the storage node) and not really useful on a version server, such as release (will always be true coming from the version server if it sends it), and timestamp, hash, and version are all pretty much interchangeable, so we really only need one.

more concretely: cmd/storagenode/main.go should know timestamp, hash, version, release for itself, but then all the version server needs to tell it is if it's okay to run. the version server only needs to send version. or maybe hash. you pick. but it doesn't need to send all 4

stefanbenten:
Okay cool, let me change that to just send the commit hashes and take those as flags for the versioning-server.

stefanbenten:
The example for checking the version is made in .cmd/storagenode/main.go.

cmd/version-control/main.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+func main() {
+	// Flags to specify required Version
+	addr := flag.String("listen", "0.0.0.0:8080", "Defines Listen Address of Webserver")
+	versions := flag.String("version", "v0.1.0,v0.1.1", "Comma separated list of Versions")
+	flag.Parse()
+
+	if flag.Parsed() {
```
brimstone: For readability, maybe invert this check and cause a reasonable `log.Fatal` including why, then pop out the rest of the conditional.

cmd/version-control/main.go
```diff
@@ -0,0 +1,68 @@
...
+		for _, subversion := range subversions {
+			instance := version.Info{
+				Version: subversion,
+			}
+			ver = append(ver, instance)
+		}
+
+		log.Printf("setting version info to: %v", ver)
+		http.HandleFunc("/", handleGet)
+		log.Println("starting Webserver")
```
brimstone: I like this, lets me know it's good to go.

cmd/version-control/main.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Only handle GET Requests
+	if r.Method == "GET" {
+		if xfor = r.Header.Get("X-Forwarded-For"); xfor == "" {
+			xfor = r.RemoteAddr
+		}
+		log.Printf("Request from: %s for %s", r.RemoteAddr, xfor)
+
+		w.Header().Set("Content-Type", "application/json")
+		response, err := json.Marshal(ver)
```
brimstone: If the version isn't ever going to change, and we want to go fast, we should only Marshal this once at start. json and go is notoriously slow.

cmd/version-control/main.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			w.WriteHeader(500)
```
brimstone: You should definitely return here and not continue writing to the writer.

cmd/version-control/main.go
jtolds: :+1:

cmd/version-control/main.go
jtolds: that's a fair point. :+1:

cmd/version-control/main.go
jtolds: you can also do http.Error(w, err.Error(), 500) if you want to provide more info, but yeah you should return afterwards

stefanbenten: Damn it, thanks!
Totally forgot to return there

cmd/version-control/main.go
stefanbenten: Done

cmd/version-control/main.go
stefanbenten: Done

cmd/version-control/config.json.example
```diff
@@ -0,0 +1,4 @@
+{
+  "listen":"0.0.0.0:8080",
+  "versions":"v0.1.0,v0.1.1,v0.2.0,v0.2.1"
```
brimstone: A common separated list in a string here is kinda weird, but i guess it's ok.

cmd/version-control/main.go
```diff
@@ -0,0 +1,117 @@
...
+			log.Fatalf("Could not read configuation file: %s", *fconfig)
+		}
+		err = json.Unmarshal(bytes, &conf)
+		if err != nil {
+			log.Fatalf("Could not parse configuation file: %s %v", *fconfig, err)
+		}
+	} else {
+		conf.Listen = *addr
+		conf.Versions = *fversions
+	}
```
jtolds: this file is good though you get almost all of this config file and log setup, etc for free if you use pkg/process. but yeah we can do that later

cmd/version-control/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+	subVersions := strings.Split(conf.Versions, ",")
```
jtolds: how do you feel about using commit hashes instead of versions? the more i've thought about it the less likely i think we'll be to shoot ourselves in the foot that way during alpha, even though you can't compare if a hash is newer

cmd/version-control/main.go
```diff
@@ -0,0 +1,117 @@
...
+	ver      []version.V
+	response []byte
+)
+
+func handleGet(w http.ResponseWriter, r *http.Request) {
+	var xfor string
+
+	// Only handle GET Requests
+	if r.Method == "GET" {
+		if xfor = r.Header.Get("X-Forwarded-For"); xfor == "" {
```
jtolds: heh nice

internal/version/version.go
```diff
@@ -0,0 +1,48 @@
...
+func (v V) Marshal() (data []byte, err error) {
+	data, err = json.Marshal(v)
+	return
+}
+
+// QueryVersionFromControlServer handles the HTTP request to gather the allowed and latest version information
+func QueryVersionFromControlServer() (ver []V, err error) {
+	resp, err := http.Get("https://version.alpha.storj.io")
+	if err != nil {
+		return []V{}, err
```
jtolds: you can do `return nil, err` here and below

internal/version/version.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return []V{}, err
+	}
+
+	err = json.Unmarshal(body, ver)
```
jtolds: this is a bit unusual idiom-wise. i think you'd normally just see `return json.Unmarshal(body, ver)`. i wish they hadn't put unnamed returns in go. they consistently end up after refactors with unintended consequences.

cmd/version-control/main.go
jtolds: versions are fine though

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// Config contains the necessary Information to check the Software Version
+type Config struct {
+	ServerAddress  string        `help:"server address to check its version against" default:"https://satellite.stefan-benten.de/version"`
```
stefanbenten: This needs to be changes still. @jtolds 
I have this server running for testing

internal/version/version.go
```diff
@@ -0,0 +1,177 @@
...
+	"strconv"
+	"strings"
+
+	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+)
+
+var (
+	mon = monkit.Package()
+	// Timestamp is the UTC timestamp of the compilation time
+	Timestamp string
```
jtolds: Timestamp, CommitHash, Version, and Release should all be lowercased, not exported, and it turns out must all be strings including release. Only Build should be exported (first letter capital).

I propose we do buildTimestamp, buildCommitHash, buildVersion, and buildRelease, and make sure to update the shell script

internal/version/version.go
```diff
@@ -0,0 +1,177 @@
...
+	Version = "v0.1.0"
+	// Release indicates whether the binary compiled is a release candidate
+	Release bool
+	// Build is a struct containing all relevant build information associated with the binary
+	Build Info
+)
+
+// Info is the versioning information for a binary
+type Info struct {
+	Timestamp  string `json:"timestamp,omitempty"`
```
jtolds: this should be a time.Time

stefanbenten: Cant we use an UNIX Timestamp better here?
Handling them in the protobuf is also easier i would suppose

internal/version/version.go
```diff
@@ -0,0 +1,177 @@
...
+	Storagenode []SemVer
+	Uplink      []SemVer
+	Gateway     []SemVer
+}
+
+// SemVerRegex is the regular expression used to parse a semantic version.
+// https://github.com/Masterminds/semver/blob/master/LICENSE.txt
+const SemVerRegex string = `v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?` +
+	`(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?` +
+	`(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?`
```
jtolds: this regex doesn't match what we're actually parsing below, which is only `v?([0-9]+)\.([0-9]+)\.([0-9]+)`

internal/version/version.go
stefanbenten: Let me do that 👍

internal/version/version.go
stefanbenten: Thats a job for @coyle

jtolds: the A-Za-z bits in particular are especially suspicious

internal/version/version.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Marshal converts the existing Version Info to any json byte array
+func (v Info) Marshal() (data []byte, err error) {
+	data, err = json.Marshal(v)
+	return
+}
+
+// parseToInt64 converts a string with schema .xxx to an int64 or returns an error
+func parseToInt64(label string) (int64, error) {
+	tmp := strings.TrimPrefix(label, ".")
```
jtolds: you could just use strconv.ParseInt directly instead of this function if you use the simplified regex

internal/version/version.go
```diff
@@ -0,0 +1,177 @@
...
+		if err != nil {
+			return nil, err
+		}
+		versions = append(versions, *sVer)
+	}
+	return versions, err
+}
+
+func init() {
+	if Version == "" {
```
jtolds: we should not return if any of the linker flag values are set. we should only return if all of them are unset

scripts/release.sh
```diff
@@ -0,0 +1,29 @@
...
+ echo -n "Tagged version: "
+VERSION=$(git describe --tags --match "v[0-9]*.[0-9]*.[0-9]*")
+echo $VERSION
+
+ echo Running "go $@"
+exec go "$1" -ldflags \
+	"-X storj.io/storj/internal/version.Timestamp=$TIMESTAMP
+         -X storj.io/storj/internal/version.CommitHash=$COMMIT
+         -X storj.io/storj/internal/version.Version=$VERSION
+         -X storj.io/storj/internal/version.Release=$RELEASE" "${@:2}"
```
jtolds: these should be updated to set the unexported values (buildTimestamp, buildCommitHash, etc)

scripts/release.sh
```diff
@@ -0,0 +1,29 @@
...
+  COMMIT=$(git rev-parse HEAD)-dirty
+  RELEASE=false
+else
+  COMMIT=$(git rev-parse HEAD)
+  RELEASE=true
+fi
+echo $COMMIT
+
+ echo -n "Tagged version: "
+VERSION=$(git describe --tags --match "v[0-9]*.[0-9]*.[0-9]*")
```
jtolds: this is an even simpler semantic version matcher. not a regex, but i think we can safely be more strict in the code because of this

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+		allowed: false,
+	}
+}
+
+// Run logs the current version information
+func (srv *Service) Run(ctx context.Context) error {
+	firstCheck := true
+	return srv.Loop.Run(ctx, func(ctx context.Context) error {
+		var err error
+		allowed, err := srv.checkVersion(ctx)
```
jtolds: gotta check the error here

jtolds: very important that if we have any errors, we let the process keep running

jtolds: reasonable go standards are to assume `allowed` could be anything (probably false) if `err != nil`

stefanbenten: Check happens in Line 71 and is just a simple error log message

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return srv.allowed
+}
+
+// CheckVersion checks if the client is running latest/allowed code
+func (srv *Service) checkVersion(ctx context.Context) (allowed bool, err error) {
+	defer mon.Task()(&ctx)(&err)
+	accepted, err := srv.queryVersionFromControlServer(ctx)
+	if err != nil {
+		return true, err
```
jtolds: okay so it's safe to do, but still surprising. my vote is to leave the determination of what to do in an error case to the caller

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+		return true, errs.New("Empty List from Versioning Server")
+	}
+	if containsVersion(list, srv.info.Version) {
+		zap.S().Infof("running on version %s", srv.info.Version.String())
+		allowed = true
+	} else {
+		zap.S().Errorf("running on not allowed/outdated version %s", srv.info.Version.String())
+		allowed = false
+	}
+	return
```
jtolds: no naked returns. please return the value explicitly

internal/version/version.go
stefanbenten: Youre right, i didnt check that at all

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return AllowedVersions{}, err
+	}
+
+	defer func() { _ = resp.Body.Close() }()
+
+	err = json.NewDecoder(resp.Body).Decode(&ver)
+	return
```
jtolds: same. please return the values

internal/version/service.go
```diff
@@ -0,0 +1,163 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	_, err = w.Write(j)
+	if err != nil {
+		zap.S().Errorf("error writing data to client %v", err)
+	}
+}
+
+func getFieldString(array *AllowedVersions, field string) []SemVer {
```
jtolds: using reflect is very eyebrow raising. okay for now but definitely prefer another way

internal/version/client.go
```diff
@@ -0,0 +1,67 @@
...
+	"google.golang.org/grpc"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/transport"
+)
+
+// VersionedClient cointains and managers a transport client, that internally checks if the software version
+// is still up to date and refuses outgoing connections in case it is outdated
+type VersionedClient struct {
```
jtolds: i'd like to take this out of this PR if that's okay

stefanbenten: Will do that 👍

internal/version/service.go
stefanbenten: Do you have a better way to take the correct struct list?
I did not found a good solution for that and a map instead of the struct feels bad.

I thought about adding a function to the struct somehow

jtolds: i think parsing the json as a map is the most expected way tbh

internal/version/version.go
stefanbenten: This cannot be done. I changed it to now set a default version of v0.0.1
Else Storj-Sim is unable to run unless we let it build binaries always with these flags.

internal/version/service.go
stefanbenten: Not sure how to go about this now. Its it fine to leave and improve later?

internal/version/version.go
coyle: I'm not understanding how this is different

stefanbenten: One is checking for Chars as well, the one from JT is strictly just Numbers, which makes parsing easier as well

cmd/versioncontrol/main.go
```diff
@@ -0,0 +1,97 @@
...
+	rootCmd.AddCommand(setupCmd)
+	cfgstruct.Bind(runCmd.Flags(), &runCfg, isDev, cfgstruct.ConfDir(confDir))
+	cfgstruct.BindSetup(setupCmd.Flags(), &setupCfg, isDev, cfgstruct.ConfDir(confDir))
+}
+
+func cmdRun(cmd *cobra.Command, args []string) (err error) {
+	log := zap.L()
+	controlserver, err := versioncontrol.New(log, &runCfg)
+	if err != nil {
+		return
```
jtolds: do we really not have a linter for naked returns? linting isn't failing here?

cmd/versioncontrol/main.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	ctx := process.Ctx(cmd)
+	err = controlserver.Run(ctx)
+	return
```
jtolds: or here?

internal/testplanet/planet.go
```diff
@@ -660,6 +683,64 @@ func (planet *Planet) newBootstrap() (peer *bootstrap.Peer, err error) {
 	return peer, nil
 }
 
+// newVersionControlServer initializes the Versioning Server
+func (planet *Planet) newVersionControlServer() (peer *versioncontrol.Peer, err error) {
+	// TODO: move into separate file
+	/*defer func() {
+		planet.peers = append(planet.peers, closablePeer{peer: peer})
+	}()*/
```
jtolds: definitely prefer not checking in commented-out code

internal/version/service.go
jtolds: k

internal/version/service.go
```diff
@@ -0,0 +1,168 @@
...
+
+	Loop *sync2.Cycle
+
+	checked chan struct{}
+	mu      sync.Mutex
+	allowed bool
+}
+
+// NewService creates a Version Check Client with default configuration
+func NewService(config *Config, info *Info, service string) (client *Service) {
```
jtolds: can we have Config and Info not be pointers? it's okay if they're copies

internal/version/service.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Run logs the current version information
+func (srv *Service) Run(ctx context.Context) error {
+	firstCheck := true
+	return srv.Loop.Run(ctx, func(ctx context.Context) error {
+		var err error
+		allowed, err := srv.checkVersion(ctx)
+		if err != nil {
+			// Log about the error, but dont crash the service and allow further operation
```
jtolds: perfect

internal/version/service.go
jtolds: okay

internal/version/version.go
stefanbenten: I am storing the Timestamp on the satellite as an INT64 as well. Makes comparison easier imo

jtolds: time.Time has a Unix method, so I'd still prefer a time.Time as the type here. if you want it to serialize to json as unix timestamp you can do:

```
type Info struct {
  Timestamp Time `json:"timestamp,omitempty"`
  ...
}

type Time time.Time

func (t Time) MarshalJSON() ([]byte, error) {
   return []byte(fmt.Sprint(t.Unix())), nil
}
```
to override json serialization

internal/version/version.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			return nil, err
+		}
+		versions = append(versions, *sVer)
+	}
+	return versions, err
+}
+
+func init() {
+	if buildVersion == "" || buildTimestamp == "" || buildCommitHash == "" {
```
jtolds: also buildRelease == ""

internal/version/version.go
```diff
@@ -0,0 +1,167 @@
...
+		return
+	}
+	timestamp, err := strconv.ParseInt(buildTimestamp, 10, 64)
+	if err != nil {
+		return
+	}
+	Build = Info{
+		Timestamp:  time.Unix(timestamp, 0),
+		CommitHash: buildCommitHash,
+		Release:    buildRelease,
```
jtolds: `Release: strings.ToLower(buildRelease) == "true",`

coyle: buildRelease is a bool

jtolds: it should not be. linker flags can only set strings

internal/version/version.go
coyle: do we not want versioning information for versions running in dev? might be useful to know what's there

jtolds: oh, i misread the if. these should all be joined with &&s and not ||s. the point is that if the release script wasn't used, we shouldn't do this init function. if any of the linker flags are set, this function should run.

internal/version/version.go
coyle: Can linker flags be private ?

jtolds: yep!

PR #1578, opened 2019-03-26
Initial Migration Draft to wipe the network gracefully

stefanbenten:
Ties in to V3-1440. It resets the network gracefully by removing all data from the storagenodes.

## Code Review Checklist

Thanks for submitting a PR!


pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -205,6 +205,13 @@ func (db *DB) Migration() *migrate.Migration {
 					`CREATE INDEX IF NOT EXISTS idx_bwa_serial ON bandwidth_agreements (serial_num)`,
 				},
 			},
+			{
+				Description: "Initiate Network reset",
+				Version:     4,
+				Action: migrate.SQL{
+					`UPDATE ttl SET expires = '1553641200' WHERE created <= '1553641200' `,
```
jtolds: can we have created be something in april? expires is fine. also aren't these numbers and not strings? we shouldn't have the quotes, right?

satellite/satellitedb/migrate.go
```diff
@@ -362,6 +362,13 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				Description: "Initiate Network reset",
+				Version:     9,
+				Action: migrate.SQL{
+					`DELETE FROM PATHDATA`,
```
jtolds: is this the only table we want to clear? okay

pkg/piecestore/psserver/psdb/psdb.go
stefanbenten: I was expecting of SQL handling the cast, but i will double check 👍

satellite/satellitedb/migrate.go
stefanbenten: If all pointers are gone, we should be fine afaik.
All other services are depending on them and i didnt want to remove any billing information.

Main issue is, that i cannot run that query on the unit tests, as they arent aware of the table..

PR #1580, opened 2019-03-26
Enable uplink TestDownloadWithSomeNodesOffline

mobyvb:
largely just removes t.Skip("flaky") now that overlay again determines if nodes are online


navillasa:
I _think_ the problem might be coming from here: https://github.com/storj/storj/blob/master/pkg/eestream/stripe.go#L97

PR #1582, opened 2019-03-26
Data repair integration test

mobyvb:
https://storjlabs.atlassian.net/browse/V3-547

## why changes
- Previously there was no integration test to confirm that data repair was working as expected.
- There was a bug where known offline nodes were saved to a segment's new pointer after the segment was repaired. This was due to the deprecated expectation that all nodes found in overlay cache are online. 

## what changes
- Now `node.Online()` is confirmed before considering a node to be a `healthyPiece`.
- New TestDataRepair located at pkg/datarepair/datarepair_test.go which works as following:
1. Data is uploaded with RS config:
```
			MinThreshold:     2,
			RepairThreshold:  3,
			SuccessThreshold: 4,
			MaxThreshold:     4,
```
2. A remote segment pointer is acquired from the pointerdb.
3. The number of nodesToKill is calculated: `len(remotePieces) - minThreshold`. That amount of nodes are then taken offline. This leaves the minimum number of nodes required to repair the data online.
4. The repair checker and repairer loops happen.
5. To ensure repair worked, the 2 remaining online nodes who were uploaded to in Step 1 are now taken offline.
6. We attempt to download the data from Step 1, and confirm that the data is the same, despite residing on different nodes than initially uploaded to.
7. We confirm that the new pointer does not contain any of the nodes that were taken offline.

## Code Review Checklist

Thanks for submitting a PR!


pkg/storage/segments/repairer.go
```diff
@@ -29,6 +30,8 @@ type Repairer struct {
 	signer     signing.Signer
 	identity   *identity.FullIdentity
 	timeout    time.Duration
+	inProgress map[storj.Path]bool
```
egonelbre: Since we will have multiple servers this won't be sufficient.

navillasa: what do you think would be better to make sure that the same injured segment doesn't have a 
repair job picked up multiple times?

egonelbre: Not quite sure how to do it easily. There are options of just letting it be repaired twice or before repairing verify that things haven't changed with the pointer. That of course still might race.

pkg/datarepair/datarepair_test.go
```diff
@@ -0,0 +1,121 @@
...
+			lostPieces = append(lostPieces, piece.GetPieceNum())
+		}
+		for _, node := range planet.StorageNodes {
+			if nodesToKill[node.ID()] {
+				err = planet.StopPeer(node)
+				assert.NoError(t, err)
+			}
+		}
+
+		time.Sleep(2 * time.Second)
```
jtolds: is there anything we can do to avoid this sleep? does testplanet have a way to wait for the condition we really care about here?

mniewrzal: Most probably satellite `satellite.Repair.Checker.Loop` with methods `Pause()` and `TriggerWait()` can be used. With those methods we can control when specific operation will be triggered. Also `Loop` needs to be added to `repairer.Service`.

Example: https://github.com/storj/storj/blob/b05cf05649d053360f22bd1e8dac7da91036bac5/storagenode/monitor/monitor_test.go#L31

pkg/storage/segments/repairer.go
```diff
@@ -21,30 +22,46 @@ import (
...
 
+	if repairer.inProgress[path] {
+		return nil
+	}
+	repairer.mutex.Lock()
+	repairer.inProgress[path] = true
+	repairer.mutex.Unlock()
+	defer func() {
+		repairer.mutex.Lock()
+		repairer.inProgress[path] = false
```
wthorp: delete(repairer.inProgress, path) might be slightly better here

pkg/datarepair/datarepair_test.go
```diff
@@ -0,0 +1,134 @@
...
+				err = planet.StopPeer(node)
+				assert.NoError(t, err)
+
+				err = satellite.Overlay.Service.Delete(ctx, node.ID())
+				assert.NoError(t, err)
+			}
+		}
+
+		satellite.Repair.Checker.Loop.Restart()
+		satellite.Repair.Checker.Loop.TriggerWait()
```
mniewrzal: `TriggerWait` should be enough here, just remove interval in milliseconds from the beginning.

pkg/storage/segments/repairer.go
mobyvb: Should we make a ticket for this and fix it in a different PR? Since it does not break the test

pkg/datarepair/datarepair_test.go
mobyvb: I added `Loop` to `repairer.Service`, but `TriggerWait` does not wait for the repairer to finish. I will try to see if I can implement `Loop` differently to get around this issue

mniewrzal: Yeah, the problem is that repairing is starting work in goroutine and triggerWait is waiting only to finish `process` method.
https://github.com/storj/storj/blob/5cd7536cb689b243bc55dffb5e0cca72de9e75c4/pkg/datarepair/repairer/repairer.go#L101

mobyvb: Fixed by exposing limiter and using `Limiter.Wait()`

mobyvb: Fixed by exposing limiter and using its `.Wait()`

PR #1584, opened 2019-03-27
Updates accounting tables models and migration

jenlij:
1. Changes “x_storage_rollup” to “x_storage__tally” to conform to existing accounting architecture
2. Removes interval_seconds from bucket_storage_tallies, storagenode_storage_tallies and (redundant information of interval_start)
3. Adds these fields bucket_storage_tallies table so we can remove the bucketUsage model eventually:
    field  remote_segments_count    uint
    field  inline_segments_count    uint
    field  object_count            uint
    field  metadata_size      uint64 

## Code Review Checklist

Thanks for submitting a PR!


pkg/accounting/bucketusage.go
```diff
@@ -10,11 +10,11 @@ import (
 	"github.com/skyrings/skyring-common/tools/uuid"
 )
 
-// BucketUsage is bucket usage rollup repository
+// BucketUsage is bucket usage tally repository
 type BucketUsage interface {
```
cam-a: You weren't there for this conversation, but I'm pretty sure `bucket_storage_tally` and `bucket_bandwith_tally` are replacing this table. However, I'm wondering... We're still going to need to roll up the `bucket_storage_tally` info somewhere, maybe this table could be it?

satellite/satellitedb/migrate.go
```diff
@@ -380,54 +380,52 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 						storage_node_id bytea NOT NULL,
 						PRIMARY KEY ( serial_number_id, storage_node_id )
 					)`,
-					`CREATE TABLE storagenode_bandwidth_rollups (
```
cam-a: I don't know much about migrations, but I think the changes should be in a new version?

satellite/satellitedb/testdata/postgres.v9.sql
```diff
@@ -25,28 +25,30 @@ CREATE TABLE accounting_timestamps (
...
+CREATE TABLE bucket_storage_tallies (
 	bucket_id bytea NOT NULL,
 	interval_start timestamp NOT NULL,
-	interval_seconds integer NOT NULL,
 	inline bigint NOT NULL,
 	remote bigint NOT NULL,
+	remote_segments integer NOT NULL,
+  inline_segments integer NOT NULL,
+  objects integer NOT NULL,
+  metadata_size bigint NOT NULL,
```
JessicaGreben: this indentation is off

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -426,60 +426,61 @@ delete serial_number (
...
+  field  inline_segments    uint
+  field  objects            uint
+  field  metadata_size      uint64
 )
 
-// --- storage node roll up tables --- //
+// --- storage node tally tables --- //
 
-model storagenode_bandwidth_rollup (
+model storagenode_bandwidth_tally (
```
egonelbre: This isn't a tally, it's a rollup, there's a difference.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -426,60 +426,61 @@ delete serial_number (
...
 	index (
-	    name storagenode_id_interval_start_interval_seconds
-	    fields storagenode_id interval_start interval_seconds
+	    name storagenode_id_interval_start
+	    fields storagenode_id interval_start
 	)
 
 	field storagenode_id   blob
 	field interval_start   utimestamp
-	field interval_seconds uint
```
egonelbre: This is not redundant, I wasn't able to find to avoid this.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -426,60 +426,61 @@ delete serial_number (
 // for preventing duplicate serial numbers
 create used_serial ()
 
-// --- bucket roll up tables --- //
+// --- bucket tally tables --- //
 
-model bucket_bandwidth_rollup (
+model bucket_bandwidth_tally (
```
egonelbre: This isn't a tally, it's a rollup.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -426,60 +426,61 @@ delete serial_number (
...
 	index (
-	    name bucket_id_interval_start_interval_seconds
-	    fields bucket_id interval_start interval_seconds
+	    name bucket_id_interval_start
+	    fields bucket_id interval_start
 	)
 
 	field bucket_id        blob
 	field interval_start   utimestamp
-	field interval_seconds uint
```
egonelbre: This isn't redundant.

satellite/satellitedb/dbx/satellitedb.dbx.postgres.sql
```diff
@@ -26,28 +26,30 @@ CREATE TABLE accounting_timestamps (
...
+CREATE TABLE bucket_storage_tallies (
 	bucket_id bytea NOT NULL,
 	interval_start timestamp NOT NULL,
-	interval_seconds integer NOT NULL,
 	inline bigint NOT NULL,
 	remote bigint NOT NULL,
+	remote_segments integer NOT NULL,
+	inline_segments integer NOT NULL,
+	objects integer NOT NULL,
+	metadata_size bigint NOT NULL,
```
egonelbre: Maybe let's use `count` in their name to be clearer on what they mean.

satellite/satellitedb/dbx/satellitedb.dbx.postgres.sql
```diff
@@ -26,28 +26,30 @@ CREATE TABLE accounting_timestamps (
...
-CREATE TABLE bucket_storage_rollups (
+CREATE TABLE bucket_storage_tallies (
 	bucket_id bytea NOT NULL,
 	interval_start timestamp NOT NULL,
-	interval_seconds integer NOT NULL,
 	inline bigint NOT NULL,
 	remote bigint NOT NULL,
+	remote_segments integer NOT NULL,
+	inline_segments integer NOT NULL,
+	objects integer NOT NULL,
```
egonelbre: Let's use `count` in their name to be clearer on what they mean.

satellite/satellitedb/migrate.go
```diff
@@ -380,54 +380,52 @@ func (db *DB) PostgresMigration() *migrate.Migration {
...
+					`CREATE TABLE storagenode_bandwidth_tallies (
```
egonelbre: Existing migrations must not be changed unless we have made an explicit exception.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -426,60 +426,61 @@ delete serial_number (
...
 	field action           uint
 
 	field allocated uint64
 	field settled   uint64
 )
 
-model storagenode_storage_rollup (
+model storagenode_storage_tally (
```
egonelbre: Ideally this wouldn't be a tally. :/

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: could you please explain why you think it is a rollup and not a tally?

We decided on this change because of the functionality of the existing tally and rollup services.  The existing tally service counts up the occurrences at some interval and then the rollup service rolls up those counts into 1 day rollups.

Now with all the recent changes, the SA is incrementing the bucket_bandwidth_rollup/tally records when it receives the orders from the SN. That is replacing the functionality of the existing tally service. So that is why we changed the name from rollup to tally.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: Could you please explain why you weren't able to avoid this?

It seems like you could just calculate the duration of the interval from the `interval_start` of the previous record.

satellite/satellitedb/migrate.go
JessicaGreben: this migration isn't in production yet, so we figured we would just modify the existing migration so that we don't have so many.

egonelbre: Sure, then it's fine.

satellite/satellitedb/migrate.go
```diff
@@ -441,6 +439,24 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					 ALTER TABLE users RENAME COLUMN last_name TO short_name;`,
 				},
 			},
+			{
+				Description: "drops interval seconds from storage_rollups, renames x_storage_rollups to x_storage_tallies, adds fields to bucket_storage_tallies",
```
egonelbre: Describe at a higher level, this level of detail probably doesn't explain too much for the end-user.

PR #1585, opened 2019-03-27
implements internal Semantic Version library

coyle:
## Code Review Checklist

Thanks for submitting a PR!


cmd/bootstrap/main.go
```diff
@@ -78,6 +78,10 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Error("Failed to initialize telemetry batcher: ", err)
 	}
 
+	if err := process.LogAndReportVersion(ctx); err != nil {
```
jtolds: why not have this as part of process.Exec? then we get it automatically everywhere

cmd/release/main.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2019 Storj Labs, Inc.
```
jtolds: i think this command (go-release) could be a bash script? 

```
#!/bin/bash
TIMESTAMP=$(date -Is)
COMMIT=$(git rev-parse HEAD)
VERSION=$(git describe --tags --match “v[0-9]*.[0-9]*.[0-9]*”)
gocommand=$1
shift
exec go "$gocommand" -ldflags "-X storj.io/storj/internal/version.Timestamp=$TIMESTAMP -X storj.io/storj/internal/version.CommitHash=$COMMIT -X storj.io/storj/internal/version.Version=$VERSION storj.io/storj/internal/version.Release=true" "$@"
```

internal/version/version.go
```diff
@@ -5,20 +5,96 @@ package version
...
-// It's goal is to limit the startup up outdated binaries as well as having an easier way to ensure the software
-// is on the latest version from a UI/UX perspective
+var (
+	// Timestamp is the UTC timestamp of the compilation time
+	Timestamp string
+	// CommitHash is the git hash of the code being compiled
+	CommitHash string
+	// Version is the semantic version set at compilation
+	// if not a valid semantic version Release should be false
+	Version = "v0.0.1"
```
jtolds: i think it's okay to leave this empty if go-release wasn't used

internal/version/version.go
```diff
@@ -5,20 +5,96 @@ package version
...
+	// Release indicates whether the binary compiled is a release candidate
+	Release bool
+	// Build is a struct containing all relevant build information associated with the binary
+	Build V
```
jtolds: wait if we have this then we shouldn't have the others. we'll just need to update the linker flag. no reason to have both spots

internal/version/version.go
```diff
@@ -5,20 +5,96 @@ package version
...
+	// Accepted is a list of accepted Version that we got from the control server
+	Accepted []V
```
jtolds: this doesn't need to be global

internal/version/version.go
jtolds: oh actually, i get it, V is parsed. let's have some sort of method that returns the Build from the go-release set values instead of a global

jtolds: oh okay okay, init(). okay carry on, ignore me.

internal/version/version.go
stefanbenten: Its more a temporary solution to run the tools 👍
will be reset to `Version string` once its working

pkg/process/version.go
```diff
@@ -0,0 +1,31 @@
...
+
+func checkVersion(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	ticker := time.NewTicker(1 * time.Hour)
+	defer ticker.Stop()
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-ticker.C:
```
jtolds: wait, check in an hour?

internal/version/version.go
stefanbenten: I am not 100% why we have that. Maybe @coyle has more explanation on that

cmd/release/main.go
stefanbenten: I can put those commands into this file and have unified 👍

jtolds:
oh one more thought for both you and @stefanbenten: perhaps the version server should have different lists of supported versions per application. like maybe we allow a big list of versions for the identity tool, a smaller list for the storage node, and a smaller list for bootstrap/satellite?

internal/version/version.go
jtolds: okay i guess i should say - i think i would prefer non-release binaries to not have a version at all

pkg/process/version.go
stefanbenten: Can be adjusted, is more of a first draft for @coyle

internal/version/version.go
jtolds: new thought to make it an API with less repetition: do the linker commands work with private values?  maybe we can make the globals release/timestamp/commithash, etc all unexported, and then init uses them to fill in the exported Build value?

cmd/release/main.go
jtolds: maybe instead of that with the `shift` it would be better to do

```
#!/bin/bash
TIMESTAMP=$(date -Is)
COMMIT=$(git rev-parse HEAD)
VERSION=$(git describe --tags --match “v[0-9]*.[0-9]*.[0-9]*”)
gocommand="$1"
exec go "$gocommand" -ldflags "-X storj.io/storj/internal/version.Timestamp=$TIMESTAMP -X storj.io/storj/internal/version.CommitHash=$COMMIT -X storj.io/storj/internal/version.Version=$VERSION storj.io/storj/internal/version.Release=true" "${@:2}"
```
and then make sure to run it through shellcheck lol

internal/version/version.go
stefanbenten: Yeah thats what i wanted to point out with that. Currently i set it as default to let the binaries start

cmd/bootstrap/main.go
coyle: Started to do it but changed because:
1) It seemed to magical. We already have a problem with people trying to debug what's happening, I think a little repeating code in different binaries for clarity is a good thing. 
2) If a satellite running 1.0.1 has a critical bug and we want to patch and fix, we release a new version 1.0.2, if we remove 1.0.1 as an acceptable version and add 1.0.2, we force every binary, to update to a new binary that doesn't impact them at all. If that ever happens we will then need to split out the config and I think having it explicit in each binary is better.

cmd/release/main.go
coyle: Are you saying instead of a release go binary we do it via shell? I'm 100% behind that idea if so

pkg/process/version.go
coyle: yes I just posted that as an example in slack

cmd/bootstrap/main.go
coyle: haha ok I see you made the same conclusion about breaking versions up

cmd/storagenode/dashboard.go
```diff
@@ -80,7 +81,7 @@ func printDashboard(data *pb.DashboardResponse) error {
 	color.NoColor = !useColor
 
 	heading := color.New(color.FgGreen, color.Bold)
-	_, _ = heading.Printf("\nStorage Node Dashboard\n")
+	_, _ = heading.Printf("\nStorage Node Dashboard ( Node Version: %s)\n", version.Version)
```
coyle: we should add a `String()` method on Version

internal/version/version.go
coyle: all the examples I've seen show Global/exported but worth a shot

pkg/process/version.go
```diff
@@ -12,20 +12,19 @@ import (
...
-}
-
-func checkVersion(ctx context.Context) (err error) {
+func LogAndReportVersion(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	ticker := time.NewTicker(1 * time.Hour)
+	err = version.CheckVersion()
+
+	ticker := time.NewTicker(15 * time.Minute)
```
stefanbenten: @coyle Shouldnt we fire this entire checker up in a seperate go routine?

pkg/process/exec.go
```diff
@@ -5,10 +5,12 @@ package process // import "storj.io/storj/pkg/process"
 
 import (
 	"fmt"
+	"github.com/spf13/cobra"
```
coyle: this needs to move outside of std lib or the linter will complain

pkg/process/version.go
```diff
@@ -0,0 +1,30 @@
...
+	defer ticker.Stop()
+
+	select {
+	case <-ctx.Done():
+		// ToDO: Handle
+	case <-ticker.C:
+		err = version.CheckVersion()
```
coyle: error check

pkg/process/exec.go
stefanbenten: Already fixed 👍

cmd/release/main.go
jtolds: oh, yes!

internal/version/version.go
```diff
@@ -4,24 +4,106 @@
...
+	Build V
+	// Allowed ensures, the client is still on the allowed versions returned by the control server
+	Allowed bool
 )
 
-// V contains all necessary values for comparing different binary version with each other.
-// It's goal is to limit the startup up outdated binaries as well as having an easier way to ensure the software
-// is on the latest version from a UI/UX perspective
+// V is the versioning information for a binary
 type V struct {
```
egonelbre: `Info` that way the full name is `version.Info`.

stefanbenten: Will do!

stefanbenten:
I'll open an new PR, this is getting to messy and works from a fork and makes further additional PRs difficult 👍

PR #1586, opened 2019-03-27
[WIP] Store bandwidth from orders on satellite

mniewrzal:
## Code Review Checklist

Thanks for submitting a PR!


satellite/satellitedb/orders.go
```diff
@@ -87,7 +98,107 @@ func (db *ordersDB) SettleRemoteOrder(ctx context.Context, orderLimit *pb.OrderL
...
+			return err
+		}
+	}
 
 	return nil
 }
+
+// GetBucketBandwidth
+func (db *ordersDB) GetBucketBandwidth(ctx context.Context, bucketID []byte, from, to time.Time) (int64, error) {
+	var sum *int64
```
JessicaGreben: does this need to be a pointer? could it be `var sum int64`?

egonelbre: Unfortunately sqlite will return NULL if there are no rows, so using `var sum int64` would fail when there are no rows.

satellite/satellitedb/orders.go
```diff
@@ -87,7 +98,107 @@ func (db *ordersDB) SettleRemoteOrder(ctx context.Context, orderLimit *pb.OrderL
 		return err
 	}
 
-	// TODO store settle bandwidth in rollup tables
+	now := time.Now()
+	intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
+
+	bucketBandwidth, err := tx.Find_BucketBandwidthRollup_By_BucketId_And_IntervalStart_And_Action(
```
egonelbre: This should be a custom query that does an upsert.

satellite/satellitedb/orders.go
JessicaGreben: oh interesting.   There are many places in the code base that use a value for the arg to `.Scan()`.  Are those incorrect then? 

Here is one example:
https://github.com/storj/storj/blob/master/pkg/piecestore/psserver/psdb/psdb.go#L401

satellite/satellitedb/orders.go
```diff
@@ -5,25 +5,30 @@ package satellitedb
...
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/zeebo/errs"
 
 	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
 	dbx "storj.io/storj/satellite/satellitedb/dbx"
 )
 
+const defaultIntervalSeconds = 60 * 60 // hour
```
kaloyan-raev: A possible alternative for counting the number of seconds in an hour is: `int(time.Hour / time.Second)`.

mniewrzal: it looks that I can do `int(time.Hour.Seconds())`

mniewrzal: eh but then it cannot be constant - `const initializer int(time.Hour.Seconds()) is not a constant`

kaloyan-raev: You can, but only as a variable, not as a constant.

kaloyan-raev: `int(time.Hour / time.Second)` is still a constant, because it is an expression of two constants.

kaloyan-raev: If you do it like `var defaultIntervalSeconds = int(time.Hour.Seconds())` then you don't need the `int()` typecast as `Seconds()` return type is `int`. I think there will be a linter to fail because of this.

mniewrzal: `Seconds()` returns float64 but I used your version after all

satellite/satellitedb/orders.go
egonelbre: In case of selecting a single row (not using SUM), we get no rows... so that should be fine and the err is properly `sql.ErrNoRows`

satellite/orders/service.go
```diff
@@ -62,6 +62,32 @@ func (service *Service) saveSerial(ctx context.Context, serialNumber storj.Seria
...
+		return nil
+	}
+	var bucketAllocation int64
+	var action pb.PieceAction
+	nodesAllocation := make(map[storj.NodeID]int64)
+	for _, limit := range limits {
+		if limit != nil {
+			bucketAllocation += limit.Limit.Limit
+			nodesAllocation[limit.Limit.StorageNodeId] = limit.Limit.Limit
+			action = limits[0].Limit.Action
```
kaloyan-raev: There is a test panicking with nil reference here. My guess is that `limit[0]` is nil.

mniewrzal: Strange because earlier I have len(limits) == 0 but i will fix it because I made mistake during copy/paste.

pkg/storage/segments/store.go
```diff
@@ -133,7 +133,17 @@ func (s *segmentStore) Put(ctx context.Context, data io.Reader, expiration time.
 			Metadata:       metadata,
 		}
 	} else {
-		limits, rootPieceID, err := s.metainfo.CreateSegment(ctx, "", "", -1, redundancy, s.maxEncryptedSegmentSize, expiration) // bucket, path and segment index are not known at this point
+		// early call to get bucket name, rest of the path cannot be determine yet
+		p, _, err := segmentInfo()
+		if err != nil {
+			return Meta{}, Error.Wrap(err)
+		}
+		bucket, _, _, err := split(p)
```
egonelbre: `split` probably should have a nicer name: `splitPathFragments` or `splitPathTokens` or similar.

satellite/orders/endpoint.go
```diff
@@ -105,21 +119,26 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
+		if endpoint.satelliteSignee.ID() == orderLimit.UplinkId {
+			uplinkSignee = endpoint.satelliteSignee
+		} else {
+			uplinkPubKey, err := endpoint.certdb.GetPublicKey(ctx, orderLimit.UplinkId)
+			if err != nil {
+				endpoint.log.Warn("unable to find uplink public key", zap.Error(err))
+				return status.Errorf(codes.Internal, "unable to find uplink public key")
+			}
+			uplinkSignee = &signing.PublicKey{
+				Self: orderLimit.UplinkId, // TODO should this be taken from public key ??
```
egonelbre: It's probably safer to take it from order limit.

satellite/orders/endpoint.go
```diff
@@ -161,6 +181,30 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
 			}
 		}
```
egonelbre: `continue` is missing here.

satellite/orders/endpoint.go
```diff
@@ -161,6 +181,30 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
 
+		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, bucketID, orderLimit.Action, order.Amount); err != nil {
```
egonelbre: When we hit error here, we should "unuse" the serial number and then return the error and not reply with neither accepted nor rejected. That will cause the storage node to retry later.

satellite/orders/endpoint.go
```diff
@@ -161,6 +181,30 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
+				Status: pb.SettlementResponse_REJECTED,
+			})
+			if err != nil {
+				return formatError(err)
+			}
+			return nil
+		}
+
+		if err := endpoint.DB.UpdateStoragenodeBandwidthSettle(ctx, orderLimit.StorageNodeId, orderLimit.Action, order.Amount); err != nil {
+			err := stream.Send(&pb.SettlementResponse{
```
egonelbre: When we hit error here, we should "unuse" the serial number and then return the error and not reply with neither accepted nor rejected. That will cause the storage node to retry later.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -438,8 +453,23 @@ model storagenode_bandwidth_rollup (
...
+create storagenode_bandwidth_rollup ()
+
+read scalar (
+	select storagenode_bandwidth_rollup
+	where  storagenode_bandwidth_rollup.storagenode_id = ?
+	where  storagenode_bandwidth_rollup.interval_start = ?
+	where  storagenode_bandwidth_rollup.action = ?
+)
+
+update storagenode_bandwidth_rollup (
```
egonelbre: Is this still needed?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -438,8 +453,23 @@ model storagenode_bandwidth_rollup (
 	field interval_seconds uint
 	field action           uint
 
-	field allocated uint64
-	field settled   uint64
+	field allocated uint64 ( updatable )
+	field settled   uint64 ( updatable )
+)
+
+create storagenode_bandwidth_rollup ()
```
egonelbre: Is this still needed?

satellite/orders/endpoint.go
mniewrzal: It makes sense for first `UpdateBucketBandwidthSettle`. If we will unuse serial number here then with another try bucket bandwidth will be updated second time.

egonelbre: for the second one yeah, we would also need to rollback the previous one

pkg/storage/segments/store.go
mniewrzal: refactored to `splitPathFragments`

satellite/orders/endpoint.go
mniewrzal: added

satellite/orders/endpoint.go
mniewrzal: fixed

satellite/satellitedb/dbx/satellitedb.dbx
mniewrzal: good catch, removed

satellite/satellitedb/dbx/satellitedb.dbx
mniewrzal: good catch, removed

pkg/storage/segments/store.go
```diff
@@ -133,7 +133,17 @@ func (s *segmentStore) Put(ctx context.Context, data io.Reader, expiration time.
...
+		p, _, err := segmentInfo()
+		if err != nil {
+			return Meta{}, Error.Wrap(err)
+		}
+		bucket, _, _, err := splitPathFragments(p)
+		if err != nil {
+			return Meta{}, err
+		}
+
+		limits, rootPieceID, err := s.metainfo.CreateSegment(ctx, bucket, "", -1, redundancy, s.maxEncryptedSegmentSize, expiration) // bucket, path and segment index are not known at this point
```
JessicaGreben: the comment at the end of this line should be removed now `// bucket, path and segment index are not known at this point`

PR #1589, opened 2019-03-27
Change crypto/rand to use math/rand in pgutil

simongui:
# Summary
Currently we are silently assuming `crypto/Rand.Read()` always succeeds when it _can_ fail. This code is only for tests and not for any production level purposes so we can safely use `math/rand.Read()` instead which has no errors to handle instead.

# Solution
This PR does the following.
* Replaces the use of `crypto/rand.Read()` with `math/rand.Read()`.
* Renames the function `RandomString()` to `CreateRandomTestingSchemaName()` so that it is obvious that this is a test related function.

## Code Review Checklist

Thanks for submitting a PR!


internal/dbutil/pgutil/schema.go
```diff
@@ -8,18 +8,46 @@ import (
...
 )
 
-// RandomString creates a random safe string
-func RandomString(n int) string {
+const (
+	randRetries = 2
+)
+
+// RandomStringFromReader creates a random safe string sourced from the specified reader.
+func RandomStringFromReader(n int, r io.Reader) string {
```
thepaul: (a) If this is meant to be of use all throughout the codebase, we should probably put it somewhere more general under `internal/`. Maybe start a new package called internal/rand or internal/ioutils or something? If this is only meant to be of use here in this package, we should probably not export it.

simongui: Good idea! I'll see if there are any other uses and refactor if I find some.

thepaul: (b) there is no (b). I don't know why I put (a) at the beginning of that comment.

egonelbre: There's a much easier fix for this, since this was used only for testing we can use `math/rand`, which doesn't fail. Then rename the method to `RandomTestSchema`.

egonelbre: All other places should have proper error handling appropriate for the exact location.

internal/dbutil/pgutil/schema.go
```diff
@@ -8,18 +8,46 @@ import (
...
+	failed := false
 	data := make([]byte, n)
-	_, _ = rand.Read(data)
+	for i := 0; i < randRetries; i++ {
```
egonelbre: If it failed once then it will probably fail all the other tries. It makes more sense to kill the program immediately when this happens, because there's something very wrong with the system in that case.

jtolds:
this pr seems to have lost all the crypto/rand changes entirely!

simongui:
Alright, so the direction of this PR changed. I spoke to @egonelbre and this code is not used for production level code, it's only used for tests. I was over complicating what is needed. I renamed the PR and changed the description to reflect the changes in direction.

There may be a future PR down the pipe because my original concern of how we are handling `crypto/rand.Read()` may exist in other production level areas of the code base. I'll sweep the code base and look for those.

jtolds:
still LGTM :+1:

PR #1590, opened 2019-03-27
Forward port Bill and Stefan's changes to master

jtolds:
We probably should make a PR for each of these commits (with more tests for Bill's certainly) but I don't want to forget about these changes.


pkg/piecestore/psserver/psdb/psdb.go
```diff
@@ -205,6 +205,13 @@ func (db *DB) Migration() *migrate.Migration {
 					`CREATE INDEX IF NOT EXISTS idx_bwa_serial ON bandwidth_agreements (serial_num)`,
 				},
 			},
+			{
+				Description: "Initiate Network reset",
+				Version:     4,
+				Action: migrate.SQL{
+					`UPDATE ttl SET expires = 1553727600 WHERE created <= 1553727600 `,
```
stefanbenten: @jtolds Should we change the timestamp here?

stefanbenten:
We should wait with merging this until we figured the Repair and Audit Issues out 👍

pkg/piecestore/psserver/psdb/psdb.go
egonelbre: ttl table is not used anymore so you can set it to all rows

pkg/audit/verifier.go
```diff
@@ -159,7 +159,7 @@ func (d *defaultDownloader) getShare(ctx context.Context, limit *pb.AddressedOrd
 	// determines number of seconds allotted for receiving data from a storage node
 	timedCtx := ctx
 	if d.minBytesPerSecond > 0 {
-		maxTransferTime := time.Duration(int32(time.Second) * bandwidthMsgSize / d.minBytesPerSecond.Int32())
+		maxTransferTime := time.Second * time.Duration(10+(bandwidthMsgSize/d.minBytesPerSecond.Int32()))
```
wthorp: This code wasn't very clear.  This morning I put a PR that simply uses int64 instead, as I was able to confirm my thoughts that this was an int32 overflow.  My new code doesn't have the +10 and still divides by ints, so I suppose it would fail if the messages was <64 bytes.  Hrmm.

wthorp: I added in a 5 second minimum check in the merge here

wthorp:
build stuck, will reopen

pkg/piecestore/psserver/psdb/psdb.go
jtolds: not if we stopped uploads

jtolds: oh ha okay

pkg/audit/verifier.go
jtolds: cool

wthorp:
I'll fix the merge linting errors I introduced.

pkg/audit/verifier.go
```diff
@@ -170,6 +170,9 @@ func (d *defaultDownloader) getShare(ctx context.Context, limit *pb.AddressedOrd
 	timedCtx := ctx
 	if d.minBytesPerSecond > 0 {
 		maxTransferTime := time.Duration(int64(time.Second) * int64(bandwidthMsgSize) / d.minBytesPerSecond.Int64())
```
littleskunk: In release-alpha7 we fixed the issue with https://github.com/storj/storj/blob/0bacce74d3e1f7c6016503595acbd4704c712720/pkg/audit/verifier.go#L172

As far as I understand the important part is time.Seconds * time.Duration. On master we now try the old code that was producing wrong values? Please don't do that.

wthorp: @littleskunk Did you tests / resolve this difference?

littleskunk: Yes I have tested it. The result looks good.

PR #1595, opened 2019-03-28
record bucket data into bucket_storage_tally table

cam-a:
https://storjlabs.atlassian.net/browse/V3-1453
In order to limit data storage during the vanguard alpha and for client invoicing we need to keep track of how much data is stored in individual buckets.
This PR adds logic for recording bucket data from the pointerDB during the tally process.

### DIFF
--- a/pkg/accounting/db.go
+++ b/pkg/accounting/db.go
@@ -52,8 +52,6 @@ type DB interface {
 	SaveRollup(ctx context.Context, latestTally time.Time, stats RollupStats) error
 	// SaveAllBucketsInfo saves the latest bucket info
 	SaveBucketTallies(ctx context.Context, intervalStart time.Time, bucketTallies map[string]*BucketTally) error
-	// LatestTallyForBucket returns the latest tally for a bucket
-	LatestTallyForBucket(ctx context.Context, bucketID string) (*BucketTally, time.Time, error)
 	// QueryPaymentInfo queries Overlay, Accounting Rollup on nodeID
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
 	// DeleteRawBefore deletes all raw tallies prior to some time
--- a/pkg/accounting/db.go
+++ b/pkg/accounting/db.go
@@ -50,7 +50,7 @@ type DB interface {
 	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
 	// SaveRollup records raw tallies of at rest data to the database
 	SaveRollup(ctx context.Context, latestTally time.Time, stats RollupStats) error
-	// SaveAllBucketsInfo saves the latest bucket info
+	// SaveBucketTallies saves the latest bucket info
 	SaveBucketTallies(ctx context.Context, intervalStart time.Time, bucketTallies map[string]*BucketTally) error
 	// QueryPaymentInfo queries Overlay, Accounting Rollup on nodeID
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
--- a/pkg/accounting/db.go
+++ b/pkg/accounting/db.go
@@ -50,8 +50,8 @@ type DB interface {
 	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
 	// SaveRollup records raw tallies of at rest data to the database
 	SaveRollup(ctx context.Context, latestTally time.Time, stats RollupStats) error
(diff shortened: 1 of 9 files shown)


pkg/accounting/db.go
```diff
@@ -50,6 +50,8 @@ type DB interface {
 	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
 	// SaveRollup records raw tallies of at rest data to the database
 	SaveRollup(ctx context.Context, latestTally time.Time, stats RollupStats) error
+	//SaveBucketInfo saves the latest bucket info
```
JessicaGreben: put a space after `//` for consistency to all the other comments

satellite/satellitedb/accounting.go
```diff
@@ -162,6 +162,28 @@ func (db *accountingDB) SaveRollup(ctx context.Context, latestRollup time.Time,
 	return Error.Wrap(err)
 }
 
+// SaveBucketInfo saves the latest bucket info
+func (db *accountingDB) SaveBucketInfo(ctx context.Context, intervalStart time.Time, bucketInfo map[string]*accounting.BucketStats) error {
+	if len(bucketInfo) == 0 {
+		return Error.New("In SaveBucketInfo with empty bucketInfo")
+	}
+	for k, v := range bucketInfo {
```
JessicaGreben: A couple thoughts here:
1. the name of this function gives the impression it is only saving the info for one bucket, but it is actually saving the info for all the buckets (or at least a list of buckets).  Should we rename the function something like `SaveAllBucketsInfo` or some name to indicate it is more than one bucket?

2. I feel like `accounting.BucketStats` should actually be called `accounting.BucketTally` and should mirror the `dbx.BucketStorageTally` struct.  Then the argument to this function would be something like:
`bucketsTallyInfo map[string]*accounting.BucketTally` 

3. can we rename `k, v` to be more descriptive like `bucketID, bucketTallyInfo`

pkg/accounting/tally/tally.go
```diff
@@ -69,14 +69,22 @@ func (t *Tally) Run(ctx context.Context) (err error) {
 //Tally calculates data-at-rest and bandwidth usage once
 func (t *Tally) Tally(ctx context.Context) error {
 	//data at rest
```
dylanlott: nit: space between // and text for consistency

JessicaGreben: throughout the code base for comments some places have space after `//` and some places do not.  It seems like in this file most/all of the comments do not have the extra space. I would argue to keep them all that way instead of having to change them all.  In a prefect world we would have consistency across the entire repo, but that seems unrealistic and already failed. In my opinion, consistency in a single file/package is the best we can do.

pkg/accounting/db.go
```diff
@@ -50,6 +50,8 @@ type DB interface {
 	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
 	// SaveRollup records raw tallies of at rest data to the database
 	SaveRollup(ctx context.Context, latestTally time.Time, stats RollupStats) error
+	// SaveAllBucketsInfo saves the latest bucket info
```
JessicaGreben: update comment to name of method

pkg/accounting/tally/tally.go
```diff
@@ -152,14 +161,17 @@ func (t *Tally) calculateAtRestData(ctx context.Context) (latestTally time.Time,
 					if currentBucket != bucketID {
 						if currentBucket != "" {
 							// report the previous bucket and add to the totals
-							currentBucketStats.Report("bucket")
-							totalStats.Combine(&currentBucketStats)
-							currentBucketStats = stats{}
+							currentBucketTally.Report("bucket")
+							totalTallies.Combine(&currentBucketTally)
+
+							// add currentBucketStats to bucketStats
```
JessicaGreben: i think this comment needs to be updated too

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -424,6 +424,20 @@ model bucket_storage_tally (
...
+	select bucket_storage_tally
+	where bucket_storage_tally.interval_start >= ?
+	where bucket_storage_tally.interval_start < ?
+)
+
+// for deleting after rollup
+delete bucket_storage_tally (
+	where bucket_storage_tally.interval_start >= ?
+	where bucket_storage_tally.interval_start < ?
+)
```
JessicaGreben: are you using these `read all` and `delete bucket_storage_tally`? If not, I feel like you can remove them here.

PR #1596, opened 2019-03-28
OverlayCache DB missing updates

aligeti:
The PR is to root cause, why overlay cache DB is not getting updated, while nodesDB(aka statDB) is. 
RootCause:
The CreateEntryIfNotExists() is used to create an entry if it doesn't exist on both nodesDB (aka statDB) 
and overlayCacheDB. nodesDB uses transactions to create an entry if it doesn't exist, but overlay cache uses two separate queries to do the very similar operation. To make sure it is transaction, I wrote a test case to test CreateEntryIfNotExists() emulation accessing the DBs parallelly and found out that without transactions modification, the function is failing very often, which proves that transactions are more robust than accessing thru individual queries. 

played with concurrency values:
1 . with concurrency set to 1000 -> without transaction it failed 40% of the time consistently and with transaction, I saw 1 failure (10%), because I ran the test 10 times. 
2. With concurrency set to 10000 -> without transaction it failed 60% of the time consistently and with transaction, I saw 20% failure. 

The recent merge/refactor fixed this problem, refactor was implemented using transaction. 

This PR's root cause is addressed, I added a stress test case testing the function. 

Thanks for submitting a PR!

### DIFF
--- a/pkg/overlay/statdb_test.go
+++ b/pkg/overlay/statdb_test.go
@@ -5,6 +5,7 @@ package overlay_test
 
 import (
 	"context"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -264,6 +265,8 @@ func testDatabase(ctx context.Context, t *testing.T, cache overlay.DB) {
 	{ // TestUpdateBatchExists
 		nodeID1 := storj.NodeID{255, 1}
 		nodeID2 := storj.NodeID{255, 2}
+		nodeID3 := storj.NodeID{255, 3}
+		nodeID4 := storj.NodeID{255, 4}
 
 		auditSuccessCount1 := int64(4)
 		auditCount1 := int64(10)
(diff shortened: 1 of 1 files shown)


pkg/overlay/statdb_test.go
```diff
@@ -339,5 +342,19 @@ func testDatabase(ctx context.Context, t *testing.T, cache overlay.DB) {
 		assert.EqualValues(t, newUptimeRatio1, stats1.UptimeRatio)
 		assert.EqualValues(t, newAuditRatio2, stats2.AuditSuccessRatio)
 		assert.EqualValues(t, newUptimeRatio2, stats2.UptimeRatio)
+
+		concurrency := 100
+		t.Run("CreateEntryIfNotExists", func(t *testing.T) {
+			nodes := []storj.NodeID{nodeID1, nodeID2, nodeID3, nodeID4}
+			for P := 0; P < concurrency; P++ {
```
dylanlott: This parallel test is a great idea 👍

satellite/satellitedb/overlaycache.go
```diff
@@ -618,12 +618,46 @@ func (cache *overlaycache) UpdateBatch(ctx context.Context, updateReqList []*ove
...
+			totalAuditCount    int64
+			auditSuccessCount  int64
+			auditSuccessRatio  float64
+			totalUptimeCount   int64
+			uptimeSuccessCount int64
+			uptimeRatio        float64
+			wallet             string
+			email              string
+		)
+		dbNode, err = tx.Create_Node(
```
aleitner: Do we normally create nodes with no other information such as wallet and email? Is it possible that these nodes could get used without getting paid?

aligeti: the wallet and email are updated separately. 

discovery/service.go L#157
// update wallet with correct info
		info, err := discovery.kad.FetchInfo(ctx, *node)

PR #1597, opened 2019-03-28
adds exponential backoff restart for storage node service calls

navillasa:
## why changes
- Occasionally an authentication handshake failed/connection reset by peer error appears and the storage nodes exit, but they should keep attempting to connect, at least a few times, with exponential back off.

## what changes
- Adds `backoffRestart` function in storagenode/peer.go which now wraps all of the storage node peer's service calls.
- On all attempts to run the function after the first, there will be a wait (time.Sleep call) before the next call. Every iteration, the wait interval will multiply by 2, until it exceeds the `maxWait` time, which has been set in the file as a const of 30 seconds. It could be configured though.

### DIFF
--- a/storagenode/peer.go
+++ b/storagenode/peer.go
@@ -295,24 +295,31 @@ func (peer *Peer) Run(ctx context.Context) error {
 	return group.Wait()
 }
 
-func (peer *Peer) exponentialBackoffBootstrap(ctx context.Context, waitInterval, maxWait time.Duration, attempts int, combined error) error {
-	if waitInterval*2 > maxWait {
-		return errs.Combine(combined, Error.New("unable to bootstrap to network after %d attempts", attempts))
-	}
-	if attempts == 1 {
-		time.Sleep(waitInterval)
-	}
-	if attempts > 1 {
-		waitInterval = waitInterval * 2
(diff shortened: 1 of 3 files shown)


storagenode/peer.go
```diff
@@ -282,6 +295,27 @@ func (peer *Peer) Run(ctx context.Context) error {
 	return group.Wait()
 }
 
+func (peer *Peer) exponentialBackoffBootstrap(ctx context.Context, waitInterval, maxWait time.Duration, attempts int, combined error) error {
```
mniewrzal: Maybe you can reorganize code to use loop instead of recursion? This should simplify code and will remove `attempts` parameter. Also you can think about extracting a generic function to be able to use is in a different places. Something like `exponentialRetry(waitInterval, maxWait time.Duration, func (ctx context.Context) error)`

navillasa: that sounds good! thanks

jtolds:
wait did this get merged somewhere else?

navillasa:
> wait did this get merged somewhere else?

sorry should've added some explanation- no, I talked to @egonelbre yesterday and decided that the restart logic needs to be added specifically to the bootstrap call instead, instead of adding another method to peer.

PR #1598, opened 2019-03-28
do not add the same segment to the repair queue multiple times

mobyvb:
There is a chance (depending on checker/repair intervals) that the data repair checker iterates over the same damaged segment multiple times before that segment has a chance to be repaired once. This PR ensures that a segment is not added to the repair queue unless it is not already in the repair queue.

Thanks for submitting a PR!


mobyvb:
this might not actually fix the problem, as the repair for a segment might be in progress even after that segment is dequeued from the repair queue

PR #1604, opened 2019-03-29
fix error handling in GetByProjectID

egonelbre:
When we hit an scanning error then this is most likely an error in code and there's no need to keep scanning the rows.


satellite/satellitedb/projectmembers.go
```diff
@@ -71,20 +71,17 @@ func (pm *projectMembers) GetByProjectID(ctx context.Context, projectID uuid.UUI
 
 		scanErr := rows.Scan(&memberIDBytes, &projectIDBytes, &pm.CreatedAt)
 		if err != nil {
-			err = errs.Combine(err, scanErr)
-			continue
+			return nil, scanErr
```
kaloyan-raev: The line above should be `if scanErr != nil`.

Perhaps, it would be less error-prone to reuse `err` in all three places instead of `scanErr` and `convertErr`.

PR #1609, opened 2019-03-29
Remove pointerdb.Server

egonelbre:
The code has migrated to either using `pointerdb.Service` and `metainfo`, so  pointerdb server and allocation signer aren't necessary anymore.


pkg/pointerdb/pointerdb_test.go
```diff
@@ -1,472 +0,0 @@
-// Copyright (C) 2019 Storj Labs, Inc.
-// See LICENSE for copying information.
-
-package pointerdb_test
```
egonelbre: Not sure what to do about these tests. We already have them covered elsewhere, however with different approaches.

egonelbre: @kaloyan-raev, @mniewrzal ?

mniewrzal: Get and Delete tests can be deleted for me. I'm looking into List and Put tests.

mniewrzal: Regarding Put tests I would add at least one test for `Number of valid pieces is lower then success threshold` somewhere. Testplanet uplink.Upload with custom redundancy scheme. List tests would be nice to move/adjust I guess. We can upload test data as inline pointers.

egonelbre: For List we do have https://github.com/storj/storj/blob/master/pkg/metainfo/kvmetainfo/objects_test.go#L293 as tests. However that is at the db level.

mniewrzal: Ok, maybe its enough. Additional value with pointerdb_test is that it's trying to check some special characters (at least at basic level).

egonelbre: It looks like we are not verifying the "lower than success threshold" at the moment.

egonelbre: I created a separate issue for the "Number of valid pieces is lower then success threshold".

egonelbre: And implemented basic utf-8 list tests.

PR #1610, opened 2019-03-29
Satellite password check simplification

crawter:


internal/testplanet/planet.go
```diff
@@ -450,13 +450,16 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
 			},
 			Mail: mailservice.Config{
-				SMTPServerAddress: "smtp.mail.example.com:587",
-				From:              "Labs <storj@example.com>",
-				AuthType:          "simulate",
+				SMTPServerAddress: "smtp.gmail.com:587",
+				From:              "Labs <andrewm@storj.io>",
+				AuthType:          "plain",
+				Password:          "Aquachester12",
+				Login:             "andrewm@storj.io",
```
egonelbre: Only `example.com` should be used in tests.

internal/testplanet/planet.go
```diff
@@ -450,13 +450,16 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
 			},
 			Console: consoleweb.Config{
-				Address:      "127.0.0.1:0",
+				Address:      "127.0.0.1:8082",
```
egonelbre: Must bind at `127.0.0.1:0`.

PR #1616, opened 2019-03-29
prevents audit's cursor.NextStripe from returning nil, nil

navillasa:
# why changes
Previously, there were 3 cases where the audit cursor's NextStripe function could return a nil (instead of a stripe) as well as a nil error. This meant that NextStripe would have to be called multiple times in some cases to acquire a valid stripe to audit.

# what changes
Adds new `getRandomValidPointer` function the pkg/audit/cursor.go to ensure that a valid stripe is found, or an error is returned after a configured number of retries.
The retry values still need to be added as a config.

Thanks for submitting a PR!

### DIFF
--- a/pkg/audit/cursor.go
+++ b/pkg/audit/cursor.go
@@ -60,6 +60,14 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
 		if err != nil {
 			return nil, err
 		}
+
+		// keep track of last path listed
+		if !more {
+			cursor.lastPath = ""
+		} else {
+			cursor.lastPath = pointerItems[len(pointerItems)-1].Path
+		}
+
 		if len(pointerItems) == 0 {
 			attempts++
 		} else {
@@ -93,13 +101,6 @@ func (cursor *Cursor) getRandomValidPointer(pointerItems []*pb.ListResponse_Item
 		pointerItem, err := getRandomPointer(pointerItems)
 		if err != nil {
 			return nil, err
-		}
-
-		// keep track of last path listed
-		if !more {
-			cursor.lastPath = ""
-		} else {
-			cursor.lastPath = pointerItems[len(pointerItems)-1].Path
 		}
 
 		pointer, err = cursor.pointerdb.Get(pointerItem.Path)
--- a/pkg/audit/cursor.go
+++ b/pkg/audit/cursor.go
@@ -79,7 +79,7 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
 		return nil, Error.New("unable to find pointers after %d attempts", maxListAttempts)
 	}
 
-	pointer, err := cursor.getRandomValidPointer(pointerItems, more, maxGetAttempts)
+	pointer, err := cursor.getRandomValidPointer(pointerItems, maxGetAttempts)
 	if err != nil {
 		return nil, err
 	}
@@ -96,7 +96,7 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
 	}, nil
 }
 
-func (cursor *Cursor) getRandomValidPointer(pointerItems []*pb.ListResponse_Item, more bool, retryAttempts int) (pointer *pb.Pointer, err error) {
+func (cursor *Cursor) getRandomValidPointer(pointerItems []*pb.ListResponse_Item, retryAttempts int) (pointer *pb.Pointer, err error) {
 	for i := 0; i < retryAttempts; i++ {
 		pointerItem, err := getRandomPointer(pointerItems)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/audit/cursor.go
```diff
@@ -49,64 +54,78 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
 	var path storj.Path
 	var more bool
 
-	pointerItems, more, err = cursor.pointerdb.List("", cursor.lastPath, "", true, 0, meta.None)
-	if err != nil {
-		return nil, err
+	var attempts = 0
+	for attempts < maxListAttempts {
+		pointerItems, more, err = cursor.pointerdb.List("", cursor.lastPath, "", true, 0, meta.None)
```
mniewrzal: Won't this always return the same list of pointers in a specific moment in time? So if `pointerItems` will be empty the first time then there is small chance that something will change?

navillasa: okay I see what you mean-- hopefully my latest change (re-assigning cursor.lastPath) within this loop will help.

navillasa:
> This change will log errors with every satellite run when pointerdb is empty:
> 
> ```
> satellite/0              09:54:31.477 | ERROR	audit	process	{"error": "audit error: unable to find pointers after 4 attempts", "errorVerbose": "audit error: unable to find pointers after 4 attempts\n\tstorj.io/storj/pkg/audit.(*Cursor).NextStripe:71\n\tstorj.io/storj/pkg/audit.(*Service).process:76\n\tstorj.io/storj/pkg/audit.(*Service).Run.func1:60\n\tstorj.io/storj/internal/sync2.(*Cycle).Run:72\n\tstorj.io/storj/pkg/audit.(*Service).Run:59\n\tstorj.io/storj/satellite.(*Peer).Run.func8:532\n\tgolang.org/x/sync/errgroup.(*Group).Go.func1:57"}
> satellite/0              09:54:31.477 | ERROR	tally	Tally failed	{"error": "Query for bandwidth failed: tally error: context canceled", "errorVerbose": "Query for bandwidth failed: tally error: context canceled\n\tstorj.io/storj/pkg/accounting/tally.(*Tally).Tally:85\n\tstorj.io/storj/pkg/accounting/tally.(*Tally).Run:58\n\tstorj.io/storj/satellite.(*Peer).Run.func6:526\n\tgolang.org/x/sync/errgroup.(*Group).Go.func1:57"}
> satellite/0              09:54:31.477 | INFO	rollup	Rollup found no new tallies
> satellite/0              09:54:31.477 | ERROR	process	{"error": "context canceled"}
> ```

hm.. should I not return that as an error then? should it just continue looping to find pointers? I'm worried that could lead to an infinite loop.

pkg/audit/cursor.go
```diff
@@ -49,64 +54,79 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
...
+		if err != nil {
+			return nil, err
+		}
+
+		// keep track of last path listed
+		if !more {
+			cursor.lastPath = ""
```
mniewrzal: does it make sense to make another attempt if there is nothing else to list?

pkg/audit/cursor.go
```diff
@@ -49,64 +54,79 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
...
-	if expiration := pointer.GetExpirationDate(); expiration != nil {
-		t, err := ptypes.Timestamp(expiration)
+	return &Stripe{
+		Index:       index,
+		Segment:     pointer,
+		SegmentPath: path,
+	}, nil
+}
+
+func (cursor *Cursor) getRandomValidPointer(pointerItems []*pb.ListResponse_Item, more bool, retryAttempts int) (pointer *pb.Pointer, err error) {
```
mniewrzal: `more `param looks to be unused

pkg/audit/cursor.go
wthorp: consider maybe having NextStripe return 'more' instead of all this retrying

navillasa:
> Maybe something like: if with first attempt List returns 0 elements then don't return error, just nil

okay, so you do think NextStripe should be returning nil, nil?

pkg/audit/cursor.go
navillasa: setting cursor.lastPath to `""` means that pointerdb will start listing from the beginning. are you saying we shouldn't start listing from the beginning? or that I need another check?

pkg/audit/cursor.go
navillasa: good point!

pkg/audit/cursor.go
mniewrzal: hmm my understanding was wrong. Code is ok and we should start listing from the beginning.

navillasa:
this is stale and requires new work: https://storjlabs.atlassian.net/browse/V3-1467

PR #1617, opened 2019-03-29
fixes infinite loop in audit TestVerifierHappyPath

navillasa:
# why changes
When there are no pointers in the pointerdb, this test caused a memory leak because of an infinite loop.

# what changes
Just makes the NextStripe call 3 times and if it still doesn't have a non-nil stripe, then the test fails. #1616 refactors NextStripe so that a more informative error is returned in the future (i.e. "unable to find pointers after %d attempts").
- Also makes sure that all audit tests use `require` instead of `assert`.

### DIFF
--- a/pkg/audit/newverifier_test.go
+++ b/pkg/audit/newverifier_test.go
@@ -7,7 +7,6 @@ import (
 	"math/rand"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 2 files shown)


pkg/audit/newverifier_test.go
```diff
@@ -38,14 +38,15 @@ func TestVerifierHappyPath(t *testing.T) {
 		cursor := audit.NewCursor(pointerdb)
```
egonelbre: use `require.NoError` for upload check

navillasa:
> LGTM, but I still wonder if its better to fix cursor.NextStripe() so that it doesn't hide the fact that it has looped over all items

I agree that's a good idea! #1616 should add some changes soon to do this

PR #1618, opened 2019-03-30
Remove bw from tally service + query bandwidth in rollup service

jenlij:
What:
• Removes bandwidth calculations from the tally service
• Remove SaveBWRaw
• Rolls up storagenode_bandwidth_rollup entries for latest 24 hr interval
• Updates UpdateStoragenodeBandwidthSettle to take intervalStart as a parameter

Why:
New storagenode_bandwidth_rollup does the tallying for each hour already. Need this info in 24 hr intervals to allow for payment queries.


pkg/accounting/rollup/rollup.go
```diff
@@ -86,35 +106,71 @@ func (r *Rollup) RollupRaws(ctx context.Context) error {
...
 	latestTally = time.Date(latestTally.Year(), latestTally.Month(), latestTally.Day(), 0, 0, 0, 0, latestTally.Location())
 	delete(rollupStats, latestTally)
 	if len(rollupStats) == 0 {
 		r.logger.Info("Rollup only found tallies for today")
-		return nil
+		return time.Now(), nil
 	}
-	err = r.db.SaveRollup(ctx, latestTally, rollupStats)
+
+	return latestTally, Error.Wrap(r.db.DeleteRawBefore(ctx, latestTally))
```
cam-a: Looks like we'll need to return rollupStats since we're not saving it in this function anymore

cam-a: Also maybe we don't want to delete the raws until the end of `SaveRollup` just in case something happens before it's recorded

pkg/accounting/rollup/rollup.go
```diff
@@ -54,30 +54,55 @@ func (r *Rollup) Run(ctx context.Context) (err error) {
...
 		return Error.Wrap(err)
 	}
-	tallies, err := r.db.GetRawSince(ctx, lastRollup)
+	rollupStats := make(accounting.RollupStats)
+	latestTally, err := r.RollupStorage(ctx, lastRollup, rollupStats)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	if len(rollupStats) == 0 {
+		r.logger.Info("RollupStats is empty I")
```
cam-a: typo?

pkg/accounting/db.go
```diff
@@ -25,7 +25,15 @@ type Raw struct {
 	CreatedAt       time.Time
 }
 
-//Rollup mirrors dbx.AccountingRollup, allowing us to use that struct without leaking dbx
+// BW mirrors dbx.StoragenodeBandwidthRollup, allowing us to use the struct without leaking dbx
+type BW struct {
```
cam-a: Since Raw is no longer atRest AND bw maybe we should get rid of the dataType field on the accounting Raw struct? Also, maybe we should consider renaming some things since raw used to include both?

jenlij: That is a good point! Let's leave it for now, and make these changes once we move the raws to the new tables.

pkg/accounting/rollup/rollup.go
jenlij: whoops yeah thanks

satellite/orders/endpoint.go
```diff
@@ -44,6 +44,9 @@ type DB interface {
 	// UpdateStoragenodeBandwidthSettle updates 'settled' bandwidth for given storage node
 	UpdateStoragenodeBandwidthSettle(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64) error
 
+	// UpdateStoragenodeBandwidthSettleWithCustomDate updates 'settled' bandwidth for given storage node for the given intervalStart time
+	UpdateStoragenodeBandwidthSettleWithCustomDate(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
```
cam-a: Not sure if we should add methods to the DB that won't actually be implemented just for the sake of testing. Maybe we can refactor the tests in a way that we don't need this?

cam-a: Maybe it's not a big deal though

jenlij: Not sure how we can do that, since we don't have access to dbx outside satellitedb.. Any ideas?

cam-a: yeah, I was just looking and I'm not sure. Either:
keep the new method,
add an interval argument to UpdateStoragenodeBandwidthSettle and set it to use time.Now() if the arg is nil,
throw out the current test architecture ¯\_(ツ)_/¯

JessicaGreben: This method makes me nervous.  There is no way of knowing (without reading all the code) that this method is supposed to be used in testing.  Im worried that someone could accidentally use it in production code instead of the `UpdateStoragenodeBandwidthSettle` which _should_ be used.  

For the test that uses this method, I'm curious if there is a way we can mock out the time package when creating the `storagenode_bandwidth_rollups` instead of adding this method. Does that make sense?

jenlij: I definitely see the issue, and Cam pointed that out too. I'm not sure how to fix it.

JessicaGreben: As an example, I'm reading this about mocking time package: https://stackoverflow.com/questions/18970265/is-there-an-easy-way-to-stub-out-time-now-globally-in-golang-during-test

jenlij: maybe something like this? https://www.reddit.com/r/golang/comments/a8yx3x/mocking_system_time_in_tests/

JessicaGreben: oops I didn't see cam's comment til now, didn't refresh.

egonelbre: Add argument `intervalStart time.Time` to `UpdateStoragenodeBandwidthSettle`.

pkg/accounting/rollup/rollup.go
```diff
@@ -54,30 +54,55 @@ func (r *Rollup) Run(ctx context.Context) (err error) {
...
 }
 
-// RollupRaws rolls up raw tally
-func (r *Rollup) RollupRaws(ctx context.Context) error {
+// Rollup aggregates storage and bandwidth amounts for the time interval
+func (r *Service) Rollup(ctx context.Context) error {
 	// only Rollup new things - get LastRollup
-	var latestTally time.Time
 	lastRollup, err := r.db.LastTimestamp(ctx, accounting.LastRollup)
+	r.logger.Info("last Rollup: " + lastRollup.String())
```
JessicaGreben: log info after the err is checked

jenlij: oh shoot, that was just a debugging statement, removing.

satellite/orders/endpoint.go
```diff
@@ -41,8 +41,8 @@ type DB interface {
 
 	// UpdateStoragenodeBandwidthAllocation updates 'allocated' bandwidth for given storage node
 	UpdateStoragenodeBandwidthAllocation(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64) error
-	// UpdateStoragenodeBandwidthSettle updates 'settled' bandwidth for given storage node
-	UpdateStoragenodeBandwidthSettle(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64) error
+	// UpdateStoragenodeBandwidthSettleWithCustomDate updates 'settled' bandwidth for given storage node for the given intervalStart time
```
cam-a: comment

satellite/orders/endpoint.go
```diff
@@ -41,8 +41,8 @@ type DB interface {
...
+	UpdateStoragenodeBandwidthSettle(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
```
cam-a: I feel like it looks a bit odd that this method takes the intervalStart arg, but none of the others do. Should we change the others to be consistent even though they don't actually need it?

jenlij: @egonelbre what do you think?

pkg/accounting/db.go
```diff
@@ -25,7 +25,15 @@ type Raw struct {
...
+	NodeID        storj.NodeID
```
JessicaGreben: any chance we can call this struct something more descriptive? `StoragenodeBandwidthRollup` or something?

pkg/accounting/db.go
```diff
@@ -42,14 +50,14 @@ type Rollup struct {
...
-	SaveBWRaw(ctx context.Context, tallyEnd time.Time, created time.Time, bwTotals map[storj.NodeID][]int64) error
 	// SaveAtRestRaw records raw tallies of at-rest-data.
 	SaveAtRestRaw(ctx context.Context, latestTally time.Time, created time.Time, nodeData map[storj.NodeID]float64) error
 	// GetRaw retrieves all raw tallies
 	GetRaw(ctx context.Context) ([]*Raw, error)
-	// GetRawSince r retrieves all raw tallies sinces
+	// GetRawSince retrieves all raw tallies since latestRollup
 	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
+	// GetBWSince retrieves all bandwidth_rollup entires since latestRollup
+	GetBWSince(ctx context.Context, latestRollup time.Time) ([]*BW, error)
```
JessicaGreben: i think this name should be more descriptive as well.  Maybe `GetStoragenodeBandwidthRollups` or `GetStoragenodeBandwidthSince`? I know the names are getting long, but its just confusing because there are `StoragenodeBandwidthRollup` and also `BucketBandwidthRollup` so I would like to be specific which we are talking about.

jenlij: good idea to both rename suggestions

PR #1625, opened 2019-04-01
satellite: don't die when subservice dies

jtolds:
The satellite often dies for all sorts of crazy things - perhaps an injured segment is found but a DB error prevents it from getting added. Satellite dies completely!

This makes it so we restart services without killing the entire process.

Thanks for submitting a PR!

### DIFF
--- a/satellite/peer.go
+++ b/satellite/peer.go
@@ -523,6 +523,21 @@ func ignoreCancel(err error) error {
 	return err
 }
 
+func (peer *Peer) restartLoop(ctx context.Context, name string, service func(context.Context) error) error {
+	for {
+		err := func() (err error) {
+			defer func() {
+				if rec := recover(); rec != nil {
+					err = errs.New("caught panic in service %s: %+v", name, rec)
+				}
+			}()
+			return service(ctx)
+		}()
+		peer.Log.Sugar().Errorf("Service %s failure: %+v", name, err)
+		time.Sleep(time.Second)
+	}
+}
+
 // Run runs storage node until it's either closed or it errors.
 func (peer *Peer) Run(ctx context.Context) error {
 	group, ctx := errgroup.WithContext(ctx)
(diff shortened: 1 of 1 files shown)


satellite/peer.go
```diff
@@ -491,6 +491,21 @@ func ignoreCancel(err error) error {
 	return err
 }
 
+func (peer *Peer) restartLoop(ctx context.Context, name string, service func(context.Context) error) error {
```
egonelbre: This seems like a really dangerous thing to do. If something panics the memory might be corrupted.

jtolds: net/http does it?

egonelbre: really?

jtolds: but sure, the panic handling isn't the most important thing about this. given that net/http's Serve does it, do you still want me to take it out?

egonelbre: Yeah, looks like it. :/

egonelbre: I'm usually in favor of crash hard and early, because tracing down the actual bug and damage after the fact can be much more difficult. But I understand if we want to keep satellite up and running no matter the cost.

satellite/peer.go
```diff
@@ -491,6 +491,21 @@ func ignoreCancel(err error) error {
...
+		err := func() (err error) {
+			defer func() {
+				if rec := recover(); rec != nil {
+					err = errs.New("caught panic in service %s: %+v", name, rec)
+				}
+			}()
+			return service(ctx)
+		}()
+		peer.Log.Sugar().Errorf("Service %s failure: %+v", name, err)
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
navillasa: I think it's important to have a storage node restart instead of immediately exiting the process when it can't get a connection, e.g. this bug that @brimstone saw occurring occasionally: https://storjlabs.atlassian.net/browse/V3-1370

satellite/peer.go
```diff
@@ -491,6 +491,21 @@ func ignoreCancel(err error) error {
 	return err
 }
 
+func (peer *Peer) restartLoop(ctx context.Context, name string, service func(context.Context) error) error {
+	for {
```
egonelbre: PS: this needs to handle context cancellation properly

navillasa: does this mean call `ignoreCancel` inside here?

jtolds: egon's saying we need to see if the ctx that's passed in is still valid before continuing through the for loop. if the context is canceled, we should stop looping

satellite/peer.go
```diff
@@ -499,25 +514,25 @@ func (peer *Peer) Run(ctx context.Context) error {
 		return ignoreCancel(peer.Kademlia.Service.Bootstrap(ctx))
 	})
 	group.Go(func() error {
-		return ignoreCancel(peer.Kademlia.Service.Run(ctx))
+		return ignoreCancel(peer.restartLoop(ctx, "kademlia", peer.Kademlia.Service.Run))
```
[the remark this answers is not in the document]
jtolds: okay with the comment below, peer.restartLoop might return ctx.Canceled, so we still need it

satellite/peer.go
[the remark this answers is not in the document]
jtolds: oh no 1597 was closed! i'll add a random sleep

satellite/peer.go
jtolds: yeah i think we do

ifraixedes:
I'm wondering why this PR has not been merged, has been forgotten?

mniewrzal:
@jtolds ping ^

PR #1626, opened 2019-04-01
Introduce NodeDossier type and cleanup overlay.DB interface

kaloyan-raev:
This a continuation of StatDB and Overlay Cache merge and first steps of refactoring the Node protobufs.

- Introduces the `NodeDossier` type
- Removes methods from the `overlay.DB` interface:
  - CreateEntryIfNotExists
  - GetStats
  - Delete
  - UpdateBatch

Thanks for submitting a PR!


pkg/overlay/cache.go
```diff
@@ -43,34 +44,26 @@ type DB interface {
 	SelectNewStorageNodes(ctx context.Context, count int, criteria *NewNodeCriteria) ([]*pb.Node, error)
```
egonelbre: The offline check needs to be part of select nodes query.

kaloyan-raev: which offline check exactly?

egonelbre: previously `SelectNewStorageNodes` worked on overlay which didn't contain nodes that were offline for the last ping. So it shouldn't select nodes where the last ping failed.

internal/testplanet/uplink_test.go
```diff
@@ -98,9 +98,6 @@ func TestDownloadWithSomeNodesOffline(t *testing.T) {
 			if nodesToKill[node.ID()] {
 				err = planet.StopPeer(node)
 				require.NoError(t, err)
-
-				err = satellite.Overlay.Service.Delete(ctx, node.ID())
```
egonelbre: I'm wondering whether we should make a function to mark nodes offline or the last ping failing instead.

pkg/overlay/cache.go
kaloyan-raev: This was already done in a previous PR: https://github.com/storj/storj/blob/master/satellite/satellitedb/overlaycache.go#L42-L43

internal/testplanet/uplink_test.go
kaloyan-raev: Good point!

pkg/overlay/cache.go
egonelbre: ah, nevermind then... missed that change

internal/testplanet/uplink_test.go
kaloyan-raev: Fixed with the latest commit.

pkg/discovery/service_test.go
```diff
@@ -36,16 +37,19 @@ func TestCache_Graveyard(t *testing.T) {
...
-		err := satellite.Overlay.Service.Delete(ctx, offlineID)
+		// mark node as offline in overlay cache
+		_, err := satellite.Overlay.Service.UpdateUptime(ctx, offlineID, false)
+		require.NoError(t, err)
+
+		node, err := satellite.Overlay.Service.Get(ctx, offlineID)
 		assert.NoError(t, err)
-		_, err = satellite.Overlay.Service.Get(ctx, offlineID)
-		assert.NotNil(t, err)
+		assert.False(t, node.Online())
```
navillasa: `node.Online()` is a cool addition

pkg/overlay/cache.go
```diff
@@ -114,16 +107,35 @@ type UpdateRequest struct {
 	IsUp         bool
 }
 
+// NodeDossier is the complete info that the satellite tracks for a storage node
+type NodeDossier struct {
+	pb.Node
```
navillasa: why doesn't this have a field name?

satellite/satellitedb/overlaycache.go
```diff
@@ -487,7 +465,7 @@ func (cache *overlaycache) UpdateOperator(ctx context.Context, nodeID storj.Node
 		return nil, Error.Wrap(tx.Rollback())
 	}
 
-	updated := getNodeStats(nodeID, updatedDBNode)
+	updated, err := convertDBNode(updatedDBNode)
```
navillasa: should this err be checked before combined and returned? or is it okay to combine nil?

satellite/satellitedb/overlaycache.go
```diff
@@ -589,46 +524,36 @@ func convertDBNode(info *dbx.Node) (*pb.Node, error) {
...
+			},
 		},
-		Metadata: &pb.NodeMetadata{
+		Type: pb.NodeType(info.Type),
+		Operator: pb.NodeOperator{
 			Email:  info.Email,
 			Wallet: info.Wallet,
 		},
-		Restrictions: &pb.NodeRestrictions{
+		Capacity: pb.NodeCapacity{
```
navillasa: this is way clearer than restrictions imo 👍

pkg/overlay/cache.go
kaloyan-raev: so if you have `var node NodeDossier` you can just do `node.Id` and `node.Address`.

satellite/satellitedb/overlaycache.go
kaloyan-raev: it's safe to pass `nil` errors to `errs.Combine`.

jtolds: sure, but we probably shouldn't run tx.Commit if err != nil, right?

pkg/overlay/cache.go
```diff
@@ -114,16 +107,35 @@ type UpdateRequest struct {
...
+	Capacity   pb.NodeCapacity
+	Reputation NodeStats
+}
+
+// Online checks if a node is online based on the collected statistics.
+//
+// A node is considered online if the last attempt for contact was successful
+// and it was within the last hour.
+func (node *NodeDossier) Online() bool {
+	return time.Now().Sub(node.Reputation.LastContactSuccess) < 1*time.Hour &&
```
jtolds: could this 1*time.Hour be configurable somehow? hmm

jtolds: at very least might be nice to promote to a top level const

kaloyan-raev: what's the use case?

jtolds: to be honest, having it be configurable could potentially help with https://storjlabs.atlassian.net/browse/V3-1479, where we don't want the tests to have to wait for an hour before a down node is considered offline

jtolds: okay if this isn't fixed in this pr though

satellite/satellitedb/overlaycache.go
kaloyan-raev: True. I will fix it.

pkg/overlay/cache.go
kaloyan-raev: I don't believe this would help the tests. Tests can call `UpdateUptime` to make nodes appear as offline after stopping them. This PR modifies the tests to do this.

jtolds: i'm still just anxious that i pulled the 1 hour number out of a hat and data science might discover that there's a better value for this and might like to try different things. i'm satisfied if it can have a top level const name and a comment

navillasa:
> I tried running this locally and it broke the kad package tests for me. Were you seeing that locally?

Update that I ran these locally with a longer timeout (2min) and it passed. Didn't realize the kademlia (not kademlia_test) package needed that long to run.

pkg/audit/newverifier_test.go
```diff
@@ -64,10 +64,10 @@ func TestVerifierHappyPath(t *testing.T) {
 		err = planet.StopPeer(planet.StorageNodes[1])
 		require.NoError(t, err)
 
-		// remove stopped nodes from overlay cache
-		err = planet.Satellites[0].Overlay.Service.Delete(ctx, planet.StorageNodes[0].ID())
+		// mark stopped nodes as offline in overlay cache
+		_, err = planet.Satellites[0].Overlay.Service.UpdateUptime(ctx, planet.StorageNodes[0].ID(), false)
 		require.NoError(t, err)
-		err = planet.Satellites[0].Overlay.Service.Delete(ctx, planet.StorageNodes[1].ID())
+		_, err = planet.Satellites[0].Overlay.Service.UpdateUptime(ctx, planet.StorageNodes[1].ID(), false)
```
mobyvb: this also should be done in `pkg/storage/segments/repairer_test.go`

satellite/satellitedb/overlaycache.go
```diff
@@ -466,8 +466,11 @@ func (cache *overlaycache) UpdateOperator(ctx context.Context, nodeID storj.Node
 	}
 
 	updated, err := convertDBNode(updatedDBNode)
+	if err != nil {
+		return nil, Error.Wrap(err)
```
kaloyan-raev: The transaction must be rolled back in this case. I pushed a fix.

pkg/overlay/cache.go
```diff
@@ -19,6 +19,8 @@ import (
 const (
 	// OverlayBucket is the string representing the bucket used for a bolt-backed overlay dht cache
 	OverlayBucket = "overlay"
+	// IsOnlineWindow is the maximum amount of time that can pass without seeing a node before that node is considered offline
+	IsOnlineWindow = 1 * time.Hour
```
kaloyan-raev: Great! I just renamed it to `OnlineWindow`. The "is" prefix in names is usually used for boolean flags and even often omitted for them. It also reads better this way, e.g. "if the duration between now and last contact success time is less than the online window...".

pkg/overlay/cache.go
kaloyan-raev: It's now available as `overlay.OnlineWindow` const.

pkg/audit/newverifier_test.go
kaloyan-raev: Good point! I updated this test too.

Do you think we should enable TestSegmentStoreRepair back? It looks stable if I run it locally a few hundred times.

kaloyan-raev: Oh. I see you already enabled it with #1557.

satellite/satellitedb/overlaycache.go
```diff
@@ -469,7 +447,7 @@ func (cache *overlaycache) UpdateStats(ctx context.Context, updateReq *overlay.U
 }
 
 // UpdateOperator updates the email and wallet for a given node ID for satellite payments.
-func (cache *overlaycache) UpdateOperator(ctx context.Context, nodeID storj.NodeID, operator pb.NodeOperator) (stats *overlay.NodeStats, err error) {
+func (cache *overlaycache) UpdateOperator(ctx context.Context, nodeID storj.NodeID, operator pb.NodeOperator) (stats *overlay.NodeDossier, err error) {
 	defer mon.Task()(&ctx)(&err)
 
 	tx, err := cache.db.Open(ctx)
```
egonelbre: I think we can remove transactions here completely.

kaloyan-raev: Not really. UpdateOperator may be invoked only with the new Wallet info and `nil` for Email.

We need the transaction, but make better use of it. I will push a fix.

kaloyan-raev: Well, actually, you are right. We don't need the transaction, but we also need to fix handing nil values.

kaloyan-raev: Wallet and Email are just strings and not pointers. So I checked all places where UpdateOperator is invoked and made sure both Email and Wallet are provided for the update. This way we won't lose one of the values if providing only the other for the update.

I also removed the transaction.

pkg/overlay/cache.go
navillasa: sounds good!

satellite/satellitedb/overlaycache.go
navillasa: I don't think I'm seeing the fix?

wthorp:
will reopen : verification/cla-signed  stalled?

satellite/satellitedb/overlaycache.go
kaloyan-raev: This is because a later commit actually removed the transaction in this method. So there is no need of commit and rollback anymore.

navillasa: ah okay!

PR #1638, opened 2019-04-02
plumb EncryptionScheme, RedundancyScheme through to buckets

thepaul:
We want to use those fields in the bucket-level Pointer objects as bucket defaults, but we need to be able to get at them first.

I don't see any strong reason not to make these available, except that it was kind of a pain, but now that's done.

Thanks for submitting a PR!


pkg/storage/objects/store.go
```diff
@@ -24,10 +24,13 @@ var mon = monkit.Package()
...
-	Modified   time.Time
-	Expiration time.Time
-	Size       int64
-	Checksum   string
+	Modified         time.Time
+	Expiration       time.Time
+	Size             int64
+	Checksum         string
+	SegmentsSize     int64
+	RedundancyScheme storj.RedundancyScheme
```
jtolds: this is so weird but we do have a bit of a mismatch where the redundancy scheme is technically per segment and not per object, so that's weird. this is fine though

thepaul: Yeah, I was wondering about that. Do we actually want to support different redundancy between segments of an object? Unless we have tests for it already, I suspect it won't work right yet.

What if we renamed this field to LastRedundancyScheme to try and make clear that it only really applies to one of the segments?

jtolds: no i don't think we do want to support it but it's kind of an unfortunate property of our data model. i think it's fine as is but maybe it's worth a comment?

thepaul: Ok. I'll get that in as part of #1568.

PR #1639, opened 2019-04-02
Add additional error handling to email message sending.

simongui:
This PR adds some additional error handling where we were ignoring errors when sending emails from the Satellite.

Thanks for submitting a PR!


internal/post/message.go
```diff
@@ -70,14 +70,24 @@ func (msg *Message) Bytes() []byte {
 		var sub io.Writer
 
 		if len(msg.PlainText) > 0 {
-			sub, _ = wr.CreatePart(textproto.MIMEHeader{
+			sub, err := wr.CreatePart(textproto.MIMEHeader{
 				"Content-Type":              []string{"text/plain; charset=UTF-8; format=flowed"},
 				"Content-Transfer-Encoding": []string{"quoted-printable"},
 			})
+			if err != nil {
+				return nil, err
```
egonelbre: Let's also use error wrapping to be able to trace back to the location where the error happened.

internal/post/message.go
```diff
@@ -70,14 +70,24 @@ func (msg *Message) Bytes() []byte {
...
 
 			enc := quotedprintable.NewWriter(sub)
-			_, _ = enc.Write([]byte(msg.PlainText))
-			_ = enc.Close()
+			_, err = enc.Write([]byte(msg.PlainText))
+			if err != nil {
+				return nil, err
+			}
+
+			err = enc.Close()
```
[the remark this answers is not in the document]
simongui: That would hide the error. That `enc.Close()` function can return an error. I don't think we want to silently ignore those.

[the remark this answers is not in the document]
thepaul: you can still capture an error inside defer. one common pattern in our code is

```go
    defer func() { err = errs.Combine(err, thing.Close()) }()
```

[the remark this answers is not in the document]
simongui: Ah interesting! I never thought of doing it that way. What happens if in later code before the defer executes an error happens? Is that what the `errs.Combine()` call is for to combine the first error with the error happening in the defer call?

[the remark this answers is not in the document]
ifraixedes: > Combine combines multiple errors into a single error.

It's the doc text of such function https://godoc.org/github.com/zeebo/errs#Combine

It's missing that if `err` are nil they are skipped.

internal/post/message.go
```diff
@@ -93,18 +105,26 @@ func (msg *Message) Bytes() []byte {
 			fmt.Fprint(sub, part.Content)
 		}
 
-		_ = wr.Close()
+		err := wr.Close()
```
ifraixedes: This should be deferred after the assignation so guarantee that it always gets close.

Use the same pattern than above `defer func() { err = errs.Combine(err, ...) }()`

internal/post/message.go
```diff
@@ -70,14 +75,21 @@ func (msg *Message) Bytes() []byte {
...
 
 			enc := quotedprintable.NewWriter(sub)
-			_, _ = enc.Write([]byte(msg.PlainText))
-			_ = enc.Close()
+			_, err = enc.Write([]byte(msg.PlainText))
+			if err != nil {
+				return nil, Error.Wrap(err)
+			}
+
+			defer func() { err = errs.Combine(err, enc.Close()) }()
```
ifraixedes: This defer should be just after the `enc` initialization for guaranteeing that always gets closed.

internal/post/message.go
```diff
@@ -93,18 +105,26 @@ func (msg *Message) Bytes() []byte {
...
 		fmt.Fprintf(&body, "Content-Transfer-Encoding: quoted-printable\r\n\r\n")
 
 		enc := quotedprintable.NewWriter(&body)
-		_, _ = enc.Write([]byte(msg.PlainText))
-		_ = enc.Close()
+		if _, err := enc.Write([]byte(msg.PlainText)); err != nil {
+			return nil, Error.Wrap(err)
+		}
+
+		if err := enc.Close(); err != nil {
```
ifraixedes: Same reason than before, defer the close after the `enc` initialization.

simongui: @ifraixedes Can I get a final review please? :) Thanks!

ifraixedes: Only this one is missing to be deferred after `enc := ...` in the line 102 using the `errs.Combine` as you used above.

simongui: Oops missed that one. Good catch. Done!

internal/post/sender.go
```diff
@@ -76,7 +76,12 @@ func (sender *SMTPSender) SendEmail(msg *Message) error {
 		err = errs.Combine(err, data.Close())
 	}()
 
-	_, err = data.Write(msg.Bytes())
+	mess, err := msg.Bytes()
+	if err != nil {
+		return err
```
ifraixedes: This should be `return Error.Wrap(err)` however all the errors of this function are returned as they so I would suggest to leave it as it's now and open an issue in Jira to revisit all the methods and functions of this package to check the ones that should be to be `Error.Wrap(..)` which they will probably be the most of them

PR #1648, opened 2019-04-02
Add Version Information into KAD Network and SatelliteDB & Change Selection Process

stefanbenten:
This PR updates node selection to consider minimum node versions. Right now the network is full of out-of-date nodes where uploads fail. This change makes it so we select nodes that will work.


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -128,6 +128,13 @@ model node (
 	field free_bandwidth int64 ( updatable )
 	field free_disk      int64 ( updatable )
 
+ 	field major             int64 ( updatable )
+	field minor             int64 ( updatable )
+	field patch             int64 ( updatable )
```
jtolds: i think  major, minor, patch should just be combined into a version string

jtolds: hmm, i guess that makes it hard to sort in the table.  nvm

stefanbenten: We figured it out and it works great now!

satellite/peer.go
```diff
@@ -222,6 +231,8 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config) (*
 			AuditCount:            config.Node.AuditCount,
 			NewNodeAuditThreshold: config.Node.NewNodeAuditThreshold,
 			NewNodePercentage:     config.Node.NewNodePercentage,
+			//ToDo: Add Custom Version for Node Selection instead of the same as the satellite runs on
+			Version: versionInfo.Version.String(),
```
stefanbenten: @jtolds This is currently setting the selection version for the Nodeselection to be equal to the Satellite Version.
Do we wanna add a flag for that?

pkg/pb/node.proto
```diff
@@ -80,3 +81,13 @@ message NodeRestrictions {
     int64 free_bandwidth = 1;
     int64 free_disk = 2;
 }
+
+// NodeVersion contains
+message NodeVersion {
+    int64 major = 1;
+    int64 minor = 2;
+    int64 patch = 3;
```
jtolds: i don't know if you saw the conversation with @brimstone in #engineering, but i think we need to make the version a string, so we can in the future support versions like "v3.0.0-beta1".

pkg/pb/node.proto
```diff
@@ -80,3 +81,13 @@ message NodeRestrictions {
...
+    string hash = 4;
+    int64 timestamp = 5;
```
jtolds: can this be `google.protobuf.Timestamp timestamp = 5;`? you'll need to use https://godoc.org/github.com/golang/protobuf/ptypes#Timestamp and https://godoc.org/github.com/golang/protobuf/ptypes#TimestampProto for conversion

satellite/peer.go
jtolds: yeah i think we want a flag that defaults to the satellite version

satellite/satellitedb/dbx/satellitedb.dbx
jtolds: yep, @brimstone's conversation convinced me this should be a string also

satellite/satellitedb/migrate.go
```diff
@@ -530,6 +530,28 @@ func (db *DB) PostgresMigration() *migrate.Migration {
...
+					ALTER TABLE nodes ADD minor bigint;
+					ALTER TABLE nodes ADD patch bigint;
+					ALTER TABLE nodes ADD hash TEXT;
+					ALTER TABLE nodes ADD timestamp bigint;
+					ALTER TABLE nodes ADD release bool;
+					UPDATE nodes SET major = 0, minor = 1, patch = 0;
+					UPDATE nodes SET hash = '';
+					UPDATE nodes SET timestamp = 0;
+					UPDATE nodes SET release = false;
+					ALTER TABLE nodes ALTER COLUMN major SET NOT NULL;
```
jtolds: we should get rid of `NOT NULL` on all of these

satellite/satellitedb/dbx/satellitedb.dbx
jtolds: oh wait the query dang it. okay, leave this as separate fields but i still think the protobuf should be a string

satellite/satellitedb/migrate.go
jtolds: they can definitely be null

satellite/peer.go
stefanbenten: The Overlay cache seems to have a default value that isnt used. Is that possible?

pkg/overlay/config.go
```diff
@@ -33,6 +33,7 @@ type LookupConfig struct {
 // NodeSelectionConfig is a configuration struct to determine the minimum
 // values for nodes to select
 type NodeSelectionConfig struct {
+	Version           string  `help:"set to minimum node software version" default:"v0.1.0"`
```
jtolds: oh wait this is going to break a lot of things, right? shouldn't we make selection by version opt-in?

stefanbenten: Thats what i think is never being used. Its always being overwritten by the peer setup

satellite/satellitedb/migrate.go
jtolds: okay nevermind, either i can't figure out dbx or dbx makes this hard. i get it

jtolds:
i think this is ready to go

satellite/peer.go
jtolds: i hooked up the flag to the existing config struct

pkg/pb/node.proto
jtolds: done

pkg/pb/node.proto
jtolds: done

pkg/overlay/cache.go
```diff
@@ -63,7 +61,7 @@ type DB interface {
 	// UpdateStats all parts of single storagenode's stats.
 	UpdateStats(ctx context.Context, request *UpdateRequest) (stats *NodeStats, err error)
 	// UpdateOperator updates the email and wallet for a given node ID for satellite payments.
```
kaloyan-raev: Update doc comment.

satellite/satellitedb/migrate.go
```diff
@@ -530,14 +530,36 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					`ALTER TABLE bucket_bandwidth_rollups ADD CONSTRAINT bucket_bandwidth_rollups_pk PRIMARY KEY (bucket_name, project_id, interval_start, action);`,
 				},
 			},
+			{
+				Description: "Add new Columns to store version information",
+				Version:     14,
+				Action: migrate.SQL{
+					`ALTER TABLE nodes ADD major bigint;
```
kaloyan-raev: You can do `ALTER TABLE nodes ADD major BIGINT NOT NULL DEFAULT 0` as a single statement instead of doing `ADD`, `UPDATE` and `ALTER COLUMN`.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -128,6 +128,13 @@ model node (
 	field free_bandwidth int64 ( updatable )
 	field free_disk      int64 ( updatable )
 
+ 	field major             int64 ( updatable )
```
kaloyan-raev: Why not `int` instead of `int64`. It seems `int` works already pretty well for the `type` field.

satellite/satellitedb/overlaycache.go
```diff
@@ -459,13 +519,36 @@ func (cache *overlaycache) UpdateStats(ctx context.Context, updateReq *overlay.U
...
+			updateFields.Email = dbx.Node_Email(nodeInfo.GetOperator().GetEmail())
+		}
+		if nodeInfo.GetCapacity() != nil {
+			updateFields.FreeDisk = dbx.Node_FreeDisk(nodeInfo.GetCapacity().GetFreeDisk())
+			updateFields.FreeBandwidth = dbx.Node_FreeBandwidth(nodeInfo.GetCapacity().GetFreeBandwidth())
+		}
+		if nodeInfo.GetVersion() != nil {
+			semVer, err := version.NewSemVer(nodeInfo.GetVersion().GetVersion())
+			if err != nil {
+				semVer = &version.SemVer{}
```
kaloyan-raev: We should return the error here. Otherwise, it can result into hard to identify bugs.

satellite/satellitedb/overlaycache.go
```diff
@@ -459,13 +519,36 @@ func (cache *overlaycache) UpdateStats(ctx context.Context, updateReq *overlay.U
...
+			}
+			pbts, err := ptypes.Timestamp(nodeInfo.GetVersion().GetTimestamp())
+			if err != nil {
+				pbts = time.Time{}
```
kaloyan-raev: Same here - return the error.

PR #1649, opened 2019-04-02
Filehealth cli

aleitner:
* Move calculations from satellite side to be done on the client side. This provides more customizability and reduces sending of unnecessary data.
* Inspector methods for checking object and segment health can both print in csv format and write to a csv file.

Segment Health
```
inspector --identity-path="satellite-identity" --address="satellite-private-address" health segment project-id -- segment-index bucket-name encrypted-path
```

Object Health
```
inspector --identity-path="satellite-identity" --address="satellite-private-address" health object project-id bucket-name encrypted-path
```

The output is by default written to `os.Stdout`, but there is also an optional `csv-path` that can be specified.

```
inspector --identity-path="satellite-identity" --address="satellite-private-address" health --csv-path="path-to-new-csv"  object project-id bucket-name encrypted-path
```
Screen Shot 2019-04-02 at 5 07 36 PM


jenlij:
Could you note the "What" and "Why" of these changes?

aleitner:
> Could you note the "What" and "Why" of these changes?

What: Inspector tool for checking object and segment stats by a pointerdb path.
Why: Devops can use this data for tracking data stability

jenlij:
Is there somewhere we can add the documentation above for reference since it'll be hard to find once this PR is merged?

aleitner:
> Is there somewhere we can add the documentation above for reference since it'll be hard to find once this PR is merged?

as far as I know our code is the documentation. There's also a help command

cmd/inspector/main.go
```diff
@@ -35,6 +36,9 @@ var (
 	// IdentityPath is the path to the identity the inspector should use for network communication
 	IdentityPath = flag.String("identity-path", "", "path to the identity certificate for use on the network")
 
+	// CSVPath is the csv path where command output is written
+	CSVPath = flag.String("csv-path", "stdout", "csv path where command output is written")
```
mniewrzal: Can you add this only to health command by `healthCmd.Flags().StringVar(&CSVPath, "csv-path", "stdout", "csv path where command output is written")` in `init` method?

egonelbre: Why do we even need an output file, people can direct the output a csv as needed.

cmd/inspector/main.go
```diff
@@ -472,20 +476,73 @@ func ObjectHealth(cmd *cobra.Command, args []string) (err error) {
 		return ErrArgs.Wrap(err)
 	}
 
+	startAfterSegment := int64(0) // start from first segment
```
egonelbre: this file is definitely too large

aleitner: you mean inspector/main.go has too many lines?

PR #1650, opened 2019-04-02
Fix repair order limit creation

mobyvb:
Repair order limits should use piece size rather than erasure share size. Counter should also be incremented upon successful addition of order limits.
Thanks for submitting a PR!


satellite/orders/service.go
```diff
@@ -393,7 +393,11 @@ func (service *Service) CreateAuditOrderLimits(ctx context.Context, auditor *ide
 func (service *Service) CreateGetRepairOrderLimits(ctx context.Context, repairer *identity.PeerIdentity, bucketID []byte, pointer *pb.Pointer, healthy []*pb.RemotePiece) (_ []*pb.AddressedOrderLimit, err error) {
 	rootPieceID := pointer.GetRemote().RootPieceId
 	redundancy := pointer.GetRemote().GetRedundancy()
-	shareSize := redundancy.GetErasureShareSize()
+	eeRedundancy, err := eestream.NewRedundancyStrategyFromProto(redundancy)
```
kaloyan-raev: Perhaps, we can avoid having both `redundancy` and `eeRedundancy` variables and initialize one single var like this:
```go
redundancy, err := eestream.NewRedundancyStrategyFromProto(pointer.GetRemote().GetRedundancy())
```
Then, if you need the total number of pieces, you can just call `redundancy.TotalCount()`.

PR #1657, opened 2019-04-03
update ProjectStorageTotals func to get all records for project

JessicaGreben:
This PR fixes an error with the query that finds the current total for at rest data for a project ID. Previously we were only getting one storage tally record, but we actually needed to get all the storage tally records for a projectID and a intervalStart. The PR updates the test and confirms this behavior.


satellite/satellitedb/accounting.go
```diff
@@ -39,14 +39,34 @@ func (db *accountingDB) ProjectBandwidthTotal(ctx context.Context, bucketID []by
...
-	if err != nil || rollup == nil {
+	if err != nil || latestRollup == nil {
 		return 0, 0, err
 	}
-	return int64(rollup.Inline), int64(rollup.Remote), err
+
+	// Sum all the inline and remote values for the project that all share the same interval_start.
+	// All records for a project that have the same interval start are part of the same tally run.
+	// This should represent the most recent calculation of a project's total at rest storage.
+	var inlineSum *int64
```
cam-a: Do these need to be pointers?

satellite/satellitedb/accounting.go
```diff
@@ -39,14 +39,34 @@ func (db *accountingDB) ProjectBandwidthTotal(ctx context.Context, bucketID []by
...
+	query := `SELECT SUM(inline) FROM bucket_storage_tallies WHERE project_id = ? AND interval_start = ?;`
+	err = db.db.QueryRow(db.db.Rebind(query), projectID, latestRollup.IntervalStart).Scan(&inlineSum)
+	if err != nil || inlineSum == nil {
+		return 0, 0, nil
+	}
+
+	var remoteSum *int64
+	query = `SELECT SUM(remote) FROM bucket_storage_tallies WHERE project_id = ? AND interval_start = ?;`
```
cam-a: Can we combine these into a single query?

satellite/satellitedb/accounting.go
JessicaGreben: see: https://github.com/storj/storj/pull/1586#discussion_r269587590

satellite/satellitedb/accounting.go
JessicaGreben: probably

satellite/satellitedb/accounting.go
jtolds: while pointers definitely work here, they unfortunately must get heap allocated for this to work. it's sliiiightly better to use the sql.NullInt64 type here. so

```
type inlineSum sql.NullInt64
....Scan(&inlineSum)
if err != nil || !inlineSum.Valid { ...
```

satellite/satellitedb/accounting.go
```diff
@@ -39,14 +39,34 @@ func (db *accountingDB) ProjectBandwidthTotal(ctx context.Context, bucketID []by
...
 		return 0, 0, err
 	}
-	return int64(rollup.Inline), int64(rollup.Remote), err
+
+	// Sum all the inline and remote values for the project that all share the same interval_start.
+	// All records for a project that have the same interval start are part of the same tally run.
+	// This should represent the most recent calculation of a project's total at rest storage.
+	var inlineSum *int64
+	query := `SELECT SUM(inline) FROM bucket_storage_tallies WHERE project_id = ? AND interval_start = ?;`
+	err = db.db.QueryRow(db.db.Rebind(query), projectID, latestRollup.IntervalStart).Scan(&inlineSum)
```
jtolds: what is Rebind doing here?

satellite/satellitedb/accounting.go
jtolds:
```
SELECT SUM(remote), SUM(inline) FROM bucket_storage_tallies WHERE project_id = ? AND interval_start = ?;
```

since you're doing a custom query anyways:

```
SELECT interval_start, SUM(remote), SUM(inline) FROM bucket_storage_tallies WHERE project_id = ? ORDER BY interval_start DESC LIMIT 1;
```
then you can get rid of the original query to find the latest interval_start also

satellite/satellitedb/accounting.go
JessicaGreben: rebind formats the queries (the `?` part) for sqlite and postgres.

satellite/satellitedb/accounting.go
JessicaGreben: i don't think we want to limit 1, I think that is the problem we were having before.  I think we want all the records with a specific `interval_start` and `project_id`

satellite/satellitedb/accounting.go
JessicaGreben: thanks for the tip jt, this `sql.NullInt64` is cool

PR #1658, opened 2019-04-03
use SerializableMeta to store bucket attributes

thepaul:
It doesn't work well to use the bucket-level pointer to store the default redundancy strategy, as such pointers are inline pointers and don't have a RemoteSegment field at all. Also, it was proving to be an enormous pain to plumb the information both up and down through existing APIs without changing too much else.

So instead, this change undoes most of commit e4a70e3, and makes use of the arbitrary-string-key-value map in the SerializableMeta structure.

This strategy has the obvious drawback of requiring a bunch of integer string formatting and string parsing every time we check the metadata on a bucket (which is, I think, every time we get any object). If it proves to be a performance problem, we can revisit.

Thanks for submitting a PR!


pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
 // convertMeta converts stream metadata to object metadata
-func convertMeta(m objects.Meta) (Meta, error) {
-	var cipher storj.Cipher
-
-	pathEncType := m.UserDefined["path-enc-type"]
+func convertMeta(m objects.Meta) (out Meta, err error) {
+	out.Created = m.Modified
+	// backwards compatibility for old buckets
+	out.PathEncryptionType = storj.AESGCM
+	out.EncryptionScheme.Cipher = storj.Invalid
```
kaloyan-raev: I doubt that defaulting the `storj.Invalid` helps backward compatibiltiy. `storj.EncAESGCM` is a much better shot.

pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
 		if err != nil {
-			return Meta{}, err
+			return Meta{}, errs.New("invalid metadata field for path-enc-type: %v", err)
+		}
+		out.PathEncryptionType = storj.Cipher(pet)
+	}
+	if defaultSegSize := m.UserDefined["default-seg-size"]; defaultSegSize != "" {
+		dss, err := strconv.ParseInt(defaultSegSize, 10, 64)
+		if err != nil {
+			return Meta{}, errs.New("invalid metadata field for path-enc-type: %v", err)
```
kaloyan-raev: Error message should be about `default-seg-size`.

pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
+			return Meta{}, errs.New("invalid metadata field for default-rs-sharsz: %v", err)
+		}
+		out.RedundancyScheme.ShareSize = int32(drs)
+	}
+	if defaultRSRequired := m.UserDefined["default-rs-reqd"]; defaultRSRequired != "" {
+		drr, err := strconv.ParseInt(defaultRSRequired, 10, 16)
+		if err != nil {
+			return Meta{}, errs.New("invalid metadata field for default-rs-reqd: %v", err)
+		}
+		out.RedundancyScheme.RepairShares = int16(drr)
```
kaloyan-raev: This value should be assigned to `RequiredShares`, not to `RepairShares`.

pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
+			return Meta{}, errs.New("invalid metadata field for default-rs-repair: %v", err)
+		}
+		out.RedundancyScheme.RepairShares = int16(drr)
+	}
+	if defaultRSOptimal := m.UserDefined["default-rs-optim"]; defaultRSOptimal != "" {
+		drr, err := strconv.ParseInt(defaultRSOptimal, 10, 16)
+		if err != nil {
+			return Meta{}, errs.New("invalid metadata field for default-rs-optim: %v", err)
+		}
+		out.RedundancyScheme.RepairShares = int16(drr)
```
kaloyan-raev: This value should be assigned to `OptimalShares`, not to `RepairShares`.

pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
+	}
+	if defaultRSTotal := m.UserDefined["default-rs-total"]; defaultRSTotal != "" {
+		drr, err := strconv.ParseInt(defaultRSTotal, 10, 16)
+		if err != nil {
+			return Meta{}, errs.New("invalid metadata field for default-rs-total: %v", err)
 		}
-		cipher = storj.Cipher(pet)
+		out.RedundancyScheme.RepairShares = int16(drr)
```
kaloyan-raev: This value should be assigned to `TotalShares`, not to `RepairShares`.

pkg/storage/buckets/store.go
```diff
@@ -166,27 +182,82 @@ func (b *BucketStore) List(ctx context.Context, startAfter, endBefore string, li
...
+	out.Created = m.Modified
+	// backwards compatibility for old buckets
+	out.PathEncryptionType = storj.AESGCM
+	out.EncryptionScheme.Cipher = storj.Invalid
 
-	if pathEncType == "" {
-		// backward compatibility for old buckets
-		cipher = storj.AESGCM
-	} else {
+	if pathEncType := m.UserDefined["path-enc-type"]; pathEncType != "" {
```
kaloyan-raev: Shouldn't we get values from the map using the pattern `value, ok := map["key"]` to avoid panics if the key does not exist?

pkg/metainfo/kvmetainfo/buckets.go
```diff
@@ -18,7 +18,10 @@ func (db *Project) CreateBucket(ctx context.Context, bucketName string, info *st
 		return storj.Bucket{}, storj.ErrNoBucket.New("")
 	}
 
-	meta, err := db.buckets.Put(ctx, bucketName, getPathCipher(info))
+	newBucketMeta := buckets.Meta{
+		PathEncryptionType: getPathCipher(info),
```
kaloyan-raev: We need to fill all other fields of `buckets.Meta`.

Also, we need to provide default values if `info == nil`.

pkg/metainfo/kvmetainfo/buckets.go
```diff
@@ -106,11 +109,8 @@ func getPathCipher(info *storj.Bucket) storj.Cipher {
...
 	return storj.Bucket{
-		Name:             bucketName,
-		Created:          meta.Created,
-		PathCipher:       meta.PathEncryptionType,
-		SegmentsSize:     meta.SegmentsSize,
-		RedundancyScheme: meta.RedundancyScheme,
-		EncryptionScheme: meta.EncryptionScheme,
+		Name:       bucketName,
+		Created:    meta.Created,
+		PathCipher: meta.PathEncryptionType,
```
kaloyan-raev: We still need this part:
```go
SegmentsSize:     meta.SegmentsSize,
RedundancyScheme: meta.RedundancyScheme,
EncryptionScheme: meta.EncryptionScheme,
```

lib/uplink/project.go
```diff
@@ -182,19 +185,18 @@ func (p *Project) OpenBucket(ctx context.Context, bucket string, access *Encrypt
...
 
 	buckets := buckets.NewStore(streams)
 
 	return &Bucket{
-		Bucket:     bucketInfo,
-		Config:     *cfg,
-		metainfo:   kvmetainfo.New(p.metainfo, buckets, streams, segments, key),
-		streams:    streams,
-		pathCipher: pathCipher,
+		BucketConfig: *cfg,
```
kaloyan-raev: We also need the `Name` field to be set here.

pkg/storage/buckets/store.go
thepaul: Invalid makes more sense as a fall-back bucket default, because we can tell whether or not there actually is a bucket default set.

pkg/storage/buckets/store.go
thepaul: wow, i must have been half asleep when i did all this

pkg/storage/buckets/store.go
thepaul: Accessing a key that doesn't exist won't cause a panic, it will only return a zero-value. The behavior we want if an empty string is really stored is the same behavior we want if the key does not exist, so I think this is probably best still

pkg/metainfo/kvmetainfo/buckets.go
thepaul: Well spotted. Thanks, will fix

lib/uplink/bucket_attrs_test.go
```diff
@@ -0,0 +1,189 @@
...
+	planet.Start(ctx)
+
+	// we only use testUplink for the free API key, until such time
+	// as testplanet makes it easy to get another way :D
+	testUplink := planet.Uplinks[0]
+	satellite := planet.Satellites[0]
+	cfg.uplinkCfg.Volatile.TLS.SkipPeerCAWhitelist = true
+
+	apiKey, err := ParseAPIKey(testUplink.APIKey[satellite.ID()])
+	if err != nil {
```
aleitner: try using require.NoError

lib/uplink/bucket_attrs_test.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Fatalf("could not parse API key from testplanet: %v", err)
+	}
+	uplink, err := NewUplink(ctx, &cfg.uplinkCfg)
+	if err != nil {
```
aleitner: try using require.NoError

lib/uplink/bucket_attrs_test.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Fatalf("could not create new Uplink object: %v", err)
+	}
+	defer ctx.Check(uplink.Close)
+	proj, err := uplink.OpenProject(ctx, satellite.Addr(), apiKey)
+	if err != nil {
```
aleitner: try using require.NoError

lib/uplink/bucket_attrs_test.go
```diff
@@ -0,0 +1,189 @@
...
+	testPlanetWithLibUplink(t, testConfig,
+		func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet, proj *Project) {
+			_, err := proj.CreateBucket(ctx, bucketName, &inBucketConfig)
+			require.NoError(t, err)
+
+			bucket, err := proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
+			defer ctx.Check(bucket.Close)
+
+			{
```
aleitner: Does this need to be indented? None of the following lines are indented

lib/uplink/uplink.go
```diff
@@ -68,6 +78,11 @@ func (c *Config) setDefaults(ctx context.Context) error {
 	if c.Volatile.MaxInlineSize.Int() == 0 {
 		c.Volatile.MaxInlineSize = 4 * memory.KiB
 	}
+	if c.Volatile.MaxMem.Int() == 0 {
+		c.Volatile.MaxMem = 4 * memory.MiB
+	} else if c.Volatile.MaxMem.Int() < 0 {
+		c.Volatile.MaxMem = 0
```
aleitner: Or should we just return an error here?

lib/uplink/uplink.go
```diff
@@ -112,14 +127,28 @@ func (u *Uplink) OpenProject(ctx context.Context, satelliteAddr string, apiKey A
...
 
-	// TODO: we shouldn't need segment or stream stores to manage buckets
-	segments := segments.NewSegmentStore(metainfo, nil, eestream.RedundancyStrategy{}, maxBucketMetaSize.Int(), maxBucketMetaSize.Int64())
-	streams, err := streams.NewStreamStore(segments, maxBucketMetaSize.Int64(), nil, 0, storj.Unencrypted)
+	// TODO: we shouldn't need segment or stream stores to manage buckets.
+	// We won't make any use of them. But you can't skimp on any of this
+	// setup, or things will break as they stand now.
+	encryptionKeyThatWeWontEvenUse := &storj.Key{}
+	whoCares := 1
+	fc, err := infectious.NewFEC(whoCares, whoCares)
```
aleitner: 🤔

lib/uplink/uplink.go
```diff
@@ -68,6 +78,11 @@ func (c *Config) setDefaults(ctx context.Context) error {
 	if c.Volatile.MaxInlineSize.Int() == 0 {
 		c.Volatile.MaxInlineSize = 4 * memory.KiB
 	}
+	if c.Volatile.MaxMem.Int() == 0 {
+		c.Volatile.MaxMem = 4 * memory.MiB
```
aleitner: maybe we should check `if c.Volatile.MaxMem.Int() <= 0`

pkg/metainfo/kvmetainfo/buckets.go
```diff
@@ -17,8 +17,19 @@ func (db *Project) CreateBucket(ctx context.Context, bucketName string, info *st
 	if bucketName == "" {
 		return storj.Bucket{}, storj.ErrNoBucket.New("")
 	}
+	if info == nil {
+		info = &storj.Bucket{PathCipher: storj.AESGCM}
+	}
+	if info.EncryptionParameters.CipherSuite == storj.EncUnspecified {
+		info.EncryptionParameters.CipherSuite = storj.EncAESGCM
+	}
```
kaloyan-raev: How about all the rest of the `storj.Bucket` fields? Shouldn't we set them to some sane default values if `info == nil` or have zero values to in `info`?

lib/uplink/uplink.go
```diff
@@ -112,14 +127,28 @@ func (u *Uplink) OpenProject(ctx context.Context, satelliteAddr string, apiKey A
 		return nil, err
 	}
 
-	// TODO: we shouldn't need segment or stream stores to manage buckets
-	segments := segments.NewSegmentStore(metainfo, nil, eestream.RedundancyStrategy{}, maxBucketMetaSize.Int(), maxBucketMetaSize.Int64())
-	streams, err := streams.NewStreamStore(segments, maxBucketMetaSize.Int64(), nil, 0, storj.Unencrypted)
+	// TODO: we shouldn't need segment or stream stores to manage buckets.
+	// We won't make any use of them. But you can't skimp on any of this
+	// setup, or things will break as they stand now.
+	encryptionKeyThatWeWontEvenUse := &storj.Key{}
```
kaloyan-raev: We have to initialize the stream store with a meaningful encryption key. Otherwise, it won't be able to decrypt the bucket metainfo for buckets that were not created with libuplink. This is one of the reasons for failing some of the gateway tests.

We can resolve this issue by having an addition `EncKey` field to `uplink.Config.Volatile`.

lib/uplink/bucket_attrs_test.go
thepaul: stylistically, I don't like require.NoError here, because it implies it's just one of the conditions we intend to test. but this function is more of a harness: it is setting up a test environment and then running the actual test.

if it _really_ bothers you, require.NoError would work fine, but it's unhelpful here imo

lib/uplink/bucket_attrs_test.go
thepaul: yes, for scope. this scoping ensures that `buf` goes away at the end of the block and that nothing else is using it or depending on it later on.

lib/uplink/uplink.go
thepaul: the intended behavior, documented up on lines 60-65, is that if you don't set this field, we get a zero value and we want to treat that as "use the library default". but we do still want to make it possible for the user to select the "use as little memory as possible" option supported by `ecclient.NewClient`. we can't use the zero value for that option because zero is the default and the default has to be to do the most reasonable thing (library default).

that's why we set MaxMem to 0 if the user passes a negative value (that's what `ecclient.NewClient` accepts if you want to tell it to allocate as little memory as possible) and we set it to 4MiB if the user passes the zero value.

lib/uplink/uplink.go
thepaul: :tada:

pkg/metainfo/kvmetainfo/buckets.go
thepaul: It looked to me like zero values in all the other positions were expected or supported. Any fields in particular?

kaloyan-raev: We should check and set defaults at least for the redundancy scheme. Otherwise, a zero redundancy scheme will be stored in the bucket metadata.

kaloyan-raev: Actually, I've just checked that we actually need all encrypted block size, redundancy scheme, and segments size set properly here.

kaloyan-raev: I opened #1710 to be merged to this PR with suggestion how to set correct default for bucket metadata.

thepaul: For my own curiosity, why does a bucket need to have a nonzero redundancy strategy?

lib/uplink/bucket_attrs_test.go
```diff
@@ -0,0 +1,189 @@
...
+		cfg.planetCfg = &testplanet.Config{SatelliteCount: 1, StorageNodeCount: 5, UplinkCount: 1}
+	}
+
+	planet, err := testplanet.NewCustom(zaptest.NewLogger(t), *cfg.planetCfg)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := testcontext.New(t)
+	defer ctx.Check(planet.Shutdown)
+	defer ctx.Cleanup() // warning: this must be deferred after planet.Shutdown so it runs before
```
egonelbre: The order should be:
```
ctx := testcontext.New(t)
defer ctx.Cleanup()

planet, err := testplanet.NewCustom(zaptest.NewLogger(t), *cfg.planetCfg)
require.NoError(t, err)
defer ctx.Check(planet.Shutdown)

planet.Start(ctx)
```

lib/uplink/uplink.go
```diff
@@ -54,6 +56,14 @@ type Config struct {
...
 		// the inline storage and require remote storage, still.)
 		MaxInlineSize memory.Size
+
+		// MaxMem is the default maximum amount of memory to be
+		// allocated for read buffers while performing decodes of
+		// objects. (This option is overrideable per Bucket if the user
+		// so desires.) If set to zero, the library default (4 MiB) will
+		// be used. If set to a negative value, the system will use the
+		// smallest amount of memory it can.
+		MaxMem memory.Size
```
egonelbre: `MaxMemory`

pkg/metainfo/kvmetainfo/buckets.go
kaloyan-raev: `libuplink.Project.OpenBucket` initializes lots of things (like redundancy strategy) based on the bucket config we store in the bucket metadata. So we have to make sure that `kvmetainfo.Project.CreateBucket` stores good defaults in the bucket metadata if the respective config is not provided by the caller.

pkg/storage/buckets/store.go
```diff
@@ -98,28 +99,43 @@ func (b *BucketStore) Get(ctx context.Context, bucket string) (meta Meta, err er
...
+		"path-enc-type":     strconv.Itoa(int(pathCipher)),
+		"default-seg-size":  strconv.FormatInt(inMeta.SegmentsSize, 10),
+		"default-enc-type":  strconv.Itoa(int(inMeta.EncryptionScheme.Cipher)),
+		"default-enc-blksz": strconv.FormatInt(int64(inMeta.EncryptionScheme.BlockSize), 10),
+		"default-rs-algo":   strconv.Itoa(int(inMeta.RedundancyScheme.Algorithm)),
+		"default-rs-sharsz": strconv.FormatInt(int64(inMeta.RedundancyScheme.ShareSize), 10),
+		"default-rs-reqd":   strconv.Itoa(int(inMeta.RedundancyScheme.RequiredShares)),
+		"default-rs-repair": strconv.Itoa(int(inMeta.RedundancyScheme.RepairShares)),
+		"default-rs-optim":  strconv.Itoa(int(inMeta.RedundancyScheme.OptimalShares)),
+		"default-rs-total":  strconv.Itoa(int(inMeta.RedundancyScheme.TotalShares)),
```
egonelbre: I'm not really liking the shorthands and strings for the keys here. I'm wondering whether we should a have a separate protobuf for bucket information that can be serialized?

lib/uplink/bucket_attrs_test.go
egonelbre: (Note if that order doesn't work, then there's a bug somewhere.)

pkg/storage/buckets/store.go
egonelbre: This would mean we need to write a migration for that later.

kaloyan-raev: We would anyway need to think about this when finish the metainfo refactoring.

PR #1659, opened 2019-04-03
releases: update alpha dockerfiles to embed release tags

jtolds:
WIP: where do the arm binaries get built?

Thanks for submitting a PR!


kaloyan-raev:
I build ARM binaries in a Ubuntu VM where I have the necessary ARM cross-compile tooling installed.

jtolds:
@kaloyan-raev right but is there a place i can update the go build command you run? it needs to change like the amd64 one here

kaloyan-raev:
I updated the build command I execute on that VM. You can check the result on quay.io if you want.

jtolds:
it would be nice to have in source control eventually

kaloyan-raev:
My assumption has been that it is a very temporary solution. But if we put this in the version control, we need to find how to setup the build environment.

jtolds:
okay that's fair

PR #1660, opened 2019-04-03
Add sequential tests to Makefile for easier use

dylanlott:
This adds another target to our makefile that runs tests 5 times in a row with caching turned off, but failfast turned on. 

Eventually I think we should consider making this the main test command for our Travis tests to help prevent flakiness. 

Thanks for submitting a PR!


Makefile
```diff
@@ -112,6 +112,10 @@ test-all-in-one: ## Test docker images locally
 	&& $(MAKE) satellite-image storagenode-image gateway-image \
 	&& ./scripts/test-aio.sh
 
+.PHONY: test-sequential
+test-sequential: ## Run tests five times with caching turned off  and failfast
+	seq 1 5 | xargs -I{} sh -c 'go test -v ./... -count=1 -race -failfast'
```
wthorp: why can't you use -count=5 instead?

brimstone:
Why do the tests need to be run multiple times?
And why 5 times and not 2 or 10 or 9001?

dylanlott:
> Why do the tests need to be run multiple times?
> And why 5 times and not 2 or 10 or 9001?

I was going to make it take an argument, so that the number of times could be configured.

As for why we want to run it multiple times - We've talked about making sure our tests are not racy or flakey. This would help with that, but would increase time for tests to run. @jtolds has mentioned in the past that we might want to try this. I was using it a lot for my local tests, so it seemed reasonable to make a target for it. I wouldn't be opposed to making our existing `make test` target take an optional param for a number of times, either. I'm not a make wizard like you are, would love a better way to do that.

Makefile
dylanlott: You are 100% right, I'll update this to take that.

egonelbre:
Note once we have tests running on multiple platforms we will have the "repeated tests" thing effectively for free. But, this means we need actually need to have multiple platforms.

egonelbre:
I think this won't be needed, since we cannot run this currently without bogging down Travis server (for specific tests you usually want to specify the test you want to rerun). Alternatively, when we have Jenkins we get repeated tests by testing on multiple platforms.

PR #1661, opened 2019-04-04
internal/version: do version check much earlier in process initialization

jtolds:
The combination of the version service with the peer services is becoming very messy. One of the primary goals of the version service is to prevent a binary from starting if it is out of date, but as it currently stands, the version check does not start until lots of other services start in parallel. This is going to be a huge foot gun and is very racy.

This change makes the version check be one of the very first things release-build processes do (and only release-build processes) before starting anything else.

### DIFF
--- a/internal/version/service.go
+++ b/internal/version/service.go
@@ -6,6 +6,7 @@ package version
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"reflect"
 	"sync"
@@ -17,10 +18,6 @@ import (
 	"storj.io/storj/internal/sync2"
 )
 
-const (
-	errOldVersion = "Outdated Software Version, please update!"
-)
-
 // Config contains the necessary Information to check the Software Version
 type Config struct {
 	ServerAddress  string        `help:"server address to check its version against" default:"https://version.alpha.storj.io"`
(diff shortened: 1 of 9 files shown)


internal/version/service.go
```diff
@@ -30,61 +27,62 @@ type Config struct {
...
 		config:  config,
-		info:    info,
 		service: service,
+		Info:    info,
 		Loop:    sync2.NewCycle(config.CheckInterval),
-		checked: make(chan struct{}, 0),
-		allowed: false,
+		allowed: true,
+	}
+	err := srv.runOnce(ctx)
```
egonelbre: Don't do this in constructor, lift this outside the service creation logic into a separate method.

As a rule `ctx` should never be passed into a constructor.

jtolds: the goal i'm trying to achieve in this PR and in this constructor is reducing the likelihood of mistakes down the road. it is more important to build robust apis that are easy to use than to follow specific one-size-fits-all rules. moving this out of the constructor means that for correct operation, everything user must call a method to run the first version check immediately after, which needlessly introduces a possible bug location where they can forget to do so.

I'm happy to rename the method to indicate that this isn't a traditional constructor, but not having this even earlier in pkg/process already feels like a huge correctness compromise for the same reasons

jtolds:
this pr is dependent on https://github.com/storj/storj/pull/1648

internal/version/service.go
jtolds: I'll change the constructor name to InitService instead of NewService unless you can think of a better name

egonelbre: Yes, but how will you create this in testplanet? This is the reason `ctx` is not allowed in the constructor, it hinders wiring up different dependencies.

`NewService` and a separate `CheckVersion` method.

egonelbre: You can also have a separate `NewServiceWithVersionCheck`, but I'm not sure whether that would be clearer than creating the service and then calling the method.

egonelbre: also `peer.New` should be built such that it doesn't do anything until `Run` is called. It might be sufficient to call `peer.Version.IsUpToDate()` prior calling run.

stefanbenten: I am with Egon on that. I think the check is early enough

jtolds: I'll make the traditional flow for testplanet and a higher level function that does both for the cmd/ code.

jtolds: i agree that it *shouldn't* do anything, but without this change, metric sending has already started

stefanbenten: @egonelbre What about adding a similar function as in testplanet like waitforVersionCheck() ?

jtolds: we need to do the version check before the peer code even starts. metrics is surely not the only thing our process will need to set up that may communicate with disk or the external world. the masterdb also is starting up outside of the peer.

egonelbre:
```
log := zap.L()

identity, err := runCfg.Identity.Load()
if err != nil {
    zap.S().Fatal(err)
}

if err := runCfg.Verify(log); err != nil {
    log.Sugar().Error("Invalid configuration: ", err)
    return err
}

db, err := bootstrapdb.New(bootstrapdb.Config{
    Kademlia: runCfg.Kademlia.DBPath,
})
if err != nil {
    return errs.New("Error starting master database on bootstrap: %+v", err)
}

defer func() {
    err = errs.Combine(err, db.Close())
}()

peer, err := bootstrap.New(log, identity, db, runCfg)
if err != nil {
    return err
}

if peer.Version.RequestUpToDateCheck(ctx) {
    return ...
}

err = db.CreateTables()
if err != nil {
    return errs.New("Error creating tables for master database on bootstrap: %+v", err)
}

ctx := process.Ctx(cmd)
if err := process.InitMetricsWithCertPath(ctx, nil, runCfg.Identity.CertPath); err != nil {
}

runError := peer.Run(ctx)
closeError := peer.Close()

return errs.Combine(runError, closeError)
```

jtolds: we've now run migrations against disk perhaps prior to the version check

egonelbre: Peers have been designed such that you can do things before you actually start them.

jtolds: the version check is inherently a property of the process and not the peer

egonelbre: in principle "Peer" is the process, without direct persistence access...

jtolds:
this pr is no longer dependent on pr 1648

internal/version/service.go
jtolds: okay i split the constructor up. i think version checking specifically is fundamental enough that it's worth making it so we can do version checks before any other logic, even if we are ruthlessly vigilant in PRs about making constructors do nothing (though i think that's a battle the codebase will ultimately lose)

egonelbre: The constructors already should be that way, if they are not then I missed something. This is the reason why `context` is not allowed as well, it indicates that something is doing a request.

internal/version/service.go
```diff
@@ -30,61 +27,67 @@ type Config struct {
...
+	log     *zap.Logger
 	config  Config
-	info    Info
 	service string
 
+	Info Info
+
 	Loop *sync2.Cycle
 
-	checked chan struct{}
```
egonelbre: this is still needed for testplanet

jtolds: how so? nothing called IsUpToDate

egonelbre: ah, right the check was disabled... nevermind, mind still foggy

internal/version/service.go
egonelbre: After thinking about it a while, if we split protocol and binary version checking then yes, the binary version checking can be completely independent from peers.

jtolds:
https://github.com/storj/storj/pull/1666

internal/version/service.go
jtolds: hahaha oh man did we just convince each other of the other's argument

jtolds: after thinking about it i decided you were right and pushed up https://github.com/storj/storj/pull/1666

jtolds: i do think (as illustrated by my addition in 1666 of cmd/identity) that figuring out how to add a version check for some commandline tool could still use work, but i'm happy with #1666 if you'd rather merge that and keep peer constructors simple for now

PR #1662, opened 2019-04-04
Tests for upload/download in parallel

mniewrzal:
Tests for uploading and downloading separate files into the same set of storage nodes in parallel.

Thanks for submitting a PR!


internal/testplanet/uplink_test.go
```diff
@@ -110,3 +112,102 @@ func TestDownloadWithSomeNodesOffline(t *testing.T) {
...
+	}
+
+	var group errgroup.Group
+	for i, data := range dataToUpload {
+		index := strconv.Itoa(i)
+		uplink := planet.Uplinks[0]
+		satellite := planet.Satellites[0]
+
+		dataCopy := make([]byte, len(data))
+		copy(data, dataCopy)
```
wthorp: Nit: did we really need this copy()?

mniewrzal: without it there will be race, I have no better idea how to handle this nicely

egonelbre: This should suffice:
```
data := data
```

The race happens because of the shared "variable name" not because of the shared data.

mniewrzal: thx, I was pretty sure I try this.

Done.

PR #1667, opened 2019-04-04
fixed test on windows

wthorp:
Fixed test that fails on windows due to different OS error text for file/directory not found


storagenode/piecestore/endpoint_test.go
```diff
@@ -265,16 +265,16 @@ func TestDownload(t *testing.T) {
...
 			require.NoError(t, err)
 			require.Equal(t, expectedData, buffer[:n])
 		}
 
 		err = downloader.Close()
-		if tt.err != "" {
+		if len(tt.errs) > 0 {
 			require.Error(t, err)
-			require.Contains(t, err.Error(), tt.err)
+			require.True(t, strings.Contains(err.Error(), tt.errs[0]) || strings.Contains(err.Error(), tt.errs[1]), err.Error())
```
mniewrzal: maybe `require.Contains(t, strings.Join(tt.errs, ""), err.Error())`?

wthorp: I was assuming that err.Error() has some other text, in which case that wouldn't work.

PR #1675, opened 2019-04-04
add test for tally calculateAtRestData

JessicaGreben:
This PR adds a few tests for tally/rollup service and other in need areas.

### DIFF
--- a/pkg/accounting/tally/tally.go
+++ b/pkg/accounting/tally/tally.go
@@ -63,11 +63,11 @@ func (t *Service) Run(ctx context.Context) (err error) {
 	}
 }
 
-// Tally calculates data-at-rest once
+//Tally calculates data-at-rest and bandwidth usage once
 func (t *Service) Tally(ctx context.Context) error {
+	// data at rest
 	var errAtRest, errBucketInfo error
-	latestTally, nodeData, bucketData, err := t.calculateAtRestData(ctx)
-
+	latestTally, nodeData, bucketData, err := t.CalculateAtRestData(ctx)
 	if err != nil {
 		errAtRest = errs.New("Query for data-at-rest failed : %v", err)
 	} else {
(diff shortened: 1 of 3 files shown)


pkg/accounting/tally/tally.go
```diff
@@ -63,11 +63,11 @@ func (t *Service) Run(ctx context.Context) (err error) {
 	}
 }
 
-// Tally calculates data-at-rest once
+// Tally calculates data-at-rest and bandwidth usage once
```
jenlij: You can remove the bandwidth part since we're doing that outside of the tally service

pkg/accounting/tally/tally.go
```diff
@@ -63,11 +63,11 @@ func (t *Service) Run(ctx context.Context) (err error) {
...
 func (t *Service) Tally(ctx context.Context) error {
+	// data at rest
```
jenlij: you may remove this comment too

satellite/satellitedb/accounting.go
```diff
@@ -225,12 +227,23 @@ func (db *accountingDB) SaveBucketTallies(ctx context.Context, intervalStart tim
 		iSegments := dbx.BucketStorageTally_InlineSegmentsCount(uint(info.InlineSegments))
 		objectCount := dbx.BucketStorageTally_ObjectCount(uint(info.Files))
 		meta := dbx.BucketStorageTally_MetadataSize(uint64(info.MetadataSize))
-		_, err := db.db.Create_BucketStorageTally(ctx, bucketName, projectID, interval, inlineBytes, remoteBytes, rSegments, iSegments, objectCount, meta)
+		dbxTally, err := db.db.Create_BucketStorageTally(ctx, bucketName, projectID, interval, inlineBytes, remoteBytes, rSegments, iSegments, objectCount, meta)
 		if err != nil {
-			return err
+			return nil, err
 		}
+		tally := accounting.BucketTally{
```
jenlij: Rather than posting to the DB and creating a slice of BucketTallies in one method, do you think we should split them into 2 methods, one for each? That would avoid returning the slice and then dropping it.

JessicaGreben: yeah that sounds like a good idea.

PR #1676, opened 2019-04-05
Build with the release script

brimstone:


Makefile
```diff
@@ -1,4 +1,4 @@
-GO_VERSION ?= 1.11
+GO_VERSION ?= 1.12.1
```
egonelbre: We should be able to use `1.12.3`, they made a new release yesterday evening.

egonelbre: Disregard this, it seems Travis doesn't still work 1.12.3, might need something more complicated.

PR #1680, opened 2019-04-05
Selecting non-reputable nodes should not result in excessive selection

kaloyan-raev:
Fixes https://storjlabs.atlassian.net/browse/V3-1487

Thanks for submitting a PR!


pkg/overlay/cache.go
```diff
@@ -225,32 +225,32 @@ func (cache *Cache) FindStorageNodesWithPreferences(ctx context.Context, req Fin
 		reputableNodeCount = req.RequestedCount
 	}
 
-	auditCount := preferences.AuditCount
-	if auditCount < preferences.NewNodeAuditThreshold {
-		auditCount = preferences.NewNodeAuditThreshold
-	}
-
-	reputableNodes, err := cache.db.SelectStorageNodes(ctx, reputableNodeCount, &NodeCriteria{
+	newNodeCount := int(float64(reputableNodeCount) * preferences.NewNodePercentage)
```
mniewrzal: can you check if `preferences.NewNodePercentage > 0` otherwise set `newNodeCount = 0`?

kaloyan-raev: Makes sense. Done.

littleskunk:
As far as I remember this PR didn't fix the issue and we used an workaround for release-alpha8.

The last error was:
```
root@kali:~/storj# head -c 1M </dev/urandom | uplink put s3://littleskunk/1M
Fatal error: segment error: ecclient error: duplicated nodes are not allowed; segment error: ecclient error: duplicated nodes are not allowed
```

pkg/overlay/cache.go
```diff
@@ -245,25 +265,11 @@ func (cache *Cache) FindStorageNodesWithPreferences(ctx context.Context, req Fin
 		return nil, err
```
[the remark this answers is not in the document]
kaloyan-raev: Done with https://github.com/storj/storj/pull/1680/commits/2e56e7393860e1b428896196c0ac0f5229c07ab7

pkg/overlay/cache.go
```diff
@@ -243,35 +270,19 @@ func (cache *Cache) FindStorageNodesWithPreferences(ctx context.Context, req Fin
...
-		FreeDisk:      req.FreeDisk,
-
-		AuditThreshold: preferences.NewNodeAuditThreshold,
-
-		Excluded: req.ExcludedNodes,
+		Excluded: excluded,
 
 		MinimumVersion: preferences.MinimumVersion,
 	})
 	if err != nil {
```
egonelbre: I'm wondering whether we should fall back to selecting any node when there aren't enough nodes at this point?

kaloyan-raev: No. The callers of this method already expect to get the exact number of the requested number of nodes. No less, no more.

egonelbre: I was thinking of when there aren't enough storage nodes that are reputable.

kaloyan-raev: Do you mean to complete the number with new vetted nodes?

I am not sure if this makes sense. We should manage our vetting process in a way that we have at least 90-95 reputable nodes in the network.

It is worth discussing this, but I think is out of scope for this PR.

egonelbre: Sure let's bring the discussion to design-drafts.

PR #1685, opened 2019-04-05
add ability to inspect version to all binaries

jtolds:
Currently, it is hard to tell if a release binary we built was built correctly. This commit makes it so running binaries with the `version` subcommand will output build version information, and the debug port now has a new /version handler that returns version information. Last, this updates the release script to allow builds without a version tag, but will not allow the release flag to be set in that case.


internal/version/version.go
```diff
@@ -35,6 +36,9 @@ type Info struct {
 	CommitHash string    `json:"commitHash,omitempty"`
 	Version    SemVer    `json:"version"`
 	Release    bool      `json:"release,omitempty"`
+
+	crcOnce       sync.Once
+	commitHashCRC uint32
```
egonelbre: Any reason we are not making it part of build flags?

internal/version/stats.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package version
+
+import "hash/crc32"
+
+func (v *Info) Stats(cb func(name string, val float64)) {
```
egonelbre: Use `reportValue` or similar. When reading the code `cb` confused me.

jtolds: callback? but okay yeah

internal/version/version.go
jtolds: uhh, yeah i mean, from a precomputation perspective that will save every process some cycles, but from a correctness perspective it adds an avenue for bugs. this way (don't let it be configured) means we're much more unlikely to get it wrong, and i think correctness here is more important than the perf gain we get from basically denormalizing and repeating the data

internal/version/version.go
```diff
@@ -153,4 +157,5 @@ func init() {
 	if Build.Timestamp.Unix() == 0 || Build.CommitHash == "" {
 		Build.Release = false
 	}
+
```
stefanbenten: Besides this empty line, it looks good 👍

PR #1686, opened 2019-04-05
Build multiarch ARM32v6 and AMD64 images

brimstone:


brimstone:
What should be done about the TRACKED_BRANCH images?

Makefile
```diff
@@ -119,20 +119,36 @@ images: satellite-image storagenode-image uplink-image gateway-image ## Build ga
...
 .PHONY: satellite-ui-image
 satellite-ui-image: ## Build satellite-ui Docker image
 	${DOCKER_BUILD} --pull=true -t storjlabs/satellite-ui:${TAG}${CUSTOMTAG} -f web/satellite/Dockerfile .
 .PHONY: storagenode-image
-storagenode-image: ## Build storagenode Docker image
-	${DOCKER_BUILD} --pull=true -t storjlabs/storagenode:${TAG}${CUSTOMTAG} -f cmd/storagenode/Dockerfile .
+storagenode-image: storagenode_linux_arm storagenode_linux_amd64 ## Build storagenode Docker image
+	${DOCKER_BUILD} --pull=true -t storjlabs/storagenode:${TAG}${CUSTOMTAG}-amd64 \
+		-f cmd/storagenode/Dockerfile .
+	${DOCKER_BUILD} --pull=true -t storjlabs/storagenode:${TAG}${CUSTOMTAG}-arm32v6 \
```
coyle: Is there a way to specify the tag here as an env var?

brimstone:
`TAG` is already allowed to be overriden with an environment variable with the same name. Is that what you're expecting?

PR #1699, opened 2019-04-08
Don't cancel fetching peer identity context when request is cancelled

egonelbre:
Currently when uplink is cutting storage nodes it could unintentionally cancel fetching peer identity from the satellite, which ends up being cached.


storagenode/trust/service.go
```diff
@@ -91,7 +91,11 @@ func (pool *Pool) VerifyUplinkID(ctx context.Context, id storj.NodeID) error {
 }
 
 // GetSignee gets the corresponding signee for verifying signatures.
+// It ignores passed in ctx cancellation to avoid miscaching between concurrent requests.
 func (pool *Pool) GetSignee(ctx context.Context, id storj.NodeID) (signing.Signee, error) {
+	// creating a new context here to avoid request context canceling fetching peer identity
+	nestedContext := context.Background()
```
coyle: seems kind of hacky. Is there a better way than creating a new context?

egonelbre: I suspect a nicer fix would be to not cache (or cache only for a limited time), when the request fails. But, that requires more changes, so this will do for now.

PR #1707, opened 2019-04-08
Uplink CLI setup wizard

mniewrzal:
What: uplink cli wizard to define: satellite address, api key, encryption passphrase

Why: setup process should be more user-friendly and there should be a way to not define encryption passphrase with command line parameter

Thanks for submitting a PR!


cmd/uplink/cmd/setup.go
```diff
@@ -55,5 +60,60 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	return process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), nil)
+	isInteractive = true
+	var override map[string]interface{}
+	if isInteractive {
```
coyle: will this ever be false ?

mniewrzal: sorry, its still wip

cmd/uplink/cmd/setup.go
```diff
@@ -55,5 +59,53 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	return process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), nil)
+	var override map[string]interface{}
+	if setupCfg.Interactive {
+		fmt.Print("Enter your Satellite address: ")
```
kaloyan-raev: I think we should check if this config has been provided as a flag by the user, and not prompt for it in such case.

This way if the user passes all required config as flags, they won't need to pass `--interactive=false`.

cmd/uplink/cmd/root.go
```diff
@@ -19,7 +19,8 @@ import (
 
 // UplinkFlags configuration flags
 type UplinkFlags struct {
-	Identity identity.Config
+	Interactive bool `help:"enable interavtive more" default:"true" setup:"true"`
```
kaloyan-raev: Typo: "interavtive"
Typo: "more" --> mode?

PR #1713, opened 2019-04-08
care if error is sql.ErrNoRows

wthorp:
What: verify that error we've assumed is not some other err

Why: for kicks


satellite/satellitedb/overlaycache.go
```diff
@@ -211,7 +211,9 @@ func (cache *overlaycache) Update(ctx context.Context, info *pb.Node) (err error
 		address = &pb.NodeAddress{}
 	}
 
-	if err != nil {
+	if err != nil && err != sql.ErrNoRows {
+		return Error.Wrap(errs.Combine(err, tx.Rollback()))
+	} else if err != nil && err == sql.ErrNoRows {
```
egonelbre: This feels complicated to read, but I don't have significant improvement to suggest either :/

egonelbre: Also is there a test we could write for it.

wthorp: I don't actually want this code.  I just don't like DBX, and wanted to point our the types of errors it introduces.

PR #1717, opened 2019-04-09
refactor injuredsegments db so that we can't have duplicates

wthorp:
What: refactor injuredsegments db so that we can't have duplicates

A queue can't do this, so we're getting rid of the queue too.

Why: because we get errors when trying to repair something that was already repaired


pkg/datarepair/repairer/repairer.go
```diff
@@ -98,6 +98,10 @@ func (service *Service) process(ctx context.Context) error {
 		if err != nil {
 			zap.L().Error("Repair failed", zap.Error(err))
 		}
+		err = service.queue.Delete(ctx, seg)
```
ifraixedes: Should we delete the segment from the queue if the repair process fail?

satellite/satellitedb/repairqueue.go
```diff
@@ -24,12 +30,57 @@ func (r *repairQueue) Insert(ctx context.Context, seg *pb.InjuredSegment) error
...
-func (r *repairQueue) Select(ctx context.Context) (seg *pb.InjuredSegment, err error) {
+func (r *repairQueue) postgresSelect(ctx context.Context) (seg *pb.InjuredSegment, err error) {
 	//todo :  add or age > some time (1 hour?)
-	err = r.db.QueryRowContext(ctx, `SELECT data FROM injuredsegments ORDER BY id LIMIT 1 WHERE attempted IS NOT NULL`).Scan(&seg)
+	err = r.db.QueryRowContext(ctx, r.db.Rebind(`
+	UPDATE injuredsegments SET attempted = ? WHERE path = (
+		SELECT path FROM injuredsegments
+		WHERE attempted IS NULL
+		ORDER BY path FOR UPDATE SKIP LOCKED LIMIT 1
+	) RETURNING data`), time.Now().UTC()).Scan(&seg)
```
ifraixedes: If I'm not wrong, this shouldn't do what's expected, because a pointer to `*pb.InjuredSegment` and it should be a non-nil pointer to `pb.InjuredSegment`

satellite/satellitedb/repairqueue.go
```diff
@@ -21,39 +22,36 @@ type repairQueue struct {
...
+	return
 }
 
-func (r *repairQueue) sqliteDequeue(ctx context.Context) (seg pb.InjuredSegment, err error) {
+func (r *repairQueue) sqliteSelect(ctx context.Context) (seg *pb.InjuredSegment, err error) {
 	err = r.db.WithTx(ctx, func(ctx context.Context, tx *dbx.Tx) error {
-		var id int64
-		err = tx.Tx.QueryRowContext(ctx, `SELECT id, info FROM injuredsegments ORDER BY id LIMIT 1`).Scan(&id, &seg)
+		var path string
+		err = tx.Tx.QueryRowContext(ctx, r.db.Rebind(`SELECT path, data FROM injuredsegments WHERE attempted IS NULL ORDER BY path LIMIT 1`)).Scan(&path, &seg)
```
ifraixedes: is `Regind` call needed for this query?

satellite/satellitedb/repairqueue.go
```diff
@@ -21,39 +22,36 @@ type repairQueue struct {
...
 		if err != nil {
 			return err
 		}
-		res, err := tx.Tx.ExecContext(ctx, r.db.Rebind(`DELETE FROM injuredsegments WHERE id = ?`), id)
+		res, err := tx.Tx.ExecContext(ctx, r.db.Rebind(`UPDATE injuredsegments SET attempted = ? WHERE path = ?`), time.Now().UTC(), path)
```
ifraixedes: is `Regind` call needed for this query?

satellite/satellitedb/repairqueue.go
```diff
@@ -21,39 +22,36 @@ type repairQueue struct {
...
 
-func (r *repairQueue) sqliteDequeue(ctx context.Context) (seg pb.InjuredSegment, err error) {
+func (r *repairQueue) sqliteSelect(ctx context.Context) (seg *pb.InjuredSegment, err error) {
 	err = r.db.WithTx(ctx, func(ctx context.Context, tx *dbx.Tx) error {
-		var id int64
-		err = tx.Tx.QueryRowContext(ctx, `SELECT id, info FROM injuredsegments ORDER BY id LIMIT 1`).Scan(&id, &seg)
+		var path string
+		err = tx.Tx.QueryRowContext(ctx, r.db.Rebind(`SELECT path, data FROM injuredsegments WHERE attempted IS NULL ORDER BY path LIMIT 1`)).Scan(&path, &seg)
 		if err != nil {
 			return err
```
ifraixedes: Shouldn't we treat `sql.ErrRows` different here?

satellite/satellitedb/repairqueue.go
```diff
@@ -72,33 +70,35 @@ func (r *repairQueue) sqliteDequeue(ctx context.Context) (seg pb.InjuredSegment,
...
+		var seg pb.InjuredSegment
+		err = rows.Scan(&seg)
+		if err != nil {
+			return
 		}
-		segments = append(segments, *seg)
+		segs = append(segs, seg)
 	}
-	return segments, nil
+	return
```
ifraixedes: We should inspect [`rows.Err`](https://golang.org/pkg/database/sql/#Rows.Err) before returning

pkg/datarepair/queue/queue.go
```diff
@@ -34,6 +36,32 @@ func NewQueue(client storage.Queue) *Queue {
 	return &Queue{db: client}
 }
 
+// // NewQueue returns a pointer to a new Queue instance with an initialized connection to Redis
```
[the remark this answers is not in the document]
navillasa: Yep, commented code shouldn't end up committed to the main repo! Sometimes people comment out code in a PR when it's a WIP. We'll be sure to remove this before final review.

pkg/datarepair/queue/queue_test.go
```diff
@@ -68,23 +71,25 @@ func TestSequential(t *testing.T) {
...
 				LostPieces: []int32{int32(i)},
 			}
-			err := q.Enqueue(ctx, seg)
+			err := q.Insert(ctx, seg)
 			assert.NoError(t, err)
 			addSegs = append(addSegs, seg)
 		}
 
-		list, err := q.Peekqueue(ctx, 100)
+		list, err := q.SelectN(ctx, 100)
```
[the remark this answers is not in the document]
navillasa: that's a good idea

pkg/datarepair/queue/queue_test.go
```diff
@@ -48,9 +45,9 @@ func TestDequeueEmptyQueue(t *testing.T) {
 
 		q := db.RepairQueue()
 
-		s, err := q.Dequeue(ctx)
-		assert.Error(t, err)
-		assert.Equal(t, pb.InjuredSegment{}, s)
+		s, err := q.Select(ctx)
+		require.Error(t, err)
+		require.Nil(t, s)
```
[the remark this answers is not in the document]
navillasa: I agree that the require.Nil check isn't necessary

pkg/datarepair/repairer/repairer.go
navillasa: probably not- good catch. instead I think we need to update the `attempted` timestamp.

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
navillasa: or actually, I think it's fine in this test case where otherwise we would use `_, err := q.Select(ctx)`. additionally I think we want to confirm that `Select` won't error but will return nil if there are no segments available.

pkg/datarepair/queue/queue_test.go
```diff
@@ -120,11 +119,14 @@ func TestParallel(t *testing.T) {
 		for i := 0; i < N; i++ {
 			go func(i int) {
 				defer wg.Done()
-				segment, err := q.Dequeue(ctx)
+				s, err := q.Select(ctx)
+				require.NoError(t, err)
+				err = q.Delete(ctx, s)
+				require.NoError(t, err)
```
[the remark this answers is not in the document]
navillasa: I tried doing this but I think messed up the second sentence^ and broke the test 😂if you want to go ahead and make the change directly on this branch feel free though, this is definitely a team affair!

pkg/datarepair/queue/queue_test.go
```diff
@@ -9,19 +9,15 @@ import (
...
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/satellite"
 	"storj.io/storj/satellite/satellitedb/satellitedbtest"
-	"storj.io/storj/storage/redis"
-	"storj.io/storj/storage/redis/redisserver"
-	"storj.io/storj/storage/testqueue"
 )
 
-func TestEnqueueDequeue(t *testing.T) {
+func TestInsertDequeue(t *testing.T) {
```
mobyvb: `TestInsertSelect`

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
ifraixedes: yes, it should be `_, err :=` if we remove the `require.Nil`.

>  additionally I think we want to confirm that Select won't error but will return nil if there are no segments available.

In case that there isn't such test case, it would be worth to add it.

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
ifraixedes: I'll do it

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
ifraixedes: > additionally I think we want to confirm that Select won't error but will return nil if there are no segments available.

Checking the whole test cases, it looks if we follow the Go common conventions, the result can be ignore in case of an error as mentioned before, but we should check that the error is of the class `ErrEmptyQueue`

I did such change, see my commit https://github.com/storj/storj/pull/1717/commits/063dec403d0a74879d7d745e941b68e5f8a79b38 if it isn't OK, let me know and I will drop it.

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
ifraixedes: Done in https://github.com/storj/storj/pull/1717/commits/0b7b4209643268a32f949c5b43c08b5c96029cc5

It's a bit different on what I commented, but after going through the test cases and wanted to make the changes, I think that's better in this way. The commit messages contains a brief explanation and the reasons, so please read it too a part of reviewing the changes.

pkg/datarepair/checker/checker.go
```diff
@@ -133,7 +133,7 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 				numHealthy := len(nodeIDs) - len(missingPieces)
 				if (int32(numHealthy) >= pointer.Remote.Redundancy.MinReq) && (int32(numHealthy) <= pointer.Remote.Redundancy.RepairThreshold) {
 					remoteSegmentsNeedingRepair++
-					err = checker.repairQueue.Enqueue(ctx, &pb.InjuredSegment{
+					err = checker.repairQueue.Insert(ctx, &pb.InjuredSegment{
```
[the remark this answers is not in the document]
wthorp: repairQueue.Insert() should return nil (no error) if the unique constraint fails.  we could a test though.

pkg/datarepair/repairer/repairer.go
wthorp: If we don't delete it, we probably need another field in the database to order the results of repairQueue.Select() by, to move it to the back of the queue.  What we don't want is the repair queue getting 'stuck' trying to repair something it can't.  I'm not quite sure where the 'irreparable' logic is, or why we don't see it here.

satellite/satellitedb/repairqueue.go
wthorp: no.  in truth its not needed for any of our queries (both postgres and sqlite accept the $1 syntax for variables).

satellite/satellitedb/repairqueue.go
wthorp: I believe so.  I don't think the '?' syntax is cross-platform.

pkg/datarepair/repairer/repairer.go
navillasa: I thought that field would be the `attempted` timestamp right? but I'm guessing the logic for moving it back in the queue hasn't been added yet (I haven't checked)

satellite/satellitedb/repairqueue.go
navillasa: hmm I didn't totally understand this comment-- I know that this function should return a `*pb.InjuredSegment`, but you're saying that `pb.InjuredSegment` should be returned instead?

satellite/satellitedb/repairqueue.go
wthorp: do you mean sql.ErrNoRows?

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
wthorp: its a test.  its not important, but its okay to test that we're complying with conventions.  we probably should test that we're complying with conventions. :D

satellite/satellitedb/repairqueue.go
wthorp: Yeah, its not immediately clear that would work, given that seg starts out as a nil pointer.  However, I believe it does work.  We use a Scanner for InjuredSegment which calls proto.Unmarshal() which calls injuredSegment,Reset(), which makes seg non-nil.

https://github.com/storj/storj/blob/904d009a732964c795518eaf1fbea1f2c580614f/pkg/pb/scannerValuer.go#L23

https://github.com/storj/storj/blob/904d009a732964c795518eaf1fbea1f2c580614f/pkg/pb/datarepair.pb.go#L32

satellite/satellitedb/repairqueue.go
wthorp: because that line is part of anonymous function that get run via `r.db.WithTx(ctx, func(ctx context.Context, tx *dbx.Tx) error) `, it is sufficient to return any error and handle ErrNoRows in the main function.

pkg/datarepair/queue/queue_test.go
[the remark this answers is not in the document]
ifraixedes: Not very important, mostly I wanted to understand if we apply those conventions because if we always have to return `nil` when an error is returned, then we have to add those asserts on any test which check that.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -172,23 +172,13 @@ read limitoffset (
...
-	field id   serial64
-	field info blob
-)
+	key path
 
-create injuredsegment ( )
-
-read first (
-	select injuredsegment
+	field path text
```
egonelbre: `path` is not `text`, it's a binary blob... `text` usually implies utf-8. however, it probably won't cause problems in the database

ifraixedes: yes, but it looks more precise, hence should we swap it?

PR #1730, opened 2019-04-10
add custom tparse that doesn't stall

egonelbre:
What: use travis that prints a dot every 5s.
Why: currently travis is stalling because tparse is not printing anything for a long time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,12 @@ matrix:
       install:
         - pushd ~
         - GOBIN=${GOPATH}/bin GOPATH=~/gotools go get github.com/mattn/goveralls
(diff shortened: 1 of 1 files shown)


jtolds:
also frustrating is that tparse obscures which test travis is working on. what are we gaining from tparse?

egonelbre:
@jtolds we have 1000s of tests and when something fails, then the failure will be in the middle of them. Making it really hard to find a specific test that failed.

However, now when the tests take too long it should print the output, rather than just printing "things took too long". It still has the problem of not showing live output... and I'm not sure how to achieve both "have failures at the top" and "show live output" in Travis.

PR #1732, opened 2019-04-10
Inspector draw routing table

Fadila82:
What: 
Add capability to inspector to ask a local node to draw its routing table graph.
When the command "inspector kad routing-graph <node-id>" is called, it saves the routing table graph as a dot file named routing-graph-<node-id>.dot

https://storjlabs.atlassian.net/browse/V3-1078

Thanks for submitting a PR!


cmd/inspector/main.go
```diff
@@ -103,6 +107,12 @@ var (
 		Short: "dump all nodes in the routing table",
 		RunE:  DumpNodes,
 	}
+	drawTableCmd = &cobra.Command{
+		Use: "routing-graph <node_id>",
+		Short: "Save the routing table graph as a dot file with name routing-graph-<node-id>",
+		Args:  cobra.MinimumNArgs(1),
```
kaloyan-raev: Can we have the `<node_id> ` argument optional? When not specified it should draw the routing-graph of the current node.

pkg/kademlia/routing.go
```diff
@@ -321,3 +323,73 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+	buf.Write([]byte("}\n"))
+
+}
+
+func (rt *RoutingTable) addLeafBucketToGraph(b bucketID, buf *bytes.Buffer, prefix string) {
+	fmt.Fprintf(buf, "b%s [label=<<b><font point-size=\"18\">%s</font></b><br />\n<i>routing:</i><br align=\"left\"/>", prefix, prefix)
+		
+	nodes, _ := rt.getNodeIDsWithinKBucket(b)
+	for _, n := range nodes {
+		fmt.Fprintf(buf, "  %s<br align=\"left\" />", hex.EncodeToString(n[:]))
```
kaloyan-raev: Instead of `hex.EncodeToString(n[:])` try calling `n.String()`, which will display the node ID in the expected way.

kaloyan-raev: Let's also display the node address together with the node id. I think this will make easier to analyze the routing table because the node ids themselves are hard to comprehend.

The node address can be added in paranthesis after the node id, i.e. `<node_id> (<node_address>)`.

cmd/inspector/main.go
```diff
@@ -244,6 +254,59 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
...
+	info, err := i.kadclient.DrawTable(context.Background(), &pb.DrawTableRequest{
+		Id:      n.GetNode().Id,
+		Address: n.GetNode().GetAddress(),
+	})
+	if err != nil {
+		return ErrRequest.Wrap(err)
+	}
+	fh, err := dotOutput(args[0])
+	if err != nil {
+		panic(err)
```
kaloyan-raev: We avoid panicking. Instead, return the error.

kaloyan-raev: We avoid panicking. Instead, return the error.

cmd/inspector/main.go
```diff
@@ -244,6 +254,59 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
...
+	}
+	defer func() {
+		err := fh.Close()
+		if err != nil {
+			panic(err)
+		}
```
kaloyan-raev: Again, instead of panicking, we usually combine `err` (from the return value declaration) with the error returned from `fh.Close()`, i.e. `defer func() { err = errs.Combine(err, fh.Close()) }()`

`errs.Combine` comes frome the `github.com/zeebo/errs` package.

pkg/kademlia/routing.go
```diff
@@ -321,3 +323,73 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+		if b&bitMask > 0 {
+			b1 = append(b1, bID)
+		} else {
+			b0 = append(b0, bID)
+		}
+	}
+	return b0, b1
+}
+
+// Prints the routing table graph as a dot graph in specified buffer
```
kaloyan-raev: The doc comment must start with the word `BufferedGraph`. Otherwise, the linter will keep complaining.

pkg/pb/overlay.proto
```diff
@@ -13,6 +13,7 @@ service Nodes {
     rpc Query(QueryRequest) returns (QueryResponse);
     rpc Ping(PingRequest) returns (PingResponse);
     rpc RequestInfo(InfoRequest) returns (InfoResponse);
+    rpc RequestGraph(GraphRequest) returns (GraphResponse);
```
egonelbre: Adding a public rpc call for this concerns me. I would like to avoid adding things here that are not essential parts of the API. Only having it internal only should suffice.

jtolds: oh, yeah, the original ticket said that we should be able to ask any node, but it looks like i changed the ticket on april 3 to be specifically only the inspector and only the internal port. so yeah, i think this should be the set of inspector-only methods? i don't know where those are defined though

egonelbre: The PR already added it to the inspector.

pkg/kademlia/kademlia.go
```diff
@@ -245,6 +246,28 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (*pb.InfoRespons
 	return info, nil
 }
 
+// FetchGraph connects to a node address and returns its routing table as a graph
+func (k *Kademlia) FetchGraph(ctx context.Context, node pb.Node) (*pb.GraphResponse, error) {
```
egonelbre: Yeah, I think since we don't want to make it public, we don't have to add it as a public thing to kademlia. This means we'll avoid accumulating diagnostic methods on kademlia.

pkg/kademlia/routing.go
```diff
@@ -321,3 +322,77 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
 		zap.L().Debug("connection success error:", zap.Error(err))
 	}
 }
+
+func splitBucket(bIDs []bucketID, bitDepth int) ([]bucketID, []bucketID) {
```
egonelbre: Let's move these methods to a separate package, maybe `kademlia/routinggraph`? To avoid making the kademlia package larger.

pkg/pb/overlay.proto
jtolds: oh, great!

Fadila82: I'm not sure I understand well what you are saying. Does it mean we shouldn't be able to call "inspector kad routing-graph <node-id>?

kaloyan-raev:
If anyone is curious, here is a real-life graph generated from the routing table of my Pi storage node.

[routing-graph.dot.gz](https://github.com/storj/storj/files/3083648/routing-graph.dot.gz)

[routing-graph.pdf](https://github.com/storj/storj/files/3083640/routing-graph.pdf)

pkg/pb/overlay.proto
kaloyan-raev: @Fadila82 As far as I get it, the inspector should not be able to ask for the routing graph of any other node except the current node, i.e. the one we are connecting the inspector to.

So we should be able to call only `inspector kad routing-graph` (without `node-id` argument), which will return the routing graph of the current node.

Thus, you should not need the `Nodes.RequestGraph` rpc method. The `KadInspector.DrawTable` should be enough to achieve this.

cmd/inspector/main.go
```diff
@@ -103,6 +107,11 @@ var (
 		Short: "dump all nodes in the routing table",
 		RunE:  DumpNodes,
 	}
+	drawTableCmd = &cobra.Command{
+		Use:   "routing-graph (<node_id>)",
```
egonelbre: The option to specify the target node id was removed

cmd/inspector/main.go
```diff
@@ -103,6 +107,11 @@ var (
...
+		Short: "Create a dot graph of the routing table saved by default as routing-graph.dot",
```
egonelbre: Specify the defaults when defining the flag, the name will automatically pop up.

cmd/inspector/main.go
```diff
@@ -39,6 +40,9 @@ var (
 	// CSVPath is the csv path where command output is written
 	CSVPath string
 
+	//DotPath is the output of command routing-graph
+	DotPath string
```
egonelbre: Rather than saving into a file, output to stdout. It's nice to use and pipe to dot e.g. `inspector kad routing-graph | dot -Tsvg > routing-graph.svg`. Less flags is clearer.

cmd/inspector/main.go
```diff
@@ -244,6 +253,47 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
 	return nil
 }
 
+func dotOutput() (*os.File, error) {
+	if DotPath == "stdout" {
```
egonelbre: usually stdout is specified as `"-"` rather than `stdout` explicitly

cmd/inspector/main.go
```diff
@@ -244,6 +253,47 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
 	return nil
 }
 
+func dotOutput() (*os.File, error) {
```
egonelbre: When creating a func that creates something use `New`, `Create` etc. prefix to make clear that it needs to be closed.

cmd/inspector/main.go
```diff
@@ -244,6 +253,47 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
...
+}
+
+// DrawTableAsGraph outputs the table routing as a graph
+func DrawTableAsGraph(cmd *cobra.Command, args []string) (err error) {
+	i, err := NewInspector(*Addr, *IdentityPath)
+	if err != nil {
+		return ErrInspectorDial.Wrap(err)
+	}
+
+	info, err := i.kadclient.DrawTable(context.Background(), &pb.DrawTableRequest{})
```
egonelbre: I'm wondering whether it makes more sense to return the graph rather than the dot file. That way the server doesn't need to import routing graph printing.

cmd/inspector/main.go
egonelbre: Hmm, I suspect `CSVPath` would've been nicer that way as well.

pkg/kademlia/kademlia.go
```diff
@@ -245,6 +246,14 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (*pb.InfoRespons
 	return info, nil
 }
 
+// FetchLocalGraph returns the routing table as a dot graph
+func (k *Kademlia) FetchLocalGraph() []byte {
```
egonelbre: Kademlia having a method to draw it as a graph feels wrong to me. There are many ways to visualize kademlia and attaching all of those to `Kademlia` itself ends up collecting different concerns, however it helps to have a general concern of "get me the full routing table" and the client decides how to visualize it.

pkg/kademlia/routing.go
```diff
@@ -321,3 +322,77 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+	}
+
+	rt.addBucketsToGraph(bucketids, 0, buf, "")
+
+	buf.Write([]byte("}\n"))
+
+}
+
+func (rt *RoutingTable) addLeafBucketToGraph(b bucketID, buf *bytes.Buffer, prefix string) {
+	fmt.Fprintf(buf, "b%s [label=<<b><font point-size=\"18\">%s </font></b><br />\n<i>routing:</i><br align=\"left\"/>", prefix, prefix)
```
egonelbre: Use dot tables or [records](https://www.graphviz.org/doc/info/shapes.html#record) since they align nicer in the final output. And prefer monospaced fonts for the same reason.

pkg/kademlia/routing.go
```diff
@@ -321,3 +322,77 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+		} else {
+			b0 = append(b0, bID)
+		}
+	}
+	return b0, b1
+}
+
+// BufferedGraph prints the routing table graph as a dot graph in the specified buffer
+func (rt *RoutingTable) BufferedGraph(buf *bytes.Buffer) {
+	buf.Write([]byte("digraph{\nnode [shape=box];edge [dir=none];\n"))
```
egonelbre: maybe, ranking left-to-right, might make the output look nicer

pkg/kademlia/routing.go
```diff
@@ -321,3 +322,77 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+
+	ids, _ := rt.GetBucketIds()
+	var bucketids []bucketID
+	for _, n := range ids {
+		bucketids = append(bucketids, keyToBucketID(n))
+	}
+
+	rt.addBucketsToGraph(bucketids, 0, buf, "")
+
+	buf.Write([]byte("}\n"))
```
egonelbre: Add this above as `defer buf.Write([]byte("}\n"))`, it helps to avoid accidentally forgetting the closing brace.

pkg/kademlia/routing.go
```diff
@@ -321,3 +322,77 @@ func (rt *RoutingTable) ConnSuccess(ctx context.Context, node *pb.Node) {
...
+	nodes, _ := rt.getUnmarshaledNodesFromBucket(b)
+	for _, n := range nodes {
+		printNodeInBuffer(n, buf)
+	}
+	fmt.Fprintf(buf, "<i>cache:</i><br align=\"left\" />")
+	cachedNodes, _ := rt.replacementCache[b]
+	for _, c := range cachedNodes {
+		printNodeInBuffer(c, buf)
+	}
+	fmt.Fprintf(buf, ">];")
```
egonelbre: defer this above, also add a new line... easier to read the output

egonelbre:
A lot of nitpicking, but a really nice to have feature :)

pkg/kademlia/kademlia.go
```diff
@@ -312,6 +312,17 @@ func (k *Kademlia) Seen() []*pb.Node {
 	return nodes
 }
 
+// GetNodesWithinKBucket returns all the routing nodes in the specified k-bucket
+func (k *Kademlia) GetNodesWithinKBucket(bID bucketID) ([]*pb.Node, error) {
+	return k.routingTable.getUnmarshaledNodesFromBucket(bID)
+	//return nodes, err
```
kaloyan-raev: Should you remove this commented line?

pkg/pb/overlay.proto
```diff
@@ -40,6 +40,11 @@ message InfoResponse {
     node.NodeVersion version = 5;
 }
 
+message GraphRequest {};
```
egonelbre: This can be removed as well.

cmd/inspector/main.go
```diff
@@ -244,6 +250,28 @@ func NodeInfo(cmd *cobra.Command, args []string) (err error) {
...
+		return ErrInspectorDial.Wrap(err)
+	}
+	// retrieve buckets
+	info, err := i.kadclient.GetBucketList(context.Background(), &pb.GetBucketListRequest{})
+	if err != nil {
+		return ErrRequest.Wrap(err)
+	}
+
+	var buf bytes.Buffer
+	bufferedGraph(&buf, info)
```
egonelbre: `routinggraph.Draw(os.Stdout, info)`

pkg/kademlia/inspector.go
```diff
@@ -136,3 +136,29 @@ func (srv *Inspector) NodeInfo(ctx context.Context, req *pb.NodeInfoRequest) (*p
 		Version:  info.GetVersion(),
 	}, nil
 }
+
+// GetBucketList returns the list of buckets with their routing nodes and their cached nodes
+func (srv *Inspector) GetBucketList(ctx context.Context, req *pb.GetBucketListRequest) (*pb.GetBucketListResponse, error) {
+
```
egonelbre: nit: remove empty line

pkg/kademlia/inspector.go
```diff
@@ -136,3 +136,29 @@ func (srv *Inspector) NodeInfo(ctx context.Context, req *pb.NodeInfoRequest) (*p
...
+	bucketIds, err := srv.dht.GetBucketIds()
+	if err != nil {
+		return nil, err
+	}
+
+	buckets := make([]*pb.GetBucketListResponse_Bucket, len(bucketIds))
+
+	for i, b := range bucketIds {
+		bucketID := keyToBucketID(b)
+		routingNodes, _ := srv.dht.GetNodesWithinKBucket(bucketID)
```
egonelbre: Don't ignore errors. Alternatively, when you do, add a comment why it was essential.

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+
+func extendPrefix(prefix string, bit bool) string {
+	if bit {
+		return prefix + "1"
+	}
+	return prefix + "0"
+}
+
+// Draw writes the routing graph obtained using a GetBucketListResponse in the specified file
+func Draw(file *os.File, info *pb.GetBucketListResponse) (err error) {
```
egonelbre: use `w io.Writer` as the receiver, it doesn't have to be a file

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+	_, err = file.Write([]byte("digraph{\nnode [shape=plaintext, fontname=\"Courier\"];edge [dir=none];rankdir=LR;\n"))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_, errWrite := file.Write([]byte("}\n"))
+		err = errs.Combine(err, errWrite)
+		err = errs.Combine(err, file.Close())
```
egonelbre: Don't move the closing responsibility, since it can cause bugs. Keep the closing of something close to where it was opened/created.

Currently this would end up closing stdout.

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+	}()
+	var buf bytes.Buffer
```
egonelbre: No need for the buffer here, the caller can buffer it wants to.

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+	buckets := info.GetBuckets()
+	addBucketsToGraph(buckets, 0, &buf, "")
+	_, err = buf.WriteTo(os.Stdout)
```
egonelbre: This should write to the argument not `stdout`

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+
+	fmt.Fprintf(buf, "<TR><TD  COLSPAN=\"2\" ALIGN=\"LEFT\"><I><B>Routing:</B></I></TD></TR>")
+	routingNodes := b.GetRoutingNodes()
+	for _, n := range routingNodes {
+		printNodeInBuffer(n, buf)
+	}
+	fmt.Fprintf(buf, "<TR><TD  COLSPAN=\"2\"></TD></TR>")
+	fmt.Fprintf(buf, "<TR><TD  COLSPAN=\"2\" ALIGN=\"LEFT\"><I><B>Cache:</B></I></TD></TR>")
+	cachedNodes := b.GetCachedNodes()
+	for _, c := range cachedNodes {
```
egonelbre: `for _, node := range b.GetCachedNodes() {`

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+		if b&bitMask > 0 {
+			b1 = append(b1, bucket)
+		} else {
+			b0 = append(b0, bucket)
+		}
+	}
+	return b0, b1
+}
+
+func addBucketsToGraph(b []*pb.GetBucketListResponse_Bucket, depth int, buf *bytes.Buffer, inPrefix string) {
```
egonelbre: `func addBucketsToGraph(buf *bytes.Buffer, inPrefix string, depth int, buckets []*pb.GetBucketListResponse_Bucket) {`

Currently the order of arguments changes a lot between the funcs. Try to make them consistent.

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/pkg/pb"
+)
+
+func extendPrefix(prefix string, bit bool) string {
```
egonelbre: Move this as last func.

More important funcs first, utilties as last. It helps to find the important things in the package/file faster.

pkg/kademlia/routing.go
Fadila82: Here are two examples: default and with left-to-right ranking. I don't know which one is better.


default
left-to-right

What do you think?

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+		err = errs.Combine(err, file.Close())
+	}()
+	var buf bytes.Buffer
+	buckets := info.GetBuckets()
+	addBucketsToGraph(buckets, 0, &buf, "")
+	_, err = buf.WriteTo(os.Stdout)
+	return err
+}
+
+func splitBucket(buckets []*pb.GetBucketListResponse_Bucket, bitDepth int) ([]*pb.GetBucketListResponse_Bucket, []*pb.GetBucketListResponse_Bucket) {
```
egonelbre: You can use named return here. `(left, right []*pb.GetBucketListResponse_Bucket)`. Also move downwards below printing funcs.

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+func addBucketsToGraph(b []*pb.GetBucketListResponse_Bucket, depth int, buf *bytes.Buffer, inPrefix string) {
+	if len(b) == 1 {
+		addLeafBucketToGraph(b[0], buf, inPrefix)
+		return
+	}
+
+	b0, b1 := splitBucket(b, depth)
+
+	outPrefix := extendPrefix(inPrefix, false)
+	fmt.Fprintf(buf, "b%s [shape=point];", inPrefix)
```
egonelbre: Do not ignore errors. You can create something to collect the error to make writing this code easier.

As a rough example how to collect the error and structure printing:

```
type dot struct {
    out io.Writer
    err error
}

func (dot *dot) printf(format string, args ...interface{}) {
    if dot.err != nil {
        return
    }
    _, dot.err = fmt.Fprintf(dot.out, format, args...)
}

func (dot *dot) Leaf(prefix string, bucket *pb.GetBucketListResponse_Bucket) {
    dot.printf(`b%s [label=< <table cellborder="0"><tr><td cellspacing="0" sides="b" border="1" colspan="2"><b> %s </b></td></tr>`, prefix, prefix)
	defer dot.printf(`</table>>];`)

    for _, node := range bucket.GetRoutingNodes() {
        dot.Node(node)
    }

    ...
}

func (dot *dot) Node(node *pb.Node) {
    dot.printf(`<tr><td align="left">%s</td><td sides="r" align="left">%s</td></tr>`, n.Id, n.Address.Address)
}
```

pkg/kademlia/routinggraph/routinggraph.go
```diff
@@ -0,0 +1,100 @@
...
+	addBucketsToGraph(b0, depth+1, buf, outPrefix)
+	fmt.Fprintf(buf, "b%s -> b%s [label=<<b>0</b>>];", inPrefix, outPrefix)
+
+	outPrefix = extendPrefix(inPrefix, true)
+	addBucketsToGraph(b1, depth+1, buf, outPrefix)
+	fmt.Fprintf(buf, "b%s -> b%s [label=<<b>1</b>>];", inPrefix, outPrefix)
+}
+
+func printNodeInBuffer(n *pb.Node, buf *bytes.Buffer) {
+	fmt.Fprintf(buf, "<TR><TD ALIGN=\"LEFT\">%s</TD><TD SIDES=\"R\" ALIGN=\"LEFT\">%s</TD></TR>", n.Id.String(), n.Address.Address)
```
egonelbre: Go has backticks for raw literals so you don't have to escape `"` character. https://golang.org/ref/spec#String_literals

pkg/kademlia/routing.go
egonelbre: Both don't look that good :D. Maybe if the font-size was smaller for the node-ids and addresses both would look better.

kaloyan-raev: The first layout looks more intuitive to me. I am used to looking at binary trees from top to down, not from left to right.

PR #1735, opened 2019-04-10
Uplink CLI setup welcome message

mniewrzal:
What: Additional welcome message when `uplink setup` wizard is finished
image

Why: Better user experience during setup 

Thanks for submitting a PR!


cmd/uplink/cmd/setup.go
```diff
@@ -115,9 +115,27 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 			"api-key":        apiKey,
 			"enc.key":        string(encKey),
 		}
+
+		fmt.Println()
+		fmt.Println("Your Uplink CLI is configured and ready to use!")
```
jtolds: this should probably be below the actual process.SaveConfig call

mniewrzal: done

cmd/uplink/cmd/setup.go
```diff
@@ -71,19 +71,37 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
 		if err != nil {
 			return err
 		}
+		satellites := []string{"mars.tardigrade.io", "jupiter.tardigrade.io", "saturn.tardigrade.io"}
 		var satelliteAddress string
 		_, err = fmt.Scanln(&satelliteAddress)
 		if err != nil {
-			return err
+			// fmt.Scanln cannot handle empty input
+			satelliteAddress = satellites[0]
```
ifraixedes: we can detect an empty input just checking `err != nil && n == 0` where `n` is from `n, err = fmt.Scanln(...`

We could use that to show an error saying that `satelliteAddress` is required and stop or loop over until one option is selected or a manual address is written.

mniewrzal: added conditions about `n == 0`

We can add loop later if needed.

cmd/uplink/cmd/setup.go
```diff
@@ -114,17 +118,17 @@ Please enter numeric choice or enter satellite address manually [1]: `)
...
 		}
 
 		if apiKey == "" {
 			return errs.New("API key cannot be empty")
 		}
 
-		_, err = fmt.Print("Enter your encryption passphrase: ")
-		if err != nil {
+		n, err = fmt.Print("Enter your encryption passphrase: ")
+		if err != nil && n != 0 {
```
ifraixedes: `n != 0` isn't needed here

mniewrzal: ah, thanks

mniewrzal:
whole PR it will be squashed before merging

cmd/uplink/cmd/setup.go
```diff
@@ -71,19 +71,41 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
+				return err
+			}
 		}
 
 		// TODO add better validation
 		if satelliteAddress == "" {
-			return errs.New("Satellite address cannot be empty")
+			return errs.New("satellite address cannot be empty")
+		} else if len(satelliteAddress) == 1 {
+			switch satelliteAddress {
```
wthorp: nit:  this logic could would be easier to maintain if it were indexed based (if we ever add other satellites)

mniewrzal: yeah, current code is quite naive :) but I think we don't need anything special for now

PR #1738, opened 2019-04-10
[STALE] Switch console to use a single zip archive for assets

brimstone:
This isn't 100% ready to merge yet, but I want to start getting thoughts about the implementation.


go.mod
```diff
@@ -87,12 +87,12 @@ require (
 	github.com/pkg/profile v1.2.1 // indirect
 	github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a // indirect
 	github.com/rs/cors v1.5.0 // indirect
-	github.com/sirupsen/logrus v1.3.0 // indirect
```
brimstone: I don't know why this happened.

satellite/mailservice/service.go
```diff
@@ -66,10 +69,39 @@ func New(log *zap.Logger, sender Sender, templatePath string) (*Service, error)
 	//	return nil, err
 	//}
 
-	service.html, err = htmltemplate.ParseGlob(filepath.Join(templatePath, "*.html"))
+	//service.html, err = htmltemplate.ParseGlob(filepath.Join(templatePath, "*.html"))
+	r, err := zip.OpenReader("/tmp/app/assets.zip")
```
brimstone: This needs to know where the static assets zip is located. It probably needs a config option.

satellite/mailservice/service.go
```diff
@@ -66,10 +69,39 @@ func New(log *zap.Logger, sender Sender, templatePath string) (*Service, error)
...
+		if !strings.HasPrefix(f.Name, "static/emails") {
+			continue
+		}
+		if !strings.HasSuffix(f.Name, ".html") {
+			continue
+		}
+		baseName := filepath.Base(f.Name)
+		fh, err := f.Open()
+		if err != nil {
+			return nil, err
```
brimstone: Is there a better way to do this err chain in this service?

go.mod
jtolds: i also don't. it's okay though

jtolds: maybe just `git add -p` to not add this removal to your commit but add the zipfs one only?

go.sum
```diff
@@ -317,11 +311,11 @@ github.com/spf13/pflag v1.0.3 h1:zPAT6CGy6wXeQ7NtTnaTerfKOsV6V6F8agHXFiazDkg=
 github.com/spf13/pflag v1.0.3/go.mod h1:DYY7MBk1bdzusC3SYhjObp+wFpr4gzcvqqNjLnInEg4=
 github.com/spf13/viper v1.2.1 h1:bIcUwXqLseLF3BDAZduuNfekWG87ibtFxi59Bq+oI9M=
 github.com/spf13/viper v1.2.1/go.mod h1:P4AexN0a+C9tGAnUFNwDMYYZv3pjFuvmeiMyKRaNVlI=
+github.com/spkg/zipfs v0.7.0 h1:Ilog8m/WwtDM+Q0LS8Yjvl3S4mPKhE4WrtScIPUWmQw=
+github.com/spkg/zipfs v0.7.0/go.mod h1:48LW+/Rh1G7aAav1ew1PdlYn52T+LM+ARmSHfDNJvg8=
```
jtolds: maybe git add -p to only add the additions and not add the removals to your commit?

satellite/console/consoleweb/server.go
```diff
@@ -87,11 +89,16 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
 	}
 
 	mux := http.NewServeMux()
-	fs := http.FileServer(http.Dir(server.config.StaticDir))
+	zipfile, err := zipfs.New(server.config.StaticArchive)
+	if err != nil {
+		logger.Fatal(err.Error())
```
jtolds: this should return an error. This means you'll want to change `NewServer` to `OpenServer` and make it return `*Server, error`

satellite/console/consoleweb/server.go
```diff
@@ -107,9 +114,22 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
 	return &server
 }
 
+// serveFile serves a single file from the assets archive to the requester
+func (s *Server) serveFile(w http.ResponseWriter, req *http.Request, path ...string) {
+	f, err := s.assets.Open(filepath.Join(path...))
+	if err != nil {
+		panic(err)
```
jtolds:
```
http.Error(w, fmt.Sprintf("Internal server error: %v", err), 500)
return
```

satellite/console/consoleweb/server.go
```diff
@@ -107,9 +114,22 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
...
+	}
+	stat, err := f.Stat()
+	if err != nil {
+		panic(err)
```
jtolds: same

satellite/console/consoleweb/server.go
```diff
@@ -167,7 +187,7 @@ func (s *Server) bucketUsageReportHandler(w http.ResponseWriter, req *http.Reque
 		return
 	}
 
-	report, err := template.ParseFiles(path.Join(s.config.StaticDir, "static", "reports", "UsageReport.html"))
+	report, err := template.New("UsageReport.html").Parse(s.getFromArchive("static/reports/UsageReport.html"))
```
jtolds: i hope this works. what if the template depends on another template? hmm. let's try it for now, but this may be a source of template bugs. we might need to figure out how to get something like ParseFiles working

satellite/console/consoleweb/server.go
```diff
@@ -270,8 +290,21 @@ func (s *Server) passwordRecoveryHandler(w http.ResponseWriter, req *http.Reques
 	}
 }
 
+func (s *Server) getFromArchive(path string) string {
+	f, err := s.assets.Open(filepath.Join(path))
+	if err != nil {
+		panic(err)
```
jtolds: should return an error

satellite/console/consoleweb/server.go
jtolds: oh looks like s.serveError also

satellite/mailservice/service.go
jtolds: no this is fine

satellite/console/consoleweb/server.go
```diff
@@ -43,7 +44,7 @@ var Error = errs.Class("satellite console error")
 // Config contains configuration for console web server
 type Config struct {
 	Address         string `help:"server address of the graphql api gateway and frontend app" default:"127.0.0.1:8081"`
-	StaticDir       string `help:"path to static resources" default:""`
+	StaticArchive   string `help:"path to static resources" default:"assets.zip"`
```
egonelbre: Does console development workflow still work? It would be inconvenient to use zips for development.

I think we want to allow for using local filesystem as well and not constrain to only zips.

stefanbenten: I agree with that

brimstone: I agree.

PR #1739, opened 2019-04-10
rollup interval fixes

cam-a:
Standardize intervals to UTC because the same day with a different timezone creates a new map key. Delete latest day's total rollup rather than deleting latest day of bandwidth and storage individually


pkg/accounting/rollup/rollup.go
```diff
@@ -102,12 +106,12 @@ func (r *Service) RollupStorage(ctx context.Context, lastRollup time.Time, rollu
 	//loop through tallies and build Rollup
 	for _, tallyRow := range tallies {
 		node := tallyRow.NodeID
-		if tallyRow.CreatedAt.After(latestTally) {
-			latestTally = tallyRow.CreatedAt
+		interval := tallyRow.IntervalEndTime.UTC()
```
coyle: what sets `IntervalEndTime` ?

cam-a: `IntervalEndTime` is set to `time.Now()` right _before_ the return of `CalculateAtRestData`. `CreatedAt` is set when `time.Now()` right _after_ the return of `CalculateAtRestData` only nanoseconds later.
This change wasn't totally necessary since we'll be refactoring anyway to use the `storagenode_storage_tally` table which only has `interval_start`

pkg/accounting/rollup/rollup.go
```diff
@@ -174,11 +171,6 @@ func (r *Service) RollupBW(ctx context.Context, lastRollup time.Time, rollupStat
 			r.logger.Info("delete order type")
 		}
 	}
-	//remove the latest day (which we cannot know is complete), then push to DB
-	latestTally = time.Date(latestTally.Year(), latestTally.Month(), latestTally.Day(), 0, 0, 0, 0, latestTally.Location())
-	delete(rollupStats, latestTally)
-	if len(rollupStats) == 0 {
-		r.logger.Info("Rollup only found data for today")
-	}
+
```
JessicaGreben: i thought you had made a change to the tests to catch this bug? I thought you added .Location() somewhere to the tests?

pkg/accounting/rollup/rollup.go
JessicaGreben: any chance we can name `interval` something more descriptive since its hard to understand whats going on here. Like `tallyEndTime` or something. Also a comment would probably be helpful too. All these timestamps are confusing, at least you got rid of createdAt, one less timestamp! yay

jenlij:
Let's add a config flag to turn on and off the delete method.

pkg/accounting/rollup/rollup.go
```diff
@@ -66,21 +66,25 @@ func (r *Service) Rollup(ctx context.Context) error {
...
-		r.logger.Info("RollupStats is empty after RollupStorage")
-	}
+
 	err = r.RollupBW(ctx, lastRollup, rollupStats)
 	if err != nil {
 		return Error.Wrap(err)
 	}
+
+	//remove the latest day (which we cannot know is complete), then push to DB
+	latestTally = time.Date(latestTally.Year(), latestTally.Month(), latestTally.Day(), 0, 0, 0, 0, latestTally.Location())
```
jenlij: what does latestTally.Location() return? Does it work ok with UTC?

jenlij: Also we can wrap lines 76 and 77 in a conditional that checks the config for enablement

pkg/accounting/rollup/rollup.go
```diff
@@ -66,21 +66,25 @@ func (r *Service) Rollup(ctx context.Context) error {
...
 		return nil
 	}
+
 	err = r.db.SaveRollup(ctx, latestTally, rollupStats)
 	if err != nil {
 		return Error.Wrap(err)
 	}
+
 	// Delete already rolled up tallies
 	err = r.db.DeleteRawBefore(ctx, latestTally)
```
jenlij: let's wrap this in a conditional statement that checks the config

pkg/accounting/rollup/rollup.go
jenlij: Update here, I'm leaving the CreatedAt timestamp in storagenode_storage_tally https://github.com/storj/storj/pull/1769

jenlij: But if we decide to no longer use the createdAt timestamp let me know

pkg/accounting/rollup/rollup.go
```diff
@@ -102,12 +106,12 @@ func (r *Service) RollupStorage(ctx context.Context, lastRollup time.Time, rollu
...
-		if tallyRow.CreatedAt.After(latestTally) {
-			latestTally = tallyRow.CreatedAt
+		interval := tallyRow.IntervalEndTime.UTC()
+		if interval.After(latestTally) {
+			latestTally = interval
 		}
 		//create or get AccoutingRollup day entry
-		iDay := tallyRow.IntervalEndTime
-		iDay = time.Date(iDay.Year(), iDay.Month(), iDay.Day(), 0, 0, 0, 0, iDay.Location())
+		iDay := time.Date(interval.Year(), interval.Month(), interval.Day(), 0, 0, 0, 0, interval.Location())
```
jenlij: let's add interval.Hour() back in here since we may keep the 6 hr rollup interval

pkg/accounting/rollup/rollup.go
```diff
@@ -146,11 +143,11 @@ func (r *Service) RollupBW(ctx context.Context, lastRollup time.Time, rollupStat
...
 		nodeID := row.NodeID
-		if row.IntervalStart.After(latestTally) {
-			latestTally = row.IntervalStart
+		interval := row.IntervalStart.UTC()
+		if interval.After(latestTally) {
+			latestTally = interval
 		}
-		day := row.IntervalStart
-		day = time.Date(day.Year(), day.Month(), day.Day(), 0, 0, 0, 0, day.Location())
+		day := time.Date(interval.Year(), interval.Month(), interval.Day(), 0, 0, 0, 0, interval.Location())
```
jenlij: here too

pkg/accounting/rollup/rollup.go
cam-a: Inside ```RollupStorage``` I made sure to convert latestTally to UTC

pkg/accounting/rollup/rollup.go
cam-a: I implemented this, but need to figure out how to set flags in testplanet because it's being set to false and testrollup is failing as a result because we assert that raws were deleted

PR #1744, opened 2019-04-11
Allow the satellite to connect to other mail servers

brimstone:
I broke sending email from Mars in a previous PR. This should fix it again.

### DIFF
--- a/cmd/satellite/Dockerfile
+++ b/cmd/satellite/Dockerfile
@@ -12,6 +12,9 @@ COPY web/satellite/ /app
 RUN npm install
 RUN npm run build
 
+FROM alpine as ca-cert
+RUN apk -U add ca-certificates
+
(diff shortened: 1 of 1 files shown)


cmd/satellite/Dockerfile
```diff
@@ -12,6 +12,9 @@ COPY web/satellite/ /app
 RUN npm install
 RUN npm run build
 
+FROM alpine as ca-cert
```
egonelbre: Should we add this to other Dockerfiles as well?

brimstone: Only if the other container images need to verify the TLS cert chain of sites that don't use Let's Encrypt. I don't think the other images do. This one does because it reaches out to a mailserver that doesn't use Let's Encrypt.

PR #1749, opened 2019-04-12
libuplink code example

kaloyan-raev:
What: 

Example program for using libuplink.

Why:

Because we need it for a blog post and reference in general.

Thanks for submitting a PR!


egonelbre:
Two important things I have here:

1. it's not a realistic example, which means it's hard to understand what's going on.
2. maybe it should be a separate repository?

As for the example I'm not sure what would be a good use case... maybe some simple tool that you can use like:

```
transfer upload somefile.txt
transfer download somefile.txt
transfer share somefile.txt
```

A simpler version of `cmd/uplink`.

Or something like:
```
backup upload -tag "images" .
backup list
# images/images-2018-03-01.tar.gz
# images/images-2018-03-02.tar.gz
# images/images-2018-03-03.tar.gz
backup download images/images-2018-03-03.tar.gz
```

kaloyan-raev:
The example in this PR is the one from a blog post that will be published in the next days. It is not meant to be a complete application built on top of libuplink, but give a basic introduction on how to use libuplink.

jtolds:
as per egon's comments, this might be better as an Example in the go test parlance: https://blog.golang.org/examples

kaloyan-raev:
How am I supposed to implement this as a testable example? Testable examples do not seem to receive `t *testing.T` like other golang tests. So I can't use TestPlanet.

egonelbre:
@kaloyan-raev `testplanet.NewWithLogger` or `testplanet.NewCustom` can be used without `testing.T`.

jtolds:
i'm also okay with it being a small tool instead of a testable example

examples/libuplink/main.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("got different object back: %q != %q", dataToUpload, receivedContents)
+	}
+	return nil
+}
+
+func main() {
+	var encryptionKey storj.Key
+	copy(encryptionKey[:], []byte(myEncryptionKey))
+
+	apiKey, err := uplink.ParseAPIKey(myAPIKey)
```
ifraixedes: What about to get the API Key from the first arg `os.Args[1]`? it's very simple to add and it wont't need to change the file for running the example.

kaloyan-raev: This is possible. As well as for the rest of the constants. But this makes it a little bit more error-prone to readers, e.g. handling spaces in program args.

This is supposed to be a simple example which marketing wants to publish in a blog post very soon. I'd not make such changes to avoid confusing them. We also want readers to have a quick win. So we want to save them from any chance of doing anything wrong.

kaloyan-raev:
I added a testable example that uses test planet.

Now the PR includes both the initial example program and the testable example. The code of the testable example is complicated due to the use of test planet.

I suggest that we keep both the example program and the testable example as they serve different purposes and both are valuable assets.

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil
+	})
+
+	fmt.Println("success!")
+
+	// Output:
+	// success!
+}
+
+func setupTestEnv(test func(context.Context, string, string, storj.RedundancyScheme) error) {
```
egonelbre: rather than use callbacks here return the planet, satellite address, api key and redundancy

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/memory"
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/lib/uplink"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/console"
+)
+
+const (
```
egonelbre: Put these inside the example, otherwise they aren't visible in godoc

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+	testProject       = "test-project"
```
egonelbre: Let's make the example about note taking, so project `notes`

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+	testBucket        = "test-bucket"
```
egonelbre: `work`

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+	testUploadPath    = "foo/bar/baz"
```
egonelbre: `presentations/2018/storj.txt`

lib/uplink/example_test.go
```diff
@@ -0,0 +1,186 @@
...
+	testDataSize      = 8 * memory.KiB
```
egonelbre: with `"Demonstrate how to upload to Storj."`

mniewrzal:
What we are doing with this PR? its approved. is it ready to merge (after conflict resolution)?

jtolds:
there are newer examples in libuplink by now

PR #1752, opened 2019-04-12
Check context cancellation more nicely

egonelbre:
What: Try to unwrap to find whether the underlying is a cancellation rather than only checking at the top-level.

Why: Sometimes tests are being flaky, because databases may return `Error.Wrap(context.Canceled)`, but it doesn't explicitly check against it.

Thanks for submitting a PR!


internal/errs2/ignore.go
```diff
@@ -0,0 +1,31 @@
...
+func IgnoreCanceled(originalError error) error {
+	err := originalError
+	for err != nil {
+		if err == context.Canceled ||
+			err == grpc.ErrServerStopped ||
+			err == http.ErrServerClosed {
+			return nil
+		}
+		unwrapped := errs.Unwrap(err)
+		if unwrapped == err {
```
mobyvb: what does this case mean?

egonelbre: It is a sanity check to avoid infinite loop when `errs.Unwrap` returns the `err` itself.

internal/errs2/ignore.go
```diff
@@ -0,0 +1,31 @@
...
+			return originalError
```
mobyvb: we could just break, but this works too

egonelbre:
@navillasa I don't think `errhelper` is much more descriptive, but I would like a better name.

navillasa:
> @navillasa I don't think `errhelper` is much more descriptive, but I would like a better name.

`errs2` makes me think it's an offshoot of zeebo/errs. I think `errhelper` or `errhandler` at least distinguishes it from that.

egonelbre:
@navillasa I'll merge as `errs2` for now and let's try to figure out a better name later, once we consolidate some of the error handling helpers it might be easier to find a name.

PR #1756, opened 2019-04-15
pkg/eestream: remove unneeded bytes copy

ifraixedes:
What: 

Remove an unneeded bytes copy and also simplify the logic on
decoderReader.Read method.

Why:

The piece of code should be simpler to understand and slightly performant.

Thanks for submitting a PR!

### DIFF
--- a/pkg/eestream/decode.go
+++ b/pkg/eestream/decode.go
@@ -86,12 +86,10 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
 		dr.currentStripe++
 	}
 
-	// copy what data we have to the output
+	// copy what data we have to the output, and shrink the remaining buffer
 	n = copy(p, dr.outbuf)
-	// slide the remaining bytes to the beginning
-	copy(dr.outbuf, dr.outbuf[n:])
-	// shrink the remaining buffer
-	dr.outbuf = dr.outbuf[:len(dr.outbuf)-n]
+	dr.outbuf = dr.outbuf[n:]
+
 	return n, nil
 }
 


pkg/eestream/decode.go
```diff
@@ -86,12 +86,10 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
...
 	}
 
-	// copy what data we have to the output
+	// copy what data we have to the output, and shrink the remaining buffer
 	n = copy(p, dr.outbuf)
-	// slide the remaining bytes to the beginning
-	copy(dr.outbuf, dr.outbuf[n:])
-	// shrink the remaining buffer
-	dr.outbuf = dr.outbuf[:len(dr.outbuf)-n]
+	dr.outbuf = dr.outbuf[n:len(dr.outbuf)]
```
egonelbre: I'm not quite certain that this will be more performant in practice. But, I'm not certain that it's slower either. The previous approach avoids reallocating buffers, and I think `p` should usually read most of the `outbuf`... so it actually won't end up copying a lot.

So, I would like to see a benchmark showing the difference, either in allocations or performance.

Also it can be just `dr.outbuf = dr.outbuf[n:]`.

ifraixedes: More performant may not, without benchmarking it isn't something known, but mostly the change is, at least for me, because it's less lines of code and simpler to understand.

> Also it can be just dr.outbuf = dr.outbuf[n:].

Yes, that's even better.

pkg/eestream/decode.go
```diff
@@ -86,12 +86,10 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
...
 	}
 
-	// copy what data we have to the output
+	// copy what data we have to the output, and shrink the remaining buffer
 	n = copy(p, dr.outbuf)
-	// slide the remaining bytes to the beginning
-	copy(dr.outbuf, dr.outbuf[n:])
-	// shrink the remaining buffer
-	dr.outbuf = dr.outbuf[:len(dr.outbuf)-n]
+	dr.outbuf = dr.outbuf[n:]
```
jtolds: i think this is a memory leak. isn't it? over time, the allocated size of the backing buffer will grow and grow, even though the slice that is referenced is smaller. perhaps append will occasionally lop off old allocations but that will increase gc pressure.

jtolds: okay not a memory leak. looks like infectious.Decode never appends to the existing outbuf but just reuses it. 

this does increase garbage collector pressure though. by sliding the outbuf out to the right the existing code will allocate a new buffer when it no longer has enough space. i'd be surprised if this actually makes things more performant

pkg/eestream/decode.go
egonelbre: The code is more complex than it should be and I would guess someone who writes the complex version, probably knows the simpler version. This suggest to me, that there was a reason why the complex version was chosen over the simpler one.

pkg/eestream/decode.go
jtolds: oh, looks like you talked it through with egon. i didn't see the earlier convo. it is indeed a matter of avoiding new buffer allocations.

pkg/eestream/decode.go
ifraixedes: it sounds reasonably, my thoughts were that it was something written that it escaped a code review and you know sometimes when focusing on the code and with some back and forth those things sometimes land without that much polishing.

pkg/eestream/decode.go
ifraixedes: > this does increase garbage collector pressure though. by sliding the outbuf out to the right the existing code will allocate a new buffer when it no longer has enough space

That's right if you append and you want to reuse the buffer, however I didn't find where that append happens with my quick view, I thought that it was only read once.

btw, I found this reading code and trying to understand some logic, not looking exactly to improve performance or anything else, I thought that it was a matter of something that escaped from a code review.

ifraixedes:
Thanks for the explanation guys!

PR #1758, opened 2019-04-15
Make tracked branch feature work with multi-arch manifests

brimstone:
This should make the `push-images` target complete successfully and push binaries to GCS again.


brimstone:
Same, if I knew how manifests worked, and could remove or overwrite a local one, then I could leave this in.

brimstone:
@stefanbenten I think i figured it out. I invalidated your approval since I changed several things and basically this whole PR.

PR #1761, opened 2019-04-15
libuplink: remove encryption key from project opening

jtolds:
What: This change moves project-level bucket metadata encryption information to the volatile section, because it is unlikely to remain in future releases

Why: Ultimately, the web user interface will allow bucket management (creation, removal, etc), but not object management as that requires an encryption key for sure and we don't want to have users give the satellite their encryption keys.

At a high level, a (*Project) type should map to all of the things you can do inside the web user interface within a project, which by necessity cannot have an encryption key. So, we really don't want an encryption key in the non-volatile section of this library.

Thanks for submitting a PR!

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -240,17 +240,20 @@ func (flags GatewayFlags) NewGateway(ctx context.Context, ident *identity.FullId
 		return nil, err
 	}
 
-	encKey := new(storj.Key)
+	var encKey storj.Key
 	copy(encKey[:], flags.Enc.Key)
 
-	project, err := uplink.OpenProject(ctx, flags.Client.SatelliteAddr, encKey, apiKey)
+	var opts libuplink.ProjectOptions
+	opts.Volatile.EncryptionKey = encKey
+
+	project, err := uplink.OpenProject(ctx, flags.Client.SatelliteAddr, apiKey, &opts)
 	if err != nil {
 		return nil, err
 	}
 
 	return miniogw.NewStorjGateway(
 		project,
-		encKey,
+		&encKey,
 		storj.Cipher(flags.Enc.PathType).ToCipherSuite(),
 		flags.GetEncryptionScheme().ToEncryptionParameters(),
 		flags.GetRedundancyScheme(),
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -240,8 +240,8 @@ func (flags GatewayFlags) NewGateway(ctx context.Context, ident *identity.FullId
 		return nil, err
 	}
 
-	var encKey storj.Key
-	copy(encKey[:], flags.Enc.Key)
+	encKey := new(storj.Key)
+	copy((*encKey)[:], flags.Enc.Key)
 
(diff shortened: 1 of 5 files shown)


jtolds:
should i move satelliteAddr and apiKey into the ProjectOptions struct too and make it required?

lib/uplink/uplink.go
```diff
@@ -158,8 +165,12 @@ func (u *Uplink) OpenProject(ctx context.Context, satelliteAddr string, encrypti
 		return nil, Error.New("failed to create redundancy strategy: %v", err)
 	}
 	segments := segments.NewSegmentStore(metainfo, nil, rs, maxBucketMetaSize.Int(), maxBucketMetaSize.Int64())
+	var encryptionKey storj.Key
+	if opts != nil {
+		encryptionKey = opts.Volatile.EncryptionKey
```
navillasa: I'm not sure if this case would arise but I feel like there should be a check that `opts.Volatile` isn't nil so that there won't be panics.

jtolds: oh i see what you mean, yeah! opts.Volatile is actually a struct, so i don't think it's possible to even compare it to nil but it might be worth a comment?

navillasa: oh I see- I think it should be fine then.

kaloyan-raev:
> should i move satelliteAddr and apiKey into the ProjectOptions struct too and make it required?

My understanding for `ProjectOptions` is that it contains optional configuration that we have reasonable defaults for if omitted. With the exception of the Volatile section, which is a temporary thing.

With regards to the above, the `apiKey` is a required parameter that we cannot have a default value for, so it does not make sense moving it to the `ProjectOptions`.

Maybe it makes sense for the `satelliteAddr` if we want to have a default satellite. But if we don't, then we should leave it as a required parameter too.

lib/uplink/uplink.go
```diff
@@ -138,8 +138,15 @@ func NewUplink(ctx context.Context, cfg *Config) (*Uplink, error) {
 	}, nil
 }
 
+// ProjectOptions allows configuration of various project options during opening
+type ProjectOptions struct {
+	Volatile struct {
+		EncryptionKey storj.Key
```
kaloyan-raev: We usually have `storj.Key` fields and params declared as pointers to avoid having many copies of this sensitive information in the memory. Shouldn't we have it declared here as a pointer too?

jtolds: oh, i see, that's why that's happening. okay, i'll update

cmd/gateway/main.go
```diff
@@ -241,9 +241,12 @@ func (flags GatewayFlags) NewGateway(ctx context.Context, ident *identity.FullId
 	}
 
 	encKey := new(storj.Key)
-	copy(encKey[:], flags.Enc.Key)
+	copy((*encKey)[:], flags.Enc.Key)
```
kaloyan-raev: Do we really need this code change?

jtolds: whoops, sorry, no

PR #1762, opened 2019-04-15
uplink setup wizard uses default port, checks errors

jtolds:
We had a neat uplink setup wizard PR and a default satellite addr PR. this PR combines them so the wizard gets the default address functionality. also checks errors because I couldn't help myself

Thanks for submitting a PR!

### DIFF
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -71,43 +71,75 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 
 	var override map[string]interface{}
 	if !setupCfg.NonInteractive {
-		fmt.Print("Enter your Satellite address: ")
+		_, err = fmt.Print("Enter your Satellite address: ")
+		if err != nil {
+			return err
+		}
 		var satelliteAddress string
-		fmt.Scanln(&satelliteAddress)
+		_, err = fmt.Scanln(&satelliteAddress)
+		if err != nil {
+			return err
+		}
 
 		// TODO add better validation
 		if satelliteAddress == "" {
 			return errs.New("Satellite address cannot be empty")
 		}
 
-		fmt.Print("Enter your API key: ")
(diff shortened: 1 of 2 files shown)


cmd/uplink/cmd/setup.go
```diff
@@ -122,49 +154,58 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
+}
+
+// ApplyDefaultHostAndPortToAddr applies the default host and/or port if either is missing in the specified address.
+func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, error) {
+	defaultHost, defaultPort, err := net.SplitHostPort(defaultAddress)
+	if err != nil {
+		return "", Error.Wrap(err)
+	}
 
 	addressParts := strings.Split(address, ":")
```
egonelbre: let's use here `net.SplitHostPort` here as well

simongui: We can't do that. See the [details here](https://github.com/storj/storj/pull/1723) why I had to do it this way.

jtolds: one problem worth considering still egon brings up is if `address` is an ipv6 address, this code won't work correctly, which net.SplitHostPort attempts to fix

simongui: Ahh right. Depends what we want to do then.`net.SplitHostPort()` won't work with `:8888`. If we want to use `net.SplitHostPort()` we lose that ability.

If we want to retain that use case we could check for multiple `:` instances. It's a bit ugly but it would work.

jtolds: i think we'll probably always have the hostname, so if we had to pick between supporting ipv6 and supporting the no-hostname case, i'd probably pick the latter. but okay to leave for another pr

egonelbre: PS: when unable to do the obvious thing leave comment for why it cannot be used, helps to avoid someone accidentally "fixing it".

PR #1763, opened 2019-04-16
sqlite3 key/value store + sqlite revocation DB

bryanchriswhite:
What: This PR adds a key/value store adapter for sqlite3 and makes sqlite3 available as an option for backing a revocation database.

Why: Bolt-backed revocation databases don't support concurrent access and redis is not a desirable dependency for some/most use cases; sqlite support should provide an option in between.

Thanks for submitting a PR!


storage/sqlitekv/sqlitekv.go
```diff
@@ -0,0 +1,167 @@
...
+	row, err := db.DB.Find_Item_Value_By_Key(context.Background(), Item_Key(key))
+	if err != nil {
+		return nil, ErrSqlitekv.Wrap(err)
+	}
+	return row.Value, nil
+}
+
+// Put adds a value to store
+func (db SqliteKV) Put(key storage.Key, value storage.Value) error {
+	// TODO: should we add contexts to the key-value store interface?
```
ifraixedes: Why not?

What I see on not adding it now is:
Normally any store should have the context because the most of them rely on I/O or network access, adding it later to the methods will be breaking changes and refactoring will have to be taken

bryanchriswhite:
Closing because this PR is old and bad.

PR #1764, opened 2019-04-16
Download is hanging when one node is not responsive

mniewrzal:
What: Change closes download connection in parallel and sets a timeout for communication between uplink and storage node (uplink side)

Why: Currently downloading even small (10MB) file can hang if one of storage node is not responding. Download operation hangs forever and the process can be only terminated. 

https://storjlabs.atlassian.net/browse/V3-1547

Thanks for submitting a PR!


pkg/eestream/decode.go
```diff
@@ -101,18 +101,24 @@ func (dr *decodedReader) Close() error {
 	// avoid double close of readers
 	errorThreshold := len(dr.readers) - dr.scheme.RequiredCount()
 	dr.close.Do(func() {
-		var errlist errs.Group
-		// close the readers
+		var errlist errgroup.Group
```
egonelbre: `errgroup` returns only the first error instead of all of them

mniewrzal: do we have something that will fit here or I just need to use for example `sync.WaitGroup`?

egonelbre: As far as I know, no, but might makes sense to implement one.

egonelbre:
Add "Why" to the issue and test for it.

mniewrzal:
Ok, in a moment. Its still wip and mainly for testing.

mniewrzal:
I think it can be reviewed but still looking for a way to test it.

pkg/eestream/decode.go
```diff
@@ -102,16 +102,29 @@ func (dr *decodedReader) Close() error {
 	errorThreshold := len(dr.readers) - dr.scheme.RequiredCount()
 	dr.close.Do(func() {
 		var errlist errs.Group
```
egonelbre: Let's make a nice type for it in `errs2.` e.g. `errs2.WaitGroup` or something.

mniewrzal: added `errs2.Group`

internal/errs2/group.go
```diff
@@ -0,0 +1,37 @@
...
+// Go calls the given function in a new goroutine.
+func (group *Group) Go(f func() error) {
+	group.wg.Add(1)
+
+	go func() {
+		defer group.wg.Done()
+
+		if err := f(); err != nil {
+			group.errOnce.Do(func() {
+				group.errors = append(group.errors, err)
```
egonelbre: This returns only one error.

internal/errs2/group.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Go calls the given function in a new goroutine.
+func (group *Group) Go(f func() error) {
+	group.wg.Add(1)
+
+	go func() {
+		defer group.wg.Done()
+
+		if err := f(); err != nil {
+			group.errors = append(group.errors, err)
```
egonelbre: this is a data race

mniewrzal: ah, sorry, I should focus on one thing at once

pkg/transport/timeout.go
```diff
@@ -22,3 +23,71 @@ func (it InvokeTimeout) Intercept(ctx context.Context, method string, req interf
...
+	})
+}
+
+func (wrapper *clientStreamWrapper) RecvMsg(m interface{}) error {
+	return wrapper.addTimeout(func() error {
+		return wrapper.stream.RecvMsg(m)
+	})
+}
+
+func (wrapper *clientStreamWrapper) addTimeout(f func() error) error {
```
egonelbre: `withTimeout`

internal/errs2/group.go
```diff
@@ -0,0 +1,39 @@
...
+
+import "sync"
+
+// Group is a collection of goroutines working on subtasks that are part of
+// the same overall task.
+type Group struct {
+	wg     sync.WaitGroup
+	errors []error
+
+	mutex sync.Mutex
```
egonelbre: Order these such it can be understood what is being protected and how.

egonelbre:
```
group  sync.WaitGroup
mu     sync.Mutex
errors []error
```

pkg/eestream/decode.go
```diff
@@ -98,21 +99,25 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
 func (dr *decodedReader) Close() error {
 	// cancel the context to terminate reader goroutines
 	dr.cancel()
-	// avoid double close of readers
 	errorThreshold := len(dr.readers) - dr.scheme.RequiredCount()
+	closeGroup := &errs2.Group{}
```
egonelbre: No need for the pointer, use `var closeGroup errs2.Group`

pkg/eestream/decode.go
```diff
@@ -98,21 +99,25 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
...
 	dr.close.Do(func() {
-		var errlist errs.Group
-		// close the readers
 		for _, r := range dr.readers {
-			err := r.Close()
-			if err != nil {
-				errlist.Add(err)
-				errorThreshold--
-			}
+			r := r
```
jtolds: this is such a fun go footgun. might be worth leaving a comment how necessary this line is and what subtle bugs will befall you if you remove it

pkg/eestream/decode.go
```diff
@@ -98,21 +99,25 @@ func (dr *decodedReader) Read(p []byte) (n int, err error) {
...
+			closeGroup.Go(func() error {
+				return r.Close()
+			})
 		}
+
 		// close the stripe reader
-		errlist.Add(dr.stripeReader.Close())
-		dr.closeErr = errlist.Err()
+		closeGroup.Go(func() error {
```
jtolds: can you just do `closeGroup.Go(dr.stripeReader.Close)`?

pkg/transport/timeout.go
```diff
@@ -22,3 +23,71 @@ func (it InvokeTimeout) Intercept(ctx context.Context, method string, req interf
...
+
+func (wrapper *clientStreamWrapper) RecvMsg(m interface{}) error {
+	return wrapper.withTimeout(func() error {
+		return wrapper.stream.RecvMsg(m)
+	})
+}
+
+func (wrapper *clientStreamWrapper) withTimeout(f func() error) error {
+	timoutTicker := time.NewTicker(wrapper.timeout)
+	defer timoutTicker.Stop()
```
jtolds:
```
doneCh := make(chan struct{})
defer close(doneCh)
```

pkg/transport/timeout.go
```diff
@@ -22,3 +23,71 @@ func (it InvokeTimeout) Intercept(ctx context.Context, method string, req interf
...
+
+	go func() {
+		select {
+		case <-timoutTicker.C:
+			wrapper.cancel()
+		case <-wrapper.Context().Done():
```
jtolds:
```case <-doneCh:
```

pkg/transport/timeout.go
jtolds: this makes it so we can clean up the goroutine early if f returns early

PR #1770, opened 2019-04-17
Fix RSConfig calculation in testplanet.Uplink.GetConfig

kaloyan-raev:
What: 

Changes `testplanet.Uplink.GetConfig` in a way to ensure that any of the redundancy properties will be higher than 0, as `0` or lower is not a valid value for them.

Why:

If a test creates test planet with less than 5 storage nodes then `testplanet.Uplink.GetConfig` will return invalid redundancy configuraiton. 

Thanks for submitting a PR!

### DIFF
--- a/internal/testplanet/uplink.go
+++ b/internal/testplanet/uplink.go
@@ -298,10 +298,10 @@ func (uplink *Uplink) GetConfig(satellite *satellite.Peer) uplink.Config {
 	config.Client.SatelliteAddr = satellite.Addr()
 	config.Client.APIKey = uplink.APIKey[satellite.ID()]
 
-	config.RS.MinThreshold = 1 * uplink.StorageNodeCount / 5
-	config.RS.RepairThreshold = 2 * uplink.StorageNodeCount / 5
-	config.RS.SuccessThreshold = 3 * uplink.StorageNodeCount / 5
-	config.RS.MaxThreshold = 4 * uplink.StorageNodeCount / 5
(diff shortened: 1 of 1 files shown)


internal/testplanet/uplink.go
```diff
@@ -298,10 +298,10 @@ func (uplink *Uplink) GetConfig(satellite *satellite.Peer) uplink.Config {
 	config.Client.SatelliteAddr = satellite.Addr()
 	config.Client.APIKey = uplink.APIKey[satellite.ID()]
 
-	config.RS.MinThreshold = 1 * uplink.StorageNodeCount / 5
-	config.RS.RepairThreshold = 2 * uplink.StorageNodeCount / 5
-	config.RS.SuccessThreshold = 3 * uplink.StorageNodeCount / 5
-	config.RS.MaxThreshold = 4 * uplink.StorageNodeCount / 5
+	config.RS.MinThreshold = atLeastOne(1 * uplink.StorageNodeCount / 5)
```
ifraixedes: I guess that `1 * ..` is to make clear something, isn't it?

egonelbre: If we move the multiplication to the end it should become more obvious `uplink.StorageNodeCount * 1 / 5`

ifraixedes: from my point of view is the same, I mostly was asking specifically for the multiplication by 1

kaloyan-raev: Comments added. Let me know if it requires more clarification.

PR #1775, opened 2019-04-17
hardcode debug ports for storj-sim

JessicaGreben:
What:  Hardcode the debug ports for storj-sim

Why: The default behavior is to pick a port at random for debug, however this is annoying because if you want to use debug endpoints, you have to go search for what the port is.

### DIFF
--- a/cmd/storj-sim/network.go
+++ b/cmd/storj-sim/network.go
@@ -53,7 +53,7 @@ const (
 // port creates a port with a consistent format for storj-sim services.
 // The port format is: "1PXXE", where P is the peer class, XX is the index of the instance, and E is the endpoint.
 func port(peerclass, index, endpoint int) string {
-	port := 10000 + int(peerclass)*1000 + index*10 + endpoint
+	port := 10000 + peerclass*1000 + index*10 + endpoint
 	return strconv.Itoa(port)
 }
 
--- a/cmd/storj-sim/network.go
+++ b/cmd/storj-sim/network.go
@@ -25,7 +25,43 @@ import (
 	"storj.io/storj/internal/processgroup"
 )
 
+type peerClass int
+
+func (p peerClass) String() string {
+	return []string{"Satellite", "Storage Node", "Version Control", "Boostrap", "Gateway"}[p]
+}
+
+const (
+	// The following values of peer class, index, and endpoints are used
+	// to create a port with a consistent format for storj-sim services.
+
+	// Peer class
+	satellitePeer peerClass = iota
+	storagenodePeer
+	gatewayPeer
+	versioncontrolPeer
+	bootstrapPeer
+
+	// Index for peers with one instance (i.e. version control and bootstrap)
+	singleIndex = 0
+
+	// Endpoint
+	publicGRPC  = 0
+	privateGRPC = 1
+	publicHTTP  = 2
+	debugHTTP   = 9
+)
+
 const folderPermissions = 0744
+
+// port creates a port with a consistent format for storj-sim services.
+// The port format is: "1PXXE", where P is the peer class, XX is the index of the instance, and E is the endpoint.
+func port(peer peerClass, index, endpoint int) (string, error) {
+	if index > 99 {
+		return "", fmt.Errorf("reached max of 100 instances of %s peer class", peer.String())
+	}
+	return strconv.Itoa(10000 + int(peer)*1000 + index*10 + endpoint), nil
+}
 
 func networkExec(flags *Flags, args []string, command string) error {
 	processes, err := newNetwork(flags)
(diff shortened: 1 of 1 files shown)


cmd/storj-sim/network.go
```diff
@@ -148,17 +148,22 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
+		gatewayDebugPort        = 9500
+		bootstrapPort           = 9999
+		bootstrapPrivatePort    = 9988
+		bootstrapDebugPort      = 9977
+		satellitePort           = 10000
+		satellitePrivatePort    = 11000
+		satelliteDebugPort      = 11500
+		storageNodePort         = 12000
+		storageNodePrivatePort  = 13000
+		storageNodeDebugPort    = 13500
```
littleskunk: Port 14500 for the storage nodes would be better. Port 13500 will allow us to run only 500 storage nodes. I don't know if I will run a storj-sim network with more than 500 storage nodes but I would like to have the option if needed.

cmd/storj-sim/network.go
```diff
@@ -148,17 +148,22 @@ func newNetwork(flags *Flags) (*Processes, error) {
...
+		consolePort             = 10100
+		bootstrapWebPort        = 10010
+		versioncontrolPort      = 10011
+		versioncontrolDebugPort = 10012
```
egonelbre: Let's figure out a stable way to set ports to things so it would be easy to figure out them.

Maybe something like:

```
func port(peer, index, subsystem int) int {
    return 10000 + peer * 1000 + index * 10 + port
}
```

where

```
satellite      = 0
storage node   = 1-
uplink         = 5-

single         = 9
   bootstrap      = 0
   versioncontrol = 1

public_grpc  = 0
private_grpc = 1

public_http  = 2

debug_http = 9
```

* a satellite 42 public http port would be `port(satellite, 42, public_http)`, which would be `1-0-42-2`. (this is the console address)
* a storage node 1 debug port would be `port(storage node, 1, debug_http)`, which would be `1-1-01-9`.
* a bootstrap debug port would be `port(single, bootstrap, debug_http)`, which would be `1-9-00-9`.
* uplink 3 public http (gateway) would be `port(uplink, 3, public_http)`, which is `1-5-03-2`

egonelbre: `1XZZW`, where `X` is the peer class, `ZZ` is the index and `W` the endpoint.

stefanbenten: With this notation you are jumping out of the max range of 65539 when running more than 100 of a kind

egonelbre: `100 * 10 == 1000`

stefanbenten: Yeah which means with your notation it ends up like: 1-1-1000-9. Else i dont understand your port collection.

egonelbre:
```
func port(peerclass, index, endpoint int) int {
    return 10000 + peerclass * 1000 + index * 10 + port
}
```

`port(storage node, 153, 3) = 12533`

egonelbre: the index overflows into the peerclass index

stefanbenten: Oh okay, yeah that works. The Notation was misleading me :)

egonelbre: I was also considering using the first number as the peer class, but that would make the ports use across a really wide range. I'm not sure whether that's nice from usability.

```
func portalternate(peerclass, index, endpoint int) int {
    return peerclass*10000 + index * 10 + port
}

satellite      = 0
storage node = 1
uplink         = 2

single         = 3
   // where index is for single is
   bootstrap      = 0
   versioncontrol = 1
```

egonelbre: Since most of the people won't be running more than 100 nodes locally anyways, I think it's better to keep the range smaller.

cmd/storj-sim/network.go
```diff
@@ -25,8 +25,56 @@ import (
...
 
+// port creates a port with a consistent format for storj-sim services.
+// The port format is: "1PXXE", where P is the peer class, XX is the index of the instance, and E is the endpoint.
+func port(peer peerClass, index, endpoint int) (string, error) {
+	const (
+		maxInstanceCount    = 100
+		maxStoragenodeCount = 200
+	)
+
+	switch {
```
egonelbre: I would move this check into flag checking. That way you don't need to return an error and can keep rest of the code cleaner.

cmd/storj-sim/network.go
```diff
@@ -25,8 +25,56 @@ import (
...
+func (p peerClass) String() string {
+	return []string{"Satellite", "Version Control", "Boostrap", "Gateway", "Storage Node"}[p]
+}
+
+const (
+	// The following values of peer class, index, and endpoints are used
+	// to create a port with a consistent format for storj-sim services.
+
+	// Peer class
+	satellitePeer peerClass = iota
```
egonelbre: Use explicit values here instead of iota. When defining constants that should be preserved from version to version, then having explicit values makes it more obvious.

cmd/storj-sim/network.go
```diff
@@ -25,8 +25,56 @@ import (
...
+	gatewayPeer
+	versioncontrolPeer
+	bootstrapPeer
+	storagenodePeer
+
+	// Index for peers with one instance (i.e. version control and bootstrap)
+	singleIndex = 0
```
egonelbre: Let's remove this and use number `0` in the code. The use of `single` here is different than I had in mind.

cmd/storj-sim/network.go
JessicaGreben: good idea. moving this to the flag checking made this much simpler.

PR #1782, opened 2019-04-18
v3-1541: Change where the encryption key is being stored for uplink

ifraixedes:

Thanks for submitting a PR!

### DIFF
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -69,7 +70,13 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	var override map[string]interface{}
+	var (
+		encKeyFilepath = filepath.Join(setupDir, ".enc.key")
+		encKey         []byte
+		override       = map[string]interface{}{
+			"enc.key-filepath": encKeyFilepath,
+		}
+	)
 	if !setupCfg.NonInteractive {
 		_, err = fmt.Print("Enter your Satellite address: ")
 		if err != nil {
@@ -90,6 +97,7 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		if err != nil {
 			return err
 		}
+		override["satellite-addr"] = satelliteAddress
 
 		_, err = fmt.Print("Enter your API key: ")
 		if err != nil {
@@ -104,12 +112,13 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		if apiKey == "" {
 			return errs.New("API key cannot be empty")
 		}
+		override["api-key"] = apiKey
 
 		_, err = fmt.Print("Enter your encryption passphrase: ")
 		if err != nil {
 			return err
 		}
-		encKey, err := terminal.ReadPassword(int(os.Stdin.Fd()))
+		encKey, err = terminal.ReadPassword(int(os.Stdin.Fd()))
 		if err != nil {
 			return err
 		}
@@ -140,12 +149,11 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 			if err != nil {
 				return err
 			}
-		}
-
-		override = map[string]interface{}{
-			"satellite-addr": satelliteAddress,
-			"api-key":        apiKey,
-			"enc.key":        string(encKey),
+		} else {
+			err = SaveEncryptionKey(encKey, encKeyFilepath)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
@@ -209,4 +217,38 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 
 	// address is host:
 	return net.JoinHostPort(addressParts[0], defaultPort), nil
+}
+
+// SaveEncryptionKey saves the key in a new file which will be stored in
+// filepath.
+// It returns an error if the directory doesn't exist, the file already exists
+// or there is an I/O error.
+func SaveEncryptionKey(key []byte, filepath string) (err error) {
+	f, err := os.OpenFile(filepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errors.New("directory path doesn't exist")
+		}
+
+		if os.IsExist(err) {
+			return errors.New("file key already exists")
+		}
+
+		return err
+	}
+
+	defer func() {
+		if err == nil {
+			err = f.Close()
+		} else {
+			_ = f.Close()
+		}
+	}()
+
+	_, err = f.Write(key)
+	if err != nil {
+		return err
+	}
+
+	return f.Chmod(0400)
 }
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -149,7 +149,8 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
-	return process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), override)
+	return process.SaveConfigWithAllDefaults(
+		cmd.Flags(), filepath.Join(setupDir, process.DefaultConfFilename), override)
 }
 
 // ApplyDefaultHostAndPortToAddrFlag applies the default host and/or port if either is missing in the specified flag name.
--- a/cmd/uplink/cmd/setup_test.go
+++ b/cmd/uplink/cmd/setup_test.go
@@ -4,10 +4,15 @@
 package cmd_test
 
 import (
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"storj.io/storj/cmd/uplink/cmd"
 )
 
@@ -114,4 +119,60 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
 
 	assert.Equal(t, "ahost:7777", setupCmd.Flags().Lookup("satellite-addr").Value.String(),
 		"satellite-addr should contain default port when no port specified")
+}
+
+func TestSaveEncriptionKey(t *testing.T) {
+	expKey := make([]byte, rand.Intn(20)+1)
+	_, err := rand.Read(expKey)
+	require.NoError(t, err)
+
+	t.Run("ok", func(t *testing.T) {
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
+		require.NoError(t, err)
+		defer func() { _ = os.RemoveAll(dir) }()
+
+		fname := filepath.Join(dir, ".enc.key")
+		err = cmd.SaveEncryptionKey(expKey, fname)
+		require.NoError(t, err)
+
+		// Read the key from the file to compare that it matches with the saved one.
+		f, err := os.Open(fname)
+		require.NoError(t, err)
+		defer func() { _ = f.Close() }()
+
+		key := make([]byte, len(expKey))
+		_, err = f.Read(key)
+		require.NoError(t, err)
+		assert.Equal(t, expKey, key)
+
+		// Check that the key file has a read only file permissions for the user
+		fi, err := f.Stat()
+		require.NoError(t, err)
+		assert.Equal(t, os.FileMode(0400), fi.Mode().Perm())
+	})
+
+	t.Run("error: unexisting dir", func(t *testing.T) {
+		// Create a directory and remove it for making sure that the path doesn't
+		// exist
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
+		require.NoError(t, err)
+		err = os.RemoveAll(dir)
+		require.NoError(t, err)
+
+		fname := filepath.Join(dir, "enc.key")
+		err = cmd.SaveEncryptionKey(expKey, fname)
+		assert.Errorf(t, err, "directory path doesn't exist")
+	})
+
+	t.Run("error: file already exists", func(t *testing.T) {
+		// Create an empty file
+		f, err := ioutil.TempFile("", "storj-test-cmd-uplink-key-*")
+		require.NoError(t, err)
+		err = f.Close()
+		require.NoError(t, err)
+		defer func() { _ = os.Remove(f.Name()) }()
+
+		err = cmd.SaveEncryptionKey(expKey, f.Name())
+		assert.Errorf(t, err, "file key already exists")
+	})
 }
--- a/uplink/config.go
+++ b/uplink/config.go
@@ -75,6 +75,7 @@ func (kfp keyFilepath) Key() (storj.Key, error) {
 // EncryptionConfig is a configuration struct that keeps details about
 // encrypting segments
 type EncryptionConfig struct {
+	Key         string
 	KeyFilepath string      `help:"the path to the file which contains the root key for encrypting the data"`
 	BlockSize   memory.Size `help:"size (in bytes) of encrypted blocks" default:"1KiB"`
 	DataType    int         `help:"Type of encryption to use for content and metadata (1=AES-GCM, 2=SecretBox)" default:"1"`
@@ -158,6 +159,8 @@ func (c Config) GetMetainfo(ctx context.Context, identity *identity.FullIdentity
 			return nil, nil, err
 		}
 	}
+
+	c.Enc.Key = string(key[:])
 
 	streams, err := streams.NewStreamStore(segments, c.Client.SegmentSize.Int64(), &key, c.Enc.BlockSize.Int(), storj.Cipher(c.Enc.DataType))
 	if err != nil {
--- a/uplink/config.go
+++ b/uplink/config.go
@@ -6,6 +6,9 @@ package uplink
 import (
 	"context"
 	"errors"
+	"fmt"
+	"io"
+	"os"
 
 	"github.com/vivint/infectious"
 	"github.com/zeebo/errs"
@@ -37,13 +40,45 @@ type RSConfig struct {
 	MaxThreshold     int         `help:"the largest amount of pieces to encode to. n." default:"95" devDefault:"10"`
 }
 
+// keyFilepath is a convenient type to load a key stored in a file.
+type keyFilepath string
+
+func (kfp keyFilepath) Key() (storj.Key, error) {
+	f, err := os.Open(string(kfp))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return storj.Key{}, Error.Wrap(fmt.Errorf("not found key file %q", kfp))
+		}
+
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	if p := fi.Mode().Perm(); p != os.FileMode(0400) {
+		return storj.Key{}, Error.Wrap(
+			fmt.Errorf("permissions '%#o' for key file %q are too open", p, kfp),
+		)
+	}
+
+	var key storj.Key
+	if _, err := f.Read(key[:]); err != nil && err != io.EOF {
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	return key, nil
+}
+
 // EncryptionConfig is a configuration struct that keeps details about
 // encrypting segments
 type EncryptionConfig struct {
-	Key       string      `help:"root key for encrypting the data"`
-	BlockSize memory.Size `help:"size (in bytes) of encrypted blocks" default:"1KiB"`
-	DataType  int         `help:"Type of encryption to use for content and metadata (1=AES-GCM, 2=SecretBox)" default:"1"`
-	PathType  int         `help:"Type of encryption to use for paths (0=Unencrypted, 1=AES-GCM, 2=SecretBox)" default:"1"`
+	KeyFilepath string      `help:"the path to the file which contains the root key for encrypting the data"`
+	BlockSize   memory.Size `help:"size (in bytes) of encrypted blocks" default:"1KiB"`
+	DataType    int         `help:"Type of encryption to use for content and metadata (1=AES-GCM, 2=SecretBox)" default:"1"`
+	PathType    int         `help:"Type of encryption to use for paths (0=Unencrypted, 1=AES-GCM, 2=SecretBox)" default:"1"`
 }
 
 // ClientConfig is a configuration struct for the uplink that controls how
@@ -113,17 +148,25 @@ func (c Config) GetMetainfo(ctx context.Context, identity *identity.FullIdentity
 		return nil, nil, err
 	}
 
-	key := new(storj.Key)
-	copy(key[:], c.Enc.Key)
+	var key storj.Key
+	{
+		kfp := keyFilepath(c.Enc.KeyFilepath)
 
-	streams, err := streams.NewStreamStore(segments, c.Client.SegmentSize.Int64(), key, c.Enc.BlockSize.Int(), storj.Cipher(c.Enc.DataType))
+		var err error
+		key, err = kfp.Key()
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	streams, err := streams.NewStreamStore(segments, c.Client.SegmentSize.Int64(), &key, c.Enc.BlockSize.Int(), storj.Cipher(c.Enc.DataType))
 	if err != nil {
 		return nil, nil, Error.New("failed to create stream store: %v", err)
 	}
 
 	buckets := buckets.NewStore(streams)
 
-	return kvmetainfo.New(metainfo, buckets, streams, segments, key, c.Enc.BlockSize.Int32(), rs, c.Client.SegmentSize.Int64()), streams, nil
+	return kvmetainfo.New(metainfo, buckets, streams, segments, &key, c.Enc.BlockSize.Int32(), rs, c.Client.SegmentSize.Int64()), streams, nil
 }
 
 // GetRedundancyScheme returns the configured redundancy scheme for new uploads
--- a/uplink/config_test.go
+++ b/uplink/config_test.go
@@ -1,0 +1,106 @@
+package uplink
+
+import (
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"storj.io/storj/pkg/storj"
+)
+
+func TestKeyFilepath_Key(t *testing.T) {
+	saveKey := func(key []byte) (filepath string, removeFile func()) {
+		f, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
+		require.NoError(t, err)
+		defer func() { _ = f.Close() }()
+
+		_, err = f.Write(key)
+		require.NoError(t, err)
+
+		err = f.Chmod(os.FileMode(0400))
+		require.NoError(t, err)
+
+		return f.Name(), func() { _ = os.Remove(f.Name()) }
+	}
+
+	t.Run("ok: file with key length less or equal than max size", func(t *testing.T) {
+		someKey := make([]byte, rand.Intn(20)+1)
+		_, err := rand.Read(someKey)
+		require.NoError(t, err)
+		fpath, cleanup := saveKey(someKey)
+		defer cleanup()
+
+		var expKey storj.Key
+		copy(expKey[:], someKey)
+
+		kfpath := keyFilepath(fpath)
+		key, err := kfpath.Key()
+		require.NoError(t, err)
+
+		assert.Equal(t, expKey[:], key[:])
+	})
+
+	t.Run("ok: file with key length greater than max size", func(t *testing.T) {
+		expKey := make([]byte, rand.Intn(10)+1+storj.KeySize)
+		_, err := rand.Read(expKey)
+		fpath, cleanup := saveKey(expKey)
+		defer cleanup()
+
+		kfpath := keyFilepath(fpath)
+		key, err := kfpath.Key()
+		require.NoError(t, err)
+
+		assert.Equal(t, expKey[:storj.KeySize], key[:])
+	})
+
+	t.Run("ok: empty file path", func(t *testing.T) {
+		fpath, cleanup := saveKey([]byte{})
+		defer cleanup()
+
+		kfpath := keyFilepath(fpath)
+		key, err := kfpath.Key()
+		require.NoError(t, err)
+		assert.Equal(t, key, storj.Key{})
+	})
+
+	t.Run("error: file not found", func(t *testing.T) {
+		// Create a temp file and delete it, to get a filepath which doesn't exist.
+		f, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
+		require.NoError(t, err)
+		err = f.Close()
+		require.NoError(t, err)
+		err = os.Remove(f.Name())
+		require.NoError(t, err)
+
+		kfpath := keyFilepath(f.Name())
+		_, err = kfpath.Key()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), fmt.Sprintf("not found key file %q", f.Name()))
+		assert.True(t, Error.Has(err), "err is not of %q class", Error)
+	})
+
+	t.Run("error: permissions are too open", func(t *testing.T) {
+		// Create a key file and change its permission for not being able to read it
+		f, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
+		require.NoError(t, err)
+		defer func() {
+			_ = f.Close()
+			_ = os.Remove(f.Name())
+		}()
+
+		err = f.Chmod(0401)
+		require.NoError(t, err)
+
+		kfpath := keyFilepath(f.Name())
+		_, err = kfpath.Key()
+		require.Error(t, err)
+		assert.Contains(t,
+			err.Error(), fmt.Sprintf("permissions '0401' for key file %q are too open", f.Name()),
+		)
+		assert.True(t, Error.Has(err), "err is not of %q class", Error)
+	})
+}
--- a/pkg/process/exec_conf.go
+++ b/pkg/process/exec_conf.go
@@ -27,6 +27,9 @@ import (
 	"storj.io/storj/internal/version"
 )
 
+// DefaultConfFilename is the default filename used for storing a configuration.
+const DefaultConfFilename = "config.yaml"
+
 var (
 	mon = monkit.Package()
 
@@ -169,7 +172,7 @@ func cleanup(cmd *cobra.Command) {
 
 		cfgFlag := cmd.Flags().Lookup("config-dir")
 		if cfgFlag != nil && cfgFlag.Value.String() != "" {
-			path := filepath.Join(os.ExpandEnv(cfgFlag.Value.String()), "config.yaml")
+			path := filepath.Join(os.ExpandEnv(cfgFlag.Value.String()), DefaultConfFilename)
 			if cmd.Annotations["type"] != "setup" || fileExists(path) {
 				vip.SetConfigFile(path)
 				err = vip.ReadInConfig()


uplink/config.go
```diff
@@ -40,10 +43,62 @@ type RSConfig struct {
 // EncryptionConfig is a configuration struct that keeps details about
 // encrypting segments
 type EncryptionConfig struct {
-	Key       string      `help:"root key for encrypting the data"`
-	BlockSize memory.Size `help:"size (in bytes) of encrypted blocks" default:"1KiB"`
-	DataType  int         `help:"Type of encryption to use for content and metadata (1=AES-GCM, 2=SecretBox)" default:"1"`
-	PathType  int         `help:"Type of encryption to use for paths (0=Unencrypted, 1=AES-GCM, 2=SecretBox)" default:"1"`
+	// TODO: WIP#if/v3-1541 read TODO in Key method why the following field cannot
```
ifraixedes: This comment must go away before merging, it's here for allowing you to know the problems that I found and get feedback in having some idea in how to do so, without breaking other packages that much, otherwise we could be tracked for a future enhancement.

uplink/config.go
```diff
@@ -40,10 +43,62 @@ type RSConfig struct {
...
+// Key returns the root key for encrypting the data which is stored in the path
+// indicated by KeyFilepath field.
+//
+// It returns an error if:
+//
+// * The file doesn't exist.
+// * The file doesn't have exactly 0400 permissions.
+// * There is an I/O error.
+//
+// TODO: WIP#if/v3-1541 refactor cache the key after the firs read. The problem
```
ifraixedes: This comment must go away before merging; it's here for the same reasons than my previous comment.

ifraixedes:
Before merging, there is at least one commit to squash (its subject starts with `FIXUP` and its comment body is saying so), but I would also like to squash the @mobyvb commits if he's OK with it.

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+// several different types as a type of storj.Key, [32]byte, etc.
+func (encCfg *EncryptionConfig) Key() (storj.Key, error) {
+	if encCfg.KeyFilepath == "" {
+		return storj.Key{}, nil
+	}
+
+	f, err := os.Open(encCfg.KeyFilepath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return storj.Key{}, Error.Wrap(fmt.Errorf("not found key file %q", encCfg.KeyFilepath))
```
mniewrzal: you can use just `Error.New("not found key file %q", encCfg.KeyFilepath)`

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+		}
+
+		return storj.Key{}, errs.Wrap(err)
```
mniewrzal: `Erros.Wrap`?

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	if p := fi.Mode().Perm(); p != os.FileMode(0400) {
```
mniewrzal: `p` -> `perm`

uplink/config_test.go
```diff
@@ -0,0 +1,120 @@
...
+		assert.True(t, Error.Has(err), "err is not of %q class", Error)
+	})
+
+	t.Run("error: permissions are too open", func(t *testing.T) {
+		// Create a key file and change its permission for not being able to read it
+		f, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
+		require.NoError(t, err)
+		defer func() {
+			_ = f.Close()
+			_ = os.Remove(f.Name())
```
mniewrzal: errors shouldn't be ignored

uplink/config_test.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		_, err = encCfg.Key()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), fmt.Sprintf("not found key file %q", f.Name()))
+		assert.True(t, Error.Has(err), "err is not of %q class", Error)
+	})
+
+	t.Run("error: permissions are too open", func(t *testing.T) {
+		// Create a key file and change its permission for not being able to read it
+		f, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
```
mniewrzal: `file`

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		defer func() { _ = os.RemoveAll(dir) }()
+
+		fname := filepath.Join(dir, ".enc.key")
+		err = cmd.SaveEncryptionKey(expKey, fname)
+		require.NoError(t, err)
+
+		// Read the key from the file to compare that it matches with the saved one.
+		f, err := os.Open(fname)
+		require.NoError(t, err)
+		defer func() { _ = f.Close() }()
```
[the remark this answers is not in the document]
ifraixedes: It's a test, I don't think that report an error for this case is relevant for the test.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+
+		key := make([]byte, len(expKey))
+		_, err = f.Read(key)
+		require.NoError(t, err)
+		assert.Equal(t, expKey, key)
```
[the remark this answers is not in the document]
ifraixedes: It's intentionally, the key isn't used from this point on and I see that's good to continue to verify the next assertions.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		assert.Equal(t, os.FileMode(0400), fi.Mode().Perm())
+	})
+
+	t.Run("error: unexisting dir", func(t *testing.T) {
+		// Create a directory and remove it for making sure that the path doesn't
+		// exist
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
+		require.NoError(t, err)
+		err = os.RemoveAll(dir)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
ifraixedes: to make sure that the path doesn't exist, I could use some random but if it has to work on different operative system I found easier and clearer to do so.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+
+		fname := filepath.Join(dir, "enc.key")
+		err = cmd.SaveEncryptionKey(expKey, fname)
+		assert.Errorf(t, err, "directory path doesn't exist")
+	})
+
+	t.Run("error: file already exists", func(t *testing.T) {
+		// Create an empty file
+		f, err := ioutil.TempFile("", "storj-test-cmd-uplink-key-*")
```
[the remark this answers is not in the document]
ifraixedes: I'm not sure if I get your comment, can you slightly elaborate it.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		require.NoError(t, err)
+		err = f.Close()
+		require.NoError(t, err)
+		defer func() { _ = os.Remove(f.Name()) }()
+
+		err = cmd.SaveEncryptionKey(expKey, f.Name())
+		assert.Errorf(t, err, "file key already exists")
```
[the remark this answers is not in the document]
ifraixedes: I don't think that matter that much, it's the last assertion of the function, however I can use `require` if you prefer so.

uplink/config_test.go
ifraixedes: Same reason than before, it's a test, these errors don't provide anything useful to report to the test if they are returned nor stop the test to run

cmd/uplink/cmd/setup.go
```diff
@@ -209,3 +218,37 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
...
+		}
+
+		return err
+	}
+
+	defer func() {
+		if err == nil {
+			err = f.Close()
+		} else {
+			_ = f.Close()
```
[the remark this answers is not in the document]
ifraixedes: Because the error returned is relevant over the the one that `Close` may return.

I didn't bear in mind `errs.Combine`, having it solve the issue because you don't miss the important one, but I'm still not sure if when checking such error in logs, I would like to see the error in `Close` when the `Write` happened, because in my opinion will be more noise than something to help

uplink/config_test.go
mniewrzal: its more a convention that we are handling all errors, even in tests to be sure we are not leaving junk or unclosed things

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
mniewrzal: sounds good

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
mniewrzal: we are trying to avoid single character variables and name them in a descriptive way

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+	expKey := make([]byte, rand.Intn(20)+1)
+	_, err := rand.Read(expKey)
+	require.NoError(t, err)
+
+	t.Run("ok", func(t *testing.T) {
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
+		require.NoError(t, err)
+		defer func() { _ = os.RemoveAll(dir) }()
+
+		fname := filepath.Join(dir, ".enc.key")
```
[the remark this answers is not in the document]
ifraixedes: which is the convention for short var names?, I'm struggling to understanding it because yes, this one could be `fileName` or `filename` but then why `f` shouldn't be `file`?

I see a difference for the `expKey` reported above, because it's used across all the subtest functions and that's a longer scope, but these 2 are in the same scope and they aren't far away from their definition and usage.

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
mniewrzal: in this case will work as expected but someone who don't know how `assert.Error` work can do copy/paste and make own test results unclear. Its more about using `assert.Error` only in places where its explicitly needed.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		// Create a directory and remove it for making sure that the path doesn't
+		// exist
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
+		require.NoError(t, err)
+		err = os.RemoveAll(dir)
+		require.NoError(t, err)
+
+		fname := filepath.Join(dir, "enc.key")
+		err = cmd.SaveEncryptionKey(expKey, fname)
+		assert.Errorf(t, err, "directory path doesn't exist")
```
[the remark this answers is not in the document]
ifraixedes: Same comment than https://github.com/storj/storj/pull/1782#discussion_r278075012

uplink/config.go
ifraixedes: I doubted it if it should use for errors not directly related to `Uplink configuration`.

Can you confirm that all the errors on this package should be of such class?
My understanding from this comment is yes, because it's mostly a way to prefix the error message, but I'm not entirely sure if that's correct.

uplink/config_test.go
ifraixedes: I commented before, about the var names conventions, but if you tell me that all the `f` variables which are of the type `file` should be `file`, and you only commented here, then it make more sense to do the rename here.

uplink/config.go
ifraixedes: Waiting for the var names convention, but for this one it's a very limited scope (one line) that you can see clear what `p` is

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
mniewrzal: sorry, most probably I just missed `f`. Most of one char vars should be more descriptive.

Yeah its simple and small function but cost of renaming this is almost zero and it will help other devs to read your code. For you it's obvious. Also code can be reorganized later and scope can be larger. Its about keeping the same standards everywhere.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +120,59 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
 		"satellite-addr should contain default port when no port specified")
 }
+
+func TestSaveEncriptionKey(t *testing.T) {
+	expKey := make([]byte, rand.Intn(20)+1)
+	_, err := rand.Read(expKey)
+	require.NoError(t, err)
+
+	t.Run("ok", func(t *testing.T) {
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
```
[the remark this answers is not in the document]
ifraixedes: I wasn't aware of such package for doing such things, I will use it instead of.
Taking a look what it has, there should be more places to use it, so I'll check them too.

uplink/config.go
mniewrzal: Question is why you want to keep it `p`? :) Is it more readable or cleaner?

uplink/config.go
mniewrzal: ok, I'm also not fully sure if this should be wrapped with uplink error

cmd/uplink/cmd/setup.go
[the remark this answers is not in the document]
kaloyan-raev: Is it possible that `f` is `nil` at this point and `f.Close()` could panic?

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
kaloyan-raev: We check all errors in tests too. So here you are supposed to have something like:
```go
defer func() { require.NoError(t, f.Close()) }()
```

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
ifraixedes: > Yeah its simple and small function but cost of renaming this is almost zero 

Not replying for not doing, because of the cost to rename, because as you said, it's miserable, I wanted to understand what it's expected, to be align.

> and it will help other devs to read your code

That's subjective, because according other group of people, short variable names are clear for not long scopes, because names could confuse more than forcing the reader to find out exactly what the var contains rather than inciting them to assume it.

> Also code can be reorganized later and scope can be larger

I'm not a big fun on doing things for the sake of being a future unknown/unplanned refactoring.


> its about keeping the same standards everywhere

This the good point for doing it, hence, I'm going to review all the changes to have longer names.

---

A part of that, I would like to know the restrictions of having explicit names.

I suffered to have to long names for exported types, functions, etc., and in my opinion they were more a problem than a help, because the lines started to be too long and reading certain parts of code where there were too many of them was making harder to understand the code despite of the descriptive way.


For example when it's fine to use abbreviations and when not? I found places where `cfg` used as `configuration`, however in this review `expKey` has been renamed to `expectedKey` so which abbreviations are good and which ones aren't?

uplink/config.go
ifraixedes: Ok, then let's ask to others

Do @kaloyan-raev @mobyvb know more how this is applied?

uplink/config.go
ifraixedes: > Question is why you want to keep it p

make the lines shorter, it's clear to realize that `p` is the result of `Perm` and make more clear to realize that it's just a var initialized in this due to `:=`

I've never used that many short variables until I started to develop in Go, at the beginning I didn't get it why of such convention, however I got used to and then I've realized that the code was as readable as code with longer names, so it's a matter of what people is use to.

Obviously, they have to use with measure, the longer the scope is the more explicit names are needed, however I also found impossible to have explicit and clear names, which are not needed to go back and see where they are coming, for variables involved in several lines of code.

>  Is it more readable or cleaner?

For the reason commented above and the fact that I found very long lines less readable (i.e. of one instructions, although you break it down in several lines) despite of the use of descriptive names.


---

At the end what I want is not what I wrote, it's to understand some conventions that they are not subjective, to have the next changes aligned on them for having to make less change due to reviews and avoid subjective discussions.

Bear in mind, this is my first more than a few lines PR, and I read the Storj conventions, however I didn't find this topic clear, so I go ahead on what I thought that's clear (not reading my code, reading go code during the last years) to actually have this discussion if somebody mentioned on them and clarify it for understanding what to follow on the future changes.

jtolds: for what it's worth, i think ivan's right that in general, go code prefers brevity (see https://talks.golang.org/2014/names.slide and https://www.reddit.com/r/golang/comments/8wxwgv/why_does_go_encourage_short_variable_names/). we have a preference for slightly longer names in cases where there might be ambiguity (https://github.com/storj/docs/blob/master/code/Style.md#variable-naming) buuut fwiw i think the p choice here is fine. it's limited to just a couple lines of scope

uplink/config_test.go
ifraixedes: I understand what you mean, because it's what I would like but in this cases my questions are:

1. Would you like that the CI stop because one temporary file created by the test isn't deleted or closing its file descriptor has failed?
2. If the test in the CI fails because the temporary file, created by the test, has not been deleted, do you think that we are going to take care to go and delete such temporary file manually? If I consider that the CI spins up a new container on each run (I'm not sure about this, it's a guess) and the disk is ephemeral, then I'm sure that no need, as it won't remain.

If the convention is, we don't want to think in if it makes sense to skip the error or not, then I'm fine with that, only note that if I wrote `_` is because I was aware of the errors that those functions can return otherwise I would have written `f.Close()` and `os.Remove(..)`.

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
ifraixedes: I knew that I could do so, but the reasons are the same that the ones commented in https://github.com/storj/storj/pull/1782#discussion_r278218338

I wait to reply on it, because it's the same case and with one thread is enough 😅

cmd/uplink/cmd/setup.go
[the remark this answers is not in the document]
ifraixedes: it cannot because if `os.Open` returns a `nil` error, then `f` cannot be `nil`

uplink/config.go
ifraixedes: 😮 I read that Storj conventions part before I've joined, however I didn't remember that it was a reference to an exact (soft limit) number of lines.

ifraixedes: btw, thanks @jtolds  for the links, I didn't post them, because I though that they were known and mostly I thought that it was another convention and my goal was trying to know which one for bearing in mind for the future.

uplink/config_test.go
kaloyan-raev: Yes, we want the CI tests to fail if any of the hypothesis you've described happen. Every time an unexpected error happens we need to know about it and investigate it. This is why we check all errors.

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
ifraixedes: 😆 my confusion was with the `keyFile`, I thought that you were referring to the name of the file rather than the var

uplink/config.go
mniewrzal: From the beginning I should mark this as a nit. I know that Go prefers brevity and acronyms but most probably it will take a long time for me to use to it. Even if var is declared very close to a place where its used, for me its always additional conversion in my head ;)

Sorry for additional buzz :)

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
mniewrzal: oh, sorry about that :)

uplink/config_test.go
ifraixedes: The important thing on those hypothesis are **"one temporary file created by the test"** if that wouldn't come from the test there wouldn't be any doubt.

However, the true is that if the test has a bug in terms of calling those functions at the wrong moment, then it could be an error but ignoring them, we will never know about such bug, obviously at the expense of the CI stoping because of an unlikely error returned by the OS, which mostly could warn about a problem with the CI machine although we would probably have other problems at that same moment too.

Adding those assertions, considering that one line (e.g. `require.NoError(t, f.Close())`) is readable/clear/understandable.

uplink/config.go
ifraixedes: Nothing to be sorry, it has clarified me a lot about Storj code conventions regarding this topic.

At the end, the language matters, if yo code in other language which another convention the most of the people expect to find the conventions that are more widely used.

ifraixedes:
I think that I haven't forgotten anything commented in the review and it has to be changed, except the several points regarding the variable names which was clarified in https://github.com/storj/storj/pull/1782#discussion_r278217103 and https://github.com/storj/storj/pull/1782#discussion_r278207932 which we are waiting for feedback.

cmd/uplink/cmd/setup.go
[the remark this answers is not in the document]
kaloyan-raev: I still think we should not ignore the error returned by Close. If Write or Chmod fails and Close returns an error too, this additional error may give us additional info for analyzing the problem.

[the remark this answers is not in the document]
egonelbre: Propagate all errors up the chain rather than hiding it. It might contain extra information on why write failed. General rule is, do not hide/ignore the error unless there's a good reason not to (e.g. something else fails as a result).

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		defer ctx.Cleanup()
+
+		fname := ctx.File("storj-test-cmd-uplink", ".enc.key")
+		err = cmd.SaveEncryptionKey(expectedKey, fname)
+		require.NoError(t, err)
+
+		// Read the key from the file to compare that it matches with the saved one.
+		f, err := os.Open(fname)
+		require.NoError(t, err)
+		defer func() { require.NoError(t, f.Close()) }()
```
egonelbre: defer ctx.Check(f.Close)

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		// Check that the key file has a read only file permissions for the user
+		fi, err := f.Stat()
+		require.NoError(t, err)
+		assert.Equal(t, os.FileMode(0400), fi.Mode().Perm())
+	})
+
+	t.Run("error: unexisting dir", func(t *testing.T) {
+		// Create a directory and remove it for making sure that the path doesn't
+		// exist
+		dir, err := ioutil.TempDir("", "storj-test-cmd-uplink")
```
egonelbre: Use testcontext for creating directories.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+	t.Run("ok", func(t *testing.T) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+
+		fname := ctx.File("storj-test-cmd-uplink", ".enc.key")
+		err = cmd.SaveEncryptionKey(expectedKey, fname)
+		require.NoError(t, err)
+
+		// Read the key from the file to compare that it matches with the saved one.
+		f, err := os.Open(fname)
```
egonelbre: ioutil.ReadFile

uplink/config.go
egonelbre: I would definitely prefer longer variable names `perm`.

By shortening the line it's not increasing readability. The way to keep things more readable is to move the evaluation to a separate line instead.

```
perm := fi.Mode().Perm(); 
if perm != os.FileMode(0400) {
    return storj.Key{}, Error.New("permissions '%#o' for key file %q are too open", p, encCfg.KeyFilepath)
}
```

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+//
+// TODO: WIP#if/v3-1541 refactor cache the key after the firs read. The problem
+// is with pkg/cfgstruct which is inflexible on accepting unexported keys nor
+// several different types as a type of storj.Key, [32]byte, etc.
+func (encCfg *EncryptionConfig) Key() (storj.Key, error) {
+	if encCfg.KeyFilepath == "" {
+		return storj.Key{}, nil
+	}
+
+	f, err := os.Open(encCfg.KeyFilepath)
```
egonelbre: `ioutil.ReadFile`

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return storj.Key{}, errs.Wrap(err)
+	}
+
+	if p := fi.Mode().Perm(); p != os.FileMode(0400) {
```
egonelbre: Why does it need to check the permissions at all? This probably will completely fail on file systems that don't have permissions.

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
kaloyan-raev: > For example when it's fine to use abbreviations and when not? I found places where cfg used as configuration, however in this review expKey has been renamed to expectedKey so which abbreviations are good and which ones aren't?

I guess it is fine using abbreviations when they are common and unambiguous. For example, it's hard to misunderstand `cfg` - it is a widely used short name for "configuration". While `expKey` may not always be clear. Is it an "expected key", or an "expired key", or something else? `expKey` is not that common and may require some thinking what is it actually while reading it.

uplink/config.go
kaloyan-raev: I think we always prefer using `Error.Wrap(err)`. Prefixing the error makes it easier to spot the context in which the error happened.

Also, I think that failing to open the file with the encryption key is an error that is very well related to the uplink configuration.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		require.NoError(t, err)
+
+		filename := filepath.Join(dir, "enc.key")
+		err = cmd.SaveEncryptionKey(expectedKey, filename)
+		require.Errorf(t, err, "directory path doesn't exist")
+	})
+
+	t.Run("error: file already exists", func(t *testing.T) {
+		// Create an empty file
+		file, err := ioutil.TempFile("", "storj-test-cmd-uplink-key-*")
```
kaloyan-raev: Use `testcontext.File` here too.

uplink/config_test.go
egonelbre: `os.Remove` gives very useful information on Windows. When it fails then there's something is still using the file, which means we have a leak in our system.

Or in other words, we have fixed production leaks by starting to check these errors.

cmd/uplink/cmd/setup.go
```diff
@@ -69,7 +70,13 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
-	var override map[string]interface{}
+	var (
+		encKeyFilepath = filepath.Join(setupDir, ".enc.key")
```
egonelbre: Use `.encryption.key` or similar. Users probably don't know what `enc` means. (I would also change flags, but we don't have aliases in place for that).

cmd/uplink/cmd/setup.go
```diff
@@ -251,3 +260,37 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
...
+
+// SaveEncryptionKey saves the key in a new file which will be stored in
+// filepath.
+// It returns an error if the directory doesn't exist, the file already exists
+// or there is an I/O error.
+func SaveEncryptionKey(key []byte, filepath string) (err error) {
+	f, err := os.OpenFile(filepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errors.New("directory path doesn't exist")
```
egonelbre: Is this the only case when this can happen? If you know for certain that on all platforms (Windows, Linux, Android, Fuchsia etc.), then this is fine, however when not, then include the original error.

cmd/uplink/cmd/setup.go
```diff
@@ -251,3 +260,37 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
...
+		}
+
+		if os.IsExist(err) {
+			return errors.New("file key already exists")
+		}
+
+		return err
+	}
+
+	defer func() {
```
egonelbre: `defer func () { err = errs.Combine(err, f.Close()) }()`

cmd/uplink/cmd/setup.go
```diff
@@ -251,3 +260,37 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 	// address is host:
 	return net.JoinHostPort(addressParts[0], defaultPort), nil
 }
+
+// SaveEncryptionKey saves the key in a new file which will be stored in
+// filepath.
+// It returns an error if the directory doesn't exist, the file already exists
+// or there is an I/O error.
+func SaveEncryptionKey(key []byte, filepath string) (err error) {
+	f, err := os.OpenFile(filepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
```
egonelbre: `ioutil.WriteFile`

cmd/uplink/cmd/setup.go
```diff
@@ -251,3 +260,37 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 	// address is host:
 	return net.JoinHostPort(addressParts[0], defaultPort), nil
 }
+
+// SaveEncryptionKey saves the key in a new file which will be stored in
+// filepath.
+// It returns an error if the directory doesn't exist, the file already exists
+// or there is an I/O error.
+func SaveEncryptionKey(key []byte, filepath string) (err error) {
```
egonelbre: If it's saving an encryption key, then take the key as the argument rather than a byte slice.

uplink/config_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"storj.io/storj/pkg/storj"
+)
+
+func TestEncryptionConfig_Key(t *testing.T) {
+	saveKey := func(key []byte) (filepath string, removeFile func()) {
+		t.Helper()
+
+		file, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
```
kaloyan-raev: Use `testcontext` to create the temp files in this test too. There are several places in this file.

uplink/config_test.go
```diff
@@ -0,0 +1,123 @@
...
+			KeyFilepath: filename,
+		}
+		key, err := encCfg.Key()
+		require.NoError(t, err)
+		assert.Equal(t, key, storj.Key{})
+	})
+
+	t.Run("error: file not found", func(t *testing.T) {
+		// Create a temp file and delete it, to get a filepath which doesn't exist.
+		file, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
```
egonelbre: Why bother creating a temp file when you are going to immediately delete it?

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+// It returns an error if:
+//
+// * The file doesn't exist.
+// * The file doesn't have exactly 0400 permissions.
+// * There is an I/O error.
+//
+// TODO: WIP#if/v3-1541 refactor cache the key after the firs read. The problem
+// is with pkg/cfgstruct which is inflexible on accepting unexported keys nor
+// several different types as a type of storj.Key, [32]byte, etc.
+func (encCfg *EncryptionConfig) Key() (storj.Key, error) {
```
egonelbre: Use `LoadKey` or `ReadKey` to make clear that it's going to read from the disk.

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+	if encCfg.KeyFilepath == "" {
+		return storj.Key{}, nil
```
egonelbre: This seems really dangerous, return a specific error `ErrKeyNotSpecified` or similar.

uplink/config.go
```diff
@@ -40,10 +43,58 @@ type RSConfig struct {
...
+	}
+
+	f, err := os.Open(encCfg.KeyFilepath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return storj.Key{}, Error.New("not found key file %q", encCfg.KeyFilepath)
```
egonelbre: `"did not find key file %q: %v", encCfg.KeyFilepath, err)`

uplink/config_test.go
ifraixedes: I wanted to use, but it doesn't create the file, as far I understood.

egonelbre: `return ioutil.WriteFile(..., ctx.File(".encryption.key", ...)`

cmd/uplink/cmd/setup.go
ifraixedes: Obviously I don't know for 100%

uplink/config_test.go
ifraixedes: It's in the comment above.
And also I replied on https://github.com/storj/storj/pull/1782#discussion_r278073976

uplink/config_test.go
```diff
@@ -0,0 +1,123 @@
...
+		require.NoError(t, err)
+		err = file.Close()
+		require.NoError(t, err)
+		err = os.Remove(file.Name())
+		require.NoError(t, err)
```
kaloyan-raev: Do we really need to create and delete a temp file here? Can't we just pass a non-existing file path to `EncryptionConfig`?

uplink/config.go
ifraixedes: My understanding is that currently it was possible to have an empty key and the configuration file contained `enc.key: ""` and this try to have the same behavior.

Not in my likes of not forcing to have a key being hones.

uplink/config_test.go
```diff
@@ -0,0 +1,123 @@
...
+	t.Run("error: permissions are too open", func(t *testing.T) {
+		// Create a key file and change its permission for not being able to read it
+		file, err := ioutil.TempFile("", "storj-test-uplink-keyfilepath-*")
+		require.NoError(t, err)
+		defer func() {
+			require.NoError(t, file.Close())
+			require.NoError(t, os.Remove(file.Name()))
+		}()
+
+		err = file.Chmod(0401)
```
kaloyan-raev: I am not sure if this will work on Windows as expected: https://golang.org/pkg/os/#Chmod

I expect this test to fail on Windows.

kaloyan-raev: Maybe setting `0600` instead of `0401` is a better test case that should work on Windows too.

uplink/config.go
ifraixedes: It was call `key` because I wanted to read it just once, but I didn't know where to host the key after the first read, and then I forgot to reconsider the name.

uplink/config.go
egonelbre: The issue I can see is that someone will use:

```
key, err := enc.Key()
if err == nil { return err }

// start using the key as if it were valid
```

cmd/uplink/cmd/setup_test.go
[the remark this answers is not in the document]
ifraixedes: you have to consider the context, it's a test, so you expect things, but I understand the point.

`cfg` I don't know if that's a common abbreviation, I used more `config` than `cfg`, but anyway, both are clear to me.

cmd/uplink/cmd/setup.go
egonelbre: Hence include the actual error, I've seen things fail in weird ways. Conceptually, I understand trying to minimize the noise, but we are still in beta and going to hit weird edge cases. So I prefer to have more information rather than cleaner output.

uplink/config.go
ifraixedes: It looks more a personal opinion.
On the other hand I agree of breaking the lines, however when to do so or not, foe example it looks a common case in the code baes to use

`defer func() { require.NoError(t, f.Close() }()` in one line.

I can perfectly read and understand it, however I don't think that it's very readable.

uplink/config_test.go
ifraixedes: > Or in other words, we have fixed production leaks by starting to check these errors.

It's a test, so this error wont' happen in production, otherwise I won't have arguments.

Anyway, after know that we are OK on stopping the CI for these matters, then it looks good to assert them.

uplink/config_test.go
ifraixedes: This is done!

uplink/config_test.go
egonelbre: I mean a test detecting a leak that was in production. It's not about leaking the specific file, but an open file descriptor to that file in an unrelated place. For example, let's say you forget to close the file in `ReadKey` then `os.Remove` will fail on Windows.

uplink/config.go
egonelbre: that code can be `defer ctx.Check(f.Close)`. I agree that the previous is not very readable.

cmd/uplink/cmd/setup_test.go
ifraixedes: The file is needed for the following check.

uplink/config.go
ifraixedes: The file is used for checking the permissions, if we remove that part, then yes, we can replace it.

uplink/config.go
ifraixedes: I don't know how this behaves in those platforms.
I was asking in the issue for the conventions, but I didn't get an answer, so I wanted to mimmic what `ssh` requires.

Do we remove it?

cmd/uplink/cmd/setup_test.go
ifraixedes: `testcontext.File` doesn't create the file but it doesn't confirm that the file doesn't exist either.

uplink/config.go
egonelbre: I would remove it. I can easily see people making a portable setup and copying their key to a FAT32 usb drive.

cmd/uplink/cmd/setup_test.go
egonelbre: `testcontext.File` ensures that it will return the path inside an empty temporary directory. So, unless you have written the file, it won't exist, unless in previous tests you write one.

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
 	assert.Equal(t, "ahost:7777", setupCmd.Flags().Lookup("satellite-addr").Value.String(),
 		"satellite-addr should contain default port when no port specified")
 }
+
+func TestSaveEncriptionKey(t *testing.T) {
```
egonelbre: typo `TestSaveEncryptionKey`

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+	t.Run("ok", func(t *testing.T) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+
+		filename := ctx.File("storj-test-cmd-uplink", ".enc.key")
+		err = cmd.SaveEncryptionKey(expectedKey, filename)
+		require.NoError(t, err)
+
+		// Read the key from the file to compare that it matches with the saved one.
+		file, err := os.Open(filename)
```
egonelbre: `ioutil.WriteFile` and in the lower part use `os.Stat`

cmd/uplink/cmd/setup_test.go
```diff
@@ -115,3 +121,58 @@ func TestDefaultPortAppliedToSatelliteAddrWithPortColonButNoPort(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, expectedKey, key)
+
+		// Check that the key file has a read only file permissions for the user
+		fileInfo, err := file.Stat()
+		require.NoError(t, err)
+		assert.Equal(t, os.FileMode(0400), fileInfo.Mode().Perm())
+	})
+
+	t.Run("error: unexisting dir", func(t *testing.T) {
```
egonelbre: `t.Run("missing dir", func(t *testing.T) {`

pkg/miniogw/integration_test.go
```diff
@@ -142,7 +144,7 @@ func TestUploadDownload(t *testing.T) {
 }
 
 // runGateway creates and starts a gateway
-func runGateway(ctx context.Context, gwCfg config, uplinkCfg uplink.Config, log *zap.Logger, ident *identity.FullIdentity) (err error) {
+func runGateway(ctx context.Context, gwCfg config, uplinkCfg uplink.Config, encKey *storj.Key, log *zap.Logger, ident *identity.FullIdentity) (err error) {
```
egonelbre: Is there a reason the encryption key isn't part of the config? To me it seems it should be part of the config rather than a separate argument.

egonelbre: Although it might be that we are conflating two different things:

1. the command line arguments to uplink
2. the configuration of the uplink

cmd/uplink/cmd/setup.go
ifraixedes: > Use .encryption.key or similar. Users probably don't know what enc means.

Before changing it, could we agree on this please, the name that I used is just to use one, actually I doubted if we wanted to use the leading `.` because that's not a hidden file in all the platforms.

> I would also change flags, but we don't have aliases in place for that

I don't understand what you're saying.

egonelbre: The uplink flags still use `--enc.key` etc. but we can leave that for now.

cmd/uplink/cmd/setup.go
ifraixedes: I was going to do without further explanation, but I appreciate your message because it gives more context on the current stage.

ifraixedes: By the way, should we use `errs.Wrap in these errors?

cmd/uplink/cmd/setup.go
ifraixedes: is it one line readable enough?

egonelbre: It's not really readable, however it's a very common thing we do... so we can easily recognize this part by learning it once. You can split into multiple lines as well.

wthorp: Bill also prefers fewer lines.  😄

cmd/uplink/cmd/setup.go
wthorp: Egon is saying use full full work "encryption" not "enc".  Dot files are a little bit of a hassle in Windows under certain circumstances, but probably not enough to actively avoid them.

cmd/uplink/cmd/setup.go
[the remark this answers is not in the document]
ifraixedes: The example doesn't help that much, but don't worry, I see the point considering that we have `errs.Combine`, I wasn't sure until that point we want to add more and more.

uplink/config_test.go
ifraixedes: That's a good example of the bug case that I mentioned above.

uplink/config_test.go
ifraixedes: Not just in windows, it also fails in FAT 32, I've tested it.

uplink/config.go
ifraixedes: I totally agree, but I guess that if we try to force a key there will be several things that can be broken.

For example I see that uplink setup ask for a passphrase, it can be empty so I left the keyfile path empty which means that there is not passphrase.

In order of not breaking the current behavior, I see that rather than doing this, we could do one of the following options:

* Return a specific error (e.g. `var NoKeyErr`) and the caller can identify it an assume the risk if it wants.
* Store a file with a zero key, so it's always expected to find a keyfile.

Do you have any other idea?

My understanding from the initial team green meeting of this sprint is to have an initial implementation and improve it from this point, so if the solution requires too many changes then it shouldn't classify on such agreement, nonetheless I don't have any issue on extending this work if we believe that's better for the sake of something like security.

cmd/uplink/cmd/setup.go
ifraixedes: > Egon is saying use full full work "encryption" not "enc"

I got that part

cmd/uplink/cmd/setup_test.go
ifraixedes: This one requires the file to exists before calling `SaveEncryptionKey`

cmd/uplink/cmd/setup.go
ifraixedes: > The uplink flags still use --enc.key etc. but we can leave that for now.

I understand that you meant the `enc.` should be `encryption`, if it's that, then yes it's in that way and I understood what you meant with the _alias_.

cmd/uplink/cmd/setup.go
ifraixedes: This kind of things are the bad things that could make any convention easy to remember and the harder are to remember the harder is to have contributions aligned with them.

cmd/uplink/cmd/setup.go
ifraixedes: I don't think that we want to truncate the file if already exists; currently the function return an error if the file already exists.

uplink/config.go
jtolds: when we filed the ticket the point was definitely to encourage people to keep their secrets private. would it acceptable @egonelbre to require permissions on linux and osx builds? i don't think people should be using their key off of a usb key directly.

jtolds: mimicking the behavior of ssh was the intention for sure

ifraixedes: Meanwhile I see that we could say that we don't support FAT32 because "who is still using that? 😄" the point is Windows.

Unfortunately, I don't have an accessible Windows machine for testing it but the [`os.Chmod`](https://golang.org/pkg/os/#Chmod) documentation looks quite clear about it, something that I missed and @kaloyan-raev brought it up in this review.

Currently, I removed the permission thinks in my local, let me know if we want to keep it, but I think that we could skip it for now, because previous to this branch we have the passphrase in the same configuration file and this is one step to secure it, hence we can delay such thing for a future enhancement and have some discussion to identify tradeoffs and agree on in which ones we can afford.

uplink/config_test.go
ifraixedes: we need that behavior, however it is done in different way using `testcontext` package.

cmd/uplink/cmd/setup_test.go
ifraixedes: I'm not convince that it was a better solution, however currently `Stat` isn't needed because the permissions set/check have been removed.

cmd/uplink/cmd/setup_test.go
ifraixedes: It's a way to mark the `t.Run` which are `ok` (i.e. happy path) and the ones which check for errors.

pkg/miniogw/integration_test.go
ifraixedes: > Is there a reason the encryption key isn't part of the config? To me it seems it should be part of the config rather than a separate argument.

It was my intention but I couldn't make it work due to the inflexibility of `cfgstruct`; I tried to explain that in the `TODO` comments which should be removed before merging it.

> Although it might be that we are conflating two different things:
> 1. the command line arguments to uplink
> 2. the configuration of the uplink

It could be; being honest after spending quite a few time trying to figure it out if I could keep the _key_ in the configuration, I wasn't able to do so if we wanted to withdraw it from the command line arguments, or at least I didn't find the way, so if you know it, it will be welcome.

uplink/config.go
```diff
@@ -41,10 +43,47 @@ type RSConfig struct {
...
+//
+// It returns an error if:
+//
+// * The file doesn't exist.
+// * There is an I/O error.
+//
+// TODO: WIP#if/v3-1541 refactor cache the key after the firs read. The problem
+// is with pkg/cfgstruct which is inflexible on accepting unexported keys nor
+// several different types as a type of storj.Key, [32]byte, etc.
+func (cfg *EncryptionConfig) LoadKey() (key storj.Key, err error) {
```
ifraixedes: I'm wondering if the returned key should be a pointer because I read in a few places that we want to keep the minimum copies of them in memory as possible.

I mostly applied such logic, however I missed this one.

cmd/uplink/cmd/cp.go
```diff
@@ -83,7 +86,10 @@ func upload(ctx context.Context, src fpath.FPath, dst fpath.FPath, showProgress
 	}
 
 	var access libuplink.EncryptionAccess
-	copy(access.Key[:], []byte(cfg.Enc.Key))
+	access.Key, err = cfg.Enc.LoadKey()
```
ifraixedes: This method returns an error if the _encryption filepath_ is empty see https://github.com/storj/storj/pull/1782/commits/8030ae2ee8c1b06ffbc7f8b6c9398aa8ae886477#diff-060bc40979507170d2ceaad8dd024efaR69 and the reason of the change in the PR https://github.com/storj/storj/pull/1848 which points to the review conversation.

If you take a look to the last Jenkins build, the current _test-sim_ doesn't have a key filepath, so the test fail when this method is called on some of the `cmd/uplink` methods.

The current upstream behavior is to accept no encryption key but the change of making the `LoadKey` to return an error when the _encryption filepath_ is empty doesn't allow to have the same behavior.

I'm wondering what's the solution that fits better to the current code base:

1. Use a package level error variable (e.g. `var ErrEncryptionFilepathEmpty = errors.New(...`) and the callers can discern over that error and proceed without returning an error.
2. Use a [errs Class](https://godoc.org/github.com/zeebo/errs#Class) which wraps the error and the callers can discern it using its [`Has` method](https://godoc.org/github.com/zeebo/errs#Class.Has).

Regarding those 2 options, I would say, to use the option 1; I'm not a huge fan of _package variables and error values_, but it's mostly a concern for big code bases, for small ones they make the job (obviously considering that people is aware per Go conventions that they don't have to modify them).

What I don't like of the 2nd option is that it isn't that simple than the 1st one; it's more awkward from a general Go point of view and the callers must use the `Has` method, plus the class will be only used for one single thing, which it doesn't look that's the purpose of them and ultimately the class must also be assigned to a package var.

__Please reply with any other idea if you have it or with the preference based on the current code base and conventions__

jtolds: #2 has been more of a convention in our codebase for what that's worth.

jtolds: the main reason is #2 gives stack traces (basically all the same reasons the go2 xerrors proposed package does it similarly)

jtolds:
in terms of being able to touch the disk, i would say it's safe to assume anything under `/cmd/*` can read files off disk, but anything under `/lib/uplink/*` or anything it depends on (including `/uplink/*`) should not assume disk access is possible

egonelbre:
I've been aiming with the design such that only few things touch io (disk / net) directly:

* any database can read files from disk (for obvious reasons) 
* cmd/*
* pkg/transport and pkg/server for network

Everything else should never invoke them directly. e.g. Storage Node depends on `storagenode.DB` interface which handles all of the file handling.

In principle the whole network should be able to run in-memory without access to networking and disk. 
Although, I there are places where those ideas are still not met. Probably because 1. they haven't been written down and 2. we haven't yet fixed all the old code.

While it's less of a problem with storage node and satellite... it's probably critical to uplink.

cmd/uplink/cmd/cp.go
ifraixedes: At the end it isn't needed to do the discern because the upstream update of the `cmd/uplink` of using `lib/uplink` has made, that those commands which could use an encryption key or the zero value, requiring a non-zero encryption key.

ifraixedes:
@jtolds 

> in terms of being able to touch the disk, i would say it's safe to assume anything under /cmd/* can read files off disk, but anything under /lib/uplink/* or anything it depends on (including /uplink/*) should not assume disk access is possible

I'm confused with `/uplink`; I got answer that such package should be merged into others, so I don't know until what point now it's worth it to worry about it.

Anyway, being this the case, please confirm me if:

1. Make a change in the current implementation because this branch has added 2 methods to one type present in the `/upconfig` package which access to the disk. Because those 2  methods are used by `/cmd/uplink` and /cmd/gateway`, I would need that you tell me where I can put that shared logic or the other possibility is to make a copy into both packages.
2. We leave as it's in this branch and we open an issue for future refactoring.

---

@egonelbre 
> I've been aiming with the design such that only few things touch io (disk / net) directly: ...

That will nice to have because it will clarify the _io_ access and avoid mess.


---

There are 3 `TODOs` in this branch that they should be away before merging; being away is that I'm waiting to tell what to do with each one:

1. Just removed.
2. Create separated issues for them and removed.
3. Do what they require in this branch (this option will potentially delay to merge the main goal of this branch for a while).

You can easy find those TODOs comments in the source just executing `grep "TODO: if/v3-1541" -rn .` (or whatever be the link in Windows) from the root of the repository.

ifraixedes:
Closed in favor of https://github.com/storj/storj/pull/1895

Please follow the conversation in such one https://github.com/storj/storj/pull/1895#issuecomment-489630276

PR #1785, opened 2019-04-18
Refactor pb.Node protobuf 

kaloyan-raev:
What: 

Reduces the number of fields in `pb.Node` protobuf to just node ID and node Address.

The most significant part of the changes can be seen in this diff: https://github.com/storj/storj/compare/860c5d8..d3fcf45

The rest of the change is mostly refactoring the setup of some tests to use `testplanet.Run` so they are easier to fix. Unfortunately, this resulted in a big code difference.

Why:

We want the KAD DHT to act only as a distributed DNS, i.e. include info about node id and node address only. All other info related to nodes should be taken directly from the node itself via the `FetchInfo` gRPC method.

Thanks for submitting a PR!


pkg/kademlia/routing_test.go
```diff
@@ -177,52 +177,12 @@ func TestConnectionSuccess(t *testing.T) {
 	}
 }
 
-func TestUpdateSelf(t *testing.T) {
```
jtolds: what happens if a storage node's local address changes? i guess we don't currently have that since we configure the external address in the config file, but eventually we may want to handle ip changes automatically

jtolds: i guess i talked myself into that it's okay for now

pkg/dht/dht.go
```diff
@@ -7,14 +7,15 @@ import (
...
 
+	"storj.io/storj/pkg/overlay"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/storage"
 )
 
 // DHT is the interface for the DHT in the Storj network
 type DHT interface {
-	FindNear(ctx context.Context, start storj.NodeID, limit int, restrictions ...pb.Restriction) ([]*pb.Node, error)
```
jenlij: How are we filtering for restrictions now?

kaloyan-raev: We don't filter restrictions/capacity in DHT as this info is not available in DHT anymore. We now filter by capacity when selecting nodes from overlay cache.

pkg/kademlia/routing_test.go
kaloyan-raev: `UpdateSelf` was invoked just to update the available capacity in the `self` node in the own routing table. DHT does not hold this info anymore, so it does not make sense to have `UpdateSelf` anymore. Plus, it caused a data race, which was the main reason to remove it.

I guess updating the address of the `self` node in DHT was and still is an open issue.

internal/testplanet/uplink_test.go
```diff
@@ -21,196 +21,177 @@ import (
 )
 
 func TestUploadDownload(t *testing.T) {
-	ctx := testcontext.New(t)
-	defer ctx.Cleanup()
+	testplanet.Run(t, testplanet.Config{
```
egonelbre: Note, some of these were specifically made to run only against sqlite to make tests run faster. But, I Jenkins might help with this.

pkg/kademlia/kademlia_planet_test.go
```diff
@@ -38,13 +39,17 @@ func TestRequestInfo(t *testing.T) {
...
 		node := planet.StorageNodes[0]
-		info, err := planet.Satellites[0].Kademlia.Service.FetchInfo(ctx, node.Local())
+		info, err := planet.Satellites[0].Kademlia.Service.FetchInfo(ctx, node.Local().Node)
 		require.NoError(t, err)
 		require.Equal(t, node.Local().Type, info.GetType())
-		require.Equal(t, node.Local().Metadata.GetEmail(), info.GetOperator().GetEmail())
-		require.Equal(t, node.Local().Metadata.GetWallet(), info.GetOperator().GetWallet())
-		require.Equal(t, node.Local().Restrictions.GetFreeDisk(), info.GetCapacity().GetFreeDisk())
-		require.Equal(t, node.Local().Restrictions.GetFreeBandwidth(), info.GetCapacity().GetFreeBandwidth())
+		require.Equal(t, node.Local().Operator.GetEmail(), info.GetOperator().GetEmail())
```
egonelbre: Doesn't `require.Equal(t, node.Local(), info)` do the same thing?

kaloyan-raev: No. Comparing protobufs usually fails due to different values in hidden XXX fields.

internal/testplanet/uplink_test.go
kaloyan-raev: I needed an easy way to execute `refresh` in the discovery service during planet start, but without affecting the TestMergePlanets test, which freezes in Travis due to the discovery refresh.

pkg/kademlia/kademlia_planet_test.go
egonelbre: `require.Empty(t, cmp.Diff(a, b, cmp.Comparer(pb.Equal)))`

kaloyan-raev: I will check this spell.

kaloyan-raev: I could not use this directly to compare `node.Local()` to `info` because they are of different types (`NodeDossier` vs `InfoResponse`). But they worked for simplifying the checks for the `Operator`, `Capacity` and `Version` fields. Thanks for the hint!

pkg/kademlia/routing_test.go
kaloyan-raev: Well... the data race was not really resolved. So I brought back the `UpdateSelf`, but it now explicitly takes care for updating the capacity in the self node: https://github.com/storj/storj/pull/1785/commits/fd278005671deafbec76f089a9851505b1872a55

PR #1786, opened 2019-04-18
kademlia Bootstrap function has exponential backoff

navillasa:
Why:
- sometimes nodes aren't able to connect to bootstrap-- getting tls handshake failed errors. this retry should make it so they re-attempt to connect.

What: 
- adds retries up to a max time interval for the bootstrap call.


pkg/kademlia/kademlia_planet_test.go
```diff
@@ -131,35 +131,45 @@ func TestBootstrapBackoffReconnect(t *testing.T) {
 	go func() {
 		// This starts the unreliable proxy server and sets it to
 		// drop connections for the first 500 milliseconds that it's up.
-		err := proxy.start(done, droppedConnInterval)
+		err := proxy.run(done, droppedConnInterval)
+		require.NoError(t, err)
+	}()
+	defer func() {
+		err := proxy.close()
```
egonelbre: This doesn't wait for the goroutine to be finished.

navillasa: @egonelbre how should I wait for the goroutine to be finished?

navillasa: it's not working when I try to add `done <- true` in this defer.

egonelbre: If something internally fails in the loop, it isn't propagated to t.Fatal / require.NoError, but instead it's only being logged at the moment.

egonelbre: To be clear. No goroutine should be running once the test finishes. All errors must be properly handled, which means either explicitly ignored (e.g. context.Canceled) or checked with require.Erorr.

pkg/kademlia/kademlia_planet_test.go
```diff
@@ -86,3 +94,138 @@ func TestPingTimeout(t *testing.T) {
...
+		return nil, err
+	}
+	return &badProxy{
+		log:      log,
+		listener: l,
+	}, nil
+}
+
+func (proxy *badProxy) close() error {
+	atomic.StoreUint32(&proxy.closed, 1)
```
wthorp: To address Egon's concerns, I think you'd need to a waitgroup / errGroup in the struct (instead of simple go routines) and wait for everything to close here.

navillasa: I think everything is properly getting shut down now because of the atomic "closed" check, I think the problem still is that I'm just not propagating and checking all possible errors if I'm understanding Egon right.

pkg/kademlia/kademlia.go
```diff
@@ -153,48 +157,60 @@ func (k *Kademlia) Bootstrap(ctx context.Context) error {
...
-
-	// TODO(dylan): We do not currently handle this last bit of behavior.
-	// ```
-	// Finally, u refreshes all k-buckets further away than its closest neighbor.
-	// During the refreshes, u both populates its own k-buckets and inserts
-	// itself into other nodes' k-buckets as necessary.
-	// ``
-
-	return err
+	errGroup.Add(Error.New("unable to start bootstrap after final wait time of %ds", waitInterval/time.Second))
```
kaloyan-raev: time.Duration automatically formats well as string: https://play.golang.org/p/nMc2lerNcnK. No need to for this division.

navillasa: cool! fixed

pkg/kademlia/kademlia_planet_test.go
navillasa: welp ended up using errGroup after all 😅thanks Bill

navillasa:
> Do you find the variable name 'alpha' as opaque as I do?

haha apparently "alpha" is a system wide concurrency parameter-- but I don't want to mess with naming since I think that's outside the scope of what I'm trying to do with this pr (I didn't add that field, just moved it around).

PR #1787, opened 2019-04-19
pkg/cfgstruct: tie defaults to releases

jtolds:
this change makes it so that by default, the flag defaults are chosen based on whether the build was built as a release build or an ordinary build (be sure to check out scripts/release.sh for how release builds work).

release builds by default get release defaults, whereas ordinary builds by default get dev defaults.

any binary can have its defaults changed by specifying

 --defaults=dev

or

 --defaults=release

Also adds a commit to make release defaults more clear, which extends cfgstruct structs to support either a 'default' tag, or a pair of 'devDefault' and 'releaseDefault' tags, but not both, for added clarity

### DIFF
--- a/cmd/identity/batch.go
+++ b/cmd/identity/batch.go
@@ -46,13 +46,13 @@ var (
 		Concurrency   int    `help:"worker concurrency" default:"4"`
 		OutputDir     string `help:"output directory to place keys" default:"."`
 	}
-	isDev bool
+	defaults cfgstruct.BindOpt
 )
 
 func init() {
-	cfgstruct.DevFlag(rootCmd, &isDev, false, "use development and test configuration settings")
+	defaults = cfgstruct.DefaultsFlag(rootCmd)
 	rootCmd.AddCommand(keyGenerateCmd)
-	cfgstruct.Bind(keyGenerateCmd.Flags(), &keyCfg, isDev)
+	cfgstruct.Bind(keyGenerateCmd.Flags(), &keyCfg, defaults)
 }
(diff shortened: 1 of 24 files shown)


cmd/identity/batch.go
```diff
@@ -46,13 +46,13 @@ var (
 		Concurrency   int    `help:"worker concurrency" default:"4"`
 		OutputDir     string `help:"output directory to place keys" default:"."`
 	}
-	isDev bool
+	defaults cfgstruct.BindOpt
```
mniewrzal: maybe it can be just local var in `init`?

cmd/uplink/cmd/setup.go
```diff
@@ -31,7 +31,7 @@ var (
 	setupCfg    UplinkFlags
 	confDir     string
 	identityDir string
-	isDev       bool
+	defaults    cfgstruct.BindOpt
```
mniewrzal: maybe it can be just local var in `init`?

cmd/identity/batch.go
jtolds: ha, i tried, but this variable is used by multiple init methods in multiple files. i could combine all the init methods if that's preferred

cmd/uplink/cmd/setup.go
jtolds: same here

cmd/identity/batch.go
mniewrzal: no need, it was just nit

PR #1788, opened 2019-04-19
pkg/process: add --debug.trace-out option to binaries

jtolds:
this adds a flag to all pkg/process binaries where if set, will write a process-level trace for all monkit-instrumented functions as an svg to the target path.

usage like:

  uplink cp --debug.trace-out=trace.svg file sj://bucket/file

Note that this feature only works with codepaths that are kicked off in a *cobra.Cmd using process.Ctx() as the starting context, and it's important to note that this is much more useful when `defer mon.Task()(&ctx)(&err)` is littered everywhere

Thanks for submitting a PR!

### DIFF
--- a/pkg/process/exec_conf.go
+++ b/pkg/process/exec_conf.go
@@ -21,8 +21,11 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"github.com/zeebo/errs"
 	"go.uber.org/zap"
 	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+	"gopkg.in/spacemonkeygo/monkit.v2/collect"
+	"gopkg.in/spacemonkeygo/monkit.v2/present"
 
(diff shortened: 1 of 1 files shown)


pkg/process/exec_conf.go
```diff
@@ -154,6 +157,9 @@ func cleanup(cmd *cobra.Command) {
 	if internalRun == nil {
 		return
 	}
+
+	traceOut := cmd.Flags().String("debug.trace-out", "", "If set, a path to write a process trace SVG to")
```
JessicaGreben: do you think its worthwhile to add this option to the cmds so we can see it as an option? that make sense?

jtolds: is that not what this does?

JessicaGreben: nvm it is there 
```
$ uplink cp --help | grep -i trace
      --debug.trace-out string                 If set, a path to write a process trace SVG to
      --log.stack                   if true, log stack traces
```

PR #1779, opened 2019-04-18
Public Jenkins

egonelbre:
There are few additions:

1. Jenkinsfile that roughly runs the same steps as Travis
2. xunit.go that outputs xunit compatible output from Go tests
3. disabling few comments
4.  adjust test-sim to allow it to be run in parallel

Why: Travis is being too slow so replace it with something speedier.

Thanks for submitting a PR!

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?

PR #1791, opened 2019-04-21
Add Worker Ability to Satellite

stefanbenten:
What: This PR adds the ability to add worker instances to increase throughput of the satellite.
This is done by disabling Repair, Audit and Accounting (setting their interval to 0)

Why: Increase Stability, Availability and Throughput of the Satellite.

Note: I Added the PR Steps as part of the Jenkins PR( #1779 ) to test Jenkins Build through the PR.

Thanks for submitting a PR!


satellite/peer.go
```diff
@@ -351,7 +351,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 		pb.RegisterBandwidthServer(peer.Server.GRPC(), peer.Agreements.Endpoint)
 	}
 
-	{ // setup datarepair
+	if config.Repairer.Interval > 0 && config.Checker.Interval > 0 { // setup datarepair
```
egonelbre: We can change `Cycle` to handle this, rather than individually specify it.

stefanbenten: Oh okay, let me take a look into that.

internal/sync2/cycle.go
```diff
@@ -68,6 +68,10 @@ func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error)
 	defer close(cycle.stop)
 
 	currentInterval := cycle.interval
+	// Return if Interval is set to "disabled"
+	if currentInterval == 0 || currentInterval == -1 {
```
egonelbre: It should start things as stopped, rather than quitting immediately.

internal/sync2/cycle.go
```diff
@@ -68,6 +68,10 @@ func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error)
...
+		cycle.Pause()
+	}
 	cycle.ticker = time.NewTicker(currentInterval)
 	if err := fn(ctx); err != nil {
```
egonelbre: Also need to skip the initial call in that case.

stefanbenten:
Lets close this for now until we figured out, how to do it the "proper way"

PR #1794, opened 2019-04-22
interface tests belong to the interface, not the implementation

egonelbre:
Interface tests belong with the interface, rather than the implementation. Of course, implementation may have separate fine-grained tests.

There's alternative approach where the interface package defines tests for itself, however that is cumbersome to work in Go due to circular references and public definition of tests.

Thanks for submitting a PR!


satellite/orders/serials_test.go
```diff
@@ -1,14 +1,13 @@
...
 
-package satellitedb_test
+package orders_test
 
 import (
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
-
```
mobyvb: shouldn't this newline still be here?

egonelbre: ah, yes, typo

PR #1796, opened 2019-04-22
Expose actual error from dbx

egonelbre:
What: Handle properly wrapped constraint violation errors.

Why: Rest of the codebase assumes actual values from the database drivers, like `pgutil.IsContraintViolation` and `sql.ErrNoRows`. This causes issues where it checks against wrong values.

Thanks for submitting a PR!


satellite/satellitedb/dbx/gen.go
```diff
@@ -13,17 +13,9 @@ import (
...
 
 func init() {
-	// catch dbx errors
-	c := errs.Class("satellitedb")
+	// do not hide the actual error, necessary for other logic to work
 	WrapErr = func(e *Error) error {
-		switch e.Code {
-		case ErrorCode_NoRows:
-			return e.Err
-		case ErrorCode_ConstraintViolation:
```
mobyvb: if I recall correctly, this was added because when there is a constraint violation, no extra information is printed to let us know details about it. Which makes those types of errors more difficult to debug if we remove this case

egonelbre: I agree, but currently this is causing bugs in rest of the system. I'm fine with adding this back later, but at the moment I'm unsure where all the bugs are.

jtolds: any chance we can preserve the stack trace though? no stack trace is going to make database errors way harder to debug

egonelbre: This means going through all the error handling of our database code. I would rather have the safer behavior be the default for now.

jtolds: fwiw, the reason dbx has WrapErr is to support stack trace attaching via something like zeebo/errs. i'd rather have code that depends on err equality do something else instead of removing stack traces

jtolds: also, what safer behavior are you expecting? do you know we're doing err equality checking with postgres errors somewhere? isn't that easier to grep for than removing stack traces?

egonelbre: This probably means having flaky tests for a few days.

jtolds: how come?

jtolds: discussed on a chat: decided we'll just fix the problems we know about and keep the error wrapping. stack traces are so useful.

internal/errs2/ignore.go
```diff
@@ -12,20 +12,29 @@ import (
...
+		return is(err)
+	}
+	for {
+		if is(err) {
+			return true
 		}
-		unwrapped := errs.Unwrap(err)
-		if unwrapped == err {
-			return originalError
+		if err = errs.Unwrap(err); err == nil {
```
jtolds: the exit condition for the for loop should be err == errs.Unwrap(err). errs.Unwrap(err) returns err if err can't be unwrapped, not nil

PR #1797, opened 2019-04-22
fix saving uplink public key on satellite

egonelbre:
What: Remove using transactions and instead handle race manually.

Why: ensure that we properly add the certificate into the database and handling race properly.

Requires #1796 to be merged.

Thanks for submitting a PR!


satellite/satellitedb/certdb.go
```diff
@@ -18,40 +19,47 @@ type certDB struct {
 	db *dbx.DB
 }
 
-func (b *certDB) SavePublicKey(ctx context.Context, nodeID storj.NodeID, publicKey crypto.PublicKey) error {
-	tx, err := b.db.Open(ctx)
+func (certs *certDB) SavePublicKey(ctx context.Context, nodeID storj.NodeID, publicKey crypto.PublicKey) error {
+	_, err := certs.db.Get_CertRecord_By_Id(ctx, dbx.CertRecord_Id(nodeID.Bytes()))
+	if err == sql.ErrNoRows {
```
jtolds: dbx error wrapping. probably need some NoRows helper

satellite/satellitedb/certdb.go
```diff
@@ -18,40 +19,47 @@ type certDB struct {
...
 	}
 
-	return Error.Wrap(tx.Commit())
+	// TODO: use upsert here instead of create
+	_, err = certs.db.Create_CertRecord(ctx,
+		dbx.CertRecord_Publickey(pubbytes),
+		dbx.CertRecord_Id(nodeID.Bytes()),
+	)
+	// another goroutine might race to create the cert record, let's ignore that error
+	if pgutil.IsConstraintError(err) || sqliteutil.IsConstraintError(err) {
```
jtolds: might make sense to have a general dbutil.IsConstraintError so we don't have to find all these places if we add support for spanner or mysql or something

satellite/satellitedb/certdb.go
egonelbre: Currently it doesn't wrap sql.ErrNoRows

satellite/satellitedb/certdb.go
egonelbre: The reason I didn't want to add `dbutil.IsConstraintError` is because that would most likely end up pulling in additional dependencies. (e.g. storage node depending on postgres)

stefanbenten: Oh yeah. Thats a good point Egon.

satellite/satellitedb/dbx/gen.go
```diff
@@ -28,19 +28,21 @@ func init() {
 }
 
 // Unwrap returns the underlying error.
-func (e *Error) Unwrap() error {
-	return e.Err
-}
+func (e *Error) Unwrap() error { return e.Err }
+
+// Cause returns the underlying error.
+func (e *Error) Cause() error { return e.Err }
```
stefanbenten: Whats the difference to Unwrap?

egonelbre: `zeebo/errs` uses `Cause` for unwrapping, Go 1.13 errors proposal uses `Unwrap`

satellite/satellitedb/certdb.go
egonelbre: No, because it doesn't help here.

satellite/satellitedb/dbx/gen.go
stefanbenten: Okay, so we remove that once 1.13 is out?
Then its fine!

egonelbre: We can remove it once we start using the new errors package.

PR #1801, opened 2019-04-22
updates the libuplink documentation with examples

aligeti:
What: 
Updates the libuplink documentation with examples

Why:
To develop new end user applications

How to run locally the godoc

```~$ mktemp -d
/tmp/tmp.kJIMZfP0x7
~$ mkdir -p /tmp/tmp.kJIMZfP0x7/go/src/storj.io/storj
~$ cd work/storj
~/work/storj master$ rsync -a --delete ./ /tmp/tmp.kJIMZfP0x7/go/src/storj.io/storj/
~/work/storj master$ GOPATH=/tmp/tmp.kJIMZfP0x7/go/ godoc -http=:6060 -play```

then open http://localhost:6060/pkg/storj.io/storj/lib/uplink/

Thanks for submitting a PR!

### CODE REVIEW

**Comment by jtolds** [comment]
the formatting for the package documentation could be improved:

image

File: lib/uplink/example_test.go [hunk 1]
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		log.Fatalln("could not parse api key:", err)
+	}
+
+	ctx := context.Background()
+
+	// Create an Uplink object with a default config
+	upl, err := uplink.NewUplink(ctx, nil)
+	if err != nil {
+		fmt.Println("could not create new Uplink object:", err)
```
jtolds: stick with log.Fatal for all of the error handling. fmt.Print will keep letting the program run

lib/uplink/example_test.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	defer func() {
```
jtolds: i would prefer making a helper for closing, like,

```
  defer logClose(upl.Close)
```

and logClose takes an error and log.Fatals

lib/uplink/example_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// It is temporarily required to set the encryption key in project options.
+	// This requirement will be removed in the future.
+	opts := uplink.ProjectOptions{}
+	opts.Volatile.EncryptionKey = &encryptionKey
+
+	// Open up the Project we will be working with
+	proj, err := upl.OpenProject(ctx, satellite, apiKey, &opts)
+	if err != nil {
+		fmt.Println("could not open project:", err)
```
jtolds: same comment for all error handling in all of these examples. this is bad because the program will continue

lib/uplink/example_test.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	defer func() {
+		if err := proj.Close(); err != nil {
+			log.Printf("Failed to close input: %scope", err)
+		}
+	}()
```
jtolds: same comment about all of the defers. we want to do the bare minimum in the defers to appease the linter without distracting from what matters in the example

lib/uplink/example_test.go
aligeti: added a helper function to logClose

lib/uplink/example_test.go
aligeti: Changed to log.Fatal.

lib/uplink/example_test.go
aligeti: changed to use log.Fatal()

lib/uplink/example_test.go
aligeti: used log.Fatal to log the error in the helper logClose()

lib/uplink/example_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+func logClose(fn func() error) {
+	err := fn()
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+// ExampleProj_CreateBucket example documentation
+func ExampleProject_CreateBucket() {
```
egonelbre: Use a single example rather than multiple.

lib/uplink/example_test.go
```diff
@@ -0,0 +1,243 @@
...
+	myBucket        = "my-first-bucket"
+	myUploadPath    = "foo/bar/baz"
+	myData          = "one fish two fish red fish blue fish"
+	myEncryptionKey = "you'll never guess this"
+)
+
+func logClose(fn func() error) {
+	err := fn()
+	if err != nil {
+		log.Fatal(err)
```
egonelbre: This kills the whole process, do not use. Instead handle logging gracefully.

lib/uplink/example_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
+
+	"storj.io/storj/lib/uplink"
+	"storj.io/storj/pkg/storj"
+)
+
+const (
```
egonelbre: Set these inside the example rather than outside.

kaloyan-raev:
How can I generate the html output from the godoc comments?

lib/uplink/example_test.go
mniewrzal: @aligeti what with this comment?

aligeti: done

PR #1802, opened 2019-04-22
save raw output and output both stderr/stdin

egonelbre:
What: tee raw output to `.json` and pipe stdin and stderr to xunit script

Why: if something is printed to stderr, it's not captured in xunit output

Thanks for submitting a PR!


jtolds:
reasoning: is stderr supposed to be json output also? whether or not stderr is json, it's going to get interleaved with the valid json output line by line, so the stderr is most likely going to be some error that will cause the json to be invalid.

egonelbre:
`xunit` already knows how to deal with unexpected input, so it will fail in those cases.

PR #1803, opened 2019-04-22
fix non-unique path for injured segments

wthorp:
What: 
migrating injured segments fails on real data

Why:
path is not unique.  unique info does not 100% mean unique path


navillasa:
> path in this case includes project id and bucket name?

what's included in the injured segments path should be the same as it was prior to the migration, but I'm not totally sure. I think at least bucket name, but not sure about including project id.

PR #1812, opened 2019-04-23
Checker: invalid and offline nodes search update

Fadila82:
What: 
Simplify checker logic around invalidNodes, combineOfflineWithInvalid: get all the nodes via SQL, and use Go to determine if they're offline / invalid. Remove old DB method.

Jira:
https://storjlabs.atlassian.net/browse/V3-1600
Why:

Thanks for submitting a PR!


pkg/overlay/cache.go
```diff
@@ -132,6 +130,10 @@ func (node *NodeDossier) Online() bool {
 		node.Reputation.LastContactSuccess.After(node.Reputation.LastContactFailure)
 }
 
+func (node *NodeDossier) Valid(maxStats *NodeStats) bool {
```
kaloyan-raev: Should we call nodes that meet the uptime and audit success criteria as "vetted" instead of "valid"?

wthorp: that seems better than valid, given the whitepaper's terminology

pkg/datarepair/checker/checker.go
```diff
@@ -187,50 +171,29 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
...
-// Find invalidNodes by checking the audit results that are place in overlay
-func (checker *Checker) invalidNodes(ctx context.Context, nodeIDs storj.NodeIDList) (invalidNodes []int, err error) {
-	// filter if nodeIDs have invalid pieces from auditing results
+func (checker *Checker) getMissingPieces(ctx context.Context, pieces []*pb.RemotePiece) (missingPieces []int32, err error) {
+	var nodeIDs storj.NodeIDList
+	for _, p := range pieces {
+		nodeIDs = append(nodeIDs, p.NodeId)
+	}
+
+	nodeDossiers, err := checker.overlay.GetAll(ctx, nodeIDs)
```
kaloyan-raev: I think just `node` (`nodes`) is better var name than `nodeDossier` (`nodeDossiers`). 

It reads better in `node.Id`, `node.Operator.Wallet`, `node.Online()`, etc.

pkg/overlay/cache.go
ifraixedes: _vetted_ sound better to me for the same reason than @wthorp mentioned

pkg/overlay/cache.go
```diff
@@ -132,6 +130,10 @@ func (node *NodeDossier) Online() bool {
 		node.Reputation.LastContactSuccess.After(node.Reputation.LastContactFailure)
 }
 
+func (node *NodeDossier) Vetted(maxStats *NodeStats) bool {
```
kaloyan-raev: How about have `VettedFor` for the method name and `stats` instead of `maxStats`?

We also need a doc comment - something like:
> VettedFor checks if the node is vetted for the given stats.

egonelbre:
The reason this was in the database to avoid transferring too much data from the database.

wthorp:
@egonelbre I'm not sure I understand what still needs optimization here.

egonelbre:
By calling `GetAll` we end up downloading from database around `100 * 180bytes` per call (180 bytes is a rough estimate, it's probably more)... whereas it would suffice to only get `~3 * 32bytes` of data from the database.

wthorp:
ah.  moving the IsOffline check into SQL. gotcha.

PR #1813, opened 2019-04-23
api keys change encoding to base32

rikysya:
What: change encoding to base 32

Why: to avoid weird characters like `-` or `_`

Thanks for submitting a PR!


satellite/metainfo/metainfo.go
```diff
@@ -77,7 +77,7 @@ func (endpoint *Endpoint) validateAuth(ctx context.Context) (*console.APIKeyInfo
 		return nil, status.Errorf(codes.Unauthenticated, "Invalid API credential")
 	}
 
-	key, err := console.APIKeyFromBase64(string(APIKey))
+	key, err := console.APIKeyFromBase32(string(APIKey))
```
egonelbre: Let's make `auth.GetAPIKey` already do this conversion, makes the code bit shorter.

egonelbre:
Also, this would be a backwards incompatible change.

satellite/metainfo/metainfo.go
rikysya: Like to make `GetAPIKey` return `console.APIKey`?

egonelbre: Yeah, it seems weird that you need to decode the API key that is fetched from context.

PS: can be a separate PR.

satellite/console/apikeys.go
```diff
@@ -47,7 +47,7 @@ type APIKey [24]byte
 
 // String implements Stringer
 func (key APIKey) String() string {
-	return base64.URLEncoding.EncodeToString(key[:])
+	return base32.HexEncoding.EncodeToString(key[:])
```
egonelbre: How do people with existing setup convert their keys?

rikysya: From `base64` to `base32`? If yes than `APIKeys` itself is the same, so by converting from `base64` to `[]byte` and then to `base32` string

egonelbre: I understand that part. We now have people using the system, so we need to communicate that and exact things that they need to do to upgrade.

rikysya: Something like a small binary, or web page that will handle the conversion?

jtolds: is this change necessary? once we switch to macaroons we're going to have really long keys. if you're trying to avoid `_` and `-` could we perhaps use base58?

@egonelbre i'm okay with api keys becoming invalidated during the alpha from time to time

rikysya: @jtolds Brandon asked me to finish this as soon as possible 
https://storjlabs.atlassian.net/browse/V3-1549

PR #1814, opened 2019-04-23
updated audit configs

wthorp:
What: 
Update defaults:  AuditCount = 500, AuditSuccessRatio = 0.4, UptimeCount = 500, UptimeRatio = 0.9
Removed / replaced NewNodeAuditThreshold with AuditCount
Removed / replaced NewNodeCriteria with NodeCriteria 

Why:
Audit was effectively disabled with the previous settings.
Previously, if NewNodeAuditThreshold not set the same as AuditCount, nodes 'disappearred' if they were above NewNodeAuditThreshold but below AuditCount
NewNodeCriteria grew very similar to NodeCriteria, so we're just removed near-duplicate code.


pkg/overlay/config.go
```diff
@@ -33,15 +34,13 @@ type LookupConfig struct {
...
-	AuditCount        int64   `help:"the number of times a node has been audited" default:"0"`
-
-	NewNodeAuditThreshold int64   `help:"the number of audits a node must have to not be considered a New Node" default:"0"`
-	NewNodePercentage     float64 `help:"the percentage of new nodes allowed per request" default:"0.05"` // TODO: fix, this is not percentage, it's ratio
-
-	MinimumVersion string `help:"the minimum node software version for node selection queries" default:""`
+	UptimeRatio       float64       `help:"a node's ratio of being up/online vs. down/offline" releaseDefault:"0.9" devDefault:"0"`
+	UptimeCount       int64         `help:"the number of times a node's uptime has been checked to not be considered a New Node" releaseDefault:"500" devDefault:"0"`
+	AuditSuccessRatio float64       `help:"a node's ratio of successful audits" releaseDefault:"0.4" devDefault:"0"`
+	AuditCount        int64         `help:"the number of times a node has been audited to not be considered a New Node" releaseDefault:"50" devDefault:"0"`
```
littleskunk: 500 instead of 50

wthorp: my argument that a bad guy could fail to save things 500x as a 'new' node didn't dissuade you?  I feel like 500 is a terrible setting.

stefanbenten: I cant fail that often due to the audit success ratio @wthorp

wthorp: @stefanbenten the audit success ratio currently does not apply to new nodes

The SQL currently is:
AND total_audit_count < ?
AND last_contact_success > ?
AND last_contact_success > last_contact_failure

Do I need to change it so that it does?

jtolds: the audit success count and the uptime count minimums should not apply to new nodes, but the audit success ratio should, when the total audit count > 0

pkg/overlay/config.go
```diff
@@ -33,15 +34,13 @@ type LookupConfig struct {
...
-	AuditSuccessRatio float64 `help:"a node's ratio of successful audits" default:"0"`
-	AuditCount        int64   `help:"the number of times a node has been audited" default:"0"`
-
-	NewNodeAuditThreshold int64   `help:"the number of audits a node must have to not be considered a New Node" default:"0"`
-	NewNodePercentage     float64 `help:"the percentage of new nodes allowed per request" default:"0.05"` // TODO: fix, this is not percentage, it's ratio
-
-	MinimumVersion string `help:"the minimum node software version for node selection queries" default:""`
+	UptimeRatio       float64       `help:"a node's ratio of being up/online vs. down/offline" releaseDefault:"0.9" devDefault:"0"`
+	UptimeCount       int64         `help:"the number of times a node's uptime has been checked to not be considered a New Node" releaseDefault:"500" devDefault:"0"`
+	AuditSuccessRatio float64       `help:"a node's ratio of successful audits" releaseDefault:"0.4" devDefault:"0"`
```
jtolds: maybe we should make a note here that for beta and beyond we intend the release default to be .9 or higher for audit success ratio

pkg/overlay/config.go
jtolds: i think 50 is okay to no longer be a "new" node. we're not storing lots of data yet. here's the average on the olio satellite: satellite=> select avg(total_audit_count) from nodes;
         avg          
----------------------
 451.6950819672131148

soooo, yeah 500 is too high for a new node threshold. the average node across the whole network isn't even there yet.

wthorp: with the addition of enforcing the audit ratio, I'll happily adjust to 100

littleskunk: Please discuss that with the datascience team. 500 was there recommendation. I trust them and I am willing to try it. If you have good reasons to use 50 we should at least get there Ok. Please don't change any numbers without talking to them.

stefanbenten: I second that ^

pkg/overlay/config.go
stefanbenten: Yeah good idea, we probably should spent some time to walk through all current ToDo's in our codebase

pkg/overlay/config.go
wthorp: 500 it is.

PR #1820, opened 2019-04-23
wired up IsHealthy to config

wthorp:
What: 
Rename Online to IsOnline, VettedFor to IsHealthy.  added IsNew.
Made node IsOnline, IsNew, IsHealthy tests part of overlay
Wired IsHealthy up to configs

Why:
I liked those names better.  :P
IsOnline, IsNew, IsHealthy need to be config driven
Valid IsHealthy configs required for checker to produce valid results.

What is not included:
IsOnline is still hard-coded amd not wired up to configs, will accomplish in another PR

### DIFF
--- a/pkg/overlay/cache.go
+++ b/pkg/overlay/cache.go
@@ -109,20 +109,6 @@ type NodeDossier struct {
 	Version    pb.NodeVersion
 }
 
-// Online checks if a node is online based on the collected statistics.
-//
-// A node is considered online if the last attempt for contact was successful
-// and it was within the last hour.
-func (node *NodeDossier) Online() bool {
-	return time.Now().Sub(node.Reputation.LastContactSuccess) < OnlineWindow &&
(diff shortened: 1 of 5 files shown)


pkg/overlay/cache.go
```diff
@@ -183,6 +169,24 @@ func (cache *Cache) Get(ctx context.Context, nodeID storj.NodeID) (_ *NodeDossie
 	return cache.db.Get(ctx, nodeID)
 }
 
+// IsNew checks if a node is 'new' based on the collected statistics.
+func (cache *Cache) IsNew(node *NodeDossier) bool {
+	return node.Reputation.AuditCount > cache.preferences.AuditCount
```
mobyvb: wouldn't it only be new if the audit count was less than the preferred audit count? Not greater than?

wthorp: nice catch!  fixed.

wthorp: (obviously IsNew() is not yet used in the codebase)

PR #1821, opened 2019-04-23
Retry getting random pointer

mobyvb:
https://storjlabs.atlassian.net/browse/V3-1467
What: 
Retry getting a random pointer in the audit service in case it is a remote pointer, it is expired, or its segment size is 0

Why:
In the cases mentioned, `cursor.NextStripe` currently returns `nil, nil`, which should not happen, especially not if there are other valid pointers in the selected pointer items list

Thanks for submitting a PR!


pkg/audit/cursor.go
```diff
@@ -54,49 +55,30 @@ func (cursor *Cursor) NextStripe(ctx context.Context) (stripe *Stripe, err error
...
 	}
 
-	// get pointer info
-	pointer, err := cursor.pointerdb.Get(path)
-	if err != nil {
-		return nil, err
-	}
+	var pointer *pb.Pointer
+	errGroup := new(errs.Group)
+	for {
```
wthorp: If you had all bad pointers, would this run indefinitely?  Maybe random.Perm() would help?

pkg/audit/cursor.go
```diff
@@ -128,11 +110,43 @@ func getRandomStripe(pointer *pb.Pointer) (index int64, err error) {
...
+			return nil, newPointerItems, cursor.pointerdb.Delete(path)
+		}
+	}
+
+	if pointer.GetType() != pb.Pointer_REMOTE {
+		return nil, newPointerItems, Error.New("not remote pointer")
+	}
+
+	if pointer.GetSegmentSize() == 0 {
+		return nil, newPointerItems, Error.New("segment size 0")
```
navillasa: I feel like the error messages for these cases might be more useful as logs instead of compiled into the error group and returned when no pointers are left, but also I'm kind of inventing a situation where they would be necessary as logs. Seems fine!

pkg/audit/cursor.go
navillasa: I was also thinking this but it seemed unlikely that the only pointers in the pointerdb all would be segment size 0 and inline? but possible!

pkg/audit/cursor.go
```diff
@@ -128,11 +110,43 @@ func getRandomStripe(pointer *pb.Pointer) (index int64, err error) {
 	return randomStripeIndex.Int64(), nil
 }
 
-func getRandomPointer(pointerItems []*pb.ListResponse_Item) (pointer *pb.ListResponse_Item, err error) {
+// getRandomPointer attempts to get a random pointer from a list, and if the selected pointer is invalid, it returns a list without that item
+func (cursor *Cursor) getRandomPointer(pointerItems []*pb.ListResponse_Item) (pointer *pb.Pointer, newPointerItems []*pb.ListResponse_Item, err error) {
```
wthorp: Maybe rename to reflect that its getting a particular type of "random" now

wthorp: possibly also break this logic into smaller pieces... delete expired items maybe be conceptually correct but when its just slipped in there its like "what the heck!?"

pkg/audit/cursor.go
mobyvb: I got rid of errors for these cases since they are not really "errors"

pkg/audit/cursor.go
mobyvb: renamed to `getRandomValidPointer`

I am still thinking of an ideal way to remove the delete logic. It is hard to know how to separate it without making things more complex

mobyvb:
@ifraixedes it would be nice if we could directly query, but the pointerdb is not using a queryable database. It is using a key value store (`storj/storage/common.go`), which does not allow for much complexity in requesting things.

pkg/audit/cursor.go
```diff
@@ -120,19 +91,72 @@ func getRandomStripe(pointer *pb.Pointer) (index int64, err error) {
 		return 0, nil
 	}
 
-	randomStripeIndex, err := rand.Int(rand.Reader, big.NewInt(pointer.GetSegmentSize()/int64(redundancy.StripeSize())))
-	if err != nil {
-		return -1, err
+	var src cryptoSource
+	rnd := rand.New(src)
```
wthorp: nit:  using a crypto random source still seems like overkill to me

mobyvb: I also think it is overkill

wthorp: lol

pkg/audit/cursor.go
```diff
@@ -120,19 +91,72 @@ func getRandomStripe(pointer *pb.Pointer) (index int64, err error) {
...
 
-	return randomStripeIndex.Int64(), nil
+	errGroup.Add(Error.New("no valid node found in selection"))
+	return nil, "", errGroup.Err()
+}
+
+// cryptoSource implements the math/rand Source interface using crypto/rand
+type cryptoSource struct{}
+
+func (s cryptoSource) Seed(seed int64) {}
```
ifraixedes: Do we expect any call to this or if there is a call to this we should say "don't do this call"?

pkg/audit/cursor.go
ifraixedes: here is where I could say that `rnd` isn't a good abbreviation, however it's totally fine to me as it's used in a few lines only and it's clear from where it comes

pkg/audit/cursor.go
ifraixedes: never mind, I've just realized that could be called by `rand.New`

PR #1822, opened 2019-04-24
add pprof to uplink

JessicaGreben:
**What:** 
Add a config flag to enable cpu and memory profiling for uplink commands.

Example)
The following command generates a `cpu.prof` and `memory.prof` files:
`uplink --profile.cpu=cpu.prof --profile.mem=memory.prof cp test.file sj://testbucket`

**Why:** 
While there is the debug endpoint for running services, this addition provides a simple way to generate pprof reports for uplink commands that wasn't easily available with current debug pprof tooling.


cmd/uplink/cmd/cat.go
```diff
@@ -22,6 +22,10 @@ func init() {
 
 // catMain is the function executed when catCmd is called
 func catMain(cmd *cobra.Command, args []string) (err error) {
+	if *debugPprof {
```
egonelbre: This should probably be added at a higher level, not inside each individual command.

JessicaGreben: i was trying to figure that out, but couldn't think of a better place. where do you recommend?

mniewrzal: maybe `cmd/uplink/main.go`?

JessicaGreben: yeah i could do that. egon?

egonelbre: Yeah, `cmd/uplink/main.go` or you can set as `PreRunE` and `PostRunE` on `RootCmd`.

mniewrzal: PreRunE and PostRunE on RootCmd sounds cleaner

JessicaGreben: hmm ok let me try one of those areas and see what looks best. thanks!

JessicaGreben: I added the `PreRunE` and `PostRunE`.  I originally tried to add that on the `RootCmd` but they were **not** executed, I think because the `RootCmd` doesn't have a `RunE`, only the subcommands have a `RunE`.  Does that sound right?

cmd/uplink/cmd/root.go
```diff
@@ -76,3 +84,40 @@ func convertError(err error, path fpath.FPath) error {
...
+func runMemoryProf() error {
+	f, err := os.Create("memory.prof")
+	if err != nil {
+		return err
+	}
+	runtime.GC()
+	if err := pprof.WriteHeapProfile(f); err != nil {
+		return err
+	}
+	err = f.Close()
```
mniewrzal: this f.Close can be deferred

JessicaGreben: when i deferred it, i got an error from the linter that i didn't check the err. 
Re: https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/PR-1822/1/pipeline

egonelbre: `return f.Close()`

cmd/uplink/cmd/root.go
```diff
@@ -76,3 +84,40 @@ func convertError(err error, path fpath.FPath) error {
...
+
+func stopCPUStartMemProf(cmd *cobra.Command, args []string) error {
+	if *debugPprof {
+		pprof.StopCPUProfile()
+		return runMemoryProf()
+	}
+	return nil
+}
+
+func runMemoryProf() error {
```
egonelbre: `writeMemoryProfile`

cmd/uplink/cmd/root.go
```diff
@@ -26,11 +30,15 @@ type UplinkFlags struct {
 
 var cfg UplinkFlags
 
+var debugPprof = flag.Bool("debug.pprof", false, "if true, creates cpu.prof and memory.prof results in current directory")
```
egonelbre: Instead of making it a boolean, make it a string one for cpu and one for memory. This is the standard solution. https://golang.org/pkg/runtime/pprof/

```
-debug.cpuprofile cpu.prof -debug.memprofile mem.prof
```

I'm also wondering, whether it should be `-profile.cpu`, `-profile.mem` or similar, rather than just `debug`.

cmd/uplink/cmd/root.go
ifraixedes: It should be deferred as commented by @mniewrzal because if `pprof.WriteHeapProfile(f)` returns an error the file won't be closed.

JessicaGreben: the file will be closed by the OS once the process exits and since this is the uplink which is a short running process, i think this should be fine to `return f.Close()`

coyle:
Would it make sense to add a pprof endpoint ?

egonelbre:
@coyle not really, because `uplink` is not a server

PR #1827, opened 2019-04-24
wired up online config

wthorp:
What: 
This make the IsOnline duration window driven by config

Why:
IsOnline was previously hardcoded


pkg/overlay/cache.go
```diff
@@ -176,7 +166,7 @@ func (cache *Cache) IsNew(node *NodeDossier) bool {
 
 // IsOnline checks if a node is 'online' based on the collected statistics.
 func (cache *Cache) IsOnline(node *NodeDossier) bool {
-	return time.Now().Sub(node.Reputation.LastContactSuccess) < OnlineWindow &&
+	return time.Now().Sub(node.Reputation.LastContactSuccess) < cache.preferences.OnlineWindow &&
```
navillasa: optional/nit: just because I feel like this is a long check, it could maybe use a comment to explain the values it's using in case the constraints ever change.

wthorp: I'm not sure I understand.  Are you saying just put a comment in that tell you what the default is?

PR #1829, opened 2019-04-24
made checker kademlia aware

wthorp:
What:
Checker now cares about Kademlia. It waits for it to bootstrap. It checks Kadmelia when the overlay doesn't know about a node.

Why:
Checker never knew about Kademlia before. Anything missing from the overlay didn't get checked correctly.

### DIFF
--- a/pkg/datarepair/checker/checker.go
+++ b/pkg/datarepair/checker/checker.go
@@ -64,7 +64,7 @@ func NewChecker(pointerdb *pointerdb.Service, repairQueue queue.RepairQueue, ove
 func (checker *Checker) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	checker.kad.WaitForBootstrap()
+	//checker.kad.WaitForBootstrap()
 	return checker.Loop.Run(ctx, func(ctx context.Context) error {
 		err := checker.IdentifyInjuredSegments(ctx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/datarepair/checker/checker.go
```diff
@@ -182,7 +186,11 @@ func (checker *Checker) getMissingPieces(ctx context.Context, pieces []*pb.Remot
 	}
 
 	for i, node := range nodes {
-		if node == nil || !checker.overlay.IsOnline(node) || !checker.overlay.IsHealthy(node) {
+		if node == nil {
+			if _, err = checker.kad.FetchPeerIdentity(ctx, nodeIDs[i]); err != nil {
```
navillasa: why are we fetching peer identity is the node is nil? just curious

egonelbre:
How could this ever happen?

When we have something in the pointerdb, then we also should have it in overlay, otherwise it wouldn't have been selected. Since we must not delete things from overlay nodes table, then it should always be there.

pkg/datarepair/checker/checker.go
wthorp: a nil node means that the overlay doesn't have a record at all.  Its likely perfectly fine, but for whatever reason we don't have a record of it.  I'm still not sure how this is even possible, given that if pointerdb knows of a node, I would think that overlay should know of the node.  
Nevertheless, we are getting nils from overlay.  :/

wthorp:
I think this is totally unnecessary / incoherent

PR #1835, opened 2019-04-25
internal/textcontext: Fail test on error & docs

ifraixedes:
What: 

* Make the `Dir` function to fail the test if the directories cannot be
  created.
* Improve the docs to clarify the behavior of the exported functions.
* Make the default timeout exported for being able to see its value in
  the documentation.

Why: it has been mentioned in reviews that errors should never be ignored and regarding the docs, I needed such clarification, but the goal is not having to go to the source for finding them out.

Thanks for submitting a PR!


internal/testcontext/context.go
```diff
@@ -20,7 +20,8 @@ import (
 	"storj.io/storj/internal/memory"
 )
 
-const defaultTimeout = 3 * time.Minute
+// DefaultTimeout is the default timeout used by new context
+const DefaultTimeout = 3 * time.Minute
```
egonelbre: Why make this public?

internal/testcontext/context.go
```diff
@@ -120,11 +123,18 @@ func (ctx *Context) Dir(subs ...string) string {
 	})
 
 	dir := filepath.Join(append([]string{ctx.directory}, subs...)...)
-	_ = os.MkdirAll(dir, 0744)
+	err := os.MkdirAll(dir, 0744)
```
egonelbre: Note, Dir may be called multiple times for the particular directory. I wonder, do we need t check for is exists with mkdirall

internal/testcontext/context.go
```diff
@@ -120,11 +123,18 @@ func (ctx *Context) Dir(subs ...string) string {
...
 	return dir
 }
 
-// File returns a filepath inside temp
+// File returns a filepath inside temp.
+//
+// subs are joined into a single path adding a separator if necessary, being the
+// last one the name of the file. Directories are created, the file isn't
+// created.
 func (ctx *Context) File(subs ...string) string {
```
egonelbre: I wonder whether subs should be renamed to elem to match https://golang.org/pkg/path/filepath/#Join declaration.

ifraixedes: I wondered the same, but I didn't want to change things that I thought that they were expressly necessary.

If you think that's better, we can change it in this same branch.

internal/testcontext/context.go
ifraixedes: The reason is commented in the PR opening comment and in the commit message.

egonelbre: Ah, sorry missed the description... I did the inital comments when you pushed the changes

internal/testcontext/context.go
ifraixedes: it won't be an error but it should be suspicious why a test is trying to create the same directory more than once.

However what happen if this is used by several test cases which they don't have any relation and they use the same values for `subs`?

egonelbre: The `Dir` is meant to return a directory path and create when it doesn't yet exist. It makes some code more convenient, rather than have a variable per directory.

internal/testcontext/context.go
egonelbre: Yeah, change. If a commit changes less than 100 lines, then don't worry too much about putting other cleanup code in the same PR. Separating them out will cause both more dev and review time.

internal/testcontext/context.go
ifraixedes: I understood that correctly when I read the code, I think that I didn't understand well what you meant with 

>  I wonder, do we need t check for is exists with mkdirall

egonelbre: Ah nevermind found the answer in documentation. https://golang.org/pkg/os/#MkdirAll

I was wondering whether `MkdirAll` returns an error when the directory already exists, but it does not so we can mark the comment as resolved.

internal/testcontext/context.go
ifraixedes: I didn't do it because of the number of changes, it was mostly because I wanted to do the changes that I believed that were objective and `subs` replaced by `elem` I didn't think that such reason was a strong for arguing the change rather than being a change of subjective taste.

Because it bubbled from you and you are the author, so good to do it.

PR #1838, opened 2019-04-25
moved invalid/offline back into SQL

wthorp:
What: 
Moved logic of finding unhealthy or offline nodes from Go to DB.
Removed GetAll() receiver.
Added two old tests back in, updated for new SQL.
Defaulted audit/uptime ratios to 1 to simplify SQL.

Why:
The main motivation here is really just to continue to clean up and reduce the logic around online/offline and reliable/unreliable nodes.  Also Egon showed dissatisfaction with doing this in Go rather than SQL.  And maybe this will be faster.


pkg/overlay/cache.go
```diff
@@ -47,6 +47,9 @@ type DB interface {
 	Get(ctx context.Context, nodeID storj.NodeID) (*NodeDossier, error)
 	// GetAll looks up nodes based on the ids from the overlay cache
 	GetAll(ctx context.Context, nodeIDs storj.NodeIDList) ([]*NodeDossier, error)
+
+	// UnhealthyOrOffline filters a set of nodes to unhealth or offlines node, independent of new
+	UnhealthyOrOffline(context.Context, *NodeCriteria, storj.NodeIDList) (storj.NodeIDList, error)
```
egonelbre: `Unhealthy` should be sufficient for the method name.

satellite/satellitedb/overlaycache.go
```diff
@@ -169,6 +169,40 @@ func (cache *overlaycache) GetAll(ctx context.Context, ids storj.NodeIDList) ([]
...
+func (cache *overlaycache) UnhealthyOrOffline(ctx context.Context, c *overlay.NodeCriteria, nodeIds storj.NodeIDList) (goodNodes storj.NodeIDList, err error) {
+	args := make([]interface{}, len(nodeIds))
+	for i, id := range nodeIds {
+		args[i] = id.Bytes()
+	}
+	args = append(args, c.AuditSuccessRatio, c.UptimeSuccessRatio, time.Now().Add(-overlay.OnlineWindow))
+
+	rows, err := cache.db.Query(cache.db.Rebind(`
+		SELECT id FROM nodes
+		WHERE id IN (?`+strings.Repeat(", ?", len(nodeIds)-1)+`)
```
egonelbre: when `len(nodeIDs) == 0` then this would fail... add a check for that above.

pkg/overlay/cache.go
wthorp: I was trying to establish strong definitions between online / healthy / new because the old definitions left logical gaps.

egonelbre: Ah, ok... Wonder whether we should have a separate name for those nodes. e.g. "Problematic", But... yeah, `UnhealthyOrOffline` makes sense in that case.

wthorp: It was 'invalid', but that seemed to vague.  We briefly used (not) 'vetted' per the white paper until I realized this was different.  I'd like to have some word with a clear antonym.  Stable / unstable?

egonelbre: `Unstable`, `Unreliable`, `Unsafe`, `Undependable`, `Unsound` (see antonyms in https://www.thesaurus.com/browse/stable?s=t)

Maybe `Unreliable` or `Unsafe`?

satellite/satellitedb/overlaycache.go
```diff
@@ -169,6 +169,43 @@ func (cache *overlaycache) GetAll(ctx context.Context, ids storj.NodeIDList) ([]
 	return infos, nil
 }
 
+// ReliableAndOnline filters a set of nodes to reliable and online nodes, independent of new
+func (cache *overlaycache) ReliableAndOnline(ctx context.Context, c *overlay.NodeCriteria, nodeIds storj.NodeIDList) (goodNodes map[storj.NodeID]bool, err error) {
```
egonelbre: I'm not quite sure why you flipped the not? We expect most of the nodes from 100 to be online, so by getting `unreliable` we are transferring less data from database.

pkg/overlay/cache.go
```diff
@@ -267,6 +256,33 @@ func (cache *Cache) GetAll(ctx context.Context, ids storj.NodeIDList) (_ []*Node
 	return cache.db.GetAll(ctx, ids)
 }
 
+// UnreliableOrOffline filters a set of nodes to unhealth or offlines node, independent of new
+func (cache *Cache) UnreliableOrOffline(ctx context.Context, nodeIds storj.NodeIDList) (badNodes map[storj.NodeID]bool, err error) {
```
egonelbre: Also return a list rather than a map. Checking against a short list might be faster on average rather than a map and involve less allocations.

satellite/satellitedb/overlaycache.go
egonelbre: Also return a list rather than a map. Checking against a short list might be faster on average rather than a map and involve less allocations.

egonelbre: Instead of `c` use `criteria`

ifraixedes: I don't think that `crit` is better than `c`.
We have to eventually materialize some conventions for var names.

egonelbre: https://github.com/storj/storj/wiki/Style#variable-naming

egonelbre: The reason I mentioned it here because while I was reading:

```
args = append(args, c.AuditSuccessRatio, c.UptimeSuccessRatio, time.Now().Add(-overlay.OnlineWindow))
```

it wasn't immediately clear what `c.` was.

ifraixedes: I read those, but they are still unclear, considering just only one of the suggestions `crit`.

Materialize I meant to tune them.

ifraixedes: `cirt` sounds to me `critical`

egonelbre: I would go with `criteria`.

pkg/overlay/cache.go
```diff
@@ -267,6 +256,33 @@ func (cache *Cache) GetAll(ctx context.Context, ids storj.NodeIDList) (_ []*Node
...
+	defer mon.Task()(&ctx)(&err)
+	badNodes = make(map[storj.NodeID]bool)
+	goodNodes, err := cache.ReliableAndOnline(ctx, nodeIds)
+	if err == nil {
```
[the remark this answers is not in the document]
ifraixedes: I feel that's easy to understand the logic
```
if err != nil {
	return nil, err
}
...
..

return badNodes, nil
```

because the logic inside of the conditional is shorter and very concise and clarifies that the error is returned at that point and otherwise there is not returned error, but not just because it's consistent with the rest of the code base.

It's a personal opinion and the existent comment motivated me to comment too.

satellite/satellitedb/overlaycache.go
wthorp: "I'm not quite sure why you flipped the not?"  Ugh.  In part, I was trying to make those test pass and apparently I stopped thinking.  The test include items which are not in the overlay cache at all... which you can't query for, so I reversed it.  Counting the `ReliableAndOnline()` nodes could also replace `.GetAll()`.

However, I can't say it was ever my intention to write the fastest code, it is to write less fragile code, which almost always means less code.  I genuinely believed that Fadila's GetAll() w/ Go logic solves this problem well.  I've seen query planners make terrible decisions with IN clauses and other criteria.  Perhaps the best answer is to leave things where they were.

satellite/satellitedb/overlaycache.go
```diff
@@ -169,6 +169,43 @@ func (cache *overlaycache) GetAll(ctx context.Context, ids storj.NodeIDList) ([]
...
+	args := make([]interface{}, len(nodeIds))
+	for i, id := range nodeIds {
+		args[i] = id.Bytes()
+	}
+	args = append(args, c.AuditSuccessRatio, c.UptimeSuccessRatio, time.Now().Add(-overlay.OnlineWindow))
+
+	rows, err := cache.db.Query(cache.db.Rebind(`
+		SELECT id FROM nodes
+		WHERE id IN (?`+strings.Repeat(", ?", len(nodeIds)-1)+`)
+		AND (audit_success_ratio > ? OR total_audit_count = 0)
```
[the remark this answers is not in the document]
wthorp: nice catch

satellite/satellitedb/overlaycache.go
wthorp: Fixed: Created ReliableAndOnline and UnreliableOrOffline SQL methods

pkg/overlay/cache.go
wthorp: Fixed

satellite/satellitedb/overlaycache.go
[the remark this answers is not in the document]
wthorp: Fixed

pkg/overlay/statdb_test.go
```diff
@@ -78,6 +78,67 @@ func testDatabase(ctx context.Context, t *testing.T, cache overlay.DB) {
...
+		}
+		criteria := &overlay.NodeCriteria{
+			AuditSuccessRatio:  0.5,
+			UptimeSuccessRatio: 0.5,
+		}
+
+		goodNodes, err := cache.ReliableAndOnline(ctx, criteria, nodeIds)
+		require.NoError(t, err)
+		invalid := make(map[storj.NodeID]bool)
+		if err == nil {
```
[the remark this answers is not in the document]
wthorp: Fixed

pkg/overlay/cache.go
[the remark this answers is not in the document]
wthorp: Fixed

pkg/overlay/cache.go
```diff
@@ -259,6 +251,31 @@ func (cache *Cache) GetAll(ctx context.Context, ids storj.NodeIDList) (_ []*Node
 	return cache.db.GetAll(ctx, ids)
 }
 
+// UnreliableOrOffline filters a set of nodes to unhealth or offlines node, independent of new
+// Note that UnreliableOrOffline will not return node ids which are not in the database at all
```
egonelbre: Can we fix the query somehow for that case?

wthorp: Lol.  We can reverse the set logic.  😛  Technically it should never happen.

I'd love to spend extra cycles in dev / test so we can explode, but I'd rather not spend those cycles in production.  ¯\_(ツ)_/¯

egonelbre: Yeah, was wondering, maybe there's an easy way to achieve it without costing too much more cycles. But if not, it's fine.

egonelbre: Wondering also whether we should include that caveat in the method name?

Oh, what if we did the positive query on the sql side and then negated the result? :P

wthorp: I'm going to have to updated my Postgres, push a ton of fake data, and look at the query plan to have anything intelligent to say here.  I'm really not concerned with the outgoing I/O at this point.  I'm concerned that the query planner will do terrible things.

satellite/satellitedb/overlaycache.go
wthorp: Renamed UnreliableOrOffline to KnownUnreliableOrOffline for clarity

pkg/overlay/cache.go
wthorp: I did some testing and thinking about this... I think I'm happy where it is right now.  I'd rather be less aggressive in labeling people as offline (particularly in impossible scenarios such as poorly written tests), as that seems to be one of our major problems at the moment.  The SQL is trivial to correct if we change out minds later.

pkg/overlay/selection_test.go
```diff
@@ -15,6 +16,39 @@ import (
...
+
+		result, err = service.UnreliableOrOffline(ctx, []storj.NodeID{
+			planet.StorageNodes[0].ID(),
+			planet.StorageNodes[1].ID(),
+			planet.StorageNodes[2].ID(),
+		})
+		require.NoError(t, err)
+		require.Empty(t, result)
+
+		result, err = service.UnreliableOrOffline(ctx, []storj.NodeID{
```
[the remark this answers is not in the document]
wthorp: I pulled this test back in from an earlier PR that removed it.  I admit it does very little anymore, but I'm also not trying to fix tests in the PR, just to limit the logical surface area that touches online/offline and reliable/unreliable.

PR #1841, opened 2019-04-25
move nodeData length check after latestTally is updated in CalculateAtRestData

cam-a:
What: move nodeData length check after latestTally is updated

Why: In the event that we only have inline segments, SaveBucketTallies fails consistently because if there are no remote segments we return out of CalculateAtRestData before the timestamp is updated. The timestamp never changes, and we get a failure on unique constraint when trying to enter the bucket tallies into the database


pkg/accounting/tally/tally_test.go
```diff
@@ -59,6 +59,58 @@ func TestDeleteRawBefore(t *testing.T) {
...
+
+		// Setup: The data in this tally should match the pointer that the uplink.upload created
+		expectedTally := accounting.BucketTally{
+			Segments:       1,
+			InlineSegments: 1,
+			Files:          1,
+			InlineFiles:    1,
+			Bytes:          expectedTotalBytes,
+			InlineBytes:    expectedTotalBytes,
+			MetadataSize:   111, // brittle, this is hardcoded since its too difficult to get this value progamatically
```
jenlij: why is this one 111 rather than 112?

pkg/accounting/tally/tally_test.go
```diff
@@ -59,6 +59,58 @@ func TestDeleteRawBefore(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		// Run calculate twice to test unique constraint issue
+		for i := 0; i < 2; i++ {
+			latestTally, actualNodeData, actualBucketData, err := tallySvc.CalculateAtRestData(ctx)
+			require.NoError(t, err)
+			assert.Len(t, actualNodeData, 0)
+
+			_, err = planet.Satellites[0].DB.Accounting().SaveBucketTallies(ctx, latestTally, actualBucketData)
+			require.NoError(t, err)
```
jenlij: Unless we want to check if SaveBucketTallies worked as expected, we don't need to call it. In TestCalculateAtRestData, we just check the result of CalculateAtRestData, not the result of the db methods.

pkg/accounting/tally/tally_test.go
cam-a: the size of the metadata is different

pkg/accounting/tally/tally_test.go
cam-a: The error that this PR fixes was coming from SaveBucketTallies

cam-a: unique constraint on attempting to insert into the database

jenlij: gotcha

PR #1843, opened 2019-04-25
nodesDB to add only the storage nodes and not satellite & bootstrap nodes

aligeti:
What: 
satellite node and bootstrap node are sent as storage node list
https://storjlabs.atlassian.net/browse/V3-1490
Why:
In our current implementation,
(1)we are blindly first writing to the nodesDB anything in the discovery queue….
(2) then reiterating all thru the nodesDB and then updating the node with right info.
(3) Lot of things can go wrong while kad refresh() is iterating to update the last newly added nodes.  we have added new nodes with invalid nodetype, we dont even know if they are really a storage node. We just added them because they responded to the ping in kad discovery.
(4)searchGraveyard() is doing same as refresh. they have same discoveryqueue () and they are writing to nodeDB parallelly and this contention is causing unpredictable behaviour.

Fix : per our kad ppt
(1) on kad discovery ,  just moved the logic of vetting and making sure the node is valid and good before I add to the nodesDB.
(2) Also, once I fetched the node info, i will add to nodesDB only if it is storage node. Invalid nodes, bootstrap and satellite nodes are not added.
(3) doing this way our NodesDB is clean and only has entries of storagenode type.
(4) I tested using the storj-sim and nodes DB looks more clean and lines up with what presented in KAD ppt
(5) removed the searchGraveyard() 

Thanks for submitting a PR!

### DIFF
--- a/internal/testplanet/planet.go
+++ b/internal/testplanet/planet.go
@@ -456,7 +456,6 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				},
 			},
 			Discovery: discovery.Config{
-				GraveyardInterval: 1 * time.Second,
 				DiscoveryInterval: 1 * time.Second,
 				RefreshInterval:   1 * time.Second,
 				RefreshLimit:      100,
--- a/pkg/datarepair/datarepair_test.go
+++ b/pkg/datarepair/datarepair_test.go
@@ -31,7 +31,6 @@ func TestDataRepair(t *testing.T) {
 		// stop discovery service so that we do not get a race condition when we delete nodes from overlay cache
 		satellite.Discovery.Service.Discovery.Stop()
 		satellite.Discovery.Service.Refresh.Stop()
-		satellite.Discovery.Service.Graveyard.Stop()
 
 		satellite.Repair.Checker.Loop.Pause()
 		satellite.Repair.Repairer.Loop.Pause()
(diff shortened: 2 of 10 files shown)


aligeti:
I totally agree. This was to a quick test for Jens to try out on his dev satellite,  to nail down the root cause- Why is the query returning a satellite’s address even though query specifically asks for node type of “storagenode”. 

 After talking to Dennis I have a better understanding . But still no idea why a satellite node is sent even though, any node that is first entered into the nodesDB by default is set as type “node invalid”, which shouldn’t be also not selected in the query.

stefanbenten:
@aligeti Could you update the PR name and description on what you want to achieve?

aligeti:
The PR is to fix the satellite nodes being sent as part of the storage nodes list.The PR is about root causing and fixing it, which I got it in.  Now I am running the CI build and it is failing running test cases in storj.io/storj/pkg/kademlia  particularly TestMergePlanets() test case in merge_test.go. I don’t seem to clearly understand the merge test case, so needed some help. I have mentioned in the PR description what is the root cause. So basically avoided adding bootstrap and satellite nodes to the overlay cache db.

mniewrzal:
@aligeti https://storjlabs.atlassian.net/browse/V3-1490 looks to be marked as `Done` what is a plan for this PR? can it be closed?

stefanbenten:
This can be closed @mniewrzal @aligeti , as there as a bug in the selection that Jens fixed.

PR #1845, opened 2019-04-25
add more monkit tracing

JessicaGreben:
What & Why:
This PR makes some changes that allow us to create a better view of whats happening during uploads/downloads for the uplink. This is helpful for debugging performance.
These changes include:
- Add more monkit tracing
- Add more details in log messages
- Fix a bug with uplink/config and how we ensure the correct value is set for `enc.block-size` setting.

### DIFF
--- a/pkg/eestream/decode.go
+++ b/pkg/eestream/decode.go
@@ -176,7 +176,8 @@ func (dr *decodedRanger) Size() int64 {
 	return blocks * int64(dr.es.StripeSize())
 }
 
-func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (io.ReadCloser, error) {
+func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (rc io.ReadCloser, err error) {
+	defer mon.Task()(&ctx)(&err)
 	// offset and length might not be block-aligned. figure out which
 	// blocks contain this request
 	firstBlock, blockCount := encryption.CalcEncompassingBlocks(offset, length, dr.es.StripeSize())
@@ -209,7 +210,7 @@ func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (io.Re
 	// decode from all those ranges
 	r := DecodeReaders(ctx, readers, dr.es, blockCount*int64(dr.es.StripeSize()), dr.mbm)
 	// offset might start a few bytes in, potentially discard the initial bytes
-	_, err := io.CopyN(ioutil.Discard, r,
+	_, err = io.CopyN(ioutil.Discard, r,
 		offset-firstBlock*int64(dr.es.StripeSize()))
 	if err != nil {
 		return nil, Error.Wrap(err)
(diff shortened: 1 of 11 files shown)


pkg/eestream/decode.go
```diff
@@ -176,7 +176,8 @@ func (dr *decodedRanger) Size() int64 {
 	return blocks * int64(dr.es.StripeSize())
 }
 
-func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (io.ReadCloser, error) {
+func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (rc io.ReadCloser, err error) {
+	defer mon.Task()(&ctx)(&err)
```
jtolds: just as a heads up, Range returns before the data actually starts being transferred, because what Range is returning is a io.ReadCloser. so in addition to this we should probably make a monkit io.ReadCloser wrapper or something that keeps track of how long a ReadCloser is open for.

uplink/config.go
```diff
@@ -110,7 +110,7 @@ func (c Config) GetMetainfo(ctx context.Context, identity *identity.FullIdentity
 	}
 	segments := segments.NewSegmentStore(metainfo, ec, rs, c.Client.MaxInlineSize.Int(), maxEncryptedSegmentSize)
 
-	if c.RS.ErasureShareSize.Int()*c.RS.MinThreshold%c.Enc.BlockSize.Int() != 0 {
+	if c.Enc.BlockSize.Int()%c.RS.ErasureShareSize.Int()*c.RS.MinThreshold != 0 {
```
jtolds: oh wait i'm sorry, i had this backwards in the presentation yesterday. the code is right. we want the stripes to be as small as possible so audits can be as small as possible

pkg/eestream/decode.go
JessicaGreben: i mostly added this tracing because there are 12 different objects that implement the Range() interface and I was struggling knowing which were being called in the code for the download

jtolds: :+1:

uplink/config.go
JessicaGreben: hmmm. ok well it still seems like _something_ is wrong here.  If the code was correct originally, then I think the error message is misleading and might need to change.  For example, before this code change, if I get this error:

```$ uplink cp go.sum sj://test3
Fatal error: Uplink configuration error: EncryptionBlockSize must be a multiple of ErasureShareSize * RS MinThreshold```

with this uplink config:

```
# size (in bytes) of encrypted blocks
enc.block-size: 8.0 KiB
# the size of each new erasure sure in bytes
rs.erasure-share-size: 1.0 KiB
# the minimum pieces required to recover a segment. k.
rs.min-threshold: 4
```

jtolds: yeah you're right, the error message is wrong

JessicaGreben: ok let me change the error message then.

jtolds: okay i was all wrong again

PR #1847, opened 2019-04-25
Console buckets page

rikysya:
What: add back-end and front-end logic for  buckets page to console web UI

Why: to display buckets.

Thanks for submitting a PR!


satellite/satellitedb/usagerollups.go
```diff
@@ -177,6 +178,153 @@ func (db *usagerollups) GetBucketUsageRollups(ctx context.Context, projectID uui
 	return bucketUsageRollups, nil
 }
 
+// GetBucketTotals retrieves bucket usage totals for period of time
+func (db *usagerollups) GetBucketTotals(ctx context.Context, projectID uuid.UUID, cursor console.BucketUsageCursor, since, before time.Time) (*console.BucketUsagePage, error) {
+	search := fmt.Sprintf("%s%%", cursor.Search)
```
egonelbre: Why not `cursor.Search + "%"`?

satellite/satellitedb/usagerollups.go
```diff
@@ -177,6 +178,153 @@ func (db *usagerollups) GetBucketUsageRollups(ctx context.Context, projectID uui
...
+	page := &console.BucketUsagePage{
+		Search: cursor.Search,
+		Limit:  cursor.Limit,
+		Offset: uint64((cursor.Page - 1) * cursor.Limit),
+	}
+
+	countQuery := db.db.Rebind(`SELECT COUNT(DISTINCT bucket_name) 
+				FROM bucket_bandwidth_rollups 
+				WHERE project_id = ? AND interval_start >= ? AND interval_start <= ?
+				AND CAST(bucket_name as TEXT) LIKE ?`)
```
egonelbre: Why is the bucket usage a like query? This should probably be an exact query instead.

satellite/satellitedb/usagerollups.go
```diff
@@ -177,6 +178,153 @@ func (db *usagerollups) GetBucketUsageRollups(ctx context.Context, projectID uui
...
 
+// GetBucketTotals retrieves bucket usage totals for period of time
+func (db *usagerollups) GetBucketTotals(ctx context.Context, projectID uuid.UUID, cursor console.BucketUsageCursor, since, before time.Time) (*console.BucketUsagePage, error) {
+	search := fmt.Sprintf("%s%%", cursor.Search)
+
+	if cursor.Limit > 50 {
+		cursor.Limit = 50
+	}
+	if cursor.Page == 0 {
+		return nil, errs.New("page can not be 0")
```
egonelbre: Don't pages usually start at `0`?

satellite/console/service.go
```diff
@@ -786,6 +786,24 @@ func (s *Service) GetProjectUsage(ctx context.Context, projectID uuid.UUID, sinc
 	return projectUsage, nil
 }
 
+// GetBucketTotals retrieves paged bucket total usages since project creation
+func (s *Service) GetBucketTotals(ctx context.Context, projectID uuid.UUID, cursor BucketUsageCursor, before time.Time) (*BucketUsagePage, error) {
```
egonelbre: Tests are missing for these.

satellite/console/usagerollups.go
```diff
@@ -14,6 +14,7 @@ import (
 type UsageRollups interface {
```
egonelbre: Tests are missing for these.

satellite/satellitedb/usagerollups.go
rikysya: This is for the search on buckets page, there is a search input where you can find bucket by name, also when you create bucket using uplink cli you give it name in a form `sj://bucket_name` so I assumed it's a text. Doesn't purpose of `bucket_name` is to give it a human readable id/reference?

satellite/satellitedb/usagerollups.go
rikysya: I was not sure where should I add one on the front-end part, or to make it start from 1 on the back-end. If the second option is more correct I can do it that way

satellite/satellitedb/usagerollups.go
egonelbre: We can assume that most people will use human readable id-s... however then there are probably the other people who will use a random byte sequence. Or some really weird encoding that isn't valid utf8.

satellite/satellitedb/usagerollups.go
egonelbre: I'm unsure. Whatever is the standard in graphql

satellite/satellitedb/usagerollups.go
rikysya: https://github.com/storj/storj/blob/7ce9a60c858820f5e95b321bde5eed5eb977c642/lib/uplink/bucket.go#L22 - `bucketName` string

I think if someone will use some strange byte sequence, then It's ok if it won't be found by search. 

Also if we can use any `byte` for `bucket_name`, then https://github.com/storj/storj/blob/7ce9a60c858820f5e95b321bde5eed5eb977c642/satellite/satellitedb/orders.go#L86 won't work if `bucket_name` contains one or more `/` runes

egonelbre: Go `string` is an arbitrary `[]byte` sequence and not UTF-8.

I'm aware that's an issue in several places. We should use a proper encoding for paths.

thepaul: `bucket_name` must not contain a `/` byte. neither can any other path components, because we use `'/'` internally as a delimiter. that would break a lot of things. I'm pretty sure we check for that in the bucket_name case at least.

> We should use a proper encoding for paths.

I strongly disagree that we should treat paths as anything but plain bytes, for the reasons outlined here: https://gist.github.com/thepaul/d4eb85dabf1a25bf1061b2d642080528 (from an early review of the v3 whitepaper)

PR #1850, opened 2019-04-26
Set connection read/write timeout 

mniewrzal:
What: Set read/write timeout for each net.Conn created in transport.go

Why: With super slow connection we need to have more granular timeout and be able to reset timeout for each read/write operation. Its different approach to #1764 
https://storjlabs.atlassian.net/browse/V3-1626


mniewrzal:
this is still in WIP but test was added with #1764

pkg/transport/timeout.go
```diff
@@ -29,68 +29,49 @@ type InvokeStreamTimeout struct {
...
+type timeoutConn struct {
+	conn    net.Conn
 	timeout time.Duration
-	stream  grpc.ClientStream
-	cancel  func()
 }
 
-func (wrapper *clientStreamWrapper) Header() (metadata.MD, error) {
-	return wrapper.stream.Header()
+func (conn *timeoutConn) Read(b []byte) (n int, err error) {
```
coyle: comment on Read ?

pkg/transport/timeout.go
```diff
@@ -29,68 +29,49 @@ type InvokeStreamTimeout struct {
...
+	if err != nil {
+		return 0, err
+	}
+	return conn.conn.Write(b)
 }
 
-func (wrapper *clientStreamWrapper) Context() context.Context {
-	return wrapper.stream.Context()
+func (conn *timeoutConn) Close() error {
+	return conn.conn.Close()
```
coyle: nit -- conn.conn feels weird to me. Could we rename conn ?

pkg/transport/transport.go
```diff
@@ -71,7 +72,13 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 		grpc.WithBlock(),
 		grpc.FailOnNonTempDialError(true),
 		grpc.WithUnaryInterceptor(InvokeTimeout{transport.requestTimeout}.Intercept),
-		grpc.WithStreamInterceptor(InvokeStreamTimeout{transport.requestTimeout}.Intercept),
+		grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", addr)
```
coyle: I just learned something ! this is neat

pkg/transport/transport.go
```diff
@@ -104,7 +111,13 @@ func (transport *Transport) DialAddress(ctx context.Context, address string, opt
...
```
coyle: would it be a benefit to have a `const TCP = "tcp"`

pkg/transport/transport.go
```diff
@@ -37,6 +37,9 @@ type Transport struct {
 	requestTimeout time.Duration
 }
 
+// TCP represents TCP network type
+const TCP = "tcp"
```
egonelbre: I'm not certain that this improves much, even more I'm more accustomed to seeing "tcp" as an argument rather than a constant.

pkg/transport/transport.go
egonelbre: To me this seems more weird. I'm expecting to see "tcp", so seeing a constant forces me to lookup what the constant value is...

pkg/transport/transport.go
mniewrzal: yeah, I was too fast with this

mniewrzal:
@littleskunk according to jira comment you have problems with uploading. Can you try this change?

PR #1852, opened 2019-04-26
avoid race in lib/uplink

egonelbre:
lib/uplink ends up modifying the config struct, however the caller may not expect that. Make an internal copy of the config struct.

Thanks for submitting a PR!

### DIFF
--- a/lib/uplink/project.go
+++ b/lib/uplink/project.go
@@ -57,38 +57,38 @@ type BucketConfig struct {
 	}
 }
 
-func (cfg *BucketConfig) clone() *BucketConfig {
-	clone := *cfg
+func (c *BucketConfig) clone() *BucketConfig {
+	clone := *c
 	return &clone
 }
 
-func (cfg *BucketConfig) setDefaults() {
-	if cfg.PathCipher == storj.EncUnspecified {
-		cfg.PathCipher = defaultCipher
+func (c *BucketConfig) setDefaults() {
+	if c.PathCipher == storj.EncUnspecified {
+		c.PathCipher = defaultCipher
 	}
-	if cfg.EncryptionParameters.CipherSuite == storj.EncUnspecified {
-		cfg.EncryptionParameters.CipherSuite = defaultCipher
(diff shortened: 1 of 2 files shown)


egonelbre:
It's often more convenient to use `NewSomething(ctx, "name", nil)`, rather than `NewSomething(ctx, "name", uplink.BucketConfig{})`... but yeah, I was trying to avoid changing the public part or mainly just being lazy.

lib/uplink/project.go
```diff
@@ -57,33 +57,38 @@ type BucketConfig struct {
...
 
-func (c *BucketConfig) setDefaults() {
-	if c.PathCipher == storj.EncUnspecified {
-		c.PathCipher = defaultCipher
+func (cfg *BucketConfig) clone() *BucketConfig {
+	clone := *cfg
+	return &clone
+}
+
+func (cfg *BucketConfig) setDefaults() {
```
thepaul: I know that your preference is more descriptive names, but I want to push back on that a bit. I have a strong preference for short names for receivers, since they are used frequently, and signal more effectively to me that they _are_ receivers. This is a pretty widely embraced convention in Go land, and it causes me confusion when we depart from it. Also I feel like we adhere to the official Go style guidelines in some places even when they are silly, so maybe we should follow them here too, where they make sense.

This isn't specific to the PR at hand, though. Should I take this over to github.com/storj/storj/wiki ?

egonelbre: In this case it was about consistency in one file it's named `cfg` in the other `c`. But, sure I can revert.

egonelbre: The reason I made this change was because I got confused by the difference of `cfg` and `c`.

thepaul: Oh dang. Consistency is more important. Good call. Keep this if you like, or we can change the other place.

egonelbre: Changing in the other place seems even more awkward.

egonelbre: re-revert

PR #1854, opened 2019-04-26
don't encrypt bucket info

jtolds:
Background: see https://github.com/storj/storj/pull/1761

Summary: we should not require an encryption key to manage buckets (though we require an encryption key to manage objects within a bucket), so that bucket management can be done in the web interface. This means the encryption key used when opening a project is going away.

We have a network wipe planned for next release, so for that wipe let's make sure data and buckets uploaded after that wipe do not require an encryption key.


jtolds:
It's just the metadata about a bucket itself that shouldn't be encrypted (and should never be encrypted). Attempts to modify the client to do anything else will be equivalent to modifying the client to upload garbage. It will break, yeah.

Objects within a bucket should always be encrypted unless explicitly opted out. 

We just have this weird thing where each bucket has a single metadata file that we store that we should probably move out of pointerdb entirely. It should always be accessible to the web interface and not be encrypted (which is okay because the metadata-about-the-bucket) won't hold anything private. 

We should probably sign the metadata, but that's for a project down the road.

ifraixedes:
> Objects within a bucket should always be encrypted unless explicitly opted out.

Last week there was come concern regarding of not allowing zero value keys, which looked to be the way to not encrypt the objects, hence, I'm wondering how now, we are going to allow the users to opt-out of encrypting the objects.

ifraixedes:
As far as I see, we could remove the `EncryptionKey` field of the [lib/uplink#ProjectOptions struct](https://godoc.org/storj.io/storj/lib/uplink#ProjectOptions) because it's a subfield of `Volatile` and it's only used by [lib/uplink#Upling.OpenProject](https://godoc.org/storj.io/storj/lib/uplink#Uplink.OpenProject) which is basically used by `cmd/uplink` and `cmd/gateway` (a part of some tests in other packages).

jtolds:
> As far as I see, we could remove the EncryptionKey field of the lib/uplink#ProjectOptions struct because it's a subfield of Volatile and it's only used by lib/uplink#Upling.OpenProject which is basically used by cmd/uplink and cmd/gateway (a part of some tests in other packages).

I agree, this is what should happen for sure.

jenlij:
Continuing work here https://github.com/storj/storj/pull/1979

jenlij:
working off this branch again

pkg/storage/buckets/store.go
```diff
@@ -126,16 +124,51 @@ func (b *BucketStore) Put(ctx context.Context, bucketName string, inMeta Meta) (
...
+	})
+
+	var exp timestamp.Timestamp
+	pointer := &pb.Pointer{
+		Type:           pb.Pointer_INLINE,
+		InlineSegment:  nil,
+		SegmentSize:    0,
+		ExpirationDate: &exp,
+		Metadata:       streamMeta,
+	}
```
jenlij: Possibly create a method to fill user meta and subsequent structs?

pkg/storage/buckets/store.go
```diff
@@ -40,8 +43,7 @@ type ListItem struct {
 
 // BucketStore contains objects store
```
JessicaGreben: this comment is out of date now that objects.Store is removed

pkg/storage/buckets/store.go
```diff
@@ -88,15 +68,29 @@ func (b *BucketStore) Get(ctx context.Context, bucket string) (meta Meta, err er
...
 		return Meta{}, err
 	}
 
-	return convertMeta(objMeta)
+	pointer, err := b.metainfo.SegmentInfo(ctx, bb, objectPath, segmentIndex)
+	if err != nil {
+		return Meta{}, err
+	}
+
+	// metadata conversion function call: pointer -> segment -> stream -> object -> bucket
```
JessicaGreben: This conversion from pointer to bucket `Meta` object would be simplified if there were a metainfo call available like `b.metainfo.BucketInfo()`, but currently that doesn't exist.  However, it does seem like that change is captured in the metainfo refactor. See design doc for details: https://github.com/storj/storj/blob/e76ecf47d8609fa1d873f9f3e00134f7853b85ad/docs/design/metainfo-rpc.md

In the meantime, since this refactor is not implemented yet, maybe what you have here is good.  Maybe add a TODO comment about once refactor is done, we can replace this conversion. Or not. up to you

pkg/storage/buckets/store.go
```diff
@@ -54,30 +56,8 @@ type Meta struct {
...
-	prefixed := prefixedObjStore{
-		store:  objects.NewStore(b.stream, m.PathEncryptionType),
-		prefix: bucket,
-	}
-	return &prefixed, nil
+func NewStore(metainfo metainfo.Client) Store {
+	return &BucketStore{metainfo: metainfo}
 }
 
 // Get calls objects store Get
```
JessicaGreben: comment out of date

pkg/storage/streams/store.go
```diff
@@ -63,7 +63,7 @@ type streamStore struct {
 	segmentSize  int64
 	rootKey      *storj.Key
 	encBlockSize int
-	cipher       storj.Cipher
+	Cipher       storj.Cipher
```
JessicaGreben: if we need the Cipher exported, could we make a method a `streamStore.GetCipher()`?

lib/uplink/uplink.go
```diff
@@ -130,11 +125,7 @@ func NewUplink(ctx context.Context, cfg *Config) (*Uplink, error) {
 }
 
 // ProjectOptions allows configuration of various project options during opening
-type ProjectOptions struct {
-	Volatile struct {
-		EncryptionKey *storj.Key
-	}
-}
+type ProjectOptions struct{}
```
JessicaGreben: i vote to remove ProjectOptions all together

pkg/storage/buckets/store.go
```diff
@@ -29,7 +33,6 @@ type Store interface {
 	Put(ctx context.Context, bucket string, inMeta Meta) (meta Meta, err error)
 	Delete(ctx context.Context, bucket string) (err error)
 	List(ctx context.Context, startAfter, endBefore string, limit int) (items []ListItem, more bool, err error)
-	GetObjectStore(ctx context.Context, bucketName string) (store objects.Store, err error)
```
JessicaGreben: im curious why remove `GetObjectStore`?

pkg/storage/buckets/store.go
```diff
@@ -54,30 +56,8 @@ type Meta struct {
...
-	return &BucketStore{store: store, stream: stream}
-}
-
-// GetObjectStore returns an implementation of objects.Store
-func (b *BucketStore) GetObjectStore(ctx context.Context, bucket string) (objects.Store, error) {
-	if bucket == "" {
-		return nil, storj.ErrNoBucket.New("")
-	}
-
-	m, err := b.Get(ctx, bucket)
```
JessicaGreben: I think we might need to call b.Get() in order to get the `PathEncryptionType`.  It looks like the `streams.Cipher` is trying to be used here instead (re: https://github.com/storj/storj/pull/1854/files#diff-19374caab19525948168b09d2e0f94aaR166), but I think  streams.cipher is used for data encryption, not for the pathCipher

pkg/storage/buckets/prefixed.go
```diff
@@ -13,54 +13,59 @@ import (
 	"storj.io/storj/pkg/storage/objects"
```
JessicaGreben: i know this isn't about your PR, but its odd this `preefixed.go` file is under `pkg/storage/buckets/` and not in `pkg/storage/objects`

pkg/storage/buckets/store.go
```diff
@@ -126,16 +124,51 @@ func (b *BucketStore) Put(ctx context.Context, bucketName string, inMeta Meta) (
...
-	var exp time.Time
-	m, err := b.store.Put(ctx, bucketName, r, pb.SerializableMeta{UserDefined: userMeta}, exp)
+
+	metadata, err := proto.Marshal(&pb.SerializableMeta{UserDefined: userMeta})
+	if err != nil {
+		return meta, err
+	}
+
+	streamInfo, err := proto.Marshal(&pb.StreamInfo{
+		NumberOfSegments: 1,
```
JessicaGreben: should this be 0?

pkg/storage/buckets/store.go
jenlij: We no longer have an ObjectStore on the BucketStore struct

JessicaGreben: but why? why remove object store?

pkg/storage/buckets/store.go
```diff
@@ -4,21 +4,25 @@
...
 
 	"storj.io/storj/pkg/encryption"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/storage/meta"
 	"storj.io/storj/pkg/storage/objects"
+	"storj.io/storj/pkg/storage/segments"
 	"storj.io/storj/pkg/storage/streams"
 	"storj.io/storj/pkg/storj"
-	"storj.io/storj/storage"
+	"storj.io/storj/uplink/metainfo"
```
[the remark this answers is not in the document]
jenlij: Yeah it is interesting that the metainfo package is nested under the uplink folder, but since it technically is a separate package that doesn't rely on uplink, I think it's ok for now.

pkg/storage/buckets/prefixed.go
```diff
@@ -13,54 +13,59 @@ import (
...
 	"storj.io/storj/pkg/storj"
 )
-
-type prefixedObjStore struct {
-	store  objects.Store
-	prefix string
+// PrefixedObjStore  ...
```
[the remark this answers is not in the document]
jenlij: I will update these once I am certain I need to export the prefixedObjStore struct and related methods

pkg/storage/buckets/prefixed.go
jenlij: Yeah that is unexpected. I suspect it was because Buckets contained an object store.

pkg/storage/buckets/store.go
jenlij: Good catch, I will update all the comments once the code is sorted out

pkg/storage/buckets/store.go
jenlij: good idea! i'll make a temp method to do this

pkg/storage/buckets/store.go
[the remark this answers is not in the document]
egonelbre: The `pkg/storage/*` (with minor exceptions) should be under `storj/uplink` as well. So, `uplink/metainfo` is in the right place.

pkg/storage/objects/store.go
```diff
@@ -136,16 +136,16 @@ func (o *objStore) List(ctx context.Context, prefix, startAfter, endBefore storj
...
 		}
 	}
 
 	return items, more, nil
 }
 
-// convertMeta converts stream metadata to object metadata
-func convertMeta(m streams.Meta) Meta {
+// ConvertMeta converts stream metadata to object metadata
+func ConvertMeta(m streams.Meta) Meta {
```
thepaul: Would be nice if we could take this opportunity to rename this function to something like `StreamMetaToObjectMeta()`. Unless you think there's already enough going on and we should strictly open new PRs for that sort of thing.

pkg/storage/buckets/store.go
```diff
@@ -222,3 +248,69 @@ func convertMeta(ctx context.Context, m objects.Meta) (out Meta, err error) {
 
 	return out, err
 }
+
+// convertTime converts gRPC timestamp to Go time
+func convertTime(ts *timestamp.Timestamp) time.Time {
```
thepaul: I'm not entirely sure, but I think you might be able to avoid the need for this entirely by using a protobuf extension: `(gogoproto.stdtime) = true`. See pkg/pb/gc.proto for usage.

pkg/storage/segments/store.go
```diff
@@ -353,8 +353,8 @@ func CalcNeededNodes(rs *pb.RedundancyScheme) int32 {
 	return needed
 }
 
-// convertMeta converts pointer to segment metadata
-func convertMeta(pr *pb.Pointer) Meta {
+// ConvertMeta converts pointer to segment metadata
+func ConvertMeta(pr *pb.Pointer) Meta {
```
thepaul: And this one could be `PointerMetaToSegmentMeta()`.

pkg/storage/streams/store.go
```diff
@@ -38,8 +38,8 @@ type Meta struct {
 	Data       []byte
 }
 
-// convertMeta converts segment metadata to stream metadata
-func convertMeta(lastSegmentMeta segments.Meta, stream pb.StreamInfo, streamMeta pb.StreamMeta) Meta {
+// ConvertMeta converts segment metadata to stream metadata
```
thepaul: SegmentMetaToStreamMeta() maybe?

PR #1858, opened 2019-04-29
made inspector segment health care about healthy nodes

wthorp:
What: 
Only consider "healthy" nodes in inspector segment health check.

Why:
It strongly appears like inspector intends to measure what checker / repairer cares about.  Previously, inspector measured segment health only by what was "online", while checker also cares about if the node uptime, if it has failed too many audits, etc.  This change makes them care about the same thing.


wthorp:
Thanks for the heart @coyle.  When I tried to talk about this in chat (perhap a little too abstractly), I wasn't sure you were onboard.

PR #1859, opened 2019-04-29
Cbindings

aleitner:
What’s left for c-bindings unless something else pops up
- [ ] (wip) Cleanup TODOs
- [x] Remove duplicate test code
- [x] Add comments to functions
- [ ] Create README.md for C Developers
- [x] Make sure we are freeing everything that’s allocated
- [x] Create Map Getter and Setter functions for interacting with Object MetaData
- [x] Standardize variable and function naming conventions
- [x] Jenkins setup
- [ ] Crossplatform compilation
- [x] Write a smaller testing framework for c integration tests
- [ ] Whatever else we discover that we forgot about


lib/uplink/ext/common.go
```diff
@@ -29,9 +40,128 @@ func GetIDVersion(number C.uint, cErr **C.char) C.struct_IDVersion {
...
 
-	if err := lib.GoToCStruct(goIDVersion, cIDVersion); err != nil {
+	if err := GoToCStruct(goIDVersion, cIDVersion); err != nil {
 		*cErr = C.CString(err.Error())
 		return cIDVersion
 	}
 	return cIDVersion
 }
+
+func GoToCStruct(fromVar, toPtr interface{}) error {
```
egonelbre: Why do we need to user reflection for this? We shouldn't have too many structs needing the mapping?
There might be also simpler implementations for this.

TopperDEL:
What are the plans for the Output of this PR? Will there be different DLLs for different CPU-architectures?

go.mod
```diff
@@ -100,8 +100,9 @@ require (
 	github.com/vivint/infectious v0.0.0-20190108171102-2455b059135b
 	github.com/yuin/gopher-lua v0.0.0-20180918061612-799fa34954fb // indirect
 	github.com/zeebo/admission v0.0.0-20180821192747-f24f2a94a40c
-	github.com/zeebo/errs v1.1.0
+	github.com/zeebo/errs v1.1.1
 	github.com/zeebo/float16 v0.1.0 // indirect
+	github.com/zeebo/goof v0.0.0-20190312211016-1ee209ef0510
```
jtolds: lol oh no

lib/uplink/common.go
```diff
@@ -5,11 +5,13 @@ package uplink
 
 import (
 	"github.com/zeebo/errs"
+	"github.com/zeebo/goof"
 	monkit "gopkg.in/spacemonkeygo/monkit.v2"
 )
 
 var (
-	mon = monkit.Package()
+	Troop goof.Troop
```
jtolds: https://media.tenor.com/images/e94f37436c75ccb038a1f10d7cca5450/tenor.gif

lib/uplink/ext/uplink.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		*cErr = C.CString(err.Error())
+		return cUplink
+	}
+
+	return C.UplinkRef_t(structRefMap.Add(goUplink))
+}
+
+//export NewUplinkInsecure
+func NewUplinkInsecure(cErr **C.char) (cUplink C.UplinkRef_t) {
```
mniewrzal: why creating new constructors and not just create struct on C side with the same fields? later just reassign them to original struct

bryanchriswhite: This is an example where we need to keep a reference to a specific go struct instance. The go struct returned from `NewUplink` has methods on it and no exported fields, so there's no need for a c struct of that type and instead we use the go struct map to store that struct instance where `cUplink` is the key to that value in the map.

mniewrzal: Sorry, I my comment was unclear. I was talking about passing `uplink.Config` C binding version and have only one `NewUplink`. Something like ```func NewUplink(cUplinkConfig *C.UplinkConfig, cErr **C.char) (cUplink C.UplinkRef_t) {```

lib/uplink/ext/mapping.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (m *mapping) Get(x token) interface{} {
+	m.lock.Lock()
+	res := m.values[x]
+	m.lock.Unlock()
+
+	return res
+}
+
+func (m *mapping) Del(x token) {
```
mniewrzal: From what I saw its never used. Shouldn't this be called when something is closed?

lib/uplink/ext/uplink.go
bryanchriswhite: Oh, yes. We would do that but the entire uplink config is inside of a `Volatile` field which we're not supposed to reference in the bindings. So, for now, the default config is always used because `nil` is being passed for the config. This unfortunately means that in the testing context we can't change the peer ca whitelist option to make things work with test planet with the production `NewUplink` function. That's why this function is here. Blame @jtolds :p.

lib/uplink/ext/mapping.go
bryanchriswhite: good idea!

bryanchriswhite:
> What are the plans for the Output of this PR? Will there be different DLLs for different CPU-architectures?

Hey man, sorry nobody (incl. myself) got back to you on this! I'm not sure what the official plan is but I'm pretty confident that the long-term goal is to have support for as many OS/arch/languages within reason. To date afaik V3 has had windows support so I would assume that we would want the cbindings to also support windows.

As far as the output of the build toolchain; currently, we're building a shared object which the end user links into their program. I'm not personally familiar with how things work in that department in .NET or on windows in general but the `c-shared` buildmode we're using to build the shared object currently [looks like it can build dlls too](https://golang.org/doc/go1.10#compiler).

If you wanted to give this a try yourself you could try with this branch or master after the bindings are merged:
```
cd storj/lib/uplink/ext
go build -o uplink-cgo.dll -buildmode=c-shared .
```

bryanchriswhite:
@TopperDEL ^

lib/uplink/ext/common_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func TestCTests(t *testing.T) {
+	ctests, err := filepath.Glob(filepath.Join(CTestsDir, "*_test.c"))
+	require.NoError(t, err)
+
+	t.Run("all", func(t *testing.T) {
+		for _, ctest := range ctests {
+			t.Run(ctest, func(t *testing.T) {
+				t.Parallel()
```
mniewrzal: btw, for some reason, some tests are not failing when `t.Parallel()` is used here. not sure why yet
when its disabled:
```
            test_utils.go:106: /home/wywrzal/git/storj/lib/uplink/ext/c/tests/object_test.c:74:TestObject:FAIL: Expected 'testing data 123' Was 'testing data 123\x10'
                
                -----------------------
                1 Tests 1 Failures 0 Ignored 
                FAIL
                
            require.go:794: 
                	Error Trace:	test_utils.go:107
                	            				test_utils.go:124
                	            				common_test.go:42
                	Error:      	Received unexpected error:
                	            	exit status 1
                	Test:       	TestCTests/all//home/wywrzal/git/storj/lib/uplink/ext/c/tests/object_test.c
FAIL
FAIL	storj.io/storj/lib/uplink/ext	32.815s
lib/uplink/ext/tests.go:5: running "go": exit status 1
```

TopperDEL:
Thanks @bryanchriswhite for reaching out! I'm actively watching this PR, but I'm not a professional regarding C and Go. But I think this should work for me.
Thank you and Keep it going! :)

TopperDEL:
Is there a very simple function, that just Returns e.g. a Version number or Version string? If not, could you add one (at least for testing)? I have difficulties to know if my build of the dll worked.

TopperDEL:
@bryanchriswhite ^^

bryanchriswhite:
Part 1 merged (#2196)

lib/uplinkc/apikey.go
```diff
@@ -27,7 +27,7 @@ func parse_api_key(val *C.char, cerr **C.char) (apikeyHandle C.APIKeyRef_t) {
 func serialize_api_key(apikeyHandle C.APIKeyRef_t) *C.char {
 	apikey, ok := universe.Get(apikeyHandle._handle).(libuplink.APIKey)
 	if !ok {
-		return C.CString("")
+		return C.CString("invalid apikey")
```
egonelbre: This should be returned as an error not as the result.

bryanchriswhite: :+1:

bryanchriswhite:
part 2 merged (#2209)

lib/uplinkc/testdata/object_test.c
```diff
@@ -0,0 +1,149 @@
...
+
+#include "require.h"
+#include "uplink.h"
+#include "helpers2.h"
+
+void handle_project(ProjectRef project);
+
+int main(int argc, char *argv[]) {
+    ProjectOptions opts = {};
+    memset(&opts.key, '\0', 32);
```
egonelbre: Ah, note, the if you make the `opts = {0}` then the memset shouldn't be needed.

bryanchriswhite:
This work was merged incrementally over multiple separate PRs:
#2196 
#2209 
#2258 
#2260

PR #1862, opened 2019-04-30
add apikey library on top of macaroons

jtolds:
What: This adds a more user-friendly Storj V3 specific library on top of a macaroon framing library

Why: So we can start to do access control!


jtolds:
added the "Do Not Merge" label because i need to retarget this PR against master once https://github.com/storj/storj/pull/1537 is merged

pkg/macaroon/apikey.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, Error.Wrap(err)
+	}
+	return &APIKey{mac: mac}, nil
+}
+
+// Check makes sure that the key authorizes the provided action given the root
+// project secret and any possible revocations, returning an error if the action
+// is not authorized. 'revoked' is a list of either revoked heads or revoked
+// tails. Keep in mind that any key generated from a key with a revoked head
+// _or_ tail will also be considered revoked.
```
ifraixedes: just curious, why `_` are around `or`?

pkg/macaroon/types.proto
```diff
@@ -0,0 +1,41 @@
...
+
+message Action {
+  enum Operation {
+    READ = 0;
+    WRITE = 1;
+    LIST = 2;
+    DELETE = 3;
+  }
+
+  Operation op = 1;
```
ifraixedes: Just for my understanding, the Operation isn't required, then if it isn't set, it defaults to `READ`, right?

pkg/macaroon/apikey.go
thepaul: Looks like the author is using a common markup pattern for emphasis: "_or_". They want to emphasize that the revocation can be on the head OR the tail.

ifraixedes: OK, it's mostly what I thought, like mimicking markdown, however this won't be rendered by godoc.

pkg/macaroon/types.proto
jtolds: operation is required, but read seemed like a good default

pkg/macaroon/apikey.go
jtolds: took the comment out due to other changes requested by @rikysya

jtolds:
added expiration tests

pkg/macaroon/types.proto
jtolds: perhaps you're saying we should make an "INVALID" operation that is always rejected and make that the zero value?

ifraixedes: sorry, I rechecked protobuf docs and I realized that `required` doesn't apply for proto 3, I bore in mind the field rule from my previous experience with protobuf which was long time ago and I used the v2.

Before writing this response I've documented myself a bit more regarding protobuf v3, for hoping to make more sense on what I've written.

In protobuf v3 all the fields are optional, that means that the sender can send it or not; when they are not sent, the fields get the [default value according its type](https://developers.google.com/protocol-buffers/docs/proto3#default) and in enums is the first defined enum value.

The actions is required for the logic, so my preference is to explicitly define the operation rather than having a default one. I always see that's more straightforward to understand because the reader doesn't need to know the default value applied when it isn't specified. So, yes, I would prefer to have an _INVALID_ operation as default that will allow to know that the user actually forget to determine the operation and we can report an error about it.

Nonetheless, I agree that if we make a default valid operation, _READ_ is a good default.

pkg/macaroon/apikey_test.go
```diff
@@ -0,0 +1,134 @@
...
+		{notBeforeMinuteFromNow, &twoMinutesFromNow, true},
+		{notAfterMinuteAgo, &twoMinutesFromNow, false},
+	} {
+		err := test.keyToTest.Check(secret, Action{
+			Time: test.timestampToTest,
+		}, nil)
+		if test.authorized {
+			require.NoError(t, err, fmt.Sprintf("test #%d", i+1))
+		} else {
+			require.False(t, !ErrUnauthorized.Has(err), fmt.Sprintf("test #%d", i+1))
```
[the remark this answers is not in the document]
jtolds: the log output is better when the test fails. it was saying "expected true for test #7" but then test #7 above was set authorized = true. true on this line in the previous version meant that it was unauthorized, which was different than the true in the test table where true means authorized. so making this false made the semantics match

pkg/macaroon/types.proto
jtolds: oh man, yeah, every single protobuf i used in v2 i marked all the fields optional cause protobuf v2 required was badness.

okay, i'll make an invalid type explicitly. i agree with you

pkg/macaroon/types.proto
```diff
@@ -0,0 +1,42 @@
...
+  // to the given values
+  repeated bytes buckets = 10;
+  repeated bytes encrypted_path_prefixes = 11;
+
+  // if set, the validity time window
+  google.protobuf.Timestamp not_after = 20 [(gogoproto.stdtime) = true];
+  google.protobuf.Timestamp not_before = 21 [(gogoproto.stdtime) = true];
+}
+
+message Action {
```
zeebo: Why is this a protobuf? Is it ever serialized?

pkg/macaroon/apikey.go
```diff
@@ -0,0 +1,185 @@
...
+func (a *APIKey) Check(secret []byte, action Action, revoked [][]byte) error {
+	if !a.mac.Validate(secret) {
+		return ErrInvalid.New("macaroon unauthorized")
+	}
+
+	caveats := a.mac.Caveats()
+	if len(caveats) == 0 {
+		// make sure we always check at least one empty caveat
+		caveats = [][]byte{[]byte("")}
+	}
```
zeebo: This is non-obvious, whereas ensuring the action has a time directly is obvious.

pkg/macaroon/apikey.go
```diff
@@ -0,0 +1,185 @@
...
+// APIKey implements a Macaroon-backed Storj-v3 API key.
+type APIKey struct {
+	mac *Macaroon
+}
+
+// ParseAPIKey parses a given api key string and returns an APIKey if the
+// APIKey was correctly formatted. It does not validate the key.
+func ParseAPIKey(key string) (*APIKey, error) {
+	data, version, err := base58.CheckDecode(key)
+	if err != nil || version != 0 {
```
coyle: why do you want the version to be `0` ?

pkg/macaroon/types.proto
jtolds: it does not need to be a protobuf, i'll change it

pkg/macaroon/apikey.go
jtolds: fair

pkg/macaroon/apikey.go
jtolds: if we ever change the version, this code doesn't know what to do, so i want it to fail

PR #1863, opened 2019-04-30
Mobile proof of concept

egonelbre:
Proof of concept for creating Java bindings.

* It uses gomobile to create the necessary jar/aar files.
* Currently upload works, however 
* download seems to skip decryption.

This is hacked together and assume there are tons of bugs, however it does upload.

Note `gomobile` doesn't work with go modules, so you need to either use https://github.com/storj/gospace or `go mod vendor`.

To create something from scratch:

```
1. install gomobile (https://godoc.org/golang.org/x/mobile/cmd/gomobile)
2. install android studio
3. install android ndk

	ensure $ANDROID_HOME is set

4. gomobile bind -target android

5. create new project in android studio
6. change app/build.gradle
	implementation fileTree(dir: 'libs', include: ['*.jar', '*.aar'])
7. AndroidManifest.xml

	<uses-permission android:name="android.permission.INTERNET" />
	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ protos/google/*
 
 *.coverprofile
 *.log
+*~*
 
 /release/
 *.swp
@@ -49,5 +50,3 @@ small-upload-testfile
 small-download-testfile
 /bin
 resource.syso
-
-mobile-out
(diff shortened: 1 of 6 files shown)


egonelbre:
Continued in #1918

PR #1865, opened 2019-04-30
add boltDB batching for Put operation, add benchmark test

JessicaGreben:
Jira [V3-1493](https://storjlabs.atlassian.net/browse/V3-1493)

**What:**
This PR modifies the `storage/boltdb/client.go` `client.Put()` method to use boltDB `db.Batch()` method instead of boltDB `db.Update()`.

This PR also adds a benchmark test that compares boltDB Put operations in the following 3 scenarios:
1. using the current boltDB `db.Update` operation
2. turning fsync off and running boltDB `db.Update` operation
3. using boltDB `db.Batch` instead of `db.Update`
4. using boltDB `db.Batch` **and** turning off fsync

**Why:**
The boltDB method `db.Update` creates a new transaction and commits the transaction right away which writes to disk and fsyncs. Kademlia RoutingTable calls `db.Update` frequently so this creates a lot of writes to disk slowing things down.  Here we test if using boltDB `db.Batch` can help us improve performance. We chose to not handle the fsync ourselves since it caused additional complicated logic without a big enough increase in performance. Benchmark test results:

```
~/storj/storage/boltdb - (jg/bolt-benchmark) $ go test -bench=.
goos: darwin
goarch: amd64
pkg: storj.io/storj/storage/boltdb
BenchmarkClientWrite-12               	      10	 157192126 ns/op
BenchmarkClientNoSyncWrite-12         	      30	  44419786 ns/op
BenchmarkClientBatchWrite-12          	    1000	   2011001 ns/op
BenchmarkClientBatchNoSyncWrite-12    	    1000	   1849665 ns/op
PASS
ok  	storj.io/storj/storage/boltdb	5.202s
```


storage/boltdb/client.go
```diff
@@ -67,6 +68,7 @@ func NewShared(path string, buckets ...string) ([]*Client, error) {
 	if err != nil {
 		return nil, Error.Wrap(err)
 	}
+	db.NoSync = true
```
egonelbre: There are other things that may use bolt.

cmd/storagenode/main.go
```diff
@@ -136,6 +136,18 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		return errs.New("Error starting master database on storagenode: %+v", err)
 	}
 
+	// Sync routing table database every 1s
+	ticker := time.NewTicker(time.Second)
```
egonelbre: Do we have handling for corruption.

cmd/storagenode/main.go
```diff
@@ -136,6 +136,18 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
...
+	defer ticker.Stop()
+	kdb, ndb := db.RoutingTable()
+	go func() {
+		for {
+			<-ticker.C
+			kdb.Sync()
```
egonelbre: This logically races with closing the database.

storage/boltdb/client_test.go
```diff
@@ -11,11 +11,146 @@ import (
...
+			}
+		}
+	}
+	kdb.db.Sync()
+	b.Logf("\n b.N: %d, TxStats Write: %v, WriteTime: %v\n", b.N, kdb.db.Stats().TxStats.Write, kdb.db.Stats().TxStats.WriteTime)
+}
+
+func BenchmarkClientBatchWrite(b *testing.B) {
+	// setup db
+	tempdir, err := ioutil.TempDir("", "storj-bolt")
```
egonelbre: Use `testcontext` to create directories.

storage/boltdb/client_test.go
```diff
@@ -11,11 +11,146 @@ import (
...
+	if err != nil {
+		fmt.Println("err:", err)
+	}
+	defer func() { _ = os.RemoveAll(tempdir) }()
```
egonelbre: Don't ignore deletion errors, they help us find bugs.

storage/redis/client.go
```diff
@@ -117,6 +117,12 @@ func (client *Client) Delete(key storage.Key) error {
 	return nil
 }
 
+// Sync writes data to disk
+func (client *Client) Sync() error {
+	// TODO: this satisfies the storage.KeyValueStore interface, implement if needed.
+	return Error.New("Sync not implemented")
```
egonelbre: So when someone uses a different backend implementation, then it will fail? The approach here should be to return `nil` because the implementation is always synced.

cmd/storagenode/main.go
JessicaGreben: I agree this part looks out of place.  I would prefer this syncing code be in the same place where `db.NoSync = true` is set so that it makes sense why we need this.  

I don't know the best way to add the `NoSync` setting and its related logic.  Seems like its  messy here and Im not sure if there is a way to avoid that, other than removing the NoSync code altogether and letting BoltDB handle the syncing. That is my preferred option, let BoltDB do the sync.

storage/boltdb/client.go
JessicaGreben: I mostly added this `db.NoSync` setting to start a discussion about if we should be handling the syncing logic in ourselves.  Personally I think we should avoid this setting and instead keep letting BoltDB handling the syncing.  While this setting does seem to improve performance a little bit, I'm not convinced it is worth the increased risk and complication added to the code base.

Anyhow, about your comment...tt looks like this function `NewShared` is currently only used for the Routing table creation. So if we decide that we want to handle syncing ourselves, then syncing logic needs to be added to the routing table code for bootstrap and SA (as it is for SN below).

If we do choose to keep this `db.NoSync = true` in here, I don't know how we can ensure that future uses of this function handle syncing correctly. I could add comment, but it still seems risky.

storage/boltdb/client.go
```diff
@@ -112,13 +115,34 @@ func (client *Client) view(fn func(*bolt.Bucket) error) error {
...
+	return err
+}
+
+// PutAndCommit adds a key/value to BoltDB and writes it to disk.
+func (client *Client) PutAndCommit(key storage.Key, value storage.Value) error {
+	if key.IsZero() {
+		return storage.ErrEmptyKey.New("")
+	}
+
+	return client.db.Update(func(tx *bolt.Tx) error {
```
thepaul: Is there a particular reason not to stick with `client.update()` here, as the other Client methods do? If no, probably best to switch back to avoid confusion. If yes, this deserves an explanatory comment, and if the same reason applies to the other methods, maybe those should be changed as well.

storage/boltdb/client_test.go
```diff
@@ -28,6 +30,7 @@ func TestSuite(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create db: %v", err)
 	}
+	store.db.MaxBatchDelay = 1 * time.Millisecond
```
thepaul: It's a little worrying anytime we have to run the test environment with a configuration that is so different from what is in prod. Are there any ways to avoid it without entirely breaking test performance? Maybe doing more test operations at the same time, so that batches would naturally tend to complete more quickly?

storage/boltdb/client_test.go
```diff
@@ -140,3 +149,155 @@ func BenchmarkSuiteLong(b *testing.B) {
...
+func BenchmarkClientBatchWrite(b *testing.B) {
+	// setup db
+	ctx := testcontext.New(b)
+	defer ctx.Cleanup()
+	dbfile := ctx.File("testbolt.db")
+	dbs, err := NewShared(dbfile, "kbuckets", "nodes")
+	if err != nil {
+		fmt.Printf("failed to create db: %v\n", err)
+	}
+	kdb := dbs[0]
```
thepaul: Why are the deferred `dbs[x].Close()` calls not needed here, as they are in the two previous functions?

storage/boltdb/client_test.go
```diff
@@ -140,3 +149,155 @@ func BenchmarkSuiteLong(b *testing.B) {
...
 }
+
+func BenchmarkClientWrite(b *testing.B) {
+	// setup db
+	ctx := testcontext.New(b)
+	defer ctx.Cleanup()
+	dbfile := ctx.File("testbolt.db")
+	dbs, err := NewShared(dbfile, "kbuckets", "nodes")
+	if err != nil {
+		fmt.Printf("failed to create db: %v\n", err)
```
thepaul: This (and all following `fmt.Printf()` calls in the benchmark functions) should be a `b.Fatalf()` call. The benchmark results are not valid if we have failed to create, close, or modify the DBs in the expected way. Plus if these are happening during automated testing, we really want to know about it.

storage/boltdb/client_test.go
```diff
@@ -140,3 +149,155 @@ func BenchmarkSuiteLong(b *testing.B) {
...
+	kdb := dbs[0]
+
+	// benchmark test: execute 1000 Put operations where each call to `PutAndCommit` does the following:
+	// 1) create a BoltDB transaction (tx), 2) execute the db operation, 3) commit the tx which writes it to disk.
+	for n := 0; n < b.N; n++ {
+		for i := 0; i < 1000; i++ {
+			key := storage.Key(fmt.Sprintf("testkey%d", i))
+			value := storage.Value("testvalue")
+
+			err := kdb.PutAndCommit(key, value)
```
thepaul: BenchmarkClientWrite() and BenchmarkClientNoSyncWrite() would be more meaningful comparisons if they also used the same WaitGroup/goroutine pattern as BenchmarkClientBatchWrite() and BenchmarkClientBatchNoSyncWrite(). It might even be worth pulling out the commonalities between them all into a single runner function that takes a callback argument for doing the actual appropriate operation.

storage/boltdb/client_test.go
JessicaGreben: yeah i can remove this if its not ideal for testing.

storage/boltdb/client_test.go
JessicaGreben: oops I will add

storage/boltdb/client_test.go
JessicaGreben: hm ok let me give that a shot

storage/boltdb/client.go
JessicaGreben: i thought the way it was written before was confusing.  For example:

```
// PutAndCommit adds a value to the provided key in boltdb, returning an error on failure.
func (client *Client) PutAndCommit(key storage.Key, value storage.Value) error {
	if key.IsZero() {
		return storage.ErrEmptyKey.New("")
	}

	return client.update(func(bucket *bolt.Bucket) error {
		return bucket.Put(key, value)
	})
}

func (client *Client) update(fn func(*bolt.Bucket) error) error {
	return Error.Wrap(client.db.Update(func(tx *bolt.Tx) error {
		return fn(tx.Bucket(client.Bucket))
	}))
}
```

I felt like we could remove the `update` method all together, but now that Im looking at it more, I guess its written like that to abstract away the boltDB method calls. I'm not sure. Anyhow, I can revert it to keep it consistent.

thepaul: Yeah, I'm not sure what the reasoning is behind the `update()` method. Probably to abstract those particular parts of Bolt world. But yeah, consistency helps readability a lot on its own. We should either use `update()` here or take it out in the other methods (and if you want to do that, a separate commit would probably be preferable).

JessicaGreben: i think i will revert the changes in this PR, then submit a different PR to remove that layer of abstraction i.e `update`, etc methods

storage/boltdb/client_test.go
```diff
@@ -140,3 +141,189 @@ func BenchmarkSuiteLong(b *testing.B) {
...
+		for i := 0; i < 1000; i++ {
+			key := storage.Key(fmt.Sprintf("testkey%d", i))
+			value := storage.Value("testvalue")
+
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				err := kdb.PutAndCommit(key, value)
+				if err != nil {
+					b.Fatal("PutAndCommit Nosync err:", err)
```
egonelbre: benchmark methods cannot be called from a goroutine it's a race.

JessicaGreben: oops I will fix that in another PR since this is merged. thanks egon.

PR #1875, opened 2019-05-01
Add image for versioncontrol server

brimstone:


brimstone:
The image builds, but I'll need @stefanbenten's help testing that it's working right.

PR #1876, opened 2019-05-01
Add Identity to Version Control Server

stefanbenten:
What: Adding allowed version to Version Control Server, to restrict older binaries from start.

Why: Currently the check for the version fails with an error message, which is not good :)

Thanks for submitting a PR!


versioncontrol/peer.go
```diff
@@ -92,6 +93,9 @@ func New(log *zap.Logger, config *Config) (peer *Peer, err error) {
 	gatewayVersions := strings.Split(config.Versions.Gateway, ",")
 	peer.Versions.Gateway, err = version.StrToSemVerList(gatewayVersions)
 
+	identityVersions := strings.Split(config.Versions.Identity, ",")
```
coyle: should you add tests ?

versioncontrol/peer.go
```diff
@@ -92,6 +93,9 @@ func New(log *zap.Logger, config *Config) (peer *Peer, err error) {
...
+	peer.Versions.Identity, err = version.StrToSemVerList(identityVersions)
```
coyle: check error ?

PR #1878, opened 2019-05-01
defaults now commented out

wthorp:
What: 
YAML files are generated with everything commented out
This caused some readability issues, so I added spaces between items

Why:
To make our configs more like other configs - in particular, to make it more obvious when a value had been changed from the default.

https://storjlabs.atlassian.net/browse/V3-1632

### DIFF
--- a/pkg/process/exec_conf.go
+++ b/pkg/process/exec_conf.go
@@ -99,17 +99,20 @@ func saveConfig(flagset *pflag.FlagSet, outfile string, overrides map[string]int
 		if overriddenValue != nil {
 			value = fmt.Sprintf("%v", overriddenValue)
 		}
+		//print usage info
 		if f.Usage != "" {
 			fmt.Fprintf(w, "# %s\n", f.Usage)
 		}
-		fmt.Fprintf(w, "%s: ", k)
+		//print commented key (beginning of value assignement line)
+		fmt.Fprintf(w, "# %s: ", k)
+		//print value (remainder of value assignement line)
 		switch f.Value.Type() {
 		case "string":
 			// save ourselves 250+ lines of code and just double quote strings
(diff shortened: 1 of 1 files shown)


pkg/process/exec_conf.go
```diff
@@ -99,17 +99,20 @@ func saveConfig(flagset *pflag.FlagSet, outfile string, overrides map[string]int
 		if overriddenValue != nil {
 			value = fmt.Sprintf("%v", overriddenValue)
 		}
+		//print usage info
 		if f.Usage != "" {
 			fmt.Fprintf(w, "# %s\n", f.Usage)
 		}
-		fmt.Fprintf(w, "%s: ", k)
+		//print commented key (beginning of value assignement line)
+		fmt.Fprintf(w, "# %s: ", k)
```
jtolds: well, wait, k might not be the default here. i think we need an if check of some kind. when saveConfig is called, we're given overrides and so on. only the flagset knows what the true default is

wthorp: k?  k is the key - variable name... but yeah... this is all very Viper specific code that gets sketchier over time.

jtolds: oh shoot, you're right, i meant the line might not want to be commented out indiscriminately - it depends on if the value is actually different than the code default

mobyvb:
Failing because the default identity dir in storagenode init() is different than the one set in storj-sim (https://github.com/storj/storj/blob/master/cmd/storagenode/main.go#L90)
But even though the default is different, the settings by storj-sim are commented out for storagenodes. My guess is that the same issue is the case for satellite. We could maybe add a special case where identity paths are explicitly not commented out. Unless there is a way to really determine whether it is the correct default or not

PR #1884, opened 2019-05-02
The Satellite hangs when misconfigured

Fadila82:
What: 
The satellite setup command should tell the user if it is unable to set a sane value and ask the user to configure one.
The satellite run command should return an error if it is unable to start and exit.

Jira task: https://storjlabs.atlassian.net/browse/V3-1465
Why:
The satellite doesn't tell me how it's misconfigured, it just hangs with the last line being `Setting up metainfo`

Thanks for submitting a PR!


satellite/metainfo/config.go
```diff
@@ -18,7 +18,7 @@ const (
 
 // Config is a configuration struct that is everything you need to start a metainfo
 type Config struct {
-	DatabaseURL          string      `help:"the database connection string to use" default:"bolt://$CONFDIR/pointerdb.db"`
+	DatabaseURL          string      `help:"the database connection string to use" default:"bolt://$CONFDIR/pointerdb.db" mandatory:"true"`
```
ifraixedes: Should `mandatory` invalidate the `default` value?

As far I understand if the `default` isn't an empty string, the `mandatory` check won't apply because the default value is used and it isn't empty.

internal/sync2/cycle.go
```diff
@@ -124,8 +143,11 @@ func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error)
...
 // Close closes all resources associated with it.
 func (cycle *Cycle) Close() {
-	cycle.Stop()
-	<-cycle.stop
+	cycle.initialize()
+	if cycle.runningState.IsRunning() {
+		cycle.Stop()
+		<-cycle.stop
+	}
 	close(cycle.control)
```
ifraixedes: there is no need to close the channel, because doing so, `Close` panics if it's called more than once.

In master, the cals after the first one, hang forever, but I think that's a bug.

Fadila82: I've removed the call to close. And this PR should remove the hanging bug :-)

ifraixedes: yes, I've realized about that it was going to be fixed if `close` got removed.

internal/sync2/cycle.go
```diff
@@ -124,9 +143,10 @@ func (cycle *Cycle) Run(ctx context.Context, fn func(ctx context.Context) error)
 
 // Close closes all resources associated with it.
 func (cycle *Cycle) Close() {
-	cycle.Stop()
-	<-cycle.stop
-	close(cycle.control)
+	if cycle.runningState.IsRunning() {
```
wthorp: Could multiple calls to IsRunning() here race?  Should that logic be moved into here to hold the Lock() longer?

pkg/process/exec_conf.go
```diff
@@ -103,6 +111,9 @@ func saveConfig(flagset *pflag.FlagSet, outfile string, overrides map[string]int
 			fmt.Fprintf(w, "# %s\n", f.Usage)
 		}
 		fmt.Fprintf(w, "%s: ", k)
+		if readBoolAnnotation(f, "mandatory") && (value == "") {
```
wthorp: Are you sure this works as intended on types where the default values is not empty string (EG: bool)?

wthorp: Also, does this logic work only during 'setup' (YAML creation), or always?

Fadila82: It works only during setup from what I understand from the code:
		if readBoolAnnotation(f, "setup") {
			continue
		}

Fadila82: And the variable "value" is always a string, it is obtained this way: 
		value := f.Value.String()

internal/sync2/cycle.go
Fadila82: We do need a mutex as the tests fail otherwise with a data race. Now for the Lock, I don't really know.

wthorp: IsRunning returns a snapshot of state not protected by the mutex.  Cycle.Stop() will run twice if Close() is called from two different threads

pkg/process/exec_conf.go
wthorp: It should technically be possible to run our programs without first running setup.  all you'd do is supply all the configuration parameter via the command line instead of via YAML.  I don't think your code overs that use case.

wthorp: It should technically be possible to run our programs without first running setup. all you'd do is supply all the configuration parameter via the command line instead of via YAML. I don't think your code overs that use case.

ifraixedes:
@Fadila82 rebase from master, there have been changes in `internal/sync2/cycle.go` so your changes may not be needed anymore.

internal/sync2/cycle.go
egonelbre: Close shouldn't be called concurrently in the first place.

kaloyan-raev:
The issue with satellite hanging is already fixed in master with https://github.com/storj/storj/pull/1936. We've discussed in the standup that we don't really need the rest of the changes in this PR. So I am closing it.

PR #1888, opened 2019-05-03
hopefully sensible satellite defaults

wthorp:
https://storjlabs.atlassian.net/browse/V3-1634

What: 
Use sensible defaults, as pulled in different from the Mars satellite
Also better ErrNotEnoughNodes debugging, because I needed it.

Why:
Part of series of efforts to have satellite configs better managed.

These changes come in part from the mars satellite yaml.  Taking these configs into account, I was able to reduce that yaml to a set of rationally required / divergent entries.  As for Docker, I don't see anything in the docker-compose.yaml settings that seems universally applicable.


pkg/overlay/config.go
```diff
@@ -39,7 +39,7 @@ type NodeSelectionConfig struct {
 	AuditSuccessRatio float64       `help:"a node's ratio of successful audits" releaseDefault:"0.4" devDefault:"0"` // TODO: update after beta
 	AuditCount        int64         `help:"the number of times a node has been audited to not be considered a New Node" releaseDefault:"500" devDefault:"0"`
 	NewNodePercentage float64       `help:"the percentage of new nodes allowed per request" default:"0.05"` // TODO: fix, this is not percentage, it's ratio
-	MinimumVersion    string        `help:"the minimum node software version for node selection queries" default:""`
+	MinimumVersion    string        `help:"the minimum node software version for node selection queries" releaseDefault:"v0.10.1" devDefault:""`
```
brimstone: This version is gonna be weird to maintain.

wthorp: Whoever created this configuration probably didn't intend for it to be set in code.  Then again, if there is a cogent value, we should probably set it, right?

pkg/kademlia/config.go
```diff
@@ -23,7 +23,7 @@ var (
 // Config defines all of the things that are needed to start up Kademlia
 // server endpoints (and not necessarily client code).
 type Config struct {
-	BootstrapAddr        string        `help:"the Kademlia node to bootstrap against" default:""`
+	BootstrapAddr        string        `help:"the Kademlia node to bootstrap against" default:"bootstrap.storj.io:8888"`
```
jtolds: we should totally not have the devDefault be this production value. the bootstrap node will get slammed with tests and storj-sims

pkg/overlay/config.go
jtolds: yeah i think this is okay to leave unset

pkg/kademlia/config.go
coyle: Agreed, do we have a reasonable dev default ?

stefanbenten: For Dev Default its best to leave it empty. Aka `releaseDefault:"bootstrap.storj.io:8888" devDefault:""`

PR #1890, opened 2019-05-03
filter duplicate node IPs

cam-a:
When requesting nodes for upload, select for distinct last seen IP addresses to increase durability and make it a bit less profitable for cheaters to run more than one node
https://storjlabs.atlassian.net/browse/V3-1635

Thanks for submitting a PR!


pkg/discovery/service.go
```diff
@@ -143,6 +143,8 @@ func (discovery *Discovery) refresh(ctx context.Context) error {
 			return ctx.Err()
 		}
 
+		// TODO: add method to update ping.LastIp? Or is it done above in cache.Put?
+
```
stefanbenten: This should be done automatically with the FetchInfo function, which is then used in UpdateNodeInfo.

stefanbenten:
Just missing the IP Filter in the selection process in the overlaycache.go file

satellite/satellitedb/dbx/satellitedb.dbx.postgres.sql
```diff
@@ -97,6 +97,7 @@ CREATE TABLE irreparabledbs (
 CREATE TABLE nodes (
 	id bytea NOT NULL,
 	address text NOT NULL,
+	last_ip text NOT NULL,
```
egonelbre: No changes to old migration files are allowed.

pkg/transport/transport.go
```diff
@@ -130,6 +138,15 @@ func (transport *Transport) WithObservers(obs ...Observer) Client {
 	return tr
 }
 
+func getIP(target string) (string, error) {
+	addrElements := strings.Split(target, ":")
```
egonelbre: This is not the proper way to split an IP address. This will fail with ipv6

pkg/transport/transport.go
```diff
@@ -130,6 +138,15 @@ func (transport *Transport) WithObservers(obs ...Observer) Client {
...
+	ipAddr, err := net.ResolveIPAddr("ip", addrElements[0])
```
egonelbre: Do we have any DNS redirect loop concerns here?

pkg/transport/transport.go
cam-a: I now see `net.SplitHostPort()`. Do you know if ipv6 targets will already have brackets around the host portion?

egonelbre: https://en.wikipedia.org/wiki/IPv6#Addressing, but we probably can assume, yes.

pkg/transport/transport.go
cam-a: I asked about this during standup. @thepaul said the go resolver might take care of this, so JT said we might not be concerned

cam-a: Would it be feasible to write a test case for this?

satellite/satellitedb/testdata/postgres.v18.sql
```diff
@@ -0,0 +1,244 @@
...
+	segmentdetail bytea NOT NULL,
+	pieces_lost_count bigint NOT NULL,
+	seg_damaged_unix_sec bigint NOT NULL,
+	repair_attempt_count bigint NOT NULL,
+	PRIMARY KEY ( segmentpath )
+);
+CREATE TABLE nodes (
+	id bytea NOT NULL,
+	address text NOT NULL,
+    last_ip text NOT NULL,
```
mobyvb: WHY IS THIS NOT ALIGNED

satellite/satellitedb/overlaycache.go
```diff
@@ -102,24 +102,39 @@ func (cache *overlaycache) queryFilteredNodes(ctx context.Context, excluded []st
...
+		type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
+		uptime_ratio, total_audit_count, audit_success_count, total_uptime_count,
+		uptime_success_count
+		FROM (SELECT id, type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
+			uptime_ratio, total_audit_count, audit_success_count, total_uptime_count, uptime_success_count,
+			Row_number() OVER(PARTITION BY last_ip ORDER BY RANDOM()) rn
+			FROM nodes
+			`+safeQuery+safeExcludeNodes+`) n
+		WHERE rn = 1
+		ORDER BY RANDOM()
```
mobyvb: I don't think this ORDER BY is necessary since you are already randomizing things in the section where you partition by last ip

satellite/satellitedb/testdata/postgres.v18.sql
cam-a: I tried fixing it. Looks like something to do with viewing on github

satellite/satellitedb/overlaycache.go
```diff
@@ -60,7 +62,23 @@ func (cache *overlaycache) SelectStorageNodes(ctx context.Context, count int, cr
 		args = append(args, v.Major, v.Major, v.Minor, v.Minor, v.Patch)
 	}
 
-	return cache.queryFilteredNodes(ctx, criteria.Excluded, count, safeQuery, args...)
+	var totalNodes []*pb.Node
+	for i := 0; i < 3; i++ {
```
mobyvb: why do we need to iterate 3 times? If we pass `count` into `cache.queryFilteredNodes`, it should return that many nodes, and if it doesn't, calling it again won't return more nodes, right?

satellite/satellitedb/overlaycache.go
```diff
@@ -85,41 +103,91 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
...
```
mobyvb: same question. In what case would calling `cache.queryFilteredNodes` a second time result in more results?

pkg/transport/transport_test.go
```diff
@@ -252,3 +254,18 @@ func TestDialNode_BadServerCertificate(t *testing.T) {
 		assert.Contains(t, err.Error(), "not signed by any CA in the whitelist")
 	})
 }
+
+func TestGetIP(t *testing.T) {
+	conn, err := net.Dial("tcp", "bootstrap.storj.io:8888")
+	require.NoError(t, err)
```
mobyvb: This test will be dependent on an external server. I don't know if it is a good idea to use non-local hosts here. If bootstrap.storj.io goes down, it means our tests will fail (locally and on jenkins)

pkg/overlay/selection_test.go
```diff
@@ -216,3 +219,119 @@ func TestNodeSelection(t *testing.T) {
...
+					AuditCount:        0,
+					NewNodePercentage: 0.5,
+					OnlineWindow:      time.Hour,
+					DistinctIP:        false,
+				},
+			},
+		}
+
+		// This sets a reputable audit count for a certain number of nodes.
+		for i, node := range planet.StorageNodes {
```
mobyvb: You shouldn't need to have this logic here. This double loop was intended to allow us to test all sorts of edge cases with reputation selection. For duplicate IPs, you just want to test selecting new vs. selecting reputable, so you can `UpdateStats` on half the nodes, and call those ones "reputable" (`preferences.AuditCount = 1`), and don't update the other half and call those ones "new".

Then you can set duplicate/non duplicate IPs accordingly to ensure you are covering the cases you care about.

satellite/satellitedb/overlaycache.go
cam-a: So the subquery gets us a number of reputable nodes equal to count. We then SELECT DISTINCT ON last_ip from that set of reputable nodes to filter out duplicate IPs. If we filtered any entries out, the number of nodes < count

pkg/transport/transport_test.go
cam-a: Hmm, yeah. Maybe we should remove this

satellite/satellitedb/overlaycache.go
```diff
@@ -85,41 +103,91 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
...
+		}
+		if len(totalNodes) == count {
+			break
+		}
+	}
+	return totalNodes, nil
+}
+
+func (cache *overlaycache) queryFilteredNodes(ctx context.Context, exclude overlay.Exclude, count int, safeQuery string, distinctIP bool, args ...interface{}) (_ []*pb.Node, err error) {
+	switch t := cache.db.DB.Driver().(type) {
```
stefanbenten: Is there no way around having a selection for each driver here?

satellite/satellitedb/overlaycache.go
stefanbenten: I agree, thats not a good idea.

satellite/satellitedb/overlaycache.go
stefanbenten: I agree with @mobyvb here, we should try to construct the query differently then.

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -172,6 +172,9 @@ kademlia.operator.wallet: ""
 # a node's ratio of successful audits
 # overlay.node.audit-success-ratio: 0.4
 
+# require distinct IPs when choosing nodes for upload
+# overlay.node.distinct-ip: true
```
stefanbenten: Please dont forget to give the Satellite Operators / Release Team/Manager a heads up, once this gets merged!

satellite/satellitedb/overlaycache.go
```diff
@@ -60,7 +62,23 @@ func (cache *overlaycache) SelectStorageNodes(ctx context.Context, count int, cr
...
-	return cache.queryFilteredNodes(ctx, criteria.Excluded, count, safeQuery, args...)
+	var totalNodes []*pb.Node
+	for i := 0; i < 3; i++ {
+		nodes, err := cache.queryFilteredNodes(ctx, criteria.Exclude, count-len(totalNodes), safeQuery, criteria.DistinctIP, args...)
+		if err != nil {
+			return nil, err
+		}
+		for _, n := range nodes {
+			totalNodes = append(totalNodes, n)
+			criteria.Exclude.Nodes = append(criteria.Exclude.Nodes, n.Id)
```
stefanbenten: These two step here should should be avoided.. We should not have to run queries multiple times.

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
```diff
@@ -88,6 +88,7 @@ CREATE TABLE irreparabledbs (
 CREATE TABLE nodes (
 	id BLOB NOT NULL,
 	address TEXT NOT NULL,
+	last_ip TEXT NOT NULL,
```
egonelbre: Shouldn't we get that information from the pointers rather than from here?

stefanbenten: No, the IPs can change quite frequently for DDNS Hosts @egonelbre

egonelbre: Pointers store node id and we can get the address information from elsewhere. You know what actually. This shouldn't be stored in the database in the first place. Let's pull it into a separate struct... this would keep the load off the database.

stefanbenten: That wont work for uploads. For uploads we dont have any pointers to look at.

satellite/satellitedb/overlaycache.go
cam-a: This query is about twice as fast as the old one, `Row_number() OVER(PARTITION BY last_ip ORDER BY RANDOM()`. @jtolds preferred speed with potential need for retrying

satellite/satellitedb/overlaycache.go
cam-a: `DISTINCT ON` is invalid with sqlite

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
egonelbre: Yeah, I agree. But, keeping it in-memory probably makes more sense anyways.

satellite/satellitedb/overlaycache.go
cam-a: We could also increase `count` by some percentage to offset the possibility of duplicates requiring a subsequent query?

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
stefanbenten: I totally agree with that. the DB is way to overloaded with simple requests that are not critical

cam-a: So query for reputable nodes, check the returned node IDs against a map of lastIPs, throw out duplicates, exclude all nodeIDs with duplicate IPs from the next query and repeat?

satellite/satellitedb/overlaycache.go
stefanbenten: @egonelbre Objections?

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
cam-a: Actually, what's the best data structure to use for storing the lastIPs?

satellite/satellitedb/overlaycache.go
jtolds: heh, cam has evaluated every possible option for this query i can think of, learning deep, dark postgres-specific secrets in the meantime. this query in the common case is much faster, and the potential for needing retries is rare. we have not been able to find another solution that provides this level of overall speed. my vote is we continue forward, unless someone can suggest a faster query that doesn't have this problem. there hasn't been a completely perfect solution found so far but this one seems to have the least downsides believe it or not

stefanbenten: Egon suggested to move this out of the DB and do it with in RAM maps. I agree with him on that.
This would remove the need for retries to optimize performance, right?

cam-a: Would that eliminate the need for retries? If last_ip isn't in the DB, how can we know that there are no duplicates in the set we retrieve? Or on the other hand, we have a map of nodeIDs and their last IPs; we choose x nodeIDs with unique IPs from that list and query for the reputable ones from that list. What if not enough of them are reputable?

pkg/overlay/cache.go
```diff
@@ -75,9 +75,10 @@ type NodeCriteria struct {
 	AuditSuccessRatio  float64
 	UptimeCount        int64
 	UptimeSuccessRatio float64
-	Excluded           []storj.NodeID
+	Exclude            Exclude
```
egonelbre: Any  reason just not to have `ExcludedNodes` and `ExcludedIPs` ?

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
egonelbre: Ah seems like I misunderstood what this `last_ip` meant, so it should be fine.

egonelbre: As for storing the last_ip you can use []byte in the binary form... see https://golang.org/pkg/net/#IP

satellite/satellitedb/overlaycache.go
```diff
@@ -137,6 +205,77 @@ func (cache *overlaycache) queryFilteredNodes(ctx context.Context, excluded []st
...
+	if len(exclude.Nodes) > 0 {
+		safeExcludeNodes = ` AND id NOT IN (?` + strings.Repeat(", ?", len(exclude.Nodes)-1) + `)`
+		for _, id := range exclude.Nodes {
+			args = append(args, id.Bytes())
+		}
+	}
+
+	safeExcludeIPs := ""
+	if len(exclude.IPs) > 0 {
+		safeExcludeIPs = ` AND last_ip NOT IN (?` + strings.Repeat(", ?", len(exclude.IPs)-1) + `)`
```
egonelbre: Did you also try `last_ip <> any(values(?,?,?,?,?,?`?

satellite/satellitedb/overlaycache.go
```diff
@@ -137,6 +205,77 @@ func (cache *overlaycache) queryFilteredNodes(ctx context.Context, excluded []st
...
 }
 
+func (cache *overlaycache) postgresQueryNodes(ctx context.Context, exclude overlay.Exclude, count int, safeQuery string, distinctIP bool, args ...interface{}) (_ []*pb.Node, err error) {
+	if count == 0 {
+		return nil, nil
+	}
+
+	safeExcludeNodes := ""
+	if len(exclude.Nodes) > 0 {
+		safeExcludeNodes = ` AND id NOT IN (?` + strings.Repeat(", ?", len(exclude.Nodes)-1) + `)`
```
egonelbre: Did you also try `last_ip <> any(values(?,?,?,?,?,?`?

satellite/satellitedb/overlaycache.go
egonelbre: The more nodes we have the less likely we will have the need for the extra nodes... however it might be still useful to always query some amount of more.

I think the more performant solution would be to keep everything in memory and using combiner queue flush things to database... depending on the query make it wait for sync to database. In essence, yes, it would remove the need for retries because we can write whatever code we like.

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
egonelbre: Also postgres supports network types natively https://www.postgresql.org/docs/9.1/datatype-net-types.html, which probably would be faster than []byte/text

satellite/satellitedb/overlaycache.go
cam-a: I did. It appeared to be on average a bit slower, however that could be due to the table I'm running the query against. I have the query plans if you'd like to compare them

pkg/overlay/cache.go
cam-a: `cache.FindStorageNodesWithPreferences()` queries once for a number of new nodes, and again for reputable nodes. In order to not select duplicates of new node IPs, I pass in excluded IPs. I merged excluded node IDs and IPs to avoid adding another argument to the function, but I'm fine with separating them

satellite/satellitedb/overlaycache.go
cam-a: Is that something you would like done in this PR, or should we make another ticket?

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
cam-a: Is this possible with DBX?

pkg/overlay/cache.go
jenlij: I think separating them is cleaner if it's not too much work for you.

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
jenlij: @cam yes it believe it would be of type blob

cam-a: I mean the postgres type `inet`

satellite/satellitedb/overlaycache.go
jenlij: Let's make another ticket for that suggestion

satellite/satellitedb/dbx/satellitedb.dbx.sqlite3.sql
cam-a: I'll do blob for now

jenlij: No dbx does not support inet

Cam: @jenlij I think you may have the wrong Cam 😉

satellite/satellitedb/overlaycache.go
stefanbenten: @cam-a Lets get your PR merged with this approach and we monitor the overhead and performance with it.
As Jen already stated, lets make a new ticket/PR for the additional node request.

satellite/satellitedb/overlaycache.go
```diff
@@ -87,41 +103,91 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
...
+		FROM (SELECT id, type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
+			uptime_ratio, total_audit_count, audit_success_count, total_uptime_count, uptime_success_count,
+			Row_number() OVER(PARTITION BY last_ip ORDER BY RANDOM()) rn
+			FROM nodes
+			`+safeQuery+safeExcludeNodes+safeExcludeIPs+`) n
+		WHERE rn = 1
+		ORDER BY RANDOM()
+		LIMIT ?`), args...)
+	} else {
+		rows, err = cache.db.Query(cache.db.Rebind(`SELECT id,
```
stefanbenten: I feel like we should move the if distinctIP Check outside of the function and have the "normal" query being a seperate function.
Ala 
* QueryNodes
* PostgresQueryNodesDistinct
* SQLiteQueryNodesDistinct

Based on the bool we then can use the same query for  SQLite and Postgres and only switch based on the driver if we want distinct nodes.

satellite/satellitedb/overlaycache.go
```diff
@@ -139,6 +205,77 @@ func (cache *overlaycache) queryFilteredNodes(ctx context.Context, excluded []st
...
+		FROM (SELECT id,
+			type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
+			uptime_ratio, total_audit_count, audit_success_count, total_uptime_count,
+			uptime_success_count
+			FROM nodes
+			`+safeQuery+safeExcludeNodes+safeExcludeIPs+`
+			ORDER BY RANDOM()
+			LIMIT ?) n`), args...)
+	} else {
+		rows, err = cache.db.Query(cache.db.Rebind(`SELECT id,
```
stefanbenten: See above comment

satellite/satellitedb/overlaycache.go
```diff
@@ -87,48 +109,205 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
...
+			for _, n := range moreNodes {
+				nodes = append(nodes, n)
+				criteria.ExcludedNodes = append(criteria.ExcludedNodes, n.Id)
+				criteria.ExcludedIPs = append(criteria.ExcludedIPs, n.LastIp)
+			}
+			if len(nodes) == count {
+				break
+			}
+		}
+	} else {
```
stefanbenten: Could we return in the if directly and save the need for the else branch?

satellite/satellitedb/overlaycache.go
```diff
@@ -62,10 +62,32 @@ func (cache *overlaycache) SelectStorageNodes(ctx context.Context, count int, cr
...
```
stefanbenten: See below comment

PR #1895, opened 2019-05-06
If/v3 1541: Change where the encryption key is being stored for uplink 

ifraixedes:

Thanks for submitting a PR!

### DIFF
--- a/cmd/uplink/cmd/root.go
+++ b/cmd/uplink/cmd/root.go
@@ -96,23 +96,23 @@ func (c *UplinkFlags) GetProject(ctx context.Context) (*libuplink.Project, error
 	cfg.Volatile.MaxInlineSize = c.Client.MaxInlineSize
 	cfg.Volatile.MaxMemory = c.RS.MaxBufferMem
 
-	uplink, err := c.NewUplink(ctx, cfg)
+	uplk, err := c.NewUplink(ctx, cfg)
 	if err != nil {
 		return nil, err
 	}
 
 	encryptionKey, err := c.Enc.LoadKey()
-	if err != nil {
+	if err != nil && err != uplink.ErrKeyFilepathEmpty {
 		return nil, err
 	}
 
 	opts := &libuplink.ProjectOptions{}
 	opts.Volatile.EncryptionKey = &encryptionKey
 
-	project, err := uplink.OpenProject(ctx, satelliteAddr, apiKey, opts)
+	project, err := uplk.OpenProject(ctx, satelliteAddr, apiKey, opts)
 
 	if err != nil {
-		if err := uplink.Close(); err != nil {
+		if err := uplk.Close(); err != nil {
 			fmt.Printf("error closing uplink: %+v\n", err)
 		}
 	}
--- a/cmd/uplink/cmd/root.go
+++ b/cmd/uplink/cmd/root.go
@@ -102,7 +102,7 @@ func (c *UplinkFlags) GetProject(ctx context.Context) (*libuplink.Project, error
 	}
 
 	encryptionKey, err := c.Enc.LoadKey()
-	if err != nil && err != uplink.ErrKeyFilepathEmpty {
+	if err != nil {
 		return nil, err
 	}
 
--- a/cmd/uplink/cmd/root.go
+++ b/cmd/uplink/cmd/root.go
@@ -101,11 +101,13 @@ func (c *UplinkFlags) GetProject(ctx context.Context) (*libuplink.Project, error
 		return nil, err
 	}
 
-	opts := &libuplink.ProjectOptions{}
+	encryptionKey, err := c.Enc.LoadKey()
+	if err != nil {
+		return nil, err
+	}
 
-	encKey := new(storj.Key)
-	copy(encKey[:], c.Enc.Key)
-	opts.Volatile.EncryptionKey = encKey
+	opts := &libuplink.ProjectOptions{}
+	opts.Volatile.EncryptionKey = &encryptionKey
 
 	project, err := uplink.OpenProject(ctx, satelliteAddr, apiKey, opts)
 
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -189,7 +189,8 @@ Some things to try next:
 		return nil
 	}
 
-	return process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), nil)
+	return process.SaveConfigWithAllDefaults(
+		cmd.Flags(), filepath.Join(setupDir, process.DefaultConfFilename), nil)
 }
 
 // ApplyDefaultHostAndPortToAddrFlag applies the default host and/or port if either is missing in the specified flag name.
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -20,6 +20,7 @@ import (
 	"storj.io/storj/internal/fpath"
 	"storj.io/storj/pkg/cfgstruct"
 	"storj.io/storj/pkg/process"
+	"storj.io/storj/pkg/storj"
 )
 
 var (
@@ -69,10 +70,10 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	var (
-		encKeyFilepath = filepath.Join(setupDir, ".enc.key")
-		encKey         []byte
-		override       = map[string]interface{}{
-			"enc.key-filepath": encKeyFilepath,
+		encryptionKeyFilepath = filepath.Join(setupDir, ".encryption.key")
+		encryptionKey         []byte
+		override              = map[string]interface{}{
+			"enc.key-filepath": encryptionKeyFilepath,
 		}
 	)
 	if !setupCfg.NonInteractive {
@@ -138,7 +139,7 @@ Please enter numeric choice or enter satellite address manually [1]: `)
 		if err != nil {
 			return err
 		}
-		encKey, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+		encryptionKey, err = terminal.ReadPassword(int(os.Stdin.Fd()))
 		if err != nil {
 			return err
 		}
@@ -160,17 +161,18 @@ Please enter numeric choice or enter satellite address manually [1]: `)
 			return err
 		}
 
-		if !bytes.Equal(encKey, repeatedEncKey) {
+		if !bytes.Equal(encryptionKey, repeatedEncKey) {
 			return errs.New("encryption passphrases doesn't match")
 		}
 
-		if len(encKey) == 0 {
+		if len(encryptionKey) == 0 {
 			_, err = fmt.Println("Warning: Encryption passphrase is empty!")
 			if err != nil {
 				return err
 			}
 		} else {
-			err = SaveEncryptionKey(encKey, encKeyFilepath)
+			key := storj.NewKey(encryptionKey)
+			err = SaveEncryptionKey(key, encryptionKeyFilepath)
 			if err != nil {
 				return err
 			}
@@ -261,34 +263,33 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 
 // SaveEncryptionKey saves the key in a new file which will be stored in
 // filepath.
-// It returns an error if the directory doesn't exist, the file already exists
-// or there is an I/O error.
-func SaveEncryptionKey(key []byte, filepath string) (err error) {
-	f, err := os.OpenFile(filepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
+//
+// key is a pointer for avoiding having several copies of the key in memory.
+//
+// It returns an error if the key is nil, the directory doesn't exist, the
+// file already exists or there is an I/O error.
+func SaveEncryptionKey(key *storj.Key, filepath string) (err error) {
+	if key == nil {
+		return errors.New("key cannot be nil")
+	}
+
+	file, err := os.OpenFile(filepath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return errors.New("directory path doesn't exist")
+			return fmt.Errorf("directory path doesn't exist. %+v", err)
 		}
 
 		if os.IsExist(err) {
-			return errors.New("file key already exists")
+			return fmt.Errorf("file key already exists. %+v", err)
 		}
 
 		return err
 	}
 
 	defer func() {
-		if err == nil {
-			err = f.Close()
-		} else {
-			_ = f.Close()
-		}
+		err = errs.Combine(err, file.Close())
 	}()
 
-	_, err = f.Write(key)
-	if err != nil {
-		return err
-	}
-
-	return f.Chmod(0400)
+	_, err = file.Write(key[:])
+	return err
 }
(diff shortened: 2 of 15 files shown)


ifraixedes:
The last concern addressed in the original PR (https://github.com/storj/storj/pull/1782) are copied below.
The goal is to follow up with them in this new PR because the previous one had too many conversations and many changes through the time it has been open and the browser was struggling to render all the HTML elements that the GH comments create.

---

@jtolds 

> in terms of being able to touch the disk, i would say it's safe to assume anything under /cmd/* can read files off disk, but anything under /lib/uplink/* or anything it depends on (including /uplink/*) should not assume disk access is possible

I'm confused with `/uplink`; I got answer that such package should be merged into others, so I don't know until what point now it's worth it to worry about it.

Anyway, being this the case, please confirm me if:

1. Make a change in the current implementation because this branch has added 2 methods to one type present in the `/upconfig` package which access to the disk. Because those 2  methods are used by `/cmd/uplink` and /cmd/gateway`, I would need that you tell me where I can put that shared logic or the other possibility is to make a copy into both packages.
2. We leave as it's in this branch and we open an issue for future refactoring.

---

@egonelbre 
> I've been aiming with the design such that only few things touch io (disk / net) directly: ...

That will nice to have because it will clarify the _io_ access and avoid mess.


---

There are 3 `TODOs` in this branch that they should be away before merging; being away is that I'm waiting to tell what to do with each one:

1. Just removed.
2. Create separated issues for them and removed.
3. Do what they require in this branch (this option will potentially delay to merge the main goal of this branch for a while).

You can easy find those TODOs comments in the source just executing `grep "TODO: if/v3-1541" -rn .` (or whatever be the link in Windows) from the root of the repository.

pkg/storj/encryption.go
```diff
@@ -131,6 +131,14 @@ const (
 	NonceSize = 24
 )
 
+// NewKey creates a new key from a passphrase
+func NewKey(passphrase []byte) *Key {
+	// TODO: if/v3-1541 This function should return an error when len(passphrase) == 0
```
kaloyan-raev: I don't think we need such an error. The Key type has the IsZero method, so we should be valid to create a empty/zero key too.

pkg/storj/encryption.go
```diff
@@ -131,6 +131,14 @@ const (
...
+	key := &Key{}
+	copy(key[:], passphrase)
```
kaloyan-raev: Perhaps, it is a good moment to improve this logic here. It's a backward incompatible change, so better do it now than later.

`copy(key[:], passphrase)` has the disadvantage that only the 32 first characters of the passphrase matter. This means that the following two will produce the same encryption key:
- ThisIsARelativelyLongButStillReasonablePassphrase
- ThisIsARelativelyLongButStillReasonableEncryptionKey

It would be better if we take the SHA-256 hash of the passphrase, which will produce `[32]byte`, which is exactly the same size as `storj.Key`.

kaloyan-raev: btw, we have a nice `pkcrypto.SHA256Hash` helper function to do this.

pkg/miniogw/integration_test.go
```diff
@@ -79,7 +79,9 @@ func TestUploadDownload(t *testing.T) {
 
 	// keys
 	uplinkCfg.Client.APIKey = "apiKey"
-	uplinkCfg.Enc.Key = "encKey"
+	const humanReadableEncKey = "encKey"
+	encKey := new(storj.Key)
+	copy(encKey[:], humanReadableEncKey)
```
kaloyan-raev: Shouldn't we use `storj.NewKey` here too?

pkg/process/exec_conf.go
```diff
@@ -30,6 +30,9 @@ import (
 	"storj.io/storj/internal/version"
 )
 
+// DefaultConfFilename is the default filename used for storing a configuration.
+const DefaultConfFilename = "config.yaml"
```
kaloyan-raev: nit: we usually use "cfg" for short of configuration. So maybe, name this `DefaultCfgFilename`.

pkg/storj/encryption.go
kaloyan-raev: @jtolds what do you think about this? does it make sense to do it now, or are we going to introduce completely different approach with the design doc for encryption key managment?

uplink/config.go
```diff
@@ -41,10 +43,84 @@ type RSConfig struct {
...
+// It returns an error if:
+//
+// * KeyFilepath field is an empty string. Err value will be ErrKeyFilepathEmpty
+// * The file doesn't exist.
+// * There is an I/O error.
+//
+// TODO: if/v3-1541 refactor cache the key after the firs read. The problem is
+// with pkg/cfgstruct which is inflexible on accepting unexported keys nor
+// several different types as a type of storj.Key, [32]byte, etc.
+func (cfg *EncryptionConfig) LoadKey() (key storj.Key, err error) {
```
kaloyan-raev: The return value should be a pointer, i.e. `*storj.Key` instead of `storj.Key`.

`storj.Key` is a 32-byte array, not a slice. We want to reference it always via a pointer to avoid many copies of the encryption key in the memory as it is sensitive data.

uplink/config.go
```diff
@@ -41,10 +43,84 @@ type RSConfig struct {
...
+		if os.IsNotExist(err) {
+			return storj.Key{}, Error.New("not found key file %q", cfg.KeyFilepath)
+		}
+
+		return storj.Key{}, Error.Wrap(err)
+	}
+
+	defer func() { err = errs.Combine(err, file.Close()) }()
+
+	if _, err := file.Read(key[:]); err != nil && err != io.EOF {
```
kaloyan-raev: After reading the file into a `[]byte` slice, we should call `storj.NewKey` to create the key from the slice.

uplink/config.go
```diff
@@ -41,10 +43,84 @@ type RSConfig struct {
...
+	return key, nil
+}
+
+// SaveKey saves the key in the file indicated by KeyFilepath field.
+//
+// key is a pointer for avoiding having several copies of the key in memory.
+//
+// It returns an error if KeyFilepath is empty, key IsZero return true, the
+// directory doesn't exist, the file already exists or there is an I/O error.
+func (cfg *EncryptionConfig) SaveKey(key *storj.Key) error {
```
kaloyan-raev: `storj.Key` is a 32-byte fixed size array, while the human readable encryption key can be of any size. This only works at the moment because we copy the human-readable encryption key into to 32-byte array and call it the encryption. This is very likely the change in the near future (if not even with this PR - see my other comment).

Therefore, I think the SaveKey method should take the human-readable encryption key as the param.

pkg/storj/encryption.go
ifraixedes: I added this after knowing that the current `cmd/uplink` doesn't work with a zero key value, because `lib/uplink` rejects it.

pkg/storj/encryption.go
ifraixedes: It makes sense to me, let's see what @jt says about it

uplink/config.go
ifraixedes: I had that question in the closed PR https://github.com/storj/storj/pull/1782/files/a77bb67c07b6f0dbde7d88329939659dbb3549f1#discussion_r278839646

I was waiting for the reply, however I forgot to moved to this PR.

pkg/miniogw/integration_test.go
ifraixedes: yes, it should, I don't know why `NewKey` wasn't used, maybe it didn't exist at that time or basically missed with one of the upstream updates.

uplink/config.go
ifraixedes: Yes, only if we `Save` the human readable key to a file (as you commented below) rather than the actual key.

pkg/storj/encryption.go
jtolds: i definitely thought we were hashing passphrases and we absolutely should for sure. let's do it

jtolds: actually, let's use PBKDF2 instead of sha256 though. that way people can't "rainbow table" a decryption.

jtolds: are you sure we aren't already doing that? i really thought we were doing PBKDF2 or Scrypt or something

ifraixedes: at the time that I created the PR which was replaced by this one, I am sure, but right now I will have to check with the last upstream updates

ifraixedes: I discover a few months ago these kind of algorithms for generating keys.

I knew that there was another one called Argon2 and I wanted to know the difference between it and PBKDF2 however I'm no  sure that the sources that I found are reliable but I wanted to mention it here.

Just to mention one of them https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016

Go has the implementation of both 

* Argon2: https://godoc.org/golang.org/x/crypto/argon2
* PBKDF2: https://godoc.org/golang.org/x/crypto/pbkdf2

I'm wondering if there is any special reason to use PBKDF2 rather than Argon2, for example being FIPS compatibility?

uplink/config.go
ifraixedes: I'm good with this, I just want to confirm that `SaveKey` will take the passphrase as a parameter and will call `storj.NewKey`(with the change commented in https://github.com/storj/storj/pull/1895#discussion_r282827259) and store the key in a file.

Is that right?

egonelbre:
> I'm confused with /uplink; I got answer that such package should be merged into others, so I don't know until what point now it's worth it to worry about it.

`/uplink` is the actual uplink implementation dealing with all the details. `/lib/uplink` is the library that tries to provide compatibility and simplified access to the code. So `/uplink` will be needed to organize the upload/download - protocol logic.

ifraixedes:
> /uplink is the actual uplink implementation dealing with all the details. /lib/uplink is the library that tries to provide compatibility and simplified access to the code. So /uplink will be needed to organize the upload/download - protocol logic.

I'm still not very clear.
Anyway, regarding to the changes here, where the save `LoadKey` and `SaveKey` should be implemented? currently they are in `uplink` but we don't want I/O on that package.

If I'm not mistaken, we'll need to load the _save_ the key in `cmd/uplink`, but _load_ it from `cmd/uplink` and `cmd/gateway`

ifraixedes:
I'm not going to carry on with these changes, I will start over, because things have been better specified and with the last changes in upstream is making this branch so messy than starting it from scratch.

I leave the PR open to clarify the las things before starting again.

uplink/config.go
ifraixedes: Only pointers will be use in the new implementation.

pkg/storj/encryption.go
ifraixedes: Do we allow an empty passphrase?

cmd/uplink/cmd/root.go
```diff
@@ -96,21 +96,23 @@ func (c *UplinkFlags) GetProject(ctx context.Context) (*libuplink.Project, error
...
+	encryptionKey, err := c.Enc.LoadKey()
+	if err != nil {
+		return nil, err
+	}
 
-	encKey := new(storj.Key)
-	copy(encKey[:], c.Enc.Key)
-	opts.Volatile.EncryptionKey = encKey
+	opts := &libuplink.ProjectOptions{}
+	opts.Volatile.EncryptionKey = &encryptionKey
```
jtolds: just a heads up that this will need to happen this next sprint: https://github.com/storj/storj/pull/1854

pkg/storj/encryption.go
jtolds: we should not allow an empty passphrase. if someone wants an empty passphrase really want they want is to disable encryption entirely. if we're dealing with passphrases they should not be empty

cmd/uplink/cmd/setup.go
```diff
@@ -124,12 +142,13 @@ Please enter numeric choice or enter satellite address manually [1]: `)
...
 			return errs.New("API key cannot be empty")
 		}
+		override["api-key"] = apiKey
 
 		_, err = fmt.Print("Enter your encryption passphrase: ")
 		if err != nil {
 			return err
 		}
-		encKey, err := terminal.ReadPassword(int(os.Stdin.Fd()))
+		encryptionKey, err = terminal.ReadPassword(int(os.Stdin.Fd()))
```
jtolds: nit: storj.NewKey takes a passphrase, so maybe this variable should be passphrase? fine to leave as-is

pkg/storj/encryption.go
jtolds: nope! argon2 seems fine. you pick what you'd like to use

jtolds:
> I'm not going to carry on with these changes, I will start over, because things have been better specified and with the last changes in upstream is making this branch so messy than starting it from scratch.

Shoot, sorry @ifraixedes! This has been a hairy review process!

> I'm confused with /uplink; I got answer that such package should be merged into others, so I don't know until what point now it's worth it to worry about it.
> Anyway, being this the case, please confirm me if:
> 1. Make a change in the current implementation because this branch has added 2 methods to one type present in the /upconfig package which access to the disk. Because those 2 methods are used by /cmd/uplink and /cmd/gateway`, I would need that you tell me where I can put that shared logic or the other possibility is to make a copy into both packages.
> 2. We leave as it's in this branch and we open an issue for future refactoring.

I'm also confused about the future goals for /uplink. I erroneously believed earlier that /uplink's EncryptionConfig type was used by /lib/uplink, but it appears to not be. I am totally fine with leaving it as is and doing something about future refactoring.

Sorry again about this PR's length!

pkg/storj/encryption.go
thepaul: We've been talking in other places about how "passphrase" needs to be differentiated from "key material". Users might want to have a high-entropy key and encrypt that with a passphrase, as opposed to deriving the key from the passphrase via some KDF. Or they might want a high-entropy key stored in the clear on their system because they trust their system security and want to have passphrase-less access. Or they might have a hardware token further confusing the situation.

The question about empty passphrases seems to contain an invalid assumption, given that.

pkg/storj/encryption.go
thepaul: I was also pretty sure we were using pbkdf2. Are we really not?

ifraixedes: @jtolds fine, I already put in place pbkdf2, I was just curious and have the chance to know something new that I didn't know or find. Thanks!

@thepaul I've never seen using any kind of these algorithms in this part of the code, so it may be use but not in this part.

ifraixedes:
Closed in favor of https://github.com/storj/storj/pull/1967

**PLEASE DO NOT delete the branch**, I will do it once this feature gets merged or the reviews of PR which replaced this one are OK to forget about this one, but meanwhile leave it because it has some parts which where approached differently.

ifraixedes:
Closed in favor of https://github.com/storj/storj/pull/1967

**PLEASE DO NOT delete the branch**, I will do it once this feature gets merged or the reviews of PR which replaced this one are OK to forget about this one, but meanwhile leave it because it has some parts which where approached differently.

pkg/storj/encryption.go
ifraixedes: I'm not sure how to apply this to the scope of the issue.
The issue has the goal of only moving the encryption key from the `config.yaml` to a separated file nothing else.

`cmd/uplink` was already asking for a "passphrase", but that has a wrong name, because it's the encryption key itself.

My new PR #1967 has changed the `cmd/uplink` to also treat the "passphrase" for not being the key, it's used for generating a key through PBKDF2; hence the "passphrase" can be forgotten because it isn't used for anything else.

We want to go further with this topic, but I don't think that all belongs to this issue, because there is already another issue to have write a design doc for the key management.

ifraixedes: I hope that we start to see the scope of the issue, because otherwise, I suggest to leave this issue aside, write a design doc with all the workflow that they have for being agreed  and then do the implementation of it.

PR #1897, opened 2019-05-06
Refactor accountingdb interface

jenlij:
What: 
1. Splits Accounting.DB interface into Accounting.StoragenodeAccounting and Accounting.ProjectAccounting. 
2. Renames accounting_raws to storagenode_storage_tallies and updates corresponding methods.

Why: 
1. Decouple Storagenode and Bucket/Project accounting for simplification of accounting interface
2. Accounting_raws no longer holds bw data. Renames model to maintain consistency with bandwidth and bucket schemas.


pkg/accounting/db.go
```diff
@@ -46,30 +44,34 @@ type Rollup struct {
...
-	SaveAtRestRaw(ctx context.Context, latestTally time.Time, created time.Time, nodeData map[storj.NodeID]float64) error
-	// GetRaw retrieves all raw tallies
-	GetRaw(ctx context.Context) ([]*Raw, error)
-	// GetRawSince retrieves all raw tallies since latestRollup
-	GetRawSince(ctx context.Context, latestRollup time.Time) ([]*Raw, error)
-	// GetStoragenodeBandwidthSince retrieves all storagenode_bandwidth_rollup entires since latestRollup
-	GetStoragenodeBandwidthSince(ctx context.Context, latestRollup time.Time) ([]*StoragenodeBandwidthRollup, error)
-	// SaveRollup records raw tallies of at rest data to the database
+// StoragenodeAccounting stores information about bandwidth and storage usage for storage nodes
+type StoragenodeAccounting interface {
```
JessicaGreben: just an idea,  could we call this `StoragenodeDB`, then since its in the accounting package it would be referred to as `accounting.StoragenodeDB`.  Though Im not sure if calling it `Storagenode` would be confusing and not descriptive enough, rather im just brainstorming how to make these names shorter.

jenlij: That's a good point, I like StoragenodeDB.

thepaul: That name would make me think it was meant to be used _by_ a storage node.

jenlij: oh true...

PR #1898, opened 2019-05-06
v3-1660: Create a test that breaks when satellite config file has changed

ifraixedes:
What: Create a dead simple test which warn when the satellite configuration is changed.

Why: Devops team wants to be warned when the configuration file has suffered a change, to add whatever is needed for the deployment of the new release, so any PR sent should report to the owner about any change on the satellite configuration and update the lock file after notifying the devops team about the changes.

Thanks for submitting a PR!


ifraixedes:
> I think these are the dev settings, and that we need the release settings.

Yes, I didn't bear in mind the `defaults` global flag.

I updated the lock file and the script to use `--defaults release`.

scripts/test-satellite-cfg-change.sh
```diff
@@ -0,0 +1,24 @@
...
+  rm "$TESTDATA_DIR/config.yaml"
+}
+trap cleanup EXIT
+
+satellite --config-dir "$TESTDATA_DIR" --defaults release setup > /dev/null
+
+
+diff "$TESTDATA_DIR/satellite-config.yaml.lock" "$TESTDATA_DIR/config.yaml"
+if [[ $? != 0 ]]; then
+  echo "Notify the Devops and PM when this test fails so they can plan for changing it in the release process before fixing it to merge your PR"
```
wthorp: Can you include directions on how to update 'satellite-config.yaml.lock' here as well?

wthorp: (Possibly adding a makefile task to make things easy)

ifraixedes: Added a script and Makefile target to update the config lock file.

wthorp:
`This also means that when we run make test-update-satellite-cfg-lock, these two flags will be incorrect and user specific.`. Doh!  You're totally right and this was a bogus request from me.

wthorp:
Ivan, per Moby's comments, you can probably remove the script I foolishly asked for, and just have say "please update scripts/testdata/satellite-config.yaml.lock" in the error message.  I'll go ahead an approve this as is, though.

ifraixedes:
I'll check that downside that I didn't detected in my machine and fixed in another PR

PR #1901, opened 2019-05-06
Open a new port on satellite for admin GUI

VinozzZ:
What: 
Create a new private port on Satellite for marketing team to have access to the configurations for various credits offers
Why:
We are using VPN to restrict access to this private port so that only people who are on our VPN can access the page
Thanks for submitting a PR!


satellite/peer.go
```diff
@@ -509,6 +517,22 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 		)
 	}
 
+	{ // setup referral offers
+		log.Debug("Setting up offers")
```
egonelbre: Don't use global loggers.

satellite/peer.go
```diff
@@ -509,6 +517,22 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
...
+		referralConfig := config.Referral
+
+		peer.Referral.Listener, err = net.Listen("tcp", referralConfig.Address)
+		if err != nil {
+			return nil, errs.Combine(err, peer.Close())
+		}
+
+		peer.Referral.Endpoint = offersweb.NewServer(
```
egonelbre: Wondering whether we should have `admin.Server` instead of offersweb... a more general purpose admin interface. I feel like we might get more of these admin panels.

The question then is, how will the permissions be managed. Or whether it makes more sense to combine this with the `console` itself?

egonelbre: Note, I think this might be fine solution for now, but I'm not sure whether there's a nicer solution that doesn't take much longer, but saves us time in the future.

VinozzZ: I added a middleware for checking request address

mniewrzal:
@VinozzZ can you add why/what description to make code review easier?

VinozzZ:
> @VinozzZ can you add why/what description to make code review easier?

@mniewrzal Yes, I just added the description

satellite/marketing/marketingweb/server.go
```diff
@@ -0,0 +1,83 @@
...
+	//TODO: serve a 404 page
+}
+
+// Run starts the server that host admin web app and api endpoint
+func (s *Server) Run(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
+	var group errgroup.Group
+	group.Go(func() error {
+		<-ctx.Done()
+		return s.server.Shutdown(nil)
```
ifraixedes: wrap the error returned by `Shutdown` with `Error.Wrap`

satellite/marketing/marketingweb/server.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+	group.Go(func() error {
+		defer cancel()
+		return s.server.Serve(s.listener)
```
ifraixedes: wrap the error returned by `Serve` with `Error.Wrap`

satellite/marketing/marketingweb/server.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+
+	return group.Wait()
+}
+
+// Close closes server and underlying listener
+func (s *Server) Close() error {
+	return s.server.Close()
```
ifraixedes: wrap the error returned by `Close` with `Error.Wrap`

brimstone:
This doesn't handle any requests yet, will it before this PR is merged?

VinozzZ:
> This doesn't handle any requests yet, will it before this PR is merged?

Would you prefer to have the service portion complete and merge this PR last?

brimstone:
I guess? Either that or gated behind a feature flag until it's ready to use.

satellite/peer.go
VinozzZ: I checked in with @jtolds and @brimstone. I think for now we will rely on the devops to restrict access to this port through VPN. In that case, the requests won't be coming in from localhost

satellite/marketing/marketingweb/server.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/gorilla/mux"
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+// Error is satellite marketing error type
+var (
+	Error       = errs.Class("satellite marketing error")
+	_, fp, _, _ = runtime.Caller(0)
```
rikysya: May be we should move this code to `storj-sim` and use config values for static locations? Because this may not work with binaries, it will return path where the file was during build. On the other hand `with storj-sim` we can more or less assume that it is built from source code and static files are in place

fuskovic: This is the code already in `storj-sim/network.go` which is similar to what I have in the server file for setting fp. Also I don't believe it returns the path on build but on invocation as stated in the godoc : https://golang.org/pkg/runtime/#Caller correct me if I'm missing something please.

		`// TODO: find source file, to set static path
		_, filename, _, ok := runtime.Caller(0)
		if !ok {
			return nil, errs.Combine(processes.Close(), errs.New("no caller information"))
		}
		storjRoot := strings.TrimSuffix(filename, "/cmd/storj-sim/network.go")`

egonelbre: Yeah, we shouldn't definitely use it in a global, for prod, it should come from a flag. For storj-sim, it might be fine.

fuskovic: @rikysya I just made sure passing 0 to the runtime.Caller() func works the way I expected by creating a test file with only that function and printing the return value by building and executing the binary. 

No matter where you run the binary it will print the absolute path to the file invoking the function. If you duplicate the above process you will see that it's not returning the path on build as you suggested but on invocation like the go-doc describes.

@egonelbre I'll add a flag to storj-sim for prod as you suggested.

fuskovic: @rikysya @egonelbre I pushed the requested changes.

satellite/marketing/marketingweb/server.go
```diff
@@ -0,0 +1,115 @@
...
+	server := Server{
+		log:      logger,
+		config:   config,
+		listener: listener,
+	}
+
+	logger.Sugar().Debugf("Starting Marketing Admin UI on %s...", server.listener.Addr().String())
+	fs := http.FileServer(http.Dir(server.config.StaticDir))
+	mux := mux.NewRouter()
+	mux.Handle("/static/", http.StripPrefix("/static",fs))
```
VinozzZ: Should we check to see if the `StaticDir` is passed in before we pass it into the mux?

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -157,6 +157,9 @@ kademlia.operator.wallet: ""
 # uri which is used when retrieving new access token
 # mail.token-uri: ""
```
VinozzZ: I think we also need to add that new config for test data @fuskovic

satellite/marketing/marketingweb/server.go
fuskovic: done

scripts/testdata/satellite-config.yaml.lock
fuskovic: done

PR #1904, opened 2019-05-06
Gateway setup interactive

coyle:
What: 

Why:

Thanks for submitting a PR!


cmd/gateway/main.go
```diff
@@ -32,6 +35,7 @@ import (
 type GatewayFlags struct {
 	Identity          identity.Config
 	GenerateTestCerts bool `default:"false" help:"generate sample TLS certs for Minio GW" setup:"true"`
+	Interactive       bool `help:"enable|disable interactive mode" default:"true" setup:"true"`
```
mniewrzal: With `uplink` we decide to add `non-interactive` flag. It's a little bit weird when to disable bool flag you need `--interactive-false`.
https://github.com/storj/storj/blob/b0f67820f683bedd8a98a0d4a4bac09ef96a4c91/cmd/uplink/cmd/root.go#L26

cmd/gateway/main.go
```diff
@@ -260,6 +276,163 @@ func (flags GatewayFlags) NewGateway(ctx context.Context, ident *identity.FullId
...
+	if err != nil {
+		return nil
+	}
+
+	return nil
+
+}
+
+// ApplyDefaultHostAndPortToAddr applies the default host and/or port if either is missing in the specified address.
+func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, error) {
```
mniewrzal: most probably this needs to be moved somewhere and shared with uplink cmd

cmd/gateway/main.go
```diff
@@ -260,6 +268,48 @@ func (flags GatewayFlags) NewGateway(ctx context.Context, ident *identity.FullId
...
+	encKey, err := cfgstruct.PromptForEncryptionKey()
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	overrides["satellite-addr"] = satelliteAddress
+	overrides["api-key"] = apiKey
+	overrides["enc.key"] = encKey
+
+	err = process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), overrides)
```
bryanchriswhite: I thought we don't want to be saving all defaults for better backwards compatibility.

pkg/cfgstruct/utils.go
```diff
@@ -0,0 +1,156 @@
...
+			satelliteAddress = satellites[0]
+		} else {
+			return "", err
+		}
+	}
+
+	// TODO add better validation
+	if satelliteAddress == "" {
+		return "", errs.New("satellite address cannot be empty")
+	} else if len(satelliteAddress) == 1 {
```
bryanchriswhite: Does "add better validation" include handling the outstanding case where `len(satelliteAddress) != 1`? Should that case be an error, and/or perhaps the condition should be `satelliteAddress <= len(satellites)`?

pkg/cfgstruct/utils.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// address is host:
+	return net.JoinHostPort(addressParts[0], defaultPort), nil
+}
+
+// PromptForSatelitte handles user input for a satelitte address to be used with wizards
+func PromptForSatelitte(cmd *cobra.Command) (string, error) {
+	_, err := fmt.Print(`
+	Pick satellite to use:
```
bryanchriswhite: I feel like this shouldn't be hard-coded; does it make sense to build this programatically?

mniewrzal: This will be indented:
image

coyle: what do you suggest ? config file, api endpoint something else ?

pkg/cfgstruct/utils.go
coyle: < 1 is checked by checking `satelliteAddress == ""`
> 1 is handled after this

cmd/gateway/main.go
coyle: That's new to me but we are doing this in Uplink and the story was do what uplink does

pkg/cfgstruct/utils.go
bryanchriswhite: Maybe a config for the short-term.. we could have a hard-coded default that we keep in the repo (like the default peer CA whitelist).

coyle: ok - I'll look into this in a follow-up PR. I need to change Uplink as well if we do this and should probably make them use the shared lib

cmd/gateway/main.go
mniewrzal: to use just `SaveConfig` we need to mark some flags with `user` tag to save them in config. We did it for storage node to limit number of flags in config file. It would be good to do the same thing for uplink but I believe its separate task.

coyle:
@bryanchriswhite @mniewrzal anything else I should do ?

PR #1908, opened 2019-05-07
add postgres support to storj-sim

JessicaGreben:
What: 
Add the ability to run storj-sim with postgres (in addition to the current default sqlite). This requires to add support for schemas in postgres code for the satellite master db and also the pointer/metainfo db.  This PR also adds an env var to test-sim.sh script to use postgres.

Examples:
```
$ storj-sim network --postgres=postgres://postgres:<pw>@localhost/testdb?sslmode=disable setup

$ storj-sim network run
```

```
$ export STORJ_SIM_POSTGRES=postgres://postgres:<pw>@localhost/testdb?sslmode=disable

$ make test-sim
```

Why:
So that we can test with postgres since that is whats used in production also so that there is consistent support for the same dbs across our test frameworks.


cmd/storj-sim/main.go
```diff
@@ -23,6 +23,11 @@ type Flags struct {
 	IsDev bool
 
 	OnlyEnv bool // only do things necessary for loading env vars
+
+	// Connection string for the satellite masterdb
+	SatelliteDB string
```
egonelbre: A simpler design would allow to specify `--postgres postgres://storj:storj-pass@localhost/teststorj?sslmode=disable` and rest happens automatically.

cmd/storj-sim/network.go
```diff
@@ -260,6 +260,10 @@ func newNetwork(flags *Flags) (*Processes, error) {
 
 		consoleAuthToken := "secure_token"
 
+		satelliteDB := flags.SatelliteDB
+		if os.Getenv("STORJ_POSTGRES_TEST") != "" {
+			satelliteDB = os.Getenv("STORJ_POSTGRES_TEST")
```
egonelbre: This is for different purpose and probably shouldn't override storj-sim behavior.

cmd/storj-sim/network.go
```diff
@@ -281,6 +285,9 @@ func newNetwork(flags *Flags) (*Processes, error) {
 
 				"--version.server-address", fmt.Sprintf("http://%s/", versioncontrol.Address),
 				"--debug.addr", net.JoinHostPort("127.0.0.1", port(satellitePeer, i, debugHTTP)),
+
+				"--database", satelliteDB,
+				"--metainfo.database-url", flags.MetainfoDB,
```
egonelbre: These won't work with multiple satellites, you need to use schemas to make it work.

cmd/storj-sim/network.go
egonelbre: Also for command line flags the usual behavior is that command line flags override environment variables not the other way around.

cmd/satellite/main.go
```diff
@@ -26,6 +26,7 @@ import (
 // Satellite defines satellite configuration
 type Satellite struct {
 	Database string `help:"satellite database connection string" default:"sqlite3://$CONFDIR/master.db"`
+	Schema   string `help:"satellite database schema name" default:""`
```
egonelbre: You should not need this. You can use connstr to pass it in.

cmd/storj-sim/network.go
```diff
@@ -281,6 +288,10 @@ func newNetwork(flags *Flags) (*Processes, error) {
 
 				"--version.server-address", fmt.Sprintf("http://%s/", versioncontrol.Address),
 				"--debug.addr", net.JoinHostPort("127.0.0.1", port(satellitePeer, i, debugHTTP)),
+
+				"--database", database,
+				"--schema", schema,
+				// TODO(jg): add connection to postgres for metainfo.databaseURL once metainfo refactor is complete
```
egonelbre: You can add the pointerdb/metainfo url already now, there shouldn't be issues with it. If there are then these should be fixed.

cmd/satellite/main.go
JessicaGreben: if i dont pass in the schema, then i need to parse it out of the connstr before I can call `db.CreateSchema`.  I thought passing this in was cleaner. thoughts?

cmd/storj-sim/network.go
JessicaGreben: yeah i was having trouble with metainfodb using postgres, but let me try again. I could have been doing something wrong.

cmd/satellite/main.go
egonelbre: Do you need to explicitly create the schema? But, I would rather handle it transparently without an extra flag.

JessicaGreben: I think I do need to create the schema explicitly. At least it was erroring before I added did. Ok let me try to change it to not need an extra flag.

cmd/satellite/main.go
```diff
@@ -141,6 +142,13 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Error("Failed to initialize telemetry batcher: ", err)
 	}
 
+	if runCfg.Schema != "" {
+		err = db.CreateSchema(runCfg.Schema)
```
egonelbre: Move this into `satellitedb` `CreateTables` method.

satellite/console/consoleweb/server.go
```diff
@@ -76,7 +77,7 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
 		mailService: mailService,
 	}
 
-	logger.Debug("Starting Satellite UI...")
+	logger.Debug(fmt.Sprintf("Starting Satellite UI on %s...", server.listener.Addr().String()))
```
egonelbre: `logger.Sugar().Debugf`

cmd/satellite/main.go
JessicaGreben: I tested out adding `CreateSchema` to the `CreateTables` method, but [`CreateTables`](https://github.com/storj/storj/blob/master/satellite/satellitedb/migrate.go#L22) doesn't have the database connstr available to it.  So I'm not sure what is best.  What I initially chose to do is [this](https://github.com/storj/storj/pull/1908/commits/eb0fb725b99132b29c47dab6b8c37d0362938f6f#diff-e7e4cbc943e802fe544ebbf762a3a459R145) just keep the CreateSchema in this file.

egonelbre: You can save the schema or connstr in https://github.com/storj/storj/blob/master/satellite/satellitedb/database.go#L47.

internal/dbutil/pgutil/schema.go
```diff
@@ -28,6 +28,15 @@ func ConnstrWithSchema(connstr, schema string) string {
 	return connstr + "&search_path=" + url.QueryEscape(schema)
 }
 
+// ParseSchemaFromConnstr returns the schema name parsed from the connection string
+func ParseSchemaFromConnstr(connstr string) (string, error) {
+	connstrParts := strings.Split(connstr, "&search_path=")
```
egonelbre: There's `url.Parse` which allows to properly decode the string. The `search_path` isn't necessarily the last argument.

JessicaGreben: good point

cmd/storj-sim/network.go
```diff
@@ -262,10 +262,13 @@ func newNetwork(flags *Flags) (*Processes, error) {
 		consoleAuthToken := "secure_token"
 
 		var schema string
-		database := "sqlite3://" + process.Directory + "/master.db"
+		sqlDatabase := "sqlite3://" + process.Directory + "/master.db"
+		kvDatabase := "bolt://" + process.Directory + "pointers.db"
```
egonelbre: You might be able to avoid this, if you add the argument dynamically when needed.

JessicaGreben: good idea, let me try that

storage/postgreskv/schema/migrate.go
```diff
@@ -37,3 +53,11 @@ func PrepareDB(db *sql.DB) error {
 	}
 	return err
 }
+
+func createSchema(db *sql.DB, schema string) error {
+	_, err := db.Exec(`create schema if not exists ` + strconv.QuoteToASCII(schema) + `;`)
+	if err != nil {
```
mniewrzal: nit: it can be just `return err`

cmd/satellite/main.go
```diff
@@ -116,7 +116,7 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Fatal(err)
 	}
 
-	db, err := satellitedb.New(log.Named("db"), runCfg.Database)
+	db, err := satellitedb.New(log.Named("satellite master db"), runCfg.Database)
```
egonelbre: Keep this as is, since it comes from satellite log, it should be obvious that it's master db.

storage/postgreskv/schema/migrate.go
```diff
@@ -7,22 +7,38 @@ package schema
...
 	if err != nil {
 		return err
 	}
+
+	schema, err := pgutil.ParseSchemaFromConnstr(dbURL)
+	if err != nil {
+		return errs.New("error parsing schema: %+v", err)
+	}
+	if schema != "" {
+		err := createSchema(db, schema)
```
egonelbre: There is `pgutil.CreateSchema`

JessicaGreben: oops you are right, i missed that. thanks

cmd/storj-sim/network.go
```diff
@@ -284,6 +285,13 @@ func newNetwork(flags *Flags) (*Processes, error) {
 			},
 			"run": {},
 		})
+		var schema string
+		if flags.Postgres != "" {
+			schema = fmt.Sprintf("satellite%d", i)
+			postgresConnstr := pgutil.ConnstrWithSchema(flags.Postgres, schema)
+			dbArgs := []string{"--database", postgresConnstr, "--metainfo.database-url", postgresConnstr}
```
egonelbre: main database and metainfo probably should use different schemas.
e.g. `satellite/0` and `satellite/0/metainfo`

JessicaGreben: good point

egonelbre: They may have some conflicts due to separate initialization. Might be better to leave them separate.

JessicaGreben: what part of the initialization will conflict?  the createSchema function executes `create schema if not exists` so that should be ok. what else would conflict?

egonelbre: There are other tables that the postgreskv makes and satellite. I'm unsure whether they conflict... easier it's to use different schemas to ensure it definitely doesn't happen. However, if it definitely doesn't conflict, then I'm fine with keeping them in the same schema.

PR #1910, opened 2019-05-07
real-time tracking of space used per project

thepaul:
Ran into difficulties trying to find the ideal solution for sharing
these counts between multiple satellite servers, so for now this is a
dumb solution storing recent space-usage changes in a big dumb in-memory
map with a big dumb lock around it. The interface used, though, should
allow us to swap out the implementation without much difficulty
elsewhere once we know what we want it to be.

This does not (yet?) account for deletions and garbage-collected objects
freeing up used space in real time. The next time a full tally is done (once
per hour by default) that will be fixed, but ideally we ought to hook in to
places that things are _removed_ from the pointerdb as well.


satellite/metainfo/metainfo.go
```diff
@@ -43,26 +44,28 @@ type APIKeys interface {
...
-	accountingDB  accounting.DB
-	maxAlphaUsage memory.Size
+	log               *zap.Logger
+	metainfo          *Service
+	orders            *orders.Service
+	cache             *overlay.Cache
+	apiKeys           APIKeys
+	accountingDB      accounting.DB
+	accountingRTCache accountingcache.Service
+	maxAlphaUsage     memory.Size
```
VinozzZ: Should we remove this `Alpha` here to be `maxUsage`?

thepaul: Good call!

pkg/accounting/tally/tally.go
```diff
@@ -65,6 +68,15 @@ func (t *Service) Run(ctx context.Context) (err error) {
 
 // Tally calculates data-at-rest usage once
 func (t *Service) Tally(ctx context.Context) error {
+	// The RT cache will only keep deltas to space/bw used relative to the
```
JessicaGreben: its hard to know what "RT" stands for. If possible, could we say "real time" in places where "RT" is?

satellite/accountingcache/accountingcache.go
```diff
@@ -0,0 +1,107 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package accountingcache
```
JessicaGreben: im curious why we would add this package here versus in `pkg/accounting` like the `tally` and `rollup` service?

pkg/accounting/tally/tally.go
thepaul: I am very amenable to that, but first can somebody come up with a better name than `accountingRTCache` for that member? I've had an especially hard time naming this whole subsystem because it doesn't really fill any normal role that people would be familiar with. It's not _really_ a cache, and it can't be too similar to `accountingDB` because that would cause confusion with a different thing that is related but entirely different. `accountingRealTimeCache` would imply that it's a "real time cache", whatever that is, instead of a collection of live accounting data used for real-time tracking purposes.

satellite/accountingcache/accountingcache.go
thepaul: `pkg/accounting` seems like the right place by its name alone, but in code it looks to me like it's pretty tightly coupled with the accounting part of satelliteDB, and this is entirely separate from that. Secondly, I've been told that `pkg/` is for things that apply to storj code in general, not any one specific subsystem, and this code definitely only applies to a satellite. But everything is in flux anyway, so I could probably be convinced that `pkg/accounting` is where it needs to be for now.

pkg/accounting/tally/tally.go
stefanbenten: What about `liveaccounting`?

thepaul: Works for me.

satellite/liveaccounting/liveaccounting.go
```diff
@@ -0,0 +1,107 @@
...
+// Config contains configurable values for the liveaccounting service.
+type Config struct {
+	StorageBackend string `help:"What to use for storing real-time accounting data"`
+}
+
+// Service represents the external interface to the liveaccounting
+// functionality.
+type Service interface {
+	GetProjectStorageUsage(ctx context.Context, projectID uuid.UUID) (int64, int64, error)
+	AddSpaceUsed(projectID uuid.UUID, inlineSpaceUsed, remoteSpaceUsed int64) error
```
JessicaGreben: seems like we are flipping between `StorageUsage` and `SpaceUsage`, I'm wondering if we want to pick one and stick with it?

thepaul: `StorageUsage` and `SpaceUsed`, actually, yeah. "Storage", to me, is the more general term; it can refer to anything that is "stored" by the system in any way, or to the act or concept of storing. In this particular case, it refers to information about multiple different types of space used. "Space", on the other hand, refers to a specific measurement: the total number of bytes that exist in the system, bytes used by a file, unused bytes on a hard disk, or—in this case—the number of bytes stored inline in the pointerDB and the number of bytes stored remotely.

thepaul: If most people don't see any such distinction between the terms, then yeah, I should change it.

satellite/liveaccounting/liveaccounting.go
```diff
@@ -0,0 +1,102 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package liveaccounting
```
egonelbre: Wondering whether there is a better name for this package? Or maybe accounting package should have a `Live` struct instead?

satellite/liveaccounting/liveaccounting.go
JessicaGreben: that distinction makes sense.  then with your definition,  it seems like they all should be named `SpaceUsage`. I don't have a preference one way or the other, I just like the naming to be consistent so its easier to read.

thepaul: Do you mean the method names or the variable names, or both? I have already changed the method name on line 25 to be `AddProjectStorageUsage()`, and everything else seems to be named according to my definitions there.

satellite/liveaccounting/liveaccounting.go
thepaul: `pkg/accounting` seems like the right place by its name alone, but in code it looks to me like ithat directory is pretty tightly coupled with the accounting part of satelliteDB, and this is entirely separate from that. Secondly, I've been told that `pkg/` is for things that apply to storj code in general, not any one specific subsystem, and this code definitely only applies to a satellite. But everything seems to be in flux anyway, so I could probably be convinced that `pkg/accounting/live` or whatever is where it needs to be for now.

satellite/liveaccounting/liveaccounting.go
```diff
@@ -0,0 +1,102 @@
...
+// Config contains configurable values for the liveaccounting service.
+type Config struct {
+	StorageBackend string `help:"What to use for storing real-time accounting data"`
+}
+
+// Service represents the external interface to the liveaccounting
+// functionality.
+type Service interface {
+	GetProjectStorageUsage(ctx context.Context, projectID uuid.UUID) (int64, int64, error)
+	AddProjectStorageUsage(projectID uuid.UUID, inlineSpaceUsed, remoteSpaceUsed int64) error
```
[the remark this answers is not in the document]
thepaul: Good point. Will fix

satellite/liveaccounting/liveaccounting.go
```diff
@@ -0,0 +1,102 @@
...
+	"sync"
+
+	"github.com/skyrings/skyring-common/tools/uuid"
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+)
+
+// Config contains configurable values for the liveaccounting service.
+type Config struct {
+	StorageBackend string `help:"What to use for storing real-time accounting data"`
```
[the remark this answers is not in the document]
thepaul: Oh. Wasn't aware of that, but you're right. Will fix

satellite/liveaccounting/liveaccounting.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(parts) == 0 || parts[0] == "" {
+		backendType = "plainmemory"
+	} else {
+		backendType = parts[0]
+	}
+	switch backendType {
+	case "plainmemory":
+		return newPlainMemoryLiveAccounting(log)
+	}
+	return nil, errs.New("Unrecognized liveaccounting backend specifier %q", backendType)
```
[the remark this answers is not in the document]
thepaul: 👍

satellite/liveaccounting/liveaccounting.go
egonelbre: Some of the packages haven’t yet been moved near their appropriate peer. That’s the reason `pkg/accounting` is still there, but there are tickets to move them.

egonelbre: Some of the packages haven’t yet been moved near their appropriate peer. That’s the reason `pkg/accounting` is still there, but there are tickets to move them.

satellite/metainfo/metainfo.go
```diff
@@ -188,6 +189,34 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
...
+	}
+	segmentSize := ptr.GetSegmentSize()
+	remote := ptr.GetRemote()
+	if remote == nil {
+		return 0, 0
+	}
+	minReq := remote.GetRedundancy().GetMinReq()
+	pieceSize := segmentSize / int64(minReq)
+	pieces := remote.GetRemotePieces()
+	return 0, pieceSize * int64(len(pieces))
```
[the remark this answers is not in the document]
thepaul: `pieceSize := segmentSize / int64(minreq)` is an inherent part of our system; pieces end up being this big as a consequence of the reed-solomon encoding. See pkg/accounting/tally/tally.go:180, pkg/eestream/encode.go:274, pkg/storage/ec/client.go:346, etc.

`pieceSize * len(pieces)` is because the post-expansion-factor space used is the size of each piece times the number of pieces we have stored.

satellite/liveaccounting/liveaccounting.go
thepaul: moved to pkg/accounting/live.

satellite/accountingcache/accountingcache.go
thepaul: moved to pkg/accounting/live.

satellite/liveaccounting/liveaccounting.go
JessicaGreben: i think its fine the way it is now.

satellite/metainfo/metainfo.go
```diff
@@ -43,26 +44,28 @@ type APIKeys interface {
...
-		accountingDB:  acctDB,
-		maxAlphaUsage: maxAlphaUsage,
+		log:            log,
+		metainfo:       metainfo,
+		orders:         orders,
+		cache:          cache,
+		apiKeys:        apiKeys,
+		accountingDB:   acctDB,
+		liveAccounting: liveAccounting,
+		maxAlphaUsage:  maxAlphaUsage,
```
JessicaGreben: there are a lot of places still using `MaxAlphaUsage` variables.  Should all that "alpha" stuff be removed since this project limiting isn't related to the alpha release anymore? Or should we create a different ticket to clean all that up?

thepaul: @iglesiasbrandon told me it still is going to be limited to the alpha release after all, so I put back all the “alpha” references I had taken out previously. See 0a8012420143bca9e987cf9458db7c9512fa6fa8

pkg/accounting/live/live.go
```diff
@@ -0,0 +1,102 @@
...
+// functionality.
+type Service interface {
+	GetProjectStorageUsage(ctx context.Context, projectID uuid.UUID) (int64, int64, error)
+	AddProjectStorageUsage(ctx context.Context, projectID uuid.UUID, inlineSpaceUsed, remoteSpaceUsed int64) error
+	ResetTotals()
+}
+
+// New creates a new live.Service instance of the type specified in
+// the provided config.
+func New(log *zap.Logger, config Config) (Service, error) {
```
egonelbre: For data backend I've been trying to pass it as an argument to peer rather than create a direct dependency on peers. This ensure we don't create a hard dependency. Also it makes more obvious when we are touching disk.

thepaul: Sorry- pass what as an argument to peer? What peer?

thepaul:
@ifraixedes Ok, I'll add that. And a test. This does of course need a test.

satellite/metainfo/metainfo.go
JessicaGreben: odd. @jtolds told me that it wasn't limited to alpha anymote. hmmm

thepaul: I tried to follow up more, but didn't get answers, so I decided to leave it all intact for now (and avoid requiring a change in the satellite config files, because they have the "alpha" in there too). If we decide to keep it, we can remove the alpha name down the road, I guess.

pkg/accounting/live/live.go
egonelbre: I mean `satellite.Peer`.

Currently this seems to implement persistance layer, which is in the realm of `satellitedb`, so it seems like it might fit more there. But, I agree it doesn't fit that mold exactly either since it's something inbetween a service and a cache.

thepaul: I think I've used the same patterns as all the other service-type things in peer.go ( https://github.com/storj/storj/blob/f74001ae9066fc84239f9cc563625793a24458df/satellite/peer.go#L311-L319 ). Is that what you mean?

egonelbre: I was mainly wondering whether this actually should be part of the database.

PR #1911, opened 2019-05-07
docs/design: template

jtolds:
Starting to put design docs in source control


ifraixedes:
Could we also add the creation/review/approve process?

PR #1912, opened 2019-05-08
Move uplink mount into separate command

mniewrzal:
What: Separate command for `storj-mount` utility

Why: Command `uplink mount` was removed from uplink but should be somehow accessible

Thanks for submitting a PR!


egonelbre:
I still think this should be a separate repository with it's own `go.mod` we want to make the core as stable as possible, by adding all different features to it it makes things more difficult.

mniewrzal:
Keeping this in separate repo is quite easy. Small problem is just with internals from master.

Separate repo would be for storj connectors?

I think we can use this PR to discuss this. @jtolds

jtolds:
One thing to keep in mind is that any uplink-side code before we launch will be relicensed to be Apache v2 instead of AGPLv3, so all libuplink code (libuplink and any dependencies, hopefully a small set) will need to be in its own repo for licensing purposes. So, we do need to break libuplink out.

It'd also be nice to have other repos for tools that use libuplink for sure. 

However, there's a lot of shared code overlap between the gateway tool and the CLI tool. In particular, we'd need to come up with a replacement for pkg/process and friends (probably easy) that includes telemetry, tracing/debug, and metrics in addition to a few other things (probably harder). In terms of prioritization, I would say breaking the gateway tool and CLI tool out into their own repos out of storj/storj is not something we really have the time to spend until after our first beta launch (we want to by the end of beta though), and, since this mount tool is also basically done and the strategy team would like to use it, is it okay if we leave this in this repo until we break those tools out?

We need to break those tools out but there's more pressing stuff to do at the current moment given pressing partner deadlines. I'm supportive of merging this as-is. We know we *must* break the tools out due to licensing issues, so we'll get to it eventually.

jtolds:
Summary: @egonelbre is absolutely right about our long term goals but because that takes more work and won't help us with short term deadlines, I'd like to punt that for now and merge as is with the goal of focusing on more pressing things.

cmd/storj-mount/mount.go
```diff
@@ -0,0 +1,561 @@
...
+		return nil, fuse.EIO
+	}
+
+	f.predictedOffset = off + int64(n)
+
+	return fuse.ReadResultData(buf[:n]), fuse.OK
+}
+
+func (f *storjFile) Write(data []byte, off int64) (uint32, fuse.Status) {
+	writer, err := f.getWriter(off)
```
jtolds: offset is not always ordered from 0. Writes might come in in the order "100, 500, 150, 0, 200" if the writing application is seeking around. The previous version of this had an expectedOffset for both reads and writes that was compared. We should do that here (or in getWriter) and throw an error if the offset does not match the expected offset

cmd/storj-mount/mount.go
```diff
@@ -0,0 +1,561 @@
...
+
+	f.size += uint64(written)
+
+	zap.S().Debug("Stops writting: ", f.name)
+
+	return uint32(written), fuse.OK
+}
+
+func (f *storjFile) getReader(off int64) (io.ReadCloser, error) {
+	if f.reader == nil {
```
jtolds: this also needs an expected offset feature

egonelbre:
I guess one of my two reasons of not merging `storj-mount` into `storj/storj`:
1. it's not critical to our short term goals
2. it's a good test-case whether people can use libuplink externally
3. I'm not certain that the `storj-mount` is done-done, AFAIK it still has a bunch of TODO-s about editing, but @mniewrzal can correct me on this...

However, for short term it probably won't affect us too much, if it does start to affect we can always split it off as needed.

mniewrzal:
`storj-mount` for sure is not fully done. We still need:
* general review (integration with libuplink was done quite fast)
* bugfixing (things commented by @jtolds and other)
* automatic tests 

My main goal was to put this 'somewhere' more or less as is because we removed it from master. We cannot use it for demoing but we have more important things to do. Maybe we can just keep it as a separate branch? Currently, we are not building binaries for this util so one way or another if someone wants to use must compile it. If I will have spare time I can improve fuse integration on this branch.

jtolds:
The mount command is critical to the strategy team's short term goals and is why they asked @mniewrzal to put what he had somewhere. I think they plan to continue development and fix it up.

jtolds:
But okay, @mniewrzal has convinced me the bug fixes aren't important yet.

mniewrzal:
closing PR but leaving branch

PR #1913, opened 2019-05-08
storage node collector

egonelbre:
What: add storage node piece collector

Why:

Thanks for submitting a PR!


storagenode/collector/service.go
```diff
@@ -1,33 +1,96 @@
...
+}
+
+func (service *Service) collect(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	now := time.Now()
+	const maxBatches = 100
+	const batchSize = 1024
+
+	for k := 0; k < maxBatches; k++ {
```
stefanbenten: Can we take all of the code below outside of the loop and avoid having a max Batch Count?
Unlikely that we have to remove 1 Mio Entries within an hour, but still feeling wrong to me.

I guess you made that to avoid having two concurrent collect functions being run?

storagenode/peer.go
```diff
@@ -247,12 +255,17 @@ func (peer *Peer) Run(ctx context.Context) error {
...
 		return errs2.IgnoreCanceled(peer.Version.Run(ctx))
 	})
+
 	group.Go(func() error {
 		return errs2.IgnoreCanceled(peer.Kademlia.Service.Bootstrap(ctx))
 	})
 	group.Go(func() error {
 		return errs2.IgnoreCanceled(peer.Kademlia.Service.Run(ctx))
 	})
+
```
stefanbenten: Nit: Either remove both of these empty lines or add one in line 275

storagenode/collector/service.go
kaloyan-raev: Shouldn't we have the loop executing until `GetExpried` returns a non-empty result?

egonelbre: We have both.

I'm not sure how to handle the situation where it is unable to delete the piece from disk (e.g. misplaced permissions). This protects against that issue. But there might be a cleaner way to handle this situation.

storagenode/collector/service.go
```diff
@@ -1,33 +1,96 @@
...
+	service.Loop.Close()
+	return nil
+}
+
+func (service *Service) collect(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	now := time.Now()
+	const maxBatches = 100
+	const batchSize = 1024
```
kaloyan-raev: Why 1024, and not 1000? We have already hardcoded limits to 1000 in several places.

storagenode/collector/service.go
```diff
@@ -1,33 +1,96 @@
...
+		}
+
+		for _, id := range ids {
+			err := service.pieces.Delete(ctx, id.SatelliteID, id.PieceID)
+			if err != nil {
+				service.log.Error("unable to delete piece", zap.Stringer("satellite id", id.SatelliteID), zap.Stringer("piece id", id.PieceID))
+				continue
+			}
+
+			err = service.pieceinfos.DeleteExpired(ctx, now, id.SatelliteID, id.PieceID)
```
kaloyan-raev: Why do we have `DeleteExpired` method at all? We already determined that this piece has expired, and already deleted it from the hard drive (line 81). So we just need to delete it from the infodb too, unconditionally. Having a call to `DeleteExpired` instead of `Delete` is just another potential source for bugs.

egonelbre: I think my initial reasoning for this was that maybe expired time gets updated inbetween GetExpired and Delete, then this would prevent it from removing... although, then current handling wouldn't be sufficient because we already deleted it from the disk.

egonelbre: Or alternatively, it was something about allowing double deletion with expired. Either way, I'll remove it for now.

storagenode/collector/service.go
kaloyan-raev: What if the first 1024 (the batch limit) failed to delete? Then the loop will always get them again and again from the db and will never be able to delete other pieces that are ordered after them.

Maybe we should have a "delete_error" column in infodb and only try to delete those again only once per cycle.

egonelbre: Yes, hence the reason there is the max batches... so it will stop after 100 tries.

egonelbre: One option would be to delete from pieceinfo table and let user cleanup the situation.

But yes, we could add a some threshold for trying deletion.

kaloyan-raev: I am not sure if it is feasible for the SNO to manually delete piece by piece only those pieces that failed deletion. I would imagine the SNO to fix the problem (permissions or other) and then wait for the collector to execute once again.

Also, if the piece has already been deleted from the hard drive, trying to delete it once again won't result in an error, right?

egonelbre: > Also, if the piece has already been deleted from the hard drive, trying to delete it once again won't result in an error, right?

Yes

egonelbre: fixed

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -88,64 +88,67 @@ func (db *pieceinfo) Get(ctx context.Context, satelliteID storj.NodeID, pieceID
...
-		WHERE piece_expiration < ? ORDER BY satellite_id `
+// GetExpired gets pieceinformation identites that are expired.
+func (db *pieceinfo) GetExpired(ctx context.Context, expiredAt time.Time, limit int64) (infos []pieces.ExpiredInfo, err error) {
+	defer db.locked()()
 
-	rows, err := db.db.QueryContext(ctx, db.Rebind(getExpiredSQL), expiredAt)
+	rows, err := db.db.QueryContext(ctx, db.Rebind(`
+		SELECT satellite_id, piece_id, piece_size
+		FROM pieceinfo
+		WHERE piece_expiration < ? AND ((deletion_failed_at IS NULL) OR deletion_failed_at <> ?)
```
kaloyan-raev: I don't understand the purpose of this part: `deletion_failed_at <> ?`. In reality, it should be always satisfied. It's very hard to pick the same timestamp.

egonelbre: Ideally it should just be `WHERE piece_expiration < ? AND deletion_failed_at <> ?`... however there's some weirdness going on with nulls...

The Collect loop uses the same timestamp for all of the batches within a timeframe. It could use a threshold, but that would make that expression even more complicated. Also not sure what will happen with clock and timezone changes here otherwise.

kaloyan-raev: Oh, I think I got it. This would prevent `GetExpired` to return the same piece with "deletion failed flag" more than once in the same cycle.

PR #1915, opened 2019-05-08
v3-1660: improvement of test-satellite-cfg-change and related things

ifraixedes:
What: 
* Clarifies that the test-satellite-cfg-change is for being run by
  Jenkins as others are.
* Move the target for updating the satellite configuration lock file to
  a new Makefile group, called "Tooling".
* Change how it's run the script update-satellite-cfg-lock.sh for
  ensuring that the outcome is the same when running from different
  machines.

Why: It wasn't clear that one of the targets is for running it from Jenkins and the other one didn't produce the same outcome from different machines.

@mobyvb this should be fixed what you spotted and commented in https://github.com/storj/storj/pull/1898#pullrequestreview-234701092
@wthorp removing it was an easy solution and if I wouldn't have come up with this solution I would do it, but this solution should be better than removing it.

Thanks for submitting a PR!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,13 +91,9 @@ test-sim: ## Test source with storj-sim (jenkins)
 	@./scripts/test-sim.sh
 
 .PHONY: test-satellite-cfg-change
-test-satellite-cfg-change: ## Test if the satellite config file has changed
+test-satellite-cfg-change: ## Test if the satellite config file has changed (jenkins)
 	@echo "Running ${@}"
 	@cd scripts; ./test-satellite-cfg-change.sh
-
-.PHONY: test-update-satellite-cfg-lock
-test-update-satellite-cfg-lock: ## Update the satellite config lock file
-	@cd scripts; ./update-satellite-cfg-lock.sh
 
 .PHONY: test-certificate-signing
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -272,3 +268,16 @@ clean-images:
 test-docker-clean: ## Clean up Docker environment used in test-docker target
 	-docker-compose down --rmi all
 
+
+##@ Tooling
+
+.PHONY: update-satellite-cfg-lock
+update-satellite-cfg-lock: ## Update the satellite config lock file
+	@docker run -ti --rm \
```
egonelbre: This docker seems a large overkill. Cannot we add something like `satellite print-config >> ./scripts/satellite.yaml.lock`?

ifraixedes: We could but it's an improvement to something that it's already in master and it doesn't do what's expected, moreover this must be run by a developer machine and it shouldn't be often used so the delay that there is in running in docker that it's less than a minute in my machine shouldn't stop to fix the problem with this approach.

Making the improvement that you mentioned would be desired, however I see that it's more work than the one planned for the current ticket, but I may be wrong with it. If the amount of work to add such sub-command to the satellite is no brainer I would like some guidance with it, because I don't have enough knowledge on the code base to see it in such way.

PR #1919, opened 2019-05-08
cmd/statreceiver: enable lua libraries in lua config

jtolds:
Change-Id: I104fa9ef6e0d3102f3213fdf2ef33c4547557f72

What: enable things like os.getenv in the statreceiver config

Why: so we can configure things based on other files or the environment if we want to

Thanks for submitting a PR!


cmd/statreceiver/luacfg/cfg.go
```diff
@@ -60,6 +60,7 @@ func (scope *Scope) register(name string, val interface{}, pusher func(l *lua.St
 // Run runs the Lua source represented by in
 func (scope *Scope) Run(in io.Reader) error {
 	l := lua.NewState()
+	lua.OpenLibraries(l)
```
egonelbre: I was reading this as `lua.OpenLibraries(1)` and kept thinking what that `1` meant.

PR #1920, opened 2019-05-08
Uplink CLI should not require identity

kaloyan-raev:
Fixes https://storjlabs.atlassian.net/browse/V3-1713

What: 

Removes the identity from the uplink configuration.

Why:

Because it is not required to have a designated identity anymore. Libuplink will internally generate a disposable identity with difficulty 0 whenever the uplink CLI is used.

Thanks for submitting a PR!

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?

PR #1921, opened 2019-05-08
Gateway should not require identity

kaloyan-raev:
What: 

Same as #1920, but for the S3 Gateway.

Why:

Because the S3 Gateway should not require designated identity anymore. Libuplink will internally generate a disposable identity with difficulty 0 whenever the gateway is started.

Thanks for submitting a PR!


kaloyan-raev:
@ifraixedes Yes. Good point. I guess this should be a separate PR in which I want to hear the opinion of @thepaul and @jtolds.

egonelbre:
Also remove identity generation from `storj-sim` in this case.

PR #1924, opened 2019-05-08
uplink share subcommand

zeebo:
What: Add a share subcommand to the uplink command

Why: To allow one to restrict macaroon based api keys


cmd/uplink/cmd/share.go
```diff
@@ -0,0 +1,147 @@
...
+	} else if date[0] == '+' {
+		d, err := time.ParseDuration(date[1:])
+		t := now.Add(d)
+		return &t, errs.Wrap(err)
+	} else if date[0] == '-' {
+		d, err := time.ParseDuration(date[1:])
+		t := now.Add(-d)
+		return &t, errs.Wrap(err)
+	} else {
+		t, err := dateparse.ParseAny(date)
```
egonelbre: My concern with this is that it picks an arbitrary resolution for dates in the form of `01.03.2019` and `01/03/2019` rather than using locale. I would rather restrict us to ISO8601 like formats (year, month, day etc.).

go.mod
```diff
@@ -117,6 +119,7 @@ require (
 	google.golang.org/grpc v1.19.0
 	gopkg.in/Shopify/sarama.v1 v1.18.0 // indirect
 	gopkg.in/cheggaaa/pb.v1 v1.0.25 // indirect
+	gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce // indirect
```
egonelbre: Where did the mongo dependency come from?

cmd/uplink/cmd/share.go
ifraixedes: or return an error when the date is ambiguous using https://godoc.org/github.com/araddon/dateparse#ParseStrict

go.mod
ifraixedes: it comes from simplereach/timeutils https://github.com/simplereach/timeutils/blob/master/time.go#L10

pkg/macaroon/apikey.go
```diff
@@ -137,8 +137,8 @@ func (a *APIKey) Tail() []byte {
 }
 
 // Serialize serializes the API Key to a string
-func (a *APIKey) Serialize() (string, error) {
-	return base58.CheckEncode(a.mac.Serialize(), 0), nil
+func (a *APIKey) Serialize() string {
```
ifraixedes: When I reviewed this code, I thought that returning always a `nil` error was because a future change of the implementation which may return some error was not going to break the exported API.

I would have probably asked rather than assuming it.

go.mod
zeebo: Gross.

cmd/uplink/cmd/share.go
zeebo: Would it be better if it printed out the time in an unambiguous format if specified? Personally, I can never remember date formats.

pkg/macaroon/apikey.go
jtolds: that was the reason but i don't care strongly

zeebo: We also have this public exported interface in lib/uplink that doesn't error: https://godoc.org/storj.io/storj/lib/uplink#APIKey.Serialize

That informed my decision to remove the error, because if the APIKey type includes a macaroon, it will have to ignore it.

jtolds: i'm just a dummy! it's okay to add it there too. very minor changes are okay for lib/uplink still. but either way

cmd/uplink/cmd/share.go
ifraixedes: For me being stuck on one format is OK, but this a user level application, and you know that can make some users uncomfortable.

egonelbre: I wouldn't call `uplink` really an user level application. I can see it only being used by programmers and scripting. Programmers can probably figure out ISO8601. Scripters probably want something consistent that doesn't break when they move code from one source to another.

go.mod
egonelbre: https://github.com/araddon/dateparse/pull/85

cmd/uplink/cmd/share.go
ifraixedes: good to me, my sentence should have been

> For me being stuck on one single format is preferable (and ISO8601 is the best) ...

go.mod
ifraixedes: Let's see if it gets accepted.

cmd/uplink/cmd/share.go
zeebo: Alright, sounds like the consensus is to just require ISO8601.

pkg/macaroon/apikey.go
zeebo: Honestly, I'd prefer to not have serialize be able to error. How important is this?

ifraixedes: Being a method it shouldn't report an error except that an instance of such types could not be serialized for some reason (e.g. zero value isn't serializable).

If we don't envision that a future change could return an error, then it looks good to me to remove it.

cmd/uplink/cmd/share.go
```diff
@@ -0,0 +1,166 @@
...
+			Bucket:              []byte(p.Bucket()),
+			EncryptedPathPrefix: []byte(encPath),
+		})
+	}
+
+	{
+		// Times don't marshal very well with MarshalTextString, and the nonce doesn't
+		// matter to humans, so handle those explicitly and then dispatch to the generic
+		// routine to avoid having to print all the things individually.
+		caveat := caveat
```
jtolds: i wish there was a way to tell go that you really do mean to make a copy here and if caveat ever becomes a pointer that it should yell

jtolds: maybe this stanza could be a helper function to make it more clear what's going on

cmd/uplink/cmd/share.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+	notAfter, err := parseHumanDate(shareCfg.NotAfter, now)
+	if err != nil {
+		return err
+	}
+
+	// TODO(jeff): we have to have the server side of things expecting macaroons
+	// before we can change libuplink to use macaroons because of all the tests.
+	// For now, just use the raw macaroon library.
```
jtolds: well, so, this PR does change libuplink to use macaroons, right?

zeebo: No. This TODO explains why this PR doesn't do that. That's #1945, which I'd like to get this merged, that rebased, and then it's good to review.

cmd/uplink/cmd/share.go
zeebo: I'll change it to `proto.Clone` the thing.

cmd/uplink/cmd/share.go
jtolds: oh i misread what https://github.com/storj/storj/pull/1924/files#diff-23f3da9beee4401f662ac539c8f7faefR143 was doing

pkg/macaroon/caveat.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"encoding/binary"
+	"time"
+)
+
+// NewCaveat returns a Caveat with a nonce initialized to the current timestamp
+// in nanoseconds.
+func NewCaveat() Caveat {
+	var buf [8]byte
+	binary.BigEndian.PutUint64(buf[:], uint64(time.Now().UnixNano()))
```
thepaul: I can imagine users not being thrilled about each share token revealing the timestamp when it was made. Can we make this random?

zeebo: Sounds good.

zeebo: #1964

ifraixedes: I've just updated the doc comment on https://github.com/storj/storj/pull/1965 after the applied changes in #1964

PR #1925, opened 2019-05-08
Add Watchtower

brimstone:
There was a report from the community that our storagenode works in places watchtower doesn't. I added watchtower here to reuse the build process we have for the storagenode, thus guaranteeing the watchtower should work in the same places as the storagenode.

The watchtower project is very unstable and has changed names in the past, or moved tags which has caused me to vendor the Docker image in our own Docker hub org. This is just the ultimate step in vendoring their project.


stefanbenten:
Why do we want that?

brimstone:
There was a report from the community that our storagenode works in places watchtower doesn't. I added watchtower here to reuse the build process we have for the storagenode, thus guaranteeing the watchtower should work in the same places as the storagenode.

brimstone:
I think this should be part of our monorepo since we're advising it as the update method. It's constrained to its own directory which should be easy to remove once we have proper updates later.

egonelbre:
Can we fix this upstream instead of forking?

Alternatively, let's make a separate repo and build it there instead.

This adds GB (probably, didn't verify) of dependencies to anyone who wants to use libuplink.

brimstone:
Maybe? The watchtower project is very unstable and has changed names in the past, or moved tags which has caused me to vendor the Docker image in our own Docker hub org. This is just the ultimate step in vendoring their project.

egonelbre:
Also the project won't always be monorepo, we will try to split the project into multiple repositories once we start having more stability in code structure.

brimstone:
The report from the community was revised to state that the current watchtower image works fine. This isn't needed, yet.

PR #1929, opened 2019-05-08
adds audit system's containment mode design draft

navillasa:
What: auditing containment design draft

Why: needs some feedback + new design draft system!


docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Abstract
+
+This design doc outlines how we will implement "containment mode," in which the auditing service monitors a node that has accepted an audit request, but refuses to send the requested data.
+Currently in the codebase, nodes that do this will be marked as having audit failures.
+However, we don't want to mark nodes as offline or immediately mark an audit failure if this specific case arises.
+Instead, we want to "contain" these nodes by retrying the original audit until they either eventually respond and pass the audit, or refuse to respond a certain number of times and ultimately receive an audit failure mark.
+
+## Main Objectives
+1. Nodes won’t be able to “escape” audits for specific data by refusing to send requested data.
```
kaloyan-raev: We need a clear definition of what is "refusing to send the requested data".
Consider the following use cases:
- The node is busy fulfilling other request and cannot respond to the audit request within the audit timeout defined on the satellite
- The node does not have the audit piece anymore due to SNO mistake to delete it (I did such mistake today on my storage node while cleaning up empty directories, don't ask - I am really stupid). In such the node will return an error that it does not have the piece anymore.
- The node cannot read the piece due to file permission issue - an SNO config mistake. The node will return an error.
- The node cannot read the piece due to bad sectors on the HDD. The node will return an error.

Which of the above are considered refusal to return the requested data?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+2. The audit service should not audit contained nodes for other data, and should verify their originally requested data.
+3. The overlay service should not select contained nodes.
```
kaloyan-raev: What's the exact motivation for not selecting contained nodes? It seems to be worse to be in containment mode than failing audit requests.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+4. A ContainmentDB holds metadata required to verify contained nodes' originally requested data.
+
+## Background
+
+The whitepaper section 4.13 talks about containment mode as follows:
+```
```
ifraixedes: Can you change the format of this paragraph to use `>`?

Otherwise when rendered all the test gets in one line with a horizontal scroll

Example

> Given a specific storage node, an audit might reveal that ...

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+    pieceNum     int
+    stripeIndex  int
+    shareSize    int64
+    expectedShareHash []byte
+    reverifyCount   int // number of times reverify has been attempted
+}
+
+type ContainmentDB interface {
+  Contain(ctx context.Context, pendingAudit pendingAudit) error
+  Get(ctx context.Context, nodeID pb.NodeID) error
```
ifraixedes: `Get` misses a return value other than error.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+
+The verifier will then attempt to download that erasure share from the contained node.
+From there, the following cases can occur:
+- Passes audit ->  `pendingAudit` removed from the ContainmentDB  -> stats updated in overlaycache/satelliteDB
+- Fails audit -> `pendingAudit` is removed from the ContainmentDB -> stats updated in overlaycache/satelliteDB
+- Refuses to respond again -> the `pendingAudit`'s `reverifyCount` is incremented in the ContainmentDB ->
+    - -> If that count exceeds the reverify limit, then the node’s stats will be updated to reflect an audit failure and the node will be removed from the ContainmentDB
+    - -> If the count does not exceed the reverify limit, the `pendingAudit` will remain in the ContainmentDB
+
+```
```
ifraixedes: add the `go` syntax highlighting on this one an other code snippets which are below

docs/design/auditing-containment.md
littleskunk: 2-4 will return an error. That should still count as audit failure.

I agree. We should make it more clear.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+    pieceID      storj.PieceID
+    pieceNum     int
+    stripeIndex  int
+    shareSize    int64
+    expectedShareHash []byte
+    reverifyCount   int // number of times reverify has been attempted
+}
+
+type ContainmentDB interface {
+  Contain(ctx context.Context, pendingAudit pendingAudit) error
```
ifraixedes: Why not `Put` rather than `Contain` as `Get` it's?

This is a question that can have a response like: "it's the common convention for Store interfaces"

docs/design/auditing-containment.md
littleskunk: The problem comes with cheaters like me. In the past I was able to avoid audit penalites by dropping the connection without sending data or keep the connection open without sending data. The audit job was just hanging. Mission completed.

In order to stop cheaters like me you have to give me audit penalties. That should be our current implementation. The downside of that is that it will hit honest storage nodes as well. The containment mode will give these nodes a second and maybe even more chances while my cheater node is out of the game.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+```
+However, we're departing from the whitepaper because offline nodes will not be moved to containment mode.
+They will just be marked as offline.
+It's only when nodes initially respond to the audit service's dial but then refuse to send the requested erasure share that they are moved to containment mode.
+
+The node will be given a `contained` flag, and if the audit service attempts to audit the node again, it will see this flag and request the same share it had originally requested.
```
littleskunk: What happens in the following example:
First audit hits a segment with 80 pieces. For some reason 10 nodes are getting the containment flag.
Second audit hits a segment with 36 pieces. Now we are not allowed to get the pieces from the 10 cointained storage nodes? Why that? What is the advantage of that?

Worst case we end up with only 26 pieces and are not allowed to run this audit even if we could request all 36 pieces.

docs/design/auditing-containment.md
kaloyan-raev: But this is the reason to introduce the containment mode, right? After a few attempts to re-request the audited erasure share, your node will fail the audit and it will get the penalty.

I am also not sure how long it will take until a contained node will be re-checked in a real-life system. 30 minutes, 1 day, 1 year?

littleskunk: We have to hit each storage nodes at least a few times per day.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+    nodeID       storj.NodeID
+    pieceID      storj.PieceID
+    pieceNum     int
+    stripeIndex  int
+    shareSize    int64
+    expectedShareHash []byte
+    reverifyCount   int // number of times reverify has been attempted
+}
+
+type ContainmentDB interface {
```
wthorp: Are we saying we'd use another database table here? I'm struggling to see why we would. We never iterate all records in ContainmentDB. Node selection seems like the 95% use case, and I see no reason for the join.

docs/design/auditing-containment.md
wthorp: I wonder if we can/should combine Contain and ReverifyFail into a simple "IncrementFail" upsert

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+  Contain(ctx context.Context, pendingAudit pendingAudit) error
+  Get(ctx context.Context, nodeID pb.NodeID) error
+  ReverifyFail(ctx context.Context, nodeID pb.NodeID) error
+  ReverifySuccess(ctx context.Context, nodeID pb.NodeID) error
```
wthorp: "Reset"? We will perform nearly the same DB routines if we finally fail the audit.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+By "uncooperative," I mean that the auditor will be able to successfully DialNode (within pkg/audit/verifier.go), but errors will occur at
+```
+downloader, err := ps.Download(timedCtx, limit.GetLimit(), offset, int64(shareSize))
+```
+or
+```
+_, err = io.ReadFull(downloader, buf)
+```
+Once these errors have occurred and we know that a `pendingAudit` entry should be made in the ContainmentDB, we should use infectious’s `f.Decode` to generate the original stripe.
+Then we should use infectious’s `f.EncodeSingle` where we input the stripe and output the missing share. We then make a hash of that missing share to be saved to the ContainmentDB as `ExpectedShareHash`.
```
wthorp: This is cool.  What happens to containment if we don't have enough nodes to Decode()?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+    auditRecords.SuccessNodeIDs = successNodes
+  }
+
+  // remove the set the contained flag to false on the NodeDossier and update the overlay (satellitedb)
+  // remove the pending audit from the ContainmentDB
+
+  return auditRecords
+```
+
+## Open Issues
```
wthorp: Can a storage node determine if they're in containment mode?  If so, can they do anything about it?  A storage node operator would want there to be a function where I can manually fix my containment mode status.

docs/design/auditing-containment.md
wthorp: Is it just upload node selection, or download/checker/repairer as well?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### Why not audit a contained node for other data?
```
wthorp: Can we do both?  Rather than storing a list of failed audits, maybe if they fail the subsequent audits we just increment the failure count even though its different data?

docs/design/auditing-containment.md
navillasa: That sounds like a good idea Bill! I originally had the interface with Get and Put, but Egon pointed out that that's susceptible to race conditions. I think it still is currently, but I think the `IncrementFail` idea could avoid that.

docs/design/auditing-containment.md
navillasa: Okay I think you're right, it should be allowed to download the pieces from the contained nodes.

docs/design/auditing-containment.md
navillasa: I probably should've included it more explicitly but yes the plan was to use another database table. But you think it would be better to add the `pendingAudit` fields to the NodeDossier or something instead?

docs/design/auditing-containment.md
navillasa: That's a good idea too!

docs/design/auditing-containment.md
navillasa: Hmm that's a good question. Currently in the existing code, an error will be returned from `verifier.Verify` if less than the required number of shares are available (this happens when `f.Correct` is called).
It seems not good to just return an error without "containing" this node. But it also seems bad to "contain" the node without saving its `ExpectedShareHash`.

Maybe we could make a case where we save the segment's path to the `pendingAudit` entry and then when reverify is attempted, it could get the latest pointer and try to download again (with hopefully more nodes available)? But maybe that is too convoluted. I would be interested in hearing other thoughts.

docs/design/auditing-containment.md
navillasa: I don't think a storage node should be able to determine if they're in containment mode. It would be a field on the NodeDossier. @kaloyan-raev do you know if there's a way that a SNO could manipulate their NodeDossier to say they're not in containment mode?

docs/design/auditing-containment.md
navillasa: Following our conversation this morning, it seems like it's fine for contained nodes to be selected (only the audit service should care about contained status), so I'll remove this objective.

docs/design/auditing-containment.md
navillasa: After our conversation this morning, it seemed like it was agreed to be a good idea that a contained node could be audited for other data as well. My initial thought was that we should keep track of which reverification failures correlate to which data to reduce possible confusion/misdirection about which pieces are missing, but just incrementing reverification failures could also be a quicker route to finding if a node needs to be disqualified. Would be open to more thoughts here.

docs/design/auditing-containment.md
navillasa: Yeah, I think Kaloyan had suggested we do experiments to find what errors are returned by DialNode in the different use cases, so I'll expand this section in the doc.

docs/design/auditing-containment.md
navillasa: I'm a little concerned though that `IncrementFail` could be confused with an "audit failure" since this isn't technically a failure yet. Does `IncrementPending` seem worse?

docs/design/auditing-containment.md
navillasa: Actually I'm just going to call this Delete since it's just deleting and not really resetting.

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+  IncrementPending(ctx context.Context, pendingAudit pendingAudit) error
+  Delete(ctx context.Context, nodeID pb.NodeID) error
+}
+```
+
+`Get` will be used in the Reverify function to get the `pendingAudit` information from the ContainmentDB needed to compare the expected audit result with the actual.
+
+`IncrementPending` is an upsert that is used when an "uncooperative" node is identified during the normal process, and when a node doesn't pass a reverification.
+
+`Delete` will be used to delete the pendingAudit from the ContainmentDB after the node either passes or fails the audit.
```
mobyvb: Maybe it could also update the node dossier

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+    expectedShareHash []byte
+    reverifyCount     int // number of times reverify has been attempted
+}
+```
+`pendingAudit` entries are created and added to the ContainmentDB when a node seems to be online and opens a connection with the satellite to be audited (this happens within the existing Verify function in the audit package), but then refuses to send the requested erasure share.
+
+Additionally, the satellite should make sure to empty the ContainmentDB of certain pending audits when segments are deleted.
+
+```go
+type ContainmentDB interface {
```
mobyvb: Shouldn't there be a `Create` function as well?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+#### 3. Contained nodes are checked for the same data that they were originally requested to respond with:
+
+```go pkg/audit/verifier.go
+func (verifier *Verifier) Reverify(ctx context.Context, node storj.NodeID) (auditRecords *RecordAuditsInfo, err error) {
+
+  pendingAudit, err := checker.containment.Get(id)
+
+  // Dial the node, then use info from the pendingAudit to download the target share from the node.
+
+  offset := node.stripeIndex * node.shareSize
```
mobyvb: this info should come from `pendingAudit`, not `node`, right?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe) (verifiedNodes *RecordAuditsInfo, err error) {
+  defer mon.Task()(&ctx)(&err)
+
+  pointer := stripe.Segment
+
+  pieces := pointer.GetRemote().GetRemotePieces()
+
+  // Get the NodeDossiers from the overlay using the pieces’ nodeIDs.
+  // Determine if they have the contained flag set to true.
+  // If so, call Reverify on them.
```
mobyvb: How should we combine the `verifiedNodes` from `Verify()` with the `verifiedNodes` returned from all the different `Reverify()` calls? What if there are duplicates between them?

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Identifying Nodes That Need to Be Contained
+In the audit verifier, we need a better system for handling the different cases in which a storage node may not respond to an audit request.
+Here are a few possibile cases:
+1. The node is busy fulfilling other requests and can't respond to the audit request within the audit timeout defined on the Satellite.
+2. The node does not have the audit piece anymore because the SNO deleted it.
+3. The node can't read the piece due to a file permission issue (a SNO config mistake).
+4. The node can't read the piece due to bad sectors on the HDD.
+
+For 1, the node doesn't necessarily need to be contained. For 2, the node does need to be contained.
```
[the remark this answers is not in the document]
navillasa: I'm not sure I did the best job but I was just trying to incorporate what Kaloyan said here: https://github.com/storj/storj/pull/1929#discussion_r282419628

docs/design/auditing-containment.md
navillasa: yeah good catch

docs/design/auditing-containment.md
```diff
@@ -0,0 +1,205 @@
...
+For 3 and 4, the node should probably also be contained.
+
+We need to find out what the error messages are for these different cases so that we don't lump the first case with the rest.
+
+### Handling Contained Nodes
+New functionality needs to be added to the audit verifier, where upon receiving a pointer, the verifier should first iterate through the nodes listed in the pointer and check if they have a `contained` field set to true on their NodeDossier or not.
+
+For nodes that are not contained, the audit will continue as normal.
+
+For nodes that are contained, the verifier will ask the ContainmentDB for the hash of the erasure share expected from the node when it was originally audited.
```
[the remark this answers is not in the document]
navillasa: yeah good point, I kind of glossed over that here.

docs/design/auditing-containment.md
navillasa: I think @wthorp suggested that `IncrementPending` (or `IncrementFail`) should be used for both creating and updating `pendingAudit` entries.

docs/design/auditing-containment.md
navillasa: that's a good idea!

docs/design/auditing-containment.md
navillasa: That's a good question... maybe instead of an array of verified nodes, it could be a map and we check to see if there's any overlap before shipping the records off to the satellitedb?

docs/design/auditing-containment.md
wthorp: Is there a reason not to make them part of NodeDossier?

docs/design/auditing-containment.md
wthorp: Why?  Clever people can certainly just read the source code and reason.  Its the not-so-clever people I'm worried about.

docs/design/auditing-containment.md
wthorp: sure!

docs/design/auditing-containment.md
navillasa: The DB as a place for keeping data specifically related to containment just seemed more intuitive to me since I don't think the `pendingAudit` fields would be needed by other services, but no I don't see a major reason for not putting them on NodeDossier. I would like to hear @kaloyan-raev 's thoughts too though.

wthorp: Audit stuff already gets written to NodeDossier, right?  I'd like to avoid treating containment mode as its own service / thing.

wthorp: assuming you'll implement design from review meeting

docs/design/auditing-containment.md
wthorp: seems to maybe be out of scope for now

docs/design/auditing-containment.md
navillasa: ended up adding step 4 to the implementation steps to explain how this would happen separately.

kaloyan-raev:
Late checking the final draft - this document is great!

PR #1933, opened 2019-05-09
jenkins: peer constraints

egonelbre:
Check whether any of the libraries import any of the peers. It's easy to introduce such a dependency right now.

Currently doesn't fail the linter since we still have violations in our code.

Thanks for submitting a PR!

### DIFF
--- a/scripts/check-peer-constraints.go
+++ b/scripts/check-peer-constraints.go
@@ -16,6 +16,7 @@ import (
 )
 
 var race = flag.Bool("race", false, "load with race tag")
+var fail = flag.Bool("fail", true, "fail on violation")
 
 func main() {
 	flag.Parse()
@@ -48,7 +49,9 @@ func main() {
 		}
 	}
 
-	os.Exit(exitcode)
+	if *fail {
+		os.Exit(exitcode)
+	}
 }
 
 func load(names ...string) ([]*packages.Package, error) {
--- a/scripts/check-peer-constraints.go
+++ b/scripts/check-peer-constraints.go
@@ -85,6 +85,7 @@ func links(source, destination []*packages.Package) bool {
 			}
 			visited[id] = true
 			if targets[id] {
+				links = true
 				fmt.Printf("import %q\n", pathstr(append(path, pkg, imp)))
 				continue
 			}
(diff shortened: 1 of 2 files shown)


scripts/check-peer-constraints.go
```diff
@@ -0,0 +1,109 @@
...
+	var buildFlags []string
+	if *race {
+		buildFlags = append(buildFlags, "-race")
+	}
+
+	return packages.Load(&packages.Config{
+		Mode:       packages.LoadImports,
+		Env:        os.Environ(),
+		BuildFlags: buildFlags,
+		Tests:      false,
```
zeebo: I think we might want to include tests as well.

scripts/check-peer-constraints.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			visit(imp, append(path, pkg))
+		}
+	}
+
+	for _, pkg := range source {
+		visit(pkg, nil)
+	}
+
+	return links
```
zeebo: when is links ever set to true?

scripts/check-peer-constraints.go
egonelbre: We a have a lot of system level tests. To test uplink fully we need to have a running storage node and satellite.

scripts/check-peer-constraints.go
egonelbre: ah, missed setting it to true.

scripts/check-peer-constraints.go
zeebo: Dang. Works for me. Additionally, if we ever have problems we can iterate on catching them here.

scripts/check-peer-constraints.go
```diff
@@ -0,0 +1,110 @@
...
+
+func main() {
+	flag.Parse()
+
+	var exitcode int
+
+	peers, err := load(
+		"storj.io/storj/satellite/...",
+		"storj.io/storj/storagenode/...",
+		"storj.io/storj/bootstrap/...",
```
stefanbenten: Should we add the versioncontrol here as well?
It doesnt have any sub packages yet, but that could happen at any stage

egonelbre: added

PR #1935, opened 2019-05-09
Remove UseIdentity and IdentityVersion from libuplink volatile config

kaloyan-raev:
What: 

- Removes `UseIdentity` and `IdentityVersion` from libuplink `Config.Volatile`. We shouldn't have aby use case for these two anymore.
- storj-sim does not generate identity for gateway anymore
- Gateway properly closes the libuplink project when verifying the satellite config

Why:

These changes address additional comments left in #1920 and #1921.

Thanks for submitting a PR!

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -141,13 +141,7 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		zap.S().Error("Failed to initialize telemetry batcher: ", err)
 	}
 
-	proj, err := runCfg.openProject(ctx)
-	if err != nil {
-		return fmt.Errorf("Failed to contact Satellite.\n"+
-			"Perhaps your configuration is invalid?\n%s", err)
-	}
-
-	_, err = proj.ListBuckets(ctx, &storj.BucketListOptions{Direction: storj.After})
+	err = checkCfg(ctx)
 	if err != nil {
 		return fmt.Errorf("Failed to contact Satellite.\n"+
 			"Perhaps your configuration is invalid?\n%s", err)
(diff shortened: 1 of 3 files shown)


cmd/gateway/main.go
```diff
@@ -165,6 +159,21 @@ func generateKey() (key string, err error) {
...
 
+func checkCfg(ctx context.Context) (err error) {
+	proj, err := runCfg.openProject(ctx)
+	if err != nil {
+		return err
+	}
+	defer func() { err = errs.Combine(err, proj.Close()) }()
+
+	_, err = proj.ListBuckets(ctx, &storj.BucketListOptions{Direction: storj.After})
+	if err != nil {
```
mniewrzal: nit: you can do just
```
_, err = proj.ListBuckets
return err
```

kaloyan-raev: You are absolutely right. Fixed.

thepaul:
I'm just seeing this now. Can we consider reverting the change? I think it needs to be possible to supply your own identity. Definitely not required, but it should be an option.

I want to be able to specify things like "uplink_identity=SomeIDHash" in macaroons. That might end up being the best way to implement access control restrictions that depend on authn of the requester. It would require some more plumbing to be user-friendly (maybe making some way of telling the satellite to associate email address A with identity B, and a way of querying that), but it seems a lot better than configuring an uplink to send a username and password or something.

Also, there might be cases where a user wants uplink setup to be as fast as possible. I imagine they'd want to pre-generate their identity and use it in the config.

ifraixedes:
It sounds an interesting use case.
I would say, rather than reverting it, adding back the identity outside of volatile making optional and add to the docs when they are good to be set.

jtolds:
> I want to be able to specify things like "uplink_identity=SomeIDHash" in macaroons.

oh, neat! that's a great idea.

jtolds:
that sounds useful enough that we would want to promote it out of volatile anyway? should we do a PR that adds identity as a first class setup option?

thepaul:
Sounds ok to me.

ifraixedes:
It has been registered under the issue [V3-1785](https://storjlabs.atlassian.net/browse/V3-1785?utm_term=&utm_source=jira-slack&utm_medium=referral-external&atlOrigin=eyJpIjoiNjZkYjE0NTE0YzM1NDBlNDkyODdiN2MwZDk1ZDcyNmEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

PR #1937, opened 2019-05-09
pkg/kademlia: fix timeout test slowness

zeebo:
The timeout tests were configured to use very short timeouts but
for some reason they took many seconds to complete. This commit
fixes two issues:

1. The transports were always using the default timeout rather than
   the timeout specified.

2. The tests were possibly calling t.FailNow inside of the non-test
   goroutine which causes it to exit, possibly losing an error from
   the error group. Additionally, it didn't seem to be testing that
   the error came back as a deadline wrapped in a transport error.

The tests run in ~3s instead of ~60s now.

### DIFF
--- a/pkg/kademlia/dialer_test.go
+++ b/pkg/kademlia/dialer_test.go
@@ -177,7 +177,7 @@ func TestSlowDialerHasTimeout(t *testing.T) {
 	// TODO: also use satellites
 	peers := planet.StorageNodes
 
-	{ // PingNode
+	func() { // PingNode
 		self := planet.StorageNodes[0]
 
 		tlsOpts, err := tlsopts.NewOptions(self.Identity, tlsopts.Config{})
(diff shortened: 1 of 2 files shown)


pkg/kademlia/dialer_test.go
```diff
@@ -177,7 +177,7 @@ func TestSlowDialerHasTimeout(t *testing.T) {
 	// TODO: also use satellites
 	peers := planet.StorageNodes
 
-	{ // PingNode
+	func() { // PingNode
```
JessicaGreben: why do you wrap these in `func`?

zeebo: So that the defer's inside execute immediately rather than all at once at the end of the test.

JessicaGreben: ah yes makes sense. good one

pkg/kademlia/dialer_test.go
```diff
@@ -203,15 +203,15 @@ func TestSlowDialerHasTimeout(t *testing.T) {
 			peer := peer
 			group.Go(func() error {
 				_, err := dialer.PingNode(ctx, peer.Local().Node)
-				require.Error(t, err, context.DeadlineExceeded)
-				require.True(t, transport.Error.Has(err))
-
+				if !transport.Error.Has(err) || errs.Unwrap(err) != context.DeadlineExceeded {
+					return errs.New("invalid error: %v", err)
+				}
```
JessicaGreben: i'm still not totall clear why this wouldn't work: 
```
require.Equal(t, errs.Unwrap(err), context.DeadlineExceeded)
require.True(t, transport.Error.Has(err))
```

egonelbre: Because `group.Go` starts things in a goroutine. And any method call on `*testing.T` results in a data race.

zeebo: Oh, I think what you wrote would work, and I over-matched on a heuristic about causing `FailNow` to happen in non-test goroutines. That's been an endless source of bugs for me, so I try to avoid it.

PR #1938, opened 2019-05-09
init ref admin server & setup 404 routing

fuskovic:
What: Initializing the GUI for the Referral Program Admin portal

Why: So Admin's will have a place to go to modify referral offers.

Thanks for submitting a PR!


satellite/referral/offersweb/pages/index.html
```diff
@@ -0,0 +1,24 @@
+{{define "head"}}
```
VinozzZ: Just an idea and would like to have some input from others for this. Should we move the templates into the `/web` folder and then create a config for the static assets folder path?

satellite/referral/offersweb/server.go
```diff
@@ -4,18 +4,24 @@
...
 	"github.com/zeebo/errs"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 )
 
 // Error is satellite referral error type
-var Error = errs.Class("satellite referral error")
+var (
+	Error = errs.Class("satellite referral error")
+	_, fp, _, _ = runtime.Caller(0)
```
VinozzZ: See this comment for file path https://github.com/storj/storj/pull/1938/files#r282598760

satellite/referral/offersweb/pages/index.html
stefanbenten: I like that idea!

satellite/referral/offersweb/pages/404.html
```diff
@@ -0,0 +1,3 @@
+{{define "main"}}
```
stefanbenten: You may want to reuse the page from the Satellite UI

satellite/referral/offersweb/server.go
stefanbenten: You may want to check if the last value as well here

satellite/referral/offersweb/pages/index.html
```diff
@@ -0,0 +1,24 @@
...
+    <title>Referral Admin Portal</title>
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <link rel="stylesheet" type="text/css" href="../static/css/style.css">
```
stefanbenten: We may be able to use the CSS code of the Kiev Team, to have the same look and feel. Having another UI with different UX experience feels wrong to me

satellite/referral/offersweb/static/css/style.css
```diff
@@ -0,0 +1,9 @@
+html{
+    height:100vh;
+    width:100vw;
+}
+
+body{
+    height:100%;
+    width:100%;
+}
```
stefanbenten: Would be nice if we can take this out and use the exisiting CSS data..

satellite/referral/offersweb/pages/index.html
fuskovic: @stefanbenten as far as structuring paths, we currently have ```satellite/referral/offersweb/server.go```
is there a reason it can't be ```satellite/referral/server.go``` just trying to get a feel for our do's and don'ts for path structuring.

satellite/referral/offersweb/pages/index.html
fuskovic: that makes sense

satellite/referral/offersweb/server.go
fuskovic: @stefanbenten May I ask why ? It's the ok bool value but is there a scenario that it would ever return false? If passing 0 to caller gets the execution stack for the file invoking it then how would it ever be able to return false ?

satellite/referral/offersweb/pages/index.html
```diff
@@ -0,0 +1,24 @@
+{{define "head"}}
+    <title>Referral Admin Portal</title>
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
```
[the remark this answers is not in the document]
fuskovic: I don't think we'll actually need this at all since @stefanbenten recommended using the styles already made for the satellite.

satellite/referral/offersweb/pages/404.html
fuskovic: It doesn't look like their is a 404 page for the satellite UI

satellite/referral/offersweb/pages/index.html
fuskovic: upon review - I disagree with this. The satellite UI uses scoped css specific for each page on the satellite. In other words, they don't have a global css file. They have file specific for the login page, another file specific to the register page, etc..... I'd like to not use their styles and keep a new set of styles for the admin gui based on the mockups Atikh is making. @stefanbenten 

With that said I think a separate static folder for this service makes sense.

satellite/referral/offersweb/static/css/style.css
fuskovic: @stefanbenten The existing CSS data is scoped and won't apply to us. The look and feel will ultimately be up to DJ so as long as we style it the way he wants, having separate CSS files shouldn't be a problem. See https://github.com/storj/storj/pull/1938/files#r283005800

satellite/referral/offersweb/server.go
stefanbenten: I am fine if there is a valid explanation for it.
Please comment your reason above it for everyone, then youre good! 👍

satellite/referral/offersweb/pages/index.html
stefanbenten: Maybe you should collaborate with @tikhbana about this. I dont think it's a good idea to have x different UI flows and looks. Its just a recommendation tho.
We might want to unifi it in an additional PR then.

fuskovic: I think you may be misinterpreting me @stefanbenten. Mockups for this are already in progress. How it looks is up to DJ and @tikhbana already. What I'm saying is the satellite UI does not use a global css file. It uses scoped css for each of it's different pages example : login.scss, register.scsss. Using the scoped css file for another page on another service doesn't make sense since those classes would not apply.

tikhbana:
Does this bootstrap file contain styling or media queries, if so it might be worth exploring the current scss files for consistency. Might be a different page but it’s all the same dashboard

--
Best,
Atikh Bana

satellite/referral/offersweb/server.go
rikysya: That file path will be bound to your local code location, that may not work for precomputed binaries. Also you set hard rule for path of your static files to be in folder where `server.go` was during compilation. But I am not sure if we care about making static files configurable this much

satellite/referral/offersweb/pages/index.html
rikysya: The idea of having a `web` was to separate server static files (`html`, `js`, etc) from the rest of `go` code

PR #1939, opened 2019-05-09
repair segment reassess it missing pieces just before repair

aligeti:
What: 
Reassess what pieces needs to be rebuilt in repair just before repairs, so newly built pieces are not just overlap and also not sent to the same storagenode. 
Why:
There is a possibility that a piece is built and sent to already existing storagenode, causing info.db UNIQUE error
https://storjlabs.atlassian.net/browse/V3-1650
Thanks for submitting a PR!

### DIFF
--- a/pkg/datarepair/repairer/repairer.go
+++ b/pkg/datarepair/repairer/repairer.go
@@ -50,7 +50,7 @@ func (c Config) GetSegmentRepairer(ctx context.Context, tc transport.Client, met
 
 // SegmentRepairer is a repairer for segments
 type SegmentRepairer interface {
-	Repair(ctx context.Context, path storj.Path, lostPieces []int32) (err error)
+	Repair(ctx context.Context, path storj.Path) (err error)
 }
 
 // Service contains the information needed to run the repair service
@@ -124,7 +124,7 @@ func (service *Service) process(ctx context.Context) error {
 		}
 
 		service.Limiter.Go(ctx, func() {
-			err := service.repairer.Repair(ctx, seg.GetPath(), seg.GetLostPieces())
+			err := service.repairer.Repair(ctx, seg.GetPath())
 			if err != nil {
 				zap.L().Error("repair failed", zap.Error(err))
 			}
--- a/pkg/storage/segments/repairer.go
+++ b/pkg/storage/segments/repairer.go
@@ -66,13 +66,19 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
 	var excludeNodeIDs storj.NodeIDList
 	var healthyPieces []*pb.RemotePiece
 	pieces := pointer.GetRemote().GetRemotePieces()
-	missingPieces, err := repairer.updateMissingPieces(ctx, pieces)
+	missingPieces, err := repairer.cache.GetMissingPieces(ctx, pieces)
 	if err != nil {
 		return Error.New("error getting missing pieces %s", err)
 	}
+
 	numHealthy := len(pieces) - len(missingPieces)
+	// irreparable piece
+	if int32(numHealthy) < pointer.Remote.Redundancy.MinReq {
+		return Error.New("piece cannot be repaired")
+	}
+
+	// repair not needed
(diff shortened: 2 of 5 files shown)


aligeti:
> Is the use case described in this ticket even possible after #1717?

The bug ticket was logged very recently. 

My understanding of the root cause to the problem is we repaired a good segment (at the time of actual repair) because the segment that was entered into the queue as injured was not injured anymore at the time of the repair. If you try to repair and upload to the same node, you will get the "UNIQUE" error. I looked at the #1717, I don't think we addressed the root cause. 

The root cause fix, I understand is at the time of repair , we actually find what are the injured segments, do similar to what the checker.go does. When you find what nodes are online, there could be a chance that the nodes which were offline at the time of checker are now online at the time of actual repair.

Let me know what you think of my thought process.

pkg/storage/segments/repairer.go
```diff
@@ -157,3 +168,23 @@ func createBucketID(path storj.Path) ([]byte, error) {
 	}
 	return []byte(storj.JoinPaths(comps[0], comps[2])), nil
 }
+
+func (repairer *Repairer) updateMissingPieces(ctx context.Context, pieces []*pb.RemotePiece) (missingPieces []int32, err error) {
```
mobyvb: this is identical to the `getMissingPieces` function in `datarepair/checker/chacker.go`
Maybe it would be good to move both to overlay cache and call the same function?

mobyvb: this is identical to the getMissingPieces function in `datarepair/checker/chacker.go`
Maybe it would be good to move both to overlay cache and call the same function?

pkg/datarepair/repairer/repairer.go
```diff
@@ -50,7 +50,7 @@ func (c Config) GetSegmentRepairer(ctx context.Context, tc transport.Client, met
 
 // SegmentRepairer is a repairer for segments
 type SegmentRepairer interface {
-	Repair(ctx context.Context, path storj.Path, lostPieces []int32) (err error)
```
mobyvb: If we are removing `lostPieces` as an argument, we may as well remove it from the proto (`pb/datarepair.proto`)

aligeti: I agree, but on a second thought, it could be used for some metrics later on. Also I wanted to make small changes. Removing adds up few more file(s) changes. I just wanted to do incremental change.

pkg/storage/segments/repairer.go
aligeti: I will make the change.

wthorp:
> > Is the use case described in this ticket even possible after #1717?

Can you create a test that proves that this is possible?  That way we know for sure it is fixed?

aligeti:
> > > Is the use case described in this ticket even possible after #1717?
> 
> Can you create a test that proves that this is possible? That way we know for sure it is fixed?
I am trying to add the testcase but I have a feeling it will be overkill and complicating a simple test case.  My changes require, I need to bring online some of the previously offline nodes. The current test case is just a sequential test case. I m not sure how to test this where in fact the inputs and outputs out of repair() function are same even after the change. Without any modification to the test case, the original test cased tested the slightly modified repair function correctly . The current test case is good enough for now, unless we want to complicate the test case by creating multi go routines to emulate the node coming back online before repair() starts. Not sure for such a small change, we need to make a considerable change in test case.

storagenode/piecestore/endpoint.go
```diff
@@ -267,8 +267,12 @@ func (endpoint *Endpoint) Upload(stream pb.Piecestore_UploadServer) (err error)
 					Uplink:          peer,
 				}
 
-				if err := endpoint.pieceinfo.Add(ctx, info); err != nil {
-					return ErrInternal.Wrap(err)
+				// check if piece already exists
+				_, err := endpoint.pieceinfo.Get(ctx, info.SatelliteID, info.PieceID)
```
egonelbre: Let's don't do it in this PR.

aligeti: Reverted the changed

PR #1940, opened 2019-05-09
transport: separate dial from request timeouts

zeebo:
## What / Why

A previous change reused the same timeout for dialing as well as
requesting in order to speed up some tests. This change introduces
a distinct timeout so that the different operations can have different
timeouts.

Discussed some previously at https://github.com/storj/storj/pull/1545#discussion_r268168062

---

**Note to reviewers:** I changed a config struct. I assume there's something I have to do about flags or whatever, but I don't know.

### DIFF
--- a/pkg/transport/transport.go
+++ b/pkg/transport/transport.go
@@ -30,24 +30,37 @@ type Client interface {
 	WithObservers(obs ...Observer) Client
 }
 
+// Timeouts contains all of the timeouts configurable for a transport
+type Timeouts struct {
+	Request time.Duration
+	Dial    time.Duration
+}
+
 // Transport interface structure
 type Transport struct {
-	tlsOpts        *tlsopts.Options
-	observers      []Observer
-	requestTimeout time.Duration
+	tlsOpts   *tlsopts.Options
+	observers []Observer
+	timeouts  Timeouts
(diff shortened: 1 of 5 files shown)


pkg/transport/transport.go
```diff
@@ -30,24 +30,37 @@ type Client interface {
...
-	requestTimeout time.Duration
+	tlsOpts   *tlsopts.Options
+	observers []Observer
+	timeouts  Timeouts
 }
 
 // NewClient returns a transport client with a default timeout for requests
 func NewClient(tlsOpts *tlsopts.Options, obs ...Observer) Client {
-	return NewClientWithTimeout(tlsOpts, defaultRequestTimeout, obs...)
+	return NewClientWithTimeouts(tlsOpts, Timeouts{}, obs...)
```
navillasa: maybe Timeouts should have the defaults set instead of being empty, unless it happens elsewhere? otherwise won't it be set to 0?

zeebo: Yep, the defaults are set in the `NewClientWithTimeouts` function so that no one can accidentally specify 0.

navillasa: OH I see. okay great!

PR #1941, opened 2019-05-09
kademlia audit gating design draft

jenlij:
What: Design document for the kademlia audit gating system.

Why: To get feedback before making JIRA tickets.

Thanks for submitting a PR!


docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,94 @@
...
+
+StorageNode B is added to StorageNode A's routing table only if StorageNode A trusts Satellite C and Satellite C as verified StorageNode B has a high enough identity-generate difficulty (CPU proof-of-work) and disk space. If StorageNode B does not have this verification, it is added to StorageNode A's Routing Table Antechamber.
+
+## Background
+
+### White paper: 4.6.1 Mitigating Sybil attacks
+While we’ve adopted the proof-of-work scheme S/Kademlia proposes to partially address Sybil attacks, we extend Kademlia with an application specific integration to further defend our network. 
+
+
+Given two storage nodes, A and B, storage node B is not allowed to enter storage node A’s routing table until storage node B can present a signed message from a Satellite C that storage node A trusts claiming that B has passed enough audits that C trusts it (sections 4.13 and 4.15). This ensures that only nodes with verified disk space have the opportunity to participate in the routing layer.
```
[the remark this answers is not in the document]
navillasa: Also, is "passed enough audits" definitely the criteria that's measured? Because I see this first goal below: "Add a signed message from Satellites to authenticate whether a Node has a high enough ID generation difficulty and disk space." While these are certain qualifications to meet, I don't think I would describe this evaluation as an audit just because that might lead to confusion with the audit system. Unless I'm misunderstanding and we are also checking for how many successful audits they have on record?

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,94 @@
...
+
+**Node Neighborhood** - *The k-closest nodes to self where distance is measured by XOR. A node is within the node neighborhood if it is closer than the furthest node in the neighborhood. The vetted node neighborhood is the k-closest nodes that are currently in the Routing Table.*
+
+[TODO Node Neighborhood Diagram]
+
+## Design
+
+1. Satellite Signatures for Node Verification
+    - Identities can sign messages already
+    - Create a Message (protobuf) to sign that Satellite C says Node B has been vetted
```
navillasa: I think a draft of the message struct could be good to see here so we can decide fields and potential methods in advance.

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,94 @@
...
+    - Get info about the Node's difficulty and disk space to add to the message
+
+2. Trusted Satellites List
+    - Create Whitelist/blacklist with an abstraction layer for trusted/untrusted Satellites
+    - These lists will live on each Node
+
+3. Routing Table Antechamber
+    - XOR ordered data structure (perhaps an ordered slice)
+    - A node can be added if it would be within the vetted node neighborhood
+    - Once a node has been verified, it broadcasts its new status to try to join routing tables
```
navillasa: how does it broadcast its status to join routing tables? also, once a node is in an antechamber, how does its status change to verified? is it periodically re-checked by the Satellite?

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,94 @@
...
+
+5. [Update FindNear](https://storjlabs.atlassian.net/browse/V3-1729)
+
+## Open issues 
+
+Q: Should closer buckets to self get refreshed more frequently?
+
+A: ?
+
+Q: If a node is removed from the Routing Table, should it be added back to the antechamber?
```
navillasa: maybe it depends on if the node got removed because of not having enough disk space or high enough ID difficulty vs. if it got removed because it was too far away? not sure, it seems like all of those factors could change, so maybe it's okay to let it back in? maybe if a node was disqualified for audit failures then it shouldn't be let back in.

docs/design/kademlia_audit_gating.md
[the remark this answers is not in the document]
jenlij: That is very true. I think initially we had intended it to be directly based on the audit system, but after further discussion, we decided to make slightly different. This background section is directly taken from the whitepaper.

docs/design/kademlia_audit_gating.md
jenlij: good idea

docs/design/kademlia_audit_gating.md
jenlij: This is a good question and the solution is still open for discussion. Hopefully we'll flesh that out  and add it to the design soon.

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,116 @@
+# Kademlia Audit Gating
+
+## Abstract
+
+StorageNode B is added to StorageNode A's routing table only if StorageNode A trusts Satellite C and Satellite C has verified StorageNode B is not disqualified and has a high enough audit count and uptime count. If StorageNode B does not have this verification, it is added to StorageNode A's Routing Table Antechamber.
```
wthorp: "... and has a high enough audit count and uptime count" is likely implicit in "is not disqualified".  I would make these details parenthetical.  Later versions of disqualification may include other criteria.

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,116 @@
...
+    - Create Whitelist/blacklist with an abstraction layer for trusted/untrusted Satellites
+    - These lists will live on each Node
+    - NB: We are using `config.Storage.WhitelistedSatelliteIDs` "a comma-separated list of approved satellite node ids" for now
+
+3. Routing Table Antechamber
+    - XOR ordered data structure (boltdb bucket)
+    - A node can be added if it would be within the vetted node neighborhood
+    - Once a node has been verified, it broadcasts its new status to try to join routing tables
+    - A Node may enter a Routing Table directly if at first contact it is already verified by a trusted Satellite.  
+    - A Node may be removed from a Routing Table if on bucket refresh it no longer meets qualifications
```
wthorp: nit: "may" is probably not the word you want here.  "should"?

docs/design/kademlia_audit_gating.md
```diff
@@ -0,0 +1,116 @@
...
+
+StorageNode B is added to StorageNode A's routing table only if StorageNode A trusts Satellite C and Satellite C has verified StorageNode B is not disqualified and has a high enough audit count and uptime count. If StorageNode B does not have this verification, it is added to StorageNode A's Routing Table Antechamber.
+
+## Background
+
+### White paper: 4.6.1 Mitigating Sybil attacks
+While we’ve adopted the proof-of-work scheme S/Kademlia proposes to partially address Sybil attacks, we extend Kademlia with an application specific integration to further defend our network. 
+
+
+Given two storage nodes, A and B, storage node B is not allowed to enter storage node A’s routing table until storage node B can present a signed message from a Satellite C that storage node A trusts claiming that B has passed enough audits that C trusts it (sections 4.13 and 4.15). This ensures that only nodes with verified disk space have the opportunity to participate in the routing layer.
```
wthorp: Kad supports multiple satellites and each satellite determines disqualification individually.  What happens when a SN is trusted by one satellite and not another?  Is it in the antechamber or the routing table?

docs/design/kademlia_audit_gating.md
wthorp: "No longer meets qualification" sounds like the satellite's concept of "disqualified", which is a permanent state.  The antechamber provides a "you're not vetted functionality" similar to the "new node" status on the satellite.  This antechamber idea seems to be slightly lacking a "disqualified" feature.

A 'bad' SN should never be trusted to provide a signed "I am disqualified" message.  However, in some cases it makes sense for a SN that is disqualified on one satellite to remain in Kad, for instance because they're not DQ'ed on another satellite.  Thus a 'good' satellite might.

Perhaps its a marginal and distributed inefficiency to _forever_ ask these 'bad' SN's if they're vetted.  Yet, it could be common for another SN to be DQ'ed on all of a SN's whitelisted satellites.  [Or to simply not whitelist those satellites at all... imagine if I started a knockoff _Fartigrade_ network.]  Hrmm.

(Going out on a limb here: Maybe this a use case for storing info in kad?)

docs/design/kademlia_audit_gating.md
cam-a: This is a good question. For now, I think it just needs to be trusted by one satellite. The vetting process is to mitigate sybil attacks, so if the node passed one satellite's vetting requirements: audit count, uptime count, not disqualified (there's also work for future parameters of minimum amount of data stored for a minimum amount of time), then it's probably a legitimate node

However, that brings up the question: what if there is a bad satellite? Maybe with the explicit purpose of setting an extremely low bar for issuing vouchers? Hopefully you wouldn't have that satellite in your trusted list and wouldn't accept a voucher from it

docs/design/kademlia_audit_gating.md
cam-a: Would falsification of the info be a concern if it's propagated through kademlia?

docs/design/kademlia_audit_gating.md
wthorp: Yeah, this would be a bit of a problem if you weren't whitelisting satellites... one faker satellite would invalidate the whole shebang.

docs/design/kademlia_audit_gating.md
wthorp: I'd expect it'd still have to be signed by satellites, but I admit I haven't really thought this through.

docs/design/kademlia_audit_gating.md
cam-a: section 4.18 of the white paper vaguely discusses a satellite reputation system, so it might be the case that these faker satellites shouldn't make it into your trusted list in the first place

docs/design/kademlia_audit_gating.md
jenlij: got it!

docs/design/kademlia_audit_gating.md
jenlij: you're right

docs/design/kademlia_audit_gating.md
jenlij: Update: When a node talks to another for discovery purposes, it also sends back all of the valid vouchers it has.
If that node does have any trustworthy vouchers, it gets "antechambered".

jenlij: Edit: rather than node discovery, we mean routing table population/refresh.

docs/design/kademlia_audit_gating.md
jenlij: Lol @ Fartigrade

Yeah, I'm thinking we don't want to implement a disqualification system, but rather the validity of the vouchers expire, and if a node doesn't have a non-expired voucher from a trusted satellite, it gets booted.

docs/design/kademlia_audit_gating.md
jenlij: Yeah right now we're relying on whitelisting satellites.

docs/design/kademlia_audit_gating.md
wthorp: ok, well get with us on monday ... we have disqualification 90% done

egonelbre:
This uploaded 1.7MB file to the repository.

PR #1945, opened 2019-05-10
server side macaroons

zeebo:
What: Adds macaroon support to the server side

Why: So that api keys are now macaroons


satellite/console/service.go
```diff
@@ -651,14 +655,20 @@ func (s *Service) CreateAPIKey(ctx context.Context, projectID uuid.UUID, name st
 		return nil, nil, ErrUnauthorized.Wrap(err)
 	}
 
-	key, err := CreateAPIKey()
+	secret := make([]byte, 32)
```
rikysya: Don't we have a function like `macaroon.NewSecret`? May be we can use it here?

satellite/console/service.go
```diff
@@ -675,14 +676,20 @@ func (s *Service) CreateAPIKey(ctx context.Context, projectID uuid.UUID, name st
 		return nil, nil, ErrUnauthorized.Wrap(err)
 	}
 
-	key, err := CreateAPIKey()
+	secret, err := macaroon.NewSecret()
```
jtolds: oh? a new secret for every key? okay

zeebo: Turns out that's easiest. /shrug

zeebo:
Those are good questions. 

From a UX perspective I could see wanting to show revoked key names in the UI? I dunno. I'm unaware of any guidance on how revocations should work, so I did what seemed obvious to me. Deleting seems more obvious.

It's a small diff to change to either. Happy to defer.

jtolds:
The UX right now is to delete API keys to revoke them, so let's just do the less-LOC approach for now and bring back stuff if we need it

thepaul:
Just to clarify, we will be able to revoke by tail, right? It doesn't seem to make sense by head.

zeebo:
Yeah. The head is just the root-most tail. 💥🤕

thepaul:
Oh. I probably have my terminology wrong. Is tail not the same thing as the HMAC/"signature" part?

zeebo:
Oh yep, you're right. Also, the head is exactly the secret. So yeah unless every macaroon has a distinct secret (as is implemented in this PR), then revoking by the head doesn't make sense.

zeebo:
Nope I'm wrong again. Secret is used twice in the function and I got confused. Head is a random secret unique to each root macaroon, so I think it makes sense to revoke it.

zeebo:
It's equivalent to revoking by the root-most tail, though, so yeah no reason to do anything but tails.

jtolds:
If you want to do the tail thing you can revert this: https://github.com/storj/storj/pull/1862/commits/87f2b67f020d6464a1adae0c1bad984f9f5bbeef

zeebo:
This is ready for a review.

PR #218, opened 2018-08-10
adds stripe_size field to pointerdb proto file

navillasa:
Adds `erasure_share_size` field to the RedundancyScheme message in the pointerdb proto file.
Erasure share size is now added to remote pointers in segments store (see makeRemotePointer function).
Adjusts tests in segmentStore accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/storj/storj/218)

PR #1955, opened 2019-05-13
Added the irreparable repair functionality

aligeti:
What: 
Irreparable logic added to fix the irreparable segments
Why:
Irreparable functionality was missing
https://storjlabs.atlassian.net/browse/V3-1720

Thanks for submitting a PR!

### DIFF
--- a/pkg/datarepair/checker/checker.go
+++ b/pkg/datarepair/checker/checker.go
@@ -213,16 +213,19 @@ func (checker *Checker) checkSegmentStatus(ctx context.Context, pointer *pb.Poin
 			return nil
 		}
 		rmtSegInfo.remoteSegmentsNeedingRepair++
-		err = checker.repairQueue.Insert(ctx, &pb.InjuredSegment{
-			Path:       path,
-			LostPieces: missingPieces,
-		})
-		if err != nil {
-			return Error.New("error adding injured segment to queue %s", err)
-		}
-		err = checker.irrdb.Delete(ctx, []byte(path))
-		if err != nil {
-			return Error.New("repair delete failed %s", err)
+		if _, err = checker.irrdb.Get(ctx, []byte(path)); err != nil {
+			err = checker.repairQueue.Insert(ctx, &pb.InjuredSegment{
+				Path:       path,
+				LostPieces: missingPieces,
+			})
+			if err != nil {
+				return Error.New("error adding injured segment to queue %s", err)
+			}
+		} else {
+			err = checker.irrdb.Delete(ctx, []byte(path))
+			if err != nil {
+				zap.L().Error("repair delete failed", zap.Error(err))
+			}
 		}
 	} else if numHealthy < redundancy.MinReq {
 		pathElements := storj.SplitPath(path)
@@ -279,10 +282,6 @@ func (checker *Checker) irrepairProcess(ctx context.Context) (err error) {
 			err := checker.checkSegmentStatus(ctx, seg[0].GetSegmentDetail(), string(seg[0].GetPath()), &rmtSegInfo)
 			if err != nil {
 				zap.L().Error("repair failed", zap.Error(err))
-			}
-			err = checker.irrdb.Delete(ctx, seg[0].GetPath())
-			if err != nil {
-				zap.L().Error("repair delete failed", zap.Error(err))
 			}
 			offset++
 		})
--- a/pkg/datarepair/checker/checker.go
+++ b/pkg/datarepair/checker/checker.go
@@ -30,40 +30,45 @@ var (
 
 // Config contains configurable values for checker
 type Config struct {
-	Interval time.Duration `help:"how frequently checker should audit segments" default:"30s"`
+	MaxRepair        int           `help:"maximum segments that can be repaired concurrently" releaseDefault:"5" devDefault:"1"`
+	Interval         time.Duration `help:"how frequently checker should audit segments" releaseDefault:"30s" devDefault:"0h0m10s"`
+	IrrepairInterval time.Duration `help:"how frequently irrepair checker should audit segments" releaseDefault:"15s" devDefault:"0h0m5s"`
 }
 
-// Checker contains the information needed to do checks for missing pieces
-type Checker struct {
-	metainfo        *metainfo.Service
-	lastChecked     string
-	repairQueue     queue.RepairQueue
-	overlay         *overlay.Cache
-	irrdb           irreparable.DB
-	logger          *zap.Logger
-	Loop            sync2.Cycle
-	durabilityStats durabilityStats
-}
-
-type durabilityStats struct {
-	remoteFilesChecked          int64
+// remoteSegmentInfo remote segment information
+type remoteSegmentInfo struct {
 	remoteSegmentsChecked       int64
 	remoteSegmentsNeedingRepair int64
 	remoteSegmentsLost          int64
 	remoteSegmentInfo           []string
 }
 
+// Checker contains the information needed to do checks for missing pieces
+type Checker struct {
+	metainfo     *metainfo.Service
+	lastChecked  string
+	repairQueue  queue.RepairQueue
+	overlay      *overlay.Cache
+	irrdb        irreparable.DB
+	logger       *zap.Logger
+	Limiter      *sync2.Limiter
+	Loop         sync2.Cycle
+	IrrepairLoop sync2.Cycle
+}
+
 // NewChecker creates a new instance of checker
-func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overlay *overlay.Cache, irrdb irreparable.DB, limit int, logger *zap.Logger, interval time.Duration) *Checker {
+func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overlay *overlay.Cache, irrdb irreparable.DB, limit int, logger *zap.Logger, repairInterval, irrepairInterval time.Duration, concurrency int) *Checker {
 	// TODO: reorder arguments
 	checker := &Checker{
-		metainfo:    metainfo,
-		lastChecked: "",
-		repairQueue: repairQueue,
-		overlay:     overlay,
-		irrdb:       irrdb,
-		logger:      logger,
-		Loop:        *sync2.NewCycle(interval),
+		metainfo:     metainfo,
+		lastChecked:  "",
+		repairQueue:  repairQueue,
+		overlay:      overlay,
+		irrdb:        irrdb,
+		logger:       logger,
+		Limiter:      sync2.NewLimiter(concurrency),
+		Loop:         *sync2.NewCycle(repairInterval),
+		IrrepairLoop: *sync2.NewCycle(irrepairInterval),
 	}
 	return checker
 }
@@ -72,13 +77,36 @@ func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overl
 func (checker *Checker) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	return checker.Loop.Run(ctx, func(ctx context.Context) error {
-		err := checker.IdentifyInjuredSegments(ctx)
+	// wait for all repairs to complete
+	defer checker.Limiter.Wait()
+	c := make(chan error)
+
+	go func() {
+		c <- checker.Loop.Run(ctx, func(ctx context.Context) error {
+			err := checker.IdentifyInjuredSegments(ctx)
+			if err != nil {
+				checker.logger.Error("error with injured segments identification: ", zap.Error(err))
+			}
+			return nil
+		})
+	}()
+
+	go func() {
+		c <- checker.IrrepairLoop.Run(ctx, func(ctx context.Context) error {
+			err := checker.irrepairProcess(ctx)
+			if err != nil {
+				checker.logger.Error("process", zap.Error(err))
+			}
+			return nil
+		})
+	}()
+
+	for err := range c {
 		if err != nil {
-			checker.logger.Error("error with injured segments identification: ", zap.Error(err))
+			return err
 		}
-		return nil
-	})
+	}
+	return nil
 }
 
 // Close halts the Checker loop
@@ -90,6 +118,8 @@ func (checker *Checker) Close() error {
 // IdentifyInjuredSegments checks for missing pieces off of the metainfo and overlay cache
 func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
+
+	var rmtSegInfo remoteSegmentInfo
 
 	err = checker.metainfo.Iterate("", checker.lastChecked, true, false,
 		func(it storage.Iterator) error {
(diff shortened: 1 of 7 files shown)


pkg/datarepair/repairer/repairer.go
```diff
@@ -135,3 +161,32 @@ func (service *Service) process(ctx context.Context) error {
...
+			return err
+		}
+
+		// irrepairabledb empty
+		if len(seg) == 0 {
+			return nil
+		}
+
+		service.Limiter.Go(ctx, func() {
+			err := service.repairer.Repair(ctx, string(seg[0].GetPath()))
```
wthorp: It seems like we should be calling checker in general instead of repairer.

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -187,15 +187,18 @@ kademlia.operator.wallet: ""
 # a node's ratio of being up/online vs. down/offline
 # overlay.node.uptime-ratio: 0.9
 
-# how frequently checker should audit segments
-# repairer.interval: 1h0m0s
+# how frequently irrepair checker should audit segments
+# repairer.irrepair-interval: 30m0s
```
wthorp: Just trying to check everything irreparable every 30 minutes seems like it would reach diminishing returns at some point... If I've checked something every 30 minutes for a year (17.5k times!)... I never give up?  I never back off?  Is there another ticket for this scope?

pkg/datarepair/repairer/repairer.go
aligeti: I am bit unclear. This is my thought process. Using the analogy of classic example, producer - consumer. 

Checker is a producer. Checker's task is to iterate thru pointerDB and distribute the pieces to two other DBs namely RepairDB and IrreparableDB. There are two processes that run parallelly to consume the entires entered by the checker into these two DBs. 

The Repair process task is to repairs the entries entered that are in repairDB. Handle these scenarios:
1. make an assessment at the time of actual repair to see what piece number are actually lost. 
    1.1. At the time of entry and at the time of actual repair, the injured segments number COULD be different. So repair the actual segments that needed to be repaired and not that were entered at some point in past time ( the assumption I made is there is a constant churn rate of nodes)
   1.2 At the time of repair, a repair segment could become irreparable or doesn't need repair.
   
The Irreparable process task does the very similar job to repair process and attempt to repair the segments that are entered in the irreparableDB. Also Handle these secenarios:
1. If the irreparable segment, at the time of repair attempt is still irreparable, then exit and retry in the next cycle.
2. if the irreparable segment at the time of repair attempt just becomes reparable, then repair it immediately and delete the entry.

The assumption made is irreparableDB is generally small and has to be constantly checked hoping that any irreparable segment becomes repairable, just repair it as soon as possible. 

Implementing these Changes as above, makes least code changes without having to refactor. 

Bill, I am not clear about you suggestion. I can give you a call we can talk more in detail.

scripts/testdata/satellite-config.yaml.lock
aligeti: I took 30min, only to wanted to run irreparable task more frequently than repairable. The time is configurable.

pkg/datarepair/repairer/repairer.go
mobyvb: @wthorp checker can't check a specific segment, but repairer can

pkg/datarepair/checker/checker.go
```diff
@@ -30,31 +30,45 @@ var (
 
 // Config contains configurable values for checker
 type Config struct {
-	Interval time.Duration `help:"how frequently checker should audit segments" default:"30s"`
+	MaxRepair        int           `help:"maximum segments that can be repaired concurrently" releaseDefault:"5" devDefault:"1"`
```
wthorp: Can we give this a better name?

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -10,6 +10,12 @@
 # how frequently checker should audit segments
 # checker.interval: 30s
 
+# how frequently irrepair checker should audit segments
+# checker.irrepair-interval: 15s
+
+# maximum segments that can be repaired concurrently
```
wthorp: Use a name that reflects that its checker and irreparable, rather than repair

pkg/datarepair/checker/checker.go
```diff
@@ -30,31 +30,45 @@ var (
...
+	Interval         time.Duration `help:"how frequently checker should audit segments" releaseDefault:"30s" devDefault:"0h0m10s"`
+	IrrepairInterval time.Duration `help:"how frequently irrepair checker should audit segments" releaseDefault:"15s" devDefault:"0h0m5s"`
```
navillasa: super nit, but I think I'd prefer to name this "IrreparableInterval" or "IrrepInterval"

pkg/datarepair/repairer/repairer.go
ifraixedes: > 2. If the irreparable segment, at the time of repair attempt is still irreparable, then exit and retry in the next cycle.

@aligeti when this loop is broken?

scripts/testdata/satellite-config.yaml.lock
aligeti: Changed to IrrdbInterval

pkg/datarepair/checker/checker.go
aligeti: changed to IrrdbInterval. Just to keep the name in consistent with short form nameing. Let me know

pkg/datarepair/checker/checker.go
```diff
@@ -30,31 +30,42 @@ var (
 
 // Config contains configurable values for checker
 type Config struct {
-	Interval time.Duration `help:"how frequently checker should audit segments" default:"30s"`
+	Interval      time.Duration `help:"how frequently checker should audit segments" releaseDefault:"30s" devDefault:"0h0m10s"`
+	IrrdbInterval time.Duration `help:"how frequently irrepairable checker should audit segments" releaseDefault:"15s" devDefault:"0h0m5s"`
```
mniewrzal: It shouldn't have db in name, it doesn't have to be related to any DB. For now maybe just use IrreparableInterval.

Also don't use `audit` in the description because it can be confused with a real audit. Maybe something like `how frequently irrepairable should recheck lost pieces` but maybe you will have better idea.

pkg/datarepair/checker/checker.go
```diff
@@ -30,31 +30,42 @@ var (
...
+}
+
+// remoteSegmentInfo remote segment information
+type remoteSegmentInfo struct {
+	remoteSegmentsChecked       int64
+	remoteSegmentsNeedingRepair int64
+	remoteSegmentsLost          int64
+	remoteSegmentInfo           []string
```
mniewrzal: no need to have prefix `remoteSegments` for fields. Type name `remoteSegmentInfo` has information about context.

pkg/datarepair/checker/checker.go
```diff
@@ -200,3 +167,113 @@ func contains(a []string, x string) bool {
...
+		})
+		if err != nil {
+			return Error.New("error adding injured segment to queue %s", err)
+		}
+
+		// if same segment previously was irreparable
+		if _, err = checker.irrdb.Get(ctx, []byte(path)); err == nil {
+			err = checker.irrdb.Delete(ctx, []byte(path))
+			if err != nil {
+				zap.L().Error("repair delete failed", zap.Error(err))
```
mniewrzal: don't use global logger, use `checker.logger`

pkg/datarepair/checker/checker.go
```diff
@@ -200,3 +167,113 @@ func contains(a []string, x string) bool {
...
+		}
+
+		// zero segments returned with nil err
+		if len(seg) == 0 {
+			break
+		}
+
+		err = checker.checkSegmentStatus(ctx, seg[0].GetSegmentDetail(), string(seg[0].GetPath()), &rmtSegInfo)
+		if err != nil {
+			zap.L().Error("repair failed", zap.Error(err))
```
mniewrzal: don't use global logger, use `checker.logger`

also method `checkSegmentStatus` not repairing so `repair failed` error message make no sense.

pkg/datarepair/checker/checker.go
```diff
@@ -63,13 +74,34 @@ func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overl
...
+			}
+			return nil
+		})
+	}()
+
+	go func() {
+		c <- checker.IrrepairLoop.Run(ctx, func(ctx context.Context) error {
+			err := checker.irrepairProcess(ctx)
+			if err != nil {
+				checker.logger.Error("process", zap.Error(err))
```
mniewrzal: just `process` message doesn't mean anything. Set meaningful message for error.

pkg/datarepair/checker/checker.go
```diff
@@ -200,3 +167,113 @@ func contains(a []string, x string) bool {
...
+		if err != nil {
+			return err
+		}
+
+		// zero segments returned with nil err
+		if len(seg) == 0 {
+			break
+		}
+
+		err = checker.checkSegmentStatus(ctx, seg[0].GetSegmentDetail(), string(seg[0].GetPath()), &rmtSegInfo)
```
mniewrzal: what should happen with a segment that it's now available and can be repaired? should it be removed from irreparabledb?

pkg/datarepair/checker/checker.go
aligeti: done

pkg/datarepair/checker/checker.go
aligeti: all the structure got changed to use durability stats by an earlier PR.

pkg/datarepair/checker/checker.go
aligeti: done

pkg/datarepair/checker/checker.go
aligeti: done

pkg/datarepair/checker/checker.go
aligeti: done

pkg/datarepair/checker/checker.go
aligeti: yes. It handled in checkSegmentStatus() when an irreparable segment becomes reparable, it is put into the repairable queue and removed from the irreparable queue. Line #208 and #218

mniewrzal:
we still missing tests for this new functionality.

aligeti:
> we still missing tests for this new functionality.

func TestIdentifyIrreparableSegments(t *testing.T) () in the checker_test.go take care of iterating along with the repair queue and verifies it. I thought this is already doing it. Anything in specific you want me to target?

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +170,120 @@ func contains(a []string, x string) bool {
...
+			Path:       path,
+			LostPieces: missingPieces,
+		})
+		if err != nil {
+			checker.logger.Error("error adding injured segment to queue: ", zap.Error(err))
+			return err
+		}
+
+		// if same segment previously was irreparable
+		if _, err = checker.irrdb.Get(ctx, []byte(path)); err == nil {
```
mniewrzal: When method `checkSegmentStatus` is triggered from `irreparableProcess()` then you are sure that this element was from `irrdb`. Can you change logic to avoid asking `irrdb` in such case to reduce DB traffic?

aligeti: irreparableProcess() runs off of irrdb and needs to read each entry and run the checksegmentstatus(). not sure how can I get pass the segment info. The irreparable process()'s reads the pointers data from irrdb and not the meta info db.

aligeti: done

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +170,114 @@ func contains(a []string, x string) bool {
...
+			LostPieces:         int32(len(missingPieces)),
+			LastRepairAttempt:  time.Now().Unix(),
+			RepairAttemptCount: int64(1),
+		}
+
+		// add the entry if new or update attempt count if already exists
+		err := checker.irrdb.IncrementRepairAttempts(ctx, segmentInfo)
+		if err != nil {
+			checker.logger.Error("error handling irreparable segment to queue: ", zap.Error(err))
+			return err
```
mniewrzal: create new error, don't duplicate log entries by logging and returning error

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +170,114 @@ func contains(a []string, x string) bool {
...
+	defer mon.Task()(&ctx)(&err)
+
+	limit := 1
+	var offset int64
+	var monStats durabilityStats
+
+	for {
+		seg, err := checker.irrdb.GetLimited(ctx, limit, offset)
+		if err != nil {
+			return err
```
mniewrzal: should this loop end with first occurred error?

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +170,114 @@ func contains(a []string, x string) bool {
...
+		err = checker.repairQueue.Insert(ctx, &pb.InjuredSegment{
+			Path:       path,
+			LostPieces: missingPieces,
+		})
+		if err != nil {
+			return err
+		}
+
+		// delete always returns nil when something was deleted and also when element didn't exists
+		_ = checker.irrdb.Delete(ctx, []byte(path))
```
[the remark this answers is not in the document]
aligeti: added the error handling

pkg/datarepair/checker/checker.go
```diff
@@ -124,78 +142,13 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
...
 				err = proto.Unmarshal(item.Value, pointer)
 				if err != nil {
-					return Error.New("error unmarshalling pointer %s", err)
-				}
-
-				remote := pointer.GetRemote()
-				if remote == nil {
-					continue
+					checker.logger.Error("error unmarshalling pointer: ", zap.Error(err))
+					return err
```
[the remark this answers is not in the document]
aligeti: updated the code to return error with Error.New()

pkg/datarepair/checker/checker.go
aligeti: updated the code to return error with Error.New()

pkg/datarepair/checker/checker.go
aligeti: we can return or make the logic complicated, by trying to analyze the error and add lot of error handling code, which we can avoid as this irreparable process runs after very short time and also the Identifyinjuredsegment() process will have taken care of it. I thought about it and found it would be clean and simple to return and come back and do it again when its cycle starts.

Also it the function failed, mostly there is good reason and if we try again , most likely we will see bunch of same errors. 

when there are no more rows, the function returns [], nil, which in the next line handles it as it knows there are no more rows to scan and exists the loop gracefully.

pkg/datarepair/checker/checker.go
```diff
@@ -72,13 +75,34 @@ func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overl
 func (checker *Checker) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	return checker.Loop.Run(ctx, func(ctx context.Context) error {
-		err := checker.IdentifyInjuredSegments(ctx)
+	c := make(chan error)
+
+	go func() {
```
wthorp: rather than the go func()'s, channels, and loggers, I think we should use an errgroup here, similar to https://github.com/storj/storj/blob/master/satellite/peer.go#L551

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +169,115 @@ func contains(a []string, x string) bool {
...
+			LostPieces: missingPieces,
+		})
+		if err != nil {
+			return Error.New("error adding injured segment to queue %s", err)
+		}
+
+		// delete always returns nil when something was deleted and also when element didn't exists
+		err = checker.irrdb.Delete(ctx, []byte(path))
+		if err != nil {
+			checker.logger.Error("error deleting piece: ", zap.Error(err))
```
mniewrzal: it should more like "error deleting entry from irreparable db"

pkg/datarepair/checker/checker.go
```diff
@@ -217,3 +169,115 @@ func contains(a []string, x string) bool {
 	}
 	return false
 }
+
+func (checker *Checker) checkSegmentStatus(ctx context.Context, pointer *pb.Pointer, path string, monStats *durabilityStats) (err error) {
```
wthorp: the naming of this function may have drifted over time.  it not only checks the segment status but also records the status in the database.  Consider renaming to something more generic, maybe 'updateSegmentStatus'?

pkg/datarepair/checker/checker_test.go
```diff
@@ -132,6 +132,29 @@ func TestIdentifyIrreparableSegments(t *testing.T) {
...
+				RootPieceId:  teststorj.PieceIDFromString("fake-piece-id"),
+				RemotePieces: pieces,
+			},
+		}
+		// put test pointer to db
+		metainfo = planet.Satellites[0].Metainfo.Service
+		err = metainfo.Put("fake-piece-id", pointer)
+		require.NoError(t, err)
+
+		// this deletes the previously entered irreparable pointer from irrdb
```
wthorp: When I first read this comment, I was like "well that certainly requires a comment!"  That's usually not a good sign.  Consider renaming IdentifyInjuredSegments.  Or maybe go for efficiency if you could call `checkSegmentStatus()` or `irrdb.Delete()` directly.

pkg/datarepair/checker/checker.go
aligeti: updated

pkg/datarepair/checker/checker.go
aligeti: I would like to take this up in another PR.

pkg/datarepair/checker/checker.go
aligeti: updated the function name as suggested.

pkg/datarepair/checker/checker_test.go
aligeti: I added the comment while I was doing development for debug thru the fmt.. It shouldnt be there ... removed the comment.

aligeti:
> Looks good, I'd approve but I really think the errgroup will improve the looks of the code.

yes, I agree with you.. I opened aPR (v3-1871) and probably will pick it up this sprint.

wthorp:
Kishore will address errgroup in another PR

PR #1957, opened 2019-05-13
adds metainfo rpc doc

navillasa:
What: metainfo rpc doc


docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,129 @@
...
+    pointerdb.Pointer pointer = 1;
+
+    // Commit Stream
+}
+```
+
+## Rationale
+
+Alternatives include individual messages (e.g. StartStream, CommitStream, StartSegment, CommitSegment), but this would result in more rpc calls overall.
+Also, trying to combine a StartStream message and StartSegment message would lead to unused fields sometimes, like stream_upload_id.
```
zeebo: Maybe this line would be more clear if it included

```protobuf
message StartStreamAndStartSegment {
	StartStream start_stream = 1;
	StartSegment start_segment = 2;
}
```

as an example of what is meant by combining.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,129 @@
...
+Our current metainfo rpc service isn't very future proof.
+If we shipped our current client code, it wouldn't be easy to update clients and add support for the following:
+- versions
+- multipart uploads
+- cleanup for uncommitted segments
+- general concurrency fixes
+
+Multipart upload can't happen with our current architecture because there's no concept of starting and ending a stream.
+
+Versions also wouldn't work, because if concurrent clients tried to upload to the same version, they would clobber each other.
```
zeebo: Maybe s/to the same version/at the same time/. The problem is that the current protobuf does not contain which version is being written, so it has to assume "latest version", which is problematic with concurrent clients (they both are writing to the "latest version").

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,129 @@
...
+This should also reduce the number of roundtrips needed to start and end both streams and segments.
+
+This design should ensure compatibility between client and server in the future as these features are implemented. Neither clients nor servers will be able to access more information than what we provide in this new proto, so we need to make sure the messages and service calls anticipate needs for the above features.
+
+## Design
+
+The following is a proto file for just the rpc portion to create segments. The read portions just need to have a stream id included, which should default to being the largest stream id.
+(When we implement versions, every upload will have a new stream, which will be given the largest stream id at that time. This will allow us to get older versions by checking the stream id.)
+
+With this design, streams will be created that contain multiple segments. Previously, there wasn't a finalization step between segments, which meant that a big global object was created at a certain path. This prevented multipart uploads because the start/end of each segment was not tracked.
```
zeebo: I think this should read "streams will be created that contain multiple **parts**." There were already RPCs to finalize individual segments, but each segment was not logically associated with a part, but instead just the specific object.

Also, I don't remember the details, but I think part id is important to the server somehow (IIRC there's some finalization step).

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,129 @@
...
+## Background
+
+Our current metainfo rpc service isn't very future proof.
+If we shipped our current client code, it wouldn't be easy to update clients and add support for the following:
+- versions
+- multipart uploads
+- cleanup for uncommitted segments
+- general concurrency fixes
+
+Multipart upload can't happen with our current architecture because there's no concept of starting and ending a stream.
```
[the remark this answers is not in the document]
navillasa: casually forgot we have multipart uploads now..

docs/design/metainfo-rpc.md
navillasa: I just added what you said here

docs/design/metainfo-rpc.md
navillasa: also added what you said here and took out the last sentence which I think was wrong

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+## Background
+
+Our current metainfo rpc service isn't very future proof.
+If we shipped our current client code, it wouldn't be easy to update clients and add support for the following:
+- versions
+- multipart uploads
+- cleanup for uncommitted segments
+- general concurrency fixes
+
+Native multipart upload can't happen with our current architecture because there's no concept of starting and ending a stream.
```
egonelbre: The issue is in out of order uploading of a single stream not starting/ending a stream.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+
+Versions also wouldn't work, because if concurrent clients tried to upload at the same time, they would clobber each other.
```
egonelbre: It's the other way around, concurrent uploading causes segments to get mangled and versions is a good way to handle concurrent uploads.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+The current protobuf does not contain which version is being written, so it has to assume "latest version", which is problematic with concurrent clients (they both are writing to the "latest version").
```
egonelbre: More accurate would be always "version 0"

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+With this design, streams will be created that contain multiple parts. There were already RPCs to finalize individual segments, but each segment was not logically associated with a part, but instead just the specific object.
+
+```protobuf
+message AddressedOrderLimit {
+    orders.OrderLimit2 limit = 1;
+    node.NodeAddress storage_node_address = 2;
+}
+
+service Metainfo {
+    rpc StartStreamAndStartSegment(StartStreamAndStartSegmentRequest) returns (StartStreamAndStartSegmentResponse);
```
egonelbre: Keep the individual RPC calls for segments.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+
+With this design, streams will be created that contain multiple parts. There were already RPCs to finalize individual segments, but each segment was not logically associated with a part, but instead just the specific object.
+
+```protobuf
+message AddressedOrderLimit {
+    orders.OrderLimit2 limit = 1;
+    node.NodeAddress storage_node_address = 2;
+}
+
+service Metainfo {
```
egonelbre: Both bucket and object manipulation missing.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+    rpc StartStreamAndStartSegment(StartStreamAndStartSegmentRequest) returns (StartStreamAndStartSegmentResponse);
+    rpc CommitSegmentAndStartSegment(CommitSegmentAndStartSegmentRequest) returns (CommitSegmentAndStartSegmentResponse);
+    rpc CommitSegmentAndCommitStream(CommitSegmentAndCommitStreamRequest) returns (CommitSegmentAndCommitStreamResponse);
+
+    // reads / deletes
+}
+
+message StartStreamAndStartSegmentRequest {
```
egonelbre: Use the specific messages inside the composite messages to make things clearer.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,147 @@
...
+message StartStream {
+    bytes stream_upload_id = 1;
+}
+message StartSegment {
+    bytes stream_upload_id = 1;
+    int64 segment_index = 2;
+    int64 part_number = 3;
+}
+```
+but this would result in more rpc calls overall.
```
egonelbre: The question isn't in the number of rpc calls per se, but rather avoiding the roundtrip. Also I suspect we might be able to do async requests without having to combine them into a single requests.

egonelbre:
See https://github.com/storj/storj/pull/1960 on what needs to be additionally included in this design doc.

docs/design/metainfo-rpc.md
ifraixedes: I'm not very clear with the added example, the test mention _stream_upload_id_ however the _message_ example doesn't contain it.

I don't bear in mind the current implementation, I've just read this document and trying to understand it with that much wider context.

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    rpc ListBuckets(BucketListRequest) returns (BucketListResponse);
+
+    rpc CreateObject(ObjectCreateRequest) returns (ObjectCreateResponse);
+    rpc CommitObject(ObjectCommitRequest) returns (ObjectCommitResponse);
+    rpc ListObjects(ObjectListRequest) returns (ObjectListResponse);
+
+    rpc DeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+    rpc FinishDeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+
+    rpc CreateSegment(SegmentCreateRequest) returns (SegmentCreateResponse);
```
jtolds: this should be BeginSegment everywhere (SegmentBeginRequest)

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+
+    google.protobuf.Timestamp created_at = 4;
+    
+    int64                default_segment_size = 5;
+    RedundancyScheme     default_redundancy_scheme = 6;
+    EncryptionParameters default_encryption_parameters = 7;
+}
+
+message BucketCreateRequest {
+    Bucket bucket = 1; // ignores created_at
```
jtolds: This should just pull appropriate settings from Bucket into the CreateRequest directly

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+}
+
+message BucketDeleteRequest {
+    bytes name = 1;
+}
+message BucketDeleteResponse {}
+
+message BucketListRequest {
+    bytes     cursor    = 1;
+    Direction direction = 2;
```
jtolds: We should eliminate Direction from the protocol entirely if nothing (miniogw) currently uses anything other than the Forward direction

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+
+    bytes  stream_id;
+
+    google.protobuf.Timestamp expires_at;
+
+    bytes  encrypted_metadata_nonce;
+    bytes  encrypted_metadata;
+
+    RedundancyScheme     redundancy_scheme;
+    EncryptionParameters encryption_parameters;
```
jtolds: expires_at through encryption_parameters (with the exception of redundancy scheme) should be moved to the ObjectCreateRequest. redundancy_scheme should be moved to SegmentBeginResponse

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+message SegmentCreateResponse {
+    repeated AddressedOrderLimit addressed_limits;
+}
+
+message AddressedOrderLimit {
+    orders.OrderLimit2 limit   = 1;
+    node.NodeAddress   address = 2;
+}
+
+message SegmentCommitRequest {
```
jtolds: This should also have redundancy scheme

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    bytes  encrypted_metadata;
+
+    RedundancyScheme     redundancy_scheme;
+    EncryptionParameters encryption_parameters;
+}
+
+message ObjectCommitRequest {
+    bytes  bucket = 1;
+    bytes  encrypted_path = 2;
+    int32  version = 3;
```
jtolds: this should include stream id

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+
+message ObjectListRequest {
+    bytes     bucket = 1;
+    bytes     encrypted_prefix = 2;
+    bytes     encrypted_cursor = 3;
+    Direction direction = 4;
+    int32     limit = 5;
+    bool      recursive = 6;
+
+    fixed32   meta_flags = 7;
```
jtolds: break meta_flags out into actual flags

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    bool      include_partial = 8;
+    bool      include_all_versions = 9;
+}
+
+message ObjectListResponse {
+    repeated Object items = 1;
+    bool more = 2;
+}
+
+message ObjectDeleteRequest {
```
jtolds: This should be ObjectBeginDeleteRequest

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    rpc CommitObject(ObjectCommitRequest) returns (ObjectCommitResponse);
+    rpc ListObjects(ObjectListRequest) returns (ObjectListResponse);
+
+    rpc DeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+    rpc FinishDeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+
+    rpc CreateSegment(SegmentCreateRequest) returns (SegmentCreateResponse);
+    rpc AbortSegment(SegmentAbortRequest) returns (SegmentAbortResponse);
+    rpc CommitSegment(SegmentCommitRequest) returns (SegmentCommitResponse);
+    rpc MakeInlineSegment(SegmentMakeInlineRequest) returns (SegmentMakeInlineResponse);
```
jtolds: missing DeleteSegment (for object deletions)

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    repeated AddressedOrderLimit addressed_limits;
+}
+
+message AddressedOrderLimit {
+    orders.OrderLimit2 limit   = 1;
+    node.NodeAddress   address = 2;
+}
+
+message SegmentCommitRequest {
+    bytes segment_id;
```
jtolds: should just use stream id, part number, and index separately

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    bytes encrypted_key_nonce;
+    bytes encrypted_key;
+
+    bytes encrypted_data_checksum;
+
+    repeated orders.PieceHash signed_piece_hashes; // TODO: add encrypted_segment_size to piece hash
+}
+
+message SegmentCommitResponse {
+    bytes segment_id; // <stream_id, part_number, index>
```
jtolds: same here

docs/design/metainfo-rpc.md
navillasa: in the meeting, I think you had this:
```
message BucketCreateRequest {
    bytes name = 1;
}
```
is that what this should be instead? I don't totally understand what "pull appropriate settings from Bucket into the CreateRequest directly" means.

docs/design/metainfo-rpc.md
navillasa: okay, Jeff searched minio, and looks like we're good to remove this.

docs/design/metainfo-rpc.md
navillasa: so like flags would be their own structs? what would be in the flags?

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,287 @@
...
+    rpc CreateBucket(BucketCreateRequest) returns (BucketCreateResponse);
+    rpc GetBucket(BucketGetRequest) returns (BucketGetResponse);
+    rpc DeleteBucket(BucketDeleteRequest) returns (BucketDeleteResponse);
+    rpc ListBuckets(BucketListRequest) returns (BucketListResponse);
+
+    rpc CreateObject(ObjectCreateRequest) returns (ObjectCreateResponse);
+    rpc CommitObject(ObjectCommitRequest) returns (ObjectCommitResponse);
+    rpc ListObjects(ObjectListRequest) returns (ObjectListResponse);
+
+    rpc DeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
```
navillasa: @zeebo  should this call be `BeginDeleteObject`? Also, should this call and `FinishDeleteObject` share `ObjectDeleteRequest` or should there be an `ObjectBeginDeleteRequest` and `ObjectFinishDeleteRequest`?

docs/design/metainfo-rpc.md
```diff
@@ -0,0 +1,317 @@
...
+    RedundancyScheme     redundancy_scheme;
+}
+
+message SegmentCommitResponse {
+    bytes stream_id;
+    int64 part_number;
+    int64 segment_index;
+}
+
+message SegmentMakeInlineRequest {}
```
navillasa: @zeebo wasn't sure what to add here and below 🤔

docs/design/metainfo-rpc.md
navillasa: I went ahead and added both `ObjectBeginDeleteRequest` and `ObjectFinishDeleteRequest`

docs/design/metainfo-rpc.md
egonelbre: Note: we have these already implemented, also listv2 does use it:
```
	var firstFull Key
	if !reverse && !opts.StartAfter.IsZero() {
		firstFull = joinKey(opts.Prefix, opts.StartAfter)
	}
	if reverse && !opts.EndBefore.IsZero() {
		firstFull = joinKey(opts.Prefix, opts.EndBefore)
	}
	err = store.Iterate(ctx, IterateOptions{
		Prefix:  opts.Prefix,
		First:   firstFull,
		Reverse: reverse,
		Recurse: opts.Recursive,
	}, iterate)
```

In code it's called end before.

PR #1959, opened 2019-05-14
add uplink-contexts design doc

zeebo:
## What / Why
Add a design doc for adding contexts to uplink (similar to kubectl) to support shares of sub paths of buckets.

[Rendered](https://github.com/storj/storj/blob/jw/uplink-contexts-design-doc/docs/design/uplink-scopes.md)


docs/design/uplink-contexts.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+Then if one asked for `sj://x/a/d`, the 2nd path would be the longest match and `"key2"` would be used to encrypt `d` and it would be appened to `"/e1"`. Similarly, if one asked for `sj://x/a/b/c/f`, then `"key2"` would be used to encrypt `b/c/f` and it would be appened to `"/e1"`. If one asked for `sj://x/f/g`, then `"key3"` would be used. `"key1"` would never be used as it is subsumed by the entry with `"key2"`.
+
+## Rationale
+
+This proposal is not a radical departure of the current method, just an extension to allowing a collection of paths to compose a bucket at any unencrypted spot. Thus, it shoudn't require any changes on the server side, reducing risk and allowing for earlier shipping.
+
+We could allow for "longest" matching. For example, in the above set of revealed paths one could use `"key1"` for `sj://x/a/b/c/d`. This begs a question about consistency. In principle be able to derive `"key1"` from the information provided by `"key2"`. If they disagree, an error could be thrown. This design takes the position that, these issues are trivial because it is very unlikely that such a share would be constructed, and even if it were, it would disagree.
```
thepaul: It probably ought to be mentioned in this doc that the "unencrypted path" part of a bucket share does not need to match the original path from the project root. Allowing for this shouldn't break anything and gives more flexibility to the users. Furthermore, I don't think anything in our system prohibits somebody from using some arbitrary unrelated encryption key for files under some path prefix, if they feel like it. Given that, it will not be possible in the general case to derive key1 from key2, and it would not be particularly useful to do so anyway.

thepaul: Also, I think "longest matching" share selection will not always give the right answer. Imagine I have macaroon A giving me read access to `a/b/`, and macaroon B giving me read/write access to `a/b/c/`, but macaroon B also has some other caveats. If B's other caveats are not satisfied, I should still be able to read the file `a/b/c/d` using macaroon A.

docs/design/uplink-contexts.md
```diff
@@ -0,0 +1,103 @@
...
+
+One could attempt to cause restricted macaroons to be "mounted". For example, if a bucket contains `sj://bucket/a/b/c` then the listing of the root would return "c". Unfortunately, this does not compose well with having multiple allowed paths in a macaroon: which "root" do you mean? It also causes confusion because the paths of objects inside of a bucket becomes a local rather than global property, even though the clients and servers must speak with the global view at all times.
+
+Additionally, there's a single global key per project. That's less than ideal for a situation like having distinct prod and staging buckets: you don't want them both using the same encryption keys so that giving access to staging does not leak the encryption keys for prod. It also doesn't allow one to re-key a bucket if a share is leaked. Even if the macaroon inside is revoked, it may be seen as a risk that the key was leaked because they are now depending on the satellite operator for the safety of their data.
```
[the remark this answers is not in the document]
zeebo: The problem I'm describing is that sj://prod and sj://staging have the same root encryption key, so all their path based keys are the same as well. So if you share sj://staging/foo/bar with someone, that also leaks the key for sj://prod/foo/bar.

Is my understanding wrong? Maybe the bucket name is included in the derivation of the root key? That still has revocation issues, though.

Like, right now, I only need one secret to get to all of the buckets inside of a project. That either gives up cross bucket secrecy, or bucket level revocation of keys.

docs/design/uplink-contexts.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Design
+
+Uplink will operate on a set of "contexts". Each context contains an API key, a satellite url, and a list of bucket shares. In Go,
```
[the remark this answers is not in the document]
zeebo: I totally agree and love "scope".

docs/design/uplink-contexts.md
```diff
@@ -0,0 +1,103 @@
...
+Satellite URL: <user puts in or picks satellite like now>
+```
+
+After a context has been created, bucket information must be added. For example,
+
+```
+$ uplink import [optionally specify a context with --context=default]
+
+Please respond with the type of information you have:
+1. Bucket with root passphrase
```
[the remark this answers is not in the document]
zeebo: That's fine. I agree it needs to be clear to the use how the provided entropy is being used and turned into key material. The UX here is only to elucidate the model.

docs/design/uplink-contexts.md
```diff
@@ -0,0 +1,103 @@
...
+
+```go
+paths := []BucketShare{
+    {Bucket: "x", EncryptedPath: "/e1/e2/e3", UnecryptedPath: "/a/b/c", EncryptionKey: "key1"},
+    {Bucket: "x", EncryptedPath: "/e1",       UnecryptedPath: "/a",     EncryptionKey: "key2"},
+    {Bucket: "x", EncryptedPath: "/e4",       UnecryptedPath: "/f",     EncryptionKey: "key3"},
+}
+```
+
+Then if one asked for `sj://x/a/d`, the 2nd path would be the longest match and `"key2"` would be used to encrypt `d` and it would be appened to `"/e1"`. Similarly, if one asked for `sj://x/a/b/c/f`, then `"key2"` would be used to encrypt `b/c/f` and it would be appened to `"/e1"`. If one asked for `sj://x/f/g`, then `"key3"` would be used. `"key1"` would never be used as it is subsumed by the entry with `"key2"`.
```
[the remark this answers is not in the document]
zeebo: Yes. This is how kubectl works right now. I usually had aliases like `alias kube-prod=kubectl --context=prod`, etc, and it was great.

docs/design/uplink-contexts.md
zeebo: Indeed. Thinking over that more, that makes me think that doing "longest" matching with ties broken by later entries would be the best, because it allows overrides.

zeebo: So that points out a restriction the data model. There's only one macaroon per context (now "scope"). If you want access to the same project under different macaroons, that's a different scope.

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
thepaul: Yeah, bucket name is included in the derivation.

> Like, right now, I only need one secret to get to all of the buckets inside of a project.

I don't know that we can avoid that scenario. It seems congruent to the AWS situation, where you only need the API key for a root account to be able to get at anything stored under that account. The best solution would seem to be "don't give people the root encryption key, then".

Revocation not being able to undo encryption key access is a thing that has bothered me, though. I don't know what solutions we can even offer. Recommend that people re-key their projects and reencrypt+reupload everything in them anytime somebody leaves the project? Add some kind of generational-key feature where everything added before time T uses key K1 but everything after T uses key K2?

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
thepaul: 👍

docs/design/uplink-contexts.md
thepaul: Ah. Right. I had the wrong model in my head.

But so this seems a little unwieldy, if people use it in the way I imagine they will. But nobody else seems to have the same expectations, so maybe it's fine.

But wouldn't it be nice if `uplink` could intelligently select which scope[s] _and_ which encryption key[s] to use, based on what you ask it to do? If we are already going to require an 'import' operation for each share, it seems like an obvious feature to provide. (Assuming, again, that we can add some kind of syntax for specifying a project in `sj://` URLs.)

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
zeebo: This proposes to at least have each bucket be independently keyed. Do you think that's worth doing?

docs/design/uplink-contexts.md
zeebo: Project & bucket isn't unique enough because project names aren't globally unique across satellites, so you'd have to specify quite a bit of information for it to automatically select it in some cases.

I think this proposal is a middle ground in that it does the right thing for selecting buckets, but requires you to be explicit about what project/access level you're interacting with.

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
zeebo: Some bullet points to help be clear about trade-offs:

1. Sharing the project root key shares into the future as well. Having each bucket keyed does not share anything except what was explicitly shared.
2. It's possible to revoke a bucket key by deleting it, giving it a new key, and uploading it again. To revoke a bucket with project level keys, the entire project would have to undergo a delete and reupload.
3. It's not possible to give a single key that works for all buckets and all time.

Point 1. doesn't require per bucket keys, it just requires not sharing the project key, but the fact that it exists means it's possible to leak. Point 3 could be considered a plus or a minus as far as I can see.

One way to think of it is that currently the bucket is just another path component as far as revocation is concerned, so we're choosing between having a single root key, or multiple root keys. I think multiple root keys is the safer option.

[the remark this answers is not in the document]
thepaul: > This proposes to at least have each bucket be independently keyed. Do you think that's worth doing?

I think I missed a step. What is proposing that?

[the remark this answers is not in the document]
zeebo: Oops, you're right in that it doesn't necessarily follow from the data model. I'm explicitly proposing we do that now, though.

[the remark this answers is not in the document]
thepaul: I think I'm going to reveal my idiocy by not understanding this comment, but oh well it's not a secret anymore

> Some bullet points to help be clear about trade-offs:

I can't seem to parse these into helpful points. Are these meant to be alternate choices, observations about the current state, observations about the code, or proposals?

> 1. Sharing the project root key shares into the future as well. Having each bucket keyed does not share anything except what was explicitly shared.

What does it mean to share into the future? What does it mean to have each bucket keyed? Why would keying a bucket cause something to be shared?

> It's possible to revoke a bucket key by deleting it, giving it a new key, and uploading it again. To revoke a bucket with project level keys, the entire project would have to undergo a delete and reupload.

Not sure we can really call that revoking, but sure, that would be a way to replace a key. I don't think we can treat it as a viable option in most cases.

> It's not possible to give a single key that works for all buckets and all time.

Why is that a thing anyone would expect to have? Are we still talking about encryption keys?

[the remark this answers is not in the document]
zeebo: It's also possible I'm an idiot so here goes:

In all of the following, I'm assuming a threat model where if someone has access to the encryption key, they have access to the data. In other words, the attacker is in control of the satellite somehow (either a 3rd party or the satellite operator themselves).

> What does it mean to share into the future? What does it mean to have each bucket keyed? Why would keying a bucket cause something to be shared?

By "shared into the future", I mean that if you have a project with some collection of buckets {a, b, c}, and you gain access to the project root key, you also gain access to the key for any bucket created in the future, not just {a, b, c}.

By "having each bucket keyed" I mean that each bucket has a distinct root key to derive all of the path keys.

By saying "having each bucket keyed does not share anything except what was explicitly shared", I mean that there is no value that you can leak that gives the keys to any buckets that have yet to be created.

> Not sure we can really call that revoking, but sure, that would be a way to replace a key. I don't think we can treat it as a viable option in most cases.

I agree, but it's strictly better than having to do the whole project.

> Why is that a thing anyone would expect to have? Are we still talking about encryption keys?

You can currently have it by having the root project key. That key derives all of the keys used for all of the buckets that exist in a project now and into the future. Yes, we're talking about encryption keys.

docs/design/uplink-contexts.md
thepaul: Right, I get that, but project name is highly likely to be unique for all projects accessible to a given user, and if they're not, the user always has the option of referring to them by their UUID (or a prefix thereof).

Assuming some sort of inspection command that shows them the names and UUIDs of projects they have access to, along with details of the specific shares, that seems relatively easy to understand from a user perspective and easy for us to code.

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
thepaul: > In all of the following, I'm assuming a threat model where if someone has access to the encryption key, they have access to the data. In other words, the attacker is in control of the satellite somehow (either a 3rd party or the satellite operator themselves). In other other words, the macaroon is useless from a security perspective.

Ok. That is definitely a valid model, but one that is harder to argue for. To compare with AWS again, I don't think anyone expects their data on S3 to remain 100% secure from an attacker that controls the S3 infrastructure. And I don't really expect our project management folks to be too worried about it.

However, there is a closely related threat model where attackers don't have access to the satellite, but do have access to data on many storage nodes (enough to have a non-negligible chance of getting some partial usable target data on their controlled nodes) and do have access to your encryption keys (because of revoked access or the like). In my opinion, we really do need a good answer for this threat model because it's easy to see how feasible such an attack could be. And most of your points here still hold in that model.

> By "having each bucket keyed" I mean that each bucket has a distinct root key to derive all of the path keys.

Possible terminology difficulty: I think they already have distinct root keys, if I am not confusing terms here, but I think you mean "independent" root keys, i.e., root keys not derived from a common source? That is already possible within the existing framework, actually; you can just declare that files in bucket B will now use a new arbitrary encryption key K if you want. To make that work with uplink, I guess you could just make a new scope for bucket B and key K.

> By saying "having each bucket keyed does not share anything except what was explicitly shared", I mean that there is no value that you can leak that gives the keys to any buckets that have yet to be created.

Ok. I don't know that that's too important. It seems to amount to "if you give away the top level secret-est root key and then you want to rescind that access so that the recipients can no longer decrypt any of the things you have stored, even if they get their hands on it, it's going to be a huge pain in the ass and you'll have to re-encrypt everything", which is a property that many other viable systems also have :)

> > Why is that a thing anyone would expect to have? Are we still talking about encryption keys?
>
> You can currently have it by having the root project key. That key derives all of the keys used for all of the buckets that exist in a project now and into the future. Yes, we're talking about encryption keys.

Ohhhh you you're talking about all buckets in that project derived from the same encryption key, not all buckets ever. Ok, makes sense. Definitely a valid concern for the given threat model.

What about this possible solution for this problem of "revocable encryption":

1. Add a field to `BucketShare`: `AlternateKeys map[int16][]byte`
2. Add a field to the pointer protobuf: `int16 alternate_key_id = N;`
3. Add some way to tell `uplink` that I want to use a new encryption key with `alternate_key_id=1` by default, for new objects in the relevant scope
4. Such new objects get `alternate_key_id=1` set in their pointer metadata
5. `uplink` uses the default `EncryptionKey` when decrypting old objects (objects with no alternate_key_id in the metadata) and the new key when decrypting new objects (because `AlternateKeys[1]` has the new key)

An attacker with revoked access could still decrypt objects stored with the old encryption key, but they could already have gotten the contents of those same objects _before_ their access was revoked anyway, so this already isn't a problem we can solve. And the user doesn't have to download, re-encrypt, or re-upload anything.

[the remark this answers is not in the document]
jtolds: hmm, that's an interesting idea! that does make it so the common case is one doesn't have to configure a passphrase when a new bucket shows up. i do think it would be good to at least have some affordance for that case (either a set of key_id to key mappings so we can persist bucket to key_id mappings or alternatively just a default key your uplink is configured with)?

in general i do think forcing all buckets to use a key derived from a project root encryption key forces some bad experiences (you must know the project root encryption key to make new buckets) but it doesn't sound like we're arguing for that so i'm happy.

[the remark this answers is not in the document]
zeebo: I don't think that proposal works. The key at the root of the bucket is used to derive a tree of paths. Changing that key would mean changing all of the encrypted paths, so now there's possibly objects that exist that are shadowed. You'd have to list N different encrypted locations if there's N possible alternate keys. That's quite a bit of complexity.

Also, that proposal is orthogonal to if each bucket has an independent root key. One could exist with or without the other.

This design doc does not depend on buckets being independently keyed, so I propose we drop it for now, and continue with a share including the bucket name so that we can support it in the future if we want.

zeebo:
I think the bucket keying UX is a distraction. I'd like to focus on the question of scopes existing. Specifically, do we

1. Have a `<satellite, macaroon>` per explicit user specified scope (this proposal).
2. Have a collection of `<satellite, macaroon, bucket share>` that the tool implicitly picks and allow the user to specify in the bucket path enough information to be unambiguous (https://github.com/storj/storj/pull/1959#discussion_r284013289).

docs/design/uplink-contexts.md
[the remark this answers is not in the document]
jtolds: :+1:

jtolds:
I definitely prefer explicit. That matches many of the tools I use. I like this proposal.

thepaul:
Ok, agreed, bucket keying UX (which should probably be called "encryption-key retirement" or something, because it doesn't necessarily have anything to do with the bucket layer) is a distraction. I'll just note for future reference that I think the alt-key-id thing can still work, and that @jtolds 's comments about it don't seem to be addressing quite the same idea I had in my head.

Remaining concerns I have here:

 * Can add a note to the doc about the plaintext path not necessarily matching the original path, to make that expectation explicit
 * I am starting to wonder if the BucketShare structure is quite right. @zeebo can you talk me through the use case where someone sends a share with a single macaroon but gives access to multiple buckets or paths?

zeebo:
I will add notes/conclusions of all the conversations

Some examples:
- A macaroon limited to list + get but you share both staging and prod buckets
- A macaroon that only allows puts but allows access to two paths in the same bucket to upload videos in one spot and thumbnails in another.

Additionally, macaroons themselves allow a whitelist of a collection of bucket+paths, so it seems like we should allow specifying the key material for a collection of bucket+paths as well.

thepaul:
If you have a macaroon limited to list + get but not limited by bucket, wouldn't that macaroon give you access to _all_ buckets in the project, and there wouldn't be any particular need to have an attached list of buckets?

And if the macaroon _is_ limited by path, wouldn't we want to get the allowed list of paths directly from the macaroon, rather than having an attached list which would presumably contain the same thing?

zeebo:
The macaroon should be limited to the buckets.

You need more than just a macaroon to access files: you also need the encryption keys. That's the whole point of the bucket shares. You can't put the encryption keys as part of the macaroon because then you leak the keys to the satellite.

thepaul:
> The macaroon should be limited to the buckets.

Okay, then as I said, wouldn't we want to get the allowed list of buckets/paths directly from the macaroon, rather than having an attached list which would presumably contain the same thing?

> You need more than just a macaroon to access files: you also need the encryption keys. That's the whole point of the bucket shares. You can't put the encryption keys as part of the macaroon because then you leak the keys to the satellite.

Heh, yes, I'm aware. But I'm not suggesting taking away the encryption keys. I'm wondering if we need the list of paths.

zeebo:
> > The macaroon should be limited to the buckets.
> 
> Okay, then as I said, wouldn't we want to get the allowed list of buckets/paths directly from the macaroon, rather than having an attached list which would presumably contain the same thing?
> 

Yeah, there is some sad duplication there.

> > You need more than just a macaroon to access files: you also need the encryption keys. That's the whole point of the bucket shares. You can't put the encryption keys as part of the macaroon because then you leak the keys to the satellite.
> 
> Heh, yes, I'm aware. But I'm not suggesting taking away the encryption keys. I'm wondering if we need the list of paths.

How do you know which key goes with which path? The macaroon can contain many paths. It's essentially using the bucket + encrypted path as an identifier to look up the unencrypted path + key in the list of bucket shares.

Maybe changing to to be 

```go
type Scope struct {
    APIKey ...
    Satellite string
    AllowedPaths map[BucketPath]BucketPathInfo
}

type BucketPath struct {
    Bucket string
    EncryptedPath string
}

type BucketPathInfo struct {
    UnencryptedPath string
    Key string
}
```

would be more clear?

thepaul:
Why wouldn't we just put the encryption keys in the order in which they appear in the macaroon?

thepaul:
Oops, I mean, in the order in which the respective paths/buckets appear in the macaroon.

zeebo:
You'd still have to duplicate the bucket and what about when a macaroon has more restrictions added to it? Like the first whitelist is for {a,b,c} and the second is just for {b,c}? I guess it's indexes for just the last whitelist group in the macaroon?

It seems more explicit to me to just have the duplication.

thepaul:
The bucket has to be specified in the macaroon already, doesn't it, if the macaroon has path restrictions? I think you can't have an encrypted path without there being a specific bucket (to module the path encryption key).

I do share a desire for explicitness, but I'm more worried about the inevitable cases when the macaroon list doesn't match the shares list, and we have a multiple-sources-of-truth problem.

thepaul:
Ok I've just realized that the same possibility-of-mismatches problem will still exist even if we take out the list of encrypted paths, because there will still be a list of unencrypted paths and the list of encryption keys. So, sorry for going down this dead-end road, I think what you have is the right thing.

zeebo:
> The bucket has to be specified in the macaroon already, doesn't it, if the macaroon has path restrictions? I think you can't have an encrypted path without there being a specific bucket (to module the path encryption key).
> 

I was incorrect. The unencrypted paths + keys would not have to duplicate the bucket.

> I do share a desire for explicitness, but I'm more worried about the inevitable cases when the macaroon list doesn't match the shares list, and we have a multiple-sources-of-truth problem.

I don't see how keeping it in an order that must match would help with those issues. In fact, it may make it worse, because you can't even detect that the encrypted paths aren't the same.

---

So I'm sympathetic to shares being shorter, and I think there's an equivalence in power, so I'm happy to do either.

I think in the interest of expediting this design doc and implementation, can we now limit it to opinions that are both (and not that this hasn't already been the case)

1. Actionable (For example: "please do this concrete thing")
2. Strongly held 

So, I'm going to go with what's written unless someone says that they strongly think that it should be some other specific way. If that's how you feel, as I said, I'm happy to do it.

zeebo:
I ended up adding/changing/realizing some things:

1. It adds an optional "project key" at the scope root for defaulting to when a bucket isn't in the list.
2. Points out some problems with key inconsistency. Namely that you can have invisible/undecryptable objects in your bucket and takes the position to hide+warn about them.
3. The parallel list vs duplication argument is actually one about serialization. The go structs themselves can contain the duplication with ~no overhead.
4. Implicit matching is also orthogonal, and I can imagine wanting both. For example, tooling may care to be very precise about the scope, and not want to force that into the url.
5. Removed some unnecessary/confusing talk about project level keys and the example command line session.

docs/design/uplink-scopes.md
```diff
@@ -8,20 +8,17 @@ This design document proposes the data we need to store inside of a share in ord
...
+There are complications when it comes to restricting to a whitelist of buckets and/or path prefixes inside of those buckets. For example, if a macaroon is limited to only allow access to paths beneath `sj://bucket/a/b`, then it would be impossible to list the root. Additionally, the clients and servers all operate on **encrypted** paths, making it impossible for the server to issue the list and filter, because one would have to reveal the root encryption key for the bucket for the client to be able to decrypt the names.
 
 ## Design
 
-Uplink will operate on a set of "scopes". Each scope contains an API key, a satellite url, and a list of bucket shares. In Go,
+Uplink will operate on a set of "scopes". Each scope contains an API key, a satellite url, an optional project key, and a list of bucket shares. In Go,
```
thepaul: I'm concluding that this project key is the project-level encryption key, not an API-type key, if it is used to derive bucket keys (which are again, presumably, encryption keys).

I'd assumed we would handle sharing at the project level by having a `BucketShare` with the empty string (or maybe `"/"`?) for the `Bucket` and `EncryptedPath` and `UnencryptedPath` fields. But your way seems fine too. I only wonder whether we shouldn't have `ProjectKey` be a `[]byte` instead of `string`, since `EncryptionKey` is `[]byte` in `BucketShare`.

docs/design/uplink-scopes.md
```diff
@@ -93,16 +46,22 @@ paths := []BucketShare{
...
+Then if one asked for `sj://x/a/d`, the 2st path would be the longest match and `"key2"` would be used to encrypt `d` and it would be appened to `"/e1"`. Similarly, if one asked for `sj://x/a/b/c/f`, then the 1st path would be the longest match and `"key1"` would be used to encrypt `f` and it would be appened to `"/e1/e2/e3"`. If one asked for `sj://x/f/g`, then `"key3"` would be used. Asking for `sj://x/g` would return not found.
+
+Note that the encrypted path does not need to match the unencrypted path, and this allows for having local views of bucket paths in the actual bucket.
 
 ## Rationale
 
-This proposal is not a radical departure of the current method, just an extension to allowing a collection of paths to compose a bucket at any unencrypted spot. Thus, it shoudn't require any changes on the server side, reducing risk and allowing for earlier shipping.
+This proposal is not a radical departure of the current method, just an extension to allowing a collection of paths to compose a bucket at any unencrypted spot. Thus, it shoudn't require any changes on the server side, reducing risk and allowing for earlier shipping. For example, any current configuration
```
thepaul: A partial sentence was appended here. A mistake?

zeebo: Good catch. I'll complete the thought.

docs/design/uplink-scopes.md
zeebo: Hmm... We could encode it by having the empty bucket be for the project-level encryption key, but then yeah the path fields are kinda meaningless. It's a special case either way I think, so I prefer the one that's more obvious to a reader, which IMO is an explicit field with documentation.

You are correct that it should be a `[]byte`.

docs/design/uplink-scopes.md
```diff
@@ -18,7 +18,7 @@ Uplink will operate on a set of "scopes". Each scope contains an API key, a sate
 type Scope struct {
     APIKey        string // Or perhaps uplink.APIKey
     SatelliteURL  string
-    ProjectKey    *string
+    ProjectKey    []byte
```
thepaul: "RootEncryptionKey", maybe? I have a tendency to get confused about unqualified "keys" in our codebase.

docs/design/uplink-scopes.md
```diff
@@ -52,7 +52,7 @@ Note that the encrypted path does not need to match the unencrypted path, and th
 
 ## Rationale
 
-This proposal is not a radical departure of the current method, just an extension to allowing a collection of paths to compose a bucket at any unencrypted spot. Thus, it shoudn't require any changes on the server side, reducing risk and allowing for earlier shipping. For example, any current configuration
+This proposal is not a radical departure of the current method, just an extension to allowing a collection of paths to compose a bucket at any unencrypted spot. Thus, it shoudn't require any changes on the server side, reducing risk and allowing for earlier shipping. For example, any current configuration can be represented as a scope with a default project key and no bucket shares.
```
thepaul: I would interpret "default project key" as using the `nil` value for `ProjectKey`, but that's probably not what you mean. What does default mean here?

docs/design/uplink-scopes.md
zeebo: "ProjectEncryptionKey"?

docs/design/uplink-scopes.md
zeebo: "can be represented as a scope with a `ProjectEncryptionKey` specified and no bucket shares."?

thepaul: Oh I get it, it was treating `ProjectKey` as a default in concept (because it gets used when no matching BucketShare is available). Ok. Yeah your suggested wording would be easier for me to understand.

thepaul:
👍

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+paths := []BucketShare{
+    {Bucket: "x", EncryptedPath: "/e1/e2/e3", UnecryptedPath: "/a/b/c", EncryptionKey: "key1"},
+    {Bucket: "x", EncryptedPath: "/e1",       UnecryptedPath: "/a",     EncryptionKey: "key2"},
+    {Bucket: "x", EncryptedPath: "/e4",       UnecryptedPath: "/f",     EncryptionKey: "key3"},
+}
+```
+
+Then if one asked for `sj://x/a/d`, the 2st path would be the longest match and `"key2"` would be used to encrypt `d` and it would be appened to `"/e1"`. Similarly, if one asked for `sj://x/a/b/c/f`, then the 1st path would be the longest match and `"key1"` would be used to encrypt `f` and it would be appened to `"/e1/e2/e3"`. If one asked for `sj://x/f/g`, then `"key3"` would be used. Asking for `sj://x/g` would return not found.
+
+Note that the encrypted path does not need to match the unencrypted path, and this allows for having local views of bucket paths in the actual bucket.
```
kaloyan-raev: Could you give some examples here? I am really confused about the relationship between `EncryptedPath` and `UnencryptedPath`. Until now I imagined that when sharing a file with someone we won't reveal its absolute path, not even the bucket it is in with the receiver.

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+
+Longest matching begs a question about consistency. In principle, it's possible to derive `"key1"` from the information provided by `"key2"`. If they disagree, an error could be thrown. Consistency does mean that you can't have differently keyed paths as part of your bucket, which may be useful. Inconsistency means that a bucket or project could contain objects that are invisible or undecryptable by clients. This design makes no requirement for consistency, and chooses to hide and warn about paths that cannot decrypt.
+
+The design does mean that if you create a new bucket on the web, you must also create the bucket in your clients, unless you have the project key. Since users are expected to manage their own encryption keys, and since creating a bucket from the command line can add the associated bucket share, this extra effort is justified.
```
kaloyan-raev: I don't really understand this paragraph. What is specific to creating a bucket on the web (I know we even don't have this at the moment)?

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+
+This design calls for explicitly choosing your scope, but that doesn't require that it do that forever. For example, one could automatically choose the scope if the bucket name is unambiguous, and provide more information in the bucket url to allow for unambiguously picking everything (satellite and api key).
```
kaloyan-raev: Could you give an example here too? I am not really sure what this is about.

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+```go
+type BucketShare struct {
+    Bucket          string
+    EncryptedPath   string
+    UnencryptedPath string
+    EncryptionKey   []byte
+}
+```
+
+If a client attempts to interact with a bucket, and no `BucketShare` matches, and there is a project key set, a key for that bucket is derived using the bucket name and the project key.
```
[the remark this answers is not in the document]
zeebo: An API key uniquely determines a project, and it's already the case that bucket names are unique per project.

docs/design/uplink-scopes.md
zeebo: I'll add an example about how it doesn't necessarily reveal the actual absolute path.

I don't know if sharing a single file is a supported use case we care about existing in this feature. For example, macaroons only support restricting to a prefix, not to an exact path. Additionally, even if we did care, I don't know if masking the bucket is part of that, and that we can't add it in a backwards compatible way later (we can always add/not set fields to a `BucketShare`).

docs/design/uplink-scopes.md
zeebo: You don't want to give your bucket's encryption key to the website when creating it, which means that you'd have to create the bucket in the web, then locally add the bucket with the encryption key.

I'll add a clarifying sentence.

kaloyan-raev: Oh, you mean "requesting a web server to create the bucket using server-side execution"? I believe we can cover the web use case without transferring the encryption key to the web server. We can have uplink executing in a client-side JavaScript.

docs/design/uplink-scopes.md
thepaul: Bucket names aren't encrypted in our paths, so I don't think it would be possible to hide the bucket name effectively when sharing.

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Design
+
+Uplink will operate on a set of "scopes". Each scope contains an API key, a satellite url, an optional project key, and a list of bucket shares. In Go,
+
+```go
+type Scope struct {
+    APIKey               string // Or perhaps uplink.APIKey
+    SatelliteURL         string
+    ProjectEncryptionKey []byte
```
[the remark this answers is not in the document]
thepaul: Do we use `storj.Key` for _encryption_ keys anywhere else in the code? `Key` is commented as being the type for "keys in a `KeyValueStore`", not for encryption keys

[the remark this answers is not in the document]
kaloyan-raev: `*storj.Key` should be the type for `ProjectEncryptionKey`, not for `APIKey`.

`APIKey` should have the type `*uplink.APIKey`.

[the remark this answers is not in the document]
thepaul: I get what you meant, but I'm asking if that is right. The docstring on `storj.Key` suggests that it is not.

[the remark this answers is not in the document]
kaloyan-raev: @thepaul This is the `storj.Key` I am referring to: https://github.com/storj/storj/blob/master/pkg/storj/encryption.go#L135

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+A bucket share contains an encrypted path, the matching unencrypted path, the key beginning at that path, and the bucket the path is contained in. In Go,
+
+```go
+type BucketShare struct {
+    Bucket          string
+    EncryptedPath   string
+    UnencryptedPath string
```
kaloyan-raev: Note that we have the `storj.Path` type (alias to `string`) that we use for paths within a bucket. It's good for both encrypted and unencrypted paths. Perhaps, use it here instead of `string`.

docs/design/uplink-scopes.md
```diff
@@ -0,0 +1,67 @@
...
+    EncryptionKey   []byte
```
kaloyan-raev: This one is also a good candidate for `*storj.Key`.

docs/design/uplink-scopes.md
kaloyan-raev: https://github.com/storj/storj/blob/master/pkg/storj/path.go#L11

docs/design/uplink-scopes.md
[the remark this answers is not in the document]
thepaul: Ok, you're right, my mistake. I was thinking of `storage.Key` as found in storage/common.go. Different thing.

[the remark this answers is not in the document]
zeebo: oops!

docs/design/uplink-scopes.md
zeebo: I'm just trying to point out that if you have a key per project, no client side information needs to be updated if a bucket is created on the server. If you allow buckets to have independent keys, then client information does need to be updated.

PR #1960, opened 2019-05-14
Rough sketch for Satellite RPC

egonelbre:
Rough sketch for all of the methods.

### DIFF
--- a/docs/design/metainfo-rpc.md
+++ b/docs/design/metainfo-rpc.md
@@ -221,7 +221,7 @@ message SegmentCommitRequest {
 }
 
 message SegmentCommitResponse {
-    bytes segment_id;
+    bytes segment_id; // <stream_id, part_number, index>
 }
 
 
--- a/docs/design/metainfo-rpc.md
+++ b/docs/design/metainfo-rpc.md
@@ -40,10 +40,11 @@ service Metainfo {
     rpc FinishDeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
 
     rpc CreateSegment(SegmentCreateRequest) returns (SegmentCreateResponse);
+    rpc AbortSegment(SegmentAbortRequest) returns (SegmentAbortResponse);
     rpc CommitSegment(SegmentCommitRequest) returns (SegmentCommitResponse);
-    rpc ContinueSegment(SegmentContinueRequest) returns (SegmentContinueResponse);
-
     rpc MakeInlineSegment(SegmentMakeInlineRequest) returns (SegmentMakeInlineResponse);
+
+    rpc Batch(BatchRequest) returns (BatchResponse);
 
     rpc ListSegments(SegmentListRequest) returns (SegmentListResponse);
     rpc DownloadSegment(SegmentDownloadRequest) returns (SegmentDownloadResponse);
@@ -231,6 +232,30 @@ message SegmentContinueRequest {
 message SegmentContinueResponse {
     SegmentCommitResponse commit = 1;
     SegmentCreateResponse next   = 2;
+}
+
+message BatchRequest {
+    message Request {
+        ObjectCreateRequest object_create;
+        ObjectCommitRequest object_commit;
+
+        SegmentCreateRequest     segment_create;
+        SegmentCommitRequest     segment_commit;
+        SegmentMakeInlineRequest segment_inline;
+    }
+    repeated Request requests;
+}
+
+message BatchResponse {
+    message Response {
+        ObjectCreateResponse object_create;
+        ObjectCommitResponse object_commit;
+
+        SegmentCreateResponse     segment_create;
+        SegmentCommitResponse     segment_commit;
+        SegmentMakeInlineResponse segment_inline;
+    }
+    repeated Response responses;
 }
 ```
 
--- a/docs/design/metainfo-rpc.md
+++ b/docs/design/metainfo-rpc.md
@@ -225,15 +225,6 @@ message SegmentCommitResponse {
 }
 
 
-message SegmentContinueRequest {
-    SegmentCommitRequest commit = 1;
-    SegmentCreateRequest next   = 2;
-}
-message SegmentContinueResponse {
-    SegmentCommitResponse commit = 1;
-    SegmentCreateResponse next   = 2;
-}
-
 message BatchRequest {
     message Request {
         ObjectCreateRequest object_create;
--- a/docs/design/metainfo-rpc.md
+++ b/docs/design/metainfo-rpc.md
@@ -210,9 +210,7 @@ message AddressedOrderLimit {
 }
 
 message SegmentCommitRequest {
-    bytes stream_id;
-    int32 part_number;
-    int32 index;
+    bytes segment_id;
 
     bytes encrypted_key_nonce;
     bytes encrypted_key;
@@ -222,7 +220,9 @@ message SegmentCommitRequest {
     repeated orders.PieceHash signed_piece_hashes; // TODO: add encrypted_segment_size to piece hash
 }
 
-message SegmentCommitResponse {}
+message SegmentCommitResponse {
+    bytes segment_id;
+}
 
 
 message SegmentContinueRequest {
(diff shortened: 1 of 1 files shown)


docs/design/metainfo-rpc.md
```diff
@@ -26,87 +26,211 @@ The following is a proto file for just the rpc portion to create segments. The r
...
+    rpc CreateObject(ObjectCreateRequest) returns (ObjectCreateResponse);
+    rpc CommitObject(ObjectCommitRequest) returns (ObjectCommitResponse);
+    rpc ListObjects(ObjectListRequest) returns (ObjectListResponse);
+
+    rpc DeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+    rpc FinishDeleteObject(ObjectDeleteRequest) returns (ObjectDeleteResponse);
+
+    rpc CreateSegment(SegmentCreateRequest) returns (SegmentCreateResponse);
+    rpc CommitSegment(SegmentCommitRequest) returns (SegmentCommitResponse);
+    rpc ContinueSegment(SegmentContinueRequest) returns (SegmentContinueResponse);
```
egonelbre: I think this is not needed and we can do an async create/commit in sequence instead.

zeebo: Since there aren't dependencies between segments, yes, there's no difference. It's probably easier and more bug free to just support the combined RPC, though, as you don't have to worry about partial failures on the client. For example, what if the start succeeds but the commit fails? Does one then need to cancel the start? What if the cancel fails? I'm sure answers can exist for those questions, but it'd be nice if they didn't need to be asked.

docs/design/metainfo-rpc.md
```diff
@@ -26,87 +26,211 @@ The following is a proto file for just the rpc portion to create segments. The r
...
-message AddressedOrderLimit {
-    orders.OrderLimit2 limit = 1;
-    node.NodeAddress storage_node_address = 2;
+service Metainfo {
+    rpc CreateBucket(BucketCreateRequest) returns (BucketCreateResponse);
+    rpc GetBucket(BucketGetRequest) returns (BucketGetResponse);
+    rpc DeleteBucket(BucketDeleteRequest) returns (BucketDeleteResponse);
+    rpc ListBuckets(BucketListRequest) returns (BucketListResponse);
+
+    rpc CreateObject(ObjectCreateRequest) returns (ObjectCreateResponse);
```
zeebo: Since I can't imagine wanting to create an object without also creating the first segment, and since creating the segment requires information from creating the object (the stream_id), there really should be a combined RPC for doing both to reduce a RTT.

docs/design/metainfo-rpc.md
```diff
@@ -26,87 +26,211 @@ The following is a proto file for just the rpc portion to create segments. The r
...
+
+    RedundancyScheme     redundancy_scheme;
+    EncryptionParameters encryption_parameters;
+}
 
-    // Start Segment
-    bytes segment_upload_id = 2;
-    repeated AddressedOrderLimit addressed_limits = 3;
-    bytes root_piece_id = 4 [(gogoproto.customtype) = "PieceID", (gogoproto.nullable) = false];
+message ObjectCommitRequest {
```
zeebo: I would prefer if this took the returned stream_id so that the server has the ability to use it like a browser cookie. It already has to depend on the create part to use the absolute version number reported from the response, and it seems like an easy bug to accidentally have it use `-1` for both that can be avoided entirely.

docs/design/metainfo-rpc.md
```diff
@@ -26,87 +26,211 @@ The following is a proto file for just the rpc portion to create segments. The r
...
-    pointerdb.Pointer pointer = 1;
+message AddressedOrderLimit {
+    orders.OrderLimit2 limit   = 1;
+    node.NodeAddress   address = 2;
+}
+
+message SegmentCommitRequest {
+    bytes stream_id;
+    int32 part_number;
+    int32 index;
```
zeebo: Similarly, I would prefer if this used a segment_upload_id or something to allow the server to use it like a browser cookie.

docs/design/metainfo-rpc.md
zeebo: Additionally, there should be a combined RPC for committing a segment and an object because you can't async send both: you don't know if the server will receive the object commit before the segment commit and have that segment not be included. That reduces a RTT for the commit as well.

navillasa: just wondering, why are multiple computers uploading different segments as part of one object? is that a normal thing? also if they're doing that, wouldn't the combined rpc still be good to ensure the create object happens before the create segment?

docs/design/metainfo-rpc.md
navillasa: wait why would the commit with version = -1 fail?

docs/design/metainfo-rpc.md
zeebo: It's important to have the common case be as fast as possible, and the common case is uploading an object from a single computer with possibly a single segment. Not having combined RPCs penalizes the common case to 2x the RTTs as it with combined RPCs.

docs/design/metainfo-rpc.md
zeebo: But that commits to the cookie's format. If it's bytes, we have room to insert information into that cookie that clients that we may not be able to update don't have to understand to return to us.

docs/design/metainfo-rpc.md
zeebo: Because perhaps a new version was started during your upload, and so -1 means something different than when you started uploading.

docs/design/metainfo-rpc.md
zeebo: The server needs to handle the partial failure mode, yes, but the client doesn't. From the client's perspective, a continue RPC could be atomic.

docs/design/metainfo-rpc.md
egonelbre: https://grpc.io/docs/tutorials/async/helloasync-cpp/ for RTT issues. I'm not quite sure how difficult it would be to support on Go. If that's really complicated I'm fine with adding the specific versions, but we still need the separate versions for concurrent upload.

docs/design/metainfo-rpc.md
egonelbre: If uplink crashes and wants to continue the upload it's pretty much the same state there will be multiple segments per stream that may have not finished uploading.

docs/design/metainfo-rpc.md
egonelbre: I realized one thing we could do is have a message for batch requests:

```
message BatchRequest {
    message Request {
        ObjectCreateRequest object_create;
        ObjectCommitRequest object_commit;
    
        SegmentCreateRequest segment_create;
        SegmentCommitRequest segment_commit;
        SegmentInlineRequest segment_inline;
    }
    repeated Request requests;
}
```

This would allow to implement:
```
object create
segment inline
object commit
```

In a single roundtrip. Also we can enforce the semantics to be such that the requests will be executed until first of them fails, which is similar to invoking them individually, but getting the RTT improvements.

docs/design/metainfo-rpc.md
egonelbre: Yeah, but I cannot see what that exact information would be. If there's more information here then a lot of the design would probably change as well... 

But, having better compatibility isn't worse and implementation wouldn't be much more difficult. So `segment_id` it is.

docs/design/metainfo-rpc.md
egonelbre: Added BatchRequest/BatchResponse

ifraixedes: > ...we still need the separate versions for concurrent upload

I'm trying to visualize a use case that requires having a version for an Object.
Would you mind to write a brief one here?

docs/design/metainfo-rpc.md
ifraixedes: How this kind of business logic errors are reported/responded in gRPC?

docs/design/metainfo-rpc.md
egonelbre: 3 concurrent uploads on the same encrypted path, one of them fails and one of them is canceled.

ifraixedes: so for avoiding to override the data on the same path without being aware?

zeebo: I think this solves all of my RTT concerns.

docs/design/metainfo-rpc.md
zeebo: I don't think this requires an additional index in the same way that the segment_upload_id wouldn't. The server can store the information as an opaque byte sequence that the client would send here, anyway.

PR #1967, opened 2019-05-15
Change where the encryption key is being stored for uplink

ifraixedes:

__NOTE__ this PR aims the same than #1895 but rebuilding from scratch the feature but  applying all the changes commented in the back and forth discussion that this feature has had and also have a simple way to resolve the conflicts with upstream.

I tried to remove the I/O added to `uplink` package on the previous PR sent for this task, but I ultimately realize that it isn't possible without having more refactoring that the one planned for this issue, because `uplink` package have some function that requires the encryption key, so if the key isn't passed by flag, which means that it has be read from the indicated filepath then it requires to read it from the file.

I think also that's better to have exported functions to read the key from disk rather than having as methods of `EncryptionConfig` type as it was on the previous PRs.

On the other hand, I've also realized that an encryption key is needed, because `lib/uplink` requires a non-empty key for some operations, so the integration tests couldn't not skip to use an encryption key and generating an encryption file brings the problem in how to delete such file when it isn't needed anymore (i.e. cleanup), hence I added a new flag to `uplink` which allows to set the encryption key to use taking priority over the one stored in a separated file and when not passed then it reads the one from the file.

Thanks for submitting a PR!


uplink/config.go
```diff
@@ -41,10 +44,11 @@ type RSConfig struct {
 // EncryptionConfig is a configuration struct that keeps details about
 // encrypting segments
 type EncryptionConfig struct {
-	Key       string      `help:"root key for encrypting the data"`
-	BlockSize memory.Size `help:"size (in bytes) of encrypted blocks" default:"1KiB"`
-	DataType  int         `help:"Type of encryption to use for content and metadata (1=AES-GCM, 2=SecretBox)" default:"1"`
-	PathType  int         `help:"Type of encryption to use for paths (0=Unencrypted, 1=AES-GCM, 2=SecretBox)" default:"1"`
+	EncryptionKey string      `help:"the root key for encrypting the data; when set, it overrides the key stored in the file indicated by the key-filepath flag"`
+	KeyFilepath   string      `help:"the path to the file which contains the root key for encrypting the data"`
```
egonelbre: I've been thinking how to handle this better. I think the major issue with the changes comes from conflating flags and configuration. Maybe, if we separate the Flags and Config struct. This means that things that do need to reuse the arguments.

So we would have something like:

```
func (flags *EncryptionFlags) Config() (*EncryptionConfig, error) {
```

This would get rid of the uplink depending on io issue. When we eventually get a proper place to store `EncryptionFlags` we can move it outside of `uplink` as well.

But, I'm also trying not to drag this PR on longer, because it has suffered long enough. So a separate PR for doing this change might make sense.

egonelbre: The annoying part is that it's part of the `uplink.Config` so we would need to split that as well.

egonelbre: Also this would allow later (in a separate PR) to clean up `EncryptionConfig` to be more specific:

```
	Key        storj.Key
	DataCipher storj.CipherSuite // or what ever the actual type was
	PathCipher storj.CipherSuite // or what ever the actual type was
	BlockSize  memory.Size
}
```

egonelbre: _PS: all of these can be done as a separate PRs to get things moving._

uplink/config.go
```diff
@@ -150,3 +156,48 @@ func (c Config) GetEncryptionScheme() storj.EncryptionScheme {
...
+			return nil, fmt.Errorf("not found key file %q", filepath)
+		}
+
+		return nil, err
+	}
+
+	defer func() { err = errs.Combine(err, file.Close()) }()
+
+	// TODO: if/v3-1541#3 should we check if the file contains a values of length
+	// storj.KeySize??
```
ifraixedes: Let me know if you think that we should add this check, if it yes, I will add it, otherwise I will just remove the this `TODO` comment.

uplink/config.go
ifraixedes: I'm ok with trying it if the time that I would need for it fits to the current sprint planning.
I'll comment about it in the today's standup.

ifraixedes:
Swapped to WIP because I didn't see that there are some tests failing because I only saw the tests which are failing due the tests which fail due to some data race present in master.
I'll check them tomorrow.

ifraixedes:
Ready to review again.

I had to remove the check for an empty key encryption filename. Read more about it in the same commit message https://github.com/storj/storj/pull/1967/commits/715d5f974372c02cc5f8411fec2141504924c890

cmd/uplink/cmd/setup.go
```diff
@@ -247,3 +277,35 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 	// address is host:
 	return net.JoinHostPort(addressParts[0], defaultPort), nil
 }
+
+// saveEncryptionKey saves the encryption key into a new file created in
+// filepath.
+func saveEncryptionKey(key *storj.Key, filepath string) error {
```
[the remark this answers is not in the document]
ifraixedes: I thought about it and I'm not agains nor to move it but I'd like to give you my rationale why I left it here.

The functions present in _common.go_ are used by different commands (i.e. cobra command functions), meanwhile this one is only used by the _setup_ one.

Let me know if you still see that it is clearer that it goes into _common.go_ and I will move it.

cmd/uplink/cmd/setup_unexported_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/storj"
+)
+
+func TestSaveEncryptionKey(t *testing.T) {
```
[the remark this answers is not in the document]
ifraixedes: The test is here because the function is implemented in _setup.go_ otherwise it would go there; my rationale why it's in _setup.go_ is as a response of your other commend.

Regarding to have an integration test between both, I'm that much sold on it, because the integration test should be more between this _save_ function and the _load_ one present in the `uplink` package, however it looks odd to me to have the integration test of 2 different package in one of them because I'd expect to have it in third one for such purpose, but

1. _save_ isn't exported
2. if it would possible, I'd be in a doubt in which one, but it doesn't matter because 1 forbid to have it in a third package.

Nonetheless, I didn't contemplated to have such test and I think that's worthwhile to have it despite of the odds mentioned above because we don't have that much test on the CLI, hence I'm going to add it.

uplink/config.go
```diff
@@ -150,3 +156,47 @@ func (c Config) GetEncryptionScheme() storj.EncryptionScheme {
...
+	return key, nil
+}
+
+// UseOrLoadEncryptionKey return an encryption key from humanReadableKey when
+// it isn't empty otherwise try to load the key from the file pointed by
+// filepath calling LoadEncryptionKey function.
+func UseOrLoadEncryptionKey(humanReadableKey string, filepath string) (*storj.Key, error) {
+	if humanReadableKey != "" {
+		var key storj.Key
+		copy(key[:], humanReadableKey)
```
kaloyan-raev: Shouldn't we use `storj.NewKey(humanReadableKey)` here?

uplink/config.go
```diff
@@ -150,3 +156,47 @@ func (c Config) GetEncryptionScheme() storj.EncryptionScheme {
...
+	}
+
+	defer func() { err = errs.Combine(err, file.Close()) }()
+
+	// TODO: if/v3-1541#3 should we check if the file contains a values of length
+	// storj.KeySize??
+	key = &storj.Key{}
+	if _, err := file.Read(key[:]); err != nil && err != io.EOF {
+		return nil, err
+	}
```
kaloyan-raev: Same here, shouldn't we use `storj.NewKey(<content-of-file>)`?

cmd/uplink/cmd/common.go
```diff
@@ -0,0 +1,39 @@
...
+	}, nil
+}
+
+// useOrLoadEncryptionAccess creates an encryption key from humanReadableKey
+// when it isn't empty otherwise try to load the key from the file pointed by
+// filepath and creates an EnryptionAccess with it.
+func useOrLoadEncryptionAccess(humanReadableKey string, filepath string) (libuplink.EncryptionAccess, error) {
+	if humanReadableKey != "" {
+		var key storj.Key
+		copy(key[:], humanReadableKey)
```
kaloyan-raev: `storj.NewKey` here too?

uplink/config.go
ifraixedes: Not for this one, this is not a passphrase, what's wanted here is to use an encryption key passed by parameter not generating one from the value passed.

uplink/config.go
ifraixedes: Because of the same than commented above, the file contains the key to use, not a passphrase

cmd/uplink/cmd/common.go
ifraixedes: Same reason than above

uplink/config.go
kaloyan-raev: I think we've discussed in one of the PRs you have closed that we don't want to use the human readable key directly as the root encryption key, but we want to derive the encryption key from it using PBKDF2 or something else. This is what `storj.NewKey` does. So we should use it here.

ifraixedes: The problem is that I haven't been able to find a way for generating file to store the encryption key and being able to clean it up on the integration tests.

I asked about how I could do it, but the response was kind of "there is not way so far" and that's one of the reasons for not having I/O on `uplink` package. Slack: https://storj.slack.com/archives/CAQV0AE2Z/p1557937208391100

This is also mentioned in the PR comment.

On the other hand, the goal of the issue wasn't about that, it was to have the encryption key outside of the `config.yaml` file, so it already contains more functionality on what the issue describe. I'm not saying that we should do more changes, but I think that new issues must be created for tackling them or at least, recreate the issue to indicate what it has to be done, because on each PR that I've sent new things are appearing and making impossible to estimate the issue on every sprint that it gets dragged.

kaloyan-raev: OK. So let's drop the PBKDF2 key derivation from this PR. Let `storj.NewKey` just do `copy(key[:], humanReadableKey)` like before, but let's have this logic just in `storj.NewKey` and nowhere else. This means that whenever we want to get `storj.Key` from the human-readable key, we should call `storj.NewKey`. The encryption file should store the human-readable key in plain text (no ending newline).

We will figure out in a separate PR how to properly do the key derivation from the human-readable key.

cmd/uplink/cmd/setup_unexported_test.go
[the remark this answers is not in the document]
mobyvb: I saw the integration test you added and it looks good to me

cmd/uplink/cmd/setup.go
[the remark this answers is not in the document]
mobyvb: Your logic makes sense. I think it is fine to do it this way.

cmd/uplink/cmd/common.go
kaloyan-raev: This still needs to be done.

cmd/uplink/cmd/common_test.go
```diff
@@ -0,0 +1,106 @@
...
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+		filename := ctx.File("encryption.key")
+
+		_, err := loadEncryptionAccess(filename)
+		require.Error(t, err)
+	})
+}
+
+func TestUsOrLoadEncryptionKeyIntoEncryptionAccess(t *testing.T) {
```
kaloyan-raev: typo: "TestUs..." should be "TestUse..."

cmd/uplink/cmd/mb.go
```diff
@@ -45,7 +46,7 @@ func makeBucket(cmd *cobra.Command, args []string) error {
 
 	project, err := cfg.GetProject(ctx)
 	if err != nil {
-		return fmt.Errorf("error setting up project: %+v", err)
+		return errs.Wrap(fmt.Errorf("error setting up project: %+v", err))
```
kaloyan-raev: Use `errs.New` instead of `errs.Wrap(fmt.Errorf(...))` if you want to create an error with stack trace.

cmd/uplink/cmd/common.go
ifraixedes: I missed yesterday to swap the `copy` by the `NewKey` after we changed.

cmd/uplink/cmd/mb.go
ifraixedes: I'm not sure from where this come from, but good spot!

cmd/uplink/cmd/setup.go
kaloyan-raev: Can we have this function save the human-readable encryption key instead of `storj.Key`?

pkg/pkcrypto/common.go
```diff
@@ -35,4 +35,6 @@ var (
 	ErrVerifySignature = errs.Class("signature verification error")
 	// ErrChainLength is used when the length of a cert chain isn't what was expected
 	ErrChainLength = errs.Class("cert chain length error")
+	// ErrSalt is used when something goes wrong while generating salt.
+	ErrSalt = errs.Class("unable to generate salt")
```
kaloyan-raev: I guess we don't need this anymore.

pkg/storj/encryption.go
```diff
@@ -3,6 +3,13 @@
 
 package storj
 
+import (
+	"github.com/zeebo/errs"
+)
+
+// ErrKey is used when something goes wrong a key.
+var ErrKey = errs.Class("key")
```
kaloyan-raev: This is not used in the latest change. So let's remove it. We can bring it back when necessary.

uplink/config.go
kaloyan-raev: This comment is still valid as per our last discussion.

pkg/pkcrypto/common.go
ifraixedes: Ive forgotten to remove it

pkg/storj/encryption.go
ifraixedes: Ive forgotten to remove it

cmd/uplink/cmd/setup.go
ifraixedes: Should do we remove the check if key is zero, which with this change will be if the human-readable key is empty?

kaloyan-raev: I think you can leave it.

uplink/config.go
ifraixedes: What the right size to use here to read the content file?
Should we read the file up to the end?

ifraixedes: For now, I've used 1024 bytes, however depending of your answers, I will have to change it.

cmd/uplink/cmd/setup.go
```diff
@@ -247,3 +272,39 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
 	// address is host:
 	return net.JoinHostPort(addressParts[0], defaultPort), nil
 }
+
+// saveEncryptionKey generates a Storj key from the inputKey and save it into a
+// new file created in filepath.
+func saveEncryptionKey(inputKey []byte, filepath string) error {
```
kaloyan-raev: This still doesn't do what I requested. This method should take the human-readable encryption key and just save it to the file as is, in plain text. The input shouldn't be changed in any way, like passing it to `storj.NewKey` or anything else.

uplink/config.go
kaloyan-raev: Just use https://golang.org/pkg/io/ioutil/#ReadFile

cmd/uplink/cmd/setup.go
ifraixedes: oh, I misunderstood your request

cmd/uplink/cmd/setup.go
```diff
@@ -247,3 +271,34 @@ func ApplyDefaultHostAndPortToAddr(address, defaultAddress string) (string, erro
...
+		}
+
+		return Error.Wrap(err)
+	}
+
+	defer func() {
+		err = Error.Wrap(errs.Combine(err, file.Close()))
+	}()
+
+	_, err = file.Write(inputKey)
```
kaloyan-raev: Could you simplify the implementation of this method by taking advantage of https://golang.org/pkg/io/ioutil/#WriteFile?

ifraixedes: I cannot if what we want is not to override an existing file, if it doesn't matter if the file exists and can be overridden then let me know and I swap over it.

kaloyan-raev: OK. Then leave it like this.

uplink/config.go
ifraixedes: Tacked in https://storjlabs.atlassian.net/browse/V3-1812

PR #1968, opened 2019-05-15
Optimize KnownUnreliableOrOffline SQL query

egonelbre:
KnownUnreliableOrOffline is a bottleneck in repair checker.

Use `ANY` check instead of `IN` for Postgres. This results in with the attached benchmark:
```
using id in ($1, ...)     1176302ns
using id = any($1)         763621ns with pq.ByteaArray(nodeIds.Bytes())
using id = any($1)         740142ns
```

_Measurements on a noisy machine._

This doesn't of course solve our problems because 0.7ms for that query is still quite terrible.

Thanks for submitting a PR!


satellite/satellitedb/overlaycache.go
```diff
@@ -159,36 +161,56 @@ func (cache *overlaycache) KnownUnreliableOrOffline(ctx context.Context, criteri
...
+			AND last_contact_success > ? AND last_contact_success > last_contact_failure
+		`), args...)
+
+	case *pq.Driver:
+		rows, err = cache.db.Query(`
+			SELECT id FROM nodes
+				WHERE id = any($1::bytea[])
+				AND audit_success_ratio >= $2 AND uptime_ratio >= $3
+				AND last_contact_success > $4 AND last_contact_success > last_contact_failure
+			`, nodeIDsArray(nodeIds),
```
egonelbre: Not quite sure whether it's worth to do this specific optimization because the improvement is in ~0.1ms range.

wthorp:
Cam was working on optimizing this query... not sure if this PR is a response to his work or not.

egonelbre:
@wthorp I didn't know that cam was working on it.

cam-a:
@wthorp @egonelbre I didn't end up making the changes to that query

satellite/satellitedb/utils.go
```diff
@@ -19,3 +23,45 @@ func bytesToUUID(data []byte) (uuid.UUID, error) {
 
 	return id, nil
 }
+
+type nodeIDsArray storj.NodeIDList
+
+// Value converts a NodeIDList to a postgres array
+func (nodes nodeIDsArray) Value() (driver.Value, error) {
```
wthorp: I love Valuers, but in this case I feel like we're enshrining database-specific serialization logic into generic Go code.

egonelbre: This is the reason it's not a method on `NodeIDList` type itself, but rather a separate type with the method. Maybe it might make sense to include `postgres` in the name?

wthorp: Ah yeah, I didn't catch that.  I like having postgres in the name

PR #1698, opened 2019-04-08
returns an error if tx.Get_Node_By_Id returns nil node

navillasa:
## why changes
- The satellite was crashing with a nil pointer dereference panic whenever `getNodeStats(nodeID, dbNode)` was called where dbNode was nil. There needed to be a check that dbNode is not nil before passing it to `getNodeStats`.
- `getNodeStats` did not use its first parameter `nodeID` so it was removed.

## what changes
- Adds a check for if dbNode is nil and returns an error if so. However, **this isn't a great fix** and we definitely need to figure out why it's returning nil, since the docs say:
```
QueryRow executes a query that is expected to return at most one row.
QueryRow always returns a non-nil value.
```
(https://golang.org/pkg/database/sql/#DB.QueryRow). I'm going to create a sticky note for this.
- Removes `nodeID` from `getNodeStats` parameters.

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?

PR #1972, opened 2019-05-15
adds db layer functions for containment

navillasa:
What: adds Get, IncrementPending, and Delete functions for the containmentDB

Why: we need to access containment db

Thanks for submitting a PR!


satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+		`INSERT INTO pending_audits (node_id, piece_id, stripe_index, share_size, expected_share_hash, reverify_count)
+		VALUES (?, ?, ?, ?, ?, ?)
+		ON CONFLICT(node_id)
+		DO UPDATE SET reverify_count = pending_audits.reverify_count + 1`,
+	)
+	_, err := containment.db.ExecContext(ctx, statement,
+		pendingAudit.NodeID.Bytes(), pendingAudit.PieceID.Bytes(), pendingAudit.StripeIndex, pendingAudit.ShareSize, pendingAudit.ExpectedShareHash, pendingAudit.ReverifyCount,
+	)
+	if err != nil {
+		return err
```
VinozzZ: Is there a reason why we are not wrapping this error?

navillasa: I think these will be wrapped when they're called by the methods in pkg/audit/containment.go

navillasa:
> This looks great! Are you creating a separate PR for the changes for Contained?

Thanks Bill! Yeah a separate PR would be needed. This is currently not linked to the existing verifier code, so nothing is calling these functions yet.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type containment struct {
+	db *dbx.DB
+}
+
+// Get gets the pending audit by node id
+func (containment *containment) Get(ctx context.Context, id pb.NodeID) (*audit.PendingAudit, error) {
+	if id.IsZero() {
+		return &audit.PendingAudit{}, audit.ContainError.New("node ID empty")
```
VinozzZ: Just a question. What's the benefit of returning a pointer here vs returning a `nil`?

navillasa: oh good catch! earlier I thought it was erroring because it wanted a pointer returned and not nil but it must've been erroring about something else lol. thanks!

pkg/audit/containment.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Delete will delete pending audit info from the db
+func (containment *Containment) Delete(ctx context.Context, nodeID pb.NodeID) error {
+	err := containment.db.Delete(ctx, nodeID)
+	if err != nil {
+		return ContainError.Wrap(err)
+	}
+	return nil
+}
```
Fadila82: Shouldn't the Delete function also set the contained flag for the node to false? Maybe I'm just missing where it's done?

pkg/audit/containment_test.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		err = planet.Satellites[0].DB.Containment().IncrementPending(ctx, info1)
+		require.NoError(t, err)
+
+		err = planet.Satellites[0].DB.Containment().Delete(ctx, info1.NodeID)
+		require.NoError(t, err)
+
+		_, err = planet.Satellites[0].DB.Containment().Get(ctx, info1.NodeID)
+		require.Error(t, err, audit.ErrContainedNotFound.New(info1.NodeID.String()))
```
ifraixedes: Shouldn't we verify that the returned error is of such specific class?

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Get gets the pending audit by node id
+func (containment *containment) Get(ctx context.Context, id pb.NodeID) (*audit.PendingAudit, error) {
+	if id.IsZero() {
+		return nil, audit.ContainError.New("node ID empty")
+	}
+
+	pending, err := containment.db.Get_PendingAudits_By_NodeId(ctx, dbx.PendingAudits_NodeId(id.Bytes()))
+	if err == sql.ErrNoRows {
```
ifraixedes: Is it a convention to not nest this checks in side of a `if err != nil` conditional?

satellite/satellitedb/containment.go
ifraixedes: yes, but not wrapping it here we would loose this line in the stack trace.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+// Delete deletes the pending audit
+func (containment *containment) Delete(ctx context.Context, id pb.NodeID) error {
+	if id.IsZero() {
+		return audit.ContainError.New("node ID empty")
+	}
+	isDeleted, err := containment.db.Delete_PendingAudits_By_NodeId(ctx, dbx.PendingAudits_NodeId(id.Bytes()))
+	if err == sql.ErrNoRows {
+		return audit.ErrContainedNotFound.New(id.String())
+	}
+	if !isDeleted {
```
ifraixedes: Shouldn't we have this conditional after the `if err != nil`?

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func convertDBPending(info *dbx.PendingAudits) (*audit.PendingAudit, error) {
+	if info == nil {
```
ifraixedes: When can this case happen?

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, Error.New("missing info")
+	}
+
+	nodeID, err := storj.NodeIDFromBytes(info.NodeId)
+	if err != nil {
+		return nil, err
```
ifraixedes: It's commented above as a reply of the @VinozzZ comment, but here it's were I see clear that we  can loose more lines in the stack trace than one

This apply for any non-wrapped error returned by this function

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+
+	pieceID, err := storj.PieceIDFromBytes(info.PieceId)
+	if err != nil {
+		return nil, err
+	}
+
+	pending := &audit.PendingAudit{
+		NodeID:            nodeID,
+		PieceID:           pieceID,
+		StripeIndex:       int(info.StripeIndex),
```
ifraixedes: After seen this, I'm wondering if it's a convention to use types without a specific length like `int` rather than the ones with specific length like `int64`

pkg/audit/containment.go
```diff
@@ -0,0 +1,71 @@
...
+
+// DB holds information about pending audits for contained nodes
+type DB interface {
+	Get(ctx context.Context, nodeID pb.NodeID) (*PendingAudit, error)
+	IncrementPending(ctx context.Context, pendingAudit *PendingAudit) error
+	Delete(ctx context.Context, nodeID pb.NodeID) error
+}
+
+// Containment allows access to pending audit info in the db
+type Containment struct {
```
ifraixedes: Where this abstraction is used or will be used?

satellite/satellitedb/containment.go
navillasa: okay so I should use errs.Wrap(err) here? or ContainError.Wrap here instead of at the layer up?

satellite/satellitedb/containment.go
navillasa: I'm not sure if it's a convention. If you think it's better, I can change the StripeIndex type to int32 to be more specific.

pkg/audit/containment.go
navillasa: it will be used by the audit verifier. should I include that in the comment?

pkg/audit/containment.go
navillasa: that's a good point! it does need to happen here I forgot to add it

pkg/audit/containment.go
```diff
@@ -0,0 +1,71 @@
...
+	log *zap.Logger
+	db  DB
+}
+
+// NewContainment creates a new containment db
+func NewContainment(log *zap.Logger, db DB) *Containment {
```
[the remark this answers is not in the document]
navillasa: I think so

pkg/audit/containment.go
```diff
@@ -0,0 +1,71 @@
...
+	return &Containment{
+		log: log,
+		db:  db,
+	}
+}
+
+// Get will get pending audit info from the db
```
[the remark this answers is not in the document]
navillasa: good point, i'll change it

pkg/audit/containment_test.go
navillasa: why would that be better than as it is currently?

satellite/satellitedb/containment.go
navillasa: I've seen a lot of other places in the codebase where this check happens first so I thought it might be better to do that.

satellite/satellitedb/containment.go
navillasa: okay i'll wrap it

satellite/satellitedb/containment.go
navillasa: okay good idea

satellite/satellitedb/containment.go
navillasa: if `containment.db.Get_PendingAudits_By_NodeId(ctx, dbx.PendingAudits_NodeId(id.Bytes()))` returns nil, nil, and these functions have returned that in the past so I thought it would be good to check.

pkg/audit/containment.go
navillasa: okay actually I think I'm going to add that in a separate PR, because it looks like there's not any functions on the overlay cache that will allow me to update the Contained field right now (UpdateNodeInfo only allows updating "info requested from the node itself like node type, email, wallet, capacity, and version").

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if err != nil {
+		return nil, audit.ContainError.Wrap(err)
+	}
+
+	return convertDBPending(pending)
+}
+
+// IncrementPending will create a new pending audit entry, or increase its reverify count if it already exists
+func (containment *containment) IncrementPending(ctx context.Context, pendingAudit *audit.PendingAudit) error {
```
mobyvb: I feel like `IncrementPending` should return an int or something that represents the current `reverify_count`. That way we don't have to follow it with a `Get` in order to determine if we've hit the max limit

navillasa: oh, that's a good idea. I would probably need to add a query to get the value here though right?

pkg/audit/containment.go
```diff
@@ -0,0 +1,83 @@
...
+		log: log,
+		db:  db,
+	}
+}
+
+// Get gets pending audit info from the db
+func (containment *Containment) Get(ctx context.Context, nodeID pb.NodeID) (*PendingAudit, error) {
+	pendingAudit, err := containment.db.Get(ctx, nodeID)
+	if err != nil {
+		return nil, ContainError.Wrap(err)
```
kaloyan-raev: nit: "Wrap returns nil if err is nil" (https://github.com/zeebo/errs/blob/master/errs.go#L30)

So I think it is safe to just do:

```go
pendingAudit, err := containment.db.Get(ctx, nodeID)
return pendingAudit, ContainError.Wrap(err)
```

Same thing for `IncrementPending` and `Delete`.

satellite/satellitedb/containment.go
kaloyan-raev: What happens if there is already a pending audit for the same node? Do we add a second pending audit, or do we return an error?

navillasa: We don't add a second pending audit, we just increment the existing pending audit's Reverify Count. We don't return an error either. Would it be better if we did?

navillasa: hey @kaloyan-raev I saw you thumbs-up'd Moby's comment, but just wanted to let you know that I decided not to write a new query inside the IncrementPending query, since it would do about the same as a Get anyway. do you think that's okay?

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Delete deletes the pending audit
+func (containment *containment) Delete(ctx context.Context, id pb.NodeID) error {
+	if id.IsZero() {
+		return audit.ContainError.New("node ID empty")
+	}
+	isDeleted, err := containment.db.Delete_PendingAudits_By_NodeId(ctx, dbx.PendingAudits_NodeId(id.Bytes()))
+	if err == sql.ErrNoRows {
+		return audit.ErrContainedNotFound.New(id.String())
```
kaloyan-raev: Do we really want to return an error in this case? I think this method should just delete everything about the given node from the containment db and not return an error in case there isn't anything about this node.

Also, I think if there is nothing to delete, the `Delete_PendingAudits_By_NodeId` will return `isDeleted == false` instead of this error, but I could be wrong.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if err != nil {
+		return audit.ContainError.Wrap(err)
+	}
+	if !isDeleted {
+		return audit.ErrContainDelete.New(id.String(), audit.ContainError.Wrap(err))
```
kaloyan-raev: At this moment `err` should be `nil`. Wrapping it will have no effect - the result will be still `nil`.

Also, see the comment above that I think we should not return an error in this case.

satellite/satellitedb/containment.go
navillasa: okay, I can check this with a test. Should I not log anything either?

satellite/satellitedb/containment.go
navillasa: should I just remove this case and be done?

satellite/satellitedb/containment.go
kaloyan-raev: I think it is OK.

satellite/satellitedb/containment.go
kaloyan-raev: Just check if in this cases that  `Delete_PendingAudits_By_NodeId` will indeed return `isDeleted == false` instead of `sql.ErrNoRows`. In such case you should just remove this if-clause.

satellite/satellitedb/containment.go
kaloyan-raev: Yes. In case you want to give info to the caller if there was something deleted or not, you can make the method return `isDeleted` boolean and just pass it over.

satellite/satellitedb/containment.go
navillasa: yep it will return `isDeleted == false`. so I'll leave the `if !isDeleted` check and remove the sql.ErrNoRows one.

navillasa: or should I remove both so that no error is returned if there's nothing to delete?

navillasa: oh, just saw your other comment thread- I'll do that. sorry for the triple reply.

satellite/satellitedb/containment.go
kaloyan-raev: I think we should return an error if there is already a pending audit for a different erasure share. This way we would catch a potential bug in our logic if we are trying the add another pending audit despite of already having one for this node.

satellite/satellitedb/containment.go
ifraixedes: Well, I always prefer to have explicit length for the variables rather than relying in something that change depending of the architecture, but that's a personal choice.

I never searched before about this topic, but I've just done it now and I found  https://yourbasic.org/golang/int-vs-int64/ that's my same personal opinion.

I cannot ask you to change that, somebody else more familiar with the code base and its evolution since the beginning could do it.

Sometimes, I drop this kind of questions in some reviews when the PR belongs to a person that is working in Storj much earlier than me to figure out things which I'm in a doubt and to learn new things from the answer.

pkg/audit/containment.go
ifraixedes: I needed to ask because I see some code that it wasn't used, so if there wasn't plan to use it, I'd have ask for removing it, but it isn't the case.

> should I include that in the comment?

Maybe, it's a good idea and then we can remove the comment once it's used by the audit verifier.

satellite/satellitedb/containment.go
navillasa: this makes sense, but I thought the point of having an `IncrementPending` upsert instead of `Create` and `Update` was so that `IncrementPending` could be used for both creating pending audits and then just incrementing their counters if the entry already existed. I thought we decided on this because if one node is not sending audit info for two different requested pieces, then we shouldn't record the second piece's info, we should just increment that node's existing reverify count as penalty.

kaloyan-raev: >  if one node is not sending audit info for two different requested pieces

How can this happen in the first place?

If a node fails in audit, we record the audited erasure share as a pending audit in the containment db. If the same node is audited for a different erasure share, we should not audit it for this new erasure share, but for the one we have as a pending audit. If it fails again, we call `IncrementPending`, but we pass the previous pending audit as a param, not a new one.

pkg/audit/containment.go
ifraixedes: for me it's clearer to have the conditional, because the reader doesn't have to bear in mind that `Wrap` returns `nil` if the passed error is `nil`.

It has never been a problem to me to understand both and once I read the _zeebo/errs_ docs, however I believe that using the common go pattern is clearer.

satellite/satellitedb/containment.go
navillasa: ah you're right, I think I had the chain of events confused. I'll make sure this returns error if there's another pending audit for a different erasure share. do you know if I'll need to add a second query to get that info or if I can add to the existing query?

pkg/audit/containment.go
ifraixedes: My previous question about the usage of this wrapper also came because it's currently a wrapper which doesn't provided anything on top of the _containment_ struct, however with this addition is justifying it aside of your response to my question.

pkg/audit/containment_test.go
ifraixedes: I didn't mean to remove this one, I was asking for having one `require` more which checks that the error is of `ErrContainedNotFound`.

Currently it doesn't do such check, your are writing the message to print out if the assertion fails, see docs https://godoc.org/github.com/stretchr/testify/require#Error

satellite/satellitedb/containment.go
ifraixedes: it looks a bit weird to me, however I don't personally have any problem in understanding the logic, so it was a question for allowing me to discover Storj code style.

satellite/satellitedb/containment.go
ifraixedes: if we want that the caller always verify so, I agree with @kaloyan-raev however when I read this logic and the method comment, I thought that we don't want that the caller had to worry of doing that check, just only call the function.

satellite/satellitedb/containment.go
ifraixedes: What @kaloyan-raev wrote looks good because if this could be call concurrently then, non-returning an error if it doesn't exist will make the usage easier.

satellite/satellitedb/containment.go
ifraixedes: Really, is it that possible?

It looks like a bug in `containment.db.Get_PendingAudits_By_NodeId(ctx, dbx.PendingAudits_NodeId(id.Bytes()))`, because my understanding is that `Get_PendingAudits_By_NodeId` always return an error if `info` is `nil`, because in case that it doesn't exists, it return the error `sql.ErrNoRows` rather than `nil`, nil`

satellite/satellitedb/containment.go
ifraixedes: One of those should be good for the stack trace matter.

Which one, it's less obvious to me and somebody could have a strong argument on which one based on how it has been used so far in other parts of the code base.

I'm torn which one should be, from one part I see that using `errs.Wrap` makes the function not to hold any responsibility about the who are the callers but at the same time it's an unexported function so the callers are from only from this package, so using `ContainError.Wrap` allow the callers to directly return from it, however it less flexible to future callers to use it if they have to return a different error class (e.g. `ErrContainedNotFound`).

I would say to seek advice from some one with more experience in the Storj code base for that answer.

kaloyan-raev: FWIW, there is no stack trace lost here, because `storj.NodeIDFromBytes` already return `zeebo/errs` error.

satellite/satellitedb/containment.go
ifraixedes: I've seen this as resolved but I wanted to post this comment just for being considered, not for being applied.

In _postgres_ there is the possibility to return an inserted/updated value https://www.postgresql.org/docs/9.5/sql-insert.html however it doesn't look that it can be done with sqlite, so we may not want to use it because it isn't going to work across SQL databases.

I'm not an expert on DB either, so I don't know if that `returning` instruction is bad from the performance point of view.

satellite/satellitedb/containment.go
ifraixedes: yes, that's right.
My bad, I didn't see that it was a Storj function so I missed to check if it was returning.

satellite/satellitedb/containment.go
navillasa: I'm not 100% that the same exact error will happen here, but here is more backstory on the bug that came up previously that I'm intending to prevent here: https://github.com/storj/storj/pull/1698

satellite/satellitedb/containment.go
wthorp: model pending_audits (
	key node_id
^^ NodeID is already unique

pkg/audit/containment_test.go
navillasa: okay thanks, I went ahead and added the extra check!

satellite/satellitedb/containment.go
navillasa: I think it looks a little weird too so I'll go ahead and move it!

satellite/satellitedb/containment.go
navillasa: I think you're right it's better to be specific, and we know these will be smaller numbers so I went ahead and changed them to type int32.

satellite/satellitedb/containment.go
navillasa: hm so should I add a check to see if the latest pending audit has the same node ID as another pending audit already in the db, but where the other fields are different? I think that would require calling Get before each IncrementPending. Would that defeat the point of the upsert or is that okay?

pkg/audit/containment.go
navillasa: opened a pr to add this functionality: https://github.com/storj/storj/pull/2013

satellite/satellitedb/containment.go
ifraixedes: it's an index, so I wouldn't expect negative index, so an unsigned type is better (e.g. `uint32`, `uint16`, etc.)

satellite/satellitedb/containment.go
kaloyan-raev: if `node_id` is already unique, my guess is that the database will return an error if you try to add another row with the same node id.

navillasa: @kaloyan-raev I added a test case for that here and it actually just increments the existing pending audit's reverify count but doesn't error

ifraixedes: It doesn't return an error because of the `ON CONFLICT` clause however it's better to have a test case for such scenario.

kaloyan-raev: I think this happens because we have `ON CONFLICT(node_id) DO UPDATE SET ...` in the SQL statement.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,116 @@
...
+	)
+	if err != nil {
+		return audit.ContainError.Wrap(err)
+	}
+
+	updateContained := dbx.Node_Update_Fields{
+		Contained: dbx.Node_Contained(true),
+	}
+
+	_, err = tx.Update_Node_By_Id(ctx, dbx.Node_Id(pendingAudit.NodeID.Bytes()), updateContained)
```
ifraixedes: `err` isn't checked and if it happens we have to rollback the transaction

satellite/satellitedb/containment.go
kaloyan-raev: I think this happens because we have `ON CONFLICT(node_id) DO UPDATE SET ...` in the SQL statement.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,116 @@
...
+		`INSERT INTO pending_audits (node_id, piece_id, stripe_index, share_size, expected_share_hash, reverify_count)
+		VALUES (?, ?, ?, ?, ?, ?)
+		ON CONFLICT(node_id)
+		DO UPDATE SET reverify_count = pending_audits.reverify_count + 1`,
+	)
+	_, err = tx.Tx.ExecContext(ctx, statement,
+		pendingAudit.NodeID.Bytes(), pendingAudit.PieceID.Bytes(), pendingAudit.StripeIndex, pendingAudit.ShareSize, pendingAudit.ExpectedShareHash, pendingAudit.ReverifyCount,
+	)
+	if err != nil {
+		return audit.ContainError.Wrap(err)
```
ifraixedes: We have to rollback the transaction here if there is an error.

The best way to have a single point for executing _commit_ or _rollback_ is to use a `defer` function call

satellite/satellitedb/containment.go
navillasa: okay, I'll go ahead and take that out.

satellite/satellitedb/containment.go
navillasa: oop that was an oversight but it's checked now

satellite/satellitedb/containment.go
navillasa: I was thinking of doing:
```
	return audit.ContainError.Wrap(errs.Combine(err, tx.Rollback()))
```
since I've seen it in a few other places. But how would you suggest using the defer?

pkg/audit/containment_test.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+func TestContainIncrementAndGet(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+
+		randBytes := make([]byte, 10)
+		_, err := rand.Read(randBytes)
+		require.NoError(t, err)
```
VinozzZ: Is this a convention to also check errors from other packages in our test?

navillasa: I wouldn't say it's a convention but I need this function to create random test data and it returns an error, so I need to handle that error in the test. You will see `require.NoError(t, err)` used for this in other test files too.

satellite/satellitedb/containment.go
ifraixedes: That's ok if it's fine to you, like me, to do the rollback on each error return since the `tx` creation.

I'm not a big fun of `defer` with logic which involve return named parameters, but it's worse to have duplicated logic because `tx.Rollback` can return an error, that's why I suggested, however using `errs` package it doesn't happen, so I prefer what you did.

pkg/audit/containment_test.go
ifraixedes: It isn't a matter of a convention is a matter of that a test can continue if the functions needed for the test return an error, so we have to abort the test if those happen and `require` do so.

satellite/satellitedb/containment.go
```diff
@@ -0,0 +1,137 @@
...
+		)
+		_, err = tx.Tx.ExecContext(ctx, statement,
+			pendingAudit.NodeID.Bytes(), pendingAudit.PieceID.Bytes(), pendingAudit.StripeIndex, pendingAudit.ShareSize, pendingAudit.ExpectedShareHash, pendingAudit.ReverifyCount,
+		)
+		if err != nil {
+			return audit.ContainError.Wrap(errs.Combine(err, tx.Rollback()))
+		}
+	} else if err == nil {
+		if !bytes.Equal(existingAudit.ExpectedShareHash, pendingAudit.ExpectedShareHash) {
+			return audit.ContainError.Wrap(errs.Combine(Error.New("pending audit already exists for nodeID %s", pendingAudit.NodeID), tx.Rollback()))
```
ifraixedes: Should the caller distinguish this error from the others for taking a different action?

satellite/satellitedb/containment.go
ifraixedes: Now that I've seen the changes, I would say what someone could prefer to use a `defer`; it would be because the logic isn't very simple (there are 4 points where the `Rollback` must be executed) so it's easy to forget to call `Rollback` in one of them or if the function get refactored in the future, so a `defer` would keep the `Rollback` and `Commit` execution in one single point just after the `tx` creation.

NOTE I'm not asking for changing in that way, it has that advantage but `defer` with named return parameters doesn't also produce a 100% clear logic, so as usual there are advantages and drawbacks.

I mostly wanted to give you the different rationales, I'm fine with both.

satellite/satellitedb/containment.go
ifraixedes: Revolved in a conversation in other means.

PR #1974, opened 2019-05-16
Bt/repair no nodes

wthorp:
What: 

Does not repair when RepairThreshold is reached if RepairThreshold is the same as SuccessThreshold
Added sanity check on repairing zero items, just in case

Why:
Otherwise we try to repair zero items, which does unnecessary work on create bogus log entries


pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +127,13 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 				}
 
 				remoteSegmentsChecked++
-				numHealthy := len(pieces) - len(missingPieces)
-				if (int32(numHealthy) >= pointer.Remote.Redundancy.MinReq) && (int32(numHealthy) <= pointer.Remote.Redundancy.RepairThreshold) {
+				numHealthy := int32(len(pieces) - len(missingPieces))
+				redundancy := pointer.Remote.Redundancy
+				if numHealthy >= redundancy.MinReq && numHealthy <= redundancy.RepairThreshold && redundancy.RepairThreshold != redundancy.SuccessThreshold {
```
JessicaGreben: any chance we can add a comment here to explain the condition where this `if` block is true, kind of a lot going on here?

pkg/datarepair/checker/checker.go
```diff
@@ -137,7 +142,7 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 					if err != nil {
 						return Error.New("error adding injured segment to queue %s", err)
 					}
-				} else if int32(numHealthy) < pointer.Remote.Redundancy.MinReq {
+				} else if numHealthy < pointer.Remote.Redundancy.MinReq {
```
JessicaGreben: "pointer.Remote.Redundancy" -> "redundancy"

satellite/metainfo/metainfo.go
```diff
@@ -457,7 +457,7 @@ func (endpoint *Endpoint) filterValidPieces(pointer *pb.Pointer) error {
 			remotePieces = append(remotePieces, piece)
 		}
 
-		if int32(len(remotePieces)) < remote.Redundancy.RepairThreshold {
+		if int32(len(remotePieces)) <= remote.Redundancy.RepairThreshold && remote.Redundancy.RepairThreshold != remote.Redundancy.SuccessThreshold {
 			return Error.New("Number of valid pieces is lower then repair threshold: %v < %v",
```
JessicaGreben: should we update error message as well with `<=` ?

PR #1975, opened 2019-05-16
Create offer table in satellitedb for free credit program

VinozzZ:
What: 
Create offer table for storing all offer information

Why:

Thanks for submitting a PR!


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -564,3 +564,48 @@ read one (
...
 delete reset_password_token ( where reset_password_token.secret = ? )
+
+
+//--- offer table ---//
+
+model offer (
+	key id
+	
+	field id blob
+	field name text ( updatable )
```
JessicaGreben: should all these fields be updatable? seems odd the name of the offer would change

VinozzZ: That's a good point. I will check in with the business to see if they actually have this need

thepaul: Note that it’s fine to leave off the “updatable” attribute until you need it. It doesn’t affect the SQL modeling at all, and we can change it later without requiring a migration or anything.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -564,3 +564,47 @@ read one (
...
+	field description text ( updatable )
+	field type int ( updatable )
+	
+	field credit int ( updatable ) 
+	field award_credit_duration_days int ( updatable )
+	field invitee_credit_duration_days int ( updatable )
+
+	field redeemable_cap int ( updatable )
+	field num_redeemed int ( updatable, autoinsert )
```
thepaul: I'm assuming here that redeemable_cap is meant to be a limit on how high num_redeemed is allowed to go. Is that accurate? It would be nice if we could add a CHECK constraint on the db table, if so. I don't think you can do that directly through dbx, but as far as I know we don't require that anything actually uses the DDL generated by dbx.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -564,3 +564,47 @@ read one (
...
+
+//--- offer table ---//
+
+model offer (
+	key id
+	
+	field id	serial 
+	field name text ( updatable )
+	field description text ( updatable )
+	field type int ( updatable )
```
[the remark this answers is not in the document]
VinozzZ: that's a great point!

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -564,3 +564,47 @@ read one (
...
+	
+	field credit int ( updatable )
```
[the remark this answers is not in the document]
VinozzZ: Yes, I was thinking to change it to something like `credit_in_cents` so the unit is clear

satellite/satellitedb/dbx/satellitedb.dbx
VinozzZ: I will add the check in the update query in the db implementation

PR #1976, opened 2019-05-16
satellite/datarepair: use reliability cache

egonelbre:
Checking every node adds a lot of load to the database.

Thanks for submitting a PR!


satellite/satellitedb/overlaycache.go
```diff
@@ -195,6 +195,32 @@ func (cache *overlaycache) KnownUnreliableOrOffline(ctx context.Context, criteri
 	return badNodes, nil
 }
 
+// Reliable returns all reliable nodes.
+func (cache *overlaycache) Reliable(ctx context.Context, criteria *overlay.NodeCriteria) (nodes storj.NodeIDList, err error) {
+	// get reliable and online nodes
+	rows, err := cache.db.Query(cache.db.Rebind(`
+		SELECT id FROM nodes
```
wthorp: This concept of reliable came into existence essentially through code drift.  These ratios will go away and this check might become "NOT disqualified".  On the plus side, it means that this list will never have false positives.  On the negative side, this list might grow so large as to not be tenable in-memory.

egonelbre: Yeah, I'm aware of the growth, but I think for now, this is better than hitting the DB hard.

stefanbenten: I totally second Egon's feeling

satellite/satellitedb/overlaycache.go
```diff
@@ -509,6 +509,31 @@ func (cache *overlaycache) KnownUnreliableOrOffline(ctx context.Context, criteri
...
+		WHERE last_contact_success >= ? AND last_contact_success >= last_contact_failure`),
```
stefanbenten: Dont we need also the disqualified flag here `disqualified IS NULL`?

pkg/datarepair/checker/online.go
```diff
@@ -0,0 +1,72 @@
...
+	"storj.io/storj/pkg/storj"
+)
+
+// ReliableCache caches the reliable nodes for the specified staleness duration
+// and updates automatically from overlay.
+type ReliableCache struct {
+	overlay    *overlay.Cache
+	staleness  time.Duration
+	lastUpdate time.Time
+	reliable   map[storj.NodeID]struct{}
```
stefanbenten: I would prefer to have a map[storj.NodeID]bool.
Reason for this is simply that if a node got an upload with the 5 minute timeframe and then that file is checked, we would account that node as down (not reliable) right?

satellite/satellitedb/overlaycache.go
stefanbenten: Actually with the comment below, this would need to be restructured to include a flag if reliable or not.

egonelbre: Yeah, for some reason I was thinking we disabled that check... but I think we just reset the disqualification statuses.

pkg/datarepair/checker/online.go
egonelbre: The check for missing pieces also handles the piece creation date:

```
missingPieces, err := checker.nodestate.MissingPieces(ctx, createdAt, pieces)

[snip]

func (cache *ReliableCache) MissingPieces(ctx context.Context, created time.Time, pieces []*pb.RemotePiece) ([]int32, error) {
	if created.After(cache.lastUpdate) || time.Since(cache.lastUpdate) > cache.staleness {
		err := cache.Refresh(ctx)
		if err != nil {
			return nil, err
		}
	}
```

So if we encounter a piece that is newer than our last update then we'll refresh the cache.

egonelbre: Using a bool could work also I guess, but would require updating one node at a time.

stefanbenten: Oh, true. Good call.

satellite/satellitedb/overlaycache.go
egonelbre: seems like this also might have fixed the intermittent failures

pkg/datarepair/checker/online.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+
+	"storj.io/storj/pkg/overlay"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+// ReliableCache caches the reliable nodes for the specified staleness duration
+// and updates automatically from overlay.
+type ReliableCache struct {
```
thepaul: Can this be "ReliabilityCache" instead? "ReliableCache" implies to me "cache that is reliable", and that caused me some confusion in reviewing here

pkg/datarepair/checker/online.go
```diff
@@ -0,0 +1,72 @@
...
+			unreliable = append(unreliable, piece.PieceNum)
+		}
+	}
+	return unreliable, nil
+}
+
+// Refresh refreshes the cache.
+func (cache *ReliableCache) Refresh(ctx context.Context) error {
+	for id := range cache.reliable {
+		delete(cache.reliable, id)
```
thepaul: This surprised me. Is this an idiom for Go mutexes? I admit I don't know. Wouldn't it be a lot faster just to reinstantiate the map entirely?

Or, if not, do you not need some sort of mutex around this? Is it ok for some other goroutine to query missing pieces while half of the items are deleted from the cache? Or maybe only one goroutine will ever use it? Either way, it is probably worth leaving a comment to avoid confusing the future reader.

egonelbre: This deletion loop gets optimized away to doing the most efficient clearing.

The implementation assumes that it's used from one goroutine only. I guess we could add a mutex around it without problems.

satellite/satellitedb/overlaycache.go
```diff
@@ -509,6 +509,32 @@ func (cache *overlaycache) KnownUnreliableOrOffline(ctx context.Context, criteri
 	return badNodes, nil
 }
 
+// Reliable returns all reliable nodes.
+func (cache *overlaycache) Reliable(ctx context.Context, criteria *overlay.NodeCriteria) (nodes storj.NodeIDList, err error) {
+	// get reliable and online nodes
+	rows, err := cache.db.Query(cache.db.Rebind(`
+		SELECT id FROM nodes
+		WHERE disqualified IS NULL
+		  AND last_contact_success >= ? AND last_contact_success >= last_contact_failure`),
```
thepaul: Several other places in the code make similar queries, but use `>` for both of these comparisons instead of `>=`. Are you making it `>=` on purpose to solve something? If so, shouldn't we make this change in the other functions as well?

egonelbre: This is using the negation of the other queries.

pkg/datarepair/checker/online.go
thepaul: I didn't know that about the optimization! That is cool.

I would be happy with just a comment somewhere saying that `ReliabilityCache` can only be used from a single goroutine.

pkg/datarepair/checker/online.go
egonelbre: renamed

pkg/datarepair/checker/online.go
egonelbre: added comment that `ReliabilityCache` is not safe for concurrent use

satellite/satellitedb/overlaycache.go
thepaul: Are you sure? The senses in which it is used seem to be the same. See e.g. the query from `KnownUnreliableOrOffline()`:

```
SELECT id FROM nodes
    WHERE id = any($1::bytea[])
    AND disqualified IS NULL
    AND last_contact_success > $2 AND last_contact_success > last_contact_failure
```

PR #1977, opened 2019-05-16
jenkins: check that lock file is updated properly

egonelbre:
`protolock status` doesn't check whether new things have been added to proto.lock

Thanks for submitting a PR!


Makefile
```diff
@@ -90,10 +90,10 @@ test-sim: ## Test source with storj-sim (jenkins)
 	@echo "Running ${@}"
 	@./scripts/test-sim.sh
 
-.PHONY: test-satellite-cfg-change
-test-satellite-cfg-change: ## Test if the satellite config file has changed (jenkins)
+.PHONY: test-satellite-config-change
```
ifraixedes: I prefer _config_ as abbreviation but the _cfg_ is commonly used in other sources.

If at the end we rename this one, then we should also rename the `update-satellite-cfg-lock` target.

egonelbre: The file is called `config.yaml` so `config` seems more appropriate. About `update-satellite-cfg-lock` sure.

PR #1978, opened 2019-05-16
Reverting uplink(V3-1717) and gateway(V3-1718) enc key changes

aligeti:
What: 
Not want to encrypt paths so we get ordered listing
Why:
With encrypted paths you can’t get alphabetical listing, which means we cant not require enc.key

Will be reverting changes associated with Tickets V3-1717(PR #1952) & V3-1718(PR #1923)
Thanks for submitting a PR!

### DIFF
--- a/cmd/uplink/cmd/setup.go
+++ b/cmd/uplink/cmd/setup.go
@@ -133,14 +133,9 @@ Please enter numeric choice or enter satellite address manually [1]: `)
 		if err != nil {
 			return err
 		}
-
 		_, err = fmt.Println()
 		if err != nil {
 			return err
-		}
-
-		if len(encKey) == 0 {
-			return errs.New("Encryption passphrase cannot be empty")
 		}
 
 		_, err = fmt.Print("Enter your encryption passphrase again: ")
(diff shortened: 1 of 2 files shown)


jtolds:
Wait wait, I'm miscommunicating to a bunch of folks.

jtolds:
Wait wait, I'm miscommunicating to a bunch of folks.

@ifraixedes @aligeti @iglesiasbrandon 

We want to support unencrypted paths in the data model and in the config file. We want to make sure people can get ordered listing if they really want.

That's separate than what the wizard does. IF there's a passphrase, we should require it. If someone is intending to encrypt their data (which is what the wizard does), we should make them choose a non-empty passphrase. If we want to extend the wizard to let people choose unencrypted paths, then that's different, but we should not allow empty passphrases.

We don't need to revert these changes. Okay to close this PR without merging?

aligeti:
> Wait wait, I'm miscommunicating to a bunch of folks.
> 
> @ifraixedes @aligeti @iglesiasbrandon
> 
> We want to support unencrypted paths in the data model and in the config file. We want to make sure people can get ordered listing if they really want.
> 
> That's separate than what the wizard does. IF there's a passphrase, we should require it. If someone is intending to encrypt their data (which is what the wizard does), we should make them choose a non-empty passphrase. If we want to extend the wizard to let people choose unencrypted paths, then that's different, but we should not allow empty passphrases. Empty passphrases are not the same thing as choosing to not encrypt your paths.
> 
> We don't need to revert these changes. Okay to close this PR without merging?

got it ...closing the PR.

ifraixedes:
> If we want to extend the wizard to let people choose unencrypted paths, then that's different

It means that at some point we are going to allow the users to configure if they want to encrypt the path or not, but the data will be always encrypted?

PR #1979, opened 2019-05-16
don't encrypt bucket info

jenlij:
Background: see https://github.com/storj/storj/pull/1761

Summary: we should not require an encryption key to manage buckets
(though we require an encryption key to manage objects within a
bucket), so that bucket management can be done in the web interface.
This means the encryption key used when opening a project is going away.

We have a network wipe planned for next release, so for that wipe let's
make sure data and buckets uploaded after that wipe do not require an
encryption key.


jenlij:
Reopening https://github.com/storj/storj/pull/1854 here

PR #1983, opened 2019-05-16
remove downloader from verifier

wthorp:
What: 
Remove downloader struct from verifier

Why:
General cleanup.  Its clearer / less code without the extra struct in the mix.


pkg/audit/verifier.go
```diff
@@ -38,50 +38,33 @@ type Share struct {
...
-
 // NewVerifier creates a Verifier
 func NewVerifier(log *zap.Logger, transport transport.Client, overlay *overlay.Cache, orders *orders.Service, id *identity.FullIdentity, minBytesPerSecond memory.Size) *Verifier {
-	return &Verifier{downloader: newDefaultDownloader(log, transport, overlay, id, minBytesPerSecond), orders: orders, auditor: id.PeerIdentity()}
+	return &Verifier{log: log, orders: orders, auditor: id.PeerIdentity(), transport: transport, overlay: overlay, minBytesPerSecond: minBytesPerSecond}
 }
 
 // Verify downloads shares then verifies the data correctness at the given stripe
-func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe) (verifiedNodes *RecordAuditsInfo, err error) {
+func (v *Verifier) Verify(ctx context.Context, stripe *Stripe) (verifiedNodes *RecordAuditsInfo, err error) {
```
ifraixedes: We have to delete such convention https://github.com/storj/docs/blob/master/code/Style.md#variable-naming

Otherwise it is misguiding

wthorp: oh, thanks! I heard you mention this in relation to https://github.com/golang/go/wiki/CodeReviewComments#receiver-names.  I misunderstood our preference.

ifraixedes: you're not he only one using short names, so I was wondering if the people didn't like to use shorter ones or basically they aren't aware of it.

I'll comment to the next reviews if I see them, because it looks that it's more about people aren't aware.

PR #1985, opened 2019-05-17
satellite/mailservice: wait for goroutines to finish

egonelbre:
What: Unattended go routines may end up logging after all the contexts and services have been shut down.

Why: There are races being reported from console.

Thanks for submitting a PR!

### DIFF
--- a/satellite/mailservice/service.go
+++ b/satellite/mailservice/service.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	htmltemplate "html/template"
 	"path/filepath"
+	"sync"
 
 	"go.uber.org/zap"
 	monkit "gopkg.in/spacemonkeygo/monkit.v2"
@@ -53,6 +54,8 @@ type Service struct {
 	html *htmltemplate.Template
 	// TODO(yar): prepare plain text version
 	//text *texttemplate.Template
+
+	sending sync.WaitGroup
 }
 
 // New creates new service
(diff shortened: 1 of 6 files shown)


satellite/mailservice/service.go
```diff
@@ -74,12 +77,28 @@ func New(log *zap.Logger, sender Sender, templatePath string) (*Service, error)
 	return service, nil
 }
 
+// Close closes and waits for any pending actions.
+func (service *Service) Close() error {
```
ifraixedes: `Close` isn't a good name for this method because it doesn't close anything, it only waits.

I usually expect that a `Close` method finalizes something and the instance gets partially/totally unutilizable.

egonelbre: Yeah, `Close` is currently used as a thing to say "please ensure that this service has completely finished". Having different names like `WaitForFinish`, `Finish` and `Stop` ended up causing more confusion... the semantics for `Close` are more clearer - if something has `Close` you are expected to call it when you are done with it.

But, yeah the comment could be clearer on what it does.

ifraixedes: I don't see it in that way, I see that's a bad name for what it does, but anyway, it's my personal perception.

egonelbre: I agree about the name not being descriptive enough. I.e. the consistency for using "Close" to finalize made working with peer code easier. Initially there were three or four different methods to show the difference, so I was never sure what was the correct method to call to tear down a service.

PR #1989, opened 2019-05-17
implement database connection debugging

egonelbre:
Database connection lifetime debugging.

Debug only.

DO NOT MERGE.


thepaul:
This is hawt.

egonelbre:
Note this can be modified to wrap all methods in https://golang.org/pkg/database/sql/#Conn and then track the last use and last use time.

egonelbre:
Closing, but this can still be found from closed [Debug] PR-s.

PR #1988, opened 2019-05-17
storage/postgreskv: use batch size 1000

egonelbre:
Currently we are using batch-size 10000, by benchmarking an unrelated thing I noticed that by using 1000 there was a significant perf win.

```
go test -v -benchtime 10x -bench BenchmarkSuiteLong -run XXX ./storage/postgreskv --test-bench-long paths.data.gz --test-bench-long-noinit --test-bench-long-noclean
```

Before:

```
BenchmarkSuiteLong/DeepRecursive-32                   10        4819822700 ns/op          10.63 MB/s
BenchmarkSuiteLong/DeepRecursiveReverse-32            10        4632797470 ns/op           9.84 MB/s
BenchmarkSuiteLong/DeepNonRecursive-32                10        1722895830 ns/op           0.00 MB/s
BenchmarkSuiteLong/DeepNonRecursiveReverse-32                 10        1340622790 ns/op           0.00 MB/s
BenchmarkSuiteLong/ShallowRecursive-32                        10         521777000 ns/op           2.37 MB/s
BenchmarkSuiteLong/ShallowRecursiveReverse-32                 10         523726730 ns/op           2.36 MB/s
```

After:

```
BenchmarkSuiteLong/DeepRecursive-32                   10         767183610 ns/op          66.81 MB/s
BenchmarkSuiteLong/DeepRecursiveReverse-32            10         735170270 ns/op          61.99 MB/s
BenchmarkSuiteLong/DeepNonRecursive-32                10        1745487930 ns/op           0.00 MB/s
BenchmarkSuiteLong/DeepNonRecursiveReverse-32                 10        1361691690 ns/op           0.00 MB/s
BenchmarkSuiteLong/ShallowRecursive-32                        10          89862570 ns/op          13.75 MB/s
BenchmarkSuiteLong/ShallowRecursiveReverse-32                 10          91827690 ns/op          13.46 MB/s
```

Thanks for submitting a PR!

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
[...]

PR #1992, opened 2019-05-17
Revert "storage/postgreskv: use batch size 1000"

thepaul:
Reverts storj/storj#1988

My own tests are showing the opposite; using 10000 is faster than 1000 with the setup I have (with PostgreSQL running on the same machine, macOS, and connecting to PG over a unix domain socket).

But as far as I can tell, defaultBatchSize _shouldn't_ affect these tests, so whether this change makes it faster or slower, I want to understand why before we commit the change.

### DIFF
--- a/storage/postgreskv/client.go
+++ b/storage/postgreskv/client.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	defaultBatchSize = 1000
+	defaultBatchSize = 10000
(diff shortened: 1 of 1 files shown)


thepaul:
Egon was using PostgreSQL running on the same machine also, but with Windows 11 and 32 cores. My test machine here is a 2018 macbook pro.

PR #1996, opened 2019-05-19
Value Attribution design doc

coyle:
What: Value Attribution Design Document

Why: Feedback on design of value attribution

Thanks for submitting a PR!


docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+### Database
+The attribution table will consist of data that allows for ease of calculating total attribution for a partner. 
+
+| Name  | Type |
+| ------------- | ------------- |
+| bucket_id (pk) | uuid  | 
+| user_id  | uuid  |
+| partner_id  | uuid  |
+| total_data | text  |
+| last_updated | timestamp  |
```
mniewrzal: What is `user_id`? What will be stored in `total_data` and why its just `text`?

coyle: I changed to `integer`, total_data is the amount of data attributed to a partner_id within a bucket

mniewrzal: And what is `user_id`?

coyle: the id of the user who owns the bucket

docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+
+	* The partner does NOT receive attribution for data stored or retrieved from that bucket. 
+
+* A client points partner connector A to an existing bucket that partner connector B had already been pointed to. 
+
+	* Partner connector B receives the attribution for all of the data that is stored and retrieved from the bucket. 
+
+## DESIGN
+### Connector
+Each partner will have a registered uuid that will identify a partners connector on the Storj network. Each connector will implement the S3 client interface to allow operations through the connector.  When a user uploads data to a specified bucket through the connector,  the connector will include the partner uuid in the context of the request. Before an upload occurs, the uplink will communicate the uuid and bucket name with the tardigrade satellite, checking for a previous attribution. If no attribution is found on the specified bucket and the bucket is currently void of data, the satellite will attribute the partners uuid to that bucket within the metadata struct.
```
mniewrzal: So Value Attribution will be limited only to S3 client interface?

docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Database
+The attribution table will consist of data that allows for ease of calculating total attribution for a partner. 
+
+| Name  | Type |
+| ------------- | ------------- |
+| bucket_id (pk) | uuid  | 
+| user_id  | uuid  |
+| partner_id  | uuid  |
+| total_data | integer  |
```
mniewrzal: In scenarios there is info that `all of the data that the client stores and retrieves from it`. will we be paying the same price for storage and download bandwidth? Or we need to distinguish it?

docs/design/value-attribution.md
jtolds: a few thoughts:
1) we probably don't need a full UUID. anything clearly identifying the connector is fine. The way GCP and Azure work is they use the User Agent string. (Here is Minio setting the User Agent for Azure: https://github.com/minio/minio/blob/master/cmd/gateway/azure/gateway-azure.go#L184, where Minio plays the part of connector and Azure is us).
2) connector attribution should work for anything that uses libuplink of any kind - so there isn't necessarily an S3 interface. But libuplink should allow for setting some sort of identifier (UUID, user agent) before performing operations on a satellite.

docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+
+	* Partner connector B receives the attribution for all of the data that is stored and retrieved from the bucket. 
+
+## DESIGN
+### Connector
+Each partner will have a registered uuid that will identify a partners connector on the Storj network. Each connector will implement the S3 client interface to allow operations through the connector.  When a user uploads data to a specified bucket through the connector,  the connector will include the partner uuid in the context of the request. Before an upload occurs, the uplink will communicate the uuid and bucket name with the tardigrade satellite, checking for a previous attribution. If no attribution is found on the specified bucket and the bucket is currently void of data, the satellite will attribute the partners uuid to that bucket within the metadata struct.
+
+After the connector has confirmed the partners uuid was successfully attributed to a bucket, the connector will notify a tardigrade satellite that the attribution was confirmed. The satellite will then look up the bucket, verify that partner has the attribution and add the necessary data to the Attribution table.
```
jtolds: isn't the previous step already enough? the satellite knows of the connection. what is the separate operation for?

docs/design/value-attribution.md
coyle: Was an extraneous sentence explaining how connectors will be built. My assumption they would implement the S3 interface was wrong but the way this works doesn't require S3 compatibility. 
`the connector will include the partner uuid in the context of the request`.

docs/design/value-attribution.md
coyle: That's a good question. @iglesiasbrandon ?

docs/design/value-attribution.md
coyle: Separation of concerns. If we had the satellite update the value attribution database when the metadata is added, we need to include information about value attribution within the meta service.

jtolds: that's fine but the performance hit is too costly. this adds another round trip from the client to the satellite.

coyle: 👍

docs/design/value-attribution.md
iglesiasbrandon: yes, but we wouldn't discount the actual price of bandwidth/ egress we would give the client credits to cover the discount; so we would need the ability to calculate how much bandwidth and egress a bucket used.

@coyle @jtolds @mniewrzal

docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+
+	* The partner does NOT receive attribution for data stored or retrieved from that bucket. 
+
+* A client points partner connector A to an existing bucket that partner connector B had already been pointed to. 
+
+	* Partner connector B receives the attribution for all of the data that is stored and retrieved from the bucket. 
+
+## DESIGN
+### Connector
+Each partner will have a registered id, (which we will refer to as the partner id) that will identify a partners connector on the Storj network.  When a user uploads data to a specified bucket through the connector,  the connector will include the partner id in the context of the request. Before an upload occurs, the uplink will communicate the partner id and bucket name with the tardigrade satellite, checking for a previous attribution. If no attribution is found on the specified bucket and the bucket is currently void of data, the satellite will attribute the partners id to that bucket within the metadata struct. Concurrently to updating the metadata struct the satelitte will add the necessary data to the Attribution table.
```
zeebo: Why isn't the partner id specified during bucket creation instead of this lazily on first write system?

coyle: because the bucket can be created months before a bucket is associated with a partner

coyle: as long as the bucket is empty and has no previous attribution, we want to allow a partner to be attributed to that bucket.

docs/design/value-attribution.md
```diff
@@ -0,0 +1,79 @@
...
+Each partner will have a registered id, (which we will refer to as the partner id) that will identify a partners connector on the Storj network.  When a user uploads data to a specified bucket through the connector,  the connector will include the partner id in the context of the request. Before an upload occurs, the uplink will communicate the partner id and bucket name with the tardigrade satellite, checking for a previous attribution. If no attribution is found on the specified bucket and the bucket is currently void of data, the satellite will attribute the partners id to that bucket within the metadata struct. Concurrently to updating the metadata struct the satelitte will add the necessary data to the Attribution table. 
+
+### Database
+The attribution table will consist of data that allows for ease of calculating total attribution for a partner. 
+
+| Name  | Type |
+| ------------- | ------------- |
+| bucket_id (pk) | uuid  | 
+| user_id  | uuid  |
+| partner_id  | uuid  |
```
jtolds: maybe this is a string instead of a uuid?

coyle: My issue with strings as IDs is historically I've run into normalization issues. If you think the space saving is a better tradeoff than clarity I'm all for swapping

jtolds: if the ID is indeed a user-agent though, it's not a UUID

jtolds: or is there another table of partner id to partner info mappings?

coyle: My thought was we would have a map of `[valid user-agent]uuid` since we authorize partners before allowing them attribution, we would add them to that list.

coyle: and not a table (although we could) but a config on satellite

coyle: benefit of this would be if a partner has multiple user-agents

docs/design/value-attribution.md
zeebo: if the bucket has been created months ago, but this only works if it's empty, why not just make them recreate the bucket?

like, the age of the bucket doesn't matter, what matters is that it's empty, and if it's empty, recreation is trivial.

docs/design/value-attribution.md
simongui: Also if the first RPC call succeeds but the file upload fails we have more potential dangling entries, am I understanding that correctly? Combining it into the existing RPC call makes that less possible?

docs/design/value-attribution.md
coyle: I guess I don't see the benefit of `create bucket x with no attribution, delete bucket x, create bucket x with attribution` vs `create bucket x with no attribution, add attribution`

docs/design/value-attribution.md
coyle: After talking with @mniewrzal I don't believe we will add it to the metadata struct anyways, since it could be changed by the user.

docs/design/value-attribution.md
simongui: My knowledge of the system is lacking still so this may be a dumb question but do we have to worry about someone say taking `uplink` and modifying it to generate millions of user-agents that would then cause our attribution service to crank on endless amounts of fake work and causing a lot of load on our database?

coyle: They would have to have a signed certificate to upload anything so this is prevented same as we would do for any user trying to ddos us. This case in particular though, the mappings of user agents to partner ids would bypass any value attribution if the user-agent was not an authorized partner.

docs/design/value-attribution.md
zeebo: So if I'm understanding correctly, if you have your uplink command set with a partner id and you upload to a bucket with data in it, it will just happily allow it. If that's the case, I can imagine a user attempting to set a partner id on a bucket that isn't empty and getting no feedback that it never worked.

If setting a partner id is an explicit action, there's an opportunity to inform the user when something went wrong. For example, a command like `uplink set-partner-id sj://bucket partner-id` could say "sorry! the bucket has to be empty".

Like, if no buckets existed, is this the design we would use? If so, great. If not, we're doing something different from ideal to handle migrating the case of empty buckets, which seems like a small thing to care about.

zeebo: I've had it explained to me why lazily allocating the bucket to a partner id is the ideal design.

JessicaGreben: @zeebo it would be nice to know why "lazily allocating the bucket to a partner id is the ideal design" for future reference.

PR #2000, opened 2019-05-19
Store repaired Segments and improve Repair Condition

littleskunk:
What: 
1. Repair job should not cutoff longtail.
2. Commit repair pieces even if not hitting the success threshold
3. Remove useless condition.

Why:
1. Main objective for the repair job is file durability and not upload speed.
2. If we managed to upload to 79 instead of 80 nodes we should still write it back to the database. There is no reason to throw an error.
3. `healty > min && healty > success` is a bit useless. `healty > success` will produce the same results because `success > min`

Thanks for submitting a PR!

### DIFF
--- a/pkg/storage/ec/client.go
+++ b/pkg/storage/ec/client.go
@@ -218,7 +218,7 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
 
 	timer := time.AfterFunc(timeout, func() {
 		if ctx.Err() != context.Canceled {
-			zap.S().Infof("Timer expired. Successfully repaired to %d nodes. Canceling the long tail...", atomic.LoadInt32(&successfulCount))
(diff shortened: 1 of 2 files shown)


pkg/storage/ec/client.go
```diff
@@ -267,8 +260,8 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
 		}
 	}()
 
-	if successfulCount < int32(optimalCount) {
-		return nil, nil, Error.New("successful nodes count (%d) does not match optimal count (%d) of erasure scheme", successfulCount, optimalCount)
+	if successfulCount == 0 {
+		return nil, nil, Error.New("successful nodes count %d", successfulCount)
```
kaloyan-raev: As `successfulCount` is 0 here for sure, perhaps we can have a clearer error message like "repair to all nodes failed".

PR #2015, opened 2019-05-21
Bugfix: Repair increase success counter

littleskunk:
What: Increase success counter.

Why: I removed an important counter with one of my latest PRs. Without that counter the repair service will always fail.

Thanks for submitting a PR!

### DIFF
--- a/pkg/storage/ec/client.go
+++ b/pkg/storage/ec/client.go
@@ -166,7 +166,7 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, r
 	}()
 
 	if int(atomic.LoadInt32(&successfulCount)) < rs.RepairThreshold() {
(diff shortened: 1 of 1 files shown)


pkg/storage/ec/client.go
```diff
@@ -261,7 +262,7 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
 	}()
 
 	if successfulCount == 0 {
```
mniewrzal: Can you use here `atomic.LoadInt32(&successfulCount)`?

littleskunk: Yes. I have no idea what the difference is. Can you explain that to me?

mniewrzal: Its making access to `successfulCount` atomic which is important when we are working with goroutines to avoid concurrent access.

PR #2018, opened 2019-05-21
using DB time only

wthorp:
What: 
Use only database time in repair queue

Why:
Trying to track down a mystery bug, this will rule out one potential

### DIFF
--- a/satellite/satellitedb/repairqueue.go
+++ b/satellite/satellitedb/repairqueue.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"time"
 
 	"github.com/lib/pq"
 	sqlite3 "github.com/mattn/go-sqlite3"
@@ -35,13 +34,12 @@ func (r *repairQueue) Insert(ctx context.Context, seg *pb.InjuredSegment) error 
 }
 
 func (r *repairQueue) postgresSelect(ctx context.Context) (seg *pb.InjuredSegment, err error) {
-	err = r.db.QueryRowContext(ctx, r.db.Rebind(`
-	UPDATE injuredsegments SET attempted = ? WHERE path = (
+	err = r.db.QueryRowContext(ctx, `
+	UPDATE injuredsegments SET attempted = now() WHERE path = (
 		SELECT path FROM injuredsegments
-		WHERE attempted IS NULL
-		OR attempted < now() - interval '1 hour'
+		WHERE attempted IS NULL OR attempted < now() - interval '1 hour'
 		ORDER BY path FOR UPDATE SKIP LOCKED LIMIT 1
-	) RETURNING data`), time.Now().UTC()).Scan(&seg)
+	) RETURNING data`).Scan(&seg)
 	if err == sql.ErrNoRows {
 		err = storage.ErrEmptyQueue.New("")
 	}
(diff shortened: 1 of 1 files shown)


satellite/satellitedb/repairqueue.go
```diff
@@ -35,13 +34,12 @@ func (r *repairQueue) Insert(ctx context.Context, seg *pb.InjuredSegment) error
 }
 
 func (r *repairQueue) postgresSelect(ctx context.Context) (seg *pb.InjuredSegment, err error) {
-	err = r.db.QueryRowContext(ctx, r.db.Rebind(`
-	UPDATE injuredsegments SET attempted = ? WHERE path = (
+	err = r.db.QueryRowContext(ctx, `
+	UPDATE injuredsegments SET attempted = now() WHERE path = (
```
thepaul: This will be a subtle problem: because `attempted` is a timestamp without time zone, it will get a different value depending on what the db server considers to be the local time zone.

If it's an option to change to `TIMESTAMP WITH TIME ZONE`, that is to be preferred-- it won't matter what the server or your machine thinks is the local time zone, since any local times will have the time zone info annotated, and the db can always find the "real" time from that.

But since that's probably not an option, I suggest you use `now() AT TIME ZONE 'utc'` instead of `now()` (both on this line and down on line 40)

littleskunk: Bill and me tried a few different statements on the database. We should use `timezone('utc', now())`

```
SELECT attempted, timezone('utc', now()) FROM injuredsegments WHERE attempted IS NOT NULL AND attempted > timezone('utc', now()) - interval '1 hour';
         attempted          |          timezone
----------------------------+----------------------------
 2019-05-21 16:08:30.609984 | 2019-05-21 16:09:43.998147
```

satellite/satellitedb/repairqueue.go
```diff
@@ -59,7 +57,7 @@ func (r *repairQueue) sqliteSelect(ctx context.Context) (seg *pb.InjuredSegment,
 		if err != nil {
 			return err
 		}
-		res, err := tx.Tx.ExecContext(ctx, r.db.Rebind(`UPDATE injuredsegments SET attempted = ? WHERE path = ?`), time.Now().UTC(), path)
+		res, err := tx.Tx.ExecContext(ctx, r.db.Rebind(`UPDATE injuredsegments SET attempted = datetime('now') WHERE path = ?`), path)
```
littleskunk: Should we use `timezone('utc', now())` here as well?

wthorp: sqlite supposed uses UTC by default

satellite/satellitedb/repairqueue.go
wthorp: updated to timezone('utc', now())

thepaul: That should be equivalent. But does that mean `now() AT TIME ZONE 'utc'` does not work? That would suggest a larger problem.

satellite/satellitedb/repairqueue.go
thepaul: sqlite doesn't have the `timezone()` function, and doesn't understand the standard `AT TIME ZONE` syntax either. Bill is right, though: sqlite3 will use UTC. `datetime('now')` should be fine. And even if it wasn't, there isn't much risk of a sqlite3 database being used by multiple clients with different local time zones set.

satellite/satellitedb/repairqueue.go
wthorp: timezone('utc', now()) works.  jen's test is successful when it returns a result.

thepaul: Right, but does `now() AT TIME ZONE 'utc'` _not_ work? That's the more standard (and also easier to read imo) way to write it

littleskunk: `now() AT TIME ZONE 'utc' works as well. Only `now()` is not working.

thepaul: Ok. Just making sure.

wthorp: oh, they should be equivalent.  I was just going the opposite direction on 'easier to read'

thepaul: To my feeble mind, "`timezone(foo)`" always seems like it's going to return a time zone. But it's totally a matter of preference and what you've got should work fine.

PR #636, opened 2018-11-13
cmd/statreceiver: lua-scriptable stat receiver

jtolds:
Change-Id: I3ce0fe3f1ef4b1f4f27eed90bac0e91cfecf22d7

PR #2019, opened 2019-05-21
cmd/statreceiver: allow for packet filtering to packet destinations

jtolds:
What: allow packetfilter to work on packet destinations instead of only on metric destinations. this will allow us to filter what applications get sent to rothko. 

Why: currently rothko is drowning in storj-sim data and it'd be nice to filter out.

Thanks for submitting a PR!


cmd/statreceiver/filter.go
```diff
@@ -5,29 +5,56 @@ package main
...
-func NewPacketFilter(applicationRegex, instanceRegex string) *PacketFilter {
+// NewPacketFilter creates a PacketFilter. It takes a packet destination,
+// an application regular expression, and an instance regular expression.
+// If the regular expression is matched, the packet will be passed through.
+func NewPacketFilter(applicationRegex, instanceRegex string, dest PacketDest) *PacketFilter {
 	return &PacketFilter{
 		application: regexp.MustCompile(applicationRegex),
 		instance:    regexp.MustCompile(instanceRegex),
+		dest:        dest,
+		scratch: sync.Pool{
```
JessicaGreben: why do you call this scratch?

cmd/statreceiver/filter.go
```diff
@@ -5,29 +5,56 @@ package main
...
+			New: func() interface{} {
+				var x [10 * kb]byte
```
JessicaGreben: I agree with this comment, using `memory.KB` is more readable and would be consistent throughout the codebase
https://github.com/storj/storj/pull/636#discussion_r234138547

cmd/statreceiver/parser.go
```diff
@@ -54,12 +54,6 @@ func (p *Parser) Packet(data []byte, ts time.Time) (err error) {
 	}
 
 	app, inst := string(appb), string(instb)
-	for _, filter := range p.filters {
-		if !filter.Filter(app, inst) {
-			return nil
```
JessicaGreben: the function `NewParser` still takes `filters` as an argument. Should that be removed so it only accepts `dest MetricDest` now since it isn't using the filters anymore?

cmd/statreceiver/example.lua
```diff
@@ -33,12 +33,7 @@ metric_handlers = mcopy(
...
-  parse(  -- parse takes one or two arguments. the first argument is
-          -- a metric handler, the remaining one is a per-packet application or
-          -- instance filter. each filter is a regex. all packets must
-          -- match all packet filters.
-    sanitize(metric_handlers), -- sanitize converts weird chars to underscores
-    packetfilter("storagenode-prod|satellite-prod|uplink-prod", ""))
+  parse(sanitize(metric_handlers)) -- sanitize converts weird chars to underscores
 
 -- pcopy forks data to multiple outputs
 -- output types include parse, fileout, and udpout
```
JessicaGreben: should this comment be updated to include `packetfilter`?

cmd/statreceiver/parser.go
jtolds: > the function NewParser still takes filters as an argument.
whoops! you're right! oversight. the parser no longer needs filters of any kind

there's basically two levels of data - packets (groups of metrics) and metrics (just the key/values themselves). a parser turns a packet into metrics. because all of the filtering now operates on packets and metrics, the parser no longer needs to do it

cmd/statreceiver/filter.go
jtolds: fair

cmd/statreceiver/filter.go
jtolds: this is a scratch memory space pool. we have a well defined memory lifecycle (parse a packet into some new area of memory, do stuff, then release the memory), so if we use a sync.Pool here, we can relieve pressure on the garbage collector by managing the memory lifecycle ourselves.

cmd/statreceiver/example.lua
jtolds: sure!

cmd/statreceiver/parser.go
JessicaGreben: this is very helpful "a parser turns a packet into metrics", i couldn't figure it out just from reading the code.

PR #2020, opened 2019-05-21
Password based key derivation design

zeebo:
What: Adds a design doc for how to derive keys from passwords

Why: To make sure we get it all right, consistent, and secure.

[Rendered](https://github.com/storj/storj/blob/jw/password-key-derivation/docs/design/password-key-derivation.md)

---

In the interest of expediting this design doc and implementation, I ask that the reviewers be mindful and limit their suggested changes to those that are

1. Actionable ("please do this concrete thing")
2. Strongly held (a showstopper for you)

Doing this will also help reduce how long it will take me to do the suggestions because I'll assume that they're very important to you. **Questions are always welcome**, and thanks for always giving such great feedback.


docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,65 @@
...
+1. It's based on the user's choice (for example, a password).
+2. If the user inputs the same password on different machines, the same root key should be derived.
+
+These requirements allow users to be in full control of their encryption, and don't require users safely transporting high entropy (hard to remember) secrets to bootstrap new uplinks.
+
+This design accomodates more requirements that allow for additional features:
+
+3. A root key can be derived for any path in a bucket.
+4. Buckets root keys should not depend on the name of the bucket.
+5. A table of keys for low entropy passwords should not exist.
```
jtolds: this could use more explanation

jtolds: oh, in the following paragraph. okay nevermind

zeebo: > 5. A table of keys for low entropy passwords should not be possible. In other words, an attacker with knowledge of the algorithm should not be able to use a dictionary of common passwords and pre-compute what keys to check in the event of a data breach.

?

jtolds: :+1:

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. It's based on the user's choice (for example, a password).
+2. If the user inputs the same password on different machines, the same root key should be derived.
+
+These requirements allow users to be in full control of their encryption, and don't require users safely transporting high entropy (hard to remember) secrets to bootstrap new uplinks.
+
+This design accomodates more requirements that allow for additional features:
+
+3. A root key can be derived for any path in a bucket.
+4. Buckets root keys should not depend on the name of the bucket.
```
thepaul: I can see a rationale for having the _possibility_ of bucket encryption keys that aren't derived from the root, but I disagree pretty strongly that that should be the only option. Since this is talking about features, does it mean to say `Bucket root keys that do not depend on the name of the bucket` instead?

zeebo: The goal of this point to not use the bucket name to derive the key in the default case so that we only need to worry about the server side to perform renames. Specifically, the encrypted paths depend on the bucket key, so if renaming the bucket changes what key would be used, then we would have to re-encrypt all of the paths with the new key.

Why is it important to support deriving a bucket key from some root key and the bucket name in particular? This design supports deriving bucket keys from some root key and some bucket entropy that isn't tied to the name.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,66 @@
...
+## Design
+
+A default password can be chosen. There is no extra entropy or information available, so it will be run through [Argon2](https://en.wikipedia.org/wiki/Argon2) with a very high work load and a fixed salt. The output of that will be stored in a secure location. The higher the workload, the better, as it increases the cost for an offline attack, and only the output is used in subsequent steps, so it only needs to happen once.
+
+Satellites will add a new field to bucket metadata called `entropy` that just contains 32 random, stable bytes.
+
+The algorithm used to derive a root key for some path in some bucket for some secret is
+
+```
+bucketEntropy = getBucketMetadata(bucketName).entropy
```
thepaul: This seems at first blush to require an extra round-trip to the satellite. Is that accurate, or are we expecting to piggyback on a request that already needs to be done at around the same time? I'm very hesitant to add anything requiring more coordination steps.

docs/design/password-key-derivation.md
thepaul: Oh okay. I misunderstood the point of this doc.

I'm not convinced that the existing bucket situation is problematic. S3 seems to have done just fine without supporting bucket renames at all. If we can support renames without any extra coordination steps, then that sounds great, but I'm not sure how it will work.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,66 @@
...
+
+The third requirement allows having multiple encrypted domains to exist within a single bucket, allowing delegation of encryption. The fourth requirement allows renaming buckets without having to re-encrypt all of the paths inside of a bucket. The fifth requirement enhances security if a satellite breach happens. This is usually accomplished with [salting](https://en.wikipedia.org/wiki/Salt_(cryptography)). The sixth requirement allows providing a finite amount of information to a third party that allows them access to the keys for any bucket created in the future.
+
+## Design
+
+A default password can be chosen. There is no extra entropy or information available, so it will be run through [Argon2](https://en.wikipedia.org/wiki/Argon2) with a very high work load and a fixed salt. The output of that will be stored in a secure location. The higher the workload, the better, as it increases the cost for an offline attack, and only the output is used in subsequent steps, so it only needs to happen once.
+
+Satellites will add a new field to bucket metadata called `entropy` that just contains 32 random, stable bytes.
```
thepaul: What about using the project UUID as the salt, to avoid having to ask the server for it? I don't recall whether we were planning to include that in our sharing bundles.

docs/design/password-key-derivation.md
zeebo: I agree. This shouldn't require extra coordination steps.

docs/design/password-key-derivation.md
zeebo: It can piggyback off of the already necessary request to get the bucket metadata.

docs/design/password-key-derivation.md
zeebo: It's not included in any sharing bundles or api key or otherwise. It would require doing an extra network request over what we already do.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,66 @@
...
+- The default password can be used for any bucket and will return a distinct key, and the output of the default password derivation can be shared with others to allow them to do the same.
+
+Some other points of consideration include
+
+- Clients can ignore the bucket entropy if they wish and know they have a strong password. Indeed, they can ignore all of this and do their own key derivation.
+- The entropy is not blindly trusted from the satellite, and is mixed with the secret in an HMAC step that ensures a hostile satellite cannot predict what the salt will be.
+- HMAC provides security against forgeries without knowing the secret, which implies that sharing the output of a HMAC does not leak information about the secret, even under attacker provided data, so I believe the usage of HMAC is safe.
+- The entropy can be used for other purposes in the future if necessary.
+- Users will already lose data if they do not back up their data pointers in the satellite (because the encryption key for the data is only stored there), so adding the bucket entropy does not add a requirement that users need to back up data in a satellite, it just adds a small amount of additional information.
```
thepaul: Not sure what this means. It sounds like it's saying the encryption key for data is stored only on a satellite?

zeebo: The entropy for the bucket used to derive the salt is stored only on the satellite, just like the actual encryption key used to encrypt the data is stored only on the satellite.

docs/design/password-key-derivation.md
thepaul: It would not require any network access if we added the project UUID to sharing bundles or api keys. Or just a prefix of the project UUID.

docs/design/password-key-derivation.md
thepaul: Is that request actually necessary? What info is contained in the bucket metadata that is necessary to access/read/upload objects in that bucket?

docs/design/password-key-derivation.md
thepaul: Oh, referring to the encrypted content keys. Gotcha.

docs/design/password-key-derivation.md
jtolds: The encryption algorithm used for path encryption.

zeebo: https://github.com/storj/storj/blob/master/pkg/storj/object.go#L26-L34
https://github.com/storj/storj/blob/master/lib/uplink/project.go#L34-L58

Looks like path encryption scheme, encryption parameters for the data, reed-solomon parameters, etc.

docs/design/password-key-derivation.md
zeebo: Indeed, at the cost of making the share larger. I agree this will work if we can't piggyback, but I believe we can.

docs/design/password-key-derivation.md
thepaul: Ok, true. I withdraw my objection.

docs/design/password-key-derivation.md
jtolds: That does make the API key longer

thepaul: Given that we need to ask the server anyway, I'm fine with per-bucket entropy. As a final question, should we not add that as a field to the bucket-level pointer object, since we need to fetch that anyway for the bucket metadata? Or are bucket-level pointers going away to be replaced by an expanded nodes table? I remember discussing that at one point.

thepaul: Another unrelated question: would it not be better to rename this `entropy` field to `bucketID` or `bucketUUID`? Because this field is essentially becoming a persistent ID to take the place of the bucket name, which will no longer be permanent. Besides being more self-explanatory in purpose, it also makes it more obvious that the bytes contained by the field are entirely public and don't need to be guarded or anything.

OK, yes, this does not actually meet the criterion of being a total showstopper for me, but I do think it would aid future spelunking expeditions.

docs/design/password-key-derivation.md
zeebo: One thing this conversation helped me realize is that we only actually need the salt information when deriving a key from a password. We could be storing the derived key in the share information and using that.

In that case, I think if the proposal changes to allow deriving the root most key with some salt help from the satellite, we can derive default bucket keys without any help, and so the only time network requests happen to the satellite is when the user is inputting a password.

I'll think on that and possibly update this.

thepaul: (I do know about `storj.Bucket` and `uplink.BucketConfig`, but none of those fields are necessary for fetching an object in the bucket, with the exception of `PathCipher` for the case where you don't already know the encrypted path. The other fields are all used as bucket defaults, if they are used at all)

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,66 @@
...
+A default password can be chosen. There is no extra entropy or information available, so it will be run through [Argon2](https://en.wikipedia.org/wiki/Argon2) with a very high work load and a fixed salt. The output of that will be stored in a secure location. The higher the workload, the better, as it increases the cost for an offline attack, and only the output is used in subsequent steps, so it only needs to happen once.
+
+Satellites will add a new field to bucket metadata called `entropy` that just contains 32 random, stable bytes.
+
+The algorithm used to derive a root key for some path in some bucket for some secret is
+
+```
+bucketEntropy = getBucketMetadata(bucketName).entropy
+bucketRootKey = argon2id(salt:hmac(hash:sha256, secret:secret, data:bucketEntropy), secret:secret)
+pathRootKey   = hmac(hash:sha256, secret:bucketRootKey, data:encryptedPath)
```
thepaul: Oh, wait, what is `encryptedPath` in this context, if we are deriving s<sub>0</sub>?

Or does the below paragraph mean that `bucketRootKey` is s<sub>0</sub>, and `pathRootKey` is just being given as an example path derivation step for an encryptedPath that has exactly one path component?

zeebo: Nope, this is for a use case where, for example, someone has a bucket, and they want to allow users to have their own paths in that bucket, and the want the users to have their own encryption keys.

In that case, the s<sub>0</sub> exists at, maybe, `sj://bucket/user1`.

zeebo:
Sorry to pull a huge 180 on the proposal mid way, but I think the conversation made me realize that we could only do network requests during project/bucket setup/imports.

I think this design is a bit simpler and more direct: the satellite is only used to store and add some entropy in the derivation of the keys.

docs/design/password-key-derivation.md
zeebo: The proposal has changed since this last comment was made to be an explicit call. I'll happily take suggestions on what to name it.

docs/design/password-key-derivation.md
thepaul: Ah, I hadn't thought of that need. Clever.

docs/design/password-key-derivation.md
thepaul: I'm fine with GetSaltEntropy.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,76 @@
...
+## Design
+
+Uplink always stores derived keys, never the input material.
+
+The satellite will add a new endpoint, `GetSaltEntropy` that uses the api key to find the project, and optionally takes a bucket name. It is important that the result is stable, even if the bucket is renamed. For example the result of `GetSaltEntropy("bucket1")` should be the same as `GetSaltEntropy("bucket2")` even after `"bucket1"` is renamed to `"bucket2"`. Note that there are still significant challenges to bucket renaming, and we don't support it right now.
+
+When deriving a root key from an api key, password, optional bucket name and optional encrypted path, the following algorithm will be used
+
+```
+saltEntropy = GetSaltEntropy(bucketName or None)
```
thepaul: What happens when bucketName is None? Is it still meaningful to ask for a salt in that case? Would this be equivalent to returning the project UUID?

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,76 @@
...
+mixedSalt   = hmac(hash:sha256, secret:password, data:saltEntropy)
+pathSalt    = hmac(hash:sha256, secret:mixedSalt, data:encryptedPath or "")
+rootKey     = argon2id(salt:salt, password:password)
```
thepaul: `salt:salt` here is presumably `salt:pathSalt` or `secret:pathSalt`? I don't know what this colon syntax is.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,76 @@
...
+- Each root key requires some secret that necessarily comes from the user in some way.
+- The algorithm is deterministic, so the user gets the same keys on any machine where she inputs the same secrets.
+- Root keys for a sub-path in a bucket are accomplished by providing a non-empty encrypted path when deriving the root key.
+- Root keys do not depend on the name of the bucket they are in: just the salt entropy associated with that bucket, allowing renames.
+- Because the entropy is added as the salt in the `argon2id` step, dictionary attacks on low-entropy passwords cannot be precomputed.
+- The default password can be used for any bucket and will return a distinct key as long as the salt entropy differs.
+
+Some other points of consideration include
+
+- Clients can ignore the salt entropy if they wish and know they have a strong password. Indeed, they can ignore all of this and do their own key derivation.
```
thepaul: Can we specify what level of library support we intend to provide for such a use case? Would they have to reimplement large parts of libuplink or will there be a way to just supply your own `rootKey` directly? I think that would probably be the route to take for people who already have high-entropy keys (a separate keyfile, a key from a hardware token, etc).

docs/design/password-key-derivation.md
zeebo: Yes, this would be some random bytes associated to the project. I'm reluctant to tie it to the UUID, but yes, it's that.

docs/design/password-key-derivation.md
zeebo: I just now realized that while I was trying to do python kwargs, it uses `=` instead of `:`. It should be `pathSalt`.

docs/design/password-key-derivation.md
zeebo: I don't think we need to have great support or anything, but a good goal would be to allow people to just input whatever bytes they want into their configuration entries for the keys and have it work.

Those details of course depend on how the keys are stored exactly, but I think we should be storing them such that minimal transformations are applied to them before use.

thepaul: Please to add a statement to that effect (at the minimum, we will allow the user to specify an already-formed encryption key if they don't want one derived from a passphrase)

docs/design/password-key-derivation.md
thepaul: What would one do with the project-level random bytes? It doesn't seem like it would help in creating buckets. I gather that the intent is probably to allow a way for a whole project to share one passphrase, but there doesn't seem to be any way left for bucket encryption keys to be derived from a project root encryption key. Maybe the whole-project-passphrase use case just reduces to storing the passphrase and using it for each bucket opened, as needed?

docs/design/password-key-derivation.md
zeebo: :+1:

docs/design/password-key-derivation.md
zeebo: > What would one do with the project-level random bytes? It doesn't seem like it would help in creating buckets.

Good point. I didn't think this through well enough.

> Maybe the whole-project-passphrase use case just reduces to storing the passphrase and using it for each bucket opened, as needed?

That's essentially what the first version of the proposal was: we stored the output of the passphrase run through argon2 at a high work factor, and used that as the passphrase for any default bucket using a specific bucket salt provided by the satellite.

The thing this second proposal attempted to avoid was requesting the salt unless a passphrase was being turned into a root key. We could do something like `bucketRootKey = hmac(projectRootKey, bucketName)` the but that then means we give up on renaming buckets.

---

I have three options and I don't know which is best:

## Option 1

Update this to bring back the whole-project-passphrase and describe that clients should generate a default bucket key using that whole-project-passphrase lazily and store the result in the scope.

Then we only make network requests for a salt when using a passphrase, or the first time a new bucket is accessed that uses the whole-project-passphrase.

## Option 2

Go back to the original proposal and have each bucket just contain some stable salt data that is provided to the client whenever they request path encryption information.

## Option 3

Give up on bucket renames forever and do something else entirely.

zeebo: Optimistically implemented option 2.

docs/design/password-key-derivation.md
```diff
@@ -26,51 +26,52 @@ The third requirement allows having multiple encrypted domains to exist within a
...
+A default password can be chosen for an api key. The following algorithm is used to derive the default password:
 
 ```
-saltEntropy = GetSaltEntropy(bucketName or None)
+saltEntropy     = getProjectSalt(apiKey)
+mixedSalt       = hmac(hash=sha256, secret=userPassword, data=saltEntropy)
+defaultPassword = argon2id(salt=mixedSalt, password=userPassword)
+```
+
+The following algorithm is used to deriving a root key from an api key, password, optional bucket name and optional encrypted path:
```
thepaul: Bucket name is no longer optional, if I'm understanding correctly.

docs/design/password-key-derivation.md
```diff
@@ -26,51 +26,52 @@ The third requirement allows having multiple encrypted domains to exist within a
...
+- Users will already lose data if they do not back up their data pointers in the satellite (because the encryption key for the data is only stored there), so adding the salts does not add a requirement that users need to back up data in a satellite, it just adds a small amount of additional information they need to back up.
 
 ## Implementation
 
-First, satellites must implement the `GetSaltEntropy` call. This can be done by adding a field on buckets and project to store the entropy. It can either run a migration and fill them all at once, or fill it on demand. It should use random bytes from the Go `crypto/rand` package, and be large enough to handle all forseeable usage (32 bytes should be sufficient).
+First, satellites must add a field on buckets and projects to store the salts. It can either run a migration and fill them all at once, or fill it on demand. It should use random bytes from the Go `crypto/rand` package, and be large enough to handle all forseeable usage (32 bytes should be sufficient).
```
thepaul: I'll try and make my case just one more time here 😄: in the case of projects, we already have a guaranteed-unique sequence of bytes associated with each one, and that is the UUID. To add another field as the salt seems to me to suggest that the new field has some different quality from the UUID bytes, and that said difference is important for our security or operating costs. But I think it can be demonstrated that that is not the case: the UUID and the proposed salt have exactly the same exposure and accessibility (both entirely public); the UUID has equivalent or stronger requirements in terms of the amount of entropy and the randomness quality; the lifetimes are identical and neither has any need to allow change; and the failure domain when either field is generated incorrectly or compromised is not meaningfully different. To add a new field containing additional entropy beyond the amount which is useful just adds more load on the db with no benefit.

My preference for the bucket salts to be referred to as "bucket IDs" or to be treated as actual UUIDs involves a similar argument, but it doesn't carry the same cost of db load, so it doesn't matter as much.

kaloyan-raev: Using the project UUID as the project salt will make the implementation easier too, as no changes would be required on the satellite.

docs/design/password-key-derivation.md
ifraixedes: I've got the same question than @thepaul without reading this question before.
However, I'm still not understanding the response because I lack of background and probably I gonna ask a silly question, so my apologies in advance.

Does the satellite store the encryption keys for each segment?

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,77 @@
...
+6. Users do not have to enter a password for every bucket they create: they can have a default password that is used for every bucket unless otherwise specified.
+
+The third requirement allows having multiple encrypted domains to exist within a single bucket, allowing delegation of encryption. The fourth requirement allows renaming buckets without having to re-encrypt all of the paths inside of a bucket. The fifth requirement enhances security if a satellite breach happens. This is usually accomplished with [salting](https://en.wikipedia.org/wiki/Salt_(cryptography)). The sixth requirement allows providing a finite amount of information to a third party that allows them access to the keys for any bucket created with the default password in the future.
+
+## Design
+
+Satellites will add a `salt` field to bucket metadata that lives with the path encryption information that already exists and that we already request. Satellites will also store a `salt` that lives with the project metadata. The salts will be 32 random, stable bytes.
+
+A default password can be chosen for an api key. The following algorithm is used to derive the default password:
```
kaloyan-raev: What is exactly the "default password"? Is it the `ProjectEncryptionKey` from #1959? If yes, can we call it "project encryption key"?

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,77 @@
...
+
+```
+saltEntropy     = getProjectSalt(apiKey)
+mixedSalt       = hmac(hash=sha256, secret=userPassword, data=saltEntropy)
+defaultPassword = argon2id(salt=mixedSalt, password=userPassword)
+```
+
+The following algorithm is used to deriving a root key from an api key, password, bucket name and optional encrypted path:
```
kaloyan-raev: The term "root key" is quite overloaded in this document. Is the "root key", referred here, the key that we derive for a specific bucket? And then we derive keys for subpath from this key.

If my understanding is correct, can we name them "bucket encryption key" and "path encryption key"?

docs/design/password-key-derivation.md
kaloyan-raev: Yes, the pointerdb stores the encryption key and nonce used the encrypt the object's content and metadata. But this encryption key is encrypted with another encryption key that is derived from the user's password/passphrase/human-readable-key/whatever-we-call-it for the object's path. This way storing those encryption keys is secure in pointerdb.

ifraixedes: Gotcha, the encryption of the encryption key is what I was missing to understand that the satellite cannot read the user's data.

docs/design/password-key-derivation.md
zeebo: :+1:

docs/design/password-key-derivation.md
zeebo: Yeah, it would be stored in the `ProjectEncryptionKey`. I agree that the terminology should be consistent.

docs/design/password-key-derivation.md
zeebo: I'll add a section that explicitly defines "root key", but it may be the key used at the root of the bucket, or it may be a key used deeper in the bucket that would not match what would be the key from an earlier root.

For example, imagine one is making a dropbox style product and they want users in control of their encryption. There may be a `userdata` bucket with some root key, and it would contain paths like `sj://userdata/user1` and `sj://userdata/user2`. Those users could choose their own password and have a new root key at those paths, delegating encryption.

In other words, it's allowing the s<sub>0</sub> key (root key) to exist at a path other than the empty path.

docs/design/password-key-derivation.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Design
+
+Satellites will add a `salt` field to bucket metadata that lives with the path encryption information that already exists and that we already request. Satellites will also store a `salt` that lives with the project metadata. The salts will be 32 random, stable bytes.
+
+A default password can be chosen for an api key. The following algorithm is used to derive the default password:
+
+```
+saltEntropy     = getProjectSalt(apiKey)
+mixedSalt       = hmac(hash=sha256, secret=userPassword, data=saltEntropy)
```
jtolds: do we store the mixedSalt as a newly created bucket's salt?

docs/design/password-key-derivation.md
thepaul: We should probably never use the term "`Key`" to refer to a password/passphrase.

thepaul: We should probably never use the term "Key" to refer to a password/passphrase.

docs/design/password-key-derivation.md
jtolds: So, for buckets, either way we need to add something to bucket metadata - either a UUID or a salt. Buckets in particular though don't need a UUID, the bucket name within a project is sufficient, so adding a UUID feels unnecessary. We do need a salt for buckets, but we don't need a UUID. Right?

For projects, if we use the project UUID as the project salt, then we don't need to add project metadata, but a previously unencumbered data type now has important consequences cryptographically. Further, I'm not convinced yet about how project UUIDs will work with Satellite import/export. Basically, I'm not actually sure projects need UUIDs. I get that they're useful now, but I'm not sure that will always be the case. It might be, but I don't know.

Separating these concerns to me does feel like the right move. 

One additional caveat - we're really hoping this design doc can be *implemented* before the release this sprint, which is... Tuesday. So perhaps we can implement the bucket metadata first and punt on the project metadata anyways?

Let's do this - let's use bucket salts (we don't need bucket UUIDs). For project salts, let's use the project UUID for now but abstract it in the API call so we can change it to generic stable project entropy, should we ever change while designing import/export to use something to identify projects other than UUIDs. And then if we can do this in such a way that we can implement the basic things ASAP, that would be great!

docs/design/password-key-derivation.md
zeebo: Yeah, it's a bit hard for me to come up with what it is: we're taking a password, passing it through a key derivation function, then using it again in the password spot of a key derivation function.

Given that it's the output of a KDF, is it alright to call it a key, even if we use in in the password portion of a KDF later?

docs/design/password-key-derivation.md
zeebo: So, bucket salts and project UUIDs with the caveat that the design refers to it as a salt throughout? IOW, the project salt will be backed by the project UUID.

jtolds: that is my proposal

zeebo:
There are some unresolved questions, but I'm just doing what is the most recent suggestion on them. It's not a commitment, just optimistic pipelining.

docs/design/password-key-derivation.md
thepaul: > adding a UUID feels unnecessary

I'm only saying that we're already essentially adding a unique identifier, in the form of the bytes to use for salting. Why not make it official? Or is the concern that a UUID is more entropy than needed?

> Bucket names are by design unique within a project and unlikely to change, both within our pointer data model and by product requirements in regards to S3 similarity.

I think the motivating factor for a lot of the design here is being able to rename buckets. If we don't care about that, I would suggest that we use the content-key derivation algorithms already in place, and limit this doc to deriving "root" encryption keys from human-supplied passphrases.

I'd rather be able to rename buckets, though, if it's all the same.

> if we use the project UUID as the project salt, then it's true that we don't need to add project metadata, but a previously unencumbered data type now has important consequences cryptographically.

The _only_ quality with any cryptographic significance whatsoever here is uniqueness. As long as salts are unique, or even _mostly_ unique, the goal of foiling rainbow tables is accomplished. It could be dangerous to depend on the entropy from a salt in any respect beyond that.

docs/design/password-key-derivation.md
thepaul: I should clarify! I meant we should probably never use the term "Key" to refer to something that contains (or might contain) a human-supplied passphrase, which presumably has fairly low entropy. The output of a KDF can certainly be called a Key as far as I'm concerned.

docs/design/password-key-derivation.md
zeebo: > I think the motivating factor for a lot of the design here is being able to rename buckets. 

I disagree. The only part of the design that changes if we don't care about renaming buckets is we can remove the salt per bucket.

---

I can imagine a world where a user can request a new salt or even store their own salt with the satellite. I can also imagine a world where we regenerate UUIDs for things (import/export into another satellite). To me, a UUID communicates different intent and properties than salt does, and it seems dangerous to mix them, even if one can be used as the other. For example, why not use project creation time as the salt? It has the same properties as the UUID, but that'd be super confusing.

docs/design/password-key-derivation.md
zeebo: OK, perhaps the "default password" terminology in this design is what's misleading then, as it's the output of a KDF.

docs/design/password-key-derivation.md
```diff
@@ -8,35 +8,41 @@ This design doc describes the way cryptographic encryption keys will be derived
...
 6. Users do not have to enter a password for every bucket they create: they can have a default password that is used for every bucket unless otherwise specified.
 
 The third requirement allows having multiple encrypted domains to exist within a single bucket, allowing delegation of encryption. The fourth requirement allows renaming buckets without having to re-encrypt all of the paths inside of a bucket. The fifth requirement enhances security if a satellite breach happens. This is usually accomplished with [salting](https://en.wikipedia.org/wiki/Salt_(cryptography)). The sixth requirement allows providing a finite amount of information to a third party that allows them access to the keys for any bucket created with the default password in the future.
 
 ## Design
 
-Satellites will add a `salt` field to bucket metadata that lives with the path encryption information that already exists and that we already request. Satellites will also store a `salt` that lives with the project metadata. The salts will be 32 random, stable bytes.
+First, a **root key** is defined to be a key for some bucket and encrypted path.
```
thepaul: I really don't want to quibble any more about terminology but this seems like it might be very misleading. We're not saying that _any_ key for a (bucket, encryptedPath) pair is a "root key", are we? I thought we were using "root key" to mean a key that isn't derived using the typical `hmac(parentKey, pathComponent)` approach.

zeebo: This is a fine quibble. 👄👌 . I'm trying to say that for any (bucket, encryptedPath) pair, any number of root keys can be created, and you can derive keys from any root key.

I'm having a hard time coming up with clear, unambiguous, and concise phrasing.

thepaul: Maybe the equation on line 33 is actually the definition of a root key? I read it as just saying that it might not match the derived key, but if it's the definition then this all seems fine

docs/design/password-key-derivation.md
thepaul: * this discussion need not block merging of the PR.

> I disagree. The only part of the design that changes if we don't care about renaming buckets is we can remove the salt per bucket.

That's true. Project salts would still be worthwhile. You're right.

> I can imagine a world where a user can request a new salt or even store their own salt with the satellite.

My imagination is not up to the task here. Could you share a scenario so we can be on the same page about what we want to provide?

> I can also imagine a world where we regenerate UUIDs for things (import/export into another satellite).

I can easily believe that there's a good reason but I'm not seeing it. Why wouldn't we use the same UUID on the other satellite? Or under what conditions would we need to regenerate a UUID?

> To me, a UUID communicates different intent and properties than salt does, and it seems dangerous to mix them, even if one can be used as the other.

We may not even be on the same page about what UUIDs and salts are. To me they both signify the intent and properties of uniqueness, nothing more or less, but with possibly varying degrees of surety that they're _actually_ unique. If you're imputing any more qualities than that to a salt, that might be a recipe for trouble.

> For example, why not use project creation time as the salt? It has the same properties as the UUID (for the purposes of being a salt), but that'd be super confusing.

If project creation time was expected to be universally unique (across multiple satellites), then yes, it might well be worth using that value for both the salt and for project identification in place of a UUID. But I don't think that's a quality we have in our creation times, so I don't think it would make a good salt.

zeebo: > My imagination is not up to the task here. Could you share a scenario so we can be on the same page about what we want to provide?

This design goes through some effort to ensure satellites controlling salts can't be used to gain information about users passwords or keys. Maybe very paranoid users that don't trust my security understanding wish to use their own salts. Maybe we eventually don't trust my security understanding and enforce that clients store salts that they generate with the servers.

> I can easily believe that there's a good reason but I'm not seeing it. Why wouldn't we use the same UUID on the other satellite? Or under what conditions would we need to regenerate a UUID?

If the UUID collides (lol). Maybe a user wants to import the same project twice in their own satellite? Maybe a backup/shadow copy?

> We may not even be on the same page about what UUIDs and salts are. To me they both signify the intent and properties of uniqueness, nothing more or less, but with possibly varying degrees of surety that they're actually unique. If you're imputing any more qualities than that to a salt, that might be a recipe for trouble.

To me, a UUID is an identifier. Indeed, it contains enough entropy to be universally unique, but what it **is** is an identifier to something. Salts do not need to be universally unique, nor is it much of a failure if they aren't totally unique even in the same satellite. To me, they're just a source of entropy to make creating tables of weak keys impossible.

In other words, to me, a UUID conveys information that there's something that can be looked up if I have it, where a salt does not. So it's not a problem of a UUID being unfit, it's a problem of it providing too much.

> If project creation time was expected to be universally unique (across multiple satellites), then yes, it might well be worth using that value for both the salt and for project identification in place of a UUID. But I don't think that's a quality we have in our creation times, so I don't think it would make a good salt.

I agree creation time may not be unique, and so I think my point was lost due to our different perspectives. If I found a database that existed with a uniqueness constraint on creation time and found that the creation time was being used as a salt, I'd find that to be a massive subversion of my expectations. 

Sure, it may save a field, but why not just have it stored distinctly like how every other system does it? In my experience, salts are stored as part of a blob containing it and the password hash (and perhaps some data about the algorithm), but that's logically the same thing as it being a separate field. The only difference our system has is the password hashes are stored on the client and the salts are stored on the server. I don't think I've ever seen a system reuse a field for the purpose of being a salt.

thepaul: (I'm carrying on with this, even though the PR is approved, just with the hopes of getting to where I feel like we understand each other. Ignore with impunity if desired.)

> This design goes through some effort to ensure satellites controlling salts can't be used to gain information about users passwords or keys. Maybe very paranoid users that don't trust my security understanding wish to use their own salts. Maybe we eventually don't trust my security understanding and enforce that clients store salts that they generate with the servers.

So, the only thing afforded by salt in our construction is that it makes rainbow tables much less effective. There isn't anything there to trust. They don't even need to trust that their salt is unique, because they can inspect the salts of other projects/buckets. If they think we can subvert their security with a specially-crafted salt, then they don't trust SHA256 or Argon2 or AES128-GCM (either because they don't trust crypto in general, or because they know about some extremely alarming weaknesses in these that no one else knows about). I don't think we can satisfy such users no matter what we do.

> If the UUID collides (lol). Maybe a user wants to import the same project twice in their own satellite? Like as a backup/shadow copy?

That sounds like it would multiply our metadata costs without the user paying for any extra space, and that would be bad. If they do want to store everything again, then fantastic--and they'll want a new UUID for the backup project.

And if we experience unexpected UUID collisions before having something like 10^15 projects, that is an event that we want to cause sirens and flashing red lights.

> Sure, it may save a field, but why not just have it stored distinctly like how every other system does it? In my experience, salts are stored as part of a blob containing it and the password hash (and perhaps some data about the algorithm), but that's logically the same thing as it being a separate field. The only difference our system has is the password hashes are stored on the client and the salts are stored on the server. I don't think I've ever seen a system reuse a field for the purpose of being a salt.

Maybe not when the field is _called_ a "salt". But using some unique property of an item to modulate a hash or an encryption is so common that I'd be surprised to see any modern crypto-using system that _doesn't_ do it somewhere. Every time someone does an HMAC with a username or a userID or a URL or an email address, they're doing what I am proposing. Perhaps the disconnect comes about here because people only tend to call such a modulator a "salt" when they don't have any _other_ unique properties they can use and they have to store excess entropy for this sole purpose.

zeebo: I think we should continue this conversation and I'm enjoying it, but

1. I don't think I can prioritize the deep thinking right now
2. It'd probably go 10x faster in person.

thepaul: Well stated.

I’m satisfied that if we ever do need to separate the concerns of project uuid and salt, at least it can be made nicely backwards compatible.

PR #2024, opened 2019-05-22
Satellite and Bootstrap nodes are sent as storagenodes to be contacted for uploading pieces

aligeti:
Satellite  and Bootstrap nodes are sent as storagenodes to be contacted for uploading pieces
https://storjlabs.atlassian.net/browse/V3-1491

Why:
In our current implementation,
(1)Whenever we do a ping to a node after discovery in the Kad, we are immediately entering the discovered node to the nodesDB before fetching info(let's call vetting) about the node.
(2)searchGraveyard() is doing same as initial part of refresh(). They have same discoveryqueue () and they are writing to nodeDB parallelly, while DB can handle multiple requests but this contention can use up a bit of CPU and memory. 

What:
(1) On initial kad discovery (list of nodes seen),  first fetch the node's info.
(2) After fetching node info, then add to nodesDB, only if it is storage node. Invalid nodes, bootstrap and satellite nodes are not added.
(3) This ensures our NodesDB is clean and only has entries of storagenode type.
(4) removed the searchGraveyard(), as it is redundant

Thanks for submitting a PR!

### DIFF
--- a/internal/testplanet/planet.go
+++ b/internal/testplanet/planet.go
@@ -457,7 +457,6 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				},
 			},
 			Discovery: discovery.Config{
-				GraveyardInterval: 1 * time.Second,
 				DiscoveryInterval: 1 * time.Second,
 				RefreshInterval:   1 * time.Second,
 				RefreshLimit:      100,
--- a/pkg/datarepair/datarepair_test.go
+++ b/pkg/datarepair/datarepair_test.go
@@ -31,7 +31,6 @@ func TestDataRepair(t *testing.T) {
 		// stop discovery service so that we do not get a race condition when we delete nodes from overlay cache
 		satellite.Discovery.Service.Discovery.Stop()
 		satellite.Discovery.Service.Refresh.Stop()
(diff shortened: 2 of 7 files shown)


littleskunk:
The reason for https://storjlabs.atlassian.net/browse/V3-1491 was a bad node selection. The node selection is fixed. What do we try to fix with this PR?

aligeti:
It’s mentioned in the PRs why & what. There is a possibility we potentially add bogus nodes to nodesDB.

littleskunk:
> Satellite and Bootstrap nodes are sent as storagenodes to be contacted for uploading pieces

That is already fixed with v0.11.3. We are selecting only storage nodes for uploads. I don't understand why we need this PR if everything is working as expected? What is the problem if we store bootstrap and satellite nodes in the database but don't select them for uploads. Which issues do we have with that?

aligeti:
The changes in this PR are good for the above PR description reasons, as it eliminates all together of adding satellites and bootstraps address and also eliminates a redundant cyclic loop which can potentially save CPU and memory. As the problem was addressed by Jens by modifying query, there changes can be picked up at a later date if necessary. This PR will be closed as the problem is addressed.

PR #2025, opened 2019-05-22
Save hash of expected share from missing audited node

kaloyan-raev:
What: 

Implements https://storjlabs.atlassian.net/browse/V3-1765

Why:

Thanks for submitting a PR!


pkg/audit/verifier.go
```diff
@@ -71,12 +72,17 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe) (verifiedN
 
 	var offlineNodes storj.NodeIDList
 	var failedNodes storj.NodeIDList
+	var containedNodes storj.NodeIDList
+	var containedPieceNums []int
```
navillasa: Having an array of ints just makes me a little wary since we had some hard-to-find pieceNum/index bugs in the repair package. Can we have a `map[int]storj.NodeID` instead?

pkg/audit/containment.go
```diff
@@ -31,10 +31,10 @@ var (
 type PendingAudit struct {
 	NodeID            storj.NodeID
 	PieceID           storj.PieceID
-	StripeIndex       uint32
-	ShareSize         int64
+	StripeIndex       int64
```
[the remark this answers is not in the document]
kaloyan-raev: We already have the `Index` field in `Stripe` type as `int64`: https://github.com/storj/storj/blob/master/pkg/audit/cursor.go#L26

So it should be the same here for consistency.

Theoretically, the index can go that high if the segment is TBs large. This doesn't sound realistic today. But, who knows, maybe in 10 years there might be a valid use case. I am not sure if this was the reason to set it to `int64` in `Stripe.Index`.

pkg/audit/reporter.go
```diff
@@ -135,3 +145,17 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
...
+	for _, pendingAudit := range pendingAudits {
+		err = reporter.containment.IncrementPending(ctx, pendingAudit)
+		if err != nil {
+			failed = append(failed, pendingAudit)
+		}
+	}
+	if len(failed) > 0 {
+		return failed, Error.New("failed to record some pending audits")
+	}
+	return nil, nil
```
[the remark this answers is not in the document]
kaloyan-raev: The execution should reach this line if all pending audits were reported successfully. So there is not failed report, and no error. Thus, both return values are `nil`.

pkg/audit/reporter.go
```diff
@@ -135,3 +145,17 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
 	}
 	return nil, nil
 }
+
+// recordPendingAudits updates the containment status of nodes with pending audits
+func (reporter *Reporter) recordPendingAudits(ctx context.Context, pendingAudits []*PendingAudit) (failed []*PendingAudit, err error) {
+	for _, pendingAudit := range pendingAudits {
+		err = reporter.containment.IncrementPending(ctx, pendingAudit)
+		if err != nil {
```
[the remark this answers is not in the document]
kaloyan-raev: Good point! I just followed the pattern of similar methods in the Reporter. But what you say totally make sense. I will modify all of them.

pkg/audit/reporter.go
```diff
@@ -66,6 +69,12 @@ func (reporter *Reporter) RecordAudits(ctx context.Context, req *RecordAuditsInf
 				errNodeIDs = append(errNodeIDs, offlineNodeIDs...)
 			}
 		}
+		if len(pendingAudits) > 0 {
+			pendingAudits, err = reporter.recordPendingAudits(ctx, pendingAudits)
+			if err != nil {
+				errNodeIDs = append(errNodeIDs, offlineNodeIDs...)
```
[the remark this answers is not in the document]
kaloyan-raev: Absolutely! Good catch!

pkg/audit/verifier.go
kaloyan-raev: You are absolutely right. Changed it.

pkg/audit/reporter.go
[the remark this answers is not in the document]
kaloyan-raev: Done.

pkg/audit/service.go
```diff
@@ -92,15 +92,15 @@ func (service *Service) process(ctx context.Context) error {
 		}
 	}
 
-	verifiedNodes, err := service.Verifier.Verify(ctx, stripe)
-	if err != nil {
-		return err
+	verifiedNodes, verifierErr := service.Verifier.Verify(ctx, stripe)
+	if verifierErr != nil && verifiedNodes == nil {
+		return verifierErr
```
navillasa: why are we being more specific with the errors here?

pkg/audit/timeout_test.go
```diff
@@ -77,8 +78,12 @@ func TestGetShareTimeout(t *testing.T) {
 			require.NoError(t, err)
 		}
 
-		_, err = verifier.Verify(ctx, stripe)
-		require.NoError(t, err)
+		verifiedNodes, err := verifier.Verify(ctx, stripe)
+		assert.Error(t, err)
```
navillasa: requires over asserts now right?

pkg/audit/verifier.go
```diff
@@ -71,12 +72,14 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe) (verifiedN
...
 	var failedNodes storj.NodeIDList
+	containedNodes := make(map[int]storj.NodeID)
 	sharesToAudit := make(map[int]Share)
 
 	for pieceNum, share := range shares {
 		if shares[pieceNum].Error != nil {
+			// TODO(kaloyan): we need to check the logic here if we correctly identify offline nodes from those that didn't respond.
 			if shares[pieceNum].Error == context.DeadlineExceeded || !transport.Error.Has(shares[pieceNum].Error) {
-				failedNodes = append(failedNodes, nodes[pieceNum])
+				containedNodes[pieceNum] = nodes[pieceNum]
```
navillasa: I understand why this addition is here, but maybe we should wait until we add more tests and implement the way for pending audits to be removed from the containmentDB before putting them in? it might not be a big deal if we add them now though.

pkg/audit/service.go
mobyvb: On a similar note, do we need to worry about the error if `verifiedNodes != nil`?

kaloyan-raev: Here the idea is that Verify can return some verified notes even if an error occurs. We want to record the audit outcome for these verified nodes. This is why on this line we return `verifyErr` only if there are no successfully verified nodes. Otherwise, we record them and combine `verifierErr` with `reporterErr`... shoot, I have bug on line 103...

kaloyan-raev: Line 103 is fixed now.

pkg/audit/timeout_test.go
kaloyan-raev: No. Here using assert over requires is intentional. I want to check both `err` and `verifiedNodes`. If the test fails here, this will give us more clue what went wrong.

pkg/audit/verifier.go
kaloyan-raev: Even if we deploy containment mode semi-ready like in this PR, it still won't be worse than the current situation for storage nodes. Here we put nodes in containment mode rather than failing the audit for them. So we won't hurt unnecessarily their reputations

PR #2031, opened 2019-05-22
create db implementation for offer table

VinozzZ:
What: 
create methods to interact with offer table for marketing service
Why:
**Default Free Tier Offer:**
  - This is the credit that all new users who join the network receive if they create an account without a Referral Link. Once a new user successfully creates their account they will have 14 days to use their Free Credit allowance.  
  - The Default Offer will remain constant, with no expiration date to the offer and does not have a redemption cap
  - The Offer Program Admin is able to adjust the default offer
  - One Default Offer per Tardigrade Satellite.
 -  When there is a lapse of the Referral Testing Offer/the Referral Testing Offer has expired or reached a cap, the Default Offer will be the default to the Referral Testing Offer that new users will redeem until a new Referral Testing Offer is generated.

**Referral Testing Offer**:
  - This is the credit awarded to the Invitee upon creating their account using a Referral Link. This credit can have a different expiration date than the awarded Referrer Credit and the Default Free Credit.
  - An admin of the Referral Program is able to easily adjust the current offer through the  Admin Satellite GUI

Thanks for submitting a PR!


satellite/marketing/service.go
```diff
@@ -0,0 +1,90 @@
...
+// ListAllOffers returns all available offers in the db
+func (s *Service) ListAllOffers(ctx context.Context) (offers []Offer, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	offers, err = s.db.Offers().ListAll(ctx)
+	if err != nil {
+		return offers, Error.Wrap(err)
+	}
+
+	return
```
navillasa: nit pick but I think this should name the return values

satellite/marketing/service.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return offer, nil
+}
+
+// InsertNewOffer inserts a new offer into the db
+func (s *Service) InsertNewOffer(ctx context.Context, offer *NewOffer) (o *Offer, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if offer.Status == Default {
+		offer.ExpiresAt = time.Now().AddDate(100, 0, 0)
```
navillasa: I think we're trying to standardize by using UTC, so maybe better to have `time.Now().UTC().AddDate..`

satellite/satellitedb/migrate.go
```diff
@@ -719,6 +719,18 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					`ALTER TABLE projects ADD usage_limit bigint NOT NULL DEFAULT 0;`,
 				},
 			},
+			{
+				Description: "Add invitee_credit_in_gb and award_credit_in_gb columns, delete type and credit_in_cents columns",
+				Version:     25,
+				Action: migrate.SQL{
+					`DELETE FROM offers WHERE id = 1;`,
```
navillasa: I feel like there's a better way to delete the data that was there before, but I'm not sure what that would be..

satellite/satellitedb/offers.go
```diff
@@ -0,0 +1,157 @@
...
+
+	_, err := offers.db.Update_Offer_By_Id(ctx, offerID, updateFields)
+	if err != nil {
+		return marketing.OffersErr.Wrap(err)
+	}
+
+	return nil
+}
+
+func offersFromDbx(offersDbx []*dbx.Offer) ([]marketing.Offer, error) {
```
navillasa: super nit, I think acronyms are all uppercase so it should be offersFromDBX

satellite/satellitedb/offers.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return offersFromDbx(offersDbx)
+}
+
+// GetCurrent returns an offer that has not expired based on offer status
+func (offers *offers) GetCurrent(ctx context.Context, offerStatus marketing.OfferStatus) (*marketing.Offer, error) {
+	var statement string
+	const columns = "id, name, description, award_credit_in_cents, invitee_credit_in_cents, award_credit_duration_days, invitee_credit_duration_days, redeemable_cap, num_redeemed, expires_at, created_at, status"
+	statement = `
+		WITH o AS (
```
navillasa: another stylistic nit, I think `o` should just be `offer`

satellite/satellitedb/offers.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return marketing.OffersErr.Wrap(err)
+	}
+
+	return nil
+}
+
+func offersFromDbx(offersDbx []*dbx.Offer) ([]marketing.Offer, error) {
+	var offers []marketing.Offer
+	var errors []error
```
navillasa: this is fine, but an optional thing that's also used in the repo is `var errList errs.Group`, where you can add errors like `errList.Add(err)`, then return all with `errList.Err()`. I'm not sure of its differences but i've just seen errs.Group more often.

satellite/satellitedb/offers.go
VinozzZ: `o` is temporary variable not a real table here. I feel it would be better to use a name that doesn't look like a table name here

satellite/marketing/offers.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// OfferType indicates the type of an offer
+type OfferType int
+
+const (
+	// FreeCredit is a type of offers used for Free Credit Program
+	FreeCredit OfferType = 0
+	// Referral is a type of offers used for Referral Program
+	Referral OfferType = 1
```
mniewrzal: You can use `iota` for enumeration, example https://github.com/storj/storj/blob/69d8b9f8286af2c7a4f32b57c036d5d84888e8e3/pkg/storj/encryption.go#L68

satellite/marketing/offers_test.go
```diff
@@ -0,0 +1,104 @@
...
+				AwardCreditDurationDays:   60,
+				InviteeCreditDurationDays: 30,
+				RedeemableCap:             50,
+				ExpiresAt:                 time.Now().UTC().Add(time.Hour * 1),
+				Status:                    marketing.Default,
+				Type:                      marketing.FreeCredit,
+			},
+		}
+
+		for i := range validOffers {
```
mniewrzal: nit: this can be also `for _, validOffer := range validOffers {`

satellite/marketing/offers_test.go
```diff
@@ -0,0 +1,104 @@
...
+				InviteeCreditDurationDays: 30,
+				RedeemableCap:             50,
+				ExpiresAt:                 time.Now().UTC().Add(time.Hour * -1),
+				Status:                    marketing.Default,
+				Type:                      marketing.Referral,
+			},
+		}
+
+		for i := range expiredOffers {
+			_, err := planet.Satellites[0].DB.Marketing().Offers().Create(ctx, &expiredOffers[i])
```
mniewrzal: you can check that returned value is `nil` in case of error

satellite/marketing/service.go
```diff
@@ -0,0 +1,94 @@
...
+	o, err = s.db.Offers().Create(ctx, offer)
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	return o, nil
+}
+
+// UpdateOffer modifies an existing offer in the db when the offer status is set to NoStatus
+func (s *Service) UpdateOffer(ctx context.Context, o *UpdateOffer) (err error) {
```
mniewrzal: nit: you can use `offer` instead `o` to be consistent with the previous method.

satellite/satellitedb/offers.go
```diff
@@ -0,0 +1,163 @@
...
+	tx, err := offers.db.Open(ctx)
+	if err != nil {
+		return nil, marketing.OffersErr.Wrap(err)
+	}
+
+	statement := offers.db.Rebind(`
+		UPDATE offers SET status=?, expires_at=?
+		WHERE status=? AND type=? AND expires_at>?;
+	`)
+	_, err = tx.Tx.ExecContext(ctx, statement, marketing.Done, currentTime, o.Status, o.Type, currentTime)
```
mniewrzal: I'm not sure I understand what's happening with this sql statement. I know its updating but can you add a comment why?

satellite/marketing/offers_test.go
VinozzZ: I was thinking that since `validOffers` is a reference type, it's better to not create copies of the pointer and just use the original pointer value in the slice

mniewrzal: sounds good :+1:

satellite/satellitedb/offers.go
VinozzZ: I added a comment for the statment

satellite/marketing/database.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package marketing
```
stefanbenten: I am not sure this is a good package name, especially with its location being in the satellite.
But i dont have a good idea how to do it better either. Just sounds wrong to me :)

satellite/peer.go
```diff
@@ -90,6 +91,8 @@ type DB interface {
 	Irreparable() irreparable.DB
 	// Console returns database for satellite console
 	Console() console.DB
+	// Marketing returns database for marketing admin GUI
+	Marketing() marketing.DB
```
stefanbenten: Same as above, not sure thats a good name for it.

satellite/satellitedb/offers.go
stefanbenten: Typo in the comment

satellite/marketing/database.go
VinozzZ: I think once we add more functionality to it, it will become more clear how we should name this package. Right now, the purpose of this package is so that marketing team can change configuration on the satellite for marketing purpose

stefanbenten: What configuration would that be? the amount of free credits?

VinozzZ: We will have a Open Source Partner Program and it will have configuration that need to be exposed to the marketing team to be able to change it. Not sure what exact configuration we will have at the moment

satellite/marketing/offers.go
```diff
@@ -0,0 +1,93 @@
...
+
+// OffersErr creates offer error class
+var OffersErr = errs.Class("offers error")
+
+// Offers holds information about offer
+type Offers interface {
+	ListAll(ctx context.Context) ([]Offer, error)
+	GetCurrentByType(ctx context.Context, offerType OfferType) (*Offer, error)
+	Create(ctx context.Context, offer *NewOffer) (*Offer, error)
+	Update(ctx context.Context, offer *UpdateOffer) error
```
egonelbre: A single update somehow feels problematic here. You could get concurrent updates and probably would want to separate these... i.e. maybe having methods `Redeem`, `Renew` and `Finish` (or similar)?

VinozzZ: Wouldn't the lock file prevent that from happening?

egonelbre: What do you mean by a lock file? If you mean a transactions, we need to reduce the number of transactions since it slows down the database.

VinozzZ:
```golang
func (m *lockedOffers) Update(ctx context.Context, offer *marketing.UpdateOffer) error {
	m.Lock()
	defer m.Unlock()
	return m.db.Update(ctx, offer)
}
```
I thought this function in the `locked.go` would prevent concurrent update to the same data

PR #2036, opened 2019-05-23
Per-project usage limiting

mniewrzal:
What: Changes to support custom usage limit for the project. With this implementation by default project usage limit is taken from configuration flag. If project DB field `usage_limit` will be set to value larger than 0 it will become custom usage limit and we will be used to verify is limit was exceeded. 

Whats changed:
* `usage_limit` (bigint) field added to `projects` table (with migration)
* things related to project usage moved from metainfo endpoint to project usage type
* `accounting.ProjectAccounting` extended with `GetProjectUsageLimits()` method

Why: We need to have different usage limits per project. https://storjlabs.atlassian.net/browse/V3-1814


pkg/accounting/projectusage.go
```diff
@@ -15,20 +23,119 @@ const (
...
+func (usage *ProjectUsage) ExceedsBandwidthUsage(ctx context.Context, projectID uuid.UUID, bucketID []byte) (_ bool, limit memory.Size, err error) {
+	// Check if this projectID has exceeded alpha usage limits for bandwidth or storage used in the past month
+	// TODO: remove this code once we no longer need usage limiting for alpha release
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
+
+	var group errgroup.Group
+	var bandwidthGetTotal int64
+	limit = usage.maxAlphaUsage
+
+	// TODO maybe project limit cache expiration time?
```
JessicaGreben: what does "project limit cache expiration time" mean?

pkg/accounting/projectusage.go
```diff
@@ -15,20 +23,119 @@ const (
...
+// for a project in the past month (30 days). The usage limit is (e.g 25GB) multiplied by the redundancy
+// expansion factor, so that the uplinks have a raw limit.
 // TODO(jg): remove this code once we no longer need usage limiting for alpha release
 // Ref: https://storjlabs.atlassian.net/browse/V3-1274
-func ExceedsAlphaUsage(bandwidthGetTotal, inlineTotal, remoteTotal int64, maxAlphaUsageGB memory.Size) (bool, string) {
-	maxUsage := maxAlphaUsageGB.Int64() * int64(ExpansionFactor)
+func (usage *ProjectUsage) ExceedsBandwidthUsage(ctx context.Context, projectID uuid.UUID, bucketID []byte) (_ bool, limit memory.Size, err error) {
+	// Check if this projectID has exceeded alpha usage limits for bandwidth or storage used in the past month
+	// TODO: remove this code once we no longer need usage limiting for alpha release
+	// Ref: https://storjlabs.atlassian.net/browse/V3-1274
```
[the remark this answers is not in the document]
mniewrzal: sure, I forgot to compare them

pkg/accounting/projectusage.go
```diff
@@ -15,20 +23,119 @@ const (
...
+
+	var group errgroup.Group
+	var bandwidthGetTotal int64
+	limit = usage.maxAlphaUsage
+
+	// TODO maybe project limit cache expiration time?
+	group.Go(func() error {
+		// TODO should we ignore err and use default limit?
```
JessicaGreben: i think keeping the err checking is my preference

satellite/metainfo/metainfo.go
```diff
@@ -54,19 +53,14 @@ type Endpoint struct {
...
-	maxAlphaUsage           memory.Size
 }
 
 // NewEndpoint creates new metainfo endpoint instance
-func NewEndpoint(log *zap.Logger, metainfo *Service, orders *orders.Service, cache *overlay.Cache,
-	apiKeys APIKeys, sdb accounting.StoragenodeAccounting,
-	pdb accounting.ProjectAccounting, liveAccounting live.Service,
-	maxAlphaUsage memory.Size) *Endpoint {
-
+func NewEndpoint(log *zap.Logger, metainfo *Service, orders *orders.Service, cache *overlay.Cache, apiKeys APIKeys, sdb accounting.StoragenodeAccounting, projectUsage *accounting.ProjectUsage, liveAccounting live.Service) *Endpoint {
```
JessicaGreben: yay removing arguments here is so nice!

pkg/accounting/projectusage.go
mniewrzal: its just an idea if we will need less DB operations. We can have cache where entry is invalidated after period of time (e.g. 1h). ofc maybe different type will fit better

satellite/peer.go
```diff
@@ -153,6 +153,9 @@ type Peer struct {
 		Database  storage.KeyValueStore // TODO: move into pointerDB
 		Service   *metainfo.Service
 		Endpoint2 *metainfo.Endpoint
+
+		// TODO(michal) most probably not best place
+		ProjectUsage *accounting.ProjectUsage
```
JessicaGreben: why not put it in Accounting?

pkg/accounting/projectusage.go
JessicaGreben: ah ok that makes sense. could we change the comment to be more clear? Something like `TODO(mn): to reduce db load, consider using a cache to retrieve the `project.UsageLimit` values`

mniewrzal: Sure, good point

pkg/accounting/projectusage.go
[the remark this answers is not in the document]
thepaul: JT has assured me that this limiting is going to stick around, and that it's not just for Alpha. Brandon said the opposite thing earlier, so I left the comments in, but now I'm pretty sure everybody's on the same page about keeping it. So feel free to take out the "TODO: remove this code" comments entirely.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -216,6 +216,7 @@ model project (
 
     field name           text
     field description    text      ( updatable )
+	field usage_limit	 int64     ( updatable )
```
thepaul: looks like a whitespace mismatch here. is this using spaces and the other lines are using tabs, or is it the other way around?

satellite/metainfo/metainfo.go
```diff
@@ -75,9 +69,8 @@ func NewEndpoint(log *zap.Logger, metainfo *Service, orders *orders.Service, cac
 		cache:                   cache,
 		apiKeys:                 apiKeys,
 		storagenodeAccountingDB: sdb,
-		projectAccountingDB:     pdb,
+		projectUsage:            projectUsage,
```
thepaul: Does Endpoint still need to have the liveAccounting member? `projectUsage` now owns the liveAccounting service, it looks like.

satellite/peer.go
JessicaGreben: what do you mean by this comment? that the `accounting.ProjectUsage` code should be moved into the metainfo package?

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: i think the indentation is off here

satellite/peer.go
mniewrzal: I had a dilemma how to organize this only in `peer.go` but `Accounting` should be ok. Comment removed.

satellite/metainfo/metainfo.go
mniewrzal: good point, `liveAccounting` removed from metainfo

pkg/accounting/projectusage.go
mniewrzal: done

pkg/accounting/projectusage.go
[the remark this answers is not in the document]
mniewrzal: thanks, removed duplicated comments and info about removing after alpha

satellite/satellitedb/dbx/satellitedb.dbx
mniewrzal: fixed

pkg/accounting/projectusage_test.go
```diff
@@ -262,3 +253,43 @@ func setUpBucketBandwidthAllocations(ctx *testcontext.Context, projectID uuid.UU
...
+		// set custom usage limit for project
+		project.UsageLimit = memory.GiB.Int64() * 10
+		err = projectsDB.Update(ctx, &project)
+		require.NoError(t, err)
+
+		projectUsage := planet.Satellites[0].Accounting.ProjectUsage
+
+		// Setup: create BucketStorageTally records to test exceeding storage project limit
+		now := time.Now()
+		err = setUpStorageTallies(ctx, project.ID, acctDB, now)
```
JessicaGreben: this function `setUpStorageTallies` creates 30GB worth of `BucketStorageTally` records to test the default 25GB project limit.  For this test, I think it would be more meaning to only create 11GB of `BucketStorageTally` records to show that the custom limit is "working"

satellite/metainfo/metainfo.go
```diff
@@ -179,19 +172,13 @@ func (endpoint *Endpoint) CreateSegment(ctx context.Context, req *pb.SegmentWrit
...
 	}
-	exceeded, resource := accounting.ExceedsAlphaUsage(0, inlineTotal, remoteTotal, endpoint.maxAlphaUsage)
 	if exceeded {
-		endpoint.log.Sugar().Errorf("monthly project limits are %s of storage and bandwidth usage. This limit has been exceeded for %s for projectID %s.",
-			endpoint.maxAlphaUsage.String(),
-			resource, keyInfo.ProjectID,
+		endpoint.log.Sugar().Errorf("monthly project limits are %s of storage and bandwidth usage. This limit has been exceeded for storage for projectID %s",
+			limit, keyInfo.ProjectID,
 		)
 		return nil, status.Errorf(codes.ResourceExhausted, "Exceeded Alpha Usage Limit")
```
JessicaGreben: can we update this comment "Exceeded Alpha Usage Limit" to remove the "alpha" part

satellite/metainfo/metainfo.go
```diff
@@ -337,19 +312,15 @@ func (endpoint *Endpoint) DownloadSegment(ctx context.Context, req *pb.SegmentDo
...
+		endpoint.log.Error("retrieving project bandwidth total", zap.Error(err))
 	}
-	exceeded, resource := accounting.ExceedsAlphaUsage(bandwidthTotal, 0, 0, endpoint.maxAlphaUsage)
 	if exceeded {
-		endpoint.log.Sugar().Errorf("monthly project usage limit has been exceeded for resource: %s, for project: %d. Contact customer support to increase the limit.",
-			resource, keyInfo.ProjectID,
+		endpoint.log.Sugar().Errorf("monthly project limits are %s of storage and bandwidth usage. This limit has been exceeded for bandwidth for projectID %s.",
+			limit, keyInfo.ProjectID,
 		)
 		return nil, status.Errorf(codes.ResourceExhausted, "Exceeded Alpha Usage Limit")
```
JessicaGreben: remove "alpha" from err, will need to update the tests as well since they have this error message hard coded

PR #2039, opened 2019-05-23
uplink: Don't save EncryptionKey in config

ifraixedes:
What: 
The encryption key configuration field MUST NOT be present in the
configuration file.

Why:
Without the `setup:"true"` the field is set in the `config.yaml`
commented out, however this value MUST NEVER be there because it's
insecure and moreover is never read.

Thanks for submitting a PR!

### DIFF
--- a/uplink/config.go
+++ b/uplink/config.go
@@ -42,7 +42,7 @@ type RSConfig struct {
 // EncryptionConfig is a configuration struct that keeps details about
 // encrypting segments
 type EncryptionConfig struct {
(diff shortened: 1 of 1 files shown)


ifraixedes:
It will require more work, the flag is needed by the gateway to run the integration test, so we'll have to do one of the following things:

1. Remove that field/flag and setup the integration test for reading the encryption key from the encryption key file
2. Find another way to skip of setting the field into the `config.yaml` file.

I'll create an issue for this for being scheduled when and who will work on it.

PR #2041, opened 2019-05-23
add Reverify to auditing

navillasa:
What:
- Adds Reverify function to audit/verifier.go

Why:
- We need to reverify contained nodes with pending audits.

Thanks for submitting a PR!


pkg/audit/reporter.go
```diff
@@ -106,6 +113,13 @@ func (reporter *Reporter) recordAuditFailStatus(ctx context.Context, failedAudit
 			failed = append(failed, nodeID)
 			errlist.Add(err)
 		}
+
+		// TODO(kaloyan): Perhaps, this should be executed in the same Tx as overlay.UpdateStats above
+		_, err = reporter.containment.Delete(ctx, nodeID)
```
mobyvb: I don't think we should delete from containment here. A node might fail a different audit than the one it was contained for, but we should still audit it for the share it was contained for (unless I'm confused)

pkg/audit/reporter.go
```diff
@@ -142,6 +156,13 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
...
```
mobyvb: same here. Just because the node passes an audit doesn't mean it should be removed from containment mode. Maybe it still fails to audit the share it was contained for. Again, not sure if my thinking is correct here

pkg/audit/containment_test.go
```diff
@@ -6,16 +6,212 @@ package audit_test
...
+				SerialNumber:    storj.SerialNumber{},
+				SatelliteId:     planet.Satellites[0].ID(),
+				UplinkId:        ul.ID(),
+				StorageNodeId:   piece.NodeId,
+				PieceId:         rootPieceID.Derive(piece.NodeId),
+				Action:          pb.PieceAction_GET_AUDIT,
+				Limit:           int64(redundancy.ErasureShareSize),
+				PieceExpiration: &timestamp.Timestamp{Seconds: time.Now().Unix() + 3000},
+				OrderExpiration: &timestamp.Timestamp{Seconds: time.Now().Unix() + 3000},
+			})
```
kaloyan-raev: I think we can simplify some code here by requesting the order limit from the Orders Service instead of replicating the same code here.

kaloyan-raev: Did it with https://github.com/storj/storj/pull/2041/commits/bcc49ba344df1a822e33ddf70c2c9eda27fad2ce

PR #2043, opened 2019-05-23
Satellite voucher service

cam-a:
https://storjlabs.atlassian.net/browse/V3-1726

Set up a service on the satellite to respond to requests from storage nodes for signed vouchers


pkg/auth/signing/encode.go
```diff
@@ -32,3 +32,11 @@ func EncodePieceHash(hash *pb.PieceHash) ([]byte, error) {
 	defer func() { hash.Signature = signature }()
 	return proto.Marshal(hash)
 }
+
+// EncodeVoucher encodes voucher into bytes for signing.
+func EncodeVoucher(voucher *pb.Voucher) ([]byte, error) {
```
aligeti: You are just serializing the bytes. Not quite encoding. renaming as "SerializeVoucher" or something in those line may be more appropriate

pkg/auth/signing/sign.go
```diff
@@ -70,3 +70,20 @@ func SignPieceHash(signer Signer, unsigned *pb.PieceHash) (*pb.PieceHash, error)
 
 	return &signed, nil
 }
+
+// SignVoucher signs the voucher using the specified signer
+// Signer is a satellite
+func SignVoucher(signer Signer, unsigned *pb.Voucher) (*pb.Voucher, error) {
```
aligeti: when I first saw it, I thought it has a datatype keyword. Might create confusion. If you can rename it to something like unsigned_vocher may be appropriate.

pkg/auth/signing/encode.go
```diff
@@ -32,3 +32,11 @@ func EncodePieceHash(hash *pb.PieceHash) ([]byte, error) {
...
+	signature := voucher.SatelliteSignature
+	voucher.SatelliteSignature = nil
+	defer func() { voucher.SatelliteSignature = signature }()
+	return proto.Marshal(voucher)
+}
```
jenlij: Let's update this method to match these https://github.com/storj/storj/pull/2084

pkg/overlay/cache.go
```diff
@@ -48,7 +48,8 @@ type DB interface {
 	KnownUnreliableOrOffline(context.Context, *NodeCriteria, storj.NodeIDList) (storj.NodeIDList, error)
 	// Paginate will page through the database nodes
 	Paginate(ctx context.Context, offset int64, limit int) ([]*NodeDossier, bool, error)
-
+	// VetNode returns whether or not the node reaches reputable thresholds
+	VetNode(ctx context.Context, id storj.NodeID, criteria *NodeCriteria) (bool, error)
```
egonelbre: This would probably benefit from a better name. It's only querying and not actually updating it, hence the active verb seems out of place. Maybe `IsVetted` (no need to specify "Node" because all of the nodes are about nodes)?

PR #2045, opened 2019-05-23
add repair monkit stats

mobyvb:
What: 
`repair_attempts` - a tracker for the number of repair attempts
`repair_segment_size` - the size of the segment being repaired
`repair_nodes_unavailable` - a tracker for the number of repair attempts that are canceled because not enough nodes are online (potential irreparable segment)
`repair_unnecessary` - a tracker for the number of repair attempts that are canceled because the number of nodes online is over the repair threshold
`repair_success` - a tracker for the number of successful repairs (healthy pieces >= success threshold at end of repair)
`repair_partial` - a tracker for the number of repairs that did not surpass the success threshold, but that surpassed the repair threshold
`repair_failed` - a tracker for the number of repairs that did not surpass the repair threshold
`healthy_ratio_before_repair` - `numHealthyPieces/numTotalPieces` before repair starts
`healthy_ratio_after_repair` - `numHealthyPieces/numTotalPieces` after repair ends

Why:

Thanks for submitting a PR!


mobyvb:
@jtolds done

pkg/storage/segments/repairer.go
```diff
@@ -74,14 +77,19 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
...
 		return Error.New("piece %v cannot be repaired", path)
 	}
 
 	// repair not needed
 	if int32(numHealthy) > pointer.Remote.Redundancy.RepairThreshold {
+		mon.Meter("repair_unnecessary").Mark(1)
 		return Error.New("piece %v with %d pieces above repair threshold %d", path, numHealthy, pointer.Remote.Redundancy.RepairThreshold)
 	}
 
+	healthyRatioBeforeRepair := float64(numHealthy) / float64(pointer.Remote.Redundancy.Total)
```
stefanbenten: Both of these value shouldnt be a float ever right?
I feel there is a better solution for that..

pkg/storage/segments/repairer.go
```diff
@@ -154,6 +162,17 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
...
 
+	if int32(len(healthyPieces)) <= pointer.Remote.Redundancy.RepairThreshold {
+		mon.Meter("repair_failed").Mark(1)
+	} else if int32(len(healthyPieces)) < pointer.Remote.Redundancy.SuccessThreshold {
+		mon.Meter("repair_partial").Mark(1)
+	} else {
+		mon.Meter("repair_success").Mark(1)
+	}
+
+	healthyRatioAfterRepair := float64(len(healthyPieces)) / float64(pointer.Remote.Redundancy.Total)
```
stefanbenten: Same as above

pkg/storage/segments/repairer.go
mobyvb: I was just casting them so that we don't do integer division

pkg/storage/segments/repairer.go
mobyvb: Same - don't want integer division here

jtolds:
I'd really like to get this in today if possible so I'm okay with not fixing it in this pr but this will panic if total is ever zero. that should never happen anyways for lots of other reasons but yeah

PR #2046, opened 2019-05-23
Change for free credit design doc

VinozzZ:
What: 
simplify database schema
add invitee_credit_in_cents column
add methods implementation for offer table
remove Delete and Update methods in offer interface
Why:
Due to changes in the user flow for the marketing offer admin gui, we need to make some changes for the offer table and service

Thanks for submitting a PR!

### DIFF
--- a/docs/design/marketing-offer-program.md
+++ b/docs/design/marketing-offer-program.md
@@ -73,7 +73,7 @@ The credit is automatically applied to the account and will have a max limit tha
     num_redeemed - integer
     created_at - timestamp
     expires_at - timestamp
-    award_credit_duration_months - int
+    award_credit_duration_days - int
     invitee_credit_duration_days - int
     status - enum[ACTIVE, DEFAULT, DONE]
     PRIMARY KEY (id)
@@ -93,6 +93,7 @@ The credit is automatically applied to the account and will have a max limit tha
     FOREIGN KEY (offer_id)
     FOREIGN KEY (referred_by)
     FOREIGN KEY (user_id)
+    PRIMARY KEY (id)
 ```
 
 **user table**
--- a/docs/design/marketing-offer-program.md
+++ b/docs/design/marketing-offer-program.md
@@ -75,7 +75,8 @@ The credit is automatically applied to the account and will have a max limit tha
     expires_at - timestamp
     award_credit_duration_days - int
     invitee_credit_duration_days - int
-    status - enum[ACTIVE, DEFAULT, DONE]
+    // ACTIVE=1, DEFAULT=2, DONE=0
+    status - int
     PRIMARY KEY (id)
 ```
 
@@ -86,7 +87,8 @@ The credit is automatically applied to the account and will have a max limit tha
     offer_id - int
     credits_earned_in_cents - int
     credits_used_in_cents - int
-    credit_type - enum[AWARD, INVITEE, NO_TYPE]
+    // AWARD=1, INVITEE=2, NO_TYPE=0
+    credit_type - int
     expires_at - timestamp
     created_at - timestamp
     referred_by - bytea (nullable)
--- a/docs/design/marketing-offer-program.md
+++ b/docs/design/marketing-offer-program.md
@@ -67,31 +67,32 @@ The credit is automatically applied to the account and will have a max limit tha
     id - int
     name -  text
     description - text  
-    credits - integer
+    award_credits_in_cents - integer
+    invitee_credits_in_cents - integer
     redeemable_cap - integer
     num_redeemed - integer
     created_at - timestamp
     expires_at - timestamp
-    award_credit_duration_days - int
+    award_credit_duration_months - int
     invitee_credit_duration_days - int
-    type - enum[FREE_TIER, REFERRAL]
-    status - enum[ON_GOING, DEFAULT, EXPIRED, NO_STATUS]
+    status - enum[ACTIVE, DEFAULT, DONE]
     PRIMARY KEY (id)
 ```
 
 **user_credit table**
 ```sql
+    id - int
     user_id - bytea
     offer_id - int
-    credits_earned - int
-    credits_used - int
+    credits_earned_in_cents - int
+    credits_used_in_cents - int
     credit_type - enum[AWARD, INVITEE, NO_TYPE]
     expires_at - timestamp
     created_at - timestamp
     referred_by - bytea (nullable)
     FOREIGN KEY (offer_id)
     FOREIGN KEY (referred_by)
-    PRIMARY KEY (user_id)
+    FOREIGN KEY (user_id)
 ```
 
 **user table**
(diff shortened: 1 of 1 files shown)


docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,145 @@ The credit is automatically applied to the account and will have a max limit tha
...
     description - text  
-    credits - integer
+    award_credits_in_cents - integer
+    invitee_credits_in_cents - integer
     redeemable_cap - integer
     num_redeemed - integer
     created_at - timestamp
     expires_at - timestamp
-    award_credit_duration_days - int
+    award_credit_duration_months - int
```
stefanbenten: Couldn't that lead to calculation problems? Lets say i started on the 17th of Feb. Do i get until the 16th of March?
Is that the same for March - April (3 days more)?

docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,145 @@ The credit is automatically applied to the account and will have a max limit tha
...
     invitee_credit_duration_days - int
-    type - enum[FREE_TIER, REFERRAL]
-    status - enum[ON_GOING, DEFAULT, EXPIRED, NO_STATUS]
+    status - enum[ACTIVE, DEFAULT, DONE]
```
stefanbenten: Could we offload this logic to the software and just store an integer for that?
I feel that is better performing and needs less resources

VinozzZ: That's the idea. I wrote it this way to show the design of the database

docs/design/marketing-offer-program.md
VinozzZ: We have been going back and forth on this. I think you are right, changed it back to be days

docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,146 @@ The credit is automatically applied to the account and will have a max limit tha
...
+
+**satellite/satellitedb/offers.go**
+```golang
+func (o *offersDB) Create(ctx context.Context, offer *marketing.Offer) error {
+  tx, err := o.db.Open(ctx)
+  if err != nil {
+    return marketing.OfferError.Wrap(err)
+  }
+
+  _, err := tx.Get_Offer_By_Status(ctx, dbx.Offer_Status(offer.Status))
```
mobyvb: I am not sure I understand how you can `Get_Offer_By_Status`, `Update_Offer_By_Status`, etc... when the status is a simple enum and not a key. How does it know which offer to update or get (less applicable for create)

VinozzZ: That's a good point. I think the idea is that there will only be one active offer at a time. But maybe to be more secure, I should get the offer that has an active status with its expiration date to be greater than `NOW()` and the latest `created_at` ? What do you think?

mobyvb: Oh, I didn't realize there would only be one active offer at a time. If this is true, then I don't actually see a problem with doing things this way. I should review the unchanged parts of the doc to get a better understanding

VinozzZ: That's a business requirement, but it's definitely a good point to make sure that we only get one offer back from the database as current offer

docs/design/marketing-offer-program.md
```diff
@@ -29,12 +29,12 @@ Existing clients give their UUID Referral Link to their friends for their friend
...
 
 - When free credit is awarded:
-It is displayed as the USD value that is equivalent to the storage & bandwidth allotted amount for the credit
+It is displayed as the Network activity value (gb) that is equivalent to the storage & bandwidth allotted amount for the credit
 The credit is automatically applied to the account and will have a max limit that can be reached within a set timeframe, as well as an expiration date for it to be used by.
   The user will be charged for activity that is beyond the Free Credit limit within the set timeframe
   The user will be charged for activity that is after the expiration date.
   Example statement to the user: 
-  “You will not be charged until all of your $XX credit is used or it expires. Your $XX credit will be applied immediately”
+  “You will not be charged until all of your XX credit is used or it expires. Your $XX credit will be applied immediately”
```
mobyvb: Forgot to remove second dollar sign?

docs/design/marketing-offer-program.md
mobyvb: Wait so there will never be two offers with different IDs associated with different users that are both active? I still think I could be misunderstanding

VinozzZ: No worries, it's complicated for sure. Ideally, there will always be two ongoing offers. One( status: active) is used for referral program, one(status: default) is for new user signup. If the referral offer is stopped or expired, it will use the default offer until a new offer starts

docs/design/marketing-offer-program.md
VinozzZ: Business just confirmed that `credits` will be using `USD` as its unit. I'm reverting it back

docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,164 @@ The credit is automatically applied to the account and will have a max limit tha
...
+      WHERE offers.status=?
+      AND NOT EXISTS (
+      SELECT * FROM o
+      ) order by offers.created_at desc;`
+  }
+
+  currentOffer := returnedValueWithLatestCreatedAt
+  return currentOffer, nil
+}
+func (o *offersDB) Create(ctx context.Context, offer *marketing.Offer) error {
```
mobyvb: On a higher level, could you explain what this function is doing? It is a little difficult for me to decipher what it's doing and why just by looking at the code.

VinozzZ: Sorry for the confusion
```golang
// First check if there is an `active` or `default` offer
// If there's one, we need to update its status to be `Done`
// then we will create the new offer
```

docs/design/marketing-offer-program.md
VinozzZ: Removed it

docs/design/marketing-offer-program.md
mobyvb: Ok that makes sense

docs/design/marketing-offer-program.md
fuskovic: nit-pick: The sentence reads fine w/o the `$xx` but I know it's just an example statement

docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,97 @@ The credit is automatically applied to the account and will have a max limit tha
...
 }
 
-func (offer *Offer) IsUpdatable() bool {
-  return offer.Status == NO_STATUS
+type OfferStatus int
+const (
+  Done OfferStatus = 0
+  Active OfferStatus = 1
+  Default OfferStatus = 2
+)
```
fuskovic: nit-pick: This is a good place to use an iota (https://github.com/golang/go/wiki/Iota)

docs/design/marketing-offer-program.md
```diff
@@ -67,54 +67,97 @@ The credit is automatically applied to the account and will have a max limit tha
...
+  }
+
+  return nil
+}
+```
+
+**satellite/marketing/offers.go**
 ```golang
-type DB interface {
+type Offers interface {
```
fuskovic: Shouldn't the StopOffer method be included in this interface?

docs/design/marketing-offer-program.md
VinozzZ: Yes, this is a rough design, not the exact implementation

docs/design/marketing-offer-program.md
VinozzZ: We won't have an explicit StopOffer method since what it would be doing is also an update

PR #2047, opened 2019-05-24
Fix flaky TestProjectUsageBandwidth

mniewrzal:
What: attempt to fix flaky TestProjectUsageBandwidth test

Why:

Thanks for submitting a PR!

### DIFF
--- a/pkg/accounting/projectusage_test.go
+++ b/pkg/accounting/projectusage_test.go
@@ -82,7 +82,6 @@ func TestProjectUsageStorage(t *testing.T) {
 }
 
 func TestProjectUsageBandwidth(t *testing.T) {
-	t.Skip("flaky")
 	cases := []struct {
 		name             string
 		expectedExceeded bool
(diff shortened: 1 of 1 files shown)


pkg/accounting/projectusage_test.go
```diff
@@ -121,7 +119,9 @@ func TestProjectUsageBandwidth(t *testing.T) {
 				expectedData := make([]byte, 50*memory.KiB)
 				_, err = rand.Read(expectedData)
 				require.NoError(t, err)
-				err := planet.Uplinks[0].Upload(ctx, planet.Satellites[0], bucketName, "test/path", expectedData)
+
+				filePath := "test/path"
+				err = planet.Uplinks[0].Upload(ctx, planet.Satellites[0], bucketName, filePath, expectedData)
```
JessicaGreben: what is changed that fixes the flakiness?

mniewrzal: problem was with uploading the same file two times (two tests). Its not related to this functionality. I just reorganized test to use new test planet for each test. I prepared another test that will show this problem and maybe will be possible to track it easier. I just need to push it.

JessicaGreben: oh ok so if its just the name of the file causing the problem, then could we just add a `filename` to the test case struct versus doing all `testplanet.Run` again since that makes tests slower?

mniewrzal: I tried but it didn't work for me. It looks to be more complicated. You can try and push version you described. Maybe you will have a better guess. In worse case scenario we can back to one testplanet per many tests when we will fix root cause.

JessicaGreben: ah ok so if it doesnt work, then these changes look good.

ifraixedes: you may run the subtests with `t.Parallel` and see if they run faster, but I think that's better to merge this and fix the flaky test than delaying this fix just for the purpose to analyze if the test run faster with the `t.Parallel`.

PR #2048, opened 2019-05-24
pkg/audit: Readdress ignored Verifier tests

ifraixedes:
What: The `Verifier` type of the `pkg/audit` had a test which was totally ignored
due to a +build tag.

Why: Ignoring test using a `+build` is confusing because it isn't clear that
the test is fully ignored and the test doesn't get compiled anymore so
all the breaking changes in the implementation don't warm to update the
test.

This changes remove the +build tag and resuscitate the tests which
haven't required that much changes, the rest of them have been deleted.

Thanks for submitting a PR!

### DIFF
--- a/pkg/audit/verifier_test.go
+++ b/pkg/audit/verifier_test.go
@@ -1,105 +1,15 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-// +build ignore
-
-package audit_test
+package audit
 
 import (
 	"context"
-	"crypto/rand"
-	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/vivint/infectious"
-
-	"storj.io/storj/internal/teststorj"
-	"storj.io/storj/pkg/audit"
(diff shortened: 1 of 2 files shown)


pkg/audit/verifier_test.go
```diff
@@ -137,22 +47,26 @@ func TestFailingAudit(t *testing.T) {
...
 		panic(err)
 	}
+
 	for i, num := range pieceNums {
 		if num != badPieceNums[i] {
 			t.Fatal("expected nums in pieceNums to be same as in badPieceNums")
 		}
 	}
+
+	assert.Equal(t, shares, correctedShares)
```
navillasa: maybe `require` instead?

ifraixedes: oh, yes, I had a loop before which was using `require` when I've realized that it could be compared directly, no idea why I used `assert`, although the output is the same for this case.

ifraixedes: I pushed a new commit with this change and realized that there was another `assert` in the test which already exists, hence I've replaced it too.

PR #2049, opened 2019-05-24
Enable Scopelint Linter

ethanadams:
What: 
add scopelint

Why:
enable them to detect bugs and issues.

Thanks for submitting a PR!


pkg/peertls/extensions/extensions_test.go
```diff
@@ -37,22 +37,23 @@ func TestHandlers_Register(t *testing.T) {
 		require.NoError(t, err)
 		chains = append(chains, identity.ToChains(chain))
 
+		index := i
```
ifraixedes: About this one I cannot say with a quick look, I'm not sure when the function passed to `extensions.NewHandleFactory` executes.

I suspect that it executed correctly because it's a test and it passed before and now with these changes.

Nonetheless it's saver to make these changes proposed by the linter.

pkg/peertls/extensions/extensions_test.go
```diff
@@ -88,22 +89,23 @@ func TestHandlers_WithOptions(t *testing.T) {
...
```
ifraixedes: Same comment than just above

satellite/console/service.go
```diff
@@ -968,7 +968,8 @@ func (s *Service) isProjectMember(ctx context.Context, userID uuid.UUID, project
 		return result, errs.New(internalErrMsg)
 	}
 
-	for _, membership := range memberships {
+	for _, projectMembership := range memberships {
+		membership := projectMembership
```
ifraixedes: This is a false positive because after the assignation of the loop declared var the loop ends due to the `return`

storage/testsuite/test_parallel.go
```diff
@@ -22,7 +22,8 @@ func testParallel(t *testing.T, store storage.KeyValueStore) {
 	defer cleanupItems(store, items)
 
 	for i := range items {
-		item := items[i]
+		index := i
```
ifraixedes: This is needed, there was a bug, however the test didn't fail because the index is only used for a test message output.

PR #2051, opened 2019-05-24
Enable gocritic linter

ethanadams:
What: 
Add gocritic linter
Why:
Detect potential  bugs and issues.

Thanks for submitting a PR!


.golangci.yml
```diff
@@ -109,7 +110,10 @@ linters-settings:
     simple: true
     range-loops: true # Report preallocation suggestions on range loops, true by default
     for-loops: false # Report preallocation suggestions on for loops, false by default
-
+  gocritic:
+    disabled-checks:
+    # Disabling appendAssign b/c it reports false positives
```
ifraixedes: Are there many false positives?

ethanadams: I reports 1 false positive.  

cmd/storj-sim/prefix.go:87:12: appendAssign: append result not assigned to the same slice (gocritic)
        buffer = append(writer.buffer, data...)
                 ^
make: *** [lint] Error 1

The assignment appears to be intentional b/c the assignment to write.buffer is deferred in the next line.

egonelbre: Note you can skip specific lines with an appropriate comment rather than disable the whole check.

ethanadams: Thanks for the tip.  I've add //nolint to the line and re-enabled the appendAssign check

cmd/storj-sim/prefix.go
```diff
@@ -84,7 +84,7 @@ func (writer *prefixWriter) Write(data []byte) (int, error) {
 
 	// buffer everything that hasn't been written yet
 	if len(writer.buffer) > 0 {
-		buffer = append(writer.buffer, data...)
+		buffer = append(writer.buffer, data...) //nolint
```
ifraixedes: I would exclude only `gocritic` not all the linters

reference: https://github.com/golangci/golangci-lint#false-positives

cmd/uplink/cmd/share.go
```diff
@@ -47,19 +47,20 @@ func init() {
 const shareISO8601 = "2006-01-02T15:04:05-0700"
 
 func parseHumanDate(date string, now time.Time) (*time.Time, error) {
-	if date == "" {
+	switch {
```
egonelbre: `switch date {`

cmd/storj-sim/prefix.go
ethanadams: Good catch.  resolved an pushed

Issue #2055, opened 2019-05-24
Minor: Size calculation in storj/internal/memory/size.go should test against absolute value to account for negative numbers


ReneSmeekes:
The size calculation checks if size is above a certain value in order to scale up to larger units. As a result all negative sizes are displayed in bytes.

Applicable code storj/internal/memory/size.go
```
// Base2String converts size to a string using base-2 prefixes
func (size Size) Base2String() string {
	if size == 0 {
		return "0"
	}

	switch {
	case size >= EiB*2/3:
		return fmt.Sprintf("%.1f EiB", size.EiB())
	case size >= PiB*2/3:
		return fmt.Sprintf("%.1f PiB", size.PiB())
	case size >= TiB*2/3:
		return fmt.Sprintf("%.1f TiB", size.TiB())
	case size >= GiB*2/3:
		return fmt.Sprintf("%.1f GiB", size.GiB())
	case size >= MiB*2/3:
		return fmt.Sprintf("%.1f MiB", size.MiB())
	case size >= KiB*2/3:
		return fmt.Sprintf("%.1f KiB", size.KiB())
	}

	return strconv.FormatInt(size.Int64(), 10) + " B"
}

// Base10String converts size to a string using base-10 prefixes
func (size Size) Base10String() string {
	if size == 0 {
		return "0"
	}

	switch {
	case size >= EB*2/3:
		return fmt.Sprintf("%.1f EB", size.EB())
	case size >= PB*2/3:
		return fmt.Sprintf("%.1f PB", size.PB())
	case size >= TB*2/3:
		return fmt.Sprintf("%.1f TB", size.TB())
	case size >= GB*2/3:
		return fmt.Sprintf("%.1f GB", size.GB())
	case size >= MB*2/3:
		return fmt.Sprintf("%.1f MB", size.MB())
	case size >= KB*2/3:
		return fmt.Sprintf("%.1f KB", size.KB())
	}

	return strconv.FormatInt(size.Int64(), 10) + " B"
}
```

Suggested change
```
// Base2String converts size to a string using base-2 prefixes
func (size Size) Base2String() string {
	if size == 0 {
		return "0 B"
	}

	switch {
	case math.Abs(size.Float64()) >= EiB.Float64()*2/3:
		return fmt.Sprintf("%.1f EiB", size.EiB())
	case math.Abs(size.Float64()) >= PiB.Float64()*2/3:
		return fmt.Sprintf("%.1f PiB", size.PiB())
[...]

PR #2056, opened 2019-05-25
Use absolute size for unit calculation to account for negative numbers

ReneSmeekes:
What: 
Currently code checks whether size is larger than the threshold set for the unit used. Because negative numbers are always smaller, they are always displayed in bytes. Using absolute numbers when comparing ensures negative numbers are displayed in the appropriate unit. To be consistent when "0" is returned it should say "0 B" as well.

Why:
This issue causes strange negative numbers to appear on the storagenode dashboard when used space goes slightly over the available space on full nodes. And might appear in other places as well.
This change will display negative sizes numbers in a compact format, similar to positive ones.

This is a fix for minor issue #2055 

Thanks for submitting a PR!

### DIFF
--- a/internal/memory/size.go
+++ b/internal/memory/size.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"math"
 )
 
 // base 2 and base 10 sizes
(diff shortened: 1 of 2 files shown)


ReneSmeekes:
Just signed the CLA.

ReneSmeekes:
Apologies for the many commits. Still new to Go and I'm not sure why this commit won't build. Instead of spamming more commits, I'll leave it to the experts from here.

PR #2059, opened 2019-05-27
audit verifier should try a fresh Kad lookup for a node before marking it as offline

Fadila82:
What: 
If a node appears offline, the audit service should attempt a new Kad lookup instead of immediately marking the node as offline.

Why:
Currently, if the audit verifier attempts to download a piece from a storage node, it will mark the node as offline if it appears to be offline. Instead, it should do a fresh Kad lookup and try to reach the node again. The whitepaper says:
>In the case of a node being offline, the audit failure may be due to the address in the node discovery cache being stale, so another, fresh Kademlia lookup will be attempted.

Only after this has been attempted, should the node be marked as offline.

Thanks for submitting a PR!

### DIFF
--- a/pkg/audit/service.go
+++ b/pkg/audit/service.go
@@ -13,6 +13,7 @@ import (
 	"storj.io/storj/internal/memory"
 	"storj.io/storj/internal/sync2"
 	"storj.io/storj/pkg/identity"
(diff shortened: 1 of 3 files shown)


ifraixedes:
Why this PR got closed?

PR #2061, opened 2019-05-28
Relocate IP Lookup to ensure it is always set

stefanbenten:
What: Move IP Lookup to the cache Put Function

Why: Ensure the IP is always up to date and even set!

Thanks for submitting a PR!


pkg/transport/transport.go
```diff
@@ -105,12 +105,6 @@ func (transport *Transport) DialNode(ctx context.Context, node *pb.Node, opts ..
 		return nil, Error.Wrap(err)
 	}
 
-	ipAddr, err := getIP(conn.Target())
-	if err != nil {
-		return nil, err
-	}
-	node.LastIp = ipAddr
```
ifraixedes: I'm not sure if this part should be removed, I understand that the IP should be updated when the node is saved in the cache, however shouldn't we set the such filed when the node is dialed successful?

stefanbenten: @ifraixedes The IP is being filled further down the code in the alertSuccessful Method, as it is calling the cache PUT function. So it is called still when dialing the node.

stefanbenten:
Testplanet doesnt allow the configuration of other hostnames for the nodes, so the tests are now failing. This is blocking the successful build

pkg/transport/transport.go
ifraixedes: always? or it depends of the observers?

stefanbenten: We were depending on the observers before already.

pkg/overlay/selection_test.go
```diff
@@ -221,17 +219,15 @@ func TestNodeSelection(t *testing.T) {
 }
 
 func TestDistinctIPs(t *testing.T) {
+	if runtime.GOOS == "darwin" {
```
stefanbenten: I have to check that on Windows, but that might not work there as well :)

internal/testplanet/reconfigure.go
```diff
@@ -21,6 +21,7 @@ type Reconfigure struct {
 
 	NewStorageNodeDB func(index int) (storagenode.DB, error)
 	StorageNode      func(index int, config *storagenode.Config)
+	NewIPCount       int
```
stefanbenten: Nice!

PR #383, opened 2018-09-25
Use context to propagate API Key

mniewrzal:
The code is a chaos right now and is far from finished but general idea is to remove api key from pointerdb api and make authentication independent from satellite elements. Api key is passed via context to request interceptor and there validation is done. When api key is valid we can return to uplink generated signature in response header. I still need to figure out many things but I want to be sure that the general idea is ok :)

PR #2067, opened 2019-05-28
add capability of logging all GRPC calls/payloads

thepaul:
If the server config has DebugLogTraffic set, info about every grpc call
will be logged. these logs include peer address, which service was
contacted, the method name, the contents of the RPC payload, and whether
an error was encountered while handling the call.

The intent is to capture logs like this on a production satellite for an
interval of time, and then use that information to replay traffic
against testing networks under controlled conditions to better
investigate performance issues.

When DebugLogTraffic is not set, there does not appear to be a
discernible impact on performance compared to that seen before this
commit.

When DebugLogTraffic _is_ set, there is an impact on performance, of
course. scripts/test-sim.sh takes about 9% more time to complete, and
the satellite is only a portion of the work done there, so the real
overhead on a satellite by itself is likely higher than that. I'm not
sure how to benchmark it effectively, though, without the data I'm
trying to make available here.

What: 

Why:

Thanks for submitting a PR!


thepaul:
One big downside here is the introduction of a new dependency in github.com/grpc-ecosystem/go-grpc-middleware. If anyone knows a simpler way to accomplish this, that would be awesome.

pkg/server/interceptors.go
```diff
@@ -43,12 +46,27 @@ func unaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServ
...
+func yesLogIt(ctx context.Context, fullMethodName string, servingObject interface{}) bool {
+	return true
+}
+
+// WithUnaryLoggingInterceptor combines interceptors for logging all GRPC unary calls with the
+// given other unary interceptors.
+func WithUnaryLoggingInterceptor(log *zap.Logger, otherInterceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
+	interceptors := append([]grpc.UnaryServerInterceptor{
+		grpc_ctxtags.UnaryServerInterceptor(),
+		grpc_zap.PayloadUnaryServerInterceptor(log, yesLogIt),
```
JessicaGreben: i feel like we might want a different zap.Logger here so that these new grpc request logs can be written somewhere else and not mixed in with normal application logs.

jtolds:
> One big downside here is the introduction of a new dependency in github.com/grpc-ecosystem/go-grpc-middleware. If anyone knows a simpler way to accomplish this, that would be awesome.

Egon's approach is that the code that's actually needed for middleware is small enough to inline it. Here's where that came up last time: https://github.com/storj/storj/pull/383#discussion_r222415027

thepaul:
Ah. Well, grpc-middleware doesn't pull in any extra dependencies now- we already have everything else, apparently. And the code being used here looks like does quite a bit more than that snippet.

But I'll see if I can pare it down to something smaller and simpler. I was more optimizing for fast dev time at the expense of this dependency.

jtolds:
lol it doesn't? we have all of its dependencies already? well i suppose that changes the decision calculus

pkg/server/interceptors.go
JessicaGreben: i feel like we might want a this `log` to be a different zap.Logger so that these new grpc request logs can be written somewhere else and not mixed in with normal application logs. I think a new logger would be the way to do that, but not 100%, reading docs now.

pkg/server/interceptors.go
```diff
@@ -43,12 +46,27 @@ func unaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServ
...
-		return a(ctx, req, info, func(actx context.Context, areq interface{}) (interface{}, error) {
-			return b(actx, areq, info, func(bctx context.Context, breq interface{}) (interface{}, error) {
-				return handler(bctx, breq)
-			})
-		})
-	}
+// the always-yes logging decider function (because grpc_zap requires a decider function)
+func yesLogIt(ctx context.Context, fullMethodName string, servingObject interface{}) bool {
+	return true
+}
```
JessicaGreben: is there a way to use this decider function instead of the if/else blocks in server.go, RE: https://github.com/storj/storj/pull/2067/files#diff-09856fe9becddf0199651f451409356aR48

pkg/server/interceptors.go
thepaul: I don't actually think that will be necessary for this one-off tool. I don't expect us to use it more than once or twice, and grep should be entirely up to the task of separating the data. We should avoid spending more time than necessary creating this feature.

pkg/server/interceptors.go
thepaul: I don't think so, no. It looks like deciders have a different signature from interceptors. If the if/else blocks are particularly offensive, we could instead have replacements for grpc_middleware.ChainUnaryServer() and grpc_middleware.ChainStreamServer() which, if one of their arguments is nil, just returns the other argument unchanged.

pkg/server/interceptors.go
JessicaGreben: if this is only going to be used "once or twice", does that mean we expect to remove this change once we are done using it?  Alternatively, if this is just a prototype, should we just keep it on its own branch and test it out there instead of merging into master?

thepaul: I would like this to be a separate branch, because there isn't any particular need for it to be in master. Stefan indicated that the changes would need to be in master before he could run it, though. I don't know whether that rule is his own or from devops. Maybe we can get an exemption?

If this does get merged in to master, I don't think it's vital that it stay there. It could come in handy for developers in some debugging scenarios, but it's reasonable to ask developers to apply a temporary patch if they need this functionality.

JessicaGreben: hmm tricky. i guess in that case sounds like it might be easier to merge to master, even though its less than ideal.

pkg/server/interceptors.go
```diff
@@ -5,17 +5,21 @@ package server
...
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 
+	"storj.io/storj/pkg/identity"
 	"storj.io/storj/storage"
 )
 
-func streamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) (err error) {
+func logOnErrorStreamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) (err error) {
```
coyle: Why change the name vs add a comment?

pkg/server/interceptors.go
```diff
@@ -52,3 +58,57 @@ func combineInterceptors(a, b grpc.UnaryServerInterceptor) grpc.UnaryServerInter
...
+		GRPCMethod:  methodName,
+		PeerAddress: "<no peer???>",
+		Msg:         req,
+	}
+	if peer, ok := peer.FromContext(ctx); ok {
+		reqLog.PeerAddress = peer.Addr.String()
+		if peerIdentity, err := identity.PeerIdentityFromPeer(peer); err == nil {
+			reqLog.PeerNodeID = peerIdentity.ID.String()
+		} else {
+			reqLog.PeerNodeID = fmt.Sprintf("<no peer id: %v>", err)
```
coyle: curious, what's the benefit of this? It seems odd to me to have the peer id possibly be a string with an error.

pkg/server/interceptors.go
```diff
@@ -30,7 +34,7 @@ func streamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamS
 	return err
 }
 
-func unaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{},
+func logOnErrorUnaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{},
 	err error) {
 	resp, err = handler(ctx, req)
```
coyle: why return an interface{} ?

pkg/server/interceptors.go
thepaul: Just because the name becomes confusing once there are other StreamInterceptor instances around.

pkg/server/interceptors.go
thepaul: This wouldn't actually set the peer ID; just what gets logged in place of the peer ID if none can be identified.

pkg/server/interceptors.go
thepaul: That's what `handler` returns (it is an instance of `grpc.UnaryHandler`).

PR #2068, opened 2019-05-29
Command line flags features and cleanup

zeebo:
What: Changes the way we handle command line flags with the process package to handle slices.

Why: Some upcoming features require configuring uplink with slices of data. See [this design doc](https://github.com/storj/storj/blob/fb86238acc9586f954e5c23fe05668da5bc01b58/docs/design/uplink-scopes.md).

---

# Summary

Before this PR, some flags were created by calling `cfgstruct.Bind` and having their fields create a flag. Once the flags were parsed, `viper` was used to acquire all the values from them and config files, and the fields in the struct were set through the flag interface.

This doesn't work for slices of things on config structs very well, since it can only set strings, and for a string slice, it turns out that the implementation in `pflag` appends an entry rather than setting it.

This changes three things:

1. Only have a `Bind` call instead of `Bind` and `BindSetup`, and make `BindSetup` an option instead.
2. Add a `process.Bind` call that takes in a `*cobra.Cmd`, binds the struct to the command's flags, and keeps track of that struct in a global map keyed by the command.
3. Use `viper` to get the values and load them into the bound configuration structs instead of using the flags to propagate the changes.

In this way, we can support whatever rich configuration we want in the config yaml files, while still getting command like flags when important.

---


pkg/process/exec_conf.go
```diff
@@ -195,30 +211,48 @@ func cleanup(cmd *cobra.Command) {
...
-					if err != nil {
-						// flag couldn't be set
-						brokenVals = append(brokenVals, key)
-					}
-
-					// revert Changed value
-					flag.Changed = oldChanged
-				}
+		configMtx.Lock()
+		configValues := configs[cmd]
```
JessicaGreben: curious why we don't use `sync. RWMutex` and `RLock` here?

pkg/process/exec_conf.go
```diff
@@ -195,30 +211,48 @@ func cleanup(cmd *cobra.Command) {
...
+		configMtx.Unlock()
+
+		var (
+			brokenKeys  = map[string]struct{}{}
+			missingKeys = map[string]struct{}{}
+			usedKeys    = map[string]struct{}{}
+			allSettings = vip.AllSettings()
+		)
+
+		// Hacky hack: these two keys are noprefix which breaks all scoping
```
JessicaGreben: do you think big picture, we should change all flags to have prefixes?

PR #2072, opened 2019-05-29
Check errors for possible outcomes from audit's DownloadShares

kaloyan-raev:
What: 

Tests that check errors from possible outcomes of the audit's `Verifier.DownloadShares` method:
- Happy path - all nodes returned an erasure share
- A node is offline and cannot be dialed
- A node is missing the requested piece and returns an error
- Dialing a node times out
- Dialing a node is successful, but downloading the share times out

To make the implementation of the above tests possible, a couple of changes are made as part of this PR:
- The piece store service returns rpc error with code `NotFound` if the requested piece for download does not exist.
- New `audit.min-download-timeout` config is defined, so tests can run faster by setting it to a lower value than the default 5s.

Why:

This will help the audit's verifier to determine whether to report the respective node as offline, failed audit, or contained.

Thanks for submitting a PR!


pkg/audit/newverifier_test.go
```diff
@@ -4,19 +4,313 @@
...
+		cursor := audit.NewCursor(planet.Satellites[0].Metainfo.Service)
+		stripe, _, err := cursor.NextStripe(ctx)
+		require.NoError(t, err)
+
+		network := &transport.SimulatedNetwork{
+			DialLatency:    200 * time.Second,
+			BytesPerSecond: 1 * memory.KiB,
+		}
+
+		slowClient := network.NewClient(planet.Satellites[0].Transport)
```
navillasa: this test will run for ~1s instead of ~10s if you give the transport client a configurable dial timeout:
```
		tlsOpts, err := tlsopts.NewOptions(planet.Satellites[0].Identity, tlsopts.Config{})
		require.NoError(t, err)

		newTransport := transport.NewClientWithTimeouts(tlsOpts, transport.Timeouts{
			Dial: 20 * time.Millisecond,
		})

		slowClient := network.NewClient(newTransport)
		require.NotNil(t, slowClient)
```
however for `TestDownloadSharesDownloadTimeout`, the `maxTransferTime` is hard coded to default to 5 seconds in the verifier code, so that will need to be changed in order for that test's run time to be reduced.

pkg/audit/newverifier_test.go
```diff
@@ -4,19 +4,313 @@
...
+		limits, err := planet.Satellites[0].Orders.Service.CreateAuditOrderLimits(ctx, planet.Satellites[0].Identity.PeerIdentity(), bucketID, stripe.Segment, nil)
+		require.NoError(t, err)
+
+		shares, _, err := verifier.DownloadShares(ctx, limits, stripe.Index, shareSize)
+		require.NoError(t, err)
+
+		for _, share := range shares {
+			assert.True(t, transport.Error.Has(share.Error))
+			assert.Equal(t, context.DeadlineExceeded, errs.Unwrap(share.Error))
+			assert.Equal(t, codes.Unknown, status.Code(errs.Unwrap(share.Error)))
```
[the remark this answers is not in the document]
kaloyan-raev: Agree. Removed in latest commit.

pkg/audit/newverifier_test.go
```diff
@@ -4,19 +4,313 @@
...
+		limits, err := planet.Satellites[0].Orders.Service.CreateAuditOrderLimits(ctx, planet.Satellites[0].Identity.PeerIdentity(), bucketID, stripe.Segment, nil)
+		require.NoError(t, err)
+
+		shares, _, err := verifier.DownloadShares(ctx, limits, stripe.Index, shareSize)
+		require.NoError(t, err)
+
+		for _, share := range shares {
+			assert.Equal(t, codes.DeadlineExceeded, status.Code(errs.Unwrap(share.Error)))
+			assert.False(t, transport.Error.Has(share.Error))
+			assert.NotEqual(t, context.DeadlineExceeded, errs.Unwrap(share.Error))
```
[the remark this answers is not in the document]
kaloyan-raev: Agree. Removed in latest commit.

pkg/audit/newverifier_test.go
kaloyan-raev: This is an awesome hint! Thank you!

storagenode/piecestore/endpoint.go
```diff
@@ -335,7 +338,13 @@ func (endpoint *Endpoint) Download(stream pb.Piecestore_DownloadServer) (err err
 
 	pieceReader, err := endpoint.store.Reader(ctx, limit.SatelliteId, limit.PieceId)
 	if err != nil {
-		return ErrInternal.Wrap(err) // TODO: report grpc status internal server error
+		if os.IsNotExist(err) {
+			return status.Error(codes.NotFound, err.Error())
+		}
+		if os.IsPermission(err) {
+			return status.Error(codes.PermissionDenied, err.Error())
```
ifraixedes: I'm not convinced that if the permission of the files doesn't allow to read the the piece the storagenode has to return this GRPC error,  because we don't use the file permission to bock the access to pieces and if that happens is because of a bad operation of the storage node operator.

For a hones node, which should run this code, for me it should be also `NotFound`.

kaloyan-raev: Yes, you are right. Now I am re-reading the godoc for `PermissionDenied` and it says:
> PermissionDenied indicates the caller does not have permission to execute the specified operation.

It's really not the caller (the satellite or the uplink) that does not have permission to access the piece. It's a problem on the storage node that the piece cannot be opened.

However, I also don't think that `NotFound` is correct too. Because the piece is there, but it cannot be opened because of an issue that can be fixed. How about returning `FailedPrecondition` or `Internal`? Which one would you choose?

ifraixedes: I agree, `NotFound` isn't good in comparison of any of the 2 that you used, however I didn't check the list of codes yesterday and I only based my reply on those 2 error codes that you were using.

Between those 2, I think that `Internal` is more appropriated because while the docs don't say that the client can do whatsoever other operation to change the state of the system for fulfilling the precondition, I think that failed preconditions are based on correct system states, so due to other operations (by the same client, other clients or even by the usual flow of the system) the system will meet the precondition at some point. For this specific case, the permissions should never have changed by our system (I'm not considering that the cause of the change is due to a bug), it has been an external factor which has changed them so I believe that the description of `Internal` fits to it

> Internal errors. This means that some invariants expected by the underlying system have been broken. This error code is reserved for serious errors.

kaloyan-raev: Deal! It will be `Internal` :)

kaloyan-raev:
@navillasa I need your approval

PR #2078, opened 2019-05-29
Add monkit stats to piecestore upload/download

mobyvb:
What: 
https://storjlabs.atlassian.net/browse/V3-1807
Adds the following values to monkit for storagenodes:
`upload_failure_size_bytes` - number of bytes uploaded on failed upload
`upload_failure_duration_ns` - number of nanoseconds taken for failed upload
`upload_failure_rate_bytes_per_sec` - failedUploadTime / failedUploadNumBytes

`upload_success_size_bytes` - number of bytes uploaded on successful upload
`upload_success_duration_ns` - number of nanoseconds taken for successful upload
`upload_success_rate_bytes_per_sec` - successfulUploadTime / successfulUploadNumBytes

`download_failure_size_bytes` - number of bytes downloaded on failed download
`download_failure_duration_ns` - number of nanoseconds taken for failed download
`download_failure_rate_bytes_per_sec` - failedDownloadTime / failedDownloadNumBytes

`download_success_size_bytes` - number of bytes downloaded on successful download
`download_success_duration_ns` - number of nanoseconds taken for successful download
`download_success_rate_bytes_per_sec` - successfulDownloadTime / successfulDownloadNumBytes

Why:

Thanks for submitting a PR!


storagenode/piecestore/endpoint.go
```diff
@@ -152,10 +153,26 @@ func (endpoint *Endpoint) Upload(stream pb.Piecestore_UploadServer) (err error)
...
 
+	var pieceWriter *pieces.Writer
 	defer func() {
+		endTime := time.Now()
+		dt := endTime.Sub(startTime)
+		uploadSize := int64(0)
+		if pieceWriter != nil {
+			uploadSize = pieceWriter.Size()
+		}
+		uploadRate := float64(uploadSize) / dt.Seconds()
```
wthorp: I'm not sure if how accurate the timers are -- if in an error case if this could be division by zero?

storagenode/piecestore/endpoint.go
```diff
@@ -320,10 +338,25 @@ func (endpoint *Endpoint) Download(stream pb.Piecestore_DownloadServer) (err err
...
 
+	var pieceReader *pieces.Reader
 	defer func() {
+		endTime := time.Now()
+		dt := endTime.Sub(startTime)
+		downloadSize := int64(0)
+		if pieceReader != nil {
+			downloadSize = pieceReader.Size()
+		}
+		downloadRate := float64(downloadSize) / dt.Seconds()
```
wthorp: Div by zero?

storagenode/piecestore/endpoint.go
mobyvb: I don't think so, unless the function stops instantaneously and no time elapses between the two `time.Now()`s

storagenode/piecestore/endpoint.go
mobyvb: see other comment

storagenode/piecestore/endpoint.go
stefanbenten: I think Bill is right, depending on the size it can take less than a second and potentially be 0 then

mobyvb: `dt.Seconds()` is a float64 but I will add a check just in case

stefanbenten: Sounds good

storagenode/piecestore/endpoint.go
```diff
@@ -125,6 +125,7 @@ func (endpoint *Endpoint) Delete(ctx context.Context, delete *pb.PieceDeleteRequ
 func (endpoint *Endpoint) Upload(stream pb.Piecestore_UploadServer) (err error) {
 	ctx := stream.Context()
 	defer mon.Task()(&ctx)(&err)
+	startTime := time.Now()
```
navillasa: I agree with Stefan that this and other places using time.Now() should probably be time.Now().UTC() for consistency

storagenode/piecestore/endpoint.go
wthorp: cool, yeah, I wasn't entirely sure but sounds good.

wthorp:
I don't think the UTC() is necessary here, particularly because we're just calculating a duration, not an absolute time.  In addition, the UTC issue we saw before was specific to SQL.

stefanbenten:
I am just making the proposal to unify it to be UTC across the entire code. Solves a LOT of headaches :)

PR #2079, opened 2019-05-29
node disqualification design draft

wthorp:
What: 
node disqualification design draft

Why:
Ours is not to reason why, ours is but to do JIRA or die.
https://storjlabs.atlassian.net/browse/V3-1852


docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+Existing calls to the DBX `UpdateNodeInfo()` method must set `disqualified` if appropriate.  This may be best accomplished by a database trigger or stored procedure.
+
+## Rationale
+
+Although disqualification is largely an atomic operation that would be handled well by an external hash, the inherent tie-ins with node selection make the above solution the most straightforward.  If we were to refactor node selection in the future, I would likely leave disqualified nodes out of the stats database, leaving them only in the overlay cache.
+
+## Implementation (Stories)
+
+- Update nodes table using DBX
+- Create nodes table migration scripts
```
mobyvb: "Update nodes table using DBX" and "Create nodes table migration scripts" seem like they could be the same story

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+- Update node selection SQL (overlaycache.go)
+- Update calls to UpdateNodeInfo()
+- Refactor tests dependent on offline / unreliable nodes to use disqualification
```
mobyvb: I think refactoring the tests should be a part of updating the node selection SQL

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+  id bytea NOT NULL,
+  ...
+  disqualified boolean NOT NULL,
+  PRIMARY KEY ( id )
+);
+```
+
+Existing SQL queries employing logic such as `WHERE audit_success_ratio >= $2 AND uptime_ratio >= $3` would change to `WHERE disqualified = FALSE`.
+
+Existing calls to the DBX `UpdateNodeInfo()` method must set `disqualified` if appropriate.  This may be best accomplished by a database trigger or stored procedure.
```
littleskunk: A database trigger will be database specific code. This will make it harder to switch to a different solution later. The nodes table is creating a high load on the database. We might want to store it somewhere else which would makes the database trigger useless.

Could we do it in the code? At the moment the audit job would like to write back the results we can check if the node is now disqualified. Same for uptime checks.

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+```sql
+CREATE TABLE nodes (
+  id bytea NOT NULL,
+  ...
+  disqualified boolean NOT NULL,
+  PRIMARY KEY ( id )
+);
+```
+
+Existing SQL queries employing logic such as `WHERE audit_success_ratio >= $2 AND uptime_ratio >= $3` would change to `WHERE disqualified = FALSE`.
```
littleskunk: How about new nodes. Can a new node with 1 failed uptime or 1 failed audit check getting disqualified or is it allowed to collect 500 audits before we judge?

docs/design/disqualification.md
wthorp: Good point.  I am a little worried about the extra db locking of getting those values from the database before setting them again... but we can write Postgres SQL at least that should be a single statement.  Other implementations might not even be SQL. Will update.

docs/design/disqualification.md
wthorp: The whitepaper specifically says 'Regardless of... the vetting process'.  It seems more sensible to me that they fail fast and try again instead of failing the moment they become 'vetted'.

wthorp: I feel like the only purpose of new nodes as described in the whitepaper is to deliver less information to new nodes to reduce the risk inherent in new / unknown actors.

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+## Abstract
+
+This design doc outlines how we will implement node disqualification, wherein a storage node is deemed permanently inelligible for storing new data via uploads.
+
+Nodes can be disqualified due to audit failures, failures to return data during repair, or poor uptime statistics.
+
+## Main Objectives
+
+1. To not send new data to nodes which are not behaving as desired.
+2. Efficiently identify and ignore these 'bad' nodes during various node selection and statistics processes.
```
kaloyan-raev: Another objective would be to improve object durability by having the repair checker consider not only offline nodes, but also online, but disqualified nodes.

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+
+> After a storage node is disqualified, the node must go back through the entire vetting process again. If the node decides to start over with a brand-new identity, the node must restart the vetting process from the beginning (in addition to generating a new nodeID via the proof of work system). This strongly disincentivizes storage nodes from being cavalier with their reputation.
+
+Further, the node will be demonetized.
+
+> Provided the storage node hasn’t been disqualified, the storage node will be paid by the Satellite for the data it has stored over the course ofthe month, per the Satellite’s records.
+
+A disqualified SNO should quickly stop particiapting with a satellite it is disqualified and demonetized on.  It may remain in Kademlia and potentially found during Overlay Cache Discover, as the kademlia network supports multiple satellites.
+
+One option that may be allowed for disqualified storage nodes is a Graceful Exit.  [Storage Node Payment and Incentives for V3](https://docs.google.com/document/d/1-Pxzk-ad-0QtF6nnTwfgzk8e_-XbBNSDxuRvnbd0QL8/edit#heading=h.rz1ehm5mbeuz) describes this feature:
```
kaloyan-raev: I think this may trigger some long discussion whether or not we should allow Graceful Exit for nodes that are already disqualified. Perhaps, we should leave it for the future design doc for Graceful Exit.

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+
+> When a node operator wants to leave the network, if they just shut down the node, the unavailability of the data on the network can contribute to repair costs.  If instead, the node triggers a function to call a Satellite, request new storage nodes to store the pieces stored on the node being shut down, then directly upload those pieces to the new nodes, file repair would be avoided.
+
+## Design
+
+### Use of Disqualified Nodes
+
+Disqualified nodes may be used during download of typical downloads from uplinks or via repair.  Therefore, their IP must be tracked in the overlay cache.
```
kaloyan-raev: I don't think DQ nodes should be used during regular downloads if there are enough other nodes to serve the download. I agree that the data on DQ nodes has still some value, especially if we are desperately trying to repair an irreparable segment. But other than that DQ nodes should be avoided.

docs/design/disqualification.md
littleskunk: Good point. Yes I agree. Lets disqualify them as early as possible.

docs/design/disqualification.md
littleskunk: "various node selections" includes the repair checker query.

docs/design/disqualification.md
```diff
@@ -0,0 +1,77 @@
...
+);
+```
+
+Existing SQL queries employing logic such as `WHERE audit_success_ratio >= $2 AND uptime_ratio >= $3` would change to `WHERE disqualified = FALSE`.
+
+Existing calls to the DBX `UpdateNodeInfo()` method must set `disqualified` if appropriate.  This may be best accomplished by a database trigger or stored procedure.
+
+## Rationale
+
+Although disqualification is largely an atomic operation that would be handled well by an external hash, the inherent tie-ins with node selection make the above solution the most straightforward.  If we were to refactor node selection in the future, I would likely leave disqualified nodes out of the stats database, leaving them only in the overlay cache.
```
kaloyan-raev: "overlay cache" and "stats database" are now merged into the "nodes" table. So they are the same thing at the moment.

docs/design/disqualification.md
wthorp: Repair checker currently does care about 'unreliable' nodes at the moment... 'unreliable' was the term we created for 'we aren't doing disqualification, but somehow we're doing this other thing'.

docs/design/disqualification.md
wthorp: I'll add a 'to be determined' statement of some sort here.

docs/design/disqualification.md
wthorp: I'm picturing this as 'Graceful Exit' procedure to be the only monetized action that a DQ'ed node can take.  In the middle of the night, I realized we would need a timestamp if we did it that way, as we currently can't track these 'unpaid downloads'.  I think its fair to expect rational SNOs to never allow 'unpaid downloads' so maybe we need to initiate repair / Graceful Exit immediately.

docs/design/disqualification.md
wthorp: At the moment.  :D

docs/design/disqualification.md
heunland: Please note that graceful exit was meant to not become an option for a SNO until a node has been operating for 15 months as described in blog post https://storj.io/blog/2019/01/sharing-storage-space-for-fun-and-profit/

docs/design/disqualification.md
wthorp: updated

docs/design/disqualification.md
wthorp: updated

docs/design/disqualification.md
wthorp: clarified as per architecture review

docs/design/disqualification.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Determining Disqualification
+
+A node is disqualified when its reputation falls below a fixed value.  We are currently envision two disinct reputation check values, one for uptime and another for audit.  These values will represent an idea value minus some some standard-deviation.  The proposed system for calculation repuations is based on four values: α0, β0, λ, and v.  Because changing these values will change the expected standard deviation of measurements, the reputation cutoff values will vary as these parameters vary.  At this phase, it is exected that these cutoffs are all configured based on numbers from the data science team.  A node will be disqualified if either the audit or the uptime reputation value falls below their disqualification cutoff value.
+
+## Rationale
+
+Although disqualification is largely an atomic operation that would be handled well by an external hash, the inherent tie-ins with node selection make the above solution the most straightforward.  If we were to refactor node selection in the future, I would likely leave disqualified nodes out of the stats database, leaving them only in the overlay cache.
```
ifraixedes: "I would likely leave disqualified nodes" you may want to write "We would likely..."

PR #2038, opened 2019-05-23
Add Stripe invoices and payment methods for projects to console

rikysya:
What: add invoices to console

Why:

Thanks for submitting a PR!

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?

PR #2080, opened 2019-05-30
Console add stripe service

rikysya:
What: add stripe service to the console

Why: first part of the https://github.com/storj/storj/pull/2038

Thanks for submitting a PR!


internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+// GetCustomer retrieves customer object from stripe network
+func (s *StripeService) GetCustomer(ctx context.Context, id string) (*stripe.Customer, error) {
+	return s.client.Customers.Get(id, nil)
+}
+
+// GetCustomerDefaultPaymentMethod retrieves customer default payment method from stripe network
+func (s *StripeService) GetCustomerDefaultPaymentMethod(ctx context.Context, customerID string) (*stripe.PaymentMethod, error) {
+	cus, err := s.client.Customers.Get(customerID, nil)
+	if err != nil {
+		return nil, err
```
VinozzZ: Should we have a `paymentErr` to wrap this?

internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+	for iterator.Next() {
+		pm := iterator.PaymentMethod()
+		paymentMethods = append(paymentMethods, pm)
+	}
+
+	return paymentMethods, nil
+}
+
+// GetPaymentMethod retrieve payment method object from stripe network
+func (s *StripeService) GetPaymentMethod(ctx context.Context, id string) (*stripe.PaymentMethod, error) {
```
VinozzZ: Is the `id` here a paymentMethod id from stripe?

internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+// - Egress
+// - ObjectsCount
+// Created invoice has AutoAdvance property set to true, so it will be finalized
+// (no further editing) and attempted to be paid in 1 hour after creation
+func (s *StripeService) CreateProjectInvoice(ctx context.Context, params CreateProjectInvoiceParams) (*stripe.Invoice, error) {
+	// create line items
+	_, err := s.client.InvoiceItems.New(&stripe.InvoiceItemParams{
+		Customer:    stripe.String(params.CustomerID),
+		Description: stripe.String("Storage"),
+		Quantity:    stripe.Int64(int64(params.Storage)),
```
VinozzZ: Will there ever be a chance that `params.Storage` overflow in the conversion here?

internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+		Currency:    stripe.String(string(stripe.CurrencyUSD)),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = s.client.InvoiceItems.New(&stripe.InvoiceItemParams{
+		Customer:    stripe.String(params.CustomerID),
+		Description: stripe.String("Egress"),
+		Quantity:    stripe.Int64(int64(params.Egress)),
```
VinozzZ: same as above

satellite/console/service.go
```diff
@@ -55,37 +56,36 @@ const (
...
-
 	if store == nil {
 		return nil, errs.New("store can't be nil")
 	}
-
 	if log == nil {
 		return nil, errs.New("log can't be nil")
 	}
-
 	if passwordCost == 0 {
```
VinozzZ: I see that we are checking all the arguments against its zero value here. Do we need to also add the check for `stripeService`?

satellite/console/service.go
```diff
@@ -119,22 +119,41 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
-		Email:        user.Email,
-		FullName:     user.FullName,
-		ShortName:    user.ShortName,
-		PasswordHash: hash,
-	})
+	// store data
+	tx, err := s.store.BeginTx(ctx)
 	if err != nil {
-		return nil, errs.New(internalErrMsg)
+		return nil, err
```
VinozzZ: Is there a reason why we aren't wrapping this error?

internal/payments/service.go
rikysya: yes

satellite/console/service.go
rikysya: I don't think we need it for internal pkg

internal/payments/service.go
rikysya: I am not sure

internal/payments/service.go
rikysya: I don't think anyone can store more than `9,223,372,036,854,775,807` `GB` in the nearest future

satellite/console/service.go
rikysya: that's not wrapping, but replacing error contents completely with user friendly, but non-descriptive error. I am not fan of this approach, but if needed it can be replaced in the future after some testing period, because it's much easier to identify errors/bugs with real error msgs

internal/payments/logger.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+
+	"github.com/stripe/stripe-go"
+	"go.uber.org/zap"
+)
+
+// zapWrapper is a wrapper for *zap.Logger.
+// Implements stripe.LeveledLoggerInterface
+type zapWrapper struct {
```
stefanbenten: Couldnt we get away with a Sugared Logger?
See below implementation proposal

internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+	Egress      float64
+	ObjectCount float64
+
+	StartDate time.Time
+	EndDate   time.Time
+}
+
+// NewService creates new instance of StripeService initialized with API key
+func NewService(log *zap.Logger, apiKey string) *StripeService {
+	stripe.DefaultLeveledLogger = wrapLogger(log)
```
stefanbenten: stripe.DefaultLeveledLogger = log.Sugar().Named("") should be valid (Didnt test it tho).
It contains the necessary interface functions.

internal/payments/service.go
```diff
@@ -0,0 +1,218 @@
...
+// CreateCustomer creates new customer from CustomerParams struct
+// sets default payment to one of the predefined testing VISA credit cards
+func (s *StripeService) CreateCustomer(ctx context.Context, params CreateCustomerParams) (*stripe.Customer, error) {
+	cparams := &stripe.CustomerParams{
+		Email:       stripe.String(params.Email),
+		Name:        stripe.String(params.Name),
+		Description: stripe.String(params.Description),
+	}
+
+	// Set default source (payment instrument)
```
stefanbenten: Remove commented code

internal/payments/service.go
stefanbenten: I like that idea, the more clear the log messages, the better!

satellite/console/service.go
```diff
@@ -55,37 +56,36 @@ const (
...
+	Signer
+	store         DB
+	stripeService payments.Service
 
 	passwordCost int
 }
 
 // NewService returns new instance of Service
-func NewService(log *zap.Logger, signer Signer, store DB, passwordCost int) (*Service, error) {
+func NewService(log *zap.Logger, signer Signer, store DB, stripeService payments.Service, passwordCost int) (*Service, error) {
```
stefanbenten: What happens if a particular Satellite Operator dont want to or can not use Stripe?
Can the satellite be operated without it?

satellite/console/service.go
```diff
@@ -55,37 +56,36 @@ const (
 
 // Service is handling accounts related logic
 type Service struct {
-	Signer
+	log *zap.Logger
 
-	store DB
-	log   *zap.Logger
+	Signer
```
stefanbenten: Can we split external from internal vars into blocks?
Dont think its necessary, but makes the code clearer imo

satellite/console/service.go
rikysya: Honestly, I don't know. When we discussed payments we wanted to use stripe wherever we can, so decoupling it gonna break everything

satellite/console/service.go
VinozzZ: That makes sense. How about using the `Error = errs.Class("console error")` and then `Error.Wrap(err)`. I saw this pattern in other services

satellite/console/service.go
```diff
@@ -119,22 +119,41 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
+		return nil, err
 	}
 
-	err = s.store.RegistrationTokens().UpdateOwner(ctx, registrationToken.Secret, u.ID)
-	if err != nil {
-		return nil, errs.New(internalErrMsg)
-	}
+	// clean up user on error
+	defer func() {
+		if err != nil {
```
jtolds: this is pretty scary. this will suddenly start having bugs if for some reason someone starts changing the names of the return values. can you put this defer at the very top of the function before any errors are defined so it's clear we mean for it to refer to the named return value?

satellite/console/consoleweb/consoleql/query_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"storj.io/storj/satellite/console/consoleauth"
 	"storj.io/storj/satellite/console/consoleweb/consoleql"
 	"storj.io/storj/satellite/mailservice"
+	"storj.io/storj/satellite/payments/internalservice"
```
jtolds: can we name this package something more descriptive? `internalservice` is surprising

satellite/console/service.go
```diff
@@ -119,22 +119,41 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
+			return errs.New(internalErrMsg)
+		}
+
+		err = tx.RegistrationTokens().UpdateOwner(ctx, registrationToken.Secret, u.ID)
+		if err != nil {
+			return errs.New(internalErrMsg)
+		}
+
+		return nil
+	})
```
jtolds: why is this create user stuff changing in this PR at all? isn't this PR about stripe?

one thing you can do to break up commits is use `git add -p` which is very helpful. it allows you to add only some changes in a file. is this change stripe related?

satellite/console/service.go
```diff
@@ -441,39 +460,38 @@ func (s *Service) CreateProject(ctx context.Context, projectInfo ProjectInfo) (p
...
-	}
-
-	transaction, err := s.store.BeginTx(ctx)
+	tx, err := s.store.BeginTx(ctx)
 	if err != nil {
 		return nil, errs.New(internalErrMsg)
 	}
 
+	// clean up project on error
 	defer func() {
```
jtolds: same about this defer. if someone changes the named return value names, everything will still compile, but this defer method will stop working right. instead, if this was at the top of the function, then it wouldn't compile if someone changed the names.

satellite/payments/internalservice/internal.go
```diff
@@ -0,0 +1,110 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package internalservice
```
jtolds: maybe this is called "internalpayments" or "localpayments"?

satellite/console/service.go
rikysya: as this is only part of original PR, those changes were added later with more stripe related stuff. And when picking changes to commit I accidently picked them up to this PR

PR #2081, opened 2019-05-30
Fix storage and object_count calculations on console

rikysya:
What:  fix storage and object values for buckets on console. Instead of storage*hours and objects*hours as we have in rollups, show the current bucket size and object count

https://storjlabs.atlassian.net/browse/V3-1791 there is still this issue so displayed values won't be correct, until it fixed

Why: https://storjlabs.atlassian.net/browse/V3-1866

Thanks for submitting a PR!


satellite/satellitedb/usagerollups.go
```diff
@@ -293,29 +298,24 @@ func (db *usagerollups) GetBucketTotals(ctx context.Context, projectID uuid.UUID
...
+		var objectCount uint64
+		err = storageRow.Scan(&inline, &remote, &objectCount)
+		if err != nil {
+			if err != sql.ErrNoRows {
+				return nil, err
+			}
 		}
 
+		// fill storage and object count
+		bucketUsage.Storage = memory.Size(inline + remote).GB()
```
mobyvb: should this still be done if `err == sql.ErrNoRows`?

satellite/console/usagerollups.go
```diff
@@ -35,7 +35,7 @@ type BucketUsage struct {
 
 	Storage     float64
 	Egress      float64
-	ObjectCount float64
+	ObjectCount uint64
```
VinozzZ: Not sure if this applies to this convention but we seem to prefer using `int` over unsigned integer https://github.com/storj/docs/blob/master/code/Style.md#integers

satellite/satellitedb/usagerollups.go
rikysya: yeah, it will return zero

satellite/console/usagerollups.go
rikysya: seems, you are right

PR #2085, opened 2019-05-30
add user credits table

fuskovic:
What: Adds db model for user credits table

Why: To keep track of user credits for the referral program

Thanks for submitting a PR!


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,26 @@ update offer (
...
 )
 
-delete offer ( where offer.id = ? )
\ No newline at end of file
+delete offer ( where offer.id = ? )
+
+//--- user credit table ---//
+
+model user_credit (
+	key		user_id
```
VinozzZ: The `user_id` won't be unique in this table, we probably need a `id` as the primary key that's just a `serial`. I have a PR up for the new schema https://github.com/storj/storj/pull/2046/files#diff-081def8b8f7fc45ec1f5c91b31f5d030

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,26 @@ update offer (
...
+
+	field	user_id user.id	cascade
```
VinozzZ: If we delete this record when a user account is deleted, then we will also loose the record of who referred this user. I think we should keep this record

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,26 @@ update offer (
...
+	field	offer_id offer.id	cascade
+	field	referred_by user.id	cascade
+	field	credits_earned	int
+	field	credits_used	int
+	field	credits_type	text
+	field	expires_at		timestamp	(autoinsert)
+	field	created_at		timestamp	(autoinsert)
+)
+
+read all ( select user_credit )
```
VinozzZ: Do we have a use case for getting all the records from this table?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,26 @@ update offer (
...
+delete offer ( where offer.id = ? )
+
+//--- user credit table ---//
+
+model user_credit (
+	key		user_id
+
+	field	user_id user.id	cascade
+	field	offer_id offer.id	cascade
+	field	referred_by user.id	cascade
```
VinozzZ: Same idea as above. When a user is deleted, we want the referred user still maintain the credits

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,26 @@ update offer (
...
+	field	credits_earned	int
+	field	credits_used	int
+	field	credits_type	text
```
VinozzZ: This type should be `int` and we will create the `const` in the code for the value

satellite/satellitedb/dbx/satellitedb.dbx
fuskovic: Not that I'm aware of. I was following convention when I added this iirc.

satellite/satellitedb/dbx/satellitedb.dbx
fuskovic: The documentation says the relation kind I should use for your suggestion is `restrict` so I will replace all instances of `cascade` with restrict.

From the dbx docs:

> Foreign Key Relation Kinds

A foreign key relation can be any of these

setnull: when the related row goes away, set this field to null. the field must be nullable.
cascade: when the related row goes away, delete this row.
restrict: do not allow the related row to go away.

satellite/satellitedb/dbx/satellitedb.dbx
fuskovic: see https://github.com/storj/storj/pull/2085#discussion_r290428198

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,25 @@ update offer (
...
+//--- user credit table ---//
+
+model user_credit (
+	key		id
+
+	field	id				serial
+	field	user_id user.id	restrict
+	field	offer_id offer.id	restrict
+	field	referred_by user.id	restrict
+	field	credits_earned	int
```
VinozzZ: Can we name this `credits_earned_in_cent` to make it clear that we are using `cent` as the unit? same with `credits_used_in_cent`

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,25 @@ update offer (
...
+	field	credits_used	int
+	field	credits_type	int
```
VinozzZ: nit pick: this can be `credit_type`

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -619,4 +619,25 @@ update offer (
...
+	field	expires_at		timestamp	(autoinsert)
+	field	created_at		timestamp	(autoinsert)
+)
+
+read one (
```
VinozzZ: This would be `read all` since it's very likely that one user will have multiple entry for an offer

fuskovic: I noticed you added this so resolving

satellite/satellitedb/dbx/satellitedb.dbx
fuskovic: Why did you delete credit_type from the model ?

satellite/satellitedb/dbx/satellitedb.dbx
fuskovic: I noticed you added this so resolving

satellite/satellitedb/dbx/satellitedb.dbx
VinozzZ: By looking at the `referred_by`, if it's nil, we know it's not an invitee credit, if it has value, we know it's a referral invitee credit. That's all we care about right now. So we don't have use of the `credit_type`

fuskovic: Gotcha

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -680,3 +680,37 @@ update offer (
...
+	field referred_by user.id	restrict ( nullable )
+
+	field credits_earned_in_cents	int
+	field credits_used_in_cents	int ( updatable, autoinsert )
+
+	field expires_at		timestamp	( updatable )
+	field created_at		timestamp	( autoinsert )
+)
+
+read all (
```
mobyvb: maybe it would be better to write your own query for this in something that wraps the dbx stuff? The functions that dbx autogenerates from this look pretty crazy/verbose.
Here is an example of what I am saying you might be able to do: https://github.com/storj/storj/blob/master/satellite/satellitedb/overlaycache.go#L360-L380

PR #2088, opened 2019-05-31
Better metainfo Create/Commit request validation

mniewrzal:
What: This PR reorganize metainfo endpoint to validate params between `CreateSegment` and `CommitSegment` to avoid committing remote segments without calling create or to avoid setting different upload parameters for those two methods.
Changes:
* all things related to validation moved to separate file `validation.go`
* user can `CommitSegment` (remote) only when `CreateSegment` was called earlier
* `CommitSegment` can be executed only once for one `CreateSegment` 
* Parameters between `create` and `commit` are compared: expiration date, redundancy scheme
* Time between `CreateSegment` and `CommitSegment` is limited (for now hardcoded 4h) 

Why: Lack of validation gives users the ability to cheat. https://storjlabs.atlassian.net/browse/V3-1673


mniewrzal:
I'm not sure if I make removing old entries overcomplicated but I didn't have a better idea at the moment.

satellite/metainfo/validation.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func newCreateRequests() *createRequests {
+	return &createRequests{
+		requests: make(map[storj.SerialNumber]*createRequest),
+		ttlItems: make([]*TTLItem, 0),
+	}
+}
+
+func (cr *createRequests) Put(serialNumber storj.SerialNumber, createRequest *createRequest) {
```
ifraixedes: According our current coding style name suggestion, it looks that `cr` should be of a more explicit name.

mniewrzal: made some changes, hope is not worse :)

PR #2090, opened 2019-05-31
uplink: enc.encryption-key flag is only available for setup command

ifraixedes:
**What:**

We moved the encryption key outside from the `uplink` configuration file and to be stored in a separated file, however at that time we didn't find a way in how to make the `test-sim` working without being able to have a flag to pass explicitly an encryption key which has preference over the one stored in the file (in case that the configuration pointed to some encryption key file path).

This PR explicitly enable only the `--enc.encryption-key` flag of the `uplink` (and `gateway`, which is bound to it) for only be available for the `setup` command, which implies that it doesn't appear anymore in the configuration file but also it isn't possible to pass the encryption key by a flag when using any other command than `setup`, hence the encryption key is always read from the encryption key file.

**Why:**

Because the current package, that we use to handle configuration files and command line flags for command-line binaries, didn't allow to have a flag which can be used for any command but excluded to be set in the configuration file the `--enc.encryption-key` flag ended up appearing in the configuration file when it shouldn't.

**NOTE** reviewers, consider to read the commit messages, before commenting or asking, you may find the answer on them.

Thanks for submitting a PR!


cmd/uplink/cmd/common_test.go
```diff
@@ -89,11 +89,12 @@ func TestUseOrLoadEncryptionKeyIntoEncryptionAccess(t *testing.T) {
 }
 
 func TestSaveLoadEncryptionKey(t *testing.T) {
-	var inputKey []byte
+	var inputKey string
```
VinozzZ: This definitely helps with readability in the code! Do you mind to elaborate more on the benefit of initializing the `inputKey` as a string?

cmd/uplink/cmd/common_test.go
```diff
@@ -89,11 +89,12 @@ func TestUseOrLoadEncryptionKeyIntoEncryptionAccess(t *testing.T) {
...
 	{
-		inputKey = make([]byte, rand.Intn(storj.KeySize)*3+1)
-		_, err := rand.Read(inputKey)
+		randKey := make([]byte, rand.Intn(storj.KeySize)*3+1)
```
mobyvb: this does not guarantee that len(randKey) > storj.KeySize. This will make the `inputKey[:storj.KeySize]` portion of the test panic in rare situations. See this related jenkins failure and corresponding fix:
https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/master/145/tests/
https://github.com/storj/storj/commit/7a9742f47b19dc88cbd6e63bb0b332bea458cb1c

cmd/gateway/main.go
```diff
@@ -100,24 +100,58 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
+		}
+	}
+
+	err := process.SaveConfigWithAllDefaults(cmd.Flags(), filepath.Join(setupDir, "config.yaml"), overrides)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	if setupCfg.Enc.EncryptionKey != "" {
+		_, _ = fmt.Printf("Your encryption key is saved to: %s\n", encryptionKeyFilepath)
```
VinozzZ: if we aren't using the return value, why do we need this assignment here?

cmd/uplink/cmd/common_test.go
ifraixedes: It isn't about readability of the test, it's because `uplink.SaveEncryptionKey` takes a string as a first parameter https://github.com/storj/storj/pull/2090/files#diff-2b3f299c98b7df4789fbfde5c3e41b40R14

cmd/uplink/cmd/common_test.go
ifraixedes: I was aware of that bug in the test which happened because of later changes that I had to do because they were requested in the review and I forgot to update the tests, an due the randomness the tests didn't always panic.

Meanwhile you're right regarding the fact of the key size, the current assertions don't always use the same length to compare the got and expected value, and if I'm not mistaken the conditional should avoid the previous panics that we had.

see https://github.com/storj/storj/pull/2090/files#diff-ca06aa73e8e63be362ca22ba3aec2222R68

cmd/gateway/main.go
ifraixedes: I mostly assign return values to the blank identifier when I want to highlight that I'm aware of the returned but I want them to be ignored; that's very important for sensitive return value like `error` where they shouldn't almost be unhandled.

However I don't usually apply it for calls `fmt.Print..` functions, but I did here because in the implementation that I found (and changed in this branch) it used to call the `fmt.Print...` functions assigning the return values to the blank identifier, and I don't see a reason why not to make my changes respecting such style.

cmd/gateway/main.go
```diff
@@ -81,17 +81,17 @@ func init() {
 func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	setupDir, err := filepath.Abs(confDir)
 	if err != nil {
-		return err
+		return Error.Wrap(err)
```
wthorp: This cleanup is awesome.  It would be even more awesome if it were in another PR, to make live easier on reviewers.  :D

ifraixedes: I try to avoid to make more changes than needed, but if the cost of opening a ticket is higher than doing the change like this one, I go ahead.

I usually create new branches for those and separated PRs, however I don't like to spend a lot of time resolving conflicts and if those changes are probably going to conflict with the ones which I was working on, I don't do; in other code sources, being possible to rebase help me a lot on having separated branches and the one which could have conflicts be on top of those, but here rebase aren't allow once the branch is pushed, because it requires to force push, so I don't do.

To wrap, I agree, but there are limitations and only do if it doesn't complicated the work.

NOTE I cannot say if that applies to those changes in this branch, but that's what stop me to act how I did in other code bases.

cmd/gateway/main.go
```diff
@@ -100,24 +100,36 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
 	if !accessKeyFlag.Changed {
 		accessKey, err := generateKey()
 		if err != nil {
-			return err
+			return Error.Wrap(err)
```
kaloyan-raev: It would be better to wrap the error inside `generateKey` instead of here.

cmd/gateway/main.go
```diff
@@ -100,24 +100,36 @@ func cmdSetup(cmd *cobra.Command, args []string) (err error) {
...
 		}
+
 		overrides[secretKeyFlag.Name] = secretKey
 	}
 
-	if !setupCfg.NonInteractive {
-		return setupCfg.interactive(cmd, setupDir, overrides)
+	// override is required because the default value of Enc.KeyFilepath is ""
+	// and setting the value directly in setupCfg.Enc.KeyFiletpathon will set the
+	// value in the config file but commented out.
```
kaloyan-raev: What is "KeyFiletpathon"? typo?

ifraixedes: yes, it's a typo.

PR #2091, opened 2019-05-31
initial value attribution check development

aligeti:
What: 
Support the partner connectorID value attribution check. This implementation will be done in 3 PRs.
**PR#1 shall implement the framework of rounting the partner info from uplink's client to satellite endpoint.**
To implement this following changes shall be made:
**ON CLIENT SIDE:**
1. update the metainfo.proto file to add two new messages: ValueAttributionRequest & ValueAttributionResponse and also a MetainfoClient rpc new interface method "ValueAttributeInfo(){}" shall be added and also corresponding MetainfoServer roc new interface method as well.

**ON SERVER(SATELLITE) SIDE:**
2. update the satellite endpoint to handle the value attribution info request. Currently, it just puts the framework in place.

**PR#2 shall implement the logic of evaluating the value attribution for client's info / request**
This PR effects the changes in the metainfo.go of the satellite side.

**PR#3 shall implement the DB access of valueattriution table logic for reading/creating/deleting** 

**THIS PR address PR#1**

Why:
new feature implementation
https://storjlabs.atlassian.net/browse/V3-1861

Thanks for submitting a PR!


satellite/metainfo/metainfo.go
```diff
@@ -506,3 +509,34 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+
+// checks if bucket has any pointers(entries)
+func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.SegmentInfoRequest) (*pb.SegmentInfoResponse, error) {
+	// not_found error indicates connectory id set but no entry exists
+	resp, err := endpoint.SegmentInfo(ctx, req)
+	return resp, err
+}
+
+// ValueAttributeInfo commits segment metadata
+func (endpoint *Endpoint) ValueAttributeInfo(ctx context.Context, req *pb.SegmentInfoRequest) (*pb.SegmentInfoResponse, error) {
```
mniewrzal: I'm not sure why we need a separate call to metainfo. Why we cannot do this as a part of commiting segment?

also I'm not sure if its a good idea to reuse `pb.SegmentInfoRequest`

satellite/satellitedb/valueattribution.go
```diff
@@ -0,0 +1,65 @@
...
+	return &pb.ConnectorKeyInfo{
+		PartnerId: dbxInfo.PartnerId,
+		BucketId:  dbxInfo.BucketId,
+		CreatedAt: info.CreatedAt,
+	}, nil
+}
+
+// Delete implements partner info
+func (keys *valueattributionDB) Delete(ctx context.Context, id uuid.UUID) (err error) {
+	return nil
```
mniewrzal: why its not implemented?

satellite/metainfo/metainfo.go
```diff
@@ -506,3 +509,34 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+			}
+			return &pb.SegmentInfoResponse{}, nil
+		default:
+			// no entry made into value attribution table
+			endpoint.log.Sugar().Info("Value attribution entry not made: \tpath=", string(req.Path), "\tbucket=", string(req.Bucket), "\terror=", err, "\n")
+			return &pb.SegmentInfoResponse{}, err
+		}
+	}
+
+	return resp, errs.New("BucketID=%v already exists", string(req.GetConnectorKeyInfo().BucketId))
```
mniewrzal: Shouldn't we support case were bucket exists and its empty?

satellite/metainfo/metainfo.go
```diff
@@ -506,3 +509,34 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
 	}
 	return storj.JoinPaths(entries...), nil
 }
+
+// checks if bucket has any pointers(entries)
+func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.SegmentInfoRequest) (*pb.SegmentInfoResponse, error) {
+	// not_found error indicates connectory id set but no entry exists
+	resp, err := endpoint.SegmentInfo(ctx, req)
```
mniewrzal: It doesn't check bucket entries but if segment exists. also don't reuse external endpoint methods, use logic from services.

aligeti:
> I'm not sure if approach is valid. Value attribution should work with each type of libuplink (go, c, java) [#1996 (comment)](https://github.com/storj/storj/pull/1996#discussion_r285801068) and I don't see any integration with this API.

I have just added the uplink's client functions, which call the satellite's endpoint. My understanding is when libuplink is developed you need to call these uplink's client function

pkg/valueattribution/valueattribution.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/satellite/console"
+)
+
+// Error messages
+const (
+	InternalErrMsg         = "It looks like we had a problem on our end. Please try again"
+	NoConnectorIDSetErrMsg = "No ConnectorID set"
```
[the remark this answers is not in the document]
aligeti: https://github.com/storj/storj/blob/f624b213a3ce7063d224964a15a42a08d6b7f104/satellite/console/service.go#L41

I followed the code in the file. 

Will change as suggested.

pkg/valueattribution/valueattribution.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// connectorID is a context value key type
+type connectorID string
```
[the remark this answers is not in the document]
aligeti: I used this file https://github.com/storj/storj/blob/master/satellite/console/auth.go as reference and used the same logic as you mentioned above. Just curious, is there any advantage of having key of type int vs something else?

I started using the context but @coyle suggested to use protobuf to send partner related info instead of context. So I left those functions there just in case if we change direction, otherwise thought of cleaning as a TODO. For now those context functions are not used.
But I will make changes to changing the key type to 'int'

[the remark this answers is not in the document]
mniewrzal: Example you point is also using `int`. Int as key is just more compact.

satellite/metainfo/metainfo.go
aligeti: Thats how I had the implementation done in commitsegment. I talked to Dennis about the DB hits, it is going to create for every segment. So after talking to Dennis, we agreed to make it as an endpoint and my understanding is on the client's side, the uplink logic should be able to handle it. 

Regarding putting in segmentInfo, we will be calling a segmentInfo function to know if any info about the bucket id exists and then save the value attribution, so either I create a separate value attribution message and call segment info or put value attribution info into segment info, it is just a matter of preference...let me know what you ur thoughts are. ... I can create a separate message called value attribution and have to call in segment info in it. let me know what your thoughts are?

satellite/satellitedb/valueattribution.go
aligeti: Initially thought I will do it in a separate PR, but as I was testing. I wasn't quite sure the use case of deletion so I left it as TODO. I will add the delete code

mniewrzal:
Libuplink consumer won't have direct access to `metainfo` client so I'm not sure how will call `ValueAttributeInfo`.

satellite/metainfo/metainfo.go
aligeti: To commit a segment, we need to have a bucket created first. my understanding is calling ValueAttributeInfo() in the bucket creation eliminates lots of redundant checks at a later time of uploads and/or downloads. What do you think?

mniewrzal: With this part I'm asking if we should have ability to assign value attribution to existing but empty bucket and reject bucket that exists but is not empty. With this code we are able to only set value attribution to non existing bucket.

pkg/valueattribution/valueattribution.go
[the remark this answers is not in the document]
aligeti: ok will change to int to be consistent... also we are not using context to pass key values..... should I still keep it until we decide  or remove all together...let me know what your thoughts are?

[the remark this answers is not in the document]
mniewrzal: Removing/leaving context values depends if you can answer my main question. How libuplink user will trigger `ValueAttributeInfo`?

See https://github.com/storj/storj/pull/2091#issuecomment-500821130

satellite/metainfo/metainfo_test.go
```diff
@@ -435,6 +435,49 @@ func TestCommitSegmentPointer(t *testing.T) {
...
+
+		keyInfo := pb.ValueAttributionRequest{
+			PartnerId: []byte("PartnerID"),
+			UserId:    []byte("userID"),
+			BucketId:  []byte("myBucketName"),
+		}
+
+		{
+			// error if pointer is nil
+			_, err = metainfo.CommitSegment(ctx, "bucket", "path", -1, nil, []*pb.OrderLimit2{})
```
mniewrzal: I'm not sure what those tests are doing because `CommitSegment` in both cases is trying to commit object not bucket. If I don't miss anything then we need tests for:
* assign value attribution - error (no bucket)
* create bucket - ok, assign value attribution - ok
* create bucket - ok, upload file - ok, assign value attribution - error
* create bucket - ok, assign value attribution partner A - ok, assign value attribution partner B - error

uplink/metainfo/client.go
```diff
@@ -223,3 +224,21 @@ func (metainfo *Metainfo) ListSegments(ctx context.Context, bucket string, prefi
 
 	return items, response.GetMore(), nil
 }
+
+// ValueAttributeInfo requests the connector key info
+func (metainfo *Metainfo) ValueAttributeInfo(ctx context.Context, bucket string, path storj.Path, segmentIndex int64, partnerID, userID string) (pointer *pb.Pointer, err error) {
```
mniewrzal: Why this method returns `pointer`? Is this needed for partner/connector case?

satellite/metainfo/metainfo.go
```diff
@@ -506,3 +506,21 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+
+// checks if bucket has any pointers(entries)
+func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.ValueAttributionRequest) (*pb.ValueAttributionResponse, error) {
+	//TODO: Logic of checking if bucket exists will be added in new PR.
+	return &pb.ValueAttributionResponse{}, nil
+}
+
+// ValueAttributeInfo commits segment metadata
+func (endpoint *Endpoint) ValueAttributeInfo(ctx context.Context, req *pb.ValueAttributionRequest) (*pb.ValueAttributionResponse, error) {
+	resp, err := endpoint.checkBucketPointers(ctx, req)
```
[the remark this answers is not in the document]
aligeti: added the arguments validation.

uplink/metainfo/client.go
aligeti: Wanted to be returned some value, and did a best guess of returning pointer. On a second thought changed it to more application specific and returned bool.

mniewrzal: Maybe just return empty response, if no error occur it will mean it's ok

aligeti: yep made suggested change

satellite/metainfo/metainfo_test.go
aligeti: added the suggested test case

satellite/metainfo/metainfo_test.go
```diff
@@ -435,6 +435,49 @@ func TestCommitSegmentPointer(t *testing.T) {
...
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 6, UplinkCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		apiKey := planet.Uplinks[0].APIKey[planet.Satellites[0].ID()]
+
+		metainfo, err := planet.Uplinks[0].DialMetainfo(ctx, planet.Satellites[0], apiKey)
+		require.NoError(t, err)
+		projects := planet.Satellites[0].DB.Console().Projects()
+
+		project, err := projects.Insert(ctx, &console.Project{
```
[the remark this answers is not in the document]
aligeti: removed

pkg/pb/metainfo.proto
```diff
@@ -103,4 +104,15 @@ message ListSegmentsResponse {
       
     repeated Item items = 1;
     bool more = 2;
+}
+
+message ValueAttributionRequest{
+    bytes bucket_id = 1;
+    bytes user_id = 2;
```
coyle: I don't think we need the user_id here. We should be pulling the user_id out of the user making the request.

pkg/pb/metainfo.proto
```diff
@@ -103,4 +104,15 @@ message ListSegmentsResponse {
...
+    bytes partner_id = 3;
+    SegmentInfoRequest segment_info_request = 4;
```
coyle: I don't think we need this either

satellite/metainfo/metainfo_test.go
```diff
@@ -435,6 +439,70 @@ func TestCommitSegmentPointer(t *testing.T) {
...
+		{
+			createInfo := storj.CreateObject{
+				RedundancyScheme: redScheme,
+				EncryptionScheme: encScheme,
+			}
+			obj, err := metainfo.CreateObject(ctx, "myBucket", "path", &createInfo)
+			require.NoError(t, err)
+
+			reader := bytes.NewReader([]byte("one fish two fish red fish blue fish"))
+			err = uploadStream(ctx, streams, obj, reader)
```
mniewrzal: To upload something you need only something like this https://github.com/storj/storj/blob/f1641af802667edc594b27a927d38adb281db31e/internal/testplanet/uplink_test.go#L37

pkg/pb/metainfo.proto
```diff
@@ -103,4 +104,15 @@ message ListSegmentsResponse {
...
+}
+
+message ValueAttributionResponse {
+    bool value_attribution = 1;
```
mniewrzal: Why `value_attribution` is needed?

satellite/metainfo/metainfo.go
```diff
@@ -506,3 +506,51 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+
+// checks if bucket has any pointers(entries)
+func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.ValueAttributionRequest) (resp bool, err error) {
+	//TODO: Logic of checking if bucket exists will be added in new PR.
+	// write into value attribution DB only if bucket exists but no segments or no bucket and no segments exits
+	defer mon.Task()(&ctx)(&err)
+
+	keyInfo, err := endpoint.validateAuth(ctx, macaroon.Action{
+		Op:            macaroon.ActionList,
+		Bucket:        req.BucketId,
```
mniewrzal: `Bucket` here is `bucket name` NOT `bucket id`

pkg/pb/metainfo.proto
mniewrzal: In fact we need only `bucket_name` (bucket_id) and `partner_id`. `BucketID` is a `project id` + `bucket name` and partner won't know our internal project id. With api keys and bucket name we are able to find coresponding bucket.

pkg/pb/metainfo.proto
aligeti: protobuf needed some response. so I added this bool . I know error value nil and not nil can be used in stead of it. Removed

satellite/metainfo/metainfo.go
aligeti: if u look at the metainfoClient
func (metainfo *Metainfo) ValueAttributeInfo(ctx context.Context, **bucket string**, path storj.Path, segmentIndex int64, partnerID string) (err error) {
	defer mon.Task()(&ctx)(&err)

	_, err = metainfo.client.ValueAttributeInfo(ctx, &pb.ValueAttributionRequest{
		PartnerId: []byte(partnerID),
		**BucketId:  []byte(bucket),**
	})

	return err
}

The field is named as **'BucketId"** it is actually the Bucket string typed to []byte and renamed. followed the spec.

aligeti: renamed to bucket name

satellite/metainfo/metainfo_test.go
```diff
@@ -435,6 +436,60 @@ func TestCommitSegmentPointer(t *testing.T) {
...
+			err = metainfoClient.ValueAttributeInfo(ctx, "myBucket1", "", -1, string(keyInfo.PartnerId))
+			require.NoError(t, err)
+		}
+		{
+			createInfo := storj.CreateObject{
+				RedundancyScheme: redScheme,
+				EncryptionScheme: encScheme,
+			}
+
+			_, err := metainfo.CreateObject(ctx, "myBucket", "path", &createInfo)
```
mniewrzal: No need to create object and createInfo

aligeti: cleaned up.

PR #2101, opened 2019-06-03
Design Doc Notification System

stefanbenten:
What: Puts up the design doc about my plan how to have a simple notification system in the satellite.

Why: SNO's need more immediate feedback about whats going on!

Thanks for submitting a PR!


stefanbenten:
Still need to add a little bit more concret information how i made the initial PoC, but i appreciate any feedback already on this!

super3:
@stefanbenten Perhaps add the scenario where the node operator can be notified if their node is out of date.

iglesiasbrandon:
@stefanbenten it would be ideal if we could build this email notification system in a way that it could be utilized by the satellites in the SNOBoard in the future. We are planning on adding a notifications center in the SNOBoard but I would love if SNOs had the ability to add their email in the SNOBoard even after they set up their node so that they can receive email notifications from things satellites send out. for example, the content policy on a satellite changes and that satellite operator wants to notify the nodes, or the satellite admin wants to notify nodes payments have been sent for the month, or the satellite admin wants to notify nodes of price changes for bandwidth/ storage. Perhaps we can have a conversation with Viktor to talk about this needs to be built so that they can also utilize it.

stefanbenten:
> @stefanbenten it would be ideal if we could build this email notification system in a way that it could be utilized by the satellites in the SNOBoard in the future. We are planning on adding a notifications center in the SNOBoard but I would love if SNOs had the ability to add their email in the SNOBoard even after they set up their node so that they can receive email notifications from things satellites send out. for example, the content policy on a satellite changes and that satellite operator wants to notify the nodes, or the satellite admin wants to notify nodes payments have been sent for the month, or the satellite admin wants to notify nodes of price changes for bandwidth/ storage. Perhaps we can have a conversation with Viktor to talk about this needs to be built so that they can also utilize it.

This email is available via the KAD information. I am just proposing to reuse exactly that address.
The SNO Board changes the local storage node configuration right? @iglesiasbrandon 
Then thats per default working already. The Function in the doc also outlines, that it can be used in whatever function call across the code (assuming all structs are passed accordingly)

docs/design/notification-system.md
```diff
@@ -0,0 +1,125 @@
...
+	* Sends a notification to the SN with the current status (rather that it entered Containment Mode, or failed the final audit)
+	
+	* In the later of above cases, it will also send the current ratio and its distance to the _disqualification bar_
+
+## DESIGN
+### Functionality
+The existing Service (mailservice) will be extended with templates for all possible notifications we want to send.
+
+### New Service
+A new service called Notificationservice is added which relies on an existing configured mailservice.
```
zeebo: pedantic but can we call it `NotificationService` since it's two words?

stefanbenten: Done, thanks for the review!

stefanbenten:
@wthorp I outlined exactly those messaging thresholds in that design doc. Its a global configuration that defines the maximum amount per hour and day.
Per default, my approach was to send both at the same time (meaning email, if configured, as well as the same message via GRPC). As both are different function calls as outlined, we can add them indepenedently as necessary.

My Goal with this Work is to prepare a framework for notifications which the Kiev Team and anyone else can build up on.

docs/design/notification-system.md
```diff
@@ -0,0 +1,118 @@
...
+### Stefan Benten
+
+
+## Definitions
+
+*SN* - _Storagenode_
+
+*SA* - _Satellite_
+
+## OVERVIEW
```
egonelbre: Use title case to be consistent with other docs.

docs/design/notification-system.md
```diff
@@ -0,0 +1,118 @@
...
+* Not preventing Notifications from every satellite currently, meaning, that you get an offline notification from every satellite you are connected to. (Possible way to mitigate this, is to first check via GRPC, if the node accepts connections)
+
+## SCENARIOS
+
+
+* The SA encounters a failed uptime check of/to SN.
+
+	* It stores the updated metrics as usual in the database. 
+
+	* Before going on, it send a notification to the SN about the change.
```
egonelbre: If there's a failed uptime check then sending a notification will probably fail as well.

PR #2102, opened 2019-06-03
add disqualified check to node selection queries

navillasa:
What:
- adds check to make sure nodes selected aren't disqualified

Why:
- we previously checked to see if audit success ratio and uptime ratio were above some limit to determine "qualification", but now we can just check the bool for `disqualified`

more info in the design draft at #2079 

Thanks for submitting a PR!


satellite/satellitedb/overlaycache.go
```diff
@@ -226,9 +225,9 @@ func (cache *overlaycache) sqliteQueryNodesDistinct(ctx context.Context, exclude
 	rows, err := cache.db.Query(cache.db.Rebind(`SELECT id,
 	type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
 	uptime_ratio, total_audit_count, audit_success_count, total_uptime_count,
-	uptime_success_count
+	uptime_success_count, disqualified
 	FROM (SELECT id, type, address, last_ip, free_bandwidth, free_disk, audit_success_ratio,
-		uptime_ratio, total_audit_count, audit_success_count, total_uptime_count, uptime_success_count,
+		uptime_ratio, total_audit_count, audit_success_count, total_uptime_count, uptime_success_count, disqualified
```
wthorp: needs comma at end

pkg/overlay/statdb_test.go
```diff
@@ -25,6 +25,8 @@ func getRatio(success, total int64) (ratio float64) {
 }
 
 func TestStatDB(t *testing.T) {
+	// TODO(team green): remove the skip when implementing V3-1843
+	t.Skip("require disqualification implementation")
```
ifraixedes: next time add the issue number in the same `skip` comment.
No need to change it in this PR, but I think that's better to easily find a reference if somebody read the code, without having to dig in the git history.

satellite/satellitedb/overlaycache.go
```diff
@@ -41,17 +41,15 @@ func (cache *overlaycache) SelectStorageNodes(ctx context.Context, count int, cr
 	nodeType := int(pb.NodeType_STORAGE)
 
 	safeQuery := `
-		WHERE type = ? AND free_bandwidth >= ? AND free_disk >= ?
+		WHERE type = ? AND NOT disqualified
+			AND free_bandwidth >= ? AND free_disk >= ?
```
ifraixedes: Align the columns of the WHERE clause to have the same indentation.

satellite/satellitedb/overlaycache.go
```diff
@@ -41,17 +41,15 @@ func (cache *overlaycache) SelectStorageNodes(ctx context.Context, count int, cr
 	nodeType := int(pb.NodeType_STORAGE)
 
 	safeQuery := `
-		WHERE type = ? AND free_bandwidth >= ? AND free_disk >= ?
+		WHERE type = ? AND NOT disqualified
```
ifraixedes: Because the rest of the columns of the WHERE clause are one per line, I would also break these 2 one per line and probably leave in the first line the `NOT disqualified`

satellite/satellitedb/overlaycache.go
```diff
@@ -97,12 +95,13 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
 	nodeType := int(pb.NodeType_STORAGE)
 
 	safeQuery := `
-		WHERE type = ? AND free_bandwidth >= ? AND free_disk >= ?
-		  AND total_audit_count < ? AND audit_success_ratio >= ?
+		WHERE type = ? AND NOT disqualified
```
ifraixedes: Same comments in this query, alignment, breaking down each column in a new line and optionally having the `NOT disqualified` as the first condition of the `WHERE` clause.

pkg/overlay/statdb_test.go
navillasa: okay, I'll just go ahead and add the PR number

PR #2103, opened 2019-06-03
makes sure all uplink cli configs get passed to libuplink, add stripeSize

JessicaGreben:
**What/Why:**
While trying to performance test the uplink cli, it appeared that some of the settings in config.yaml weren't getting read in by the code. This PR makes some changes to the uplink related code to fix those issues. The changes include:
- in the `cmd/uplink/cmd/root.go` file, config/cfg variable names are changed to be more explicit so that you can tell when we are referring to the uplink cli config versus the libuplinkConfig 
- some of the uplink cli flags from the config.yaml file were not getting passed to libuplink code, for example the configs in `cmd/uplink/cmd/mb.go`, now hopefully they all are
- Fix the encryption block size/stripeSize check and also add comments hoping its easier for others to understand
- remove the uplink config option `enc.block-size` since that should be calculated by redundancy scheme settings ShareSize and MinRequired
- Update tests


cmd/uplink/cmd/root.go
```diff
@@ -109,10 +105,10 @@ func (c *UplinkFlags) GetProject(ctx context.Context) (*libuplink.Project, error
 }
 
 // GetProjectAndBucket returns a *libuplink.Bucket for interacting with a specific project's bucket
-func (c *UplinkFlags) GetProjectAndBucket(ctx context.Context, bucketName string, access libuplink.EncryptionAccess) (project *libuplink.Project, bucket *libuplink.Bucket, err error) {
-	project, err = c.GetProject(ctx)
+func (cliCfg *UplinkFlags) GetProjectAndBucket(ctx context.Context, bucketName string, access libuplink.EncryptionAccess) (project *libuplink.Project, bucket *libuplink.Bucket, err error) {
+	project, err = cliCfg.GetProject(ctx)
 	if err != nil {
-		return nil, nil, err
+		return project, bucket, err
```
VinozzZ: I have seen pre-defined return values in other places in the code. Is there a convention that we would always return the pre-defined variables vs we would be explicit to return `nil`?

JessicaGreben: im not sure if there is a convention. Personally I just like using the name of the return value if it has a name since its more clear what "nil" is for and saves me a step trying to figure out what nil is suppose to be

lib/uplink/project.go
```diff
@@ -93,23 +94,46 @@ func (cfg *BucketConfig) setDefaults() {
...
+	shareSize := cfg.Volatile.RedundancyScheme.ShareSize
+	requiredShares := int32(cfg.Volatile.RedundancyScheme.RequiredShares)
+	blockSize := cfg.EncryptionParameters.BlockSize
+
+	// number of bytes needed to recreate a stripe
+	stripeSize := shareSize * requiredShares
+
+	if blockSize%stripeSize != 0 {
+		return fmt.Errorf("encryption BlockSize (%d) must be a multiple of RS ShareSize (%d) * RS RequiredShares (%d)",
+			shareSize, requiredShares, blockSize,
```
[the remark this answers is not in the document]
JessicaGreben: oops good catch

pkg/metainfo/kvmetainfo/objects.go
```diff
@@ -105,14 +107,27 @@ func (db *DB) CreateObject(ctx context.Context, bucket string, path storj.Path,
...
+
+		// BlockSize should be an integer multiple of stripe size.
+		// If the provided EncryptionScheme.BlockSize isn't a multiple of the
+		// DefaultRS stripeSize, then overwrite the EncryptionScheme with the DefaultES values.
+		stripeSize := int32(DefaultRS.RequiredShares) * DefaultRS.ShareSize
+		if info.EncryptionScheme.BlockSize%stripeSize != 0 {
+			fmt.Printf(`encryption BlockSize (%d) must be a multiple of RS ShareSize (%d) * RS RequiredShares (%d).
+			Overwriting Blocksize with kvmetainfo.DefaultES.BlockSize (%d)`,
+				info.EncryptionScheme.BlockSize, DefaultRS.RequiredShares,
+				DefaultRS.ShareSize, DefaultES.BlockSize,
```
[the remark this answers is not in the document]
JessicaGreben: wow i really got all these wrong lol

pkg/metainfo/kvmetainfo/buckets.go
```diff
@@ -48,6 +49,10 @@ func (db *Project) CreateBucket(ctx context.Context, bucketName string, info *st
 		info.SegmentsSize = db.segmentsSize
 	}
 
+	if err := validateBlockSize(info.RedundancyScheme, info.EncryptionParameters.BlockSize); err != nil {
```
VinozzZ: Just curious on the reason why we moved this function into `kvmetainfo`. More organized code?

JessicaGreben: Good question. I guess two reasons: 
1) `validateBlockSize()` was happening in two different packages and I wondered if I could move it to a shared package
2) I realized that the flow of data as it relates to metainfo goes like this:
```
libuplink -> pkg/metainfo/kvmetainfo -> pkg/storage -> network layer/grpc -> SA metainfo endpoint -> SA pointerDB
```
Previously the `validateBlockSize` check was happening in two places:
1. `lib/uplink` when we created a bucket and 
2. `pkg/metainfo/kvmetainfo` when we created an object

This meant that `validateBlockSize` was occurring in two different layers, the highest layer `lib/uplink` and the next layer down in `pkg/metainfo/kvmetainfo`. So I figured we could move the check to the same layer `pkg/metainfo/kvmetainfo`.

pkg/metainfo/kvmetainfo/objects.go
```diff
@@ -40,9 +40,10 @@ var DefaultRS = storj.RedundancyScheme{
 }
 
 // DefaultES default values for EncryptionScheme
+// BlockSize should default to the size of a stripe
 var DefaultES = storj.EncryptionScheme{
 	Cipher:    storj.AESGCM,
-	BlockSize: 1 * memory.KiB.Int32(),
+	BlockSize: DefaultRS.StripeSize(),
```
jtolds: delayed comment but might be worth making this maybe double the stripe size? every encryption block gets some padding overhead. for a 1MB file, with a 1K blocksize, we end up actually storing 1MB + 16K. If we do a 2*DefaultRS.StripeSize() (where StripeSize() is 1K), then the overhead is only 8K instead of 16K.

JessicaGreben: i'm have a different branch with some changes that relate to this.  I was thinking since the `BlockSize` config option is removed, that I would replace it with a `StripesPerBlock` config setting like so ...

```
[...]
+      StripesPerBlock int    `help:"the integer of how many stripes should fit in an encryption block" default:"2"
````

jtolds: oh that's a neat idea

JessicaGreben: orrr we can just hardcode 2 stripes per block and call it good.  Im cool with either.

jtolds: shrug emoji

jtolds: i think we should default to a multiple, but i don't care strongly more than that about if it's hardcoded or configurable or what. i do think it's valuable to not have people assume it's just the stripe size always

JessicaGreben: I prefer fewer config options when possible, so if you are /shrug about it, i will hardcode it.  I will gete a PR up with that change in a bit here and will run it by you.

jtolds: i am /shrug about it

PR #2112, opened 2019-06-04
V3-1819 Storage node operator server and service started

crawter:
What: 
This is very initial backend for storagenode operator dashboard (SNOBoard). 
The only thing it does right now - just returning static files with our markup.
All static files will be included in our binary. To achieve this - we generate this binaries from static files via go-bindata.

Why:
We really need this PR to be reviewed and merged to unblock our frontend developers, that are almost created markup for storagenode operator

Thanks for submitting a PR!


web/operator/dist/public/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
stefanbenten: Same as above. It feels very fragile, if SNO's have to download a binary and a dist folder..

cmd/storj-sim/network.go
```diff
@@ -414,9 +414,19 @@ func newNetwork(flags *Flags) (*Processes, error) {
 			process.WaitForStart(satellite)
 		}
 
+		// TODO: reuse for satellite and storagenode operator
+		// TODO: find source file, to set static path
+		_, filename, _, ok := runtime.Caller(0)
+		if !ok {
+			return nil, errs.Combine(processes.Close(), errs.New("no caller information"))
+		}
+		storjRoot := strings.TrimSuffix(filename, "/cmd/storj-sim/network.go")
```
stefanbenten: Could we put these files inline/ dont require external paths?

rikysya: that's for running with `storj-sim only`, but, yeah, we won't need if we bundle assets with the binary

stefanbenten: But the same needs to be applied in the storagenode binary directly, right?

rikysya: as `storj-sim` using pre-built binaries, we won't these lines if we bundle assets directly to the binary as there won't be any path for resources at all. This code originally was for `satellite` running with `storj-sim`, to bypass setting config values for external resources to be able to run `storj-sim`(which assumes in some way your building from sources)

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,117 @@
...
+		log.Error(err)
+		return
+	}
+}
+
+// appHandler is web app http handler function
+func (s *Server) appStaticHandler(w http.ResponseWriter, req *http.Request) {
+	resourceName := strings.TrimPrefix(req.RequestURI, "/static/")
+
+	data, err := Asset(s.config.StaticDir + resourceName)
```
rikysya: why do u need static dir here? doesn't actual request will something like `/static/resourcename`?

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/prometheus/common/log"
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+
+	"storj.io/storj/storagenode/operator"
+)
+
+// TODO: improve embedded resources generation
+//go-bindata -pkg operatorserver -o storagenode/operator/operatorserver/static.go web/operator/dist web/operator/dist/public/
```
stefanbenten: I like that soo much!

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Error is storagenode operator web error type
+var Error = errs.Class("storagenode operator web error")
+
+// Config contains configuration for storagenode operator web server
+type Config struct {
+	Address   string `help:"server address of the graphql api gateway and frontend app" default:"127.0.0.1:14002"`
+	StaticDir string `help:"path to static resources" default:""`
```
stefanbenten: I think we should set the default to `web/operator` here, to avoid problems in the static.go file

storagenode/operator/operatorserver/server.go
stefanbenten: This could also be a Make Target. that we add to our release process.
That should then include the most recent assets into our binary.

egonelbre: Oh, nooo... this will bloat our github repository. I've seen repositories end up as 2GB as an end result.

cmd/storj-sim/network.go
```diff
@@ -420,6 +420,7 @@ func newNetwork(flags *Flags) (*Processes, error) {
 		process.Arguments = withCommon(process.Directory, Arguments{
 			"setup": {
 				"--identity-dir", process.Directory,
+				"--operator.address", net.JoinHostPort(host, port(storagenodePeer, i, publicHTTP)),
```
egonelbre: Let's call it `console` or `dashboard` rather than `operator`.

storagenode/operator/operatorserver/static.go
```diff
@@ -0,0 +1,125 @@
+// Code generated by go-bindata. DO NOT EDIT.
+
+package operatorserver
```
egonelbre: No embedded data, it will end up bloating our repository. This should be embedded at build stage not here.

storagenode/operator/diskspace.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package operator
```
egonelbre: Maybe `dashboard` or `console`... calling it `operator` seems weird for some reason.

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Close closes server and underlying listener
+func (s *Server) Close() error {
+	return s.server.Close()
+}
+
+// appHandler is an entry point for storagenode operator web interface
+func (s *Server) appHandler(w http.ResponseWriter, req *http.Request) {
+	data, err := Asset(s.staticDir + "dist/public/index.html")
```
egonelbre: Don't use addition to create filepaths use `filepath`.

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	_, err = w.Write(data)
+	if err != nil {
+		log.Error(err)
```
egonelbre: Don't use global logger.

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+// appStaticHandler is needed to return static resources
+func (s *Server) appStaticHandler(w http.ResponseWriter, req *http.Request) {
+	resourceName := strings.TrimPrefix(req.RequestURI, "/static/")
+
+	data, err := Asset(s.staticDir + resourceName)
+	if err != nil {
+		log.Error(err)
```
egonelbre: Don't use global logger.

storagenode/operator/operatorserver/server.go
```diff
@@ -0,0 +1,118 @@
...
+		return
+	}
+
+	_, err = w.Write(data)
+	if err != nil {
+		log.Error(err)
```
egonelbre: Don't use global logger.

storagenode/operator/service.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/memory"
+	"storj.io/storj/storagenode/bandwidth"
+	"storj.io/storj/storagenode/pieces"
+)
+
+// Service is handling storage node operator related logic
```
egonelbre: Why do we need this service when we already have `bandwidth`, `pieces` and other services for this logic.

crawter: @egonelbre just to separate our business logic from api layer
operator.Service uses bandwidth, pieces, etc to combine all data together

storagenode/console/service.go
```diff
@@ -0,0 +1,100 @@
...
+type Service struct {
+	log *zap.Logger
+
+	bandwidth bandwidth.DB
+	pieceInfo pieces.DB
+	kademlia  *kademlia.Kademlia
+
+	allocatedBandwidth memory.Size
+	allocatedDiskSpace memory.Size
+	walletNumber       string
```
egonelbre: `walletAddress`

storagenode/console/service.go
```diff
@@ -0,0 +1,100 @@
...
+// Service is handling storage node operator related logic
+type Service struct {
+	log *zap.Logger
+
+	bandwidth bandwidth.DB
+	pieceInfo pieces.DB
+	kademlia  *kademlia.Kademlia
+
+	allocatedBandwidth memory.Size
+	allocatedDiskSpace memory.Size
```
egonelbre: *monitor.Service has the necessary information already, so rather ask that service.

storagenode/console/service.go
```diff
@@ -0,0 +1,100 @@
...
+		bandwidth:          bandwidth,
+		pieceInfo:          pieceInfo,
+		kademlia:           kademlia,
+		allocatedBandwidth: allocatedBandwidth,
+		allocatedDiskSpace: allocatedDiskSpace,
+		walletNumber:       walletNumber,
+		upTime:             time.Now(),
+	}
+
+	return &service, nil
```
egonelbre: you can inline the previous statement

storagenode/console/service.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return FromUsage(usage, s.allocatedBandwidth.Int64())
+}
+
+// GetUsedStorage returns all info about storagenode disk space usage
+func (s *Service) GetUsedStorage(ctx context.Context) (*DiskSpaceInfo, error) {
+	spaceUsed, err := s.pieceInfo.SpaceUsed(ctx)
```
egonelbre: `*monitor.Service` can tell you this information.

storagenode/operator/operatorserver/server.go
stefanbenten: @egonelbre How do you want to install and release binaries without that?

egonelbre: Created an issue to discuss this: https://storjlabs.atlassian.net/browse/V3-2002

web/operator/dist/build.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (C) 2019 Storj Labs, Inc.
```
egonelbre: Is there a reason this folder is called `operator` and not `storagenode`?

storagenode/console/service.go
crawter: @egonelbre but, as i understand, the main purpose of monitoring service - is to update kademlia in loop. It is separate service with defined responsibility.

In our case we created new service, main responsibility of which - is to return data to our web ui

Won't we break SRP if we will use monitor for this purpose? 

Moreover, we have some specific requirements in SNO operator. And we will have more than one reason to modify monitoring service in future each time when we will receive new specifications for SNO dashboard

egonelbre: Monitoring service is responsible for monitoring the storage node disk usage, bandwidth etc. and updating kademlia.

I didn't mean that `console.Service` should be removed, but rather that it can rely `monitor.Service` and other services for the information about the disk usage / bandwidth information. That way we don't duplicate the logic and accidentally end up displaying different information for dashboard and the console.

storagenode/console/service.go
crawter: @egonelbre - also storagenode/inspector/inspector.go has such functionality. But, as console/service and monitoring/service they serves for different purposes

egonelbre: Inspector in this case is an Endpoint, so only for responding to network requests.

storagenode/console/service.go
egonelbre: But I get your point, the responsibilities at the moment are a bit fuzzy.

web/operator/dist/build.js
crawter: @egonelbre this folder is created only for testing purposes. @NikolaiYurchenko  and @VitaliiShpital  are responsible for frontend part and they used exactly this name as you mentioned. But they are blocked by this pull request

storagenode/operator/operatorserver/server.go
crawter: @egonelbre truly to say - we have an SPA with minified js bundle ~ 1mb and a few mb of media resources, that also could be minified with FFMPEG. I don't think that we will have more that 5mb of data

@stefanbenten from the other side - storagenode operator - is a local server. I don't think that we will have problems with not embedded static files there

storagenode/console/service.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if pieceInfo == nil {
+		return nil, errs.New("pieceInfo can't be nil")
+	}
+
+	if version == nil {
+		return nil, errs.New("version can't be nil")
+	}
+
+	if kademlia == nil {
```
egonelbre: This could be condensed into a single check since the information that something is missing should be sufficient to figure out what is the problem.

storagenode/operator/operatorserver/server.go
egonelbre: This will be eventually committed 1000x or more times and minified files to compress.. so if 5MB it'll end up as 5GB.

crawter: @egonelbre 
By the way - i've just checked and all assets and js from deployed satellite console weights ~1.8 mb
Node operator will be much more lightweight

jtolds: i agree with egon - if we check in bundles that are generated from the code, then every time we change a line in the relevant parts of the code it's a new couple megabytes. it does balloon quickly

PR #2113, opened 2019-06-04
pkg/bloomfilter: implementation and benchmark results

Fadila82:
The benchmark will not be merged. It can be found as a gist here: https://gist.github.com/Fadila82/9f54c61b5f91f6b1a6f9207dfbb5dd2d

What: 
Benchmark of different filters library

If we assume that a typical storage nodes has 2 TB capacity, and a typical piece is ~2 MB, this means that a such node with a full hard drive would have 1 million pieces.
We're testing the behavior with 1 million pieces

For a beginning, we will  generate 1 million piece ids, add 95% of them to the bloom filter, and then check all 1 million piece ids with the bloom filter. Let's measure times, memory allocation, false positives, etc.

Why:
To decide which filter we should use

Thanks for submitting a PR!


pkg/filters/filter_test.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}()
+	Init()
+
+	names := []string{"Zeebo", "Willf", "Steakknife", "Custom"}
+
+	_, err = file.WriteString("# p\t")
+	if err != nil {
+		b.Log(err.Error())
+		b.Fail()
```
kaloyan-raev: We can replace `b.Log` and `b.Fail` with just `b.Fatal(err)`.

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,68 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
...
+### Steakknife's bloom filters
+- Parameters:
+    - **maxElements**: max number of elements in the set
+    - **p**: probability of false positive
+- Serialization available
+- murmur3 hash function
+
+### Custom bloom filter
+- Parameters:
+    - **maxElements**: max number of elements in the set
```
kaloyan-raev: Does this mean that when a bloom filter is created, the caller needs to know that maximum possible number of element that will be added to the filter?

If this is the case then I am not sure if this will work well with the satellite. Initially, we will make the data repair checker to create the bloom filters as part of its usual work. I am not sure if it knows how many pieces are stored on each storage node.

Is it possible to construct the custom bloom filter in a way that the `maxElements` is not known?

Fadila82: I thought we would know the bloom filter size beforehand. The custom bloom filter does not handle that (because then we cannot compute the optimal values for k and m), I guess in this case using Zeebo's implementation would make more sense.

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,68 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
...
+Three Bloom filter implementations are considered:
+- **BF1**: Zeebo's bloom filters (github.com/zeebo/sbloom)
+- **BF2**: Willf's Bloom filters (github.com/zeebo/sbloom)
+- **BF3**: Steakknife's Bloom filters (github.com/golang/leveldb/bloom)
+- **BF4**: Custom bloom filter
+
+### Zeebo's bloom filters
+-Parameters:
+    - **k**: The Bloom filter will be built such that the probability of a false positive is less than (1/2)**k
+    - **h**: hash functions
```
kaloyan-raev: This does not render well in markdown: `(1/2)**k - **h**`. I am not sure what the exact formula is.

docs/design/garbage-collection.md
egonelbre: We can keep the last number around and then use 2x of that number for the next filter.

kaloyan-raev: What happens if we add more than `maxElements` to the filter. Will it become corrupted?

egonelbre: No, no corruption, just the false positive rate will be terrible.

kaloyan-raev: So, let's say, if we initialize the bloom filter with 1000 max elements, but add 2000, it would still be worth sending it to the storage node. It will be just less effective, right?

egonelbre: It will be less effective and at some point it would make sense to just drop it for that iteration.

egonelbre: e.g. if you have FP=2%, and 1M items, then inserting 2M will give you FP=14%, with 3M the FP=33%

docs/design/garbage-collection.md
Fadila82: I've changed it.

docs/design/garbage-collection.md
kaloyan-raev: OK. Cool. So we can actually just use the last known piece count for this node from the last iteration. It will just result in higher FP if it happens that this node stores more pieces. No need to waste memory with 2x last known piece count.

docs/design/garbage-collection.md
kaloyan-raev: Thanks!

docs/design/garbage-collection.md
egonelbre: Yeah, that would work as well.

egonelbre:
Note for merging, it's probably sufficient just to keep one implementation. If we need to we can make a separate repository for keeping these sorts of benchmarks.

ifraixedes:
> Note for merging, it's probably sufficient just to keep one implementation. If we need to we can make a separate repository for keeping these sorts of benchmarks.

That's a good idea, however what's the problem of merging it, the extra dependencies?

egonelbre:
Yeah, the extra dependencies.

ifraixedes:
> Yeah, the extra dependencies.

Are they a problem if they were only used in  a test? at the end they shouldn't be used when building non-test binaries.

pkg/filters/filter.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Filter is an interface for filters
+type Filter interface {
+	Contains(pieceID []byte) bool
+	Add(pieceID []byte)
+	Encode() []byte
+}
+
+// CuckooFilter is a cuckoo filter
+type CuckooFilter struct {
```
ifraixedes: I'll move this implementation to a different file, for example to the one currently named _bloom_filters.go_ and then rename it to something like _external_filters.go_ or _third_party_filters.go_.

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,185 @@
...
+	if !initDone {
+		totalNbPieces = 1000000
+		nbPiecesInFilter = 950000
+		pieceIDs = GenerateIDs(totalNbPieces)
+		initDone = true
+		falsePositiveProbability = 0.1
+	}
+}
+
+func benchmarkAdd(filter Filter, pieceIDs [][]byte, b *testing.B) {
```
ifraixedes: I usually see that this kind of helper functions to get `b *testing.B` as a first parameter. 
Also add `b.Helper()` call.

These apply for the rest of the helper functions.

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,185 @@
...
+var nbPiecesInFilter int
+var totalNbPieces int
+var falsePositiveProbability float64
+
+//  generates 1 million piece ids
+// adds 95% of them to the bloom filter,
+// and then checks all 1 million piece ids with the bloom filter
+// measure times, memory allocation, false positives
+
+func Init() {
```
ifraixedes: There is no need to be capitalized.

However we can achieve what `Init` does using a [`TestMain`](https://golang.org/pkg/testing/#hdr-Main) function and avoid any potential data race and also influence in the first bench march function which runs and call `Init`.

```go
func TestMain(m *testing.M) {                                                                                              
  totalNbPieces = 1000000                                                                                                  
  nbPiecesInFilter = 950000                                                                                                
  pieceIDs = GenerateIDs(totalNbPieces)                                                                                    
  initDone = true                                                                                                          
  falsePositiveProbability = 0.1                                                                                           
  os.Exit(m.Run())                                                                                                         
} 
```

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,185 @@
...
+	falsePositiveP := float64(nbIn) / float64(len(pieceIDs[nbPiecesInFilter:]))
+	if falsePositiveP > falsePositiveProbability {
+		b.Log("False positive ratio: ", falsePositiveP, " - greater than expected :", falsePositiveProbability)
+	}
+	b.Log("False positive ratio: ", falsePositiveP)
+	return falsePositiveP
+}
+
+func BenchmarkInit(b *testing.B) {
+	Init()
```
ifraixedes: Do we need to benchmark `Init()`?

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,185 @@
...
+	if nbIn < nbPiecesInFilter {
+		// we have a false negative - it should not happen
+		b.Fatal("False negative!")
+	}
+	nbIn = benchmarkContains(filter, pieceIDs[nbPiecesInFilter:], b)
+	falsePositiveP := float64(nbIn) / float64(len(pieceIDs[nbPiecesInFilter:]))
+	if falsePositiveP > falsePositiveProbability {
+		b.Log("False positive ratio: ", falsePositiveP, " - greater than expected :", falsePositiveProbability)
+	}
+	b.Log("False positive ratio: ", falsePositiveP)
```
ifraixedes: Shouldn't be this in an `else` clause?

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, pieceID := range pieceIDs {
+		if filter.Contains(pieceID) {
+			nbPiecesIn++
+		}
+	}
+	return
+}
+
+func benchmarkFilter(filter Filter, pieceIDs [][]byte, b *testing.B) (p float64) {
+	b.ReportAllocs()
```
ifraixedes: I think that this should be controlled by the proper Bench function rather than the helper one.

egonelbre:
> Are they a problem if they were only used in a test? at the end they shouldn't be used when building non-test binaries.

Yes, they are still downloaded when somebody uses our code as a library.

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,68 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
     - Storage node should be able to receive a Delete request from a Satellite
 - Storage node should use the filter from the Delete request to decide which pieces to delete, then delete them
 - Eventually, this service should iterate over a db snapshot instead of being integrated with the data repair checker
+
+## Bloom filters benchmark
```
egonelbre: Let's add a link to the gist.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package filters
```
egonelbre: Wondering whether there's a better name for it. But, maybe just `bloom` or `bloomfilter` since it doesn't contain any other filters. `liveset`, `liveness` also come to mind.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"encoding/binary"
+	"math"
+	"math/big"
+)
+
+// CustomFilter is a custom filter
+type CustomFilter struct {
```
egonelbre: `Filter` since there are no other implementations here anymore.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+	seed              int
+	k                 int
+	nbBitsPerElements int
+	table             []byte
+}
+
+// NewCustomFilter returns a new custom filter
+func NewCustomFilter(nbElements int, p float64) *CustomFilter {
```
egonelbre: `nbElements` to `expectedElements`

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/binary"
+	"math"
+	"math/big"
+)
+
+// CustomFilter is a custom filter
+type CustomFilter struct {
+	seed              int
+	k                 int
+	nbBitsPerElements int
```
egonelbre: `bitsPerElement` or `size`

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+	table             []byte
```
egonelbre: Wondering whether there is a benefit from using `[]uint64`

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+	cf.seed = 8 // TODO allow another seed value
+	cf.nbBitsPerElements = int(-1.44*math.Log2(p)) + 1
+	cf.k = int(float64(cf.nbBitsPerElements)*math.Log(2)) + 1
+	m := (nbElements * cf.nbBitsPerElements) // total number of bits in the array
+	cf.table = make([]byte, m/8)
+	return &cf
+}
+
+// Add adds an element to the bloom filter
+func (cf *CustomFilter) Add(pieceID []byte) {
```
egonelbre: Use the concrete type `storj.PieceID` here.

pkg/filters/filter.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package filters
+
+// Filter is an interface for filters
+type Filter interface {
```
egonelbre: I don't think we need the interface here.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+// Encode encodes the filter
+func (cf *CustomFilter) Encode() []byte {
+	filterInfos := append(int2bytes(cf.seed), int2bytes(cf.nbBitsPerElements)...)
+	filterInfos = append(filterInfos, int2bytes(cf.k)...)
+	return append(filterInfos, cf.table...)
+}
+
+func int2bytes(num int) (b []byte) {
+	b = make([]byte, 4)
+	binary.BigEndian.PutUint32(b, uint32(num))
```
egonelbre: Wondering whether uint32 is sufficent to keep the number of pieces on a storage node.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package filters
+
+import (
+	"encoding/binary"
+	"math"
+	"math/big"
```
egonelbre: This can be done without using `big.Int`... using `big.Int` in hot-path is complicated.

pkg/filters/custom_filter.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// CustomFilter is a custom filter
+type CustomFilter struct {
+	seed              int
+	k                 int
```
egonelbre: I would hardcode `k=4` because it's good enough for our FP rate. Mainly for performance.

egonelbre:
@Fadila82 is this just about the benchmark and not the actual implementation? And the actual implementation will come later? We can merge the analysis now and make a separate PR for the implementation.

ifraixedes:
> Yes, they are still downloaded when somebody uses our code as a library.

Yes, downloaded yes, but I don't see it as a big deal for now.
Obviously, if we can have a repo for those today/tomorrow, it's OK, but delaying for so long isn't good, so I would prefer to merge it and moved out when possible than leaving it open for many days.

pkg/filters/custom_filter.go
ifraixedes: `live**` variants doesn't mean that much to me, perhaps something that involves `sn`, `pieces`, otherwise `bloomfilter` looks better.

pkg/filters/filter.go
ifraixedes: It's used for the benchmark however, I agree, it should be moved to the `_tet.go` file or being unexported.

I prefer the first option.

Fadila82:
> @Fadila82 is this just about the benchmark and not the actual implementation? And the actual implementation will come later? We can merge the analysis now and make a separate PR for the implementation.



> @Fadila82 is this just about the benchmark and not the actual implementation? And the actual implementation will come later? We can merge the analysis now and make a separate PR for the implementation.

Well, I did it thinking about the benchmark. It's the start of the implementation, but was not meant to the final. It was mainly to test using the piece id as the hash function, and it can be greatly optimized I think.
Merging the analysis and make a separate PR for the implementation sounds good to me. I would then put all the files in a gist (including the custom filter), and create a PR for later with the changes you suggest for the for-now called "CustomFilter". Should I do that?

kaloyan-raev:
There is already a good code review discussion on the implementation going on in this PR. So maybe let's finish it. Opening a separate PR will break the discussion started here. Merging the PR without an actual implementation won't help us making progress on the Garbage Collection topic.

pkg/filters/custom_filter.go
kaloyan-raev: It should be good for storing a few PB of data on the storage node. If not enough, then the bloom filter for such node would be larger than 4 GB. I am not sure how practical this is. If we hit this limit then we should be looking for alternatives like partial GC instead of just increasing the size of the bloom filter.

pkg/filters/filter.go
```diff
@@ -1,11 +1,98 @@
...
+	seed           int
+	k              int
+	bitsPerElement int
+	table          []byte
+}
+
+// NewFilter returns a new custom filter
+func NewFilter(expectedElements int, p float64) *Filter {
+	var f Filter
+	f.seed = 8 // TODO allow another seed value
```
ifraixedes: how it should be set by a function parameter? if that's the case why not to add it now? otherwise it isn't a TODO for me.

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,74 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
...
+
+Three Bloom filter implementations are considered:
+- **BF1**: Zeebo's bloom filters (github.com/zeebo/sbloom)
+- **BF2**: Willf's Bloom filters (github.com/zeebo/sbloom)
+- **BF3**: Steakknife's Bloom filters (github.com/golang/leveldb/bloom)
+- **BF4**: Custom bloom filter
+
+### Zeebo's bloom filters
```
ifraixedes: I forgot to comment in the previous review, we could link this titles to the repos for being more exact what ones are referring for.

pkg/filters/filter.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package bloomfilter
```
ifraixedes: Rename the `pkg/filter` directory to `pkg/bloomfilter`

pkg/filters/filter.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"encoding/binary"
+	"math"
+	"math/big"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Filter is a custom filter
```
ifraixedes: `// Filter is a bloom filter implementation`

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,74 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
...
+|0.13|	7340620|	0.12|	530832|	    0.11|	530904|	0.12|	625012|	0.08|
+|0.14|	7340620|	0.12|	511552|	    0.12|	511624|	0.13|	625012|	0.08|
+|0.15|	7340620|	0.12|	493600|	    0.14|	493672|	0.14|	500012|	0.16|
+|0.16|	7340620|	0.12|	476816|	    0.15|	476888|	0.15|	500012|	0.16|
+|0.17|	7340620|	0.12|	461040|	    0.15|	461112|	0.16|	500012|	0.16|
+|0.18|	7340620|	0.12|	446168|	    0.17|	446240|	0.17|	500012|	0.16|
+|0.19|	7340620|    0.12|	432104|	    0.18|	432176|	0.18|	500012|	0.16|
+|0.20|	7340620|	0.12|	418760|	    0.19|	418832|	0.19|	500012|	0.16|
+
+The benchmark code is available as a gist [here](https://gist.github.com/Fadila82/9f54c61b5f91f6b1a6f9207dfbb5dd2d).
```
ifraixedes: I would add a `go.mod` file to the gist to freeze the used version.

pkg/filters/filter_test.go
```diff
@@ -0,0 +1,52 @@
...
+//  generates 1 million piece ids
+// adds 95% of them to the bloom filter,
+// and then checks all 1 million piece ids with the bloom filter
+
+func TestMain(m *testing.M) {
+	totalNbPieces = 1000000
+	nbPiecesInFilter = 950000
+	pieceIDs = GenerateIDs(totalNbPieces)
+	falsePositiveProbability = 0.1
+	m.Run()
```
ifraixedes: this must be `os.Exit(m.Run())`

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,74 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
     - Storage node should be able to receive a Delete request from a Satellite
 - Storage node should use the filter from the Delete request to decide which pieces to delete, then delete them
 - Eventually, this service should iterate over a db snapshot instead of being integrated with the data repair checker
+
+## Bloom filters benchmark
+
+Three Bloom filter implementations are considered:
+- **BF1**: Zeebo's bloom filters (github.com/zeebo/sbloom)
+- **BF2**: Willf's Bloom filters (github.com/zeebo/sbloom)
```
wthorp: bad link?

pkg/filters/filter.go
Fadila82: it's because it was for benchmarking alone. I did not take the time to handle the case where the seed is higher (we can then reach the end of the byte array representing the piece id, and have to loop over it)

ifraixedes: I would add a short warning in the comment, otherwise it's very tentative to just add a new parameter to the function and set it without doing any other change.

On the other hand, if for now we don't plan to change it or making it configurable, I would remove the `TODO` and just add the mentioned warning.

kaloyan-raev:
@Fadila82 could you change the PR title to reflect that it includes the actual bloom filter implementation, not just benchmarks?

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,85 @@
...
+	f.seed = 8 // TODO allow another seed value - have to handle the reach of the end of the array
+	f.bitsPerElement = int(-1.44*math.Log2(p)) + 1
+	f.k = int(float64(f.bitsPerElement)*math.Log(2)) + 1
+	m := (expectedElements * f.bitsPerElement) // total number of bits in the array
+	f.table = make([]byte, m/8)
+	return &f
+}
+
+// Add adds an element to the bloom filter
+func (cf *Filter) Add(pieceID storj.PieceID) {
```
ifraixedes: I didn't see it yesterday, but `cf` isn't a good name.
Following our [current conventions]https://github.com/storj/docs/blob/master/code/Style.md#variable-naming) my guess is that it has to be `filter`

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,85 @@
...
+	offsetAsByteArray := pieceID[cf.seed : cf.seed+cf.k]
+	offset := binary.BigEndian.Uint64(append(make([]byte, len(offsetAsByteArray)), offsetAsByteArray...))
```
ifraixedes: do we actually need to pass a copy of the slice here?

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Encode encodes the filter
+func (cf *Filter) Encode() []byte {
+	filterInfos := append(int2bytes(cf.seed), int2bytes(cf.bitsPerElement)...)
+	filterInfos = append(filterInfos, int2bytes(cf.k)...)
+	return append(filterInfos, cf.table...)
+}
+
+func int2bytes(num int) (b []byte) {
```
ifraixedes: I wouldn't use a named return here

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,85 @@
...
+		cf.table[byteIndex] |= 0x1 << bitIndex
+		i++
+	}
+
+}
+
+// Contains return true if pieceID may be in the set
+func (cf *Filter) Contains(pieceID storj.PieceID) bool {
+	offsetAsByteArray := pieceID[cf.seed : cf.seed+cf.k]
+	offset := binary.BigEndian.Uint64(append(make([]byte, len(offsetAsByteArray)), offsetAsByteArray...))
```
ifraixedes: Same question than above, do we need to pass a copy?

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,141 @@
...
+	"storj.io/storj/pkg/storj"
+)
+
+const (
+	version1 = 1
+)
+
+// rangeOffsets contains offsets for selecting subranges
+// that minimize overlap in the first hash functions
+var rangeOffsets = [...]byte{9, 13, 19, 23}
```
egonelbre: visualization for the range offsets: https://play.golang.org/p/iRRULODRyL5

pkg/bloomfilter/filter.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+// NewOptimal returns a filter based on expected element count and false positive rate.
+func NewOptimal(expectedElements int, falsePositiveRate float64) *Filter {
+	seed := byte(rand.Intn(255))
+
+	// calculation based on https://en.wikipedia.org/wiki/Bloom_filter#Optimal_number_of_hash_functions
+	bitsPerElement := int(-1.44*math.Log2(falsePositiveRate)) + 1
+	hashCount := int(float64(bitsPerElement)*math.Log(2)) + 1
```
kaloyan-raev: Why do we have the `+ 1` in the above two formulas? I don't see this coming from the Wikipedia link.

Perhaps, you mentioned it yesterday, but I don't remember...

kaloyan-raev: @egonelbre ^

kaloyan-raev: Also the formula for hashCount does not seem correct. Shouldn't it be: **k = -log<sub>2</sub>p**

egonelbre: Not sure. I guess we should separate this logic out and implement tests for the optimal params. @Fadila82

Fadila82: For the +1, it's to be sure to have enough bits per element and hashes to achieve the desired probability. It's the same than ceil(float)

For the hashCount, it also says that k= (m/n) ln 2, which is what is used.

egonelbre: Based on https://hur.st/bloomfilter

```
filterSizeInBits = ceil((elementCount * log(p)) / log(1 / pow(2, log(2))));
hashCount = round((filterSizeInBits / elementCount) * log(2));
```

The `+1` yeah looks wrong. But it might be to avoid zero condition.

egonelbre: For `ceil(x)` is (almost) the same as `floor(x + 1)`... however here we should be ceiling to the next byte or larger, since we cannot create a bucket with less bits... i.e. `ceil(x/32)*32`. And we probably cannot remove the size for that same reason.

For the `round`, that would need to be `int((filterSize / elementCount) * math.Log(2) + 0.5)`.

egonelbre: (I'm not sure how large the practical difference is.)

egonelbre: https://play.golang.org/p/iWPAwX_uDIN

egonelbre: and with rounding to nearest uint64 https://play.golang.org/p/feVrJ7HyWHK

Fadila82: It seems to achieve better in terms of size of the bloom filter, but it gives a false positive rate higher than the targeted one. 
For instance, if I try it with an false positive rate of 0.1, the current one gives:
0.0964
0.0878
0.087
0.1018
0.0916

With the second one:
0.1274
0.124
0.1136
0.1164
0.119

And with the third:
0.1274
0.124
0.1136
0.1164
0.119

docs/design/garbage-collection.md
```diff
@@ -145,3 +146,74 @@ see: [Bloom filter math](http://pages.cs.wisc.edu/~cao/papers/summary-cache/node
...
+|0.11|	7864829|	0.06|	574296|	    0.10|	574376|	0.10|	625012|	0.08|
+|0.12|	7864829|	0.06|	551656|	    0.11|	551736|	0.11|	625012|	0.08|
+|0.13|	7340620|	0.12|	530832|	    0.11|	530904|	0.12|	625012|	0.08|
+|0.14|	7340620|	0.12|	511552|	    0.12|	511624|	0.13|	625012|	0.08|
+|0.15|	7340620|	0.12|	493600|	    0.14|	493672|	0.14|	500012|	0.16|
+|0.16|	7340620|	0.12|	476816|	    0.15|	476888|	0.15|	500012|	0.16|
+|0.17|	7340620|	0.12|	461040|	    0.15|	461112|	0.16|	500012|	0.16|
+|0.18|	7340620|	0.12|	446168|	    0.17|	446240|	0.17|	500012|	0.16|
+|0.19|	7340620|    0.12|	432104|	    0.18|	432176|	0.18|	500012|	0.16|
+|0.20|	7340620|	0.12|	418760|	    0.19|	418832|	0.19|	500012|	0.16|
```
zeebo: So one disadvantage of the scalable bloom filter is that it may allocate a new, 2x as large filter, right before you're done adding values. Sort of like having a byte slice of size 4096, appending one value, and ending up with one of capacity 8192.

I do expect the sbloom implementation to be larger, but not 10x larger. I wonder if it's very sparse and would compress well. I don't expect the other filters to compress well if they're tuned correctly: at optimal tuning, a bloom filter is 50% 1 bits randomly distributed. 

Can we try a quick gzip around the sbloom output and see if that makes a big difference?

pkg/bloomfilter/filter.go
zeebo: I think you may want to try keeping it as a float all the way to `sizeInBytes` and only then rounding up.

```go
	// calculation based on https://en.wikipedia.org/wiki/Bloom_filter#Optimal_number_of_hash_functions
	bitsPerElement := -1.44 * math.Log2(falsePositiveRate)
	hashCount := bitsPerElement * math.Log(2)
	if hashCount > 255 {
		// it will never be larger, but just in case to avoid overflow
		hashCount = 255
	}
	sizeInBytes := int(math.Ceil(float64(expectedElements) * bitsPerElement / 8))

	return newExplicit(seed, byte(hashCount), sizeInBytes)
```

egonelbre:
> There was some discussion about cuckoo filters, but it was dropped without much justification.

The reason Cuckoo filter was dropped:

1. we cannot use deletion
2. when FPP is > 3%, then bloom filter is smaller (https://bdupras.github.io/filter-tutorial/)
3. https://bdupras.github.io/filter-tutorial/, suggests Bloom filters when insertion speed is of concern

So, overall: it'll be probably worse in performance and size for our use-case.

pkg/bloomfilter/filter.go
egonelbre: I wouldn't have expected this. I guess we can leave it as is then.

But, this suggests there is some problem somewhere.

egonelbre: Anyways, I reran the calculations:

Stats from different implementations, also uploaded the diagnostic tool:

```
mse = mean-squared-error from the expected false positive
hc  = hash count
sz  = size of the table
```

current:
```
n=    1000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=1.2 KiB
n=   10000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=12.2 KiB
n=  100000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=125.0 KB
n= 1000000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=1.2 MB
n=    1000 p=0.05 avg=0.03 min=0.02 mean=0.03 max=0.05 mse=0.000 hc=5 sz=0.9 KiB
n=   10000 p=0.05 avg=0.03 min=0.02 mean=0.04 max=0.05 mse=0.000 hc=5 sz=8.5 KiB
n=  100000 p=0.05 avg=0.03 min=0.02 mean=0.03 max=0.05 mse=0.000 hc=5 sz=85.4 KiB
n= 1000000 p=0.05 avg=0.03 min=0.02 mean=0.03 max=0.06 mse=0.000 hc=5 sz=0.9 MB
n=    1000 p=0.10 avg=0.09 min=0.07 mean=0.09 max=0.11 mse=0.000 hc=4 sz=625 B
n=   10000 p=0.10 avg=0.09 min=0.07 mean=0.09 max=0.11 mse=0.000 hc=4 sz=6.1 KiB
n=  100000 p=0.10 avg=0.09 min=0.07 mean=0.09 max=0.12 mse=0.000 hc=4 sz=61.0 KiB
n= 1000000 p=0.10 avg=0.09 min=0.07 mean=0.09 max=0.12 mse=0.000 hc=4 sz=625.0 KB
n=    1000 p=0.20 avg=0.15 min=0.12 mean=0.15 max=0.20 mse=0.003 hc=3 sz=500 B
n=   10000 p=0.20 avg=0.15 min=0.12 mean=0.15 max=0.18 mse=0.003 hc=3 sz=5.0 KB
n=  100000 p=0.20 avg=0.15 min=0.12 mean=0.15 max=0.17 mse=0.003 hc=3 sz=50.0 KB
n= 1000000 p=0.20 avg=0.15 min=0.11 mean=0.15 max=0.18 mse=0.003 hc=3 sz=500.0 KB
n=    1000 p=0.30 avg=0.25 min=0.22 mean=0.25 max=0.29 mse=0.002 hc=3 sz=375 B
n=   10000 p=0.30 avg=0.25 min=0.22 mean=0.25 max=0.28 mse=0.002 hc=3 sz=3.7 KiB
n=  100000 p=0.30 avg=0.25 min=0.22 mean=0.25 max=0.28 mse=0.002 hc=3 sz=36.6 KiB
n= 1000000 p=0.30 avg=0.25 min=0.22 mean=0.25 max=0.28 mse=0.003 hc=3 sz=375.0 KB
```

version 1:
```
n=    1000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=10 sz=1.2 KiB
n=   10000 p=0.01 avg=0.01 min=0.01 mean=0.01 max=0.02 mse=0.000 hc=10 sz=11.7 KiB
n=  100000 p=0.01 avg=0.01 min=0.01 mean=0.01 max=0.02 mse=0.000 hc=10 sz=117.0 KiB
n= 1000000 p=0.01 avg=0.01 min=0.01 mean=0.01 max=0.02 mse=0.000 hc=10 sz=1.1 MiB
n=    1000 p=0.05 avg=0.05 min=0.04 mean=0.05 max=0.07 mse=0.000 hc=6 sz=0.8 KiB
n=   10000 p=0.05 avg=0.06 min=0.04 mean=0.05 max=0.08 mse=0.000 hc=6 sz=7.6 KiB
n=  100000 p=0.05 avg=0.06 min=0.04 mean=0.05 max=0.07 mse=0.000 hc=6 sz=76.1 KiB
n= 1000000 p=0.05 avg=0.06 min=0.04 mean=0.06 max=0.08 mse=0.000 hc=6 sz=0.7 MiB
n=    1000 p=0.10 avg=0.11 min=0.09 mean=0.11 max=0.14 mse=0.000 hc=5 sz=599 B
n=   10000 p=0.10 avg=0.11 min=0.09 mean=0.11 max=0.14 mse=0.000 hc=5 sz=5.8 KiB
n=  100000 p=0.10 avg=0.11 min=0.09 mean=0.11 max=0.14 mse=0.000 hc=5 sz=58.5 KiB
n= 1000000 p=0.10 avg=0.12 min=0.09 mean=0.12 max=0.15 mse=0.000 hc=5 sz=585.0 KiB
n=    1000 p=0.20 avg=0.21 min=0.18 mean=0.21 max=0.25 mse=0.000 hc=3 sz=418 B
n=   10000 p=0.20 avg=0.21 min=0.17 mean=0.21 max=0.24 mse=0.000 hc=3 sz=4.1 KiB
n=  100000 p=0.20 avg=0.21 min=0.18 mean=0.20 max=0.24 mse=0.000 hc=3 sz=40.9 KiB
n= 1000000 p=0.20 avg=0.21 min=0.17 mean=0.21 max=0.24 mse=0.000 hc=3 sz=408.9 KiB
n=    1000 p=0.30 avg=0.34 min=0.30 mean=0.34 max=0.39 mse=0.002 hc=3 sz=313 B
n=   10000 p=0.30 avg=0.34 min=0.29 mean=0.34 max=0.37 mse=0.002 hc=3 sz=3.1 KiB
n=  100000 p=0.30 avg=0.34 min=0.29 mean=0.34 max=0.39 mse=0.002 hc=3 sz=30.6 KiB
n= 1000000 p=0.30 avg=0.34 min=0.30 mean=0.34 max=0.38 mse=0.002 hc=3 sz=305.9 KiB
```

jeffs calculation:
```
n=    1000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=1.2 KiB
n=   10000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=11.7 KiB
n=  100000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=116.8 KiB
n= 1000000 p=0.01 avg=0.01 min=0.00 mean=0.01 max=0.02 mse=0.000 hc=7 sz=1.1 MiB
n=    1000 p=0.05 avg=0.05 min=0.03 mean=0.05 max=0.07 mse=0.000 hc=5 sz=0.8 KiB
n=   10000 p=0.05 avg=0.05 min=0.03 mean=0.05 max=0.07 mse=0.000 hc=5 sz=7.6 KiB
n=  100000 p=0.05 avg=0.05 min=0.04 mean=0.05 max=0.07 mse=0.000 hc=5 sz=76.0 KiB
n= 1000000 p=0.05 avg=0.05 min=0.04 mean=0.05 max=0.07 mse=0.000 hc=5 sz=0.7 MiB
n=    1000 p=0.10 avg=0.10 min=0.07 mean=0.10 max=0.13 mse=0.000 hc=4 sz=598 B
n=   10000 p=0.10 avg=0.10 min=0.08 mean=0.10 max=0.13 mse=0.000 hc=4 sz=5.8 KiB
n=  100000 p=0.10 avg=0.10 min=0.09 mean=0.10 max=0.13 mse=0.000 hc=4 sz=58.4 KiB
n= 1000000 p=0.10 avg=0.10 min=0.08 mean=0.10 max=0.13 mse=0.000 hc=4 sz=583.9 KiB
n=    1000 p=0.20 avg=0.21 min=0.17 mean=0.21 max=0.24 mse=0.000 hc=3 sz=418 B
n=   10000 p=0.20 avg=0.21 min=0.17 mean=0.21 max=0.25 mse=0.000 hc=3 sz=4.1 KiB
n=  100000 p=0.20 avg=0.21 min=0.18 mean=0.21 max=0.25 mse=0.000 hc=3 sz=40.8 KiB
n= 1000000 p=0.20 avg=0.21 min=0.17 mean=0.21 max=0.23 mse=0.000 hc=3 sz=408.2 KiB
n=    1000 p=0.30 avg=0.30 min=0.27 mean=0.30 max=0.35 mse=0.000 hc=2 sz=313 B
n=   10000 p=0.30 avg=0.30 min=0.27 mean=0.31 max=0.34 mse=0.000 hc=2 sz=3.1 KiB
n=  100000 p=0.30 avg=0.30 min=0.27 mean=0.30 max=0.34 mse=0.000 hc=2 sz=30.5 KiB
n= 1000000 p=0.30 avg=0.30 min=0.26 mean=0.30 max=0.34 mse=0.000 hc=2 sz=305.3 KiB
```

Fadila82: Well, I still wonder if it's ok to have a fp rate higher than the targeted one. Are we ok with that?

egonelbre: Yeah, it's fine.

PR #2116, opened 2019-06-04
Value Attribution DBX Model

coyle:
What: 
Creates the DBX model for Value Attribution

Why:

Thanks for submitting a PR!


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -1,5 +1,18 @@
...
+//--- Value Attribution ---//
+model value_attribution (
+	key bucket_id
+
+	field bucket_id		blob
+	field user_id		blob
+	field partner_id	blob
+	field at_rest_data	int64
+	field egress_data	int64
+	field ingress_data  int64
```
JessicaGreben: indentation is off

JessicaGreben:
For reference, here is the design doc that describes this table:
https://github.com/storj/storj/blob/master/docs/design/value-attribution.md#database

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -1,5 +1,18 @@
 // dbx.v1 golang satellitedb.dbx .
 
+//--- Value Attribution ---//
+model value_attribution (
+	key bucket_id
+
+	field bucket_id		blob
+	field user_id		blob
```
jtolds: which user?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -1,5 +1,18 @@
...
+	field partner_id	blob
+	field at_rest_data	int64
+	field egress_data	int64
+	field ingress_data  	int64
```
jtolds: these values should probably not be denormalized here. instead, when collections are normally due, we should just join with this bucket when doing the usage report to see who gets a cut. so i propose dropping the three int64 fields

satellite/satellitedb/migrate.go
```diff
@@ -736,6 +736,22 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					`ALTER TABLE offers ALTER COLUMN expires_at SET NOT NULL;`,
 				},
 			},
+			{
+				Description: "Create value attribution table",
+				Version:     27,
+				Action: migrate.SQL{
+					`CREATE TABLE IF NOT EXISTS value_attributions (
```
mniewrzal: we are avoiding using `IF NOT EXISTS` because if it exists then it indicates a bigger problem with migration and we can miss it.

satellite/satellitedb/dbx/satellitedb.dbx
coyle: I don't understand the why

satellite/satellitedb/dbx/satellitedb.dbx
coyle: I thought each bucket had one user_id associated with it

jtolds: nope, buckets are associated to projects

satellite/satellitedb/dbx/satellitedb.dbx
jtolds: keeping these fields up to date increases the database load for no reason - we only pay partners when we're doing invoicing anyway, so let's delay the running calculation of how much usage there is for the invoicing and usage processes.

satellite/satellitedb/dbx/satellitedb.dbx
JessicaGreben: but projects have users. /shrug

jtolds: but a bucket does not have a single user

JessicaGreben: yeah i guess it would need to be the project owner if such a thing exists

JessicaGreben: but also Im curious because this data model is described in the design doc and was approved ref: https://github.com/storj/storj/pull/1996

so im not sure what to make of this.  I guess we should update the design doc? Or how does that work?

jtolds: shoot, didn't catch that in the design doc but we should probably update it. we currently don't have any concept of user roles. everyone's an owner right now

PR #2117, opened 2019-06-04
db: set max open conns, conn max lifetime, add db stat monitoring

jtolds:
This sets the postgres connection limit. Currently during tally runs we're seeing spikes of 20k connections concurrently, which is bad. This doesn't fix the root cause but maybe helps us find it.
This also makes these values configurable, which is important for debugging and tuning. We'll need to clean up the configuration so it's not in flags and in our config system later.

### DIFF
--- a/internal/dbutil/defaults.go
+++ b/internal/dbutil/defaults.go
@@ -16,6 +16,7 @@ var (
 	connMaxLifetime = flag.Duration("db.conn_max_lifetime", -1, "default value for database connections, -1 means the stdlib default")
 )
 
+// Configure Sets Connection Boundaries and adds db_stats monitoring to monkit
 func Configure(db *sql.DB, mon *monkit.Scope) {
 	if *maxIdleConns >= 0 {
 		db.SetMaxIdleConns(*maxIdleConns)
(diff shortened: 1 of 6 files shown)


internal/dbutil/defaults.go
```diff
@@ -3,5 +3,31 @@
...
+	"database/sql"
+	"flag"
+
+	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+)
+
+var (
+	maxIdleConns    = flag.Int("db.max_idle_conns", 50, "default value for database connections, -1 means the stdlib default")
+	maxOpenConns    = flag.Int("db.max_open_conns", 100, "default value for database connections, -1 means the stdlib default")
+	connMaxLifetime = flag.Duration("db.conn_max_lifetime", -1, "default value for database connections, -1 means the stdlib default")
```
jtolds: yes, i know this configuration situation sucks. i'm open to suggestions for how to easily apply configuration to all db connections. long term we should do something new with db setup, but for now this is a short term attempt at a remedy for our connection pooling woes

thepaul:
Aw yiss

PR #2131, opened 2019-06-05
cbindings (1/5)

bryanchriswhite:
What: 

Why:

Thanks for submitting a PR!


internal/testplanet/planet.go
```diff
@@ -76,9 +76,10 @@ type Config struct {
 	StorageNodeCount int
 	UplinkCount      int
 
-	Identities      *testidentity.Identities
-	IdentityVersion *storj.IDVersion
-	Reconfigure     Reconfigure
+	Identities         *testidentity.Identities
+	IdentityVersion    *storj.IDVersion
+	Reconfigure        Reconfigure
+	UsePeerCAWhitelist bool
```
mniewrzal: Why this needs to be added here? For testing it should be possible to manipulate this param with `Reconfigure`

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,155 @@
...
+	srcGlobs = append([]string{
+		libuplink,
+		filepath.Join(cTestsDir, "unity.c"),
+		filepath.Join(cTestsDir, "helpers.c"),
+		filepath.Join(cSrcDir, "*.c"),
+	}, srcGlobs...)
+	testBinPath := ctx.CompileC(srcGlobs...)
+	commandPath := testBinPath
+
+	if dir, ok := os.LookupEnv("STORJ_DEBUG"); ok {
```
mniewrzal: please add comment what's happening here

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,155 @@
...
+	return nil
+}
+
+func runCTest(t *testing.T, ctx *testcontext.Context, filename string, envVars ...string) {
+	runCTests(t, ctx, envVars, filepath.Join(cLibDir, "tests", filename))
+}
+
+func startTestPlanet(t *testing.T, ctx *testcontext.Context) *testplanet.Planet {
+	planet, err := testplanet.NewCustom(
+		zap.NewNop(),
```
mniewrzal: Use `"go.uber.org/zap/zaptest"` and `zaptest.NewLogger(t)`

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,155 @@
...
+			UsePeerCAWhitelist: false,
+		},
+	)
+	require.NoError(t, err)
+
+	planet.Start(ctx)
+	return planet
+}
+
+func newProject(t *testing.T, planet *testplanet.Planet) *console.Project {
```
mniewrzal: I think it should be possible to avoid creating new project and api keys. You can try to generate `Uplink` with test planet and later use `planet.Uplinks[0].APIKey[planet.Satellites[0].ID()]` to extract api key.

internal/testplanet/planet.go
bryanchriswhite: Nice, I forgot about that

lib/uplink/ext/mapping_test.go
```diff
@@ -0,0 +1,62 @@
...
+	{
+		t.Log("string")
+		testMap := newMapping()
+
+		str := "testing 123"
+		strToken := testMap.Add(str)
+
+		gotStr, ok := testMap.values[strToken]
+		require.True(t, ok)
+		assert.Equal(t, str, gotStr)
```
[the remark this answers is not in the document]
bryanchriswhite: I think we should use `assert` instead of `require` for assertions that don't prevent the test from finishing. This way, if there are multiple failing assertions, we will see them all at once instead of having to fix each thing and re-run the tests repeatedly.

lib/uplink/ext/mapping_test.go
```diff
@@ -0,0 +1,62 @@
...
+	{
+		t.Log("pointer")
+		testMap := newMapping()
+
+		str := "testing 123"
+		strToken := testMap.Add(&str)
+
+		gotStr, ok := testMap.values[strToken]
+		require.True(t, ok)
+		assert.Equal(t, str, *gotStr.(*string))
```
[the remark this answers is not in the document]
bryanchriswhite: ^

lib/uplink/ext/mapping_test.go
```diff
@@ -0,0 +1,62 @@
...
+	{
+		t.Log("string")
+		testMap := newMapping()
+
+		str := "testing 123"
+		strToken := token(1)
+		testMap.values[strToken] = str
+
+		gotStr := testMap.Get(strToken)
+		assert.Equal(t, str, gotStr)
```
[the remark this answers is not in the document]
bryanchriswhite: ^

lib/uplink/ext/mapping_test.go
```diff
@@ -0,0 +1,62 @@
...
+	{
+		t.Log("pointer")
+		testMap := newMapping()
+
+		str := "testing 123"
+		strToken := token(1)
+		testMap.values[strToken] = &str
+
+		gotStr := testMap.Get(strToken)
+		assert.Equal(t, str, *gotStr.(*string))
```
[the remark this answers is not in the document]
bryanchriswhite: ^

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,155 @@
...
+
+func runCTests(t *testing.T, ctx *testcontext.Context, envVars []string, srcGlobs ...string) {
+	srcGlobs = append([]string{
+		libuplink,
+		filepath.Join(cTestsDir, "unity.c"),
+		filepath.Join(cTestsDir, "helpers.c"),
+		filepath.Join(cSrcDir, "*.c"),
+	}, srcGlobs...)
+	testBinPath := ctx.CompileC(srcGlobs...)
+	commandPath := testBinPath
```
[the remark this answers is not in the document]
bryanchriswhite: I think this is left-over from a refactor or something. Will remove.

lib/uplink/ext/test_utils.go
bryanchriswhite: Sure; this is just for debugging. Because we're using a test context for all the bin paths (and we're deleting them after the tests finish), if you need to debug a test binary (with gdb, for example), this lets you copy those bins to the path in the `STORJ_DEBUG` env var.

If it makes sense to keep this, that env var probably needs a better name. Thoughts?

lib/uplink/ext/test_utils.go
bryanchriswhite: Nice, I didn't know that was a thing. What does it do differently?

lib/uplink/ext/mapping_test.go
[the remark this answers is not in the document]
mniewrzal: yeah but output of such multiple failures in one test case can be sometimes unreadable. also unit tests are fast to execute and re-running shouldn't be a problem. If I didn't convince you treat this as a nit :)

lib/uplink/ext/test_utils.go
mniewrzal: hmm I don't know if we should keep this. If you used it at least once then most probably yes. Regarding name, a better name would be nice. Maybe something with `STORJ_TEST_BINARY_OUTPUT` or `STORJ_TEST_BINARY_COPY` but I don't have a strong opinion. If you decide to leave it in the code add a comment to code why its there and how it can be used.

lib/uplink/ext/test_utils.go
mniewrzal: `zap.NewNop()` will print no logs. In general, I think it better integrates with tests. Logs are printed only when test will fail or verbose mode will be enabled.

lib/uplink/ext/apikey_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"storj.io/storj/lib/uplink"
+)
+
+func TestParseAPIKey(t *testing.T) {
+	var cErr Cchar
+	apikeyString := "testapikey123"
+	cAPIKeyRef := ParseAPIKey(stringToCCharPtr(apikeyString), &cErr)
+	require.Empty(t, cCharToGoString(cErr))
+
+	apikey, ok := structRefMap.Get(token(cAPIKeyRef)).(uplink.APIKey)
```
egonelbre: Test the external behavior rather than internals, testing internals are usually fragile.

lib/uplink/ext/apikey_test.go
```diff
@@ -0,0 +1,40 @@
...
+	assert.Equal(t, apikeyString, apikey.Serialize())
+}
+
+func TestSerialize(t *testing.T) {
+	apikeyString := "testapikey123"
+	apikey, err := uplink.ParseAPIKey(apikeyString)
+	require.NoError(t, err)
+	require.NotEmpty(t, apikey)
+
+	cAPIKeyRef := CAPIKeyRef(structRefMap.Add(apikey))
```
egonelbre: Test the external behavior rather than internals, testing internals are usually fragile.

Tests probably shouldn't know anything about `structRefMap`.

lib/uplink/ext/c/tests/unity.c
```diff
@@ -0,0 +1,1570 @@
+/* =========================================================================
```
egonelbre: This adds quite a lot of code, which could be written using regular C.

lib/uplink/ext/common_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+// TODO: Start up test planet and call these from bash instead
+func TestCCommonTests(t *testing.T) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	planet := startTestPlanet(t, ctx)
+	defer ctx.Check(planet.Shutdown)
+
+	runCTest(t, ctx, "common_test.c")
```
egonelbre: Separate tests would end up conflicting here, instead start one testplanet per test.

egonelbre: Maybe something like:

```
cfiles, err := filepath.Glob("*_test.c")
require.NoError(t, err)
for _, cfile := range cfiles {
    cfile := cfile
    t.Run(cfile, func(t *testing.T){
        t.Parallel()
        
        ctx := testcontext.New(t)
        defer ctx.Cleanup()

        bin, err := ctx.CompileC(cfile)
        require.NoError(t, err)

        ...
    })
}
```

lib/uplink/ext/c/tests/common_test.c
```diff
@@ -0,0 +1,42 @@
...
+#include "unity.h"
+#include "../../uplink-cgo.h"
+
+void TestGetIDVersion(void)
+{
+    char *_err = "";
+    char **err = &_err;
+    uint8_t id_version_number = 0;
+
+    IDVersion_t id_version = GetIDVersion(id_version_number, err);
```
egonelbre: Is PascalCase standard code writing in C?

bryanchriswhite: PascalCase for types seems to be a c convention; however, we currently are also using it for cgo exported functions and some helper functions. This is not conventional and we plan to change those to snake_case, which is.

egonelbre: Wasn't PascalCase only for C++... although, I've seen PascalCase C code as well, but not very commonly.

lib/uplink/ext/mapping_test.go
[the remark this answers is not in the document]
bryanchriswhite: If we're taking the fact that these are unit tests into consideration then i think it's also fair to consider that these asserts aren't in any large loops or anything and the most number of errors you'd see simultaneously is 2.

lib/uplink/ext/c/tests/unity.c
bryanchriswhite: Are you suggesting we write our own c test assertion library!?

mniewrzal: I think Egon wants to say that we are using a limited number of assertions from `unity` and maybe it would be possible to avoid including it and write our assertions with regular C.

lib/uplink/ext/common_test.go
aleitner: The go tests are more focused on unit tests. The C tests are more focused on integration tests.

lib/uplink/ext/c/tests/unity.c
egonelbre: Yeah, currently it seems a regular `==` or `memcmp` would suffice to get the same effect.

lib/uplink/ext/common_test.go
egonelbre: Based on the code it doesn't look like that, the `unity` thing seems to run unit test like things.

lib/uplink/ext/test_utils.go
bryanchriswhite: Interesting idea. I don't see any particular benefit to doing it that way and what we have now works so I'm inclined to leave it alone if it's all the same to you.

mniewrzal: I'm ok with leaving it for now. Its additional code that needs to be maintained and it's redundant. Please add a comment that this should be removed and replaced with a simpler solution. When uplink config will be used testing with multiple satellites will be easier. Also when tests will be working on my machine I can help you removing this.

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,176 @@
...
+func runCTests(t *testing.T, ctx *testcontext.Context, envVars []string, srcGlobs ...string) {
+	srcGlobs = append([]string{
+		LibuplinkSO,
+		filepath.Join(CTestsDir, "unity.c"),
+		filepath.Join(CTestsDir, "helpers.c"),
+		filepath.Join(CSrcDir, "*.c"),
+	}, srcGlobs...)
+	testBinPath := ctx.CompileC(srcGlobs...)
+
+	if dir, ok := os.LookupEnv("STORJ_C_TEST_BIN_DIR"); ok {
```
mniewrzal: Can you add comment how this can be used?

lib/uplink/ext/c/tests/unity.c
mniewrzal: Its resolved but no conclusion was done. I agree with Egon that currently, we are using only the simplest assertions from Unity and this code is not super necessary in our repo.

egonelbre: Sorry, forgot to write the conclusion, discussed this with Bryan yesterday directly. They currently have tests and things written out using `unity.c`. Yeah, it's not necessary, however bigger priority is to get c-bindings working. We'll revisit this later once c-bindings are done.

mniewrzal: Thanks, I think that's the best approach.

lib/uplink/ext/test_utils.go
bryanchriswhite: o right, yea!

lib/uplink/ext/apikey.go
```diff
@@ -0,0 +1,38 @@
...
+//   #include "c/headers/main.h"
+// #endif
+import "C"
+import (
+	"storj.io/storj/lib/uplink"
+)
+
+// ParseAPIKey parses an API Key
+//export ParseAPIKey
+func ParseAPIKey(val *C.char, cErr **C.char) (cAPIKey C.APIKeyRef_t) {
```
egonelbre: I just realized that this error handling approach probably isn't too C-like.

```
func ParseAPIKey(val *C.char, cErr **C.char) (cAPIKey C.APIKeyRef_t) {

vs.

func ParseAPIKey(val *C.char, apikey *C.APIKeyRef_t) (err C.APIKeyRefError_t) {
```

Most likely we want the latter, however might be difficult to implement right now.

egonelbre: _Something to keep in mind._

lib/uplink/ext/main.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+//go:generate go build -o uplink-cgo.so -buildmode=c-shared .
```
egonelbre: Creating `.so` with go generate feels very wrong to me. It probably should be part of a makefile or some other approach. I think this is the reason why it was also globbering the source directory.

lib/uplink/ext/test_utils.go
```diff
@@ -0,0 +1,180 @@
...
+	// CTestsDir is the path to the storj/lib/uplink/ext/c/src directory.
+	CTestsDir string
+	// LibuplinkSO is the path to the storj/lib/uplink/ext/uplink-cgo.so shared object.
+	LibuplinkSO string
+
+	testConfig = new(uplink.Config)
+)
+
+func init() {
+	// TODO: is there a cleaner way to do this?
```
egonelbre: Yes, build the `.so` into a temporary directory and the working directory points to the package.

lib/uplink/ext/uplink-cgo.h
```diff
@@ -0,0 +1,105 @@
+/* Code generated by cmd/cgo; DO NOT EDIT. */
```
egonelbre: I don't think this should be checked in, it seems like a by production of compilation.

lib/uplink/ext/tests.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+//go:generate go run .
+//go:generate go test .
```
egonelbre: Is this a leftover? `go generate` shouldn't run tests.

bryanchriswhite: It was convenient during development to have a make target that built the .so and ran the c api tests. Currently, the `test-bindings` does this by calling generate on this file. What would you suggest doing instead?

lib/uplink/ext/uplink-cgo.h
bryanchriswhite: My thought was that it was very similar to the sort of thing that is produced by dbx and/or protobuf and we do commit that stuff.

lib/uplink/ext/main.go
bryanchriswhite: I don't have a preference on how it gets built. Currently a make target (`libuplink`) is calling `go generate` on this file.

bryanchriswhite: What are you talking about with the source directory?

lib/uplink/ext/test_utils.go
bryanchriswhite: Cgo needs the paths to included c files (in the `import C` comment) to be valid but so do the c tests when we build them. The problem is that the header file produced by cgo (uplink-cgo.h) includes those paths so you can't move the `.so` and it's header file relative to the package root without either breaking cgo compilation or the c test compilation

bryanchriswhite: Unless we rewrite the `uplink-cgo.h` header file to change the paths after it's written, I'm not sure how else we could move it.

egonelbre: I'm not following.

As I understand the tests are fully in C, so cgo doesn't really matter here. Since you are compiling tests separately anyways you can set include paths as needed.

lib/uplink/ext/uplink-cgo.h
egonelbre: This is more similar to an `.exe` rather than anything else. I.e. nothing in our Go codebase should directly depend on it, AFAIU.

PR #2134, opened 2019-06-05
satellite/orders: TODOs about transactions

jtolds:
potentially related to https://storjlabs.atlassian.net/browse/V3-1796


jtolds:
perhaps @zeebo can propose a dbx pattern that allows us to have functions operate on existing batched transactions, instead of having each function have its own transaction(s)

satellite/orders/service.go
```diff
@@ -84,6 +84,7 @@ func (service *Service) updateBandwidth(ctx context.Context, bucketID []byte, ad
 	now := time.Now().UTC()
 	intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
 
+	// TODO: all of this below should be a single db transaction. in fact, this whole function should probably be part of an existing transaction
 	if err := service.orders.UpdateBucketBandwidthAllocation(ctx, bucketID, action, bucketAllocation, intervalStart); err != nil {
 		return err
 	}
```
jtolds: the for loop below appears to be a huge component of @thepaul's current CreateSegment performance ticket

jtolds:
oh I'm a dummy. we need to just be passing the current live transaction around and reusing it

satellite/orders/endpoint.go
```diff
@@ -199,21 +201,22 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
 
 		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, bucketID, orderLimit.Action, order.Amount, intervalStart); err != nil {
+			// TODO: we should use the serial number in the same transaction we settle the bandwidth? that way we don't need this undo in an error case?
 			if err := endpoint.DB.UnuseSerialNumber(ctx, orderLimit.SerialNumber, orderLimit.StorageNodeId); err != nil {
 				log.Error("unable to unuse serial number", zap.Error(err))
 			}
 			return err
 		}
 
+		// TODO: whoa this should also be in the same transaction
```
egonelbre: Wouldn't using transactions make it slower?

The current logic is implemented to try and avoid transactions for performance, but yeah it could be batched or a single transaction.

jtolds: i mean ideally it's all in the same query even, but yes sharing a transaction will drastically improve performance

egonelbre: Oh, interesting.

PR #2135, opened 2019-06-06
Reputation / node selection doc

wthorp:
What: 
Its a design doc for reputation and node selection!

Why:
We like design docs!


docs/design/node-selection.md
```diff
@@ -0,0 +1,82 @@
...
+
+The node selection SQL queries will also change.  Twice as many nodes must returned from these functions to satisfy the "Power of Two Choices" requirement, which gives preference to nodes with better reputation.  For every two nodes returned, the one with the higher reputation scores will be selected returned and the other discarded.
+
+Note that the initial implementation has two different reputation statistics:  audit and uptime.  For the purposes of node selection, we assume that these two reputation can be combined by scaling one of them by some constant.  We further assume that operations may weigh these reputations differently.  The initial configuration should include a uptime_repair_weight, audit_repair_weight, uptime_uplink_weight, and audit_uplink_weight constants.
+
+> Total Repair Reputation = uptime_repair_weight · uptime R(n) + audit_repair_weight · audit R(n)
+>
+> Total Uplink Reputation = uptime_uplink_weight · uptime R(n) + audit_uplink_weight · audit R(n)
```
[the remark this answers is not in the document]
ifraixedes: The document needs some clarification about the 2 total reputations, the _repair_ and the _uplink_ one.

I may lack in how the audits and uptime checks are done, but I don't personally get this separation and specifically what the _uplink_ reputation is.

docs/design/node-selection.md
```diff
@@ -0,0 +1,82 @@
...
+## Implementation
+
+* Create configuration elements for audit_a0, audit_β0, audit_λ, audit_w, audit_repair_weight, and audit_uplink_weight
+* Create configuration elements for uptime_a0, uptime_β0, uptime_λ, uptime_w, uptime_repair_weight, and uptime_uplink_weight
+* Alter DBX model removing audit_success_count, audit_success_ratio, uptime_success_count, uptime_ratio
+* Alter DBX model adding audit_reputation_a, audit_reputation_b, uptime_reputation_a, uptime_reputation_b
+* Alter SQL node selection queries to consider new values
+* Alter SQL node selection queries to return 2x more nodes
+* Implement "Power of Two Choices" logic in node selection query
+* Update disqualification code to use reputation
```
ifraixedes: I'm not entirely sure about this task.
Is the disqualification implemented? my main understanding is that that disqualification is mostly a flag to mark a node when it is disqualified, currently there is not such logic, there is only the database column to mark the node and the node selection queries has been updated to ignore those nodes which are marked as disqualified.

With the creation of this document, I see that the [Node Disqualification document](https://github.com/storj/storj/blob/master/docs/design/disqualification.md) could have some modifications/adaptations for example in the section [Determining Disqualification](https://github.com/storj/storj/blob/master/docs/design/disqualification.md#determining-disqualification) could be simplified that it's based in the reputation and link to this document.

docs/design/node-selection.md
```diff
@@ -0,0 +1,82 @@
...
+
+The reputation _R(n)_ is calculated based on some measured success _v_, two persisted values a and β, and fixed configuration values λ and _w_.  
+
+> α(n) = λ·α(n-1) + _w_(1+_v_)/2
+>
+> β(n) = λ·β(n-1) + _w_(1-_v_)/2
+>
+> R(n) = α(n) / (α(n) + β(n))
+
+Initial values for a and β - a0 and β0 - will also be configuration values.  While _v_ may need to vary depending on our "easing" implementation, it will likely vary algorithmically and should not require configuration.  The initial implement may simply assume that _v_ = 1.
```
benjaminsirb: v is a scalar selected from the interval [-1,1]. It must vary based on the success/failure of the transaction.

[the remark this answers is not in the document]
wthorp: I do not agree with the Data Science team's assertion that we need to 'ease' into a reputation.  That's why I do not have these equations in there.  Its also why it took me three days to write this; it was a difficult decision to ignore their guidance.  Feedback on that decision would be appreciated.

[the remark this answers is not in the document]
benjaminsirb: @wthorp "easing" to a reputation was one of two suggestions. Either way v cannot be fixed to be equal to 1. In the simplest case, v=1 for a success and v=-1 for a failure. See https://github.com/storj/datascience/blob/master/reputation/choosing_v/main.pdf for a quick explanation.

Sorry if there was confusion on this point.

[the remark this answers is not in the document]
ifraixedes: Clarified the disagreement, the only to change is that _v_ can be 1 or -1 one as commented above.

docs/design/node-selection.md
```diff
@@ -0,0 +1,82 @@
+# Title: Reputation and Node Selection
```
ifraixedes: Remove "Title:"

docs/design/node-selection.md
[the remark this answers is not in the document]
wthorp: Better?

docs/design/node-selection.md
wthorp: Better?

docs/design/node-selection.md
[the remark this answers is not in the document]
ifraixedes: If you're referring to the variable formatting of the above paragraph, yes they are more readable, however it doesn't reply my doubt.

~~Meanwhile I understand the `Total Repair Reputation` as the reputation of the node when the satellite is doing repairs and use contact to that specific network (uptime) I'm wondering what an audit means in the context of a repair?~~

~~On the other hand, I don't know when a satellite does the uptime and audits for the `Total Uplink Reputation` calculation, because I don't understand what `Uplink` is in the Satellite context.~~

While writing the above, I've understood what this is, mostly there are 2 process, the uptime and audit which run in the satellite, then each time that both are completed, there will be registered this 2 different total reputations, one will be used for the repair process to select nodes based on the `Total Repair Reputation` and the other (`Total Uplink Reputation`) will be used for returning the list of nodes to uplink when it request orders to upload new files.

docs/design/node-selection.md
ifraixedes: Yes, it's clearer now.

However I'm still thinking that the task should be in the _Node Disqualification document_ because as far as I know it isn't already implemented, right?

docs/design/node-selection.md
```diff
@@ -0,0 +1,87 @@
...
+
+The node selection SQL queries will also change.  Twice as many nodes must returned from these functions to satisfy the "Power of Two Choices" requirement, which gives preference to nodes with better reputation.  For every two nodes returned, the one with the higher reputation scores will be selected returned and the other discarded.
+
+Note that the initial implementation has two different reputation statistics:  audit and uptime.  For the purposes of node selection, we assume that these two reputation can be combined by scaling one of them by some constant.  We further assume that different operations may weigh these reputations differently. For instance, repair may be more concerned that a node is reliable than it is speedy.  New file uploads coming from an uplink may have different criteria.   The initial configuration should include a `uptime_repair_weight`, `audit_repair_weight`, `uptime_uplink_weight`, and `audit_uplink_weight` constants.
+
+> Total Repair Reputation = uptime_repair_weight · uptime R(n) + audit_repair_weight · audit R(n)
```
ifraixedes: I'm thinking in the implementation of these totals and how `uptime R(n)` and `audit R(n)` are going to be correlated, because each check runs in a different process but we need both results.

My understanding is that `n` is the number of the check which if they were executed at the same moment `n` will be value of the counter (henceforth `N`), however the checks aren't coordinated so we could store the last one and then combined when the both `n` are registered, but it won't work because the frequency of each check could be (or it may already be) different, so the one with lower frequency will never catch the one with higher.

For example:

* Audit runs every 60 seconds
* Uptime runs every 10 seconds.

__After 10 seconds__

* Audit_N = 0
* Uptime_N = 1
	* Uptime(1): OK

__After 20 seconds__

* Audit_N = 0
* Uptime_N = 2
	* Uptime(2): OK

...

__After 60 seconds:__

* Audit_N = 1
	* Audit(1): OK
* Uptime_N = 6
	* Uptime(6): Error

Total Reputation Repair (1) = uptime_repair_weight * uptime R(1) + audit_repair_weight * audit R(1)
Total Uplink Reputation (1) = uptime_uplink_weight * uptime R(1) + audit_uplink_weight * audit R(1)

...

__After 90 seconds:__

* Audit_N = 1
* Uptime_N = 9
	* Uptime(9): OK

...

__After 120 seconds:__

* Audit_N = 1
	* Audit(2): Error
* Uptime_N = 12
	* Uptime(12): OK

Total Reputation Repair (2) = uptime_repair_weight * uptime R(2) + audit_repair_weight * audit R(2)
Total Uplink Reputation (2) = uptime_uplink_weight * uptime R(2) + audit_uplink_weight * audit R(2)


Storing the values for each `n` doesn't look feasible and coordinate both process doesn't seem a good idea.

wthorp: While the `n` in `R(n)` effectively equals `TotalAuditCount` or `TotalUptimeCount`, these values are not used when calculating `R(n)`.  At any given point in time, audit and uptime are expected to be using different effective values of `n`, due to them being measured at different rates.

docs/design/node-selection.md
wthorp: I don't have a strong opinion on where this work lands.  It can be one PR or two, I assumed it'd be two PRs.

ifraixedes: Well, then it must exist in order to be updated, but if it doesn't exist I'm not sure what we'll have to update.
Anyway, we can figure it out at the time to work on it.

docs/design/node-selection.md
ifraixedes: so basically it's

`Total Repair Reputation = uptime_repair_weight · uptime R(n) + audit_repair_weight · audit R(m)`

The question is now how to implement it, because those uptimes and audits run in different Go routines so it could be done in different way, but we can have that definition in the ticket that will tackle such task.

PR #2142, opened 2019-06-06
wip: scopes in lib/uplink

zeebo:
What: Adds the ability for a project to be initialized with a slice of `EncryptionAccess`

Why: This lets us do scopes as described in https://github.com/storj/storj/blob/master/docs/design/uplink-scopes.md


lib/uplink/project.go
```diff
@@ -147,20 +148,37 @@ func (p *Project) GetBucketInfo(ctx context.Context, bucket string) (b storj.Buc
...
 	bucketInfo, cfg, err := p.GetBucketInfo(ctx, bucketName)
 	if err != nil {
 		return nil, err
 	}
 
-	if access == nil || access.Key == (storj.Key{}) {
-		return nil, Error.New("No encryption key chosen")
+	// Construct a searcher from the accesses. This allows finding the appropriate key for
+	// any path in the bucket.
+	if len(accesses) != 1 {
```
kaloyan-raev: Shouldn't this be `if len(accesses) > 1`?

lib/uplink/project.go
```diff
@@ -147,20 +148,37 @@ func (p *Project) GetBucketInfo(ctx context.Context, bucket string) (b storj.Buc
...
+		return nil, Error.New("At least one encryption access must be provided")
 	}
-	if err != nil {
-		return nil, err
+
+	searcher := vpath.NewSearcher()
+	for _, access := range accesses {
+		if access == nil || access.Key == (storj.Key{}) {
```
kaloyan-raev: nit: `access.Key.IsZero()` can be used instead of `access.Key == (storj.Key{})`.

pkg/encryption/path.go
```diff
@@ -58,24 +58,10 @@ func DecryptPath(path storj.Path, cipher storj.Cipher, key *storj.Key) (decrypte
...
-	if len(path) == 0 {
-		return key, nil
-	}
-
-	comps := storj.SplitPath(path)
-	if depth > len(comps) {
-		return nil, Error.New("depth greater than path length")
-	}
-
+func DerivePathKey(path storj.Path, key *storj.Key) (derivedKey *storj.Key, err error) {
```
kaloyan-raev: Could you update the godoc comment to reflect what this method now exactly does?

pkg/vpath/helpers.go
```diff
@@ -0,0 +1,38 @@
+package vpath
```
kaloyan-raev: Just wonder why the package is called `vpath`?

pkg/vpath/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"strings"
+)
+
+// pathWalker efficiently pops components off of a path and keeps track of the
+// amount of path consumed and remaining.
+type pathWalker struct {
+	path     string
```
kaloyan-raev: Any reason for not using `storj.Path` instead of `string`?

kaloyan-raev: Or perhaps, you need this is as a more efficient way to traverse through path components than using `storj.SplitPath`?

pkg/vpath/helpers_test.go
```diff
@@ -0,0 +1,34 @@
...
+		{Path: "u1/u2/u3", Parts: []string{"u1", "u2", "u3"}},
+	}
+
+	for _, test := range cases {
+		path := newPathWalker(test.Path)
+
+		for !path.Empty() {
+			if exp, got := test.Parts[0], path.Next(); exp != got {
+				t.Fatal("exp:", exp, "got:", got)
+			}
```
kaloyan-raev: Wouldn't `assert.Equal(t, test.Parts[0], path.Next())` do the same? It also won't stop the test here and will allow the rest of the test cases to execute.

pkg/vpath/helpers_test.go
```diff
@@ -0,0 +1,34 @@
...
+			test.Parts = test.Parts[1:]
+		}
+
+		if len(test.Parts) != 0 {
+			t.Fatal("extra parts:", test.Parts)
+		}
```
kaloyan-raev: `assert.Empty(test.Parts)` ?

pkg/storage/streams/store.go
```diff
@@ -61,43 +65,65 @@ type Store interface {
...
+	defer mon.Task()(&ctx)(&err)
+
+	// TODO(jeff): we can do stuff with the returned map if base is nil because it allows us
+	// to decrypt specific components. This is important during list.
+	_, consumed, base := s.searcher.Lookup(path)
+	if base == nil {
+		return "", nil, Error.New("cannot find encryption information for: %q", path)
+	}
+
+	remainingPath := path[len(consumed):]
```
kaloyan-raev: I am quite anxious about such kind of operations. Does `remainingPath` include a leading slash now?

i.e. if `path` is `a/b/c` and consumed is `a` then `remainingPath` will be `/b/c`, not `b/c`. I guess this can lead to some problems.

pkg/encryption/path_test.go
```diff
@@ -46,46 +46,38 @@ func TestEncryption(t *testing.T) {
...
 				continue
 			}
+
+			secondEncrypted, err := EncryptPath(path[1], cipher, derivedKey)
 			if !assert.NoError(t, err, errTag) {
 				continue
 			}
 
-			shared := storj.JoinPaths(storj.SplitPath(encrypted)[tt.depth:]...)
-			decrypted, err := DecryptPath(shared, cipher, derivedKey)
```
kaloyan-raev: There was some value in testing if a subpath can be decrypted with a derived key. Why aren't we testing this anymore?

lib/uplink/project.go
zeebo: Yep.

lib/uplink/project.go
zeebo: Cool. I was porting from how it was already written, but I'm glad you told me.

pkg/encryption/path.go
zeebo: Of course. I'm not sure this is staying this way, either.

pkg/vpath/helpers.go
zeebo: Temporary name. Sorta like "virtual path"? I'll probably rename/move this into some other package before I'm done.

pkg/vpath/helpers.go
zeebo: `storj.Path` is a type alias to `string`, so they're really the same. The answer is I wrote this path walker a long time before trying to integrate it into the code base, and forgot to update it when I learned about `storj.Path`.

pkg/vpath/helpers_test.go
zeebo: Sure, though, I'd probably want to break out of this inner `for` loop.

pkg/vpath/helpers_test.go
zeebo: Neat!

pkg/storage/streams/store.go
zeebo: Yeah, I'm also anxious about it. I'm also anxious about how sometimes a `storj.Path` has a bucket prefix and sometimes it doesn't.

I'm probably going to be doing a larger refactoring to create some type that encapsulates all of this complexity and makes it very type safe. Like a `storj.BucketPath` or something.

pkg/encryption/path_test.go
zeebo: As far as I could tell, this is the only place that the `depth` parameter was used. Everywhere else always passed in the length of the components.

Additionally, decryption coverage is handled in the previous test.

In other words, if `enc(prefix, rootKey) + "/" + enc(part, prefixKey) == enc(prefix + "/" + part, rootKey)` for all combinations of `prefix`, `part` and `rootKey`, then we already know that decryption works properly at every subpath.

zeebo:
Got merged in smaller pieces.

PR #2143, opened 2019-06-06
Ensure Storage Nodes collect expired used serial numbers

egonelbre:
What: Ensure Storage Nodes collect expired used serial numbers.

Why: Otherwise the used serials numbers table keeps growing.

Thanks for submitting a PR!


storagenode/collector/service_test.go
```diff
@@ -61,9 +64,37 @@ func TestCollector(t *testing.T) {
...
+		require.Equal(t, serialsPresent, 2)
+
+		serialsPresent = 0
+
+		// imagine we are 48 days in the future
+		for _, storageNode := range planet.StorageNodes {
+			usedSerials := storageNode.DB.UsedSerials()
+
+			// collect all the data
+			err = storageNode.Collector.Collect(ctx, time.Now().Add(48*24*time.Hour))
```
VinozzZ: I know this is tests, but should we try to use `UTC` time across the board?

egonelbre: I'm honestly not sure. Initially I was thinking that using UTC in DB will avoid problems, however I think datetime without timezone created some problems, I didn't know about. So I currently just used with timezone and let DB handle it. But this UTC and no timezone vs non-UTC + timezone needs to be revisited.

thepaul: “with timezone” will store a unix time, essentially, so it might as well be UTC. The benefit is that no code on the setting or fetching sides needs to remember to do UTC conversions. See https://wiki.postgresql.org/wiki/Don't%5FDo%5FThis#Don.27t_use_timestamp_.28without_time_zone.29

egonelbre: Note this is sqlite code.

egonelbre: Also we have https://storjlabs.atlassian.net/browse/V3-1882

thepaul: My point was intended to apply in general, not just here.

PR #2146, opened 2019-06-07
satellite/orders: batch update storage node allocations

egonelbre:
What: each allocation for storage node was updated individually making it slow

Why:

Thanks for submitting a PR!


satellite/orders/endpoint.go
```diff
@@ -39,9 +39,9 @@ type DB interface {
 	UpdateBucketBandwidthInline(ctx context.Context, bucketID []byte, action pb.PieceAction, amount int64, intervalStart time.Time) error
 
 	// UpdateStoragenodeBandwidthAllocation updates 'allocated' bandwidth for given storage node
-	UpdateStoragenodeBandwidthAllocation(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
+	UpdateStoragenodeBandwidthAllocation(ctx context.Context, storageNode []storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
 	// UpdateStoragenodeBandwidthSettle updates 'settled' bandwidth for given storage node
-	UpdateStoragenodeBandwidthSettle(ctx context.Context, storageNode storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
+	UpdateStoragenodeBandwidthSettle(ctx context.Context, storageNodes storj.NodeID, action pb.PieceAction, amount int64, intervalStart time.Time) error
```
thepaul: It looks like this argument rename was supposed to be done on UpdateStoragenodeBandwidthAllocation instead.

satellite/satellitedb/orders.go
```diff
@@ -125,20 +129,44 @@ func (db *ordersDB) UpdateBucketBandwidthInline(ctx context.Context, bucketID []
...
-	)
-	_, err = db.db.ExecContext(ctx, statement,
-		storageNode.Bytes(), intervalStart, defaultIntervalSeconds, action, uint64(amount), 0, uint64(amount),
-	)
-	if err != nil {
-		return err
+
+	switch t := db.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		statement := db.db.Rebind(
```
thepaul: db.db.Rebind() is a noop here, since sqlite3 wants the `?` bind markers that are already being used.

Still, probably harmless to keep it here.

satellite/satellitedb/orders.go
```diff
@@ -125,20 +129,44 @@ func (db *ordersDB) UpdateBucketBandwidthInline(ctx context.Context, bucketID []
...
+			`INSERT INTO storagenode_bandwidth_rollups (storagenode_id, interval_start, interval_seconds, action, allocated, settled)
+			VALUES (?, ?, ?, ?, ?, ?)
```
thepaul: If you want, we could add a row of `, (?, ?, ?, ?, ?, ?)` here for each storageNode, and perform all the inserts at the same time in the same statement. Sqlite supports that ok.

like

    INSERT INTO storagenode_bandwidth_rollups (storagenode_id, interval_start, interval_seconds, action, allocated, settled)
    VALUES (?, ?, ?, ?, ?, ?)
    , (?, ?, ?, ?, ?, ?)
    , (?, ?, ?, ?, ?, ?)
    , (?, ?, ?, ?, ?, ?)
    ON CONFLICT (storagenode_id, interval_start, action)
    DO UPDATE SET allocated = storagenode_bandwidth_rollups.allocated + excluded.allocated

With sqlite being an in-memory db where round-trips aren't very costly, this might not have any meaningful effect on performance (and we may not particularly care about performance here anyway), but at least it would be more obvious that the inserts are meant to be done all-at-once, in the same transaction?

egonelbre: They aren't required to be done in the same transaction and the values change would complicate the code here quite a bit. So, I would prefer keep the code simpler for sqlite.

thepaul: Sure, it's your call.

Just for consideration though, the more the semantic details diverge between the sqlite side (which powers most of our tests) and the postgresql side (mostly used in production), the effectiveness of our tests suffers. I don't see how it could matter here, so I won't push very hard, but "we do these few hundred things in a single transaction in production but many separate transactions in test" is pretty scary to me.

egonelbre: Most of the tests do run against sqlite and postgres both in Jenkins. i.e. `testplanet.Run` and `satellitedbtest.Run` iterate over database configurations.

There's only a difference, if you haven't setup testing postgres locally.

PR #2150, opened 2019-06-07
reputation: Add configuration parameters

ifraixedes:
__What__:

Add the configuration parameters which will be used by the algorithm
which will calculate the storage node reputation.

__Why:__

Because the algorithm to calculate the SNs reputation require them as stated in draft [Reputation & Node Selection design doc](https://github.com/storj/storj/pull/2135)

__Other concerns:__

The reputation calculation is based on audit and uptime check
results hence some configuration parameters are in pkg/audit, others in
pkg/discovery and others in the satellite which will combine the both
reputation results to obtain the storage node reputation for repair and
uplink.

__NOTE__: only a few default configuration values are specified in the document, the rest of them I don't know which are the good default values, so I set them to `1.0` and waiting in the review to know which are the correct default values.

Thanks for submitting a PR!


ifraixedes:
The CI fails due to satellite configuration has changed and the config lock file isn't updated so far, because the review will confirm if this configuration parameters are correct and are in the right place.

pkg/audit/service.go
```diff
@@ -30,6 +30,11 @@ type Config struct {
 	MinBytesPerSecond  memory.Size   `help:"the minimum acceptable bytes that storage nodes can transfer per second to the satellite" default:"128B"`
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"5s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
+
+	ReputationAlpha0 float64 `help:"the initial shape 'alpha' used to calculate audit SNs reputation" default:"1.0"`
```
navillasa: why Alpha0 and Beta0 instead of just Alpha and Beta?

wthorp: The zero is meant to show that this is the initial value, rather than the value at the Nth measurement.

navillasa:
since these reputation ratios will be used for node selection, I'm thinking they could potentially be better located in the `pkg/overlay/config.go` file.

ifraixedes:
> since these reputation ratios will be used for node selection, I'm thinking they could potentially be better located in the pkg/overlay/config.go file.

At the end they need to be used for calculating the _uptime_ and _audit_ reputation which will be used for calculating the total _repair_ and _uplink_ reputation, these 2 last ones will be the one used for selecting and disqualify nodes.

If _pkg/overlay_ package is the package which will do those calculations, then yes, they should be there.

I cannot say because I'm not still very clear in the how the sources are split across packages in the project, so I cannot have an opinion on why not to move them to the _pkg/overlay_, but I would like that another reviewer have the same opinion before doing the change.

ifraixedes:
As commented in other means, the configuration parameters have been moved to _pkg/overlay_ as @navillasa suggested.

I'm going to warn our Ops team about the new configuration parameters.

PR #2152, opened 2019-06-07
satellite payments local add db

rikysya:
What: Add db for local payments.

Why: For storing invoices, so that console can operate normally(with ability to show billing items with correct numbers, create project invoices)

I am not sure that placing `local_invoice` table to `satellite.DB` is correct.

Also, I am wondering if moving to `project_invoice_stamp` from `console` to `stripepayments` and exposing a `GetProjectInvoices(ctx, ...)` method from `payments.Service` interface is a better architectural decision?

Thanks for submitting a PR!


egonelbre:
I think having it in `satellitedb` is fine. Technically it could be handled separate to not conflate development and release tables, but separating these would add more confusion and more work than necessary. We can always separate them when needed. The main downside is that we now need to handle migrations etc. with that table for testing.

I'm not sure whether I was following the `project_invoice_stamp` completely, but if it's only applicable to `stripepayments` and not `localpayments` then moving it into `stripepayments` makes sense. However, when `payments.Service` needs to expose some method for it, it seems that it might belong to `payments` package instead.

rikysya:
@egonelbre, ok, I got it about db placement, that `satellitedb` is fine, thanks!

For the other part, `project_invoice_stamp` creates tie between `console.Project` and `stripe.Invoice`(it can be any payments service actually). As there are no such thing as our `project` on any payments service. Right know `project_invoice_stamp` also help preventing double invoicing by recording billing period for the invoice.(And you can query invoices by billing period like that, which can not be done by just using `stripe` api). This responsibility is on `console` as it is interact with `project_invoice_stamps` directly. But as `payments` can be referred as `storj.Payments` we can delegate those rules to the implementation, and expose a method like `payments.Service.GetProjectInvoices` which encapsulate the actual implementation(depends on `payments service` api). But that also means that if we pick some other payments provider(with similar behavior as `stripe`) to work with we may end up two separate tables serving the same reason, also we would be able to handle payments providers separately(your can restart your `satelllite` with different payments setup and your invoicing data won't be corrupted). However it would be possible to create two different invoices for the same billing period on separate payments services.

satellite/payments/localpayments/local.go
```diff
@@ -98,27 +102,78 @@ func (*service) GetCustomerPaymentsMethods(ctx context.Context, customerID []byt
...
+	amount += egress.Amount
+	amount += objectCount.Amount
+
+	inv, err := s.db.CreateInvoice(ctx,
+		payments.Invoice{
+			PaymentMethodID: params.PaymentMethodID,
+			Amount:          amount,
+			Currency:        payments.CurrencyUSD,
+			Status:          payments.InvoiceStatusPaid,
+			LineItems:       lineItems,
```
egonelbre: You can use `[]payments.LineItem{storage, egress, objectCount}` directly, the append probably isn't needed.

satellite/payments/localpayments/local.go
```diff
@@ -98,27 +102,78 @@ func (*service) GetCustomerPaymentsMethods(ctx context.Context, customerID []byt
...
+	objectCount := payments.LineItem{
+		Key:      payments.LineItemObjectCount,
+		Quantity: int64(params.ObjectCount),
+		Amount:   int64(params.ObjectCount),
+	}
+
+	// custom fields
+	billingPeriod := payments.CustomField{
+		Name:  payments.CustomFieldBillingPeriod,
+		Value: fmt.Sprintf("%s - %s", params.StartDate, params.EndDate),
```
egonelbre: Slightly worried about localization here, but I guess it's fine for now.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -361,6 +361,67 @@ read all (
...
+)
+
+create local_invoice()
+
+read one (
+    select local_invoice
+    where local_invoice.id = ?
+)
+
+model local_invoice_line_item (
```
egonelbre: Maybe if the line items were serialized via protobuf or something, life would be much easier.

rikysya:
outdated

PR #2153, opened 2019-06-07
Move from ratios to the data science team's reputation model

wthorp:
THIS PR WILL NOT BE USED.  IT GREW TOO LARGE IN SCOPE AND IS BEING BROKEN UP.

What: 
Alters DBX model removing audit_success_count, audit_success_ratio, uptime_success_count, uptime_ratio
Alters DBX model adding audit_reputation_α, audit_reputation_β, uptime_reputation_α, uptime_reputation_β
Creates migration scripts for SQL table changes
Alters node selection SQL queries to return α, β instead of ratios

Why:
Move from ratios to the data science team's reputation model... Because ratios are too simple.

Thanks for submitting a PR!


stefanbenten:
Should there be no migration of the current "reputation" to the new schema?

satellite/satellitedb/overlaycache.go
```diff
@@ -824,13 +817,15 @@ func getNodeStats(dbNode *dbx.Node) *overlay.NodeStats {
 	return nodeStats
 }
 
-func updateRatioVars(newStatus bool, successCount, totalCount int64) (int64, int64, float64) {
+func updateReputation(newStatus bool, α, β, λ, w float64, totalCount int64) (int64, float64, float64) {
```
kaloyan-raev: It's really cool having Greek letters in the code, but let's stick to ASCII characters. We can have `alpha`, `beta`, `lambda` and `omega` instead of `α, β, λ, w`. I am even not sure if the last one is the Greek "omega" or the English "double u".

kaloyan-raev:
Could you remove from this PR `go.mod`, `go.sum` and all `*.pb.go` files, which proto file was not actually changed? This will reduce the number of changed lines in the PR.

PR #2154, opened 2019-06-07
pkg/audit: Rename verifier test files

navillasa:
What:
- renames `verifier_test.go` to `verifier_unit_test.go` since it had only unit tests
- renames `newverifier_test.go` to `verifier_test.go`
- that's it 

Why:
- seemed weird to have `verifier_test.go` and `newverifier_test.go` because the distinction wasn't clear.
- I guess they could be added to one file, but it would be ~800 loc so this seemed like an okay compromise.
- `verifier_test.go` is already kind of long, and maybe could be broken into a `downloadshares_test.go` file, but maybe that's too many test files 🤷🏻‍♀️

Thanks for submitting a PR!


navillasa:
> Did you use 'git mv'? I would have expected the "Files changed" info to be more straight forward.

@wthorp oh shoot I forgot `git mv` is a thing. is it okay as is?

thepaul:
`git mv` is essentially just `git rm oldname && git add newname`, with better sanity checking. It doesn't add any metainfo indicating the name change. When github doesn't recognize slightly-changed moved files, like this, you can always check it out locally and use something like `git diff -M75` to see the changes more clearly.

wthorp:
Thanks for clarifying that Paul!

PR #2155, opened 2019-06-07
Enforce our Minimum Requirements for Node Operators and sanity check them

stefanbenten:
What: Enforces 500GB Bandwidth and Diskspace to ensure we have proper nodes on the network.

Why: Windows and OSX have issues with directory mapping sometimes, but the node still starts up.
To prevent writing to a small internal docker directory (<5GB), we check that the space is actually at least 500GB.

Thanks for submitting a PR!


storagenode/monitor/monitor.go
```diff
@@ -107,6 +112,16 @@ func (service *Service) Run(ctx context.Context) (err error) {
 		service.log.Warn("Disk space is less than requested. Allocating space", zap.Int64("bytes", service.allocatedDiskSpace))
 	}
 
+	// Ensure the disk is at least 500GB in size, which is our current minimum required to be an operator
+	if service.allocatedDiskSpace < minimumDiskSpace {
+		service.log.Fatal("Total disk space less than required minimum", zap.Int64("bytes", minimumDiskSpace))
```
jtolds: don't use fatal. we should return an error

stefanbenten: Does that stop the node directly?

jtolds: i don't know, but functions should not as a side effect call os.exit. if a service dies, we should have the service running in the peer know that certain errors are fatal errors. but it should be up to the peer code about whether to exit. having this run method decide the process should end is overstepping it's contractual bounds in my opinion

jtolds: so i think if you want the node to die you should check out the peer code in the storage node and make sure that if the monitor Run method errors that it kills the storage node

jtolds: maybe the error group the peer has should have two types of errors, non-fatal and fatal? and this is a fatal error.

littleskunk:
Test looks good.
```
2019-06-09T17:22:50.603-0400	ERROR	piecestore:monitor	Total disk space less than required minimum	{"bytes": 500000000000}
Fatal error: piecestore monitor: disk space requirement not met
```

internal/testplanet/planet.go
```diff
@@ -619,6 +620,10 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatelliteIDs []strin
 					Interval: time.Hour,
 					Timeout:  time.Hour,
 				},
+				Monitor: monitor.Config{
+					MinimumBandwidth: memory.GB * 500,
+					MinimumDiskSpace: memory.GB * 25,
```
egonelbre: For testplanet, let's set it to 0.

egonelbre: Or something like `100MB`.

stefanbenten: Done

internal/testplanet/planet.go
```diff
@@ -619,6 +620,10 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatelliteIDs []strin
 					Interval: time.Hour,
 					Timeout:  time.Hour,
 				},
+				Monitor: monitor.Config{
+					MinimumBandwidth: memory.MB * 100,
```
egonelbre: `100 * memory.MB` usually reads better :).

PR #2156, opened 2019-06-07
Moved monkit tracing after the limit nil check & fix node id formatting.

simongui:
# What 
- Fixed a crash that could occur if limit was nil.
- Fixed the node id format in the monkit tracing to display nicer.

# Why
In some cases in `client.go:putPiece()` it was possible for limit to be nil and this was causing a crash when the monkit line was trying to include the Node ID.

Thanks for submitting a PR!


pkg/storage/ec/client.go
```diff
@@ -274,14 +274,15 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
...
 func (ec *ecClient) putPiece(ctx, parent context.Context, limit *pb.AddressedOrderLimit, data io.ReadCloser, expiration time.Time) (hash *pb.PieceHash, err error) {
-	defer mon.Task()(&ctx, "node: ", limit.GetLimit().StorageNodeId.String()[0:8])(&err)
 	defer func() { err = errs.Combine(err, data.Close()) }()
 
 	if limit == nil {
 		_, _ = io.Copy(ioutil.Discard, data)
 		return nil, nil
 	}
 
+	defer mon.Task()(&ctx, "node: "+limit.GetLimit().StorageNodeId.String()[0:8])(&err)
```
thepaul: (A) I think it would still be worth keeping records on calls to this function and how long they take, even in the limit=nil case.
(B), moving this defer after the other means switching the order they get run, so errors from data.Close() would not be reported correctly

How about putting something like this at the top instead?

    nodeName := "nil"
    if limit != nil {
        nodeName = limit.GetLimit().StorageNodeId.String()[0:8]
    }
    defer mon.Task()(&ctx, "node: "+nodeName)(&err)

jtolds: :+1:

simongui: I like that idea! I've updated it.

pkg/storage/ec/client.go
```diff
@@ -274,7 +274,11 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
 }
 
 func (ec *ecClient) putPiece(ctx, parent context.Context, limit *pb.AddressedOrderLimit, data io.ReadCloser, expiration time.Time) (hash *pb.PieceHash, err error) {
-	defer mon.Task()(&ctx, "node: ", limit.GetLimit().StorageNodeId.String()[0:8])(&err)
+	nodeName := "nil"
```
JessicaGreben: you can initialize to a zero value with `var nodeName string`

simongui: By using `nil` we can search in the traces for empty traces a lot easier or count occurrences easier than if we just leave it blank.

PR #2157, opened 2019-06-07
dbutil: reduce defaults so sum is < 100

jtolds:
What: Reduce the defaults

Why: Michal says this helps! 

> I ran this test once again with modified open/idle connection setting for our DB. I set it to 60/30 open/idle (I read somehere that sum needs to be less than postgres max, default 100) and after more than 4h it still uploading (earlier after 2h I had "too many clients already" and satellite was unresponsive). I uploaded more than 130k files and its still working.


stefanbenten:
As i am not reaching this limit on my database, i cannot report if thats actually accurate :)

stefanbenten:
@jtolds Also dont forget to alert our channel for config-changes please!

mniewrzal:
Egon had a good suggestion that we have two db pools (master + pointerdb) so with this limit, we still can hit `"too many clients already"`. Maybe we can set even half of the new settings but maybe someone with better postgres knowledge can say more about it.

stefanbenten:
@mniewrzal Reduced the numbers even more. They should be still more than plenty. 👍

mniewrzal:
:+1

mniewrzal:
:+1:

PR #2158, opened 2019-06-08
add storage node voucher request service

cam-a:
What: Add a service on the storage node to request vouchers from satellites and insert them into a DB table

Why: For kademlia audit gating. To prevent sybil attacks
https://storjlabs.atlassian.net/browse/V3-1868

(Discussion)
- How many trusted satellites do we expect a storage node to potentially have in the future? This may impact how we want to design db queries

- Right now, the voucher service iterates through the trusted satellites list and calls `NeedVoucher` on each satellite, which determines if a voucher is needed based on whether a voucher exists for this satellite and if that voucher is going to expire. With this method, the number of queries equals the number of satellites in the trusted list (could be thousands in the future?)

- In order to cut down on queries, we could query the vouchersDB for all expiring vouchers, but we would also need to determine which satellites in the trusted list are not in the vouchersDB yet, which would require making another query to return all satellites in the vouchersDB and filter them from the trusted satellites list

- Another idea is to only make one query, which returns all satellite ids and expirations from the table. Since in the idea above we need to return all satellite ids in the table, why not just return all satellite ids and expirations with one query? However, this could be quite large. cpu resources vs db resources?

Thanks for submitting a PR!


internal/testplanet/planet.go
```diff
@@ -51,13 +51,14 @@ import (
...
 	"storj.io/storj/satellite/satellitedb"
-	"storj.io/storj/satellite/vouchers"
+	satVouchers "storj.io/storj/satellite/vouchers"
 	"storj.io/storj/storagenode"
 	"storj.io/storj/storagenode/collector"
 	"storj.io/storj/storagenode/monitor"
 	"storj.io/storj/storagenode/orders"
 	"storj.io/storj/storagenode/piecestore"
 	"storj.io/storj/storagenode/storagenodedb"
+	"storj.io/storj/storagenode/vouchers"
```
jenlij: can we name this one snVouchers?

satellite/vouchers/service.go
```diff
@@ -49,29 +49,29 @@ func NewService(log *zap.Logger, satellite signing.Signer, cache *overlay.Cache,
...
 
 // Request receives a voucher request and returns a voucher and an error
-func (service *Service) Request(ctx context.Context, req *pb.VoucherRequest) (_ *pb.Voucher, err error) {
+func (service *Service) Request(ctx context.Context, req *pb.VoucherRequest) (_ *pb.VoucherResponse, err error) {
 	defer mon.Task()(&ctx)(&err)
 
 	peer, err := identity.PeerIdentityFromContext(ctx)
 	if err != nil {
-		return &pb.Voucher{}, Error.Wrap(err)
+		return nil, Error.Wrap(err)
```
jenlij: Why are we returning nil here instead of an empty voucher? Just wondering

cam-a: Not sure if there is a particular reason to choose one over the other

storagenode/peer.go
```diff
@@ -110,6 +114,9 @@ type Peer struct {
 		Sender    *orders.Sender
 	}
 
+	//TODO: Where to put this? Storage2/Kademlia?
```
jenlij: What do you mean?

cam-a: Other services have been compartmentalized into the Storage2 and Kademlia structs, but Vouchers don't really fall into either of those so, I could probably get rid of this comment

internal/testplanet/planet.go
```diff
@@ -517,12 +518,13 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				PasswordCost:    console.TestPasswordCost,
 				AuthTokenSecret: "my-suppa-secret-key",
 			},
+			Vouchers: satVouchers.Config{
```
jenlij: Remember to let devops know about these config changes in the config-changes slack channel

satellite/vouchers/service.go
jenlij: Yeah the internet says it doesn't really matter, but then some people say nil is better than a pointer to an empty struct I think because it may be clearer. But in either case, you're fine here!

storagenode/peer.go
jenlij: ok yeah even though vouchers are tied to kad, it's kind of its own thing

storagenode/storagenodedb/vouchers.go
```diff
@@ -56,35 +56,27 @@ func (db *vouchersdb) Put(ctx context.Context, voucher *pb.Voucher) (err error)
 	return err
 }
 
-// GetExpiring retrieves all vouchers that are expired or about to expire
-func (db *vouchersdb) GetExpiring(ctx context.Context, expirationBuffer time.Duration) (satellites []storj.NodeID, err error) {
+// NeedVoucher returns true if a satellite's voucher is expired/about to expire or does not exist
```
jenlij: super nit: "NeedVoucher returns true if a voucher from a particular satellite is expired, about to expire, or does not exist"

storagenode/vouchers/vouchers_test.go
```diff
@@ -0,0 +1,204 @@
...
+		expiration, err := ptypes.TimestampProto(time.Now().UTC().Add(24 * time.Hour))
+		require.NoError(t, err)
+
+		voucher := &pb.Voucher{
+			SatelliteId:   satellite.ID,
+			StorageNodeId: storagenode.ID,
+			Expiration:    expiration,
+		}
+
+		// GetValid with no entry
```
jenlij: Could we maybe elaborate more on these test case descriptions? Like "Tests that GetValid returns nil when there is no voucher corresponding to satellite.ID"

storagenode/vouchers/vouchers_test.go
```diff
@@ -0,0 +1,204 @@
...
+		result, err := vdb.GetValid(ctx, []storj.NodeID{satellite.ID})
+		require.NoError(t, err)
+		assert.Nil(t, result)
+
+		// basic Put test
+		err = vdb.Put(ctx, voucher)
+		require.NoError(t, err)
+
+		// basic NeedVoucher test
+		expirationBuffer := 48 * time.Hour
```
jenlij: Why does case result in NeedVoucher true?

jenlij:
Going through the checklist, I realized we should update the design doc and or create new documentation with all the audit gating information once that's all implemented.

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+}
+
+// Service is a service for requesting vouchers
+type Service struct {
+	log *zap.Logger
+
+	kademlia  *kademlia.Kademlia
+	transport transport.Client
+
+	vdb   DB
```
cam-a: call voucherDB

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
-	GetExpiring(context.Context, time.Duration) ([]storj.NodeID, error)
 	// GetValid returns one valid voucher from the list of approved satellites
 	GetValid(context.Context, []storj.NodeID) (*pb.Voucher, error)
+	// NeedVoucher returns true if a voucher from a particular satellite is expired, about to expire, or does not exist
+	NeedVoucher(context.Context, storj.NodeID, time.Duration) (bool, error)
+}
+
+// Config defines configuration for requesting vouchers.
+type Config struct {
+	Interval         int `help:"number of days between voucher service iterations" default:"7"`
```
JessicaGreben: if interval is 7 days then that means a new node can never be vetted faster than 7 days. Is that what we want? Seems long to me.

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+	ExpirationBuffer int `help:"buffer period of X days into the future. If a voucher would expire within this period, send a request for renewal" default:"7"`
```
JessicaGreben: we might want to hard code `ExpirationBuffer` to be some value longer than `Interval`

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+	defer mon.Task()(&ctx)(&err)
+	return service.Loop.Run(ctx, service.RunOnce)
+}
+
+// RunOnce runs one iteration of the voucher request service
+func (service *Service) RunOnce(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	service.log.Info("Checking vouchers")
+
+	satellites := service.trust.GetSatellites((ctx))
```
JessicaGreben: can we call this `trustedSatellites`?

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+
+	if len(satellites) > 0 {
+		var group errgroup.Group
+		ctx, cancel := context.WithTimeout(ctx, time.Hour)
+		defer cancel()
+		for _, sat := range satellites {
```
JessicaGreben: can we call this `satellite` instead of `sat`?

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+			sat := sat
+			need, err := service.vdb.NeedVoucher(ctx, sat, service.expirationBuffer)
```
JessicaGreben: can we call this `needVoucher`

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+				group.Go(func() error {
+					service.Request(ctx, sat)
+					return nil
+				})
+			}
+		}
+		_ = group.Wait() // doesn't return errors
+	} else {
+		service.log.Debug("No voucher requests to send")
+	}
```
JessicaGreben: can we change this to check `if len(satellites) == 0; service.log.Debug("No voucher requests to send"); return nil` first, then we dont need the rest indented

storagenode/vouchers/service.go
```diff
@@ -7,16 +7,165 @@ import (
...
+
+	resp, err := pb.NewVouchersClient(conn).Request(ctx, &pb.VoucherRequest{})
+	if err != nil {
+		return VoucherError.New("failed to start request: %v", err)
+	}
+
+	// TODO: separate status for disqualified nodes?
+	switch resp.GetStatus() {
+	case pb.VoucherResponse_REJECTED:
+		service.log.Info("Voucher request denied. Node does not meet voucher requirements")
```
JessicaGreben: is there anyway we can provide more info about why it was rejected? Even if we can pass the specific values, but just generally what is being checked

storagenode/vouchers/vouchers_test.go
```diff
@@ -0,0 +1,204 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
```
JessicaGreben: yay so many tests!

storagenode/vouchers/vouchers_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+		// run service and check vouchers have been received
+		err := node.Vouchers.RunOnce(ctx)
+		require.NoError(t, err)
+
+		for _, sat := range planet.Satellites {
+			voucher, err := node.DB.Vouchers().GetValid(ctx, []storj.NodeID{sat.ID()})
+			require.NoError(t, err)
+			assert.NotNil(t, voucher)
+		}
```
JessicaGreben: this is totally unnecessary since your test coverage is already amazing, but just as a fun idea it might be cool to write a test that runs the voucher service when the storagenode is not vetted, then runs it again later and show that it is vetted

storagenode/vouchers/vouchers_test.go
```diff
@@ -0,0 +1,204 @@
...
+	})
+}
+
+func TestVerifyVoucher(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 2, StorageNodeCount: 1, UplinkCount: 1,
```
JessicaGreben: do you need uplink?

storagenode/vouchers/vouchers_test.go
cam-a: The voucher's expiration, `time.Now().UTC().Add(24 * time.Hour)` falls within the expiration buffer of 48 hours from now

storagenode/vouchers/service.go
jtolds: can we make this be a time.Duration instead of an int plz? the default value (if 7 days) would then be "168h"

storagenode/vouchers/service.go
jtolds: also can this be a time.Duration instead of an int?

storagenode/vouchers/service.go
cam-a: I was torn between time.Duration for code simplicity and int for simplicity of user configuration, but changing it sounds good!

storagenode/vouchers/service.go
cam-a: Jessica and I were thinking it might be better to hard-code this to be == `Interval` or `Interval` + (some extra duration). What do you think?

pkg/pb/vouchers.proto
```diff
@@ -21,6 +21,19 @@ message Voucher {
 
 message VoucherRequest {}
 
+message VoucherResponse {
+    enum Status {
+        INVALID  = 0;
```
JessicaGreben: do we ever use this invalid status? seems like it might not be needed

pkg/pb/vouchers.proto
```diff
@@ -21,6 +21,19 @@ message Voucher {
...
+    Voucher voucher = 1;
+
+    Status status = 2;
+}
+
+
 service Vouchers {
-    rpc Request(VoucherRequest) returns (Voucher) {}
+    rpc Request(VoucherRequest) returns (VoucherResponse) {}
 }
```
JessicaGreben: add new line at end please

storagenode/peer.go
```diff
@@ -265,6 +271,29 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, ver
...
+		intervalDuration, err := time.ParseDuration(fmt.Sprintf("%dh", intervalHours))
+		if err != nil {
+			return nil, err
+		}
+
+		if config.Vouchers.ExpirationBuffer <= 0 {
+			return nil, errs.New("voucher service expiration buffer (%d) must be > 0", config.Vouchers.ExpirationBuffer)
+		}
+		bufferHours := config.Vouchers.ExpirationBuffer * 24
+		bufferDuration, err := time.ParseDuration(fmt.Sprintf("%dh", bufferHours))
```
jtolds: you don't have to do any of this if you just make the config type be a time.Duration off the bat. our config system knows how to handle time.Durations

storagenode/vouchers/service.go
JessicaGreben: im running storj-sim right now and its not clear what this message means

```
DEBUG	vouchers	vouchers/service.go:107	No voucher requests to send
```
Maybe we could add something like, "No trusted satellites so not requesting any vouchers to approve storing data".  What do you think?

pkg/pb/vouchers.proto
cam-a: Yeah, we can probably get rid of it

internal/testplanet/planet.go
```diff
@@ -517,12 +518,13 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
+				Expiration: 30,
```
jtolds: is this a time.Duration? if it is, then this means 30 nanoseconds. if it's not a time.Duration, can we make it one, and then change this to 30*24*time.Day?

storagenode/peer.go
```diff
@@ -265,6 +271,17 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, ver
 		)
 	}
 
+	{ // setup vouchers
+		interval := config.Vouchers.Interval
+		bufferHours := interval.Hours() + 1.0
+		bufferDuration, err := time.ParseDuration(fmt.Sprintf("%fh", bufferHours))
+		if err != nil {
+			return nil, errs.Combine(err, peer.Close())
+		}
```
jtolds: you can just do bufferDuration := interval + time.Hour?

internal/testplanet/planet.go
cam-a: ok

pkg/pb/vouchers.proto
jtolds: it's nice to have the zero value default to "invalid"

jtolds: basically, if you construct this protobuf in go, all the values are initialized to zero values, and it's nice in many cases to have the status default to something besides something you opted in to

PR #2162, opened 2019-06-10
Fix smtp sender return error on close

rikysya:
What: Fix smtp sender returns an error because of incorrect resources closing

Why: 

Thanks for submitting a PR!


internal/post/sender.go
```diff
@@ -45,8 +46,11 @@ func (sender *SMTPSender) SendEmail(ctx context.Context, msg *Message) (err erro
 		return err
 	}
 	// close underlying connection
+	// if any unexpected error occurred
 	defer func() {
-		err = errs.Combine(err, client.Close())
+		if err != nil {
```
egonelbre: Currently when an error happens earlier then the connection won't be closed.

internal/post/sender.go
```diff
@@ -73,26 +77,31 @@ func (sender *SMTPSender) SendEmail(ctx context.Context, msg *Message) (err erro
 		}
 	}
 
-	data, err := client.Data()
+	mess, err := msg.Bytes()
```
egonelbre: Why was this moved upward?

internal/post/sender.go
egonelbre: It feels like it always should call `client.Close`, but depending on the outer error might ignore it.

rikysya: @egonelbre  if there is no error that means we called `client.Quit` successfully, which closes the connection. If error some error occurs during sending process then we close it manually in defer

internal/post/sender.go
rikysya: for `data` to be closer to function that invokes it

internal/post/sender.go
rikysya: could you please describe a situation when it won't be closed?

egonelbre: Let's say msg.Bytes() returns a parsing error.

rikysya: @egonelbre so then we return this `err`, defer is called with non-nil `err` and we close the connection, right? May be I am not quite following you

egonelbre: if the error is non-nil it doesn't mean the underlying connection is closed

internal/post/sender.go
```diff
@@ -35,22 +36,27 @@ func (sender *SMTPSender) FromAddress() Address {
...
-	// close underlying connection
-	defer func() {
-		err = errs.Combine(err, client.Close())
-	}()
+
+	if err = sender.communicate(ctx, client, msg); err != nil {
+		return errs.Combine(err, client.Close())
+	}
+
+	return nil
```
egonelbre: keep the `client.Quit` here, that way it's more clear that it gets closed, also that way you can remove write data and keep it inside

rikysya: Then I need to remove `StartTls` and `Auth` out from there, so then I can possibly reuse communicate method on open connection, because calling `client.Quit` here makes me think that I can still use the `client` before quitting it. Another reason is that if `client.Quit` returns an`error` than `client.Close` should be called explicitly  

`writedata` makes sure that `writer` is closed immediately  after it is used, also it makes `error` more clear as `writer.Close` error will be combined only with `writer.Write` error

PR #2163, opened 2019-06-10
add const stripesPerBlock const to calc blocksize

JessicaGreben:
**What/Why:**
This primary change in this PR does the following:
* adds a constant `stripesPerBlock` to the uplink/config.go that is used to calculate blocksize. We want stripes to be a multiple of encryption blocksize so that performance for seeking objects doesn't pull down extra unneeded data.
* reduces default erasure share size to 256B.  Originally this was 1KiB which seems to be too large since to get stripeSize is this value times the minThreshold setting. We will set this to 256B for now, but will require more benchmarking in the future to confirm ideal settings. This means for the dev config the related settings will be the following:
-- stripe size is 1KiB 
-- encryption blocksize is 2Kib, 
and for release config:
-- stripe size is 7424B and 
-- encryption blocksize is 14,848B.

This PR also adds are comments to add more clarity and this adds the average speed to the progressbar for uploads/downloads since that is what aws does and its useful to see. 

Ref: convo here on a different PR about adding stripesPerBlock https://github.com/storj/storj/pull/2103#discussion_r291353266


uplink/config.go
```diff
@@ -155,9 +155,10 @@ func (c Config) GetPathCipherSuite() storj.CipherSuite {
 
 // GetEncryptionScheme returns the configured encryption scheme for new uploads
 func (c Config) GetEncryptionScheme() storj.EncryptionScheme {
+	const stripesPerBlock = 2
 	return storj.EncryptionScheme{
 		Cipher:    storj.Cipher(c.Enc.DataType),
-		BlockSize: c.GetRedundancyScheme().StripeSize(),
+		BlockSize: c.GetRedundancyScheme().StripeSize() * stripesPerBlock,
```
jtolds: i don't know where else this would matter (tests maybe?) but in your original PR wasn't there more than one place where the block size was set to the stripe size?

JessicaGreben: good point, let me make a run through the tests and check they reflect this change

PR #2164, opened 2019-06-10
Audit service should download erasure shares in parallel

kaloyan-raev:
What: 

Audit service will download erasure shares in parallel (each share in a separate goroutine) instead of serially (one-by-one).

Why:

Better performance.

Thanks for submitting a PR!

### DIFF
--- a/pkg/audit/verifier.go
+++ b/pkg/audit/verifier.go
@@ -39,6 +39,7 @@ var (
 type Share struct {
 	Error    error
 	PieceNum int
+	NodeID   storj.NodeID
 	Data     []byte
 }
 
@@ -91,7 +92,7 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
 		verifier.log.Debug("Verify: order limits not created for some nodes (offline)", zap.Strings("Node IDs", offlineNodes.Strings()))
 	}
 
-	shares, nodes, err := verifier.DownloadShares(ctx, orderLimits, stripe.Index, shareSize)
+	shares, err := verifier.DownloadShares(ctx, orderLimits, stripe.Index, shareSize)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/audit/verifier.go
```diff
@@ -225,32 +226,40 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
...
-		}
+			ch <- &share
+		}(i, limit)
+	}
 
-		shares[share.PieceNum] = share
-		nodes[share.PieceNum] = limit.GetLimit().StorageNodeId
+	for range limits {
+		share := <-ch
+		if share != nil {
```
ifraixedes: Can a share be `nil` here? or only if there is a bug in `verifier.GetShare`?

kaloyan-raev: Yes. Look at line 237. `limits` is ordered by piece number. So it can have `nil` values for those piece numbers that no storage node is storing a piece.

pkg/audit/verifier.go
```diff
@@ -225,32 +226,40 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
 }
 
 // DownloadShares downloads shares from the nodes where remote pieces are located
-func (verifier *Verifier) DownloadShares(ctx context.Context, limits []*pb.AddressedOrderLimit, stripeIndex int64, shareSize int32) (shares map[int]Share, nodes map[int]storj.NodeID, err error) {
+func (verifier *Verifier) DownloadShares(ctx context.Context, limits []*pb.AddressedOrderLimit, stripeIndex int64, shareSize int32) (shares map[int]Share, err error) {
 	defer mon.Task()(&ctx)(&err)
 
 	shares = make(map[int]Share, len(limits))
-	nodes = make(map[int]storj.NodeID, len(limits))
+	ch := make(chan *Share, len(limits))
```
Fadila82: Out of curiosity, why use channels and not waitGroup?

kaloyan-raev: Because of my limited knowledge in Golang. If you can do this better with WaitGroup, feel free to create a PR.

PR #2169, opened 2019-06-10
add db implementation for user_credits table

VinozzZ:
What: 
database implementation for user_credits table
Why:

user_credit table will need to store all user's credits information. console needs access to those data in order to display how many credits users have earned and used through marketing programs
Link to Jira: https://storjlabs.atlassian.net/browse/MRTCH-33?atlOrigin=eyJpIjoiOWI3MzQ4MDZjMmFkNGY4ZTk2N2VlZWJiYjZmODBhOTkiLCJwIjoiaiJ9

Thanks for submitting a PR!


satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,142 @@
...
+	availableCredits, err := tx.All_UserCredit_By_UserId_And_ExpiresAt_Greater_And_CreditsUsedInCents_Less_CreditsEarnedInCents_OrderBy_Asc_ExpiresAt(ctx,
+		dbx.UserCredit_UserId(id[:]),
+		dbx.UserCredit_ExpiresAt(expirationEndDate),
+	)
+	if err != nil {
+		return errs.Wrap(errs.Combine(err, tx.Rollback()))
+	}
+
+	for _, credit := range availableCredits {
+		if appliedCredits == 0 {
```
coyle: `appliedCredits -= credit.CreditsEarnedInCents - credit.CreditsUsedInCents` could be less than 0?

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	referredByID, err := bytesToUUID(userCreditDBX.ReferredBy)
+	if err != nil {
+		return nil, err
+	}
+
+	credit := console.UserCredit{
```
coyle: nit --
```
return &console.UserCredit{ ...}, nil
```

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+func test(ctx context.Context, t *testing.T, store satellite.DB) {
+	consoleDB := store.Console()
+
+	user, referrer, offer, err := setupData(ctx, store)
+	require.NoError(t, err)
+	randomID, err := uuid.New()
+	require.NoError(t, err)
+
+	var inValidUserCredits = []struct {
```
coyle: Awesome!

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+}
+
+func setupData(ctx context.Context, store satellite.DB) (user *console.User, referrer *console.User, offer *marketing.Offer, err error) {
+	consoleDB := store.Console()
+	marketingDB := store.Marketing()
+	// create user
+	var userPassHash [8]byte
+	_, err = rand.Read(userPassHash[:])
```
coyle: `require.NoError(t, err)` ?

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	var referrerPassHash [8]byte
+	_, err = rand.Read(referrerPassHash[:])
```
coyle: `require.NoError(t, err)` ?

satellite/satellitedb/usercredits.go
VinozzZ: The statement on line 84 should make sure that never happen

satellite/satellitedb/usercredits.go
```diff
@@ -73,28 +79,54 @@ func (c *usercredits) UpdateAvailableCredits(ctx context.Context, appliedCredits
...
+	var infos []updateInfo
 
 	for _, credit := range availableCredits {
-		if appliedCredits == 0 {
+		if chargedCredits == 0 {
 			break
 		}
 
-		updatedUsedCredit := credit.CreditsEarnedInCents
+		updateCredit := credit.CreditsEarnedInCents - credit.CreditsUsedInCents
```
fuskovic: Should this be `updatedCredit` ?

VinozzZ: It's not updated yet at this point so I don't want to name it as `updated`. Maybe I will change it to `creditsForUpdate` or something like that

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func test(ctx context.Context, t *testing.T, store satellite.DB) {
+	consoleDB := store.Console()
+
+	user, referrer, offer := setupData(ctx, t, store)
+	randomID, err := uuid.New()
+	require.NoError(t, err)
+
+	var inValidUserCredits = []console.UserCredit{
```
mobyvb: `inValid` -> `invalid`
Also, some comments describing what makes each case invalid would be very beneficial imo

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{
+			userCredit: console.UserCredit{
+				UserID:               user.ID,
+				OfferID:              offer.ID,
+				ReferredBy:           referrer.ID,
+				CreditsEarnedInCents: 100,
+				ExpiresAt:            time.Now().UTC().AddDate(0, 1, 0),
+			},
+			chargedCredits: 120,
+			expected:       0,
```
mobyvb: it doesn't seem like expected should be 0 here to me. Either they should have -20 credits, or they should still have 100 and an error or something should happen when we try to charge them more than they have

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{
+			userCredit: console.UserCredit{
+				UserID:               user.ID,
+				OfferID:              offer.ID,
+				ReferredBy:           referrer.ID,
+				CreditsEarnedInCents: 100,
+				ExpiresAt:            time.Now().UTC().AddDate(0, 0, -5),
+			},
+			chargedCredits: 60,
+			expected:       0,
```
mobyvb: similarly, I think expected should be nonzero here as well. Seems like it should be 40 (if I understand how this works, which I might not)

satellite/satellitedb/usercredits_test.go
VinozzZ: Good point! I changed the API to always return the remaining charges so the caller can know if the charge is complete or there are some leftover

satellite/satellitedb/usercredits_test.go
VinozzZ: I added a comment for this entry to highlight that it should not be applied since the `expires_at` is set to be before `Now()`

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+	var infos []updateInfo
+	remainingCharge = creditsToCharge
+	for _, credit := range availableCredits {
+		if remainingCharge == 0 {
+			break
+		}
+
+		creditsForUpdate := credit.CreditsEarnedInCents - credit.CreditsUsedInCents
+
+		if remainingCharge < creditsForUpdate {
```
coyle: what if they are equal ?

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+		infos = append(infos, updateInfo{
+			id:      credit.Id,
+			credits: creditsForUpdate,
+		})
+
+		remainingCharge -= creditsForUpdate
+	}
+
+	statement := `UPDATE user_credits SET
+		credits_used_in_cents = CASE id ` + convertToSQLFormat(infos)
```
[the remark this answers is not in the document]
VinozzZ: I know the Jenkins is not working right now. But locally the tests are passing in both Postgres and sqlite

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+	_, err = tx.Tx.ExecContext(ctx, c.db.Rebind(statement))
+	if err != nil {
+		return creditsToCharge, errs.Wrap(errs.Combine(err, tx.Rollback()))
+	}
+	return remainingCharge, errs.Wrap(tx.Commit())
+}
+
+func convertToSQLFormat(updateInfo []updateInfo) (updateStr string) {
+	for i, info := range updateInfo {
+		updateStr += `WHEN ` + strconv.Itoa(info.id) + ` THEN ` + strconv.Itoa(info.credits)
```
egonelbre: No string addition of values for sql queries, it's easy to get the escaping wrong, use statement arguments. While it's probably fine here, I don't want to have a situation where somebody modifies this statement and accidentally adds something that causes a SQL injection.

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return creditsToCharge, errs.Wrap(errs.Combine(err, tx.Rollback()))
+	}
+	if len(availableCredits) == 0 {
+		return creditsToCharge, errs.Combine(errs.New("No available credits"), tx.Commit())
+	}
+
+	var infos []updateInfo
+	remainingCharge = creditsToCharge
+	for _, credit := range availableCredits {
```
egonelbre: What happens if there are many concurrent queries? Can this be written without transactions?

satellite/satellitedb/usercredits.go
egonelbre: See https://www.owasp.org/index.php/SQL_Injection for more information.

satellite/satellitedb/usercredits.go
VinozzZ: The update happens after a user has paid their bill. Will there be a scenario that user pays multiple times in a row?

egonelbre: Unlikely. But, I would like as much of the code work with concurrent requests... it helps to avoid debugging weird scenarios.

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+func (c *usercredits) GetAvailableCredits(ctx context.Context, referrerID uuid.UUID, expirationEndDate time.Time) ([]console.UserCredit, error) {
+	availableCredits, err := c.db.All_UserCredit_By_UserId_And_ExpiresAt_Greater_And_CreditsUsedInCents_Less_CreditsEarnedInCents_OrderBy_Asc_ExpiresAt(ctx,
+		dbx.UserCredit_UserId(referrerID[:]),
+		dbx.UserCredit_ExpiresAt(expirationEndDate),
+	)
+	if err != nil {
+		return nil, errs.Wrap(err)
+	}
+
+	return fromDBX(availableCredits)
```
egonelbre: Let's use a better name than `fromDBX` this package contains other implementations. e.g. `userCreditsFromDBX`

satellite/satellitedb/usercredits.go
```diff
@@ -0,0 +1,172 @@
...
+		dbx.UserCredit_ExpiresAt(userCredit.ExpiresAt),
+		dbx.UserCredit_Create_Fields{
+			ReferredBy: dbx.UserCredit_ReferredBy(userCredit.ReferredBy[:]),
+		},
+	)
+	if err != nil {
+		return nil, errs.Wrap(err)
+	}
+
+	return convertDBX(createdUser)
```
egonelbre: Let's use a better name than `convertDBX`

satellite/satellitedb/usercredits.go
egonelbre: It seems like this could be done as a single query for Postgres, leaving tx for sqlite won't be as much of a problem.

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	satellitedbtest.Run(t, func(t *testing.T, db satellite.DB) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+
+		test(ctx, t, db)
+	})
+}
+
+func test(ctx context.Context, t *testing.T, store satellite.DB) {
```
egonelbre: no need for the separate func

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,202 @@
...
+		Email:        "john@example.com",
+		PasswordHash: userPassHash[:],
+		Status:       console.Active,
+	})
+	require.NoError(t, err)
+
+	//create an user as referrer
+	referrer, err = consoleDB.Users().Insert(ctx, &console.User{
+		FullName:     "referrer",
+		Email:        "referrer@example.com",
```
egonelbre: `mail.test`

satellite/satellitedb/usercredits_test.go
```diff
@@ -0,0 +1,202 @@
...
+	require.NoError(t, err)
+
+	var referrerPassHash [8]byte
+	_, err = rand.Read(referrerPassHash[:])
+	require.NoError(t, err)
+
+	// create an user
+	user, err = consoleDB.Users().Insert(ctx, &console.User{
+		FullName:     "John Doe",
+		Email:        "john@example.com",
```
egonelbre: `mail.test`

satellite/satellitedb/usercredits.go
egonelbre: but i guess we can leave it as a tx for now.

VinozzZ: I had to make separate queries for `sqlite3` and `postgres` to use `statement` to pass in values. Maybe I will try to just use single query for Postgres and leave the sqlite for tx

satellite/satellitedb/usercredits.go
VinozzZ: I made the change

satellite/satellitedb/usercredits.go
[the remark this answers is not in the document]
VinozzZ: It's working now! I had to make a change to avoid sql injection

PR #2177, opened 2019-06-12
Fix marketingweb code

egonelbre:
* Only listen on 127.0.0.1 (see https://github.com/storj/storj/wiki/Testing#automatically-choose-ports)
* Move bundled files to the same folder as all other files.
* Server shouldn't panic on misconfigured template.
* Fix path handling. For filepaths `filepath` is the correct thing. String addition for filepath handling is problematic.
* Don't rewrite content on template writing error. Writing header twice can cause problems or continuing from a failed write can cause problems.
* Write correct error code. `404` is for not found, correct one in this case is internal server error. (See https://www.codetinkerer.com/2015/12/04/choosing-an-http-status-code.html)

Not fixed in this one:
* Templates shouldn't be parsed on every request for a production server.

Thanks for submitting a PR!


satellite/marketing/marketingweb/server.go
```diff
@@ -32,65 +32,89 @@ type Server struct {
 
 	listener net.Listener
 	server   http.Server
+
+	templateDir string
+}
+
+func (s *Server) templateFilePath(name string) string {
```
ifraixedes: This method isn't used, I would delete it.

satellite/marketing/marketingweb/server.go
```diff
@@ -32,65 +32,89 @@ type Server struct {
...
+	return s
 }
 
-// appHandler is web app http handler function
-func (s *Server) appHandler(w http.ResponseWriter, req *http.Request) {
+// ServeHTTP handles index request
+func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if req.URL.Path != "/" {
-		s.serveError(w, req)
+		s.serveInternalError(w, req)
```
ifraixedes: Why the correct one is an internal server error when the path doesn't match the ones exposed?

egonelbre: Ah, right... yeah, my mistake.

PR #2180, opened 2019-06-12
only parse templates for the first requests

VinozzZ:
What: 
add a global templates variable to hold parsed templates
Why:
templates shouldn't be parsed on every requests

Thanks for submitting a PR!


satellite/marketing/marketingweb/server.go
```diff
@@ -16,7 +17,10 @@ import (
 )
 
 // Error is satellite marketing error type
-var Error = errs.Class("satellite marketing error")
+var (
+	Error     = errs.Class("satellite marketing error")
+	templates *template.Template
```
egonelbre: This will have race condition, shouldn't be global.

satellite/marketing/marketingweb/server.go
```diff
@@ -74,22 +80,25 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-	)
-
-	home, err := template.New("landingPage").ParseFiles(files...)
-	if err != nil {
-		s.serveInternalError(w, req)
-		return
+	var err error
+
+	if templates == nil {
+		files := append(s.commonPages(),
```
egonelbre: This should be in `NewServer` not ServeHTTP.

satellite/marketing/marketingweb/server.go
VinozzZ: Good point. I changed it to be on `Server`

satellite/marketing/marketingweb/server.go
```diff
@@ -83,13 +76,27 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		filepath.Join(s.templateDir, "foModal.html"),
 	)
 
-	home, err := template.New("landingPage").ParseFiles(files...)
+	s.templates, err = template.New("landingPage").ParseFiles(files...)
```
egonelbre: This should be propagated up, if this fails here serving in ServeHTTP will panic.

satellite/marketing/marketingweb/server.go
```diff
@@ -83,13 +76,27 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+}
+
+// ServeHTTP handles index request
+func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.URL.Path != "/" {
+		s.serveNotFound(w, req)
+		return
+	}
+
+	if s.templates == nil {
```
egonelbre: There are other places where the templates are parsed.

satellite/marketing/marketingweb/server.go
VinozzZ: changed it to return the error

satellite/marketing/marketingweb/server.go
VinozzZ: added all templates parsing into a method

satellite/marketing/marketingweb/server.go
```diff
@@ -34,6 +40,7 @@ type Server struct {
 	server   http.Server
 
 	templateDir string
+	templates   customTemplate
```
egonelbre: you can an anonymous struct here, it will avoid the extra type

PR #2182, opened 2019-06-12
Removing bucket_id and adding project_id and bucket_name as primary key

ethanadams:
What: 
Removing bucket_id and adding project_id and bucket_name as primary key
Why:
Allows us to join to the bucket BW rollup and storage tallies table more efficiently.

Thanks for submitting a PR!


satellite/satellitedb/migrate.go
```diff
@@ -785,6 +785,19 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					);`,
 				},
 			},
+			{
+				Description: "Alter value attribution table. Remove bucket_id. Add project_id and bucket_name as primary key",
+				Version:     30,
+				Action: migrate.SQL{
+					`ALTER TABLE value_attributions DROP CONSTRAINT value_attributions_pkey`,
+					`ALTER TABLE value_attributions ADD project_id bytea;`,
+					`UPDATE value_attributions SET project_id=SUBSTRING(bucket_id FROM 1 FOR 16);`,
```
mniewrzal: I don't think you need logic to move data from one schema to another. We might deploy new table to satellite but I'm pretty sure nobody is using this yet. So just removing column + adding 2 new one should be ok.

mniewrzal: ah sorry i might be wrong, without this migration can fail because we have one entry in testdata

satellite/satellitedb/dbx/satellitedb.dbx.postgres.sql
```diff
@@ -188,10 +188,11 @@ CREATE TABLE users (
 	PRIMARY KEY ( id )
 );
 CREATE TABLE value_attributions (
-	bucket_id bytea NOT NULL,
+	project_id bytea NOT NULL,
+	bucket_name bytea NOT NULL,
 	partner_id bytea NOT NULL,
 	last_updated timestamp NOT NULL,
-	PRIMARY KEY ( bucket_id )
+	PRIMARY KEY ( project_id, bucket_name )
```
aligeti: It is possible for one partner to have multiple buckets. In that case can you make multiple entries for the same partner id ?

PR #2183, opened 2019-06-12
Always encrypt inline segments without padding

JessicaGreben:
**Background:**
When uploading a file, we can encrypt data in two different ways:
1. call `encryption.Encrypt()` directly which encrypts the file as one byte slice, or
2. create a `TransformedReader` which encrypts the file into `encBlockSize` chunks

In the code, there was an inconsistency when we stored files inline where sometimes we encrypted by method 1 listed above while other times we encrypted by method 2.  This happened because we were checking inline vs remote in one layer, while checking if file size was less than `encBlockSize` in a different layer.  

For example, in `pkg/streams/store.go upload()` we checked if the file size was less than `encBlockSize`,  so we used method 1 from above. Then one layer down in `pkg/segments/store.go` we checked if the file should be inlined or not.

It it was small enough to be inline and also less than `encBlockSize`, then we directly called `encryption.Encrypt()`. **But** when the file was small enough to inline, but **larger than** `encBlockSize`, we used  a `TransformedReader` instead.

It was also the case that inline files that were smaller than one block size were not padded to match the block size, while those that were larger than one block size were padded to match a block-size multiple. This padding is not necessary, and increases the cost of storing the pointer.

**What/Why:** 
This PR makes it so files that are small enough to be inlined are always just encrypted directly, not chunked, and not padded to match block sizes. In order to accomplish this, we moved the inline/remote logic up one layer into `pkg/streams/` from `pkg/segment`.


pkg/storage/streams/store.go
```diff
@@ -109,157 +111,203 @@ func (s *streamStore) Put(ctx context.Context, path storj.Path, pathCipher storj
...
-			}
-			transformedReader = bytes.NewReader(cipherData)
+			return Meta{}, currentSegmentIndex, err
 		}
 
-		putMeta, err = s.segments.Put(ctx, transformedReader, expiration, func() (storj.Path, []byte, error) {
-			encPath, err := EncryptAfterBucket(ctx, path, pathCipher, s.rootKey)
+		// Deal with inline first
+		if !isRemote {
+			data := peekReader.Buf()
```
mobyvb: I think it might make sense to use `data, err := ioutil.ReadAll(peekReader)` rather than adding a `Buf()` function to `peekReader`

Alternatively, if you still want to go with the `Buf()` route, maybe there should be a check within that function to verify that `IsLargerThan` was called? Otherwise there is the risk of returning an empty buffer if someone forgot to call `IsLargerThan` first

JessicaGreben: yeah good point, i think we can just `readAll`.  I'm trying to remember if there is a reason we didn't do that in the first place. Either way, yes to adding either `readAll` or adding the check to `IsLargerThan`

JessicaGreben:
@mobyvb thank for testing the code! We discussed backwards compatibility, and I don't think we need to worry about that.  I believe we still have some data wipes coming.

JessicaGreben:
@kaloyan-raev I feel like this PR already makes those changes.
1. https://github.com/storj/storj/pull/2183/files#diff-72b734f4a2e408a544fef071d2565fc8R186
2. https://github.com/storj/storj/pull/2183/files#diff-60b2cf8b2412ad57a05f99445aec4663R278
What do you think needs to change?

kaloyan-raev:
Please, revert all refactoring. The PR was very hard to follow and time-consuming. Let's just focus on the fix.

This whole code will be deleted after we go back to the metainfo refactoring. If you still want to do the refactoring, open a new PR for it.

pkg/storage/streams/store.go
```diff
@@ -59,15 +59,16 @@ type Store interface {
...
+	segmentSize     int64
+	rootKey         *storj.Key
+	encBlockSize    int
+	cipher          storj.Cipher
+	inlineThreshold int
 }
 
 // NewStreamStore stuff
-func NewStreamStore(segments segments.Store, segmentSize int64, rootKey *storj.Key, encBlockSize int, cipher storj.Cipher) (Store, error) {
+func NewStreamStore(segments segments.Store, segmentSize int64, rootKey *storj.Key, encBlockSize int, cipher storj.Cipher, inlineThreshold int) (Store, error) {
```
egonelbre: Note: the number of arguments is getting large here.

kaloyan-raev: True, the list is growing. But we are going to replace this code with a brand new one around the Metainfo Refactoring, right?

egonelbre: Yeah, eventually.

PR #2186, opened 2019-06-12
satellite/marketing: Create New Offer

fuskovic:
What: Renders offer table data to UI by type and Implements handler for creating new offers. 

Why: So referral admins can create new offers in the admin-portal.

Thanks for submitting a PR!


satellite/marketing/marketingweb/server.go
```diff
@@ -8,18 +8,31 @@ import (
...
 	"github.com/gorilla/mux"
 )
 
-// Error is satellite marketing error type
-var Error = errs.Class("satellite marketing error")
+// Time converter is used by decoder to format expiration date field from form input
+var (
+	Error       = errs.Class("satellite marketing error")
+	decoder     = schema.NewDecoder()
+	timeConverter = func(value string) reflect.Value {
```
VinozzZ: What's the benefit of having an anonymous function here?

satellite/marketing/marketingweb/server.go
```diff
@@ -28,16 +41,33 @@ type Config struct {
...
-
 	listener net.Listener
 	server   http.Server
-
+	service  *marketing.Service
 	templateDir string
 }
 
-// commonPages returns templates that are required for everything.
+// Struct used to render each offer table
```
VinozzZ: https://github.com/storj/storj/wiki/Style#comments I think we are following this convention for commenting

satellite/marketing/marketingweb/server.go
```diff
@@ -28,16 +41,33 @@ type Config struct {
...
+type offerSet struct {
+	RefOffers,FreeCredits []marketing.Offer
+}
+
+// Takes db result of get offers, organizes them by type and returns them
+// Used by the index handler to build render the tables
+func organizeOffers(offers []marketing.Offer) offerSet{
+	os := new(offerSet)
```
VinozzZ: It's a little weird that we are initializing a pointer and then returning its value. Maybe we can do `var os offerSet` and then directly return its value at the end of this function

satellite/marketing/marketingweb/server.go
```diff
@@ -70,31 +102,74 @@ func NewServer(logger *zap.Logger, config Config, listener net.Listener) *Server
 	return s
 }
 
-// ServeHTTP handles index request
-func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+// Serves index page and renders offer and credits tables
```
VinozzZ: https://github.com/storj/storj/wiki/Style#comments

satellite/marketing/marketingweb/server.go
```diff
@@ -70,31 +102,74 @@ func NewServer(logger *zap.Logger, config Config, listener net.Listener) *Server
...
+// Helper function used by createOffer to turn post form input into a new offer
+func (s *Server) formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffer, e error){
+	err := req.ParseForm()
 	if err != nil {
-		s.log.Error("failed to execute template", zap.Error(err))
+		return o, err
+	}
+	defer req.Body.Close()
+
+	decoder.RegisterConverter(time.Time{}, timeConverter)
```
egonelbre: This might race.

satellite/marketing/marketingweb/server.go
egonelbre: Why is this not written as a regular func?

satellite/marketing/marketingweb/server.go
```diff
@@ -8,18 +8,31 @@ import (
...
+	"storj.io/storj/satellite/marketing"
 	"github.com/gorilla/mux"
 )
 
-// Error is satellite marketing error type
-var Error = errs.Class("satellite marketing error")
+// Time converter is used by decoder to format expiration date field from form input
+var (
+	Error       = errs.Class("satellite marketing error")
+	decoder     = schema.NewDecoder()
```
egonelbre: Is it thread-safe?

satellite/marketing/marketingweb/server.go
```diff
@@ -70,31 +102,74 @@ func NewServer(logger *zap.Logger, config Config, listener net.Listener) *Server
...
 	if err != nil {
-		s.serveInternalError(w, req)
+		s.serveInternalError(w,req,err)
 		return
 	}
+}
 
-	err = home.ExecuteTemplate(w, "base", nil)
+// Helper function used by createOffer to turn post form input into a new offer
+func (s *Server) formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffer, e error){
```
VinozzZ: Does this function need to have the receiver here?

satellite/marketing/marketingweb/server.go
```diff
@@ -48,20 +78,22 @@ func (s *Server) commonPages() []string {
...
 
 	logger.Sugar().Debugf("Starting Marketing Admin UI on %s...", s.listener.Addr().String())
-
 	fs := http.StripPrefix("/static/", http.FileServer(http.Dir(s.config.StaticDir)))
 	mux := mux.NewRouter()
 	if s.config.StaticDir != "" {
+		mux.HandleFunc("/", s.getOffers)
 		mux.PathPrefix("/static/").Handler(fs)
-		mux.Handle("/", s)
+		mux.HandleFunc("/createFreeCredit", s.createOffer)
```
egonelbre: Use more conventional endpoint naming. All lower-case only, with maybe "-"... the internet tends to mangle casing.

satellite/marketing/marketingweb/server.go
fuskovic: Is it because decoder is global ?

satellite/marketing/marketingweb/server.go
fuskovic: No good reason. Updating.

satellite/marketing/marketingweb/server.go
fuskovic: Same thoughts as https://github.com/storj/storj/pull/2186#discussion_r293191793.

satellite/marketing/marketingweb/server.go
fuskovic: https://github.com/storj/storj/pull/2186#discussion_r293191828

satellite/marketing/marketingweb/server.go
fuskovic: You're right.

satellite/marketing/marketingweb/server.go
fuskovic: Since I'm not using anything from `s` no it doesn't.

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: I localized decoder into the only block it was needed. I believe that solves the concern. Correct me if i'm wrong.

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: https://github.com/storj/storj/pull/2186#discussion_r293196093

satellite/marketing/marketingweb/server.go
```diff
@@ -79,16 +99,23 @@ func NewServer(logger *zap.Logger, config Config, listener net.Listener) (*Serve
 	return s, nil
 }
 
-// ServeHTTP handles index request
-func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+// Serves index page and renders offer and credits tables
```
egonelbre: The comment still doesn't follow the convention. i.e.
```
// <funcname> ...
```

web/marketing/css/style.css
```diff
@@ -60,12 +60,13 @@ hr{
...
 
-.data-row{
+.data-row,
+.data-row .col a{
     font-weight: normal;
     font-size: 14px;
     line-height: 49px;
     letter-spacing: -0.100741px;
-    color: #656565;
+    color: #656565 !important;
```
egonelbre: Why `!important`? This seems like ordering in css is messed up.

satellite/marketing/marketingweb/server.go
```diff
@@ -28,12 +30,10 @@ type Config struct {
 // Server represents marketing offersweb server
 type Server struct {
 	log *zap.Logger
-
 	config Config
-
 	listener net.Listener
 	server   http.Server
-
+	service  *marketing.Service
```
egonelbre: Leave whitespace between logical chunks, it helps with readability.

satellite/marketing/marketingweb/server.go
```diff
@@ -42,7 +42,25 @@ type Server struct {
...
 }
 
-// commonPages returns templates that are required for everything.
+// Struct used to render each offer table
+type offerSet struct {
+	RefOffers,FreeCredits []marketing.Offer
+}
+
+// Organizes offers from database by type.
+func organizeOffers(offers []marketing.Offer) offerSet{
```
egonelbre: You can use filter functions in templates.

satellite/marketing/marketingweb/server.go
```diff
@@ -53,20 +71,22 @@ func (s *Server) commonPages() []string {
...
 	logger.Sugar().Debugf("Starting Marketing Admin UI on %s...", s.listener.Addr().String())
-
 	fs := http.StripPrefix("/static/", http.FileServer(http.Dir(s.config.StaticDir)))
 	mux := mux.NewRouter()
 	if s.config.StaticDir != "" {
+		mux.HandleFunc("/", s.getOffers)
 		mux.PathPrefix("/static/").Handler(fs)
-		mux.Handle("/", s)
+		mux.HandleFunc("/create-free-credit", s.createOffer)
+		mux.HandleFunc("/create-ref-offer", s.createOffer)
```
egonelbre: `create-reference-offer`?

satellite/marketing/marketingweb/server.go
```diff
@@ -53,20 +71,22 @@ func (s *Server) commonPages() []string {
...
 
 	logger.Sugar().Debugf("Starting Marketing Admin UI on %s...", s.listener.Addr().String())
-
 	fs := http.StripPrefix("/static/", http.FileServer(http.Dir(s.config.StaticDir)))
 	mux := mux.NewRouter()
 	if s.config.StaticDir != "" {
+		mux.HandleFunc("/", s.getOffers)
 		mux.PathPrefix("/static/").Handler(fs)
-		mux.Handle("/", s)
+		mux.HandleFunc("/create-free-credit", s.createOffer)
```
egonelbre: Are `/create-free-credit`  and `/create-ref-offer` supposed to link to the same handler?

satellite/marketing/marketingweb/server.go
egonelbre: The globalness wasn't the problem, it was that it was calling `Register` from a func that is called concurrently. If it were done in an init func it might have been fine.

fuskovic: I see. I will move the `decoder` var back in the global vars block since I need it in an init func for registering the `timeConverter` and also `formToStruct ()` for the actual decoding.

I will use the decoder to register the timeConverter in the `NewServer` whee it should be safe no?

satellite/marketing/marketingweb/server.go
fuskovic: This is for an older commit. This comment isn't present in latest.

web/marketing/css/style.css
fuskovic: removed

satellite/marketing/marketingweb/server.go
fuskovic: done

satellite/marketing/marketingweb/server.go
fuskovic: Yes, there is logic in the handler that assigns the offer type for newly created offer based on the endpoint the request came from.

satellite/marketing/marketingweb/server.go
fuskovic: I did this originally and it made the template look messy imo. Handling the filtering logic on the server side helped me write cleaner more easily-readable template logic.

satellite/marketing/marketingweb/server.go
fuskovic: If this is good I pushed it already.

egonelbre: No, in an init func... so:

```
func init() {
   decoder.Register(...
```

satellite/marketing/marketingweb/server.go
```diff
@@ -128,25 +160,78 @@ func (s *Server) parseTemplates() (err error) {
...
+	o, err := formToStruct(w,req)
+	if err != nil{
+		s.log.Error("err from createFreeCredit Handler", zap.Error(err))
+		s.serveInternalError(w,req,err)
+		return
+	}
+
+	o.Status = marketing.Active
+
+	if req.URL.Path == "/create-referral-offer" {
```
VinozzZ: Can we pass a url param to indicate the type in the request instead of creating two endpoints?

satellite/marketing/marketingweb/server.go
```diff
@@ -128,25 +160,78 @@ func (s *Server) parseTemplates() (err error) {
...
 	w.WriteHeader(http.StatusInternalServerError)
 
-	err := s.templates.internalError.ExecuteTemplate(w, "base", nil)
-	if err != nil {
+	if err := s.templates.internalError.ExecuteTemplate(w, "base", e); err != nil {
 		s.log.Error("failed to execute template", zap.Error(err))
 	}
 }
 
-// Run starts the server that host admin web app and api endpoint
```
VinozzZ: Maybe this is an accident?

satellite/marketing/marketingweb/server.go
```diff
@@ -162,7 +247,6 @@ func (s *Server) Run(ctx context.Context) error {
 	return group.Wait()
 }
 
-// Close closes server and underlying listener
```
VinozzZ: Accident delete?

fuskovic: pushed

satellite/marketing/marketingweb/server.go
fuskovic: pushed

satellite/marketing/marketingweb/server.go
fuskovic: pushed

satellite/marketing/marketingweb/server.go
fuskovic: Gotcha. I understood this better after reading [this](https://tutorialedge.net/golang/the-go-init-function/.) I pushed the change

satellite/marketing/marketingweb/server.go
fuskovic: I pushed an update so only one endpoint is used for create which takes an `{offer_type}` param which is then used in a conditional in the handler to set the offer type https://github.com/storj/storj/pull/2186#discussion_r293925878.

satellite/marketing/marketingweb/server.go
```diff
@@ -42,7 +45,29 @@ type Server struct {
...
+
+func init(){
+	decoder.RegisterConverter(time.Time{}, timeConverter)
+}
+
+// Organizes offers from database by type.
+func organizeOffers(offers []marketing.Offer) offerSet{
+	var os offerSet
+	for _,offer := range offers {
+		if offer.Type == marketing.FreeCredit {
```
VinozzZ: Do you think that a `switch` statement here would have a better readability?

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,86 @@
...
+		form.Add("AwardCreditInCents","50")
+		form.Add("AwardCreditDurationDays","50")
+	}
+	return form
+}
+
+func buildResources(endpoint string, isReferralOffer bool) (url.URL, url.Values){
+	URL, err := url.ParseRequestURI("http://127.0.0.1:10003")
+	if err != nil{
+		log.Fatalf("URL parsing Err : %v\n", err)
```
VinozzZ: I don't think we want to exit the process if there is a error here

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,86 @@
...
+	form := buildForm(isReferralOffer)
+	URL.RawQuery = form.Encode()
+	return *URL, form
+}
+
+func callServer(t *testing.T,endpoint,params string, form url.Values) string{
+	c := http.Client{}
+	req, err := http.NewRequest("POST", endpoint, strings.NewReader(params))
+	if err != nil {
+		t.Fatalf("Err building request : %v\n", err)
```
VinozzZ: We probably wan to use `require.NoError` here

satellite/marketing/marketingweb/server.go
VinozzZ: This is still not following the comment pattern `// <function name> .....`

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +163,74 @@ func (s *Server) parseTemplates() (err error) {
 	return nil
 }
 
+// Time converter is used by decoder in formToStruct to format expiration date field from form input.
```
VinozzZ: Same as above

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +163,74 @@ func (s *Server) parseTemplates() (err error) {
...
+	}
+	defer req.Body.Close()
+
+	if err := decoder.Decode(&o, req.PostForm);err != nil {
+		return o, err
+	}
+	return o,nil
+}
+
+// Handler for POST requests to create, insert, and start a new offer or credit.
```
VinozzZ: This is still not following the comment pattern // <function name> .....

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,86 @@
+package marketingweb
```
VinozzZ: You can see an example of how to use `testplanet` for testing in `containment_test.go`

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if isReferralOffer {
+		form.Add("AwardCreditInCents","50")
+		form.Add("AwardCreditDurationDays","50")
+	}
+	return form
+}
+
+func buildResources(endpoint string, isReferralOffer bool) (url.URL, url.Values){
+	URL, err := url.ParseRequestURI("http://127.0.0.1:10003")
```
egonelbre: Don't use fixed ports in tests.

egonelbre: Don't use fixed ports in tests.

satellite/marketing/marketingweb/server_test.go
fuskovic: How come? Should I just log the error and return it ?

VinozzZ: Because we also want other tests continue to run. If you check other tests files, you will see we are using the `"github.com/stretchr/testify/require"` assertion package to do testing

satellite/marketing/marketingweb/server.go
fuskovic: This looks like an older commit. I just pushed up updates to the formatting of the comments again. If you pull down latest you should see they all follow the correct format now.

satellite/marketing/marketingweb/server.go
fuskovic: Yes. Pushed.

satellite/marketing/marketingweb/server.go
fuskovic: Pushed. Pull down latest to see all updated comment formatting.

satellite/marketing/marketingweb/server.go
fuskovic: pushed

satellite/marketing/marketingweb/server_test.go
fuskovic: pushed

satellite/marketing/marketingweb/server_test.go
fuskovic: pushed appending of `_test` to pkg name.

satellite/marketing/marketingweb/server_test.go
fuskovic: pushed

satellite/marketing/marketingweb/server.go
VinozzZ: I still don't see the new changes

fuskovic: Did you pull down latest for `fh/create-new-offer` ?

VinozzZ: `// Serves index page and renders offer and credits tables` I'm seeing this on github

satellite/marketing/marketingweb/server_test.go
fuskovic: pushed

fuskovic: pushed

satellite/marketing/marketingweb/server_test.go
fuskovic: Added test planet and followed proper http server unit test practices pushed

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+			URL, form, err := buildResources(s.Config.Address, endpoint, isReferralOffer)
+			require.NoError(t, err, "failed to build request resources")
+
+			urlStr := URL.String()
+
+			req, err := http.NewRequest("POST", urlStr, strings.NewReader(URL.RawQuery))
+			require.NoError(t, err, "failed to create new POST request")
+
+			req.PostForm = form
+			req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
egonelbre: Make an actual request over the network.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+			rr := httptest.NewRecorder()
+
+			go func(recorder *httptest.ResponseRecorder, request *http.Request) {
```
egonelbre: All goroutines that have been started must be waited on.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func buildResources(address, endpoint string, isReferralOffer bool) (url.URL, url.Values, error) {
+	URL, err := url.ParseRequestURI("http://" + address)
+	if err != nil {
+		fmt.Printf("err from buildResources : %v\n", err)
+		return *URL, url.Values{}, err
+	}
+	URL.Path = endpoint
+	form := buildForm(isReferralOffer)
+	URL.RawQuery = form.Encode()
```
egonelbre: Why is this request going over put params?

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+				isReferralOffer = true
+			case marketing.FreeCredit:
+				isReferralOffer = false
+				endpoint += "/free-credit"
+			}
+
+			URL, form, err := buildResources(s.Config.Address, endpoint, isReferralOffer)
+			require.NoError(t, err, "failed to build request resources")
+
+			urlStr := URL.String()
```
egonelbre: See https://golang.org/pkg/net/http/, specifically:

```
resp, err := http.PostForm("http://example.com/form",
	url.Values{"key": {"Value"}, "id": {"123"}})
```

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+		SatelliteCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		validOffers := []marketing.NewOffer{
+			{Type: marketing.Referral},
+			{Type: marketing.FreeCredit},
+		}
+
+		s := planet.Satellites[0].Marketing.Endpoint
+
+		for _, offer := range validOffers {
```
egonelbre: You can make this parallel by running it with:

```
t.Run(offer.String(), func(t *testing.T) {
    t.Parallel()
})
```

satellite/marketing/marketingweb/server.go
```diff
@@ -9,31 +9,37 @@ import (
...
 type Server struct {
-	log *zap.Logger
-
-	config Config
-
-	listener net.Listener
-	server   http.Server
-
+	log         *zap.Logger
+	Config      Config
```
egonelbre: why is this public?

satellite/marketing/marketingweb/server.go
```diff
@@ -79,16 +114,23 @@ func NewServer(logger *zap.Logger, config Config, listener net.Listener) (*Serve
...
-func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+// getOffers renders the tables for free credits and referral offers to the UI
+func (s *Server) getOffers(w http.ResponseWriter, req *http.Request) {
 	if req.URL.Path != "/" {
 		s.serveNotFound(w, req)
 		return
 	}
 
-	err := s.templates.home.ExecuteTemplate(w, "base", nil)
+	offers, err := s.service.ListAllOffers(context.Background())
```
egonelbre: `context.Background()` shouldn't be used: https://github.com/storj/storj/wiki/Style#contexts

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
 	return nil
 }
 
+// timeConverter formats form time input as time.Time.
+func timeConverter(value string) reflect.Value {
```
egonelbre: convertStringToTime

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
+	v, err := time.Parse("2006-01-02", value)
+	if err != nil {
+		return reflect.Value{}
```
egonelbre: Add a note that this is the invalid value for decoder.

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
+	}
+	return o, nil
+}
+
+// CreateOffer handles requests to create new offers.
+func (s *Server) CreateOffer(w http.ResponseWriter, req *http.Request) {
+	o, err := formToStruct(w, req)
+	if err != nil {
+		s.log.Error("failed to convert form to struct", zap.Error(err))
+		s.serveInternalError(w, req, err)
```
egonelbre: Use the correct http error code. This is not a internal server error.

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
 		s.log.Error("failed to execute template", zap.Error(err))
+		s.serveInternalError(w, req, err)
+		return
 	}
 }
 
-func (s *Server) serveInternalError(w http.ResponseWriter, req *http.Request) {
+// serveInternalError handles 500 errors and renders err to the internalErr template.
+func (s *Server) serveInternalError(w http.ResponseWriter, req *http.Request, e error) {
+
```
egonelbre: remove empty line

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
 	w.WriteHeader(http.StatusInternalServerError)
 
-	err := s.templates.internalError.ExecuteTemplate(w, "base", nil)
-	if err != nil {
+	if err := s.templates.internalError.ExecuteTemplate(w, "base", e); err != nil {
```
egonelbre: the line is quite long so keep it on a separate line

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
 	if err != nil {
 		s.log.Error("failed to execute template", zap.Error(err))
+		s.serveInternalError(w, req, err)
+		return
 	}
 }
 
-func (s *Server) serveInternalError(w http.ResponseWriter, req *http.Request) {
+// serveInternalError handles 500 errors and renders err to the internalErr template.
+func (s *Server) serveInternalError(w http.ResponseWriter, req *http.Request, e error) {
```
egonelbre: use a better name than `e` here

satellite/marketing/marketingweb/server.go
egonelbre: Do we really want to expose our internal error? If we open the endpoint at some point to the public with authentication then we would need to remove this. We still have the log with the appropriate error.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+	form := buildForm(isReferralOffer)
+	URL.RawQuery = form.Encode()
+	return *URL, form, nil
+}
+
+func TestCreateOffer(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		validOffers := []marketing.NewOffer{
```
egonelbre: Instead of making this of `marketing.NewOffer` here create a table of:

```
type CreateRequest struct {
    Path string
    Values url.Values
    ExpectedStatus http.StatusCode
}

requests := []CreateRequest{
    {
        Path: "/create/referral-offer,
        Values: url.Values{
            "Name", "May Credit",
            "Description", "desc",
            "ExpiresAt", "2019-06-27",
            "InviteeCreditInCents", "50",
            "InviteeCreditDurationDays", "50",
            "RedeemableCap", "150",
        },
        ExpectedStatus: http.StatusOK,
        RedirectedTo: "/",
    }
}
```

satellite/marketing/marketingweb/server.go
```diff
@@ -42,7 +48,35 @@ type Server struct {
...
+
+		case marketing.Referral:
+			os.RefOffers = append(os.RefOffers, offer)
+		}
+
+	}
+	return os
+}
+
+// CommonPages returns templates that are required for everything.
```
VinozzZ: Nitpick: it should be `commonPages` since it's a private method

satellite/marketing/marketingweb/server.go
```diff
@@ -53,24 +87,25 @@ func (s *Server) commonPages() []string {
...
 	}
 
 	logger.Sugar().Debugf("Starting Marketing Admin UI on %s...", s.listener.Addr().String())
-
-	fs := http.StripPrefix("/static/", http.FileServer(http.Dir(s.config.StaticDir)))
+	fs := http.StripPrefix("/static/", http.FileServer(http.Dir(s.Config.StaticDir)))
 	mux := mux.NewRouter()
-	if s.config.StaticDir != "" {
+	if s.Config.StaticDir != "" {
+		mux.HandleFunc("/", s.getOffers)
```
VinozzZ: why the `getOffers()` method is private but `CreateOffer()` is public?

satellite/marketing/marketingweb/server.go
VinozzZ: `req.Context()` ?

satellite/marketing/marketingweb/server.go
```diff
@@ -128,20 +170,73 @@ func (s *Server) parseTemplates() (err error) {
...
+	o.Status = marketing.Active
+	reqType := mux.Vars(req)["offer_type"]
+
+	if reqType == "referral-offer" {
+		o.Type = marketing.Referral
+	} else {
+		o.Type = marketing.FreeCredit
+	}
+
+	if _, err := s.service.InsertNewOffer(context.Background(), &o); err != nil {
```
VinozzZ: `req.Context()`?

satellite/marketing/marketingweb/server.go
fuskovic: The referral-admin portal isn't ever going to be open to the public. Only a couple admins on the marketing team who will use it to manage offers behind a VPN. The reason we are exposing the error is if someone from Marketing gets an error, they can file a Jira ticket and include the error message in the screenshot which will help us fix it for them. This is an internal service.

satellite/marketing/marketingweb/server.go
fuskovic: I don't mind changing it but I would like to understand why I would use one over the other ?

VinozzZ: https://github.com/storj/storj/wiki/Style#contexts

satellite/marketing/marketingweb/server_test.go
fuskovic: `http.StatusCode` doesn't exist but `http.ResponseWriter` has a `StatusCode` field so I'm thinking you meant that or `http.StatusSeeOther` which you also can't use because it's not a type so wouldn't the right thing to do be to type-set `ExpectedStatus` as an int? 

The Path key in your `CreateRequest` struct literal has a string value that's not terminated by a closing double-quote.

In the `Values` key which is of type `map[string][]string` your formatting there looks wrong to me because your map literal has no key.

@egonelbre

fuskovic: .........But I understand what you wan't me to do.

satellite/marketing/marketingweb/server.go
fuskovic: If the server fails to convert the form to a struct because a referral admin input invalid form values I wan't that to be displayed on the page `serveInternalError` serves so when they file a ticket and attach that message I know to tell them that they are not inputting correct form values. If they get the 404 page instead that doesn't help me help them figure out what went wrong.

satellite/marketing/marketingweb/server_test.go
VinozzZ: https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

fuskovic: @VinozzZ can you elaborate please ?

fuskovic: Ok I believe I know how to correct it.

satellite/marketing/marketingweb/server.go
fuskovic: When you told me to not use a fixed address earlier in my unit test I made this public so I could use the address string from the server config in my unit test.

satellite/marketing/marketingweb/server_test.go
fuskovic: line 87 is just adding the form to the request. I'm actually executing the http request on `s.CreateOffer`

satellite/marketing/marketingweb/server.go
VinozzZ: Maybe try adding this in the /marketingweb/server.go` ?
```golang
func (s *Server) Addr() string { return s.config.Address }
```

satellite/marketing/marketingweb/server_test.go
VinozzZ: I think he meant that we should make an actual `POST` request to the endpoint `/create` instead of manually creating a `Request` data and passing into `CreateOffer`. Ideally, if we make a post request to `/create`, we won't need to call `CreateOffer()` in this test ourselves

satellite/marketing/marketingweb/server.go
fuskovic: I can do that but it's not clear to my why I would create a new method for the purpose of returning one of it's receiver's fields when I can just export the field to begin with ? 

Why is it bad to export Config?

egonelbre: There's no reason to export it. The `s.config.Address` doesn't contain the actual address, but the template for the address... e.g. it could be `":0"`

egonelbre: If do an actual request against that address it should fail.

satellite/marketing/marketingweb/server.go
egonelbre: The correct error code here would be `400` and you can still show what wasn't parsed correctly.

satellite/marketing/marketingweb/server_test.go
egonelbre: Yeah, my typos.

satellite/marketing/marketingweb/server_test.go
fuskovic: that makes sense

satellite/marketing/marketingweb/server.go
VinozzZ: Hey @fuskovic, maybe try this?
```golang
s := planet.Satellites[0].Marketing.Listener.Addr()
resp, err := http.PostForm("http://"+s.String()+"/create", nil)
```

satellite/marketing/marketingweb/server_test.go
fuskovic: No worries. I think you were just showing me the general idea. I can't expect you to take the time to spell it out for me every time. Thank you

web/marketing/pages/err.html
```diff
@@ -0,0 +1,3 @@
+{{define "main"}}
```
VinozzZ: We need copyright in this file

web/marketing/pages/free-offers-modal.html
```diff
@@ -1,3 +1,4 @@
+
```
VinozzZ: extra empty line

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,119 @@ func (s *Server) parseTemplates() (err error) {
...
+	if err != nil {
+		s.log.Error("failed to convert form to struct", zap.Error(err))
+		s.serveBadRequest(w, req, err)
+		return
+	}
+
+	o.Status = marketing.Active
+	reqType := mux.Vars(req)["offer_type"]
+
+	if reqType == "referral-offer" {
```
VinozzZ: What if someone puts in `/create/foo`? I would use a switch statement here since we don't want to create `FreeCredit` offer by default.

web/marketing/pages/free-offers.html
```diff
@@ -13,49 +13,38 @@
...
-            <div class="col">06/01/19</div>
-            <div class="col"><strong>Live</strong></div>
-            <div class="col col-heading edit-offer">EDIT</div>
-        </div>
-        <hr>
+        {{range .FreeCredits}}
+            {{if eq .Status 2}}
+                <div class="row data-row">
+                    <div class="col ml-3">{{.Name}}</div>
+                    <div class="col">${{.InviteeCreditInCents}}</div>
```
VinozzZ: We should either let user know this is a `cent` field or we do the conversion ourselves in the handler.

web/marketing/pages/free-offers.html
```diff
@@ -13,49 +13,38 @@
...
-            <div class="col ml-3">May Referral</div>
-            <div class="col">$50</div> 
-            <div class="col">$50</div>
-            <div class="col">48</div>
-            <div class="col">200</div>
-            <div class="col">05/01/19</div>
-            <div class="col">06/01/19</div>
-            <div class="col">Off</div>
+            <div class="col ml-3">{{.Name}}</div>
+            <div class="col">${{.InviteeCreditInCents}}</div>
```
VinozzZ: same as above

satellite/marketing/marketingweb/server.go
```diff
@@ -216,11 +216,12 @@ func (s *Server) CreateOffer(w http.ResponseWriter, req *http.Request) {
 	o.Status = marketing.Active
 	reqType := mux.Vars(req)["offer_type"]
 
-	if reqType == "referral-offer" {
-		o.Type = marketing.Referral
-	} else {
-		o.Type = marketing.FreeCredit
-	}
+	switch reqType {
```
VinozzZ: We need a default case here

satellite/marketing/marketingweb/server.go
VinozzZ: I think @egonelbre meant to add a comment here? I don't think we need to log that

satellite/marketing/marketingweb/server.go
```diff
@@ -6,43 +6,79 @@ package marketingweb
...
+	RefOffers, FreeCredits []marketing.Offer
+}
+
+// init safely registers convertStringToTime for the decoder.
+func init() {
+	decoder.RegisterConverter(time.Time{}, convertStringToTime)
+}
+
+// organizeOffers organizes offers by type.
+func organizeOffers(offers []marketing.Offer) offerSet {
```
coyle: do you have a unit test for this ?

satellite/marketing/marketingweb/server.go
```diff
@@ -6,43 +6,79 @@ package marketingweb
...
 		pageNotFound  *template.Template
 		internalError *template.Template
+		badRequest    *template.Template
 	}
 }
 
-// commonPages returns templates that are required for everything.
+// offerSet provides a separation of marketing offers by type.
+type offerSet struct {
+	RefOffers, FreeCredits []marketing.Offer
```
coyle: nit -- I think the long form is more readable

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,124 @@ func (s *Server) parseTemplates() (err error) {
...
-	err := s.templates.internalError.ExecuteTemplate(w, "base", nil)
-	if err != nil {
+	if err := s.templates.internalError.ExecuteTemplate(w, "base", errMsg); err != nil {
+		s.log.Error("failed to execute template", zap.Error(err))
+	}
+}
+
+// serveBadRequest handles 400 errors and renders err to the bad-request template.
+func (s *Server) serveBadRequest(w http.ResponseWriter, req *http.Request, errMsg error) {
+
```
coyle: same nit as @egonelbre made above

satellite/marketing/marketingweb/server.go
fuskovic: This is a helper function for `GetOffers` which is unit-tested.

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,122 @@ func (s *Server) parseTemplates() (err error) {
...
+	}
+
 	return nil
 }
 
+// convertStringToTime formats form time input as time.Time.
+func convertStringToTime(value string) reflect.Value {
+	v, err := time.Parse("2006-01-02", value)
+	if err != nil {
+		log.Println("invalid decoder value")
```
egonelbre: No need to log, just a comment suffices.

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,122 @@ func (s *Server) parseTemplates() (err error) {
...
+func (s *Server) CreateOffer(w http.ResponseWriter, req *http.Request) {
+	o, err := formToStruct(w, req)
+	if err != nil {
+		s.log.Error("failed to convert form to struct", zap.Error(err))
+		s.serveBadRequest(w, req, err)
+		return
+	}
+
+	o.Status = marketing.Active
+	reqType := mux.Vars(req)["offer_type"]
```
egonelbre: Why is it called `reqType` and not `offerType`?

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,122 @@ func (s *Server) parseTemplates() (err error) {
...
+
+	if err := decoder.Decode(&o, req.PostForm); err != nil {
+		return o, err
+	}
+	return o, nil
+}
+
+// CreateOffer handles requests to create new offers.
+func (s *Server) CreateOffer(w http.ResponseWriter, req *http.Request) {
+	o, err := formToStruct(w, req)
```
egonelbre: `o` -> `offer`

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,122 @@ func (s *Server) parseTemplates() (err error) {
...
+	v, err := time.Parse("2006-01-02", value)
+	if err != nil {
+		log.Println("invalid decoder value")
+		return reflect.Value{}
+	}
+	return reflect.ValueOf(v)
+}
+
+// formToStruct decodes POST form data into a new offer.
+func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffer, e error) {
```
egonelbre: Since this is only parsing offers: `parseOfferForm`

satellite/marketing/marketingweb/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+				}
+
+				_, err = http.Get(baseURL)
+				if err != nil {
+					return err
+				}
+
+				return nil
+			})
+			err := group.Wait()
```
egonelbre: group wait should be the last thing this does.

web/marketing/pages/referral-offers.html
```diff
@@ -11,50 +11,39 @@
...
-        <div class="col col-heading edit-offer">EDIT</div>
+        <p class="offer-type">default&nbsp;offer</p>
     </div>
-    <hr>
+    {{range .ReferralOffers}}
+        {{if eq .Status 2}}
+            <div class="row data-row">
+                <div class="col ml-3">{{.Name}}</div>
+                <div class="col">${{.InviteeCreditInCents}}</div> 
+                <div class="col">${{.AwardCreditInCents}}</div>
```
egonelbre: Why is the credit in cents? However $ indicates it's in non-cents? Or is it something else than dollars?

satellite/marketing/marketingweb/server.go
```diff
@@ -102,46 +146,122 @@ func (s *Server) parseTemplates() (err error) {
...
+}
+
+// serveNotFound handles 404 errors and defaults to 500 if template parsing fails.
 func (s *Server) serveNotFound(w http.ResponseWriter, req *http.Request) {
 	w.WriteHeader(http.StatusNotFound)
 
 	err := s.templates.pageNotFound.ExecuteTemplate(w, "base", nil)
 	if err != nil {
 		s.log.Error("failed to execute template", zap.Error(err))
+		s.serveInternalError(w, req, err)
```
egonelbre: This will write the header twice and cause an error in http server.

satellite/rewards/rewards.go
```diff
@@ -5,9 +5,28 @@ package rewards
...
 	"context"
+	"fmt"
+	"math"
+	"strconv"
 	"time"
 )
 
+// ToCents converts USD credit amounts to cents.
+func ToCents(dollar int) int {
+	cents := float64(dollar) * 100
```
VinozzZ: Why do we need to convert `dollar` into `floast64` here and then convert it back to `int` in return?

satellite/rewards/rewards.go
```diff
@@ -5,9 +5,28 @@ package rewards
...
+    return int(math.Round(cents))
+}
+
+// ToDollars converts credit amounts in cents to USD.
+func ToDollars(cents int) int {
+	formattedAmount := fmt.Sprintf("%d.%d",(cents / 100),(cents % 100))
+	dollars, err := strconv.ParseFloat(formattedAmount,64)
```
VinozzZ: Same as above

satellite/rewards/rewards.go
fuskovic: `Math.Round ` only takes a `float64`

VinozzZ: If we are using integer. Why do we need to round it?

satellite/rewards/rewards.go
fuskovic: Modulo and integer division has to be used per our group slack thread but results of those two have to be formatted and you can't format `%d.%d` as an int without truncating. Also this helps prevent needing to create a custom `Cent` type which I also realized would result in lots of breaking changes in the db logic.

satellite/rewards/rewards.go
fuskovic: You're right, we don't. I removed it and pushed.

satellite/rewards/rewards.go
```diff
@@ -5,9 +5,28 @@ package rewards
...
+	if err != nil {
+		return 0
+	}
+	return int(dollars)
```
VinozzZ: This will truncate the formatted amount

satellite/rewards/rewards.go
VinozzZ: I think this is a display value. We just need to format it return the `formattedAmount`

satellite/rewards/rewards.go
fuskovic: Changed this as well.

VinozzZ: Maybe you pushed to the other branch? I'm not seeing the changes here

satellite/marketingweb/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+				}
+
+				_, err = http.Get(baseURL)
+				if err != nil {
+					return err
+				}
+
+				return nil
+			})
+			err := group.Wait()
```
egonelbre: This group.Wait needs to be outside the for loop.

satellite/rewards/rewards.go
fuskovic: Should be good now

satellite/rewards/rewards.go
```diff
@@ -5,26 +5,17 @@ package rewards
...
-func ToCents(dollar int) int {
-	cents := float64(dollar) * 100
-    return int(math.Round(cents))
+func ToCents(dollars int) int {
+    return dollars * 100
 }
 
 // ToDollars converts credit amounts in cents to USD.
 func ToDollars(cents int) int {
-	formattedAmount := fmt.Sprintf("%d.%d",(cents / 100),(cents % 100))
```
VinozzZ: We still need this format here. Otherwise, `5550` cents would return `55` instead of `55.50`

fuskovic: hmm, so then I would be returning either a `float64` or a `string`, neither of which I can re-assign back to  `...CreditInCents` without type-casting and truncating the output.

fuskovic: I could do it in the template but then I would definitely have to create the custom type no ?

satellite/rewards/rewards.go
fuskovic: I'm going to write a new interface `Converter` to help invoke ToDollars as a method from the template without the need to change the type of `...CreditInCents`

satellite/rewards/rewards.go
fuskovic: https://github.com/storj/storj/pull/2186#discussion_r298308612

satellite/marketingweb/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+					"ExpiresAt":                 {"2119-06-27"},
+					"InviteeCreditInCents":      {"50"},
+					"InviteeCreditDurationDays": {"50"},
+					"AwardCreditInCents":        {"50"},
+					"AwardCreditDurationDays":   {"50"},
+					"RedeemableCap":             {"150"},
+				},
+			}, {
+				Path: "/create/free-credit-offer",
+				Values: url.Values{
```
VinozzZ: We need to add a test case that the user will have input like `InviteeCreditInCents: "50.5"`

web/marketing/pages/referral-offers-modal.html
```diff
@@ -2,51 +2,51 @@
...
+                            <label for="ExpiresAt">Credit Exp Date</label>
+                            <input type="date" class="form-control" id="ExpiresAt" name="ExpiresAt" placeholder="06/01/19" required>
                         </div>
                     </div>
                     <div class="form-row">
                         <div class="form-group col-md-2">
-                            <label for="give-credit">Give Credit</label>
-                            <input type="number" class="form-control" id="give-credit" placeholder="$50" required>                        
+                            <label for="InviteeCreditInCents">Give Credit(¢)</label>
+                            <input type="number" class="form-control" name="InviteeCreditInCents" id="InviteeCreditInCents" placeholder="$50" required>
```
VinozzZ: This `input` field needs `step="0.1"` or whatever accuracy Andi would want

satellite/marketingweb/server.go
```diff
@@ -102,32 +148,102 @@ func (s *Server) parseTemplates() (err error) {
...
+		s.serveBadRequest(w, req, err)
+		return
+	}
+	inviteeCredits := offer.InviteeCreditInCents
+	awardCredits := offer.AwardCreditInCents
+
+	offer.Status = rewards.Active
+	offerType := mux.Vars(req)["offer_type"]
+
+	offer.AwardCreditInCents = rewards.ToCents(awardCredits)
```
VinozzZ: This converting needs to happen before `decoder`, otherwise it won't be able to parse it into our struct

satellite/rewards/rewards.go
VinozzZ:
```golang
	s.templates.home, err = template.New("landingPage").Funcs(template.FuncMap{
		"ToDollars": func(cents int) string {
			return fmt.Sprintf("%d.%d", cents/100, cents%100)
		},
	}).ParseFiles(homeFiles...)
```
You can pass in a function like this as well and in the template you can do `{{ ToDollars .InviteeCreditInCents }}`

egonelbre: this won't format properly with 505.

VinozzZ: If we only allow `integer` value for dollar amount. We should never have a cent value to be `505`?

web/marketing/pages/free-offers-modal.html
```diff
@@ -12,33 +12,33 @@ <h5 class="modal-title" id="free-offers-modal-lbl">Create Free Credit</h5>
...
+                            <label for="ExpiresAt">Credit Exp Date</label>
+                            <input type="date" class="form-control" name="ExpiresAt" id="ExpiresAt" placeholder="06/01/19" required>
                         </div>
                     </div>
                     <div class="form-row">
                         <div class="form-group col-md-4">
-                            <label for="give-credit">Give Credit</label>
-                            <input type="number" class="form-control" id="give-credit" placeholder="$50" min="1" required>                        
+                            <label for="AwardCreditInCents">Award Credit(¢)</label>
+                            <input type="number" class="form-control" name="AwardCreditInCents" id="AwardCreditInCents" placeholder="$50" min="1" required>
```
VinozzZ: Let's add a `min=0` for all credits input field

PR #2190, opened 2019-06-13
pkg/process/metrics: add an instance prefix

jtolds:
the distinction between which satellite is sending which data should go in the instance field, not the suffix or application fields. (un)fortunately, the instance id is deliberately not configurable because we don't want it to be easy to accidentally have multiple applications collide with the same instance id.

so we're currently stuffing the human readable instance in the suffix. :(

perhaps a reasonable tradeoff would be an optional instance prefix that allows operators to put their domain name in the instance

Change-Id: I6fcc8498be908c5740439cc00f77474ad151febd

What: 

Why:

Thanks for submitting a PR!


egonelbre:
Is there a reason we are not sending a signed node id on connection to metrics backend? This would uniquely identify the satellite or anything without adding an extra field.

jtolds:
signing doesn't really give us much security without a handshake. these packets have no timestamps or anything in them, and we don't, or shouldn't, rely on telemetry for actual application logic anywhere. the node id is currently sent just fine but it's bad UX. you have to know it in grafana/redash. this PR allows us to add a little "mars-", etc. tag to the front. see https://github.com/storj/kube-infra/pull/38

egonelbre:
Yeah, the handshake was implied. If we are sending it already cannot we have a lookup table on grafana/redash?

jtolds:
A lookup table is probably a good idea. I'll check with data science tomorrow morning to see how feasible that is before we merge this.

pgerbes1:
I like the idea of a lookup table. Pretty sure Redash let's you use Google Sheets as a data source so can build it manually if need be.

jtolds:
okay so philip pointed out a good point - if we have multiple application servers serving the same satellite node id, we can't distinguish between app servers without being able to add something to the instance id

egonelbre:
Shouldn't the differentiating things be part of the suffix not the prefix?

jtolds:
you mean make an instance suffix as well as an application suffix?

the major benefit of having an instance prefix is so the ops team can make instances show up ordered by human name alphabetically. it makes the ux when there's no lookup table much better

PR #2192, opened 2019-06-13
config: update release values to match prod

jtolds:
Looking at the config in k8s, these flags are all manually overridden to these values. Let's just make the code match so we know and the custom configs can go away.

Thanks for submitting a PR!


jtolds:
the most unusual config change here is that we don't do distinct IPs in production?

pkg/datarepair/repairer/repairer.go
```diff
@@ -36,7 +36,7 @@ var (
 type Config struct {
 	MaxRepair    int           `help:"maximum segments that can be repaired concurrently" releaseDefault:"5" devDefault:"1"`
 	Interval     time.Duration `help:"how frequently repairer should try and repair more data" releaseDefault:"1h" devDefault:"0h5m0s"`
-	Timeout      time.Duration `help:"time limit for uploading repaired pieces to new storage nodes" default:"10m0s"`
+	Timeout      time.Duration `help:"time limit for uploading repaired pieces to new storage nodes" devDefault:"10m0s" releaseDefault:"2h"`
```
egonelbre: So `2h` to upload single piece?

satellite/console/consoleweb/server.go
```diff
@@ -47,7 +47,7 @@ var (
 
 // Config contains configuration for console web server
 type Config struct {
-	Address         string `help:"server address of the graphql api gateway and frontend app" default:"127.0.0.1:8081"`
+	Address         string `help:"server address of the graphql api gateway and frontend app" devDefault:"127.0.0.1:8081" releaseDefault:"0.0.0.0:10100"`
```
egonelbre: Do we really want to expose these things by default rather than let devops manually specify them? It seems like something that we shouldn't do, but I'm not too strongly against it either.

pkg/accounting/live/live.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Config contains configurable values for the live accounting service.
 type Config struct {
-	StorageBackend string `help:"what to use for storing real-time accounting data"`
+	StorageBackend string `help:"what to use for storing real-time accounting data" default:"plainmemory"`
```
egonelbre: Any reason it's not called `inmemory` instead of `plainmemory`. Is there something that makes it more plainer than inner? :D

pkg/datarepair/repairer/repairer.go
littleskunk: The problem is that repair uploads are running into some strange error message and we want to make sure it has nothing to do with the timeout.

pkg/overlay/config.go
```diff
@@ -41,7 +41,7 @@ type NodeSelectionConfig struct {
 	NewNodePercentage float64       `help:"the percentage of new nodes allowed per request" default:"0.05"` // TODO: fix, this is not percentage, it's ratio
 	MinimumVersion    string        `help:"the minimum node software version for node selection queries" default:""`
 	OnlineWindow      time.Duration `help:"the amount of time without seeing a node before its considered offline" default:"1h"`
-	DistinctIP        bool          `help:"require distinct IPs when choosing nodes for upload" releaseDefault:"true" devDefault:"false"`
+	DistinctIP        bool          `help:"require distinct IPs when choosing nodes for upload" releaseDefault:"false" devDefault:"false"`
```
littleskunk: Please keep true here. I have tested the IP filter. It doesn't work the way we need it but it is still an inprovement. I missed to activate it in production.

satellite/console/consoleweb/server.go
stefanbenten: I second that, this should stay on localhost

pkg/accounting/live/live.go
stefanbenten: @egonelbre Thats a question for @thepaul

pkg/overlay/config.go
stefanbenten: Agreeing here as well!

satellite/console/consoleweb/server.go
jtolds: the devops config still specifies this config value directly, but i thought if devops wanted it this way maybe other satellite operators would benefit from the default. our default server address is already :7777

pkg/overlay/config.go
jtolds: okay, so, this should stay true, and we should activate in prod. that's kind of what i thought

satellite/console/consoleweb/server.go
jtolds: maybe this shouldn't specify 0.0.0.0 though if that messes up ipv6

pkg/accounting/live/live.go
thepaul: One potential future alternative would be something like memcache, which would still be in-memory, but would also be distributed and structured more intelligently.

satellite/console/consoleweb/server.go
egonelbre: The default server is password protected so there's less concern for it.

jtolds: the default grpc server? no it's not

PR #2194, opened 2019-06-13
Repair should not corrupt files

kaloyan-raev:
What: 

Fixes https://storjlabs.atlassian.net/browse/V3-1924 by forcing the repair to always download at least `k+1` pieces.

Why:

Error detection during erasure decoding is only possible if we have `k+1` pieces. If only `k` pieces are available the data can still be decoded, but we cannot say if it was corrupted or not.

Thanks for submitting a PR!


littleskunk:
We need to update the repair checker to call a file with k pieces irreparabel. k+1 is required now. I can push that commit.

Is it possible to remove the bad piece? Right now we keep the bad piece allive. Over time we will collect more and more bad pieces until the repairer is unable to find any healthy pieces.

kaloyan-raev:
> Is  it possible to remove the bad piece? Right now we keep the bad piece  allive. Over time we will collect more and more bad pieces until the  repairer is unable to find any healthy pieces.

@littleskunk It should be possible, but I think it would require an improvement in the infectious library to report the bad pieces with the Decode function. I suggest merging this PR to fix the corruption issue and opening another ticket for removing bad pieces from the pointer. Perhaps, we can also report a failed audit for the storage node too.

ifraixedes:
I'm wondering if it's possible to select `k+1` pieces and 2 of them are bad, could that happen? if it happens, wouldn't we get the same problem without this change?

littleskunk:
> I'm wondering if it's possible to select `k+1` pieces and 2 of them are bad, could that happen? if it happens, wouldn't we get the same problem without this change?

Yes that is exactly what will happen. For now lets merge this PR and I will open an new jira ticket. I will cherry pick and release it. If we are lucky it solves our current problem.

kaloyan-raev:
In this case, the repair will fail. It won't corrupt the file. If we select `k+1` pieces, the only way corruption can happen is if all of them are corrupted in such a way that they can be decoded successfully. This requires very strict coordination on all of these k+1 storage nodes.

PR #2198, opened 2019-06-13
Add timestamp and piece size to piece hash

mniewrzal:
What: Add timestamp and piece size to `PieceHash` to validate that several things:
* upload wasn't done long time ago (more than 2h) and uplink isn't postponing commiting segment on purpose
* use piece size to verify that uplink is not cheating on upload size
* additional beyond scope: compare `PieceHash` pieceID with `OrderLimit` pieceID to be sure that we received correct hash for given `OrderLimit`

Why: We want to avoid different scenarios of cheating.
https://storjlabs.atlassian.net/browse/V3-1652

This PR should be merged in coordinance to satellite and storage node deploy. First it should be deployed on storage node and later satellite can be updated.


pkg/pb/orders.proto
```diff
@@ -60,6 +60,10 @@ message PieceHash {
     bytes hash = 2;
     // signature either satellite or storage node
     bytes signature = 3;
+    // size of uploaded piece
+    int64 piece_size = 4;
+    // timestamp when upload occur
+    google.protobuf.Timestamp timestamp = 5;
```
kaloyan-raev: Add `[(gogoproto.stdtime) = true]` like here: https://github.com/storj/storj/blob/15e74c8c3d68821eac13f4d2cccf89ba4604b237/pkg/macaroon/types.proto#L27

This will set the type to `time.Time` in the generated Go code, which is more convenient and we won't need to convert from Timestamp to Time.

satellite/metainfo/metainfo.go
```diff
@@ -28,6 +28,9 @@ import (
 	"storj.io/storj/storage"
 )
 
+// TODO make it configurable ?
+const pieceHashExpiration = time.Hour * 2
```
kaloyan-raev: nit: `2 * time.Hour` reads better.

satellite/metainfo/metainfo.go
```diff
@@ -467,9 +472,39 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
+			err = endpoint.validatePieceHash(ctx, piece, limits)
+			if err != nil {
+				// TODO should this be logged also to uplink somehow ?
+				endpoint.log.Sugar().Warn(err)
+				continue
+			}
+
+			// TODO maybe minimal PieceSize should be bigger
+			if piece.Hash.PieceSize <= 0 || (maxPieceSize > 0 && maxPieceSize != piece.Hash.PieceSize) {
+				allSizesValid = false
```
kaloyan-raev: break here?

satellite/metainfo/metainfo.go
```diff
@@ -448,12 +451,14 @@ func createBucketID(projectID uuid.UUID, bucket []byte) []byte {
...
 
-func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Pointer) (err error) {
+func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Pointer, limits []*pb.OrderLimit2) (err error) {
 	defer mon.Task()(&ctx)(&err)
 
 	if pointer.Type == pb.Pointer_REMOTE {
 		var remotePieces []*pb.RemotePiece
 		remote := pointer.Remote
+		allSizesValid := true
+		maxPieceSize := int64(0)
```
kaloyan-raev: `maxPieceSize` is misleading var name. It looks like here you record the size of the first piece and then check if all other pieces have the same size.

satellite/metainfo/metainfo.go
mniewrzal: might be a good idea

satellite/metainfo/metainfo.go
mniewrzal: I cannot use first piece size because some pieces could be nil e.g first one. I will think to make more straightforward.

mniewrzal: renamed to `lastPieceSize` and removed condition to update it only with larger piece size

satellite/metainfo/metainfo.go
```diff
@@ -467,9 +472,37 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
+			}
+			lastPieceSize = piece.Hash.PieceSize
+
 			remotePieces = append(remotePieces, piece)
 		}
 
+		if allSizesValid {
+			redundancy, err := eestream.NewRedundancyStrategyFromProto(pointer.GetRemote().GetRedundancy())
+			if err != nil {
+				return Error.New("unable to verify piece size, invalid redundancy scheme values")
```
kaloyan-raev: Here we lose the original error and it may be harder to debug what was exactly wrong with the redundancy. Perhaps, it's better to just return the error.

pkg/pb/orders.proto
```diff
@@ -60,6 +60,10 @@ message PieceHash {
     bytes hash = 2;
     // signature either satellite or storage node
     bytes signature = 3;
+    // size of uploaded piece
+    int64 piece_size = 4;
+    // timestamp when upload occur
+    google.protobuf.Timestamp timestamp = 5 [(gogoproto.stdtime) = true];
```
kaloyan-raev: Oh, I've just learnt this can be actually annotated with `[(gogoproto.stdtime) = true, (gogoproto.nullable) = false]`, which will make `timestamp` of type `time.Time` instead of `*time.Time`. This will resolve the issue in some places to get a pointer to `time.Now()`.

mniewrzal: even better, nice!

satellite/metainfo/metainfo.go
ifraixedes: I'm wondering if it would be useful to have a kind of function that allows to wrap the error and provides an error.

satellite/metainfo/validation.go
```diff
@@ -244,13 +244,12 @@ func (endpoint *Endpoint) validatePointer(ctx context.Context, pointer *pb.Point
 
 	// TODO does it all?
```
ifraixedes: I know that it has not been added in this branch, but do you know if the question is still applying?

mniewrzal: We didn't have one story to verify and fill gaps in `pointer` struct validation. Lots of these rules were added with tasks only related to validation. I will remove this comment and make jira ticket to review metainfo validation.

PR #2202, opened 2019-06-13
NodeSelectionConfig inits and test cleanup

wthorp:
What: 
Added new reputation fields to NodeSelectionConfig initialization
Cleaned up NodeSelectionConfig related tests to be more readable

Why:
General cleanup after adding NodeSelectionConfig repuation fields


internal/testplanet/planet.go
```diff
@@ -450,13 +450,25 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
-					OnlineWindow:      time.Hour,
-					DistinctIP:        false,
+					UptimeRatio:                  0,
+					UptimeCount:                  0,
+					AuditSuccessRatio:            0,
+					AuditCount:                   0,
+					NewNodePercentage:            0,
+					OnlineWindow:                 time.Hour,
+					DistinctIP:                   false,
+					ReputationAuditRepairWeight:  1,
```
egonelbre: Let's add some whitespace to group things more clearly.

PR #2206, opened 2019-06-13
NodeStats adding reputation fields

wthorp:
What: 
Adding reputation related fields to pb.NodeStats
Removed duplicate overlay.NodeStats

Why:
We'll be removing the ratio fields later, we need to replace them with something.
We learned the 'gogoproto.stdtime' trick to make protocol buffer Timestamps map directly to time.Time fields... thus the two structs were literally the same.


pkg/pb/node.proto
```diff
@@ -49,8 +49,14 @@ message NodeStats {
...
-    google.protobuf.Timestamp last_contact_success = 9;
-    google.protobuf.Timestamp last_contact_failure = 10;
+    google.protobuf.Timestamp last_contact_success = 9 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    google.protobuf.Timestamp last_contact_failure = 10 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    //new reputation and disqualification stuff
+    double audit_reputation_alpha = 11;
+    double uptime_reputation_alpha = 12;
+    double audit_reputation_beta = 13;
+    double uptime_reputation_beta = 14;
+	bool disqualified = 15;
```
navillasa: maybe it's github but does the indentation need to be fixed for this line?

wthorp: fixed

mobyvb:
I'm not quite sure why you removed `overlay.NodeStats`. If I recall correctly, the original reason we had a separate struct (it was originally in a protobuf like here), is because those values are only passed around on one component (satellite) and don't need to be sent over any grpc requests. Since we don't need to send stats over grpc requests, we bloat the codebase by using protobuf generated structs rather than manually defining the struct in the overlay package

pkg/pb/node.proto
```diff
@@ -49,8 +49,14 @@ message NodeStats {
...
-    google.protobuf.Timestamp last_contact_success = 9;
-    google.protobuf.Timestamp last_contact_failure = 10;
+    google.protobuf.Timestamp last_contact_success = 9 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    google.protobuf.Timestamp last_contact_failure = 10 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    //new reputation and disqualification stuff
+    double audit_reputation_alpha = 11;
+    double uptime_reputation_alpha = 12;
+    double audit_reputation_beta = 13;
+    double uptime_reputation_beta = 14;
+    bool disqualified = 15;
```
kaloyan-raev: Why did you add the `disqualified` field here? It is already part of `NodeDossier`.

wthorp:
Based on Moby's feedback and review at standup, this PR will likely be closed in favor of https://github.com/storj/storj/pull/2226

wthorp:
went with logic from https://github.com/storj/storj/pull/2226 instead

PR #2209, opened 2019-06-14
Uplink C bindings part 2

bryanchriswhite:
What: 

Why:

Thanks for submitting a PR!


internal/testcontext/compile.go
```diff
@@ -34,28 +35,28 @@ func (ctx *Context) Compile(pkg string) string {
 }
 
 // CompileShared compiles pkg as c-shared.
-func (ctx *Context) CompileShared(name string, pkg string) Include {
-	ctx.test.Helper()
+func (ctx *Context) CompileShared(t *testing.T, name string, pkg string) Include {
```
egonelbre: I think the fix here should be to make the running in separate folders work. Let's add a TODO note about it for now.

lib/uplinkc/apikey.go
```diff
@@ -12,21 +12,22 @@ import (
...
 }
 
 //export serialize_api_key
 // serialize_api_key serializes the API Key to a string
-func serialize_api_key(apikeyHandle C.APIKey) *C.char {
+func serialize_api_key(apikeyHandle C.APIKeyRef_t, cerr **C.char) *C.char {
 	apikey, ok := universe.Get(apikeyHandle._handle).(libuplink.APIKey)
 	if !ok {
+		*cerr = C.CString("invalid apikey")
 		return C.CString("")
```
egonelbre: Here we should return nil. If I get an error I shouldn't have to free the result.

lib/uplinkc/project.go
```diff
@@ -7,48 +7,48 @@ package main
...
 import (
-	"storj.io/storj/lib/uplink"
+	libuplink "storj.io/storj/lib/uplink"
 )
 
 // Project is a scoped uplink.Project
 type Project struct {
 	scope
-	*uplink.Project
+	*libuplink.Project
```
egonelbre: Let's keep the `uplink` thing, we'll revert `libuplink` changes in mobile.

lib/uplinkc/apikey.go
```diff
@@ -12,21 +12,22 @@ import (
 
 //export parse_api_key
 // parse_api_key parses an API Key
-func parse_api_key(val *C.char, cerr **C.char) (apikeyHandle C.APIKey) {
+func parse_api_key(val *C.char, cerr **C.char) (apikeyHandle C.APIKeyRef_t) {
 	apikey, err := libuplink.ParseAPIKey(C.GoString(val))
 	if err != nil {
 		*cerr = C.CString(err.Error())
 		return apikeyHandle
```
egonelbre: Let's make it explicit that it's returning an empty value. So `C.APIKey{}`.

lib/uplinkc/uplink_definitions.h
```diff
@@ -4,14 +4,65 @@
 #include <stdbool.h>
 #include <stdlib.h>
 
-typedef struct APIKey   { long _handle; } APIKey;
-typedef struct Uplink   { long _handle; } Uplink;
-typedef struct Project  { long _handle; } Project;
+typedef struct APIKey   { long _handle; } APIKeyRef_t;
+typedef struct Uplink   { long _handle; } UplinkRef_t;
+typedef struct Project  { long _handle; } ProjectRef_t;
+typedef struct Bucket   { long _handle; } BucketRef_t;
```
egonelbre: Note from posix, `_t` are reserved. http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

egonelbre: But also everybody seems to be ignoring that. :/

egonelbre: https://stackoverflow.com/a/3225396/192220

lib/uplinkc/uplink.go
```diff
@@ -19,24 +19,24 @@ type Uplink struct {
 // an error in cerr, when there is one.
 //
 // Caller must call close_uplink to close associated resources.
-func new_uplink(cfg C.UplinkConfig, cerr **C.char) C.Uplink {
+func new_uplink(cfg C.UplinkConfig_t, cerr **C.char) C.UplinkRef_t {
 	scope := rootScope("inmemory") // TODO: pass in as argument
 
 	libcfg := &uplink.Config{} // TODO: figure out a better name
-	libcfg.Volatile.TLS.SkipPeerCAWhitelist = cfg.Volatile.TLS.SkipPeerCAWhitelist == 1
+	libcfg.Volatile.TLS.SkipPeerCAWhitelist = cfg.Volatile.TLS.SkipPeerCAWhitelist == C.bool(true)
```
egonelbre: Is casting `true` to `C.bool` the correct behavior?

lib/uplinkc/project.go
bryanchriswhite: I think this is better

lib/uplinkc/bucket.go
```diff
@@ -0,0 +1,193 @@
...
+		config = &uplink.BucketConfig{
+			PathCipher: storj.CipherSuite(bucketConfig.path_cipher),
+			EncryptionParameters: storj.EncryptionParameters{
+				CipherSuite: storj.CipherSuite(bucketConfig.encryption_parameters.cipher_suite),
+				BlockSize:   int32(bucketConfig.encryption_parameters.block_size),
+			},
+		}
+		config.Volatile.RedundancyScheme = storj.RedundancyScheme{
+			Algorithm:      storj.RedundancyAlgorithm(bucketConfig.redundancy_scheme.algorithm),
+			ShareSize:      int32(bucketConfig.redundancy_scheme.share_size),
```
mniewrzal: I'm not sure how it works with cgo but will `bucketConfig.redundancy_scheme` be always allocated? or it can be a nil?

egonelbre: Here it cannot be nil.

egonelbre: So, I've not yet conclusively decided. We can do it two ways:
1. keep a pointer, this means that from C it will be more annoying to use - you need to alloc/free the config
2. no pointer, this means in Go we need to detect `0` case and handle that instead of missing RS.

bryanchriswhite: We didn't make it a pointer because the libuplink code used a value. It might also be worth noting that this is in the volatile field which is subject to change and we weren't supposed to really mess with anyways.

lib/uplinkc/uplink.go
bryanchriswhite: I believe so, we changed the type of `SkipPeerCAWhitelist` to be a boolean.

internal/testcontext/compile.go
bryanchriswhite: Can you elaborate on the folders bit, I don't quite follow. Afaik the issue here was that the test in the testcontext and the test that's calling `CompileC`.

egonelbre: I mean, for compilation we should be able to use different testcontexts and folders for compilation. However, currently it's easier if the shared lib is in the same folder as the executable.

bryanchriswhite: Gotcha, I'll add that Todo.

PR #2210, opened 2019-06-14
Improve stability of TestDownloadSharesDownloadTimeout

kaloyan-raev:
What: 

Changes the way how we simulate timeout during the download of an erasure share during an audit. In this case, we want dialing the node to be always successful and the download of the data to always timeout.

So far we used a slow transport client on the satellite. This slows down the network communication for all storage nodes, and for both dialing the node and the actual download of the data. It turned out that it is hard to adjust the test in a way that it runs quick and at the same time guarantee that always the download will timeout, but not dialing the node.

The alternative approach used here is to replace the blob store on a specific storage node with a slow blob node -- one that will delay its operations with configurable duration. This way dialing the node always run as fast as possible, while downloading the actual data will be slow.

Why:

Because TestDownloadSharesDownloadTimeout occasionally fails because the download the audit share fails with an error from dial node timeout, where it is expected an error from download timeout.

Thanks for submitting a PR!


storagenode/pieces/store.go
```diff
@@ -81,6 +81,12 @@ func NewStore(log *zap.Logger, blobs storage.Blobs) *Store {
 	}
 }
 
+// WithBlobs replaces the blob store with the provided blobs.
+func (store *Store) WithBlobs(blobs storage.Blobs) *Store {
```
kaloyan-raev: @egonelbre do you have a better idea how to do this?

internal/testblobs/slow.go
```diff
@@ -0,0 +1,51 @@
...
+func NewSlowBlobs(blobs storage.Blobs, delay time.Duration) *SlowBlobs {
+	return &SlowBlobs{
+		blobs: blobs,
+		delay: delay,
+	}
+}
+
+// Create creates a new blob that can be written
+// optionally takes a size argument for performance improvements, -1 is unknown size
+func (b *SlowBlobs) Create(ctx context.Context, ref storage.BlobRef, size int64) (storage.BlobWriter, error) {
```
ifraixedes: According to our current code conventions, the `b` isn't a good name for a receiver.

pkg/audit/verifier_test.go
```diff
@@ -299,18 +298,12 @@ func TestDownloadSharesDownloadTimeout(t *testing.T) {
 		require.NoError(t, err)
 
 		upl := planet.Uplinks[0]
-		testData := make([]byte, 32*memory.KiB)
+		testData := make([]byte, 8*memory.KiB)
 		_, err = rand.Read(testData)
 		require.NoError(t, err)
 
 		// Upload with larger erasure share size to simulate longer download over slow transport client
```
ifraixedes: is this comment still apply?

internal/testblobs/slow.go
kaloyan-raev: True. Renamed it.

pkg/audit/verifier_test.go
kaloyan-raev: Good catch. It's outdated, removed it.

storagenode/pieces/store.go
egonelbre: Reconfigure allows you to use a custom storage node db. There you can wrap the blobs as needed.

``` go
type SlowDB struct { storagenode.DB }
func (db *SlowDB) Pieces() storage.Blobs { return &SlowBlobs{db.Latency, db.DB.Pieces()} }
```

If you need it to depend on the call... you can pass that information via context... e.g.

```
// do regular stuff
ctx := context.WithValue(ctx, slowblobs.Enabled, true)
// now it's going to be slow
```

kaloyan-raev: Thanks! Indeed this would look better.

I am trying to reconfigure like this:

```go
		Reconfigure: testplanet.Reconfigure{
			NewStorageNodeDB: func(index int) (storagenode.DB, error) {
				return &SlowDB{
					DB: <db-to-wrap>,
					Latency: 110 * time.Millisecond,
				}, nil
			},
		},
	}
```

Where am I supposed to get `<db-to-wrap>` from? This should be the storagenode.DB created by the testplanet, but I don't find a way to acces it.

egonelbre: `NewStorageNodeDB` is the func that creates the database. So you would need to call `storagenodedb.New` manually. It might actually make sense to give the "default db constructor" as an argument to `NewStorageNodeDB`.

kaloyan-raev: @egonelbre I am trying to follow the idea with the `context.WithValue`, but it seems it won't work in this test case.

The issue is that I want the blob store to be fast until I want to download the shares. At this moment I derive a child context with value `slowblobs.Enabled` and pass with to `verifier.DownloadShares`.

Unfortunately, `verifier.DownloadShares` is exectued on the satellite. The respective storage nodes is not affected at all by this derived context.

egonelbre: Ah, right... yeah... then in that case I would keep a references to the created slow blob stores and call "enable", or something, before doing the slow test.

kaloyan-raev: Cool. This worked: https://github.com/storj/storj/pull/2210/commits/d976aff07fd682d15de8e4bd2c8158843e19a103

internal/testblobs/slow.go
```diff
@@ -0,0 +1,104 @@
...
+	"storj.io/storj/storagenode"
+)
+
+// SlowDB implements slow storage node DB.
+type SlowDB struct {
+	storagenode.DB
+	blobs *SlowBlobs
+}
+
+// NewSlowDB creates a new slow storage node DB wrapping apping the provided db.
```
navillasa: "wrapping apping" lol

kaloyan-raev: tell me how I've done it :)

navillasa: oh sorry unclear-- the comment probably should just say "DB wrapping the provided db" instead of "wrapping apping the provided db"

kaloyan-raev: I totally got it the first time. I wanted to say that I had no idea how I made such a mistake.

kaloyan-raev:
For some reason, sometimes the latency does not apply to SlowBlobs.

Logs from latest test failure on Jenkins:
```
logger.go:130: 2019-06-26T09:59:20.614Z INFO storage0.piecestore download started {"Piece ID": "LFDUHAL65DV7J65GUUZGVM4VY6HTEUXILCRVJ22VVA26PS24WUVA", "SatelliteID": "1ZgnjCicAfiGjJfKh7rnGG5if1Ps7i39THiEjMe1SgZs9xHYGr", "Action": "GET_AUDIT"}
logger.go:130: 2019-06-26T09:59:20.623Z INFO storage0.piecestore downloaded {"Piece ID": "LFDUHAL65DV7J65GUUZGVM4VY6HTEUXILCRVJ22VVA26PS24WUVA", "SatelliteID": "1ZgnjCicAfiGjJfKh7rnGG5if1Ps7i39THiEjMe1SgZs9xHYGr", "Action": "GET_AUDIT"}
```

The share was downloaded for just 9 ms, while the latency was set to 200 ms.

I don't reproduce it locally.

@egonelbre do you have any idea what could be wrong?

egonelbre:
@kaloyan-raev so `time.Sleep` is specified as to sleeping for `amount +- system granularity`, but as far as I know the granularity should be below 100ms on Docker/Linux.

You can try debugging/logging whether it actually hits that place and has the right value for the sleep.

kaloyan-raev:
@egonelbre I added additional logs. It turns out that the `storageNodeDB.SetLatency(delay)` is not executed at all in the problematic runs. I have no idea how this would happen...

I have a log right before calling this line:
```
DEBUG satellite0 TestDownloadSharesDownloadTimeout SetLatency {"delay": "200ms", "Type": "*testblobs.SlowDB"}
```
And a log that is the first thing executed by `SlowDB.SetLatency`:
```
DEBUG slowdb SlowDB SetLatency {"delay": "200ms"}
```
During the problematic runs the first log is displayed, but the second is missing!

Any idea how this could happen at all?

egonelbre:
`testplanet.Run` runs two tests in parallel

egonelbre:
Maybe try something along the lines of:
```
		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 1,
		Reconfigure: testplanet.Reconfigure{
			NewStorageNodeDB: func(index int, db storagenode.DB, log *zap.Logger) (storagenode.DB, error) {
				return testblobs.NewSlowDB(log.Named("slowdb"), db), nil
			},
		},
		slowblobs := planet.StorageNodes[0].DB.(*slowblobs.DB)
		err := planet.Satellites[0].Audit.Service.Close()
		require.NoError(t, err)
```

kaloyan-raev:
@egonelbre Thanks! This seems to work.

PR #2212, opened 2019-06-14
[v3-1914] Storage node disqualification: Change type from bool to timestamp

ifraixedes:
What: Disqualification column must be of a timestamp type not a bool one

Why: In the last architecture meeting we realize that using a timestamp column for disqualification rather than a bool was ideal for being able to undisqualify nodes which could be disqualified by an unexpected cause such a bug, hard network disconnection, etc.

Thanks for submitting a PR!


satellite/satellitedb/migrate.go
```diff
@@ -815,6 +815,13 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					);`,
 				},
 			},
+			{
+				Description: "Change type of disqualified column of nodes table to timestamp",
+				Version:     32,
+				Action: migrate.SQL{
+					`ALTER TABLE nodes ALTER COLUMN disqualified TYPE timestamp with time zone`,
```
ifraixedes: I assume that currently the `disqualified` field was not used, otherwise it should be done in different way (using several queries and alter statements) to set the current `disqualified = true` to something like `disqualified = <<some timestamp in the past>>` rather than setting all of them to `NULL`

satellite/satellitedb/overlaycache.go
```diff
@@ -803,7 +810,7 @@ func convertDBNode(ctx context.Context, info *dbx.Node) (_ *overlay.NodeDossier,
 			Release:    info.Release,
 		},
 		Contained:    info.Contained,
-		Disqualified: info.Disqualified,
+		Disqualified: disqualified,
```
ifraixedes: Not sure if currently use `*time.Time` for business logic types or use a `time.Time`, which is what it has been used here and due to the DB uses a `*time.Time` it must be mapped accordingly.

On the other hand, there may be some helper function in our current code base for this conversions, but I don't know as I asked in other means and I didn't get an answer at that time and I moved one.

satellite/satellitedb/testdata/postgres.v32.sql
```diff
@@ -0,0 +1,314 @@
...
+INSERT INTO "project_invoice_stamps" ("project_id", "invoice_id", "start_date", "end_date", "created_at") VALUES (E'\\022\\217/\\014\\376!K\\023\\276\\031\\311}m\\236\\205\\300'::bytea, E'\\363\\311\\033w\\222\\303,'::bytea, '2019-06-01 08:28:24.267934+00', '2019-06-29 08:28:24.267934+00', '2019-06-01 08:28:24.267934+00');
+
+INSERT INTO "value_attributions" ("project_id", "bucket_name", "partner_id", "last_updated") VALUES (E'\\363\\311\\033w\\222\\303Ci\\265\\343U\\303\\312\\204",'::bytea, E''::bytea, E'\\363\\342\\363\\371>+F\\256\\263\\300\\273|\\342N\\347\\014'::bytea,'2019-02-14 08:07:31.028103+00');
+
+-- NEW DATA --
+
+INSERT INTO "user_credits" ("id", "user_id", "offer_id", "referred_by", "credits_earned_in_cents", "credits_used_in_cents", "expires_at", "created_at") VALUES (1, E'\\363\\311\\033w\\222\\303Ci\\265\\343U\\303\\312\\204",'::bytea, 1, E'\\363\\311\\033w\\222\\303Ci\\265\\343U\\303\\312\\204",'::bytea, 200, 0, '2019-10-01 08:28:24.267934+00', '2019-06-01 08:28:24.267934+00');
```
mobyvb: this shouldn't be new data

satellite/satellitedb/overlaycache.go
wthorp: I would make everything a *time.Time

ifraixedes: ok!

satellite/satellitedb/testdata/postgres.v32.sql
ifraixedes: I actually copied the `postgres.v31.sql` into this new one file and modified the the things related to disqualified, I haven't taken a look to the rest.

I'll moved to it should be

satellite/satellitedb/migrate.go
ifraixedes: Never mind, I updated the migration query and it currently sets a specific timestamp for the nodes which could have the `disqualified` column set to `FALSE`

ifraixedes:
I've tested in local with a PostgreSQL connection however I wanted to see if the CI passed, but it's currently unavailable so I ask for a review without waiting it.

satellite/satellitedb/migrate.go
```diff
@@ -815,6 +815,20 @@ func (db *DB) PostgresMigration() *migrate.Migration {
...
+			{
+				Description: "Change type of disqualified column of nodes table to timestamp",
+				Version:     32,
+				Action: migrate.SQL{
+					`ALTER TABLE nodes
+						ALTER COLUMN disqualified DROP DEFAULT,
+						ALTER COLUMN disqualified DROP NOT NULL,
+						ALTER COLUMN disqualified TYPE timestamp with time zone USING
+							CASE disqualified
+								WHEN true THEN TIMESTAMP WITH TIME ZONE '2019-06-15 00:00:00+00'
```
mobyvb: is it ok for us to hardcode this specific date in? Does it matter if it is not the correct date right now?

ifraixedes: I should have written my reason in the commit message or in a comment here.

At the beginning I had a `now()` call for setting the date, however I thought that a fix date would be better because if there is a bug somewhere (in this migration or even there are already nodes in the DB with `disqualified = FALSE`) we could easily identify the nodes which where disqualified by this migration.

The date could be any in the past, but I applied my random logic thinking in one that it isn't too old but it's before the release of this and used the 15th because it was exactly the half of the month.

mobyvb: That makes sense. Thank you for the explanation :)

PR #2215, opened 2019-06-14
nodes db - adds alpha and beta reputation fields

wthorp:
What: 
Added alpha and beta reputation fields to Nodes database
Updated SQL to retrieve alpha and beta reputation stats
Added migrate scripts

Why:
So we can track the new alpha and beta reputation stats


satellite/satellitedb/testdata/postgres.v32.sql
```diff
@@ -0,0 +1,317 @@
...
+	uptime_success_count bigint NOT NULL,
+	total_uptime_count bigint NOT NULL,
+	uptime_ratio double precision NOT NULL,
+	created_at timestamp with time zone NOT NULL,
+	updated_at timestamp with time zone NOT NULL,
+	last_contact_success timestamp with time zone NOT NULL,
+	last_contact_failure timestamp with time zone NOT NULL,
+	contained boolean NOT NULL,
+	disqualified boolean NOT NULL,
+	reputation_audit_alpha double precision NOT NULL,
```
ifraixedes: This and the following new columns miss the `DEFAULT`  value attribute set in the migration https://github.com/storj/storj/pull/2215/files#diff-af38612c15ad33aeba391857c3ec07c8R822

satellite/satellitedb/testdata/postgres.v32.sql
```diff
@@ -0,0 +1,317 @@
...
+INSERT INTO "nodes"("id", "address", "last_ip", "protocol", "type", "email", "wallet", "free_bandwidth", "free_disk", "major", "minor", "patch", "hash", "timestamp", "release","latency_90", "audit_success_count", "total_audit_count", "audit_success_ratio", "uptime_success_count", "total_uptime_count", "uptime_ratio", "created_at", "updated_at", "last_contact_success", "last_contact_failure", "contained", "disqualified", "reputation_audit_alpha", "reputation_audit_beta", "reputation_uptime_alpha", "reputation_uptime_beta") VALUES (E'\\363\\342\\363\\371>+F\\256\\263\\300\\273|\\342N\\347\\014', '127.0.0.1:55517', '', 0, 4, '', '', -1, -1, 0, 1, 0, '', 'epoch', false, 0, 0, 0, 1, 0, 0, 1, '2019-02-14 08:07:31.028103+00', '2019-02-14 08:07:31.108963+00', 'epoch', 'epoch', false, false, 1, 0, 1, 0);
```
ifraixedes: If you remove the `"reputation_audit_alpha", "reputation_audit_beta", "reputation_uptime_alpha", "reputation_uptime_beta"` columns from this `INSERT` and also the values `1, 0, 1, 0` this insert should allow to verify the column default values

docs/design/node-selection.md
```diff
@@ -48,12 +48,12 @@ This design may be refined in the future to prefer storage nodes based on speed,
 ```DBX
 model node (
 ...
-	field audit_reputation_α   float64   ( updatable )
-	field audit_reputation_β   float64   ( updatable )
-	field total_audit_count    int64     ( updatable )
-	field uptime_reputation_α  float64   ( updatable )
-	field uptime_reputation_β  float64   ( updatable )
-	field total_uptime_count   int64     ( updatable )
+	field reputation_audit_alpha  float64   ( updatable )
```
navillasa: this is very nitpicky and okay as is, but I prefer `audit_reputation_alpha` and `uptime_reputation_alpha` because I find it easier to distinguish the fields when they're all in a list.
e.g.
```
			dbx.Node_AuditReputationAlpha(1),
			dbx.Node_AuditReputationBeta(0),
			dbx.Node_UptimeReputationAlpha(1),
			dbx.Node_UptimeReputationBeta(0),
```
vs
```
			dbx.Node_ReputationAuditAlpha(1),
			dbx.Node_ReputationAuditBeta(0),
			dbx.Node_ReputationUptimeAlpha(1),
			dbx.Node_ReputationUptimeBeta(0),
```
idk hopefully other people also perceive a difference 😂

navillasa: actually, a better supporting argument for this suggestion is the fact that the NodeStats fields are `AuditReputationAlpha`, `UptimeReputationAlpha`, etc. in this PR (https://github.com/storj/storj/pull/2226), so I think it seems reasonable for them to be consistent.

satellite/satellitedb/testdata/postgres.v32.sql
wthorp: yeah, the auto-generated stuff is created by DBX and does not have the ability to set default vaules.  We have to manually add that to the migrate.go scripts.

satellite/satellitedb/testdata/postgres.v32.sql
wthorp: this is the v32 of the script.  it should require those values to be set, so as to verify them.  the v31 version of the script is essentially what you're asking for...the baseline against which we see if the default values get set.

docs/design/node-selection.md
wthorp: sure!  and fixed!

satellite/satellitedb/testdata/postgres.v32.sql
ifraixedes: I don't get your answer.
I thought that the test applies the migration and after it compares the data of the database with the data inserted by this one, hence if you don't insert the values here, the defaults should apply as they are applied by the migration.

satellite/satellitedb/testdata/postgres.v32.sql
ifraixedes: yes, but don't we have to add them here?

satellite/satellitedb/testdata/postgres.v32.sql
wthorp: v31 is run, the migrate scripts are applied, and the result is compared to v32.  thus v31 should not have defaults hard-coded, the migrate script should append default values and columns, and v32 should have everything hard-coded.

satellite/satellitedb/testdata/postgres.v32.sql
wthorp: no.  these files are never used.

satellite/satellitedb/testdata/postgres.v32.sql
ifraixedes: > thus v31 should not have defaults hard-coded

This one was obvious to me that it has not to be touched.

> the migrate scripts are applied

That implies that the existing values of the database will get the defaults

> and the result is compared to v32.

For that reason I was saying that if the table contained the defaults, having an insert without the columns will check those defaults.

Howe due to the comment https://github.com/storj/storj/pull/2215#discussion_r294725233 it cannot be in that way because they don't have to be set on them.

satellite/satellitedb/testdata/postgres.v32.sql
ifraixedes: Ok, I thought that this file had to have the exactly representation of the DB schema at such version.

PR #2222, opened 2019-06-17
Create protocol buffer for garbage collection

Fadila82:
What: Protobuf for exchanging bloom filters between the satellite and the storage node

Why: to be able to exchange bloom filters

Thanks for submitting a PR!


pkg/pb/gc.proto
```diff
@@ -0,0 +1,28 @@
...
+
+message BloomFilter {
+    uint32 seed = 1;
+    uint32 k = 2;
+    uint32 bits_per_element = 3;
+    bytes table = 4;
+}
+
+message DeleteRequest {
+    google.protobuf.Timestamp creation_date = 1;
```
kaloyan-raev: Add `[(gogoproto.stdtime) = true]` like here: https://github.com/storj/storj/blob/15e74c8c3d68821eac13f4d2cccf89ba4604b237/pkg/macaroon/types.proto#L27

This will set the type to `time.Time` in the generated Go code, which is more convenient and we won't need to convert from Timestamp to Time.

pkg/pb/gc.proto
```diff
@@ -0,0 +1,28 @@
...
+
+package gc;
+
+import "google/protobuf/timestamp.proto";
+
+service GarbageCollection {
+    rpc Delete(DeleteRequest) returns (DeleteResponse);
+}
+
+message BloomFilter {
```
kaloyan-raev: Why not just `Filter` as we defined it in the design doc?

Fadila82: I don't know :) Changed.

pkg/pb/gc.proto
```diff
@@ -0,0 +1,30 @@
...
+option go_package = "pb";
+
+package gc;
+
+
+import "gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+service GarbageCollection {
+    rpc Delete(DeleteRequest) returns (DeleteResponse);
```
egonelbre: I'm not sure `Delete` is the right word for this. Collect or Retain might be better.

Fadila82: I think Retain is indeed better. I've changed it.

Fadila82: I think Retain is indeed better. I've changed it.

docs/design/garbage-collection.md
```diff
@@ -53,7 +53,7 @@ What could be sent to the node:
 ### Service
 ```protobuf
 service GarbageCollection {
-    rpc Delete(DeleteRequest) returns (DeleteResponse);
+    rpc Retain(DeleteRequest) returns (RetainResponse);
```
kaloyan-raev: DeleteRequest -> RetainRequest?

pkg/pb/gc.proto
```diff
@@ -0,0 +1,30 @@
...
+
+service GarbageCollection {
+    rpc Retain(RetainRequest) returns (RetainResponse);
+}
+
+message Filter {
+    uint32 seed = 1;
+    uint32 k = 2;
+    uint32 bits_per_element = 3;
+    bytes table = 4;
```
kaloyan-raev: How is the storage node supposed to create a `bloomfilter.Filter` struct from this protobuf?

`bloomfilter.NewFilter` from #2113 is taking `expectedElements int, p float64` params only.

pkg/pb/gc.proto
```diff
@@ -0,0 +1,30 @@
...
+}
+
+message RetainRequest {
+    google.protobuf.Timestamp creation_date = 1 [(gogoproto.stdtime) = true];
```
kaloyan-raev: I've just learnt this can be actually annotated with `[(gogoproto.stdtime) = true, (gogoproto.nullable) = false]`, which will make `timestamp` of type `time.Time` instead of `*time.Time`.

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/jtolds/monkit-hw v0.0.0-20190108155550-0f753668cf20
 	github.com/klauspost/cpuid v0.0.0-20180405133222-e7e905edc00e // indirect
 	github.com/klauspost/reedsolomon v0.0.0-20180704173009-925cb01d6510 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
egonelbre: Note, revert the `godebug`

PR #2225, opened 2019-06-17
satellite/marketing: Stop Offer

fuskovic:
What: Implements a hander that stops the current referral offer or current free credit and replaces it with the default offer for its offer type.

Why: The referral admin should be able to manually stop the current offer before its redeem cap is reached if they feel the need to do so.

Thanks for submitting a PR!


satellite/marketing/marketingweb/server.go
```diff
@@ -186,11 +190,23 @@ func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffe
 	return o,nil
 }
 
+func (s *Server) reRenderTables(w http.ResponseWriter){
+    req,err := http.NewRequest("GET","/",nil)
```
VinozzZ: Shouldn't `Redirect` already making a request to `/`?

VinozzZ: This will be the same user who made the `POST` request tho. Does it matter if we carry it over?

satellite/marketing/marketingweb/server.go
```diff
@@ -186,11 +190,23 @@ func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffe
 	return o,nil
 }
 
+func (s *Server) reRenderTables(w http.ResponseWriter){
```
egonelbre: Don't use shorthands, use properly named things. `re` usually means `regular expression`.

satellite/marketing/marketingweb/server.go
egonelbre: This `NewRequest` shouldn't be here, it doesn't make sense to make a new request and then redirect. Redirect the original request.

satellite/marketing/marketingweb/server.go
```diff
@@ -186,11 +190,23 @@ func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffe
...
+    req,err := http.NewRequest("GET","/",nil)
+    if err != nil {
+        s.log.Error("Rerender tables err", zap.Error(err))
+        s.serveInternalError(w,req,err)
+        return
+    }
+
+    http.Redirect(w,req,"/",http.StatusFound)
```
egonelbre: Use `http.StatusSeeOther` to prevent propagating `POST`.

satellite/marketing/marketingweb/server.go
```diff
@@ -186,11 +190,23 @@ func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffe
...
+}
+
+
 // Handler for POST requests to create, insert, and start a new offer or credit.
 func (s *Server) createOffer(w http.ResponseWriter, req *http.Request) {
 	o, err := formToStruct(w,req)
 	if err != nil{
-		s.log.Error("err from createFreeCredit Handler", zap.Error(err))
+		s.log.Error("create handler error", zap.Error(err))
```
egonelbre: `parse form failed in createOffer`

satellite/marketing/marketingweb/server.go
jtolds: i think faris just meant "rerender" as one word?

satellite/marketing/marketingweb/server.go
```diff
@@ -186,11 +190,23 @@ func formToStruct(w http.ResponseWriter, req *http.Request) (o marketing.NewOffe
...
 		s.serveInternalError(w,req,err)
```
egonelbre: I think actually this should respond with `http.StatusBadRequest`

satellite/marketing/marketingweb/server.go
```diff
@@ -205,12 +221,12 @@ func (s *Server) createOffer(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if _, err := s.service.InsertNewOffer(context.Background(), &o); err != nil {
-		s.log.Error("createdHandler error", zap.Error(err))
+		s.log.Error("create handler error", zap.Error(err))
```
egonelbre: `failed to insert new offer`

satellite/marketing/marketingweb/server.go
```diff
@@ -225,6 +241,23 @@ func (s *Server) serveNotFound(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (s *Server) stopOffer(w http.ResponseWriter, req *http.Request){
+	offerID,err := strconv.Atoi(mux.Vars(req)["offer_id"])
+	if err != nil {
+		s.log.Error("stop handler error", zap.Error(err))
+		s.serveInternalError(w,req,err)
```
egonelbre: Use correct error code.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -83,4 +83,22 @@ func TestGetOffers(t *testing.T){
 	if resp.Status != "200 OK" {
 		t.Fatalf("response err : %v\n", err)
 	}
+}
+
+func TestStopOffer(t *testing.T){
+	url := "http://127.0.0.1:10003"
```
egonelbre: Don't use fixed port numbers in tests.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -83,4 +83,22 @@ func TestGetOffers(t *testing.T){
...
+	req,err := http.NewRequest("PUT",url+"/stop/1",nil)
```
egonelbre: This doesn't test the server.

web/marketing/pages/referral-offers.html
```diff
@@ -12,11 +12,26 @@
         <div class="col col-heading">Created</div>
         <div class="col col-heading">Expiration</div>
         <div class="col col-heading">Status</div>
-        <div class="col col-heading"></div>
     </div><hr>
     <div class="row offer-heading ">
         <p class="offer-type">DEFAULT&nbsp;OFFER</p>
```
egonelbre: Use css to create proper casing.

web/marketing/pages/free-offers.html
```diff
@@ -3,47 +3,66 @@
...
-            <div class="col">0</div>
-            <div class="col">{{.RedeemableCap}}</div>
-            <div class="col">{{printf "%.10s" .CreatedAt}}</div>
-            <div class="col">{{printf "%.10s" .ExpiresAt}}</div>
-            <div class="col"><a href="#">{{if eq .Status 1}}<strong>Live &#183;</strong>{{else}}off{{end}}</a></div>
-            <div class="col col-heading edit-offer">EDIT</div>
-        </div><hr>
+    {{end}}
+    <div class="row offer-heading ">
+        <p class="offer-type">CURRENT&nbsp;OFFER</p>
```
egonelbre: Use css to create proper casing.

web/marketing/pages/free-offers.html
```diff
@@ -3,47 +3,66 @@
...
+        <div class="col col-heading">Name</div>
+        <div class="col col-heading">Give Credit</div>
+        <div class="col col-heading">Referrals&nbsp;Used</div> 
+        <div class="col col-heading">Redeemable&nbsp;Capacity</div>
+        <div class="col col-heading">Created</div>
+        <div class="col col-heading">Expiration</div>
+        <div class="col col-heading">Status</div>
+    </div><hr>
+    <div class="row offer-heading ">
+        <p class="offer-type">DEFAULT&nbsp;OFFER</p>
```
egonelbre: Use css to create proper casing.

web/marketing/css/style.css
```diff
@@ -178,4 +178,8 @@ input{
 .edit-offer:hover{
     text-decoration: underline;
     color: #656565;
+}
+
+.stop-offer:hover {
```
egonelbre: the `:hover` part is not necessary here, you can default it to `cursor:pointer`.

satellite/marketing/marketingweb/server.go
egonelbre: I think it might mean `redirect`.

satellite/marketing/marketingweb/server.go
fuskovic: For our use-case it doesn't really matter since it's behind a VPN but I think its a good practice in general for security.

satellite/marketing/marketingweb/server.go
fuskovic: It's redirecting to "/" which re-renders to table so both. Maybe I'll go with `refreshTables` ?

egonelbre: There's no need for the separate func, a `http.Redirect(w, r, "/", http.StatusSeeOther)` should suffice.

satellite/marketing/marketingweb/server.go
fuskovic: Right because if you're debugging it's more clear to get back exactly where it failed in the create  handler. I get it. Will do.

satellite/marketing/marketingweb/server.go
egonelbre: There's no difference security wise.

satellite/marketing/marketingweb/server.go
fuskovic: I was originally re-assigning the method type for the request in the line before (i.e. add `req.Method = "GET"` before `http.Redirect(....`) and that would also prevent propagating a POST and I was looking for a way to do it on one line but I missed `http.StatusSeeOther`. Will use that instead.

satellite/marketing/marketingweb/server.go
fuskovic: will do.

satellite/marketing/marketingweb/server.go
fuskovic: `Failed to parse offer ID` ?

satellite/marketing/marketingweb/server.go
fuskovic: Ok I'll get rid of it then.

satellite/marketing/marketingweb/server.go
VinozzZ: This doesn't seem to be changed to use `http.StatusSeeOther`

satellite/marketing/marketingweb/server_test.go
```diff
@@ -76,3 +78,46 @@ func TestCreateOffer(t *testing.T) {
 		}
 	})
 }
+
+func TestStopOffer(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
```
VinozzZ: We don't need to run another `testplanet` instance here. We can just add these test cases into the first one

satellite/marketing/marketingweb/server.go
```diff
@@ -52,7 +53,7 @@ type Server struct {
 
 // offerSet provides a separation of marketing offers by type.
 type offerSet struct {
-	RefOffers, FreeCredits []marketing.Offer
+	ReferralOffers, FreeCredits []marketing.Offer
```
egonelbre: Define the fields on separate lines.

web/marketing/pages/free-offers.html
```diff
@@ -27,24 +26,45 @@
...
+                    <span data-toggle="modal" data-target=".stop-free-credit-modal">
+                        <strong class="stop-offer">Live &#183;</strong>
+                    </span>
+                </div>    
+            </div><hr>
         {{end}}
+    {{end}}
+    {{template "stopFreeCredit" .}}
+    <div class="row offer-heading ">
+        <p class="offer-type">other&nbsp;offers</p>
```
egonelbre: `Other&nbsp;Offers`, write the header as if it were without styling. Title case is usual for titles. So the HTML contains title-case, but css transforms it as needed. This makes screen readers understand better what's happening.

satellite/marketing/marketingweb/server_test.go
```diff
@@ -69,6 +72,11 @@ func TestCreateOffer(t *testing.T) {
 					return err
 				}
 
+				_, err = http.Get(baseURL + "/stop/" + id)
```
VinozzZ: Should it be a `PUT` request if we are updating the data ?

egonelbre: Also, if this is working, then we are missing the appropriate method check.

satellite/marketing/marketingweb/server.go
```diff
@@ -248,7 +252,25 @@ func (s *Server) serveNotFound(w http.ResponseWriter, req *http.Request) {
...
+// StopOffer expires the current offer and replaces it with the default offer.
+func (s *Server) StopOffer(w http.ResponseWriter, req *http.Request) {
+	offerID, err := strconv.Atoi(mux.Vars(req)["offer_id"])
+	if err != nil {
+		s.log.Error("failed to parse offer id", zap.Error(err))
+		s.serveBadRequest(w, req, err)
+		return
+	}
+
+	if err := s.service.FinishOffer(context.Background(), offerID); err != nil {
```
VinozzZ: `req.Context()` ?

satellite/marketing/marketingweb/server_test.go
fuskovic: I'm not using a `PUT` request I'm using a `GET` and generating dynamic params
`<a href="/stop/{{range .FreeCredits}}{{if eq .Status 2}}{{.ID}}{{end}}{{end}}">`

There's no need to use a form for this and even if we did you can't use `PUT` methods in an html form.

web/marketing/pages/referral-offers.html
```diff
@@ -7,14 +7,30 @@
...
+                <div class="col">{{printf "%.10s" .CreatedAt}}</div>
+                <div class="col">&#8734;</div>
+            </div><hr>
+        {{end}}
+    {{end}}
+    <div class="row offer-heading ">
+        <p class="offer-type">current offer</p>
     </div>
     {{range .ReferralOffers}}
         {{if eq .Status 2}}
```
egonelbre: `{{if .Status.IsXYZ}}`, you can call methods on types to avoid these magic number.s

web/marketing/pages/stop-referral-offer.html
```diff
@@ -0,0 +1,36 @@
...
+{{define "stopReferralOffer"}}
+<div class="modal fade stop-referral-offer-modal mt-5" tabindex="-1" role="dialog" aria-labelledby="confirm" aria-hidden="true">
+    <div class="modal-dialog modal-lg">
+        <div class="modal-content">
+            <div class="row d-flex justify-content-between">
+                <img class="mt-3 ml-5" src="/static/img/warning.png" height="30" width="30"/>
+                <p class="mt-3">
+                        Replace 
+                        <strong>
+                            {{range .ReferralOffers}}{{if eq .Status 2}}{{.Name}}{{end}}{{end}}
```
egonelbre: Shouldn't these be separated somehow?

satellite/marketing/marketingweb/server_test.go
egonelbre: If this is doing a modification then it must not use GET. Different proxies may cache the results or the browser may open the link automatically without clicking (prefetching). Either use PUT or POST.

web/marketing/pages/stop-referral-offer.html
fuskovic: In this case the reason it works is because there will only ever be one current offer and one default offer per offer type. Still I don't like the way I did this. I'm going to use your suggestion https://github.com/storj/storj/pull/2225#discussion_r298130490 and create a few boolean evaluator methods I can invoke in my templates to specifically get what I want and avoid messy template logic.

satellite/marketing/marketingweb/server_test.go
fuskovic: Because I can't change the request method of an `<a>` tag that means I have to use a form instead which only allows the `POST` method so I'll go with that.

egonelbre: Yeah, there's no nice way of doing it without form. But the form one isn't too complicated:

````
<form method="DELETE" action="/foo/{{ .ID}}">
    <input type="submit" value="do it">
</form>
````

And you can use post as well:

````
<form method="POST" action="/stop">
    <input type="hidden" id="offer-id" value="{{.ID}}">
    <input type="submit" value="do it">
</form>
````

There are also hacks that use onclick + javascript to attach itself to `a` tag. And I suspect they might have bizarre problems on certain devices.

PR #2226, opened 2019-06-17
draft of removing stats from inspector

wthorp:
What: 
Add reputation fields to overlay.NodeStats.
Removed pb.NodeStats.
Removed CreateStats and GetStats from Inspector.

Why:
Refactoring for reputation, we noticed NodeStats was in the code twice, once as a protocol buffer, once not.
Digging in, it seemed like Inspectors CreateStats and GetStats functions should be using the protocol buffer version of NodeStats.
However, as it turns out, CreateStats and GetStats are unused legacy calls.


pkg/overlay/cache.go
```diff
@@ -108,15 +108,19 @@ type NodeDossier struct {
...
+	Latency90             int64
+	AuditSuccessRatio     float64
+	AuditSuccessCount     int64
+	AuditCount            int64
+	UptimeRatio           float64
+	UptimeSuccessCount    int64
+	UptimeCount           int64
+	LastContactSuccess    time.Time
+	LastContactFailure    time.Time
+	AuditReputationAlpha  float64
```
egonelbre: Is this change supposed to part of this PR?

wthorp: Yes.  It probably shouldn't be, but this tiny change is why we started examining this code in the first place.  https://github.com/storj/storj/pull/2206 was the original scope, but we realize that nobody actually used the protocol buffer version of NodeStats

PR #2227, opened 2019-06-17
incorporate new reputation model into overlay cache methods

navillasa:
What: 
- This PR incorporates new node reputation calculations as well as disqualification into pkg/overlay/cache.go and satellite/satellitedb/overlaycache.go.

Why:
- We want the overlay cache to use the new reputation model for node selection and not the old way.
- **Note** that the way NodeSelectionConfig values are passed around (specifically alpha, beta, lambda, and weight values) is not ideal right now, but we think this is the path of least resistance in working with the existing cache methods in pkg/overlay/cache.go and satellite/satellitedb/overlaycache.go. We do think a refactor for better propagation of NodeSelectionConfig values will be due after this PR though.

Thanks for submitting a PR!


satellite/satellitedb/overlaycache.go
```diff
@@ -575,24 +573,18 @@ func (cache *overlaycache) CreateStats(ctx context.Context, nodeID storj.NodeID,
 		return nil, Error.Wrap(errs.Combine(err, tx.Rollback()))
 	}
 
-	if startingStats != nil {
-		auditSuccessRatio, err := checkRatioVars(ctx, startingStats.AuditSuccessCount, startingStats.AuditCount)
-		if err != nil {
-			return nil, errAuditSuccess.Wrap(errs.Combine(err, tx.Rollback()))
-		}
+	if startingStats == nil {
+		// TODO: add possible sanity checking for alpha and beta values
```
ifraixedes: Why not adding them rather leaving a `TODO`? the amount looks that should be minimal

satellite/satellitedb/locked.go
```diff
@@ -845,10 +845,10 @@ func (m *lockedOverlayCache) UpdateStats(ctx context.Context, request *overlay.U
 }
 
 // UpdateUptime updates a single storagenode's uptime stats.
-func (m *lockedOverlayCache) UpdateUptime(ctx context.Context, nodeID storj.NodeID, isUp bool) (stats *overlay.NodeStats, err error) {
+func (m *lockedOverlayCache) UpdateUptime(ctx context.Context, nodeID storj.NodeID, isUp bool, alpha float64, beta float64, lambda float64, weight float64) (stats *overlay.NodeStats, err error) {
```
ifraixedes: For curiosity, how this file is updated and what it does?

navillasa: you can update this file by using `go generate` inside the `satellite/satellitedb` directory. I think it just puts a lock on the relevant database until the respective methods are finished, but I don't totally understand how it works.

satellite/satellitedb/overlaycache.go
navillasa: that's fair! let me put a wip label on this PR-- this TODO was more of a reminder for myself to address within this PR.

satellite/satellitedb/locked.go
wthorp: Ivan: There are issues with multi-threading in SQLite.  This file essentially generates a wrapper of the database code with a mutex to prevent those issues.  For more context, see this old, unfinished PR: https://github.com/storj/storj/pull/1513

satellite/satellitedb/overlaycache.go
```diff
@@ -565,10 +565,19 @@ func (cache *overlaycache) CreateStats(ctx context.Context, nodeID storj.NodeID,
 		return nil, Error.Wrap(errs.Combine(err, tx.Rollback()))
 	}
 
-	if startingStats == nil {
-		// TODO: add possible sanity checking for alpha and beta values
-
-		startingStats = &overlay.NodeStats{}
+	if startingStats != nil {
+		if startingStats.AuditReputationAlpha <= 0 {
```
ifraixedes: I would do these checks before starting the transaction

navillasa: it turns out we never used this function, oops

satellite/satellitedb/overlaycache.go
```diff
@@ -673,33 +623,50 @@ func (cache *overlaycache) UpdateStats(ctx context.Context, updateReq *overlay.U
...
-		UptimeRatio:        dbx.Node_UptimeRatio(uptimeRatio),
+		TotalAuditCount:       dbx.Node_TotalAuditCount(totalAuditCount),
+		TotalUptimeCount:      dbx.Node_TotalUptimeCount(totalUptimeCount),
+		AuditReputationAlpha:  dbx.Node_AuditReputationAlpha(auditAlpha),
+		AuditReputationBeta:   dbx.Node_AuditReputationBeta(auditBeta),
+		UptimeReputationAlpha: dbx.Node_UptimeReputationAlpha(uptimeAlpha),
+		UptimeReputationBeta:  dbx.Node_UptimeReputationBeta(uptimeBeta),
+	}
+
+	if dbNode.Disqualified == nil {
```
wthorp: did we already check this above?

navillasa: oops good call

satellite/satellitedb/overlaycache.go
```diff
@@ -536,7 +534,7 @@ func (cache *overlaycache) Paginate(ctx context.Context, offset int64, limit int
 }
 
 // Update updates node address
-func (cache *overlaycache) UpdateAddress(ctx context.Context, info *pb.Node) (err error) {
+func (cache *overlaycache) UpdateAddress(ctx context.Context, info *pb.Node, defaults overlay.NodeSelectionConfig) (err error) {
```
aligeti: https://github.com/storj/storj/blob/b638d2560af0ca98d004443c4fb85e5adcec2e7e/satellite/satellitedb/overlaycache.go#L556

should there be any distinction be made between "no sql rows" vs any db get err?


similarly other places when Get_Node_By_Id() is called
dbNode, err := tx.Get_Node_By_Id(ctx, dbx.Node_Id(nodeID.Bytes()))
	if err != nil {
		return nil, Error.Wrap(errs.Combine(err, tx.Rollback()))
	}
like that. the err could be no sql rows and could abort the operation. Just an observation.

satellite/satellitedb/overlaycache.go
```diff
@@ -909,42 +877,30 @@ func convertDBNode(ctx context.Context, info *dbx.Node) (_ *overlay.NodeDossier,
...
+// updateReputation uses the Beta distribution model to determine a node's reputation.
+// lambda is the "forgetting factor" which determines how much past info is kept when determining current reputation score.
+// w is the normalization weight that affects how severely new updates affect the current reputation distribution.
+func updateReputation(isSuccess bool, alpha, beta, lambda, w float64, totalCount int64) (newAlpha, newBeta float64, updatedCount int64) {
+	// v is a single feedback value that allows us to update both alpha and beta
+	var v float64 = -1
+	if isSuccess {
+		v = 1
+	}
+	newAlpha = lambda*alpha + w*(1+v)/2
```
aligeti: --> just a thought, thought can it be simplified?
func updateReputation(isSuccess bool, alpha, beta, lambda, w float64, totalCount int64) (newAlpha, newBeta float64, updatedCount int64) {
	// v is a single feedback value that allows us to update both alpha and beta
	if isSuccess {
		// v= 1
		newAlpha = lambda*alpha + w
		newBeta = lambda * beta
	} else {
		//v = -1
		newAlpha = lambda * alpha
		newBeta = lambda*beta + w
	}
	return newAlpha, newBeta, totalCount + 1
}

pkg/overlay/config.go
```diff
@@ -34,16 +34,19 @@ type NodeSelectionConfig struct {
...
-	AuditReputationLambda        float64 `help:"the forgetting factor used to calculate the audit SNs reputation" default:"1.0"`
-	AuditReputationWeight        float64 `help:"the normalization weight used to calculate the audit SNs reputation" default:"1.0"`
+	AuditReputationRepairWeight float64 `help:"weight to apply to audit reputation for total repair reputation calculation" default:"1.0"`
+	AuditReputationUplinkWeight float64 `help:"weight to apply to audit reputation for total uplink reputation calculation" default:"1.0"`
+	AuditReputationAlpha0       float64 `help:"the initial shape 'alpha' used to calculate audit SNs reputation" default:"1.0"`
+	AuditReputationBeta0        float64 `help:"the initial shape 'beta' value used to calculate audit SNs reputation" default:"0.0"`
+	AuditReputationLambda       float64 `help:"the forgetting factor used to calculate the audit SNs reputation" default:"1.0"`
+	AuditReputationWeight       float64 `help:"the normalization weight used to calculate the audit SNs reputation" default:"1.0"`
+	// TODO(nat): figure out default for DQs
```
ifraixedes: Are we going to know this before merging?

pkg/overlay/cache.go
```diff
@@ -92,6 +90,16 @@ type UpdateRequest struct {
 	NodeID       storj.NodeID
 	AuditSuccess bool
 	IsUp         bool
+
+	// n.b. these are set values from the satellite.
+	// They are part of the UpdateRequest struct in order to be
+	// more easily accessible in satellite/satellitedb/overlaycache.go.
+	AuditLambda  float64
```
ifraixedes: I think that I'm missing something, why do we store the lambadas and the weights for each node?

Shouldn't those to be equal for each node?

I've read the comment but I don't fully understand and thinking if there is a way to access to them without having to store them for each node.

pkg/overlay/config.go
mobyvb: Yes, we are waiting on some initial values from Dr. Ben

pkg/overlay/cache.go
mobyvb: They should be equal for for each node. The issue is we couldn't find an easy way to set those values on the satellite config and access them from the satellitedb (`satellite/satellitedb/overlaycache.go`) without passing them down from the overlay cache (`pkg/overlay/cache.go`)

ifraixedes: What about passing them to [`pkg/overlay.NewCache`](https://godoc.org/storj.io/storj/pkg/overlay#NewCache) and satellite can provide them from the [`Config`](https://godoc.org/storj.io/storj/satellite#Config) when the [_Overlay cache_ is instantiated](https://github.com/storj/storj/blob/master/satellite/peer.go#L263)

mobyvb: That is already happening. We updated the config with values in `pkg/overlay/config.go`, and the `Node` struct in that file is being passed into `NewCache`. Then, we pass those values (stored in `cache.preferences` down into the db layer

navillasa: yeah the issue is that the two cache structs are different. foiled by wrapper functions once again!

ifraixedes: I don't get it then why we have to store those values in the DB, but anyway, if you're clear about it it's fine to me as I'm less familiar than you with the code base.

navillasa: we're not storing lambda or weight in the db, we're only storing alpha and beta in the db. we're passing lambda and weight to the "db-level" to use to calculate the reputation/determine if disqualified.

ifraixedes: oh, I see, I got confused as my first question was referencing to that and I didn't get this response :sweat_smile: 

:thinking: I prefer to have those calculations on the business layer, but I guess that you're doing in the DB layer because you do it in the same `Update..` function rather than having to call `Get` and `Update`

navillasa: right, I also agree it's not ideal but hopefully we'll refactor the overlay cache to separate these kinds of concerns soon.

pkg/audit/verifier.go
```diff
@@ -93,9 +93,10 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
 		return nil, err
 	}
 
+	// note: offlineNodes here will include disqualified nodes
```
kaloyan-raev: Hmm... this means that DQ nods will appear as offline in the report. DQ nodes should not appear in the report at all.

navillasa: this should probably be better handled but Moby and I had the justification that when UpdateStats is called when reporting, the disqualified nodes will be ignored/their stats won't be updated.

navillasa: alternately, CreateAuditOrderLimits could also a return a list of nodes that were disqualified, and those could be skipped but ignored by the report.

kaloyan-raev: I think you made the right call for now. We need to think about it. Because it won't be DQ nodes only. Soon we will have nodes that exited gracefully too.

PR #2231, opened 2019-06-18
support value attribution endpoint

aligeti:
What: 
Support value attribution endpoint to be able to read/write partner info into value attribution DB.
Why:
support the value attribution feature
https://storjlabs.atlassian.net/browse/V3-1998

Thanks for submitting a PR!

### DIFF
--- a/satellite/attribution/db.go
+++ b/satellite/attribution/db.go
@@ -12,8 +12,8 @@ import (
 	"github.com/zeebo/errs"
 )
 
-// ErrNoAttributionFound is returned if a node does not exist in database
-var ErrNoAttributionFound = errs.Class("bucket not attributed")
+// ErrBucketNotAttributed is returned if a requested bucket not attributed(entry not found)
+var ErrBucketNotAttributed = errs.Class("bucket not attributed")
 
 // Info describing value attribution from partner to bucket
 type Info struct {
--- a/satellite/attribution/db.go
+++ b/satellite/attribution/db.go
@@ -12,8 +12,8 @@ import (
 	"github.com/zeebo/errs"
 )
 
-// ErrBucketNameNotFound is returned if a node does not exist in database
-var ErrBucketNameNotFound = errs.Class("bucket name not found")
+// ErrNoAttributionFound is returned if a node does not exist in database
+var ErrNoAttributionFound = errs.Class("bucket not attributed")
 
 // Info describing value attribution from partner to bucket
 type Info struct {
--- a/satellite/attribution/db.go
+++ b/satellite/attribution/db.go
@@ -9,7 +9,11 @@ import (
 	"time"
 
 	"github.com/skyrings/skyring-common/tools/uuid"
+	"github.com/zeebo/errs"
 )
+
+// ErrBucketNameNotFound is returned if a node does not exist in database
+var ErrBucketNameNotFound = errs.Class("bucket name not found")
 
 // Info describing value attribution from partner to bucket
 type Info struct {
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -536,12 +536,16 @@ func (endpoint *Endpoint) SetAttribution(ctx context.Context, req *pb.SetAttribu
 				return nil, err
 			}
 
-			items, _, err := endpoint.metainfo.List(ctx, prefix, "", "", true, 1, 0)
+			items, _, err := endpoint.metainfo.List(ctx, prefix, "", "", true, 2, 0)
 			if err != nil {
 				return nil, err
 			}
 
-			if len(items) > 0 {
+			if len(items) == 0 {
+				return nil, Error.New("bucket doesnt exists")
+			}
+
+			if len(items) > 1 {
 				return nil, Error.New("Bucket(%s) , PartnerID(%s) cannot be attributed", string(req.BucketName), string(req.PartnerId))
 			}
 
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -528,7 +528,7 @@ func (endpoint *Endpoint) SetAttribution(ctx context.Context, req *pb.SetAttribu
 		if attribution.ErrBucketNameNotFound.Has(err) {
 			partnerID, err := bytesToUUID(req.GetPartnerId())
 			if err != nil {
-				return &pb.SetAttributionResponse{}, nil
+				return nil, err
 			}
 			_, err = endpoint.partnerinfo.Insert(ctx, &attribution.Info{
 				ProjectID:  *projectID,
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -539,6 +539,7 @@ func (endpoint *Endpoint) SetAttribution(ctx context.Context, req *pb.SetAttribu
 	}
 
 	if !attribution.ErrBucketNotAttributed.Has(err) {
+		// try only to set the attribution, when it's missing
 		return nil, Error.Wrap(err)
 	}
 
@@ -553,7 +554,7 @@ func (endpoint *Endpoint) SetAttribution(ctx context.Context, req *pb.SetAttribu
 	}
 
 	if len(items) > 0 {
-		return nil, Error.New("Bucket(%s) , PartnerID(%q) cannot be attributed", string(req.BucketName), req.PartnerId)
+		return nil, Error.New("Bucket(%q) , PartnerID(%s) cannot be attributed", req.BucketName, req.PartnerId)
 	}
 
 	_, err = endpoint.partnerinfo.Insert(ctx, &attribution.Info{
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -516,6 +516,12 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
 func (endpoint *Endpoint) SetAttribution(ctx context.Context, req *pb.SetAttributionRequest) (_ *pb.SetAttributionResponse, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	// try to add an attribution that doesn't exist
+	partnerID, err := bytesToUUID(req.GetPartnerId())
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
 	keyInfo, err := endpoint.validateAuth(ctx, macaroon.Action{
 		Op:            macaroon.ActionList,
 		Bucket:        req.BucketName,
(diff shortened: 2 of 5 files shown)


satellite/metainfo/metainfo.go
```diff
@@ -548,7 +551,23 @@ func (endpoint *Endpoint) ValueAttributeInfo(ctx context.Context, req *pb.ValueA
...
+	_, err = endpoint.partnerinfo.Get(ctx, req.GetBucketName())
+	if err != nil {
+		// no rows
+		if valueattribution.ErrBucketNameNotFound.Has(err) {
+			_, err = endpoint.partnerinfo.Insert(ctx, &valueattribution.PartnerInfo{
+				PartnerID:  req.GetPartnerId(),
+				BucketName: req.GetBucketName(),
+			})
+			if err != nil {
+				endpoint.log.Sugar().Info("related partner info could not be attributed \n")
```
mobyvb: does the logger need a newline at the end?

satellite/metainfo/metainfo_test.go
```diff
@@ -453,29 +453,29 @@ func TestValueAttributeInfo(t *testing.T) {
...
 
 		keyInfo := pb.ValueAttributionRequest{
-			PartnerId:  []byte("PartnerID"),
-			BucketName: []byte("myBucketName"),
+			PartnerId:  []byte("I am the new PartnerID"),
+			BucketName: []byte("myProjectID  +  myBucketName"),
 		}
 
 		{
 			// bucket with no items
```
mobyvb: I do not think this is a bucket with no items. I do not think it is a bucket at all (which is the test case immediately below). The bucket that is created is called "myBucket", but `string(keyInfo.GetBucketName())` here is "myProjectID + myBucketName"

satellite/metainfo/metainfo_test.go
```diff
@@ -453,29 +453,29 @@ func TestValueAttributeInfo(t *testing.T) {
...
 			require.NoError(t, err)
 		}
 		{
 			expectedData := make([]byte, 1*memory.MiB)
 			_, err = rand.Read(expectedData)
 			assert.NoError(t, err)
 
-			err = planet.Uplinks[0].Upload(ctx, planet.Satellites[0], "myBucket", "path", expectedData)
+			err = planet.Uplinks[0].Upload(ctx, planet.Satellites[0], string(keyInfo.GetBucketName()), "path", expectedData)
 			assert.NoError(t, err)
```
mobyvb: use `require` unless `assert` is necessary

satellite/metainfo/metainfo.go
aligeti: refactored the code and removed the logs are they are redundant

satellite/metainfo/metainfo_test.go
aligeti: code got refactored and renamed.

satellite/metainfo/metainfo_test.go
aligeti: code refactored and test cases updated.

satellite/metainfo/metainfo.go
```diff
@@ -511,20 +514,37 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-	// TODO: add valueattribution DB access functions added in new PR.
-
+	// check if already bucket exists
+	_, err = endpoint.partnerinfo.Get(ctx, *projectID, req.GetBucketName())
+	if err != nil {
+		// no rows found for the bucket name
+		if attribution.ErrBucketNameNotFound.Has(err) {
+			partnerID, err := bytesToUUID(req.GetPartnerId())
+			if err != nil {
+				return &pb.SetAttributionResponse{}, nil
```
egonelbre: Return an error to the caller.

satellite/metainfo/metainfo.go
```diff
@@ -511,20 +514,37 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-		endpoint.log.Sugar().Debug("related bucket id already attributed \n")
-		return &pb.SetAttributionResponse{}, err
+		return nil, err
 	}
 
-	// TODO: add valueattribution DB access functions added in new PR.
-
+	// check if already bucket exists
+	_, err = endpoint.partnerinfo.Get(ctx, *projectID, req.GetBucketName())
+	if err != nil {
```
egonelbre: Check for the correct error.

satellite/metainfo/metainfo.go
aligeti: done

aligeti:
updated the doc accordingly with the dbx model fields.

satellite/metainfo/metainfo.go
```diff
@@ -536,22 +556,34 @@ func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.SetAt
...
 
 	items, _, err := endpoint.metainfo.List(ctx, prefix, string(""), string(""), true, 1, 0)
 	if err != nil {
-		return false, err
+		return nil, err
 	}
 
 	if len(items) > 0 {
-		return false, errors.New("already attributed")
+		return nil, errors.New("already attributed")
```
mniewrzal: it doesn't mean that it's `attributed`, it means that cannot be attributed because its not empty.

satellite/metainfo/metainfo.go
```diff
@@ -536,22 +556,34 @@ func (endpoint *Endpoint) checkBucketPointers(ctx context.Context, req *pb.SetAt
...
+		return nil, status.Errorf(codes.Unauthenticated, err.Error())
 	}
 
 	prefix, err := CreatePath(ctx, keyInfo.ProjectID, -1, req.BucketName, []byte(""))
 	if err != nil {
-		return false, err
+		return nil, err
 	}
 
 	items, _, err := endpoint.metainfo.List(ctx, prefix, string(""), string(""), true, 1, 0)
```
mniewrzal: such conversion is unnecessary (`string("")`)

satellite/metainfo/metainfo.go
```diff
@@ -511,20 +514,37 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-		// TODO: return correct status code for GRPC
-		endpoint.log.Sugar().Debug("related bucket id already attributed \n")
-		return &pb.SetAttributionResponse{}, err
+		return nil, err
 	}
 
-	// TODO: add valueattribution DB access functions added in new PR.
-
+	// check if already bucket exists
+	_, err = endpoint.partnerinfo.Get(ctx, *projectID, req.GetBucketName())
```
mniewrzal: I feel that logic in SetAttribution is a little bit awkward. I believe it would be more optimal to have such steps:
* check if attribution is set for given bucket
* return error if yes
* check if bucket exists 
* return error if no
* check if bucket is empty
* return error if no
* set attribution

satellite/satellitedb/attribution.go
```diff
@@ -24,6 +25,9 @@ func (keys *attributionDB) Get(ctx context.Context, projectID uuid.UUID, bucketN
 		dbx.ValueAttribution_ProjectId(projectID[:]),
 		dbx.ValueAttribution_BucketName(bucketName),
 	)
+	if err == sql.ErrNoRows {
+		return nil, attribution.ErrBucketNameNotFound.New(string(bucketName))
```
mniewrzal: its not that bucket is not found, it means that there are no attribution for such bucket name.

satellite/metainfo/metainfo.go
aligeti: updated to return bit more info --> return nil, errors.New("bucket: " + string(req.BucketName) + " cannot be attributed")

satellite/metainfo/metainfo.go
aligeti: updated

satellite/metainfo/metainfo.go
aligeti: refactored the handling of bucket attribution as suggested

satellite/satellitedb/attribution.go
aligeti: updated the error to ErrNoAttributionFound = errs.Class("bucket not attributed").

satellite/attribution/db.go
```diff
@@ -9,8 +9,12 @@ import (
 	"time"
 
 	"github.com/skyrings/skyring-common/tools/uuid"
+	"github.com/zeebo/errs"
 )
 
+// ErrNoAttributionFound is returned if a node does not exist in database
+var ErrNoAttributionFound = errs.Class("bucket not attributed")
```
mniewrzal: variable name should reflect error message e.g. `ErrBucketNotAttributed  = errs.Class("bucket not attributed")`. Also I think comment is copy pasted

aligeti: oops... !!!fixed

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,63 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-	items, _, err := endpoint.metainfo.List(ctx, prefix, string(""), string(""), true, 1, 0)
-	if err != nil {
-		return false, err
+			items, _, err := endpoint.metainfo.List(ctx, prefix, "", "", true, 1, 0)
+			if err != nil {
+				return nil, err
+			}
+
+			if len(items) > 0 {
+				return nil, Error.New("Bucket(%s) , PartnerID(%s) cannot be attributed", string(req.BucketName), string(req.PartnerId))
```
egonelbre: Use `%q`, partnerID should be an uuid, so casting it to a string will end up badly.

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,63 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+			}
+
+			partnerID, err := bytesToUUID(req.GetPartnerId())
```
egonelbre: do this earlier, so you avoid mistakes

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,63 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
 	})
 	if err != nil {
-		return false, status.Errorf(codes.Unauthenticated, err.Error())
+		return nil, status.Errorf(codes.Unauthenticated, err.Error())
 	}
 
-	prefix, err := CreatePath(ctx, keyInfo.ProjectID, -1, req.BucketName, []byte(""))
+	// check if attribution is set for given bucket
+	_, err = endpoint.partnerinfo.Get(ctx, keyInfo.ProjectID, req.GetBucketName())
 	if err != nil {
```
egonelbre:
```
if err == nil {
    return nil, Error.New("Bucket(%s) , PartnerID(%s) cannot be attributed", string(req.BucketName), string(req.PartnerId))
}
```

to reduce identation

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,63 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-		return false, err
-	}
+		if attribution.ErrBucketNotAttributed.Has(err) {
```
egonelbre: same here

```
if !attribution.ErrBucketNotAttributed.Has(err) {
   return nil, err
}

// try to add an attribution that doesn't exist
```

satellite/metainfo/metainfo.go
coyle: uuid supports the stringer interface

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,63 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+			if err != nil {
+				return nil, err
+			}
+			_, err = endpoint.partnerinfo.Insert(ctx, &attribution.Info{
+				ProjectID:  keyInfo.ProjectID,
+				BucketName: req.GetBucketName(),
+				PartnerID:  partnerID,
+			})
+			if err != nil {
+				return nil, err
```
egonelbre: wrap the naked errors

satellite/metainfo/metainfo.go
egonelbre: Here `req.PartnerId` is still a `[]byte` and `string(xyz)` doesn't use `xyz.String()`.

aligeti: updated

satellite/metainfo/metainfo.go
aligeti: moved above

satellite/metainfo/metainfo.go
aligeti: wrapped the naked errors

satellite/metainfo/metainfo.go
aligeti: updated

satellite/metainfo/metainfo.go
aligeti: updated as suggested

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,65 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
+		return nil, status.Errorf(codes.Unauthenticated, err.Error())
+	}
+
+	// check if attribution is set for given bucket
+	_, err = endpoint.partnerinfo.Get(ctx, keyInfo.ProjectID, req.GetBucketName())
+	if err == nil {
+		return nil, Error.New("Bucket(%s) , PartnerID(%s) cannot be attributed", string(req.BucketName), string(req.PartnerId))
+	}
+
+	if !attribution.ErrBucketNotAttributed.Has(err) {
```
egonelbre: add a comment, something like:
```
// try only to set the attribution, when it's missing
```

since the statement itself is a double negation, which is slightly harder to understand.

satellite/metainfo/metainfo.go
```diff
@@ -511,47 +514,65 @@ func CreatePath(ctx context.Context, projectID uuid.UUID, segmentIndex int64, bu
...
-	items, _, err := endpoint.metainfo.List(ctx, prefix, string(""), string(""), true, 1, 0)
+	items, _, err := endpoint.metainfo.List(ctx, prefix, "", "", true, 1, 0)
 	if err != nil {
-		return false, err
+		return nil, Error.Wrap(err)
 	}
 
 	if len(items) > 0 {
-		return false, errors.New("already attributed")
+		return nil, Error.New("Bucket(%s) , PartnerID(%q) cannot be attributed", string(req.BucketName), req.PartnerId)
```
egonelbre: I meant use %q for bucket name, and %s for partner id is probably fine.

I'm not sure whether %q works for uuids

satellite/metainfo/metainfo.go
aligeti: felt that was pretty self explanatory. Will modify as suggested

PR #2233, opened 2019-06-18
Avoid reporting audit failures for files deleted during audit

kaloyan-raev:
What: 

The audit verifier will check if the segment wasn't deleted while the erasure shares were being downloaded from storage nodes. If this is cases, the audit is aborted as unreliable.

In a similar way, if a contained node is being reverified, after erasure share is downloaded, it is again verified if the node is still contained. If not contained, this means that the segment was deleted meanwhile, and the audit results are unreliable.

Why:

Fixes https://storjlabs.atlassian.net/browse/V3-1918. We don't want to report audit failures if a file is deleted while being audited.

Thanks for submitting a PR!


pkg/audit/verifier.go
```diff
@@ -470,6 +500,30 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+
+	pointer, err := verifier.metainfo.Get(ctx, stripe.SegmentPath)
+	if err != nil {
+		if storage.ErrKeyNotFound.Has(err) {
+			return ErrSegmentDeleted.New(stripe.SegmentPath)
+		}
+		return err
+	}
+
+	if pointer == nil {
```
ifraixedes: Out of curiosity, when does this case happen?

I think that I asked this before, but I don't remember why.

kaloyan-raev: It should not happen. If such pointer does not exist, `metainfo.Get` should return `storage.ErrKeyNotFound` error. But I felt it would be safer if I check `pointer` for `nil` in case something changes in the metainfo service in future.

Should I remove this check?

ifraixedes: From the safety point of view it's good because the service won't crash, but if that happens because of a bug, we may want to know, hence or we crash or we could inform about this unexpected behavior logging the case.

My understanding is that if we do change the behavior of public API, we should try not introducing breaking changes and if we do, we should care about them, however crashing could be to bad for a node operator, so the logging line could avoid it and inform of it.

To be honest, up to you, I thought there was a case that it could happen but I couldn't remembered.

kaloyan-raev: I removed the nil check. We don't do it other places in the code. Instead, we are using the Get* methods of the Pointer protobuf, which are nil-safe, instead of accessing the fields directly.

PR #2242, opened 2019-06-19
satellite: remove from kademlia

jtolds:
This PR
1. cleans up some kademlia peer discovery logic and fixes some kademlia bugs. Specifically, it makes sure the discovery queue works as specified by the kademlia paper (should use K instead of alpha for queue depth, queried nodes should take queue space, etc). these adjustments combined should make kademlia lookups both faster and more secure
2. sends the satellite's external address in order limits
3. has the storage node use the satellite's external address directly instead of attempting a kademlia lookup
4. introduces a non-participant kademlia client
5. removes the satellite from the kademlia network entirely

The point of all of this is to make it much easier to break a satellite into multiple subservices, as described in https://docs.google.com/document/d/1YelTiTvAFnUKgrFr4DLDv73heuQFXGA4XTCtcs4z5AQ/edit. As it stands, having a microservice-oriented cluster satellite keep a routing table up to date is more work than this route.

Some of the constituent commits could be committed individually I suppose


pkg/kademlia/client.go
```diff
@@ -0,0 +1,92 @@
...
+	config ClientConfig
+	dialer *Dialer
+
+	mu             sync.Mutex
+	resolved       bool
+	bootstrapNodes []*pb.Node
+}
+
+// NewClient returns a newly configured Kademlia client
+func NewClient(log *zap.Logger, transport transport.Client, bootstrapNodes []*pb.Node, config ClientConfig) *Client {
```
egonelbre: Close is missing for Client

pkg/kademlia/kademlia.go
```diff
@@ -36,20 +36,13 @@ var (
...
-}
-
 // Kademlia is an implementation of kademlia adhering to the DHT interface.
 type Kademlia struct {
 	log            *zap.Logger
 	alpha          int // alpha is a system wide concurrency parameter
 	routingTable   *RoutingTable
 	bootstrapNodes []pb.Node
-	dialer         *Dialer
+	Dialer         *Dialer
```
egonelbre: Does this need to be public?

satellite/orders/service.go
```diff
@@ -596,15 +604,16 @@ func (service *Service) CreatePutRepairOrderLimits(ctx context.Context, repairer
 		}
 
 		orderLimit, err := signing.SignOrderLimit(ctx, service.satellite, &pb.OrderLimit2{
```
egonelbre: It looks like orderlimit content has been changed but the signing and verification not.

storagenode/trust/service.go
```diff
@@ -131,7 +132,7 @@ func (pool *Pool) GetSignee(ctx context.Context, id storj.NodeID) (_ signing.Sig
 	defer info.mu.Unlock()
 
 	if info.identity == nil {
-		identity, err := pool.kademlia.FetchPeerIdentity(ctx, id)
+		identity, err := pool.dialer.FetchPeerIdentityUnverified(ctx, address.Address)
```
egonelbre: How do we verify that the uplink has sent us the correct address and signature?

pkg/kademlia/config.go
```diff
@@ -36,17 +39,65 @@ type Config struct {
...
+//	 node ::= ( <ipv4> | `[` <ipv6> `]` ) `:` <port> [ `#` <nodeid> ]
+//
+// Examples of individual nodes:
+//
+//   33.20.0.1:7777
+//   33.20.0.1:7777#ekC4dHif4NAGTTFtniBbcLuhPGoujdgNIJf313
+//	 [2001:db8:1f70::999:de8:7648:6e8]:7777
+//	 [2001:db8:1f70::999:de8:7648:6e8]:7777#ekC4dHif4NAGTTFtniBbcLuhPGoujdgNIJf313
+//
+func NodesFromConfig(cfg string) (nodes []*pb.Node, err error) {
```
egonelbre: Tests are missing

pkg/kademlia/client.go
```diff
@@ -0,0 +1,92 @@
...
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/transport"
+)
+
+type ClientConfig struct {
+	K     int `help:"K is the system-wide Kademlia neighborhood size" default:"20"`
+	Alpha int `help:"alpha is a system wide concurrency parameter" default:"5"`
+}
+
+type Client struct {
```
egonelbre: Tests are missing

pkg/kademlia/peer_discovery.go
```diff
@@ -147,57 +151,37 @@ type queueItem struct {
 }
 
 // newDiscoveryQueue returns a items with priority based on XOR from targetBytes
-func newDiscoveryQueue(size int) *discoveryQueue {
+func newDiscoveryQueue(target storj.NodeID, K int) *discoveryQueue {
```
egonelbre: why is K uppercase here? in the context of a queue `size` makes more sense than `k`

jtolds: addressed in https://github.com/storj/storj/pull/2252

jtolds:
i aimed to make the review easier (with multiple dependent changes) by making the commits in this branch fairly contained and standalone. you should be able to review the commits independently, but i'll push up separate prs.

egonelbre:
Yeah, I understand. The issue is that currently there are comments for various steps... so effectively all the different commits will start getting changes. Effectively, it's hard to tell which comments have been addressed and which not. i.e. the initial picture is clear, however after adding fixes on top of current implementation, it becomes hard to track... or in case of fixing history, github breaks, and we cannot see which comments have been addressed.

storagenode/trust/service.go
jtolds: whoops, i should have compared identity.ID to the `id` variable. if they don't match, then we failed to get the public key for the id we actually cared about. good catch

satellite/orders/service.go
jtolds: what else needs to change?

pkg/kademlia/config.go
egonelbre: I created https://github.com/storj/storj/pull/2379 for this

jenlij:
closing PR

PR #2243, opened 2019-06-19
cmd/storj-sim: ensure prefix writer can be used concurrently

egonelbre:
What: fixes race https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/PR-2238/10/pipeline

Why:

Thanks for submitting a PR!


cmd/storj-sim/prefix.go
```diff
@@ -54,7 +56,13 @@ func (writer *PrefixWriter) Prefixed(prefix string) io.Writer {
 	writer.prefixlen = max(writer.prefixlen, len(prefix))
 	writer.mu.Unlock()
 
-	return &prefixWriter{writer, prefix, "", make([]byte, 0, writer.maxline)}
+	return &prefixWriter{
+		PrefixWriter: writer,
+		prefix:       prefix,
+		local:        sync.Mutex{},
```
mniewrzal: just curious, do we need explicitly create new `sync.Mutex`?

egonelbre: ah, not anymore, initially it wasn't using named fields, so yes... we can do without it

PR #2244, opened 2019-06-19
Bucket name validation

mniewrzal:
What: Adds more restrictive validation for bucket name.
```
Bucket names must comply with DNS naming conventions.
Bucket names must be at least 3 and no more than 63 characters long.
Bucket names must not contain uppercase characters or underscores.
Bucket names must start with a lowercase letter or number.
Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
Bucket names must not be formatted as an IP address (for example, 192.168.5.4).
```
Why: We need to have strict rules about bucket names. This is part of changes for https://storjlabs.atlassian.net/browse/V3-1787


satellite/metainfo/validation.go
```diff
@@ -225,12 +230,56 @@ func (endpoint *Endpoint) validateBucket(ctx context.Context, bucket []byte) (er
 	if len(bucket) == 0 {
 		return errs.New("bucket not specified")
 	}
-	if bytes.ContainsAny(bucket, "/") {
-		return errs.New("bucket should not contain slash")
+
+	if len(bucket) < 3 || len(bucket) > 63 {
+		return errs.New("bucket name must be at least 3 and no more than 63 characters long")
```
ifraixedes: Shouldn't we use for this errors the `Error` class?

This apply to any other error returned in this function and unexported functions like `validateBucketLabel`

satellite/metainfo/validation.go
```diff
@@ -225,12 +230,56 @@ func (endpoint *Endpoint) validateBucket(ctx context.Context, bucket []byte) (er
...
 	}
+
+	labels := bytes.Split(bucket, []byte("."))
```
ifraixedes: I have the feeling that we could check if the bucket name is correct with this regular expression `^([a-z0-9]+([a-z0-9\-][a-z0-9])*)+(.[a-z0-9]+([a-z0-9\-][a-z0-9])*)*$`

The above checks and the check `ipRegexp.MatchString` must be kept.

mniewrzal: I did a fast comparison with benchmark and if I didn't mess up something it looks that non regexp way is faster for valid bucket names. I think we will be processing more valid bucket names than invalid.
https://gist.github.com/mniewrzal/49de3af95f36e63e88fac24f565e444c
```
BenchmarkValidation/simple-size-5-8         	30000000	        51.7 ns/op	      32 B/op	       1 allocs/op
BenchmarkValidation/regexp-size-5-8         	 5000000	       252 ns/op	       5 B/op	       1 allocs/op
BenchmarkValidation/simple-size-29-8        	20000000	       100 ns/op	      32 B/op	       1 allocs/op
BenchmarkValidation/regexp-size-29-8        	 2000000	       991 ns/op	      32 B/op	       1 allocs/op
BenchmarkValidation/simple-size-59-8        	20000000	       121 ns/op	      32 B/op	       1 allocs/op
BenchmarkValidation/regexp-size-59-8        	 1000000	      1857 ns/op	      65 B/op	       1 allocs/op
PASS
ok  	storj.io/storj/satellite/metainfo	12.664s
```

satellite/metainfo/validation.go
mniewrzal: you are right, done

satellite/metainfo/validation.go
ifraixedes: Let's add a comment to remember why a regex isn't used and link to the gist, where you could add a text file with this result.

satellite/metainfo/metainfo_test.go
```diff
@@ -523,3 +518,52 @@ func createTestPointer(t *testing.T) *pb.Pointer {
 	}
 	return pointer
 }
+
+func TestBucketNameValidation(t *testing.T) {
+	t.Skip("Skip until bucket name validation is not enabled")
```
egonelbre:
```
if BucketNameRestricted {
    t.Skip(...
```

mniewrzal: done

PR #2203, opened 2019-06-13
Adds test coverage to CalculateBucketAtRestData

jenlij:
This was supposed to fix flakiness in TestCalculateBucketAtRestData but I couldn't pinpoint the actual cause of flakiness. I did add more test cases so I changed this PR to address that instead.

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Could the PR be broken into smaller PRs?
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?

PR #2250, opened 2019-06-19
fix tally test flakiness

JessicaGreben:
What/Why:

This PR makes changes to the test for calculating bucket tally data.  This test was originally flaky because the Uplink.Upload would intermittently return an error from a failed upload (it was hitting this error https://github.com/storj/storj/blob/master/satellite/metainfo/metainfo.go#L475).  

Since we are just trying to test the tally code here, I removed the uplink.upload part and created a pointer directly instead. 

See jira ticket and previous PR for details:
- https://github.com/storj/storj/pull/2203
- https://storjlabs.atlassian.net/browse/V3-1963


pkg/accounting/tally/tally_test.go
```diff
@@ -159,123 +162,144 @@ func TestCalculateNodeAtRestData(t *testing.T) {
 }
 
 func TestCalculateBucketAtRestData(t *testing.T) {
-	t.Skip("TODO: this test is flaky")
+	var testCases = []struct {
+		name         string
```
jenlij: awesome, I appreciate when the test cases have names

pkg/accounting/tally/tally_test.go
```diff
@@ -159,123 +162,144 @@ func TestCalculateNodeAtRestData(t *testing.T) {
...
+		projectID    string
+		segmentIndex string
+		bucketName   string
+		objectName   string
+		inline       bool
+		last         bool
+	}{
+		{"inline, same project, same bucket", "mockProjectID", "l", "mockBucketName", "mockObjectName", true, true},
```
jenlij: oh I forgot L meant inline segment, good reminder

pkg/accounting/tally/tally_test.go
```diff
@@ -159,123 +162,144 @@ func TestCalculateNodeAtRestData(t *testing.T) {
...
+		{"remote, same project, same bucket", "mockProjectID", "s0", "mockBucketName", "mockObjectName1", false, false},
+		{"last segment, same project, different bucket", "mockProjectID", "l", "mockBucketName1", "mockObjectName2", false, true},
```
jenlij: why do we want to specifically test the last segment here? just wondering.

pkg/accounting/tally/tally_test.go
```diff
@@ -159,123 +162,144 @@ func TestCalculateNodeAtRestData(t *testing.T) {
...
+		{"different project", "mockProjectID1", "s0", "mockBucketName", "mockObjectName", false, false},
+	}
```
jenlij: what about including a test case where there are no objects in the bucket?

pkg/accounting/tally/tally_test.go
JessicaGreben: l means last segment i believe

pkg/accounting/tally/tally_test.go
JessicaGreben: because the tally code has different behavior if its the last segment. re: https://github.com/storj/storj/blob/master/pkg/accounting/bucketstats.go#L70

pkg/accounting/tally/tally_test.go
JessicaGreben: sounds good. let me add that.

pkg/accounting/tally/tally_test.go
JessicaGreben: me too!

PR #2252, opened 2019-06-19
pkg/kademlia: clean up peer discovery

jtolds:
What: this change cleans up some kademlia peer discovery logic and fixes some kademlia bugs. Specifically, it makes sure the discovery queue works as specified by the kademlia paper (should use K instead of alpha for queue depth, queried nodes should take queue space, etc). 

Why: these adjustments combined should make kademlia lookups both faster and more secure

Thanks for submitting a PR!


jenlij:
cool I'm excited to take a look at this

egonelbre:

pkg/kademlia/dialer.go
```diff
@@ -47,29 +48,37 @@ func (dialer *Dialer) Close() error {
 }
 
 // Lookup queries ask about find, and also sends information about self.
-func (dialer *Dialer) Lookup(ctx context.Context, self pb.Node, ask pb.Node, find pb.Node) (_ []*pb.Node, err error) {
+// If self is nil, pingback will be false.
```
jenlij: When will self be nil?

jenlij:
I can't figure out which test is failing. I must be missing it when I look through the Jenkins logs, and everything passes locally...

pkg/kademlia/dialer.go
```diff
@@ -47,29 +48,37 @@ func (dialer *Dialer) Close() error {
...
+func (dialer *Dialer) Lookup(ctx context.Context, self *pb.Node, ask pb.Node, find storj.NodeID, limit int) (_ []*pb.Node, err error) {
```
jenlij: Just curious, why is the Node slice an anonymous variable?

pkg/kademlia/dialer.go
```diff
@@ -47,29 +48,37 @@ func (dialer *Dialer) Close() error {
...
 	defer mon.Task()(&ctx)(&err)
 	if !dialer.limit.Lock() {
 		return nil, context.Canceled
 	}
 	defer dialer.limit.Unlock()
 
+	req := pb.QueryRequest{
+		Limit:  int64(limit),
+		Target: &pb.Node{Id: find}, // TODO: should not be a Node protobuf!
```
jenlij: What should it be if not a node protobuf?

pkg/kademlia/kademlia.go
```diff
@@ -280,37 +283,33 @@ func (k *Kademlia) FindNode(ctx context.Context, nodeID storj.NodeID) (_ pb.Node
...
 	defer mon.Task()(&ctx)(&err)
 	if !k.lookups.Start() {
-		return pb.Node{}, context.Canceled
+		return nil, context.Canceled
 	}
 	defer k.lookups.Done()
 
-	kb := k.routingTable.K()
-	var nodes []*pb.Node
-	if isBootstrap {
```
jenlij: We don't need to tell it to distinguish between bootstrap or not?

pkg/kademlia/dialer.go
jtolds: perhaps having "self" indicate whether we want ping backs is confusing, but most kademlia lookups don't need ping backs, and therefore, don't need to send self along in the request. there are only a few kademlia lookups that do need pingbacks, and those are specifically kademlia-driven bucket refreshes. having pingbacks enabled for other lookups (pkg/discovery, etc), is too resource intensive

pkg/kademlia/dialer.go
jtolds: during the monkit flurry of additions, i needed a way of naming the `err` return value without doing much else to the semantics of the function. go won't let me just name one return value without naming all of them. we could name the node slice. that would probably be better

pkg/kademlia/dialer.go
jtolds: it should just be a node id. a node also includes an address and so on. all we care about for the target is the id.

pkg/kademlia/kademlia.go
jtolds: the only time bootstrap nodes matter is when your routing table is empty, and only then to bootstrap the lookup. as soon as you have any nodes in your routing table or lookup queue, the normal kademlia operation is totally sufficient. we don't really need other logic.

the key thing in this pr for getting the bootstrap nodes into the mix is making sure we call the transport observers with their IDs, which gets them added to routing tables, node databases, etc.

jtolds:
the storagenode/trust test is failing. it took me a long time to figure out (i just figured this out over the weekend!) that with our jenkins setup, the row that starts of "go test -vet=off" which is marked red is pretty useless, but the row *after it* that's green actually has all the test failures color coded. weird?
i'll fix it

egonelbre:
@jtolds in the upper right corner there is also "Tests" which shows each test individually.

PR #2256, opened 2019-06-19
scripts/tag-release.sh: libuplink release tagging

jtolds:
a couple of arch review meetings ago we discussed how to make sure that it is much easier to get the release defaults for binaries, libraries, and so on. we already imperfectly solved the binary problem with the release.sh script, but (until now!) have not solved the problem of getting release defaults for people building from source.

the solution we seemed to all prefer was to make sure our tagged version commits check the release state into the source code.

this script aides in tagging commits with version tags and updating the source defaults. it still plays nicely with release.sh and our other build processes.

after this is merged we should configure github/go modules to prefer people use one of our tags instead of master (which will keep dev defaults). UPDATE: go modules does indeed already do the right thing.

Please describe the tests: no tests
 
Please describe the performance impact: no runtime performance impact


scripts/tag-release.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
egonelbre: Does this have to be a bash script :(

egonelbre:

egonelbre:
Do we have documentation on our release process?

jtolds:
I dunno, @littleskunk do we have anything?

I can add a comment about how this should be used in place of `git tag` somewhere I suppose

littleskunk:
No we haven't documented it.

PR #2260, opened 2019-06-20
Uplink C bindings part 4

bryanchriswhite:
What: 

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

lib/uplinkc/convert.go
```diff
@@ -14,30 +14,54 @@ func newBucketInfo(bucket *storj.Bucket) C.BucketInfo {
...
 }
+
+// bytes_to_cbytes converts a byte array to a C uint8_t array
+func bytes_to_cbytes(bytes []byte) (data *C.uint8_t, _ C.uint64_t) {
+	length := len(bytes)
+	ptr := C.malloc(C.size_t(length))
+	mem := (*[1 << 30]uint8)(ptr)
+	copy((*mem)[:], bytes)
+
+	return data, C.uint64_t(length)
```
kaloyan-raev: If I am not wrong `data` is nil at this point. Shouldn't we return `mem ` instead?

bryanchriswhite: good catch!... this explains a todo in the object test :p

lib/uplinkc/object.go
```diff
@@ -65,7 +135,7 @@ func upload_write(uploader C.UploaderRef, bytes *C.uint8_t, length C.int, cErr *
 
 	n, err := upload.wc.Write(buf)
 	if err == io.EOF {
-		return C.EOF
+		return C.int(0)
```
kaloyan-raev: Shouldn't it return `n` instead of `0`? `n` should be the number of bytes successfully written before the error occurred.

kaloyan-raev: Shouldn't we also set `cErr` to `err`, if `err != nil && err != io.EOF`?

lib/uplinkc/object.go
```diff
@@ -134,7 +252,7 @@ func download_read(downloader C.DownloaderRef, bytes *C.uint8_t, length C.int, c
 
 	n, err := download.rc.Read(buf)
 	if err == io.EOF {
-		return C.EOF
+		return C.int(0)
```
kaloyan-raev: Same comments like on `upload_write`.

lib/uplinkc/object.go
bryanchriswhite: yea - just talked with alex about it and we arrived at the same conclusion

bryanchriswhite: good catch

lib/uplinkc/convert.go
```diff
@@ -14,26 +14,40 @@ func newBucketInfo(bucket *storj.Bucket) C.BucketInfo {
 	return C.BucketInfo{
 		name:         C.CString(bucket.Name),
 		created:      C.int64_t(bucket.Created.Unix()),
-		path_cipher:  C.uint8_t(bucket.PathCipher),
+		path_cipher:  C.Cipher(bucket.PathCipher),
```
egonelbre: nit: Ah, this kind of cast worries me, I would rather have a func that does an explicit conversion. It seems would be too easy to cast a `storj.Cipher` to `C.Cipher`.

lib/uplinkc/convert.go
```diff
@@ -14,26 +14,40 @@ func newBucketInfo(bucket *storj.Bucket) C.BucketInfo {
...
 		encryption_parameters: convertEncryptionParameters(&bucket.EncryptionParameters),
 		redundancy_scheme:     convertRedundancyScheme(&bucket.RedundancyScheme),
 	}
 }
 
+// newObjectInfo returns a C object struct converted from a go object struct.
+func newObjectInfo(object *storj.Object) C.ObjectInfo {
+	return C.ObjectInfo{
+		version:      C.uint32_t(object.Version),
+		bucket:       newBucketInfo(&object.Bucket),
```
egonelbre: Do we need to get the bucket info for every object? If the user wants bucket info it can ask it explicitly.

lib/uplinkc/uplink_definitions.h
```diff
@@ -6,10 +6,30 @@
...
+    STORJ_AESGCM,
+    STORJ_SECRET_BOX,
+    STORJ_INVALID
+} Cipher;
+
+typedef enum CipherSuite {
+    STORJ_ENC_UNSPECIFIED,
+    STORJ_ENC_NULL,
+    STORJ_ENC_AESGCM,
+    STORJ_ENC_SECRET_BOX
```
egonelbre: Let's add specific numbers to these...

```
    STORJ_ENC_UNSPECIFIED = 0,
    STORJ_ENC_NULL        = 1,
    STORJ_ENC_AESGCM      = 2,
    STORJ_ENC_SECRET_BOX  = 3
```

egonelbre: This is to make it clear that it's just not an enum, but rather a enumeration where the values need to be kept as is.

egonelbre: Note to self: we need to do the same on Go side.

lib/uplinkc/uplink_definitions.h
```diff
@@ -70,7 +91,48 @@ typedef struct EncryptionAccess {
     char key[32];
 } EncryptionAccess;
 
+typedef struct ObjectInfo {
+    uint32_t   version;
+    BucketInfo bucket;
```
egonelbre: Same here, can we avoid the bucket info?

lib/uplinkc/uplink_definitions.h
```diff
@@ -70,7 +91,48 @@ typedef struct EncryptionAccess {
...
+    char    *prefix;
+    char    *cursor;
+    char    delimiter;
+    bool    recursive;
+    int8_t  direction;
+    int64_t limit;
+} ListOptions;
+
+typedef struct ObjectMeta {
+    char     *bucket;
```
egonelbre: Do we need the bucket here?

lib/uplinkc/uplink_definitions.h
```diff
@@ -70,7 +91,48 @@ typedef struct EncryptionAccess {
...
+    char     *path;
+    bool     is_prefix;
```
egonelbre: Is prefix doesn't make sense to me on a type that is not part of list.

lib/uplinkc/uplink_definitions.h
```diff
@@ -70,7 +91,48 @@ typedef struct EncryptionAccess {
...
+typedef struct ListOptions {
+    char    *prefix;
+    char    *cursor;
+    char    delimiter;
+    bool    recursive;
+    int8_t  direction;
+    int64_t limit;
+} ListOptions;
+
+typedef struct ObjectMeta {
```
egonelbre: Wondering whether we could get rid of `ObjectMeta` altogether and just use `ObjectInfo`?

jtolds: i don't know if this is related, but i tend to prefer APIs where it's very clear what I can and can't set. I would much prefer keeping return types (that have more computed information) and argument types (strictly the things i can change) separated

lib/uplinkc/convert.go
```diff
@@ -14,30 +14,49 @@ func newBucketInfo(bucket *storj.Bucket) C.BucketInfo {
...
+		modified:     C.int64_t(object.Modified.Unix()),
+		expires:      C.int64_t(object.Expires.Unix()),
+	}
+}
+
 // convertEncryptionParameters converts Go EncryptionParameters to C.
 func convertEncryptionParameters(goParams *storj.EncryptionParameters) C.EncryptionParameters {
 	return C.EncryptionParameters{
-		cipher_suite: C.uint8_t(goParams.CipherSuite),
+		cipher_suite: C.CipherSuite(goParams.CipherSuite),
```
egonelbre: use a func here as well, so I can sleep nicer today :)

lib/uplinkc/uplink_definitions.h
egonelbre: Yeah, I agree... but I'm trying to figure out the difference.

egonelbre: I would expect the types be more like:

```
typedef struct ObjectList {
    char       *bucket;
    char       *prefix;
    bool       more;
    ObjectListItem *items;
    int32_t    length;
} ObjectList;

typedef struct ObjectListItem {
    bool       is_prefix;

    uint32_t   version;
    char       *path;
    char       *content_type;

    int64_t    created;
    int64_t    modified;
    int64_t    expires;
} ObjectListItem;

typedef struct ObjectInfo {
    uint32_t   version;
    char       *path;
    char       *content_type;

    int64_t    created;
    int64_t    modified;
    int64_t    expires;
    uint64_t   size;

    uint8_t    *checksum_bytes;
    uint64_t   checksum_length;
} ObjectInfo;
```

lib/uplinkc/uplink_definitions.h
bryanchriswhite: :+1:

lib/uplinkc/uplink_definitions.h
egonelbre: Either way it seems that something can be improved in `lib/uplink` as well.

lib/uplinkc/convert.go
aleitner: is this a nitpick?

bryanchriswhite: I'm not sure what you're asking for here. The go type is `CipherSuite`. This is the correct cast from the go to the C respective types, we're not converting between cipher and cipher suite here and there's no way to know the difference.

lib/uplinkc/uplink_definitions.h
bryanchriswhite: I don't disagree with you guys on this but we were just trying to mirror the go api and this is what the go api looks like. Can we create a separate issue to change libuplink and address this design change?

lib/uplinkc/uplink_definitions.h
bryanchriswhite: this is mirroring the go api (see comment below)

lib/uplinkc/uplink_definitions.h
aleitner: leave as-is

lib/uplinkc/uplink_definitions.h
egonelbre: resolution: looks like there are quite a few bits that need improvements on `lib/uplink` side as well

PR #2262, opened 2019-06-20
reject invalid orders

littleskunk:
What: Reject invalid orders to make sure we accept the valid orders at the end of the queue.

Why: Mars and friends got restored from an old backup. At the moment storage nodes are unable to submit any orders because we error out instead of rejecting the invalid orders. (they are valid but we lost the data to confirm that)
 

mniewrzal:
so from what I understand we will now reject orders also when we are not able to find `uplink` public key?

stefanbenten:
> so from what I understand we will now reject orders also when we are not able to find `uplink` public key?

Yeah, thats correct. This is the only way to let our SN submit their orders for the faulty DBs of Mars and Jupiter.

PR #2265, opened 2019-06-20
[v3-1952 test 1 & 3] pkg/audit: Add DQ test for too many failed audits

ifraixedes:
What:

Add an integration test which checks that a node which fails to many
audits get disqualified but not before it reaches the audit reputation
disqualification cut-off.

This should complete [v3-1952](https://storjlabs.atlassian.net/browse/V3-1952) _high risk test 1 and 3_.

Why:

We want to have several integrations tests for the disqualification feature.

Please describe the tests:
	It has been added a test case and it does what basically is described in the `what` block but also in the same test https://github.com/storj/storj/pull/2265/files#diff-aa4bcffb7ce996ca2c5aa3d24268c191R20
 
Because we are adding a new test and it uses `testplanet` the CI time will increase however as an usual expectation. 


ifraixedes:
There is one [`TODO`](https://github.com/storj/storj/pull/2265/files#diff-aa4bcffb7ce996ca2c5aa3d24268c191R36) in the new test to mark the part that will go away once https://github.com/storj/storj/pull/2271 be merged.\

This PR is ready for a review by it's blocked until https://github.com/storj/storj/pull/2271 be merged.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,135 @@
...
+			5*time.Second,
+		)
+
+		report, err := verifier.Verify(ctx, stripe, nil)
+		require.NoError(t, err)
+		require.Len(t, report.Offlines, 0)
+		require.Len(t, report.PendingAudits, 0)
+		require.Len(t, report.Fails, 1)
+		require.Equal(t, nodeID, report.Fails[0])
+
```
kaloyan-raev: We have enough tests that the audit report will contain a failed audit if a storage node is missing the piece. So we can simplify this test by getting rid of everything above this comment.

For the rest of the test, we can manually create the report like:
```go
report := &audit.Report{
	Fails: storj.NodeIDList{nodeID},
}
```
where `nodeID := planet.StorageNodes[0].ID()`.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,135 @@
...
+		dossier, err := sat.Overlay.Service.Get(ctx, nodeID)
+		require.NoError(t, err)
+
+		var prevReputation float64
+		{
+			alpha := dossier.Reputation.AuditReputationAlpha
+			beta := dossier.Reputation.AuditReputationBeta
+			prevReputation = alpha / (alpha + beta)
+		}
```
kaloyan-raev: How about extract this as a function? So you can use it both here and on lines 107-112.

pkg/audit/disqualification_test.go
ifraixedes: I gonna change it, however it looks that we could have this test without using testplanet and it should run faster, currently it takes 1 second something in my machine and my guess is because of using testplanet.

pkg/audit/disqualification_test.go
ifraixedes: It's very minimal operation but I'm ok with that.

Initially I wanted to use the function that we use because if we change the way to calculate the reputation we won't need to change this test, but the function isn't exported.

ifraixedes: I've understood better now what you said, passing dossier as a parameter to that function

pkg/audit/disqualification_test.go
ifraixedes: I could remove quite a few things more just creating the `report` as you commented, and now the test execution has been reduced to half the time than before.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,94 @@
...
+// * Create a failed audit report for a storagenode
+// * Record the audit report several times and check that the node isn't
+//	 disqualified until the audit reputation reaches the cut-off value.
+func TestDisqualificationTooManyFailedAudits(t *testing.T) {
+	var auditDQCutOff float64
+
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, Reconfigure: testplanet.Reconfigure{
+			Satellite: func(log *zap.Logger, index int, config *satellite.Config) {
+				auditDQCutOff = config.Overlay.Node.AuditReputationDQ
```
kaloyan-raev: You don't really need this to be in the Reconfigure. You can get this value in the test body too.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+		prevReputation := calculateStorageNodeReputation(dossier)
+
+		// Report the audit failure until the node gets disqualified due to many
+		// failed audits
+		for n := 0; ; n++ {
+			_, err := sat.Audit.Service.Reporter.RecordAudits(ctx, report)
+			require.NoError(t, err)
+
+			dossier, err := sat.Overlay.Service.Get(ctx, nodeID)
```
kaloyan-raev: This can also go inside the `calculateStorageNodeReputation` (how about just `calcReputation` as function name) function, which can take `planet` and `nodeID` as arguments, and return reputation and error.

ifraixedes: I prefer `calcReputation` too, however I did the long name because I'm not understanding when to use abbreviations and when not and when to use more explicit and less explicit names, etc.

I'll do those changes.

ifraixedes: Well, no!

I need the dossier in the loop so doing so it will require to get the dossier  twice in the loop.

kaloyan-raev: Oh, yes. You are right.

kaloyan-raev: Regarding the names, I think the rule of thumb is that we use longer names if a shorter one would be ambiguous.

pkg/audit/disqualification_test.go
ifraixedes: That's what I wanted but I didn't find how.

Could you tell me how to access to it please?

pkg/audit/disqualification_test.go
ifraixedes: yeah! but ambiguity and abbreviations (except fa few very specific) are quite subjective so it's difficult to see when yes and no unless that you see some that they are applied.

pkg/audit/disqualification_test.go
kaloyan-raev: Oh, I assumed wrong! I thought it is possible to access peer config in the test body, but this is not possible. Perhaps, this could be a nice future improvement for testplanet.

ifraixedes: I thought the same, I'll open a ticket for it

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestDisqualificationTooManyFailedAudits(t *testing.T) {
+	var auditDQCutOff float64
+
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, Reconfigure: testplanet.Reconfigure{
+			Satellite: func(log *zap.Logger, index int, config *satellite.Config) {
+				auditDQCutOff = config.Overlay.Node.AuditReputationDQ
+				// TODO: if/v3-1952-h1 this setting will go away once
+				// https://github.com/storj/storj/pull/2271 be merge
+				config.Audit.MaxRetriesStatDB = 1
```
mobyvb: PR has been merged

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,94 @@
...
+				)
+
+				require.True(t, time.Now().Sub(*dossier.Disqualified) >= 0,
+					"Disqualified should be in the past",
+				)
+
+				break
+			}
+
+			require.Nil(t, dossier.Disqualified, "Disqualified")
```
mobyvb: we don't guarantee that this branch of the test gets hit if the node is disqualified after the first loop. Maybe we should pick dq values that ensure we iterate at least twice?

pkg/audit/disqualification_test.go
ifraixedes: I was waiting for it!

internal/testplanet/satellite.go
```diff
@@ -115,12 +115,14 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 					AuditReputationBeta0:         0,
 					AuditReputationLambda:        1,
 					AuditReputationWeight:        1,
+					AuditReputationDQ:            0.5,
```
navillasa: I think the default values in the non-test config is 0.6. Is there a reason we're using 0.5 here?

ifraixedes: I used the defaults that was set yesterday, but those changed later IRCC

I gonna update them to match the current defaults

pkg/audit/disqualification_test.go
ifraixedes: I added an assertion for ensuring whatever values are used, there will be at least 2 iterations.

internal/testplanet/satellite.go
ifraixedes: I updated these values to match the current defaults in `pkg/overlay.Config`

ifraixedes:
I had to set custom values because the default ones only made one iteration.
I also added 3 new assertions:

* Assertion to ensure that a a node which haven't got any audit, its dossier has `alpha` equal to `alpha0`
* Assertion to ensure that a a node which haven't got any audit, its dossier has `beta` equal to `beta0`
* Assertion that it has been more than 1 iteration

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,105 @@
...
+		require.NoError(t, err)
+
+		require.Equal(t, alpha0, dossier.Reputation.AuditReputationAlpha)
+		require.Equal(t, beta0, dossier.Reputation.AuditReputationBeta)
+
+		prevReputation := calcReputation(dossier)
+
+		// Report the audit failure until the node gets disqualified due to many
+		// failed audits
+		iterations := 1
```
mobyvb: I think this should start at 0. After the loop, you check if `iterations > 1`, which will happen after a single iteration

ifraixedes: No, `iteration++` is executed after the loop body finishes (without breaking/returning)

mobyvb: Oh, did not realize

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,105 @@
...
+	var (
+		auditDQCutOff float64 = 0.4
+		alpha0        float64 = 1
+		beta0         float64 = 0
+	)
+
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, Reconfigure: testplanet.Reconfigure{
+			Satellite: func(log *zap.Logger, index int, config *satellite.Config) {
+				config.Overlay.Node.AuditReputationAlpha0 = alpha0
```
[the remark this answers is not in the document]
ifraixedes: I want to guarantee the 2 iteration, we could use only some of them but this test will be susceptible to testplanet changes.

If we want to check the default ones, then the test should be changed to work fine with them without any modification, but halfway doesn't provide any benefit.

pkg/audit/disqualification_test.go
ifraixedes: https://storjlabs.atlassian.net/browse/V3-2011

PR #2268, opened 2019-06-20
satellite: rearrange marketing package

VinozzZ:
What: 
- Remove `marketing package`
- Create `Offers` package

Why:

- We need to expose offer table to the `console` package and `marketing` package. It's not idea to tie `marketing` package with `console` package. 
- A satellite operator should be able to choose which marketing campaign they would like to run in stead of only have a high-level marketing package
- The `value attribution` has its own package for open source program so `marketing` package should not exists

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/offers/offers.go
```diff
@@ -0,0 +1,189 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package offers
```
egonelbre: Is there a better name than `offers`?

e.g. `kickback`, `rewards`...?

VinozzZ: I do like `rewards`!

egonelbre: I suspect there will eventually be some other logic related to `rewards`...

with regards to package naming:
* https://golang.org/doc/effective_go.html#package-names
* https://rakyll.org/style-packages/
* https://github.com/golang/go/wiki/CodeReviewComments#package-names
* https://blog.golang.org/package-names

`rewardsconfig` seems overly specific. The package responsibility is to organize logic about `rewards`.

satellite/satellitedb/offers.go
```diff
@@ -10,26 +10,26 @@ import (
...
 
-// ListAll returns all offers from the db
-func (offers *offers) ListAll(ctx context.Context) ([]marketing.Offer, error) {
-	offersDbx, err := offers.db.All_Offer(ctx)
+// ListAll returns all offersDB from the db
+func (db *offersDB) ListAll(ctx context.Context) ([]offers.Offer, error) {
+	offersDbx, err := db.db.All_Offer(ctx)
 	if err != nil {
-		return nil, marketing.OffersErr.Wrap(err)
+		return nil, offers.Err.Wrap(err)
```
egonelbre: Note the wrapping error should be from this package not the imported one.

satellite/offers/offers.go
VinozzZ: That sounds reasonable. I changed it

VinozzZ:
The only reason I don’t want to change the struct type name right now is
that I want to keep it consistent with the table name which is offers.
We also use the term free credit offer. Either way, I think we can make
these changes for now and if we do see a big confusion or advantage of
changing all naming to rewards. We can do it in another PR

internal/testplanet/satellite.go
```diff
@@ -26,7 +26,7 @@ import (
 	"storj.io/storj/satellite/console"
 	"storj.io/storj/satellite/console/consoleweb"
 	"storj.io/storj/satellite/mailservice"
-	"storj.io/storj/satellite/marketing/marketingweb"
```
coyle: curious, why the change ?

satellite/peer.go
```diff
@@ -605,6 +605,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 		peer.Marketing.Endpoint, err = marketingweb.NewServer(
 			peer.Log.Named("marketing:endpoint"),
 			marketingConfig,
+			// peer.DB.Offers()
```
coyle: should this still be here ?

satellite/satellitedb/offers.go
```diff
@@ -10,26 +10,31 @@ import (
...
 	"github.com/zeebo/errs"
 
-	"storj.io/storj/satellite/marketing"
+	"storj.io/storj/satellite/rewards"
 	dbx "storj.io/storj/satellite/satellitedb/dbx"
 )
 
-type offers struct {
+var (
+	// Error the default offers errs class
```
coyle: should be in the form of `offerErr is the default offer errors class`

internal/testplanet/satellite.go
VinozzZ: Yea, from the conversation we had last week about the package structure, the service layer is not needed since both `console` and `marketingweb` server needs access to the `offers` table. It feels weird to put `offers` under marketing only.

satellite/peer.go
VinozzZ: Yes, I will remove this

PR #2271, opened 2019-06-20
pkg/audit: fix maxRetries bug

navillasa:
What:
- changes loop condition from `retries < reporter.maxRetries` to `retries <= reporter.maxRetries`

Why:
- because maxRetries had a default set to 0 in testplanet which meant that an initial "try" wasn't happening at all

Please describe the tests:
 Test 1: TestRecordAuditsAtLeastOnce
- sets maxRetries to 0, but then expects that the audit report was recorded
- idk this is very minor and basically only tests testplanet. maybe we can take out the test. yolo.
 
- now we know


pkg/audit/reporter_test.go
```diff
@@ -53,3 +53,30 @@ func TestReportPendingAudits(t *testing.T) {
...
+		reporter := audit.NewReporter(zap.L(), overlay, containment, 0, 3)
+
+		// expect RecordAudits to try recording at least once
+		failed, err := reporter.RecordAudits(ctx, &report)
+		require.NoError(t, err)
+		assert.Zero(t, failed)
+
+		node, err := overlay.Get(ctx, nodeID)
+		require.NoError(t, err)
+		assert.Equal(t, int64(1), node.Reputation.AuditCount)
```
mobyvb: you can use `EqualValues` and you don't have to cast

mobyvb: also use require

pkg/audit/reporter_test.go
```diff
@@ -53,3 +53,30 @@ func TestReportPendingAudits(t *testing.T) {
...
+		overlay := planet.Satellites[0].Overlay.Service
+		containment := planet.Satellites[0].DB.Containment()
+
+		// set maxRetries to 0
+		reporter := audit.NewReporter(zap.L(), overlay, containment, 0, 3)
+
+		// expect RecordAudits to try recording at least once
+		failed, err := reporter.RecordAudits(ctx, &report)
+		require.NoError(t, err)
+		assert.Zero(t, failed)
```
mobyvb: require

pkg/audit/reporter_test.go
navillasa: cool, I didn't realize EqualValues does that

pkg/audit/reporter.go
```diff
@@ -66,7 +66,7 @@ func (reporter *Reporter) RecordAudits(ctx context.Context, req *Report) (failed
 	var errlist errs.Group
 
 	retries := 0
-	for retries < reporter.maxRetries {
+	for retries <= reporter.maxRetries {
 		if len(successes) == 0 && len(fails) == 0 && len(offlines) == 0 && len(pendingAudits) == 0 {
```
Fadila82: Maybe it would be clearer by renaming "retries" to "tries", because as it is, when I will have forgotten this PR and if I come across this part of code, I will think there's a mistake.

navillasa: okay I will!

PR #2272, opened 2019-06-20
[v3-1952 tests 4 & 5] DQ nodes should not be used for download or upload 

Fadila82:
What: 
Add integration tests to check a disqualified node is not used for download or upload
This should complete [v3-1952](https://storjlabs.atlassian.net/browse/V3-1952) _high risk test 4 and 5_.

Why:
We want to have several integrations tests for the disqualification feature.


pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+	// - uploads random data
+	// - mark a node as disqualified
+	// - check we don't get it when we require order limit
+
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		satellite := planet.Satellites[0]
+		disqualifiedNode := planet.StorageNodes[0]
+		uplinkNode := planet.Uplinks[0]
```
kaloyan-raev: `uplinkNode` is confusing. We usually call "nodes" only the storage nodes. You can use `uplink`, or `upl` if it is not possible to use `uplink` due to conflicts.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+		err := satellite.Audit.Service.Close()
+		require.NoError(t, err)
+
+		testData := make([]byte, 8*memory.KiB)
+		_, err = rand.Read(testData)
+		require.NoError(t, err)
+
+		err = uplinkNode.UploadWithConfig(ctx, satellite, &uplink.RSConfig{
```
kaloyan-raev: I guess you are uploading with custom RS config to be sure that `planet.StorageNodes[0]` gets a piece.

This is not really required. Below you get the pointer of the uploaded file. You can use `pointer.GetRemote().GetRemotePieces()[0].NodeId` to get a storage node that has a piece for sure. Then you can disqualify this storage node.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+		require.NoError(t, err)
+
+		projects, err := satellite.DB.Console().Projects().GetAll(ctx)
+		require.NoError(t, err)
+
+		bucketID := []byte(storj.JoinPaths(projects[0].ID.String(), "testbucket"))
+
+		disqualifyNode(t, ctx, satellite, disqualifiedNode.ID())
+
+		listResponse, _, err := satellite.Metainfo.Service.List(ctx, "", "", "", true, 0, 0)
```
kaloyan-raev: It is simpler to use `satellite.Metainfo.Service.Get` to get the pointer. You already know the bucket and path.

kaloyan-raev:
Please remove go.mod and go.sum changes from this PR.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+			require.NoError(t, err)
+			if pointer.GetType() == pb.Pointer_REMOTE {
+				break
+			}
+		}
+		limits, err := satellite.Orders.Service.CreateGetOrderLimits(ctx, uplinkNode.Identity.PeerIdentity(), bucketID, pointer)
+		require.NoError(t, err)
+
+		for _, orderLimit := range limits {
+			require.NotEqual(t, orderLimit.StorageNodeAddress.Address, disqualifiedNode.Addr())
```
kaloyan-raev: It would be better to check node IDs instead of addresses.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+		AuditSuccess: false,
+		AuditLambda:  0,
+		AuditWeight:  1,
+		AuditDQ:      0.5,
+		UptimeLambda: 1,
+		UptimeWeight: 1,
+		UptimeDQ:     0.5,
+	})
+	require.NoError(t, err)
+	reputable, err := satellite.Overlay.Service.IsVetted(ctx, nodeID)
```
kaloyan-raev: Why do we check this?

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+		for _, orderLimit := range limits {
+			require.NotEqual(t, orderLimit.StorageNodeAddress.Address, disqualifiedNode.Addr())
+		}
+
+	})
+}
+
+func TestDisqualifiedNodesGetNoUpload(t *testing.T) {
+
+	// - uploads random data
```
kaloyan-raev: We don't really do this in this test.

pkg/audit/disqualification_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+		expirationDate, err := ptypes.Timestamp(pointer.ExpirationDate)
+		require.NoError(t, err)
+
+		apiKey := uplinkNode.APIKey[satellite.ID()]
+
+		metainfo, err := uplinkNode.DialMetainfo(ctx, satellite, apiKey)
+		require.NoError(t, err)
+
+		_, _, err = metainfo.CreateSegment(ctx, "testbucket", "file/path", -1, pointer.Remote.Redundancy, memory.MiB.Int64(), expirationDate)
```
kaloyan-raev: It would be simpler if you call `planet.Satellites[0].Overlay.Service.FindStorageNodes`. This is what `metainfo.CreateSegment` does too.

Also, you can call `FindStorageNodes` 10 times in a loop, so we are surer that node selection works properly.

mobyvb: If you call `FindStorageNodes` and request 4 nodes, you should get `overlay.ErrNotEnoughNodes` back as the error, but it will still return a list of nodes even though it errors. That way, you can request 4, expect an error, expect that the node list returned is not empty, and expect that none of the nodes in the list are disqualified.

kaloyan-raev: Oh, this is a great idea @mobyvb. This way only one call would be enough. No need for that 10 calls in a loop idea.

pkg/audit/disqualification_test.go
navillasa: I think to make sure the disqualification worked? with `require.False(t, reputable)`?

Fadila82: That was the aim yes

pkg/audit/disqualification_test.go
Fadila82: sorry, badly modified copy/paste

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +109,136 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+			FreeDisk:             0,
+			ExcludedNodes:        nil,
+			MinimumVersion:       "", // semver or empty
+		}
+		nodes, err := satellite.Overlay.Service.FindStorageNodes(ctx, request)
+		require.True(t, overlay.ErrNotEnoughNodes.Has(err))
+
+		require.NotEmpty(t, nodes)
+		for _, node := range nodes {
+			reputable, err := satellite.Overlay.Service.IsVetted(ctx, node.Id)
```
mobyvb: alternatively or in addition, you could check `require.NotEqual(t, node.Id, disqualifiedNode)`

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +109,136 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+
+		request := overlay.FindStorageNodesRequest{
+			MinimumRequiredNodes: 4,
+			RequestedCount:       0,
+			FreeBandwidth:        0,
+			FreeDisk:             0,
+			ExcludedNodes:        nil,
+			MinimumVersion:       "", // semver or empty
+		}
+		nodes, err := satellite.Overlay.Service.FindStorageNodes(ctx, request)
```
mobyvb: it looks like your tests got mixed up. This is the Download after dq test, so you should be testing that `limits, err := satellite.Orders.Service.CreateGetOrderLimits(ctx, uplinkNode.Identity.PeerIdentity(), bucketID, pointer)` does not include the dqed node (you removed that in this commit: https://github.com/storj/storj/pull/2272/commits/a1106b943de20465aa38b0880944f17693829f0a )
In the test below (`TestDisqualifiedNodesGetNoUpload`), you should be doing the `FindStorageNodes` step. And you should be able to remove all the stuff after `disqualifyNode(t, ctx, satellite, disqualifiedNode.ID())` in that test

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +109,136 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+		satellite := planet.Satellites[0]
+		disqualifiedNode := planet.StorageNodes[0]
+		upl := planet.Uplinks[0]
+
+		err := satellite.Audit.Service.Close()
+		require.NoError(t, err)
+
+		disqualifyNode(t, ctx, satellite, disqualifiedNode.ID())
+
+		rs := &pb.RedundancyScheme{
```
mobyvb: for this test, you can remove from here to the bottom and replace it with the stuff including and after `nodes, err := satellite.Overlay.Service.FindStorageNodes(ctx, request)` in the test above

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +107,111 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+			RequestedCount:       0,
+			FreeBandwidth:        0,
+			FreeDisk:             0,
+			ExcludedNodes:        nil,
+			MinimumVersion:       "", // semver or empty
+		}
+		nodes, err := satellite.Overlay.Service.FindStorageNodes(ctx, request)
+		require.True(t, overlay.ErrNotEnoughNodes.Has(err))
+
+		require.NotEmpty(t, nodes)
```
kaloyan-raev: We require exactly 3 nodes, right? Let's assert for this instead of not empty, i.e. `assert.Len(t, nodes, 3)`.

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +107,111 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+		for _, node := range nodes {
+			reputable, err := satellite.Overlay.Service.IsVetted(ctx, node.Id)
+			require.NoError(t, err)
+			require.True(t, reputable)
+			require.NotEqual(t, node.Id, disqualifiedNode)
```
kaloyan-raev: Let's use `assert` instead of `require` for the above two checks.

We usually use `require` when there is no sense to continue the check if the assertion does not pass, e.g. when checking errors to avoid panics on the next checks. Here, if the test fails, we may have additional info from checking all nodes.

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +107,111 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+		lastSegPath := storj.JoinPaths(projects[0].ID.String(), "l", encryptedAfterBucket)
+		pointer, err := satellite.Metainfo.Service.Get(ctx, lastSegPath)
+		require.NoError(t, err)
+
+		disqualifiedNode := pointer.GetRemote().GetRemotePieces()[0].NodeId
+		disqualifyNode(t, ctx, satellite, disqualifiedNode)
+
+		limits, err := satellite.Orders.Service.CreateGetOrderLimits(ctx, upl.Identity.PeerIdentity(), bucketID, pointer)
+		require.NoError(t, err)
+
```
kaloyan-raev: let's check that `limits` is not empty, or even better -- has the expected number of items, i.e. `len(pointer.GetRemote().GetRemotePieces()) - 1`

pkg/audit/disqualification_test.go
kaloyan-raev: `IsVetted` checks not only if a node has been disqualified, but also if it has at least a number of audit and uptime checks about a configured threshold.

A much cleaner check if a node is disqualified would be:

```go
node, err := satellite.Overlay.Service.Get(ctx, nodeID)
require.NoError(t, err)
assert.NotNil(t, node.Disqualified)
```

pkg/audit/disqualification_test.go
```diff
@@ -103,3 +108,114 @@ func calcReputation(dossier *overlay.NodeDossier) float64 {
...
+		disqualifyNode(t, ctx, satellite, disqualifiedNode)
+
+		limits, err := satellite.Orders.Service.CreateGetOrderLimits(ctx, upl.Identity.PeerIdentity(), bucketID, pointer)
+		require.NoError(t, err)
+		assert.Len(t, limits, len(pointer.GetRemote().GetRemotePieces())-1)
+
+		for _, orderLimit := range limits {
+			reputable, err := satellite.Overlay.Service.IsVetted(ctx, orderLimit.Limit.StorageNodeId)
+			assert.NoError(t, err)
+			assert.True(t, reputable)
```
kaloyan-raev: Why are we checking `IsVetted` for these nodes?

Fadila82: Sorry, I forgot to change it

PR #2278, opened 2019-06-21
satellite: send external address in order limits

jtolds:
What/why: in the ongoing effort to remove satellites from kademlia (and only have them be observers instead of participants), we need another way for storage nodes to be able to find the public key of a satellite. this change sends the satellite's public address in orders, so that storage nodes can dial the storage node directly without kademlia to get the satellite's public key over the tls handshake.

Please describe the tests: just adding a field, integration tests come when the storage node starts using these fields
 
Please describe the performance impact: a couple bytes more data to send over the wire and to hash prior to signing 


egonelbre:
Since storage nodes have to pre-approve satellites that they talk to and they also need to keep information about the satellite updated anyways. I don't think there should be a need to add the satellite address in the order limit. Or am I missing something?

egonelbre:

PR #2281, opened 2019-06-21
Add node stats GRPC endpoint to the satellite

rikysya:
What:  add node stats GRPC endpoint to the satellite

Why: to retrieve additional info for the `SNO`
Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/nodestats.proto
```diff
@@ -0,0 +1,21 @@
...
+import "google/protobuf/timestamp.proto";
+
+service NodeStats {
+    rpc UptimeCheck(UptimeRequest) returns (UptimeResponse) {}
+}
+
+message UptimeRequest { }
+
+message UptimeResponse {
+    double ratio = 1;
```
egonelbre: What's ratio here?

satellite/nodestats/endpoint.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil, NodeStatsEndpointErr.Wrap(err)
+	}
+
+	node, err := e.overlay.Get(ctx, peer.ID)
+	if err != nil {
+		return nil, NodeStatsEndpointErr.Wrap(err)
+	}
+
+	return &pb.UptimeResponse{
+		Ratio: node.Reputation.UptimeRatio,
```
mobyvb: We are no longer updating `UptimeRatio` and will likely remove it in the near future. If you still want to use it, I would recommend calculating it yourself in this function with `node.Reputation.UptimeSuccessCount / node.Reputation.UptimeCount`. Otherwise it will be inaccurate.

If you want to send the reputation value we are actually using to determine whether a node is good or bad, you can return `reputation = node.Reputation.UptimeReputationAlpha / (node.Reputation.UptimeReputationAlpha + node.Reputation.UptimeReputationBeta)`, which will be a number bounded between 0 and 1, where closer to 1 is a better reputation and closer to 0 is a worse reputation.

rikysya: thx for clarification!

PR #2291, opened 2019-06-21
Don't require encryption keys for project or bucket management

jenlij:
What: 
Cleaner & rebased version of https://github.com/storj/storj/pull/1854
We don't want to require encryption keys for project or bucket management.

Why: 
Security & Privacy. We don't want to users to send their encryption keys to the satellite.

Please describe the tests:
Tests were just updated to compile. No new test cases were written.
 
Please describe the performance impact: No obvious performance improvements here. More UX to not require users to provide encryption parameters to the satellite GUI to manage buckets.

Note: https://storjlabs.atlassian.net/browse/V3-2015 will take care of removing kvmetainfo/tempfile


lib/uplink/uplink.go
```diff
@@ -157,29 +150,18 @@ func (u *Uplink) OpenProject(ctx context.Context, satelliteAddr string, apiKey A
...
-		memory.KiB.Int(), storj.AESGCM, maxBucketMetaSize.Int(),
-	)
-	if err != nil {
-		return nil, Error.New("failed to create stream store: %v", err)
-	}
+
+	// volatile warning: we're setting an encryption key of all zeros when one isn't provided.
+	// TODO: fix before the final alpha network wipe
+	encryptionKey := new(storj.Key)
+	streams, err := streams.NewStreamStore(segments, maxBucketMetaSize.Int64(), encryptionKey, memory.KiB.Int(), storj.AESGCM, maxBucketMetaSize.Int())
```
zeebo: missing error check?

pkg/miniogw/gateway_test.go
```diff
@@ -710,8 +710,9 @@ func initEnv(ctx context.Context, planet *testplanet.Planet) (minio.ObjectLayer,
 	if err != nil {
 		return nil, nil, nil, err
 	}
-
-	kvmetainfo := kvmetainfo.New(metainfo, streams, segments, encKey, int32(blockSize), rs, 64*memory.MiB.Int64())
+	//TODOOOOO
+	p := kvmetainfo.NewProject(streams, int32(blockSize), rs, 64*memory.MiB.Int64())
```
JessicaGreben: whats up with this todo?

lib/uplinkc/object.go
```diff
@@ -87,6 +87,7 @@ func get_object_meta(cObject C.ObjectRef, cErr **C.char) C.ObjectMeta {
 	}
 }
 
+// Upload TODO needs a comment
```
JessicaGreben: add a comment?

lib/uplinkc/project.go
```diff
@@ -35,16 +32,8 @@ func open_project(uplinkHandle C.UplinkRef, satelliteAddr *C.char, apikeyHandle
...
-	if unsafe.Pointer(cProjectOpts) != nil {
-		opts = &libuplink.ProjectOptions{}
-		opts.Volatile.EncryptionKey = new(storj.Key)
-		for i := range cProjectOpts.key {
-			opts.Volatile.EncryptionKey[i] = byte(cProjectOpts.key[i])
-		}
-	}
-
-	project, err := uplink.OpenProject(scope.ctx, C.GoString(satelliteAddr), apikey, opts)
+	// TODO: add project options argument
```
jtolds: hmm, this makes me think we should either
1) remove the C.ProjectOptions type also (like we did with the mobile bindings)
or 2) leave the ProjectOptions struct entirely (across the board) and only remove the EncryptionKey field, leaving it a weird, empty struct (which is still good for forwards compatibility? it's an easy place to add things in the future)

jenlij: Oh I forgot I added that TODO... Hmm I think I actually want to remove the C.ProjectOptions type entirely.

pkg/metainfo/kvmetainfo/temputils.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, Error.New("failed to create erasure coding client: %v", err)
+	}
+	rs, err := eestream.NewRedundancyStrategy(eestream.NewRSScheme(fc, whoCares), whoCares, whoCares)
+	if err != nil {
+		return nil, Error.New("failed to create redundancy strategy: %v", err)
+	}
+	maxBucketMetaSize := 10 * memory.MiB
+	segment := segments.NewSegmentStore(m, nil, rs, maxBucketMetaSize.Int(), maxBucketMetaSize.Int64())
+
+	// volatile warning: we're setting an encryption key of all zeros when one isn't provided.
```
JessicaGreben: so does this mean that the satellite will have to decrypt bucket metadata with a key of all zeros to display bucket info in the web UI?

jenlij: All zeros is essentially a nil encryption key, so the bucket metadata isn't supposed to be encrypted

egonelbre:
nice

PR #2294, opened 2019-06-21
add functions for password/root key derivation

zeebo:
This commit adds two functions that implement the algorithms
described in the password key derivation design document. They
will be used during setup to derive bucket level root keys or
default passwords to use when buckets do not have their own
independent key.

The code is not yet used, so there is no testing or performance impact.

### DIFF
--- a/pkg/encryption/password.go
+++ b/pkg/encryption/password.go
@@ -1,0 +1,61 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package encryption
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"runtime"
+
+	"github.com/zeebo/errs"
+	"golang.org/x/crypto/argon2"
+
+	"storj.io/storj/pkg/storj"
+)
+
+func sha256hmac(key, data []byte) ([]byte, error) {
+	h := hmac.New(sha256.New, key)
+	if _, err := h.Write(data); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


pkg/encryption/password.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, err
+	}
+
+	pathSalt, err := sha256hmac(mixedSalt, []byte(path))
+	if err != nil {
+		return nil, err
+	}
+
+	// use a time of 1, 64MB of ram, and all of the cores.
+	keyData := argon2.IDKey(password, pathSalt, 1, 64<<10, uint8(runtime.GOMAXPROCS(-1)), 32)
```
ifraixedes: here, we could use `storj.KeySize`
And perhaps also [`64*memory.MiB`](https://godoc.org/storj.io/storj/internal/memory#Size).

pkg/encryption/password.go
```diff
@@ -0,0 +1,61 @@
...
+// used to derive all of the bucket level secrets. See the password key derivation design
+// doc.
+func DeriveDefaultPassword(password, salt []byte) ([]byte, error) {
+	mixedSalt, err := sha256hmac(password, salt)
+	if err != nil {
+		return nil, err
+	}
+
+	// use a time of 1, 64MB of ram, and all of the cores.
+	return argon2.IDKey(password, mixedSalt, 1, 64<<10, uint8(runtime.GOMAXPROCS(-1)), 32), nil
```
ifraixedes: Same here than above.

PR #2297, opened 2019-06-22
value attribution integration with libuplink

aligeti:
What: 
value attribution integration with libuplink
Why:
Value attribution feature functionality
https://storjlabs.atlassian.net/browse/V3-2012

Please describe the tests:
 - Test 1: **TestPartnerBucketAttrs()**
**WITH VALID PARTNER ID**
--> create a bucket using lib uplink
--> create partner ID using uuid.New()
--> open the bucket
       --> Check valid PartnerID
       --> make partner bucket attribution and set the Bucket's attribution = true can verify this attribution .

**WITH INVALID PARTNER ID**
 --> set the partner ID to empty/not valid
--> create a bucket using lib uplink
--> open the bucket
       --> Check valid PartnerID. --> returns false
       --> make partner bucket attribution and set the Bucket's attribution = false can verify this attribution .
 - Test 2:
 
### DIFF
--- a/lib/uplink/bucket.go
+++ b/lib/uplink/bucket.go
@@ -19,8 +19,9 @@ import (
 // Bucket represents operations you can perform on a bucket
 type Bucket struct {
 	BucketConfig
-	Name    string
-	Created time.Time
+	Name        string
+	Attribution bool
+	Created     time.Time
 
 	bucket   storj.Bucket
 	metainfo *kvmetainfo.DB
--- a/lib/uplink/bucket.go
+++ b/lib/uplink/bucket.go
@@ -19,9 +19,8 @@ import (
 // Bucket represents operations you can perform on a bucket
 type Bucket struct {
 	BucketConfig
-	Name        string
-	Attribution bool
-	Created     time.Time
+	Name    string
+	Created time.Time
 
 	bucket   storj.Bucket
 	metainfo *kvmetainfo.DB
--- a/lib/uplink/bucket_attrs_test.go
+++ b/lib/uplink/bucket_attrs_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/skyrings/skyring-common/tools/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -57,14 +56,6 @@ func testPlanetWithLibUplink(t *testing.T, cfg testConfig,
 func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
 	copy(access.Key[:], encKey)
 	return access
-}
-
-func createPartnerID() (partnerID string, err error) {
-	id, err := uuid.New()
-	if err != nil {
-		return partnerID, err
-	}
-	return id.String(), err
 }
 
 // check that partner bucket attributes are stored and retrieved correctly.
--- a/lib/uplink/bucket_attrs_test.go
+++ b/lib/uplink/bucket_attrs_test.go
@@ -97,7 +97,7 @@ func TestPartnerBucketAttrs(t *testing.T) {
 			require.NoError(t, err)
 
 			_, err = db.Get(ctx, project.ID, []byte(bucketName))
-			require.Error(t, err)
+			require.NoError(t, err)
 
 			// partner ID NOT set
 			proj.uplinkCfg.Volatile.PartnerID = ""
--- a/lib/uplink/bucket_attrs_test.go
+++ b/lib/uplink/bucket_attrs_test.go
@@ -16,6 +16,7 @@ import (
 	"storj.io/storj/internal/memory"
 	"storj.io/storj/internal/testcontext"
 	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/internal/testrand"
 	"storj.io/storj/pkg/storj"
 )
 
@@ -78,8 +79,7 @@ func TestPartnerBucketAttrs(t *testing.T) {
 			_, err := proj.CreateBucket(ctx, bucketName, nil)
 			require.NoError(t, err)
 
-			partnerID, err := createPartnerID()
-			require.NoError(t, err)
+			partnerID := testrand.UUID().String()
 
 			consoleProjects, err := planet.Satellites[0].DB.Console().Projects().GetAll(ctx)
 			assert.NoError(t, err)
--- a/lib/uplink/bucket_attrs_test.go
+++ b/lib/uplink/bucket_attrs_test.go
@@ -69,38 +69,13 @@ func createPartnerID() (partnerID string, err error) {
 // check that partner bucket attributes are stored and retrieved correctly.
 func TestPartnerBucketAttrs(t *testing.T) {
 	var (
-		access          = simpleEncryptionAccess("voxmachina")
-		bucketName      = "mightynein"
-		shareSize       = memory.KiB.Int32()
-		requiredShares  = 2
-		stripeSize      = shareSize * int32(requiredShares)
-		stripesPerBlock = 2
-		inBucketConfig  = BucketConfig{
-			PathCipher: storj.EncSecretBox,
-			EncryptionParameters: storj.EncryptionParameters{
-				CipherSuite: storj.EncAESGCM,
-				BlockSize:   int32(stripesPerBlock) * stripeSize,
-			},
-			Volatile: struct {
-				RedundancyScheme storj.RedundancyScheme
-				SegmentsSize     memory.Size
-			}{
-				RedundancyScheme: storj.RedundancyScheme{
-					Algorithm:      storj.ReedSolomon,
-					ShareSize:      shareSize,
-					RequiredShares: int16(requiredShares),
-					RepairShares:   3,
-					OptimalShares:  4,
-					TotalShares:    5,
-				},
-				SegmentsSize: 688894,
-			},
-		}
+		access     = simpleEncryptionAccess("voxmachina")
+		bucketName = "mightynein"
 	)
 
-	testPlanetWithLibUplink(t, testConfig{}, &access.Key,
+	testPlanetWithLibUplink(t, testConfig{},
 		func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet, proj *Project) {
-			_, err := proj.CreateBucket(ctx, bucketName, &inBucketConfig)
+			_, err := proj.CreateBucket(ctx, bucketName, nil)
 			require.NoError(t, err)
 
 			partnerID, err := createPartnerID()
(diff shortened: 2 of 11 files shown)


lib/uplink/bucket.go
```diff
@@ -19,8 +19,9 @@ import (
 // Bucket represents operations you can perform on a bucket
 type Bucket struct {
 	BucketConfig
-	Name    string
-	Created time.Time
+	Name        string
+	Attribution bool
```
jtolds: should be a string?

lib/uplink/project.go
```diff
@@ -150,6 +152,8 @@ func (p *Project) GetBucketInfo(ctx context.Context, bucket string) (b storj.Buc
 func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *EncryptionAccess) (b *Bucket, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	attribution, _ := p.CheckBucketAttribution(ctx, bucketName)
```
jtolds: don't ignore errors plz

lib/uplink/project.go
```diff
@@ -202,3 +207,20 @@ func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *Enc
 func (p *Project) Close() error {
 	return nil
 }
+
+// CheckBucketAttribution Checks the bucket attribution
+func (p *Project) CheckBucketAttribution(ctx context.Context, bucketName string) (bool, error) {
+	partnerID, err := uuid.Parse(p.uplinkCfg.Volatile.PartnerID)
```
jtolds: if there isn't a partner id set, we should just return out with no error

lib/uplink/project.go
```diff
@@ -202,3 +207,20 @@ func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *Enc
...
+	if err != nil {
+		return false, Error.Wrap(err)
+	}
+
+	err = p.metainfo.SetAttribution(ctx, bucketName, *partnerID)
```
jtolds: maybe we should only do this if the BucketInfo struct doesn't already have an attribution set?

in terms of RPC calls, the common case isn't the case where we need to set attribution, so maybe we say:
 1) get me bucket info
 2) bucket info says the bucket isn't attributed and i have a partner id set, so let's set it in the bucket info.

so that the next time
 1) get me bucket info
 2) bucket info says the bucket is already attributed, nothing to do

lib/uplink/bucket.go
aligeti: Added **Attribution** field  for two purposes (1)to test the  bucket attribution during open bucket operations(). We don't exit if attribution is false, but just log the info and continue.So to test I need a variable to verify. Also (2) thought having a separate field that identifies the bucket's attribution, and save it as part of bucket info, which end user app developer could have some use for it. My understanding was either bucket is attributed or not( boolean state). If I use it as a string, what info I can save in it?

lib/uplink/project.go
aligeti: added the error handling.

lib/uplink/project.go
aligeti: Added the logic to handle the not set partnerID and just log the info and return.

lib/uplink/project.go
aligeti: All the bucket attribution logic is handled at the satellite side, and handles very similarly as suggested.

SetAttribution() on satellite side returns **nil** for:
- when no bucket or just bucket with no file segment exists (no entry exists in value attributionDB, with valid set partner ID )
- when already bucket attributed
SetAttribution() return **err** for:
- when already bucket with file segment exists 
- any value attribution error (like db read and/or Insert error occurs)

SetAttribution() on the lib uplink side, handles the satellite retuned err and nil with relevant log/debug info.

lib/uplink/bucket_attrs_test.go
```diff
@@ -59,6 +60,73 @@ func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
...
+	if err != nil {
+		return partnerID, err
+	}
+	return id.String(), err
+}
+
+// check that partner bucket attributes are stored and retrieved correctly.
+func TestPartnerBucketAttrs(t *testing.T) {
+	var (
+		access          = simpleEncryptionAccess("voxmachina")
```
mniewrzal: you don't have to create such custom bucket config. it doesn't matter for attribution testing.

lib/uplink/project.go
coyle: Is your suggestion to store in the local meta bucket info the attribution ?

lib/uplink/project.go
```diff
@@ -150,6 +152,11 @@ func (p *Project) GetBucketInfo(ctx context.Context, bucket string) (b storj.Buc
 func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *EncryptionAccess) (b *Bucket, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	attribution, err := p.CheckBucketAttribution(ctx, bucketName)
+	if err != nil {
+		zap.S().Warn("Attribution error")
```
coyle: It's not an attribution error, just no attribution was performed.

lib/uplink/bucket.go
coyle: I'm not understanding why you need attribution on the bucket

coyle:
We will also need to update the c bindings

lib/uplink/bucket.go
aligeti: removed it .. upon discussion

lib/uplink/bucket_attrs_test.go
aligeti: updated the test case

lib/uplink/project.go
aligeti: updated the error

lib/uplink/bucket_attrs_test.go
```diff
@@ -57,6 +58,45 @@ func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
...
+			got, err := proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
+
+			got, err = proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
+
+			// partner ID NOT set
+			proj.uplinkCfg.Volatile.PartnerID = ""
+			got, err = proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
```
bryanchriswhite: Same comment as on line 87.

bryanchriswhite:
https://github.com/storj/storj/pull/2326/files

aligeti:
Coordinated with Bryan and he is working on it.

lib/uplink/bucket_attrs_test.go
```diff
@@ -57,6 +58,45 @@ func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
...
+			_, err := proj.CreateBucket(ctx, bucketName, nil)
+			require.NoError(t, err)
+
+			partnerID, err := createPartnerID()
+			require.NoError(t, err)
+
+			// partner ID set
+			proj.uplinkCfg.Volatile.PartnerID = partnerID
+			got, err := proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
coyle: Might be able to just directly query the db both before and after to check status has changed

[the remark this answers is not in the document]
aligeti: yes. I added a new attribute DB access function GetByBucket() and read the DB before and after the attribution.

lib/uplink/bucket_attrs_test.go
```diff
@@ -57,6 +58,45 @@ func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
...
+
+			got, err = proj.OpenBucket(ctx, bucketName, &access)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
aligeti: The second openBucket actually returns from a different part of the code which is when a bucket is already attributed , just returns nil with a debug info logged on the satellite side, which it was from the previous OpenBucket(), but there isn't anyway to test it.  I removed it just to avoid confusion.

lib/uplink/bucket_attrs_test.go
aligeti: This is return with no partenID set. Previous test case was with a valid partner id set.

lib/uplink/project.go
aligeti: Yes that was the intent, but removed it as it was creating confusion.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -15,7 +15,6 @@ delete value_attribution (
 	where value_attribution.project_id = ?
 	where value_attribution.bucket_name = ?
 )
-
```
bryanchriswhite: nit: can this be reverted?

aligeti: updated

lib/uplink/project.go
```diff
@@ -151,6 +152,11 @@ func (p *Project) GetBucketInfo(ctx context.Context, bucket string) (b storj.Buc
 func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *EncryptionAccess) (b *Bucket, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	err = p.CheckBucketAttribution(ctx, bucketName)
+	if err != nil {
+		return nil, err
```
coyle: more a question than comment @egonelbre when should we wrap errors and when should we just return them?

egonelbre: When in doubt Wrap.

There are minor cases where we don't want to hide the underlying error, which is where we skipped the wrapping. With `errs.IsFunc` and `errs.Is` we should be able to fix those situations.

lib/uplink/bucket_attrs_test.go
```diff
@@ -57,6 +58,55 @@ func simpleEncryptionAccess(encKey string) (access EncryptionAccess) {
...
+		access     = simpleEncryptionAccess("voxmachina")
+		bucketName = "mightynein"
+	)
+
+	testPlanetWithLibUplink(t, testConfig{},
+		func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet, proj *Project) {
+			_, err := proj.CreateBucket(ctx, bucketName, nil)
+			require.NoError(t, err)
+
+			partnerID, err := createPartnerID()
```
egonelbre: There is `testrand.UUID().String()` now.

aligeti: updated the code to use testrand

jtolds:
we should be really careful about adding functions to lib/uplink - it's the public interface we're committing to. if there's any chance something should change or it isn't strictly necessary it should not be exported: https://github.com/storj/storj/pull/2352

PR #2299, opened 2019-06-23
lib/uplink: remove NewReader

jtolds:
there's a few reasons for this but the main one is:
a seekable interface simply doesn't map well to the semantics of loading large streams of data over the network. a seekable interface has a surprising amount of subtle challenges when trying to optimize performance. as lib/uplink is our public interface, i'm not eager to give our downstream users any ideas or try to do anything this early that might lead anyone to believe we will efficiently support randomized seeking.

this PR takes a chapter out of the https://godoc.org/gocloud.dev/blob interface and simply does not provide a seekable type. Instead, where gocloud.dev/blob offers a NewRangeReader, we stick with the existing DownloadRange method that does the same thing. perhaps we can do something better down the road but for now i'd rather not increase the API surface area more than we've already exposed.

Change-Id: I65fe739411068a576d8ad787f0f7688c4d649ea3


lib/uplink/bucket.go
```diff
@@ -178,25 +178,6 @@ func (b *Bucket) NewWriter(ctx context.Context, path storj.Path, opts *UploadOpt
...
 
-// ReadSeekCloser combines interfaces io.Reader, io.Seeker, io.Closer
-type ReadSeekCloser interface {
-	io.Reader
-	io.Seeker
-	io.Closer
-}
-
-// NewReader creates a new reader that downloads the object data.
-func (b *Bucket) NewReader(ctx context.Context, path storj.Path) (_ ReadSeekCloser, err error) {
```
egonelbre: return here `io.ReadCloser`

lib/uplinkc/object.go
```diff
@@ -230,14 +229,22 @@ func download(bucketRef C.BucketRef, path *C.char, cErr **C.char) (downloader C.
 
 	scope := bucket.scope.child()
 
-	rc, err := bucket.NewReader(scope.ctx, C.GoString(path))
```
egonelbre: Still use `bucket.NewReader`, there would be a different method `bucket.NewRangeReader`. Currently `DownloadRange` has special arguments and goes through ranger rather than using stream directly.

lib/uplinkc/object.go
```diff
@@ -270,6 +277,7 @@ func download_close(downloader C.DownloaderRef, cErr **C.char) {
 	defer download.cancel()
 
 	err := download.rc.Close()
+	err = errs.Combine(err, download.obj.Close())
```
egonelbre: `err := errs.Combine(download.rc.Close(), download.obj.Close())`

mobile/bucket.go
```diff
@@ -322,23 +324,28 @@ type ReaderOptions struct {
...
+	object    io.Closer
+	reader    io.ReadCloser
 }
 
 // NewReader returns new reader for downloading object
 func (bucket *Bucket) NewReader(path storj.Path, options *ReaderOptions) (*Reader, error) {
 	scope := bucket.scope.child()
 
-	reader, err := bucket.lib.NewReader(scope.ctx, path)
+	object, err := bucket.lib.OpenObject(scope.ctx, path)
```
egonelbre: same here use `NewReader`

egonelbre:
I would rather see `NewRangeReader` method directly rather than go through `DownloadRange`, which uses sentinel values for arguments.

lib/uplinkc/object.go
jtolds: the point of this pr is to remove bucket.NewReader

lib/uplinkc/object.go
jtolds: fair

egonelbre:
```
type Bucket interface {
    Info(ctx context.Context, path storj.Path) (ObjectInfo, error)
    Download(ctx context.Context, path storj.Path) (Object, error)
    DownloadRange(ctx context.Context, path storj.Path, start, limit int64) (Object, error)
    Close() error
}

type Object interface {
    Info() ObjectInfo
    io.ReadSizeCloser
}
```

ifraixedes:
Are the changes in this branch still good for getting merge after applying the review requested changes?

egonelbre:
No, the last comment is what the design should end up being.

PR #2301, opened 2019-06-23
do ip filtering in a more correct way?

thepaul:
This doesn't solve much of the performance difficulty but ought to be
lots more correct in terms of proper node selection semantics.

(this does not yet split unique IPs into a separate table, and does not yet truncate IPv4 addresses to their /24 subnets, as discussed with Stefan. It also does not avoid sequential table scans, and does not provide good strength of randomness.)

What: makes ip filtering work in what is arguably a more correct way.

Why: because we want ip filtering to work in a correct way

Please describe the tests:
 - (existing tests should still pass and cover behavior)
 
- should take maybe 65% of the time of the old implementation

### DIFF
--- a/satellite/satellitedb/overlaycache.go
+++ b/satellite/satellitedb/overlaycache.go
@@ -299,14 +299,28 @@ func (cache *overlaycache) postgresQueryNodesDistinct(ctx context.Context, exclu
 	}
 	args = append(args, count)
 
-	rows, err := cache.db.Query(cache.db.Rebind(`SELECT DISTINCT ON (last_ip) id,
-	type, address, last_ip, free_bandwidth, free_disk, total_audit_count, audit_success_count,
-	total_uptime_count, uptime_success_count, audit_reputation_alpha, audit_reputation_beta, 
-	uptime_reputation_alpha, uptime_reputation_beta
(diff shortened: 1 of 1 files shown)


satellite/satellitedb/overlaycache.go
```diff
@@ -299,14 +299,28 @@ func (cache *overlaycache) postgresQueryNodesDistinct(ctx context.Context, exclu
...
 	args = append(args, count)
 
-	rows, err := cache.db.Query(cache.db.Rebind(`SELECT DISTINCT ON (last_ip) id,
-	type, address, last_ip, free_bandwidth, free_disk, total_audit_count, audit_success_count,
-	total_uptime_count, uptime_success_count, audit_reputation_alpha, audit_reputation_beta, 
-	uptime_reputation_alpha, uptime_reputation_beta
-	FROM (SELECT * FROM nodes
+	rows, err := cache.db.Query(cache.db.Rebind(`
+	WITH candidates AS (
+		SELECT * FROM nodes
```
stefanbenten: I like the Idea of the CTE!

egonelbre:
@thepaul are there tests that should go with it?

PR #2302, opened 2019-06-23
lib/uplink: encryption context

jtolds:
What: this PR exposes an encryption context type that works somewhat like the existing API keys. It can be restricted by path, serialized, parsed, etc. Unlike API keys, encryption contexts can be merged as someone gains access to new folders/buckets.

Blocked on https://github.com/storj/storj/pull/2293

Why: The EncryptionAccess type was underfeatured.

Please describe the tests:
 - Test 1: An example test where one admin gives another admin full access
 - Test 2: An example test where one admin gives a user restricted access
 
Please describe the performance impact: Generating keys from passphrases will be deliberately both CPU intensive and rare. We also need to retrieve a project salt from the server when generating a key from a passphrase.


lib/uplink/encryption.go
```diff
@@ -4,20 +4,179 @@
...
+}
+
+// Import merges the other encryption context into this one. In cases
+// of conflicting path decryption settings (including if both contexts have
+// a default key), the new settings are kept.
+func (s *EncryptionCtx) Import(other *EncryptionCtx) error {
+	if key := other.store.GetDefaultKey(); key != nil {
+		s.store.SetDefaultKey(key)
+	}
+	return other.store.Iterate(s.store.Add)
```
jtolds: neat

lib/uplink/encryption.go
```diff
@@ -4,20 +4,179 @@
...
-type EncryptionAccess struct {
-	// Key is the base encryption key to be used for decrypting objects.
-	Key storj.Key
-	// EncryptedPathPrefix is the (possibly empty) encrypted version of the
-	// path from the top of the storage Bucket to this point. This is
-	// necessary to have in order to derive further encryption keys.
-	EncryptedPathPrefix storj.Path
+// EncryptionCtx represents an encryption context. It holds information about
+// how various buckets and objects should be encrypted and decrypted.
+type EncryptionCtx struct {
```
egonelbre: This should be probably part of `uplink` not `lib/uplink`.

zeebo: Why? I don't know how to make this decision.

lib/uplink/encryption.go
zeebo: cute, right? i thought it was cute.

lib/uplink/encryption.go
egonelbre: The difference is in `lib/uplink` is the higher-level easy-to-use interface, `uplink` is the lower level that allows to have access all the complexity of things. In principle everything should be implementable using only `uplink`, but might be more comfortable with `lib/uplink`.

The `EncryptionCtx` seems more of a lower-level primitive rather than something of higher-level convenience thing.

lib/uplink/encryption.go
jtolds: it's so cute

lib/uplink/encryption.go
jtolds: encryption.Store is the lower level primitive. EncryptionCtx is intended to be a higher level convenience thing

jtolds: overall one goal i have is to make lib/uplink usable almost entirely with types defined in lib/uplink for better discoverability when reading the documentation. specifically, doing paginated listing right now is especially brutal to figure out using the documentation with the types not all in the same package

thepaul: @egonelbre can you clarify what the benefit is to the separation of `uplink` and `lib/uplink`? I confess I don't understand.

egonelbre: `lib/uplink` is the stable simplified interface for interacting with the network.

`uplink` is the complicated inner workings which allow customization at different levels. e.g. you could switch out how the streaming, encryption etc. work.

lib/uplink/ex_restrict_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Print it!
+	fmt.Fprintln(out, string(data))
+	return nil
+}
+
+func Example_restrictAccess() {
```
egonelbre: Let's use testplanet here instead and ensure that this runs as a test.

lib/uplink/encryption.go
thepaul: I _do_ understand the rules you're proposing for what to put in each one, but what I don't understand is _why_. Does it avoid some import cycles or lessen dependencies for a libuplink user? Or is it just that we want some sort of semantic separation between "simple things" and "advanced things" so that users know when they're crossing the line into Hard Mode?

egonelbre: More of stable-ish API and unstable-ish API. But also yeah... when you cross over into `uplink` you might lose compatibility with `lib/uplinkc` and so on.

jtolds: due to licensing issues, eventually we need to separate uplink code from storj/storj entirely. when we do that, i would also like to fix the naming. having lib/uplink and uplink is very confusing, independent of the motivation. my plan is that storj.io/uplink will be storj.io/storj/lib/uplink, and storj.io/uplink/internal (or something) will be storj.io/storj/uplink

lib/uplink/ex_restrict_test.go
jtolds: we want this as documentation. testplanet does run this though. see examples_test.go

zeebo:
Please do further reviews on #2349. Sorry for the confusion!

PR #2305, opened 2019-06-24
Mask IP Addresses to subnets

stefanbenten:
What: This PR updates the last_ip column to contain only the /24 network the IP of the host is residing in. That ensures we are not vulnerable to someone running 200 nodes on different IPs within the same subnet.

Why: Improves data distribution across entire IP Ranges.

Please describe the tests:
 - Test 1: Done, ensures that the IPv4 and IPv6 Masking works as expected.
 - Test 2: Not needed.
 
Please describe the performance impact: Slightly more conversion, but improved grouping performance for the node selection


pkg/overlay/cache.go
```diff
@@ -408,3 +408,21 @@ func getIP(ctx context.Context, target string) (_ string, err error) {
...
+		return "", err
+	}
+
+	addr := net.ParseIP(ip)
+	if addr == nil {
+		return "", errors.New("invalid ip")
+	}
+
+	//Filter all IP Addresses into /24 Subnet's
+	mask := net.CIDRMask(24, 32)
```
thepaul: We should check first if `len(addr) == 4`. If `len(addr) == 16` instead, we should not apply a mask (as discussed, there doesn’t seem to be any very reasonable way to guess the appropriate subnet size with ipv6 addresses, so we will punt on masking those for now)

stefanbenten: Ill add the check and the migration 👍

thepaul: For the migration, `UPDATE nodes SET last_ip = host(network(set_masklen(last_ip::INET, 24))) WHERE last_ip <> '' AND family(last_ip::INET) = 4;` ought to make the necessary changes

stefanbenten: I included a grouping per /64 subnet. I know that is not the best way, but better than allowing it entirely there..

littleskunk:
@thepaul we need one more approval.

pkg/overlay/cache.go
```diff
@@ -399,7 +399,7 @@ func (cache *Cache) GetMissingPieces(ctx context.Context, pieces []*pb.RemotePie
 	return missingPieces, nil
 }
 
-func getIP(ctx context.Context, target string) (_ string, err error) {
+func getIP(ctx context.Context, target string) (ip string, err error) {
```
jtolds: can we have this be a net.IPAddr return value? that way GetNetwork doesn't need to call net.ParseIP

stefanbenten:
> someday we should rename the last_ip column to last_ip_net or something

Renamed it to last_net 👍

pkg/overlay/cache.go
```diff
@@ -399,15 +399,39 @@ func (cache *Cache) GetMissingPieces(ctx context.Context, pieces []*pb.RemotePie
 	return missingPieces, nil
 }
 
-func getIP(ctx context.Context, target string) (_ string, err error) {
+func getIP(ctx context.Context, target string) (ip net.IPAddr, err error) {
```
ifraixedes: the `ip` return parameter is declared and never used, it was not needed to be changed from `_`

PR #2314, opened 2019-06-24
fix ordersDB methods to take correct args

JessicaGreben:
What/Why: 
All methods in orders.go are fixed so that they take projectID and bucketID as separate arguments with appropriate types instead of parsing bucketID in orderDB methods. Also fixed accountingDB where it had this same anti-pattern. 

Please describe the tests:
Updated tests to take projectID and bucketID as separate arguments. Moved the splitBucketID tests to the satellite/orders package.

N/A


pkg/accounting/tally/tally.go
```diff
@@ -142,9 +143,13 @@ func (t *Service) CalculateAtRestData(ctx context.Context) (latestTally time.Tim
 					bucketID := storj.JoinPaths(project, bucketName)
 
 					bucketTally := bucketTallies[bucketID]
+					projectID, err := uuid.Parse(project)
+					if err != nil {
+						return err
```
egonelbre: wrap error

pkg/accounting/tally/tally.go
```diff
@@ -142,9 +143,13 @@ func (t *Service) CalculateAtRestData(ctx context.Context) (latestTally time.Tim
...
+					}
 					if bucketTally == nil {
 						bucketTally = &accounting.BucketTally{}
-						bucketTally.ProjectID = []byte(project)
+						bucketTally.ProjectID = projectID[:]
```
egonelbre: Why is ProjectID here not `uuid`?

satellite/orders/endpoint.go
```diff
@@ -208,7 +209,11 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
 		now := time.Now().UTC()
 		intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
 
-		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, bucketID, orderLimit.Action, order.Amount, intervalStart); err != nil {
+		projectID, bucketName, err := SplitBucketID(bucketID)
```
egonelbre: UseSerialNumber should be returning project id and bucket name separately.

satellite/orders/orders_test.go
```diff
@@ -174,3 +176,27 @@ func noLongTailRedundancy(planet *testplanet.Planet) uplink.RSConfig {
 	redundancy.SuccessThreshold = redundancy.MaxThreshold
 	return redundancy
 }
+
+func TestSpliteBucketID(t *testing.T) {
```
egonelbre: typo `TestSpliteBucketID`

satellite/orders/service.go
```diff
@@ -204,7 +205,11 @@ func (service *Service) CreateGetOrderLimits(ctx context.Context, uplink *identi
 		return nil, Error.Wrap(err)
 	}
 
-	if err := service.updateBandwidth(ctx, bucketID, limits...); err != nil {
+	projectID, bucketName, err := SplitBucketID(bucketID)
```
egonelbre: The bucket id shouldn't need to be split the argument should already be split.

satellite/orders/service.go
```diff
@@ -664,27 +685,42 @@ func (service *Service) CreatePutRepairOrderLimits(ctx context.Context, repairer
...
 	intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
 
-	return service.orders.UpdateBucketBandwidthInline(ctx, bucketID, pb.PieceAction_PUT, amount, intervalStart)
+	return service.orders.UpdateBucketBandwidthInline(ctx, projectID, bucketName, pb.PieceAction_PUT, amount, intervalStart)
+}
+
+// SplitBucketID takes a bucketID, splits on /, and returns a projectID and bucketName
+func SplitBucketID(bucketID []byte) (projectID *uuid.UUID, bucketName []byte, err error) {
+	pathElements := bytes.Split(bucketID, []byte("/"))
+	bucketName = pathElements[1]
```
egonelbre: This panics if there isn't enough elements.

pkg/accounting/tally/tally.go
JessicaGreben: I attempted to change it to uuid here, but it became kind of messy so I decided to leave it at []byte

egonelbre: Add a note somewhere that this needs to be fixed.

satellite/orders/endpoint.go
JessicaGreben: the serial number tables stores the bucketID so thats what it returns. I would have to split the bucketID

egonelbre: ah, right... I thought we already had split that as well

satellite/orders/service.go
JessicaGreben: bucketID gets passed in as an arg to this function

egonelbre: Ah, meaning `CreateGetOrderLimits` should probably already get projectID and bucket name separately. But yeah, we can leave these for future.

PR #2315, opened 2019-06-24
account for disqualified nodes in data repair tests

navillasa:
What:
- instead of only killing nodes in the data repair tests (TestDataRepair and TestSegmentStoreRepair), this also disqualifies one node and expects successful repair

Why:
- the data repair tests didn't explicitly account for disqualified nodes
- #2 in this story: https://storjlabs.atlassian.net/browse/V3-1952

Please describe the tests:
 - Test 1: TestDataRepair - checks that repair happens correctly when nodes are turned off or disqualified. More of a test about how data repair works within the "ecosystem" (with other services running).
 - Test 2: TestSegmentStoreRepair - also checks that repair happens correctly when nodes are turned off or disqualified, but meant to help zero-in on testing the data repair functionality (less about integration)
- Test 3: TestRepairMultipleDisqualified (new test) - checks that repair happens correctly when multiple nodes are disqualified
 

pkg/datarepair/datarepair_test.go
```diff
@@ -126,6 +140,139 @@ func TestDataRepair(t *testing.T) {
...
+				assert.NoError(t, err)
+			}
+		}
+
+		// we should be able to download data without any of the original nodes
+		newData, err := ul.Download(ctx, satellite, "testbucket", "test/path")
+		assert.NoError(t, err)
+		assert.Equal(t, newData, testData)
+
+		// updated pointer should not contain any of the killed nodes
```
ifraixedes: shouldn't be `// updated pointer should not contain any of the disqualified nodes`?

pkg/datarepair/datarepair_test.go
```diff
@@ -126,6 +140,139 @@ func TestDataRepair(t *testing.T) {
...
+
+		satellite.Repair.Checker.Loop.Restart()
+		satellite.Repair.Checker.Loop.TriggerWait()
+		satellite.Repair.Checker.Loop.Pause()
+		satellite.Repair.Repairer.Loop.Restart()
+		satellite.Repair.Repairer.Loop.TriggerWait()
+		satellite.Repair.Repairer.Loop.Pause()
+		satellite.Repair.Repairer.Limiter.Wait()
+
+		// kill nodes kept alive to ensure repair worked
```
ifraixedes: I'm missing something here.
If repair is executed before this nodes are killed, the only pieces to repair are the ones which were store in the disqualified nodes, so if we kill these nodes and repair isn't run again, how is it guarantee that the number of pieces which remain in the network (nodes alive an not disqualified) are the minimum for being able to retrieve the data?

pkg/datarepair/datarepair_test.go
```diff
@@ -68,17 +70,25 @@ func TestDataRepair(t *testing.T) {
 		remotePieces := pointer.GetRemote().GetRemotePieces()
 		minReq := redundancy.GetMinReq()
 		numPieces := len(remotePieces)
-		toKill := numPieces - (int(minReq) + 1)
+		// disqualify one storage node
+		toDisqualify := 1
+		toKill := numPieces - (int(minReq+1) + toDisqualify)
```
kaloyan-raev: Let's check that we are killing at least one node after all these calculations.

pkg/datarepair/datarepair_test.go
```diff
@@ -68,17 +70,25 @@ func TestDataRepair(t *testing.T) {
...
 		// we should have enough storage nodes to repair on
-		assert.True(t, (numStorageNodes-toKill) >= numPieces)
+		assert.True(t, (numStorageNodes-toKill-toDisqualify) >= numPieces)
```
kaloyan-raev: I think this one should be `require` instead of `assert`. If this check is not satisfied then we have a bad test setup and we don't really want the test to continue.

pkg/datarepair/datarepair_test.go
kaloyan-raev: Here is what happens:
- Test data is uploaded to 7 nodes
- 3 of them are disqualified
- Data repair is triggered. It repairs the data from the remaining 4 nodes to additional 3 new nodes
- The 4 nodes from which the data was repaired are shut down.
- Now we have just the 3 new nodes to which the data was repaired.
- We try to download the data from these 3 nodes. It is successful because 3 nodes are enough for download.

pkg/storage/segments/repairer_test.go
```diff
@@ -68,23 +70,36 @@ func TestSegmentStoreRepair(t *testing.T) {
 		remotePieces := pointer.GetRemote().GetRemotePieces()
 		minReq := redundancy.GetMinReq()
 		numPieces := len(remotePieces)
-		toKill := numPieces - (int(minReq) + 1)
+		// disqualify one storage node
+		toDisqualify := 1
+		toKill := numPieces - (int(minReq+1) + toDisqualify)
```
kaloyan-raev: Like in TestDataRepair, add a check that we kill at least one node. I am not sure this will pass with the 2/3/4/4 RS scheme in this test.

navillasa: it did pass!

navillasa: oh I was looking at the wrong thing.. no it didn't.

navillasa: well I updated the RS setting to be the same as TestDataRepair, but now I'm kind of wondering what the benefit is of having two separate tests that seem to do about the same thing. do you think it would be okay to remove one of them?

kaloyan-raev: These two tests really look to be the same. Let's remove the TestSegmentStoreRepair test.

pkg/datarepair/datarepair_test.go
```diff
@@ -36,9 +44,11 @@ func TestDataRepair(t *testing.T) {
 		satellite.Repair.Checker.Loop.Pause()
 		satellite.Repair.Repairer.Loop.Pause()
 
-		testData := testrand.Bytes(1 * memory.MiB)
+		testData := make([]byte, 1*memory.MiB)
+		_, err := rand.Read(testData)
+		require.NoError(t, err)
```
egonelbre: There's testrand package that makes this part more convenient.

pkg/datarepair/datarepair_test.go
```diff
@@ -125,7 +148,147 @@ func TestDataRepair(t *testing.T) {
...
+		// stop discovery service so that we do not get a race condition when we delete nodes from overlay cache
+		satellite.Discovery.Service.Discovery.Stop()
+		satellite.Discovery.Service.Refresh.Stop()
+
+		satellite.Repair.Checker.Loop.Pause()
+		satellite.Repair.Repairer.Loop.Pause()
+
+		testData := make([]byte, 1*memory.MiB)
+		_, err := rand.Read(testData)
+		require.NoError(t, err)
```
egonelbre: There's testrand package that makes this part more convenient. Might have gotten accidentally reverted.

pkg/datarepair/datarepair_test.go
navillasa: oh shoot my mistake. I can make a tiny pr fix if needed

navillasa: https://github.com/storj/storj/pull/2408

PR #2318, opened 2019-06-24
Creates Routing Table Antechamber

jenlij:
What: 
• Adds a bolt bucket called antechamber to the Routing Table struct
• Implements antechamberAddNode, antechamberRemoveNode, antechamberFindNear
• Stubs method nodeHasValidVoucher to check that a node has at least one valid voucher
Why: 
To fulfill the routing table antechamber feature of https://github.com/storj/storj/blob/master/docs/design/kademlia_audit_gating.md

Please describe the tests:
• Test 1: TestAntechamberAddNode - Checks a node is added to antechamber even if there are no neighborhood nodes (empty rt). Checks that a node that is not in the RT neighborhood should not get added to antechamber. A node that is in neighborhood should.
• Test 2: TestAntechamberRemoveNode - Checks removing nonexistent node from antechamber doesn't break. Checks that a node in the antechamber is removed.
• Test 3: TestAntechamberFindNear - Checks 0 nodes are returned from empty antechamber. Checks that the correct nodes are selected when there are nodes.
 
Please describe the performance impact: None yet, but once these methods are integrated into the routing table methods, there may be a subtle performance decrease when populating the routing tables since the antechamber methods will be called as well. But since this doesn't happen very often, this impact will be negligible.


bootstrap/bootstrapdb/database.go
```diff
@@ -22,19 +22,20 @@ type Config struct {
 
 // DB contains access to different database tables
 type DB struct {
-	kdb, ndb storage.KeyValueStore
+	kdb, ndb, adb storage.KeyValueStore
```
egonelbre: Let's make a proper interface for the storage from the start rather than pushing it forward. Kademlia should be doing that as well, but we haven't gotten around to it.

I.e. ideally this should read `kademlia kademlia.DB`.

jenlij: Ok I'll make separate PR to get that in first.

pkg/kademlia/antechamber.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		return nil
+	})
+	return closestNodes, Error.Wrap(err)
+}
+
+// checks whether the node in question has a valid voucher.
+// If true, call addNode
+// If false, call antechamberAddNode
+func (rt *RoutingTable) nodeHasValidVoucher(ctx context.Context, node *pb.Node, vouchers []*pb.Voucher) bool {
```
stefanbenten: Would it be possible to move this function and commented code into a future PR/draft instead of commiting it to master as is?

storage/boltdb/client.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/boltdb/bolt"
 	"github.com/zeebo/errs"
-	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+	"gopkg.in/spacemonkeygo/monkit.v2"
```
stefanbenten: `monkit` should not be removed

pkg/kademlia/antechamber.go
jenlij: sure!

storage/boltdb/client.go
jenlij: Idk what happened here, ill put it back

VinozzZ: I'm not sure we need that alias for the package?

PR #2331, opened 2019-06-25
satellite/orders: don't panic

egonelbre:
What: orders.Service implementation was using DPanicOnInvalid, which should be removed.


pkg/pb/utils.go
```diff
@@ -63,15 +62,6 @@ func CopyNode(src *Node) (dst *Node) {
 	return &node
 }
 
-// DPanicOnInvalid panics if NodeType is invalid if zap is in development mode,
```
VinozzZ: Shouldn't this only panic when it's in dev mode?

egonelbre: Yes, but the rest of the code is invalid if it doesn't panic, the error should be properly handled. Either way, removing this completely so people won't accidentally use this for production things.

PR #2334, opened 2019-06-25
Fix some leaks and add notes about close handling

egonelbre:
What: fix some leaks and add notes about things that need fixing.

`// TODO(leak): unclear ownership semantics` means that there should be clearer ownership semantics. For `New*` the assumption is that the caller is responsible for closing things passed as arguments. For `Dial*` the assumption is that the callee closes everything.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

JessicaGreben:
I'm curious if there is some way to ensure people follow this standard? Re: "For New* the assumption is that the caller is responsible for closing things passed as arguments. For Dial* the assumption is that the callee closes everything."

egonelbre:
Not sure. The more general principle is that the caller is responsible for any arguments, unless explicitly specified otherwise. E.g Dial internally makes a connection, so it is responsible for closing it.

PR #2335, opened 2019-06-25
remove flakiness in TestDataRepair and TestSegmentStoreRepair

navillasa:
What/why:
- stops audit loop in TestDataRepair and TestSegmentStoreRepair
    - to prevent possible conflicts between audit and repair jobs. We don't want nodes we are trying to download from during repair to be blocked by audit jobs
- increases repair timeout from 10s to 1m in testplanet
    - adding the satellite logger to ecclient revealed that the reason for the test failures was the repair job timing out. Testplanet configured this timeout to 10s, and this wasn't sufficient fairly often


lib/uplink/project.go
```diff
@@ -164,7 +165,7 @@ func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *Enc
 	}
 	encryptionScheme := cfg.EncryptionParameters.ToEncryptionScheme()
 
-	ec := ecclient.NewClient(p.tc, p.uplinkCfg.Volatile.MaxMemory.Int())
+	ec := ecclient.NewClient(zap.L(), p.tc, p.uplinkCfg.Volatile.MaxMemory.Int())
```
egonelbre: This should be passed down from somewhere.

pkg/metainfo/kvmetainfo/buckets_test.go
```diff
@@ -314,7 +315,7 @@ func newMetainfoParts(planet *testplanet.Planet) (*kvmetainfo.DB, streams.Store,
 	}
 	// TODO(leak): call metainfo.Close somehow
 
-	ec := ecclient.NewClient(planet.Uplinks[0].Transport, 0)
+	ec := ecclient.NewClient(zap.L(), planet.Uplinks[0].Transport, 0)
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

pkg/miniogw/gateway_test.go
```diff
@@ -690,7 +691,7 @@ func initEnv(ctx context.Context, planet *testplanet.Planet) (minio.ObjectLayer,
 	}
 	// TODO(leak): close m metainfo.Client somehow
 
-	ec := ecclient.NewClient(planet.Uplinks[0].Transport, 0)
+	ec := ecclient.NewClient(zap.L(), planet.Uplinks[0].Transport, 0)
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

pkg/storage/ec/client.go
```diff
@@ -86,8 +89,8 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, r
 		return nil, nil, Error.New("duplicated nodes are not allowed")
 	}
 
-	zap.S().Infof("Uploading to storage nodes using ErasureShareSize: %d StripeSize: %d RepairThreshold: %d OptimalThreshold: %d",
-		rs.ErasureShareSize(), rs.StripeSize(), rs.RepairThreshold(), rs.OptimalThreshold())
+	ec.log.Info(fmt.Sprintf("Uploading to storage nodes using ErasureShareSize: %d StripeSize: %d RepairThreshold: %d OptimalThreshold: %d",
```
egonelbre: There is `ec.log.Sugar().Infof`, same for other things.

pkg/storage/ec/client_planet_test.go
```diff
@@ -47,7 +48,7 @@ func TestECClient(t *testing.T) {
 
 	planet.Start(ctx)
 
-	ec := ecclient.NewClient(planet.Uplinks[0].Transport, 0)
+	ec := ecclient.NewClient(zap.L(), planet.Uplinks[0].Transport, 0)
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

pkg/storage/segments/repairer_test.go
```diff
@@ -96,7 +99,7 @@ func TestSegmentStoreRepair(t *testing.T) {
 		// repair segment
 		os := satellite.Orders.Service
 		oc := satellite.Overlay.Service
-		ec := ecclient.NewClient(satellite.Transport, 0)
+		ec := ecclient.NewClient(zap.L(), satellite.Transport, 0)
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

pkg/storage/segments/store_test.go
```diff
@@ -270,7 +271,7 @@ func runTest(t *testing.T, test func(t *testing.T, ctx *testcontext.Context, pla
 		require.NoError(t, err)
 		defer ctx.Check(metainfo.Close)
 
-		ec := ecclient.NewClient(planet.Uplinks[0].Transport, 0)
+		ec := ecclient.NewClient(zap.L(), planet.Uplinks[0].Transport, 0)
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

uplink/config.go
```diff
@@ -104,7 +105,7 @@ func (c Config) GetMetainfo(ctx context.Context, identity *identity.FullIdentity
 		return nil, nil, Error.New("failed to create project: %v", err)
 	}
 
-	ec := ecclient.NewClient(tc, c.RS.MaxBufferMem.Int())
+	ec := ecclient.NewClient(zap.L(), tc, c.RS.MaxBufferMem.Int())
```
egonelbre: This should be passed down from somewhere. Or use `zaptest.NewLogger`.

lib/uplink/project.go
mobyvb: should we create a `zap.L()` in `cmd/uplink/cmd/root.go` and pass it down from there? As far as I know, there is no zap logger created on the uplink side right now

mobyvb: I cannot find a way to pass it in from a higher level without needing to add `zap.L()` in more places. I could add it to the `libuplink.Uplink` struct and pass it down into the `Project` struct that way, but then, everywhere that calls `NewUplink` which is a lot of places will need to pass in `zap.L()`. Most places that call `NewUplink` do not have loggers already created. Should I still create them and pass them in from a higher level? I can do it but I don't think it's very pretty either way.

jtolds: if you want libuplink to do logging, i think *Uplink should not take a new constructor argument but instead should have a SetLogger method

jtolds: the reason is because most users won't care, and this is our public api

mobyvb: I actually don't care about uplink logging at all in this case. I specifically care about ecclient logs on the satellite. Unfortunately, I can't add a logger to ecclient without having to handle it on the uplink side as well. And it can't just be `nil` otherwise we deal with lots of exceptions. So it doesn't seem like the `SetLogger` method is a solution for this case

egonelbre: For uplink there probably should be two constructors. `New` and `NewWithLogger`, where `New` defaults to some `zap.L()`.

pkg/metainfo/kvmetainfo/buckets_test.go
```diff
@@ -274,14 +275,14 @@ func runTest(t *testing.T, test func(*testing.T, context.Context, *testplanet.Pl
...
-		db, streams, err := newMetainfoParts(planet)
+		db, streams, err := newMetainfoParts(planet, t)
 		require.NoError(t, err)
 
 		test(t, ctx, planet, db, streams)
 	})
 }
 
-func newMetainfoParts(planet *testplanet.Planet) (*kvmetainfo.DB, streams.Store, error) {
+func newMetainfoParts(planet *testplanet.Planet, t *testing.T) (*kvmetainfo.DB, streams.Store, error) {
```
egonelbre: Make `t` the first argument rather than last, usually `testing.T` is the first argument.

pkg/miniogw/gateway_test.go
```diff
@@ -649,15 +650,15 @@ func runTest(t *testing.T, test func(context.Context, minio.ObjectLayer, storj.M
...
+	layer, m, strms, err := initEnv(ctx, planet, t)
 	if !assert.NoError(t, err) {
 		return
 	}
 
 	test(ctx, layer, m, strms)
 }
 
-func initEnv(ctx context.Context, planet *testplanet.Planet) (minio.ObjectLayer, storj.Metainfo, streams.Store, error) {
+func initEnv(ctx context.Context, planet *testplanet.Planet, t *testing.T) (minio.ObjectLayer, storj.Metainfo, streams.Store, error) {
```
egonelbre: Make `t` the first argument rather than last, usually `testing.T` is the first argument.

lib/uplink/project.go
jtolds: SetLogger doesn't work?

```
ul, err := NewUplink(ctx, cfg)
ul.SetLogger(log)
// do stuff
```

mobyvb: What I meant is - what if someone creates a new uplink and does not set a logger? Then in all the places where we use `uplink.logger.Debug` or something, we will get a nil pointer exception. 
Full disclosure, I do not know much about `SetLogger`, but I have implemented Egon's suggestion above, with `NewUplink` and `NewUplinkWithLogger`

pkg/miniogw/gateway_test.go
mobyvb: should it come before `ctx`? I will put it there unless I hear otherwise

navillasa: did this for the above, but can't for this one because the linter wants ctx as the first param

mobyvb: actually, the linter complained so I did `ctx, t, planet`

navillasa: me too lol

lib/uplink/project.go
zeebo: there's a zap.NewNop() that you would use by default.

lib/uplink/project.go
```diff
@@ -172,6 +172,9 @@ func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *Enc
 	}
 	encryptionScheme := cfg.EncryptionParameters.ToEncryptionScheme()
 
+	if p.log == nil {
+		p.log = zap.L()
```
egonelbre: Assume that this OpenBucket might be called in parallel.

internal/testplanet/satellite.go
```diff
@@ -154,7 +154,7 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 			Repairer: repairer.Config{
 				MaxRepair:    10,
 				Interval:     time.Hour,
-				Timeout:      10 * time.Second, // Repairs can take up to 4 seconds. Leaving room for outliers
+				Timeout:      1 * time.Minute, // Repairs can take up to 10 seconds. Leaving room for outliers
```
egonelbre: The question is whether the flakiness was fixed by the go test parallelism changes or by this change?

mobyvb: We added the logger last week to narrow down the cause of the tests when they failed (before the parallelism changes), and the cause of the failing was always due to repair being cutoff because of the 10s timeout (which we didn't see before because the repairer logs were not visible in jenkins).
After increasing the timeout, we were able to get the test to always pass. I believe 10s was the timeout when the testplanet test had fewer storagenodes, so it makes sense that it is not always sufficient now.

egonelbre: From the logs it seems that the tests started passing after merging in the parallelism changes. So is the 10s is at the cusp of passing or was something else going on?

mobyvb: The tests were also consistently passing before jenkins started failing a lot (for reasons I think unrelated to this change). I believe 10s was on the cusp before. This test was failing on like 50% of builds before. We can set it back to 10s and see what happens to the builds if you'd like

egonelbre: Yeah, would be nice to know how much difference that made. But yeah, 1min is fine as well if it makes sense for the test.

mobyvb: We tried it a bunch of times again with 10s and it seemed to work most of the time - but we did see the timeout error one time over the course of 10 builds. I think it is dependent on how many simultaneous builds are going on and how much load Jenkins is under. 1 minute shouldn't be necessary, but it will prevent the test from failing frequently like it has been lately.

egonelbre: Jenkins doesn't run multiple jobs concurrently, so the load should be similar. There are slight differences between the jenkins nodes though.

PR #2341, opened 2019-06-25
satellite/console: add get current reward offer to console

VinozzZ:
What: 
we need to expose reward offer data to satellite GUI
Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/consoleweb/consoleql/query.go
```diff
@@ -83,6 +86,19 @@ func rootQuery(service *console.Service, mailService *mailservice.Service, types
...
 			},
+			ActiveRewardQuery: &graphql.Field{
+				Type: types.reward,
+				Args: graphql.FieldConfigArgument{
+					FieldType: &graphql.ArgumentConfig{
+						Type: graphql.NewNonNull(graphql.Int),
+					},
+				},
+				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					rewardType, _ := p.Args[FieldType].(int)
```
egonelbre: Is there a reason this isn't a string instead of int? It'll mean we need to lock in the iota values in offer type values to not break requests.

VinozzZ: It can be a string. I'm not sure I understand the difference here? Could you explain more?

egonelbre: We have currently defined in our code:
```
// OfferType indicates the type of an offer
type OfferType int

const (
	// FreeCredit is a type of offers used for Free Credit Program
	FreeCredit = OfferType(iota)
	// Referral is a type of offers used for Referral Program
	Referral
)
```

When we change

```
const (
	// FreeCredit is a type of offers used for Free Credit Program
	FreeCredit = OfferType(iota)
	SomeNewOfferType
	// Referral is a type of offers used for Referral Program
	Referral
)
```

Then all old clients using `1` will start sending the wrong request. This can be avoided by using named strings. e.g. letting requests use the "free-credit" instead of a number.

Or locking the indices to a specific value:

```
const (
	// FreeCredit is a type of offers used for Free Credit Program
	FreeCredit = OfferType(0)
	// Referral is a type of offers used for Referral Program
	Referral = OfferType(1)
)
```

This makes it harder to accidentally change these.

VinozzZ: I see. That makes sense. I will open another PR to change it to hard-coded integers

egonelbre: Note for making requests from graphql side it probably feels nicer to use "referral" rather than 0.

Also for enumerations keep the 0 value as "invalid" or the default... it helps to avoid bugs about forgetting to set the value.

PR #2345, opened 2019-06-25
Replace base64 encoding for path segments

mniewrzal:
What: Replace base64 encoding for path segments with encoding algorithm described in #2386 :

Why: Main motivation is to reduce the overhead from 33% for base64 encoding.
https://storjlabs.atlassian.net/browse/V3-1787

Please describe the tests:
 - Test 1: all automatic tests should pass without major modification
 - Test 2: additional test case added
 
Please describe the performance impact: should be faster than base64 (benchmark added) and should have much smaller overhead


pkg/encryption/path.go
```diff
@@ -159,3 +155,49 @@ func decryptPathComponent(comp string, cipher storj.Cipher, key *storj.Key) (str
...
 	return string(decrypted), nil
 }
+
+func encodeSegment(segment []byte) []byte {
+	if len(segment) == 0 {
+		return segment
+	}
+
+	// escape 0x00 to `{0x00, 0x01}`
+	// escape `/` to `{0x00, 0x02}`
```
kaloyan-raev: Can we use constants instead of magic numbers?

mniewrzal: sure

mniewrzal:
No, its not covered. 

In which cases we are displaying encrypted paths?

kaloyan-raev:
I don't think we display them to end users, but I imagine they often appear in logs.

mniewrzal:
then I think it depends how valuable will be having a human-readable encrypted path in logs. I believe we can do this in separate change.

egonelbre:
With the path handling the idea is to move towards `EncryptedPath []string`, rather than keep it in encoded form. Only at the last layer, when storing in the database, would this encoding be used. It seems keeping it in encoded form keeps us making mistakes.

egonelbre:
Jeff has a design doc for a slightly better idea https://github.com/storj/storj/pull/2386 for path encoding

mniewrzal:
I pushed basic implementation of Jeff encoding but it looks that we have problem with `InjuredSegment` because `path` there is a `string` and it fails repair tests:
```
boltdb error: checker error: error adding injured segment to queue sql: converting argument $2 type: Protobuf Valuer: proto: field "repair.InjuredSegment.Path" contains invalid UTF-8
```
https://github.com/storj/storj/blob/29d16b4d68aa5b067b5503d1de4438b61ea41db6/satellite/satellitedb/repairqueue.go#L27

First idea is to migrate `path` from `string` to `byte` but maybe someone will have a better idea.
@zeebo @egonelbre

egonelbre:
@mniewrzal yes, all paths, bucket names should be defined as bytes in database and protobuf.

pkg/encryption/examples_test.go
```diff
@@ -27,7 +27,7 @@ func ExampleEncryptPath() {
 		panic(err)
 	}
 	fmt.Println("path to encrypt:", path)
-	fmt.Println("encrypted path: ", encryptedPath)
+	fmt.Println("encrypted path: ", []byte(encryptedPath))
```
zeebo: Can we do `hex.EncodeToString` or something instead of `[]byte` so that the output is more human readable?

pkg/encryption/path.go
```diff
@@ -159,3 +164,59 @@ func decryptPathComponent(comp string, cipher storj.CipherSuite, key *storj.Key)
...
+
+func encodeSegment(segment []byte) []byte {
+	if len(segment) == 0 {
+		return emptyComponent
+	}
+
+	result := make([]byte, 0, len(segment)*2+1)
+	result = append(result, notEmptyComponentPrefix)
+	for i := 0; i < len(segment); i++ {
+		switch {
```
zeebo: You can change this to

```go
switch segment[i] {
case escape1:
case escape1+1:
case escape2:
...
```

which seems cleaner to me.

pkg/encryption/path.go
```diff
@@ -159,3 +164,59 @@ func decryptPathComponent(comp string, cipher storj.CipherSuite, key *storj.Key)
...
+
+func decodeSegment(segment []byte) []byte {
+	if len(segment) == 0 {
+		return segment
+	}
+	if segment[0] == emptyComponent[0] {
+		return []byte{}
+	}
+
+	// TODO should first byte different than x02 should be invalid?
```
zeebo: I think that sequences that aren't properly escaped should be errors when attempting to unescape.

pkg/encryption/path_test.go
```diff
@@ -98,3 +100,60 @@ func forAllCiphers(test func(cipher storj.CipherSuite)) {
...
+		[]byte{'/', '/', '/', '/', '/'},
+		[]byte{0, 0, 0, 0, 0, 0, 0, 0, 0},
+
+		[]byte{'a', '/', 'a', '2', 'a', 'a', 0, '1', 'b', 255},
+		[]byte{'/', '/', 'a', 0, 'a', 'a', 0, '1', 'b', 'g', 'a', 'b', '/'},
+		[]byte{0, '/', 'a', '0', 'a', 'a', 0, '1', 'b', 'g', 'a', 'b', 0},
+	}
+
+	for _, segment := range segments {
+		encoded := encodeSegment(segment)
```
zeebo: Can we also require that `bytes.IndexByte(encoded, {0, 255, '/'})` is always `-1`?

zeebo:
With respect to the paths containing bytes that can't be displayed: the `paths.{Encrypted,Unencrypted}` types have `String` methods that can be used to transform them into something with only printable characters. They also have a `Raw` method that I've been trying to use everywhere when pulling the actual contents out.

pkg/encryption/path.go
mniewrzal: done

pkg/encryption/path.go
mniewrzal: good idea

pkg/encryption/examples_test.go
mniewrzal: use `hex.EncodeToString` for now

pkg/encryption/path_test.go
mniewrzal: done

mniewrzal:
@zeebo yes, using `String/Raw` methods sounds nice. I guess we need to make the code more consistent to widely use this approach.

pkg/encryption/path.go
```diff
@@ -294,3 +302,73 @@ func decryptPathComponent(comp string, cipher storj.CipherSuite, key *storj.Key)
...
+			segment[currentIndex] = segment[i] + segment[i+1] - 1
+			i++
+		case segment[i] == escape3:
+			segment[currentIndex] = segment[i+1] - 1
+			i++
+		default:
+			segment[currentIndex] = segment[i]
+		}
+		currentIndex++
+	}
```
zeebo: I think we need to be more strict in what we disallow, otherwise we can have distinct strings that map to the same value. For example, `\x02\x01\x23` and `\x02\x22` both map to `\x22`. This could possibly cause some havoc if a malicious client uploaded to the same path in multiple ways.

Can we enforce for some segment:

1. The last byte is not in `{escape1, escape2, escape3}`
2. Any byte after an escape character is `\x01` or `\x02`
3. It does not contain any characters in `{\x00, \xff, \x2f}`
4. It is non-empty
5. It begins with a character in `{\x01, \x02}`

I believe that's sufficient to enforce uniqueness, and `decodeSegment` can just call that function. Additionally, can we export a function that the satellite can call to check that a full encoded path is valid? That function would iterate over the segments and check that each segment is valid.

mniewrzal: Good idea, only thing I'm not sure if we should incorporate those rules as a separate function and use it in `decodeSegment` or put those rules directly into `decodeSegment` (might be faster). 

I think segment validation on satellite side is a good candidate for separate PR.

mniewrzal: I pushed method for encoded segment validation. I feel it can be more optimal but I would prefer to update this later.

pkg/encryption/path.go
```diff
@@ -15,6 +14,16 @@ import (
 	"storj.io/storj/pkg/storj"
 )
 
+var (
+	emptyComponentPrefix    = byte('\x01')
+	notEmptyComponentPrefix = byte('\x02')
+	emptyComponent          = []byte{emptyComponentPrefix}
+
+	escape1 = byte('\x2e')
```
egonelbre: Let's use clearer naming here... maybe `escapeSlash`, `escapeFF`, `escape00` or something? Or `escapeLow`, `escapeHigh`?

pkg/encryption/path_test.go
```diff
@@ -63,3 +66,97 @@ func forAllCiphers(test func(cipher storj.CipherSuite)) {
 		test(cipher)
 	}
 }
+
+func TestSegmentEncoding(t *testing.T) {
```
egonelbre: Let's test:
* encoding/decoding random data paths,
* parsing completely random data
* parsing all byte slices up to length 3. (or if it takes too long, then 2)

pkg/encryption/path.go
mniewrzal: done

pkg/encryption/path_test.go
mniewrzal: done

pkg/encryption/path_test.go
```diff
@@ -63,3 +66,135 @@ func forAllCiphers(test func(cipher storj.CipherSuite)) {
...
+					combination(i+1, j+1)
+				}
+			}
+			return
+		}
+		combination(0, 0)
+	}
+
+	allCombinations(func(b []byte) {
+		segments = append(segments, b)
```
egonelbre: this appends the same slice all the time

egonelbre:

PR #2349, opened 2019-06-26
lib/uplink: encryption context

zeebo:
What: this PR exposes an encryption context type that works somewhat like the existing API keys. It can be restricted by path, serialized, parsed, etc. Unlike API keys, encryption contexts can be merged as someone gains access to new folders/buckets.

~Blocked on #2293~

Why: The EncryptionAccess type was underfeatured.

Please describe the tests:

    Test 1: An example test where one admin gives another admin full access
    Test 2: An example test where one admin gives a user restricted access

Please describe the performance impact: Generating keys from passphrases will be deliberately both CPU intensive and rare. We also need to retrieve a project salt from the server when generating a key from a passphrase.


internal/testplanet/uplink.go
```diff
@@ -321,3 +327,75 @@ func atLeastOne(value int) int {
...
+
+	blockSize := config.GetEncryptionScheme().BlockSize
+	if int(blockSize)%config.RS.ErasureShareSize.Int()*config.RS.MinThreshold != 0 {
+		err = errs.New("EncryptionBlockSize must be a multiple of ErasureShareSize * RS MinThreshold")
+		return nil, nil, err
+	}
+
+	// TODO(jeff): there's some cycles with libuplink and this package in the libuplink tests
+	// and so this package can't import libuplink. that's why this function is duplicated
+	// in some spots.
```
egonelbre: There shouldn't be import cycles when:
1. only test packages use testplanet
2. tests are written using `package xxx_test`

But I guess resolving the deps would be too much here.

lib/uplink/encryption.go
```diff
@@ -4,20 +4,184 @@
...
+	}
+
+	p := new(pb.EncryptionCtx)
+	if err := proto.Unmarshal(data, p); err != nil {
+		return nil, errs.New("unable to unmarshal encryption context: %v", err)
+	}
+
+	encCtx := NewEncryptionCtx()
+
+	if len(p.DefaultKey) > 0 {
```
egonelbre: nit: seems like we should ensure that protobuf already unserializes into the right type so we don't have to check this. e.g. There's NodeID that already does this.

lib/uplink/ex_create_bucket_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"io"
+	"os"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/lib/uplink"
+)
+
+func CreateBucketExample(ctx context.Context, satelliteAddress string, apiKey string, cfg *uplink.Config, out io.Writer) (err error) {
+	errCatch := func(fn func() error) { err = errs.Combine(err, fn()) }
```
egonelbre: nit: I would just write it inline.

lib/uplink/project.go
```diff
@@ -204,11 +195,40 @@ func (p *Project) OpenBucket(ctx context.Context, bucketName string, access *Enc
...
+}
+
+// SaltedKeyFromPassphrase returns a key generated from the given passphrase using a stable, project-specific salt
+func (p *Project) SaltedKeyFromPassphrase(ctx context.Context, passphrase string) (_ *storj.Key, err error) {
+	defer mon.Task()(&ctx)(&err)
+	salt, err := p.retrieveSalt(ctx)
+	if err != nil {
+		return nil, err
+	}
+	key, err := encryption.DeriveDefaultPassword([]byte(passphrase), salt)
```
egonelbre: Is there a reason derive isn't returning storj.Key?

lib/uplink/encryption.go
```diff
@@ -4,20 +4,184 @@
...
-type EncryptionAccess struct {
-	// Key is the base encryption key to be used for decrypting objects.
-	Key storj.Key
-	// EncryptedPathPrefix is the (possibly empty) encrypted version of the
-	// path from the top of the storage Bucket to this point. This is
-	// necessary to have in order to derive further encryption keys.
-	EncryptedPathPrefix storj.Path
+// EncryptionCtx represents an encryption context. It holds information about
+// how various buckets and objects should be encrypted and decrypted.
+type EncryptionCtx struct {
```
egonelbre: hmm... Is there a reason the name changed? ctx seems more precise but ends up being more convoluted to talk about and write. e.g. `access` vs. `encCtx` for variable names, one seems much easier to communicate.

internal/testplanet/uplink.go
zeebo: The comment is also wrong. It's not actually duplicated. I thought it would have to be, but it turns out only `testplanet` and tests that use `testplanet` ever call this.

lib/uplink/encryption.go
zeebo: @jtolds ?

lib/uplink/ex_create_bucket_test.go
jtolds: this is for documentation so i don't want to distract from the semantics

lib/uplink/project.go
zeebo: No, that's fine. It's a `[]byte` because it originally wasn't designed to be used as a key, but that changed. I also have to update some design docs.

zeebo:
I'm happy to do another round of PRs to get the names to be something everyone is happy about. I'm a bit unhappy about `encCtx` as well.

The style isn't very clear, but the examples do run: https://github.com/storj/storj/pull/2349/files#diff-f9e0f390d36aac0c32ffc9b55cb7dbc8R17

PR #2357, opened 2019-06-26
referral-manager design doc

VinozzZ:
What: 
Right now, referral program only works per satellite meaning a user on us-east-1 satellite can only refer people to join us-ease-1. 
However, it's crucial for the tardigrade branded network to be able to handle cross-satellite referrals. Therefore, we plan to have a program running in
GCP along with tardigrade satellites to handle this issue.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/design/referral-manager.md
```diff
@@ -0,0 +1,48 @@
...
+####Definition
+*Referral Link* - _A static link that contains a unique referral code_
+
+*Referral Code* - _An uuid generated by referral manager program for each user to be able to participate in our referral program_
+
+#### User Flow
+1. Sign up
+
+    After a user clicks on the create account button:
+    - Satellite sends a request to referral manager for a referral code and store that code along with the user information inside the user table.
```
jtolds: It might be nice to be more specific about where tables are. Is this the correct interpretation?
> After a user clicks on the create account button on Satellite A:
>  - Satellite A asks referral manager B for a brand new referral code. Satellite A stores that referral code in its own database along with the user information in the user table.

docs/design/referral-manager.md
```diff
@@ -0,0 +1,48 @@
...
+    - If the user signed up through a referral link, satellite will also store the referrer's referral code in user_credits table under referred_by column.
```
jtolds: if the referral link came from satellite B from user Bob, how does Bob get credit from Alice signing up on Satellite A? Doesn't the referral manager need to store credits used and credits available also?

docs/design/referral-manager.md
```diff
@@ -0,0 +1,48 @@
...
+    - After generating a unique referral code, referral manager will store that code along with the satellite id into its own database and send the code back to satellite.
```
jtolds: is this part of the first bullet point?

docs/design/referral-manager.md
VinozzZ: Yes, you are absolutely right! I should use a letters to represent different players in this flow for sure

docs/design/referral-manager.md
```diff
@@ -0,0 +1,48 @@
...
+
+2. Award referrer credits
```
jtolds: i'm confused about the goal of this section entirely

VinozzZ: Oh noooo, let me revise it and then I will reach out for review

jtolds: lol it's okay! i'm trying to imagine in my head where all the bytes are going and when

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
+
+# Title: Referral Manager
+
+## Abstract
+
+Right now, referral program only works per satellite meaning a user on us-east-1 satellite can only refer people to join us-ease-1.
```
stefanbenten: Typo _east_

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
+
+# Title: Referral Manager
+
+## Abstract
+
+Right now, referral program only works per satellite meaning a user on us-east-1 satellite can only refer people to join us-ease-1. 
+However, it's crucial for the tardigrade branded satellites to be able to handle cross-satellite referrals. Therefore, we plan to have a program running in GCP along with tardigrade satellites to handle this issue.
```
stefanbenten: It does not need to be GCP, correct? 
Let's call it party in the network instead of _program_

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
...
+
+    After new user Alice chose to register on satellite B and clicked on the create account button:
+    - Satellite B sends a request with its satellite id to referral manager for a referral code for Alice
+    - Referral manager receives the request from satellite B, generates an unique referral code 11111111, and then store that code, 11111111, along with the satellite B's address into referral manager's database 
+    - After successfully storing the referral code and satellite B's address, referral manager sends a response with the generated referral code, 11111111, back to satellite B 
+    - Satellite B receives the referral code, 11111111, from referral manager and then store it along with Alice's information into user table in satellitedb
+        - Signed up WITHOUT referral link:
+            - Satellite B will store Alice's credit into user_credit table with a null value in referred_by column
+        - Signed up WITH referral link:
+            - If user Bonnie from satellite C referred Alice with a referral link, https://C.tardigrade.io/signup/22222222
```
stefanbenten: Just ensuring, the token can be used on whatever satellite i want? How would Bonnie know the Satellite that Alice wants to use? In this example she would refer to the same satellite as she is on. Wouldnt it need to be a "general" valid sign up page, something like https://signup.tardigrade.io/signup/222222 and there Alice can chose her satellite?

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
...
+1. Will this overload satellites?
+2. Should we just have satellites communicate with each other directly?
+
+## Implementation
+
+#### node id authentication 
+_to make sure that the node id the referral manager is talking to is one of our approved satellites_
+
+````golang
+var approvedSatellites  map[storj.NodeID]satelliteAddress
```
stefanbenten: Not sure if that plays nicely with the newest changes with Opt-In. I think the referal manager should accept all satellites per default?

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
...
+
+	// status has three possible value: NoStatus=0, Default=1, Active=2.
+	field status int ( updatable )
+	// type has two possible value: FREE_CREDIT=0, REFERRAL=1
+	field type int ( updatable )
+````
+
+## Open issues (if applicable)
+1. Should we also store referral offers information in referral manager so all tardigrade satellites will have the same offer running? What technical concerns that would arise?
+2. Should we use signed message for the communication between satellites and referral manager?
```
stefanbenten: Definitely yes, else anyone can tamper referrals.

docs/design/referral-manager.md
```diff
@@ -0,0 +1,206 @@
...
+        - If a satellite address, C, is found through the referral code 22222222
+            - referral manager sends the referral code 22222222 to satellite C
+            - referral manager sends back a OK response to satellite B
+            - satellite C receives the referral code 22222222
+            - satellite C looks up the user Bonnie that's associated with the referral code 22222222
+            - satellite C injects a new entry into user_credits table for Bonnie for their earned referrer credits
+            
+## Rationale
+1. Will this overload satellites?
+2. Should we just have satellites communicate with each other directly?
```
stefanbenten: No, that creates more problems than a single "service", like our bootstrap service currently.

docs/design/referral-manager.md
VinozzZ: Yes, we are developing a generic sign up page that has satellite selection in it

docs/design/referral-manager.md
VinozzZ: I thought referral manager is not open-sourced? So I assumed that we will only have tardigrade satellites using it

egonelbre:
I just started wondering, is there a reason that the satellites couldn't communicate directly?

In this design, the referral manager basically keeps a mapping from: `referral code -> <satellite, user>`... when the referral code contained the node id, like `<satellite node id>-<user referral id>` we wouldn't need to store it centrally.

The only difference is that we would need to specify which satellites trust each other for the requests. Which would be mainly keeping a centralized list of "tardigrade satellites", which we have to have anyways.

CC: @jtolds

docs/design/referral-manager.md
VinozzZ: @egonelbre I have listed having the satellites communicating with each other directly as an question but seems like most people are against it

egonelbre: Yeah, if we intend to add more responsibilities to the referral manager then, yes, it makes sense.

jtolds:
on hold while we're figuring out new business objectives about this

docs/design/referral-manager.md
stefanbenten: Thats fair yeah. Lets wait what our further discussions result in :)

PR #2358, opened 2019-06-27
jenkins: Stop docker container after 18m

egonelbre:
What: Previously docker containers were left running for some reason. Add an explicit timeout to the container itself to ensure it gets shutdown.

Note the limit for Jenkins is still 10min, but if it doesn't properly cleanup after 10min, it will fallback to stopping the main process at 18min and killing at 20min.

Also removes some test code from kademlia.

Why:

Please describe the tests:
 - Test 1: manually tested the `timeout -k20m 18m` part.
 - Test 2:
 
### DIFF
--- a/pkg/kademlia/peer_discovery.go
+++ b/pkg/kademlia/peer_discovery.go
@@ -96,11 +96,6 @@ func (lookup *peerDiscovery) Run(ctx context.Context) (_ []*pb.Node, err error) 
 						)
 					}
 				} else {
-					if next.Id == lookup.target {
-						lookup.cond.L.Lock()
-						allDone = true
-						lookup.cond.L.Unlock()
-					}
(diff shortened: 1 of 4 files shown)


pkg/kademlia/peer_discovery.go
```diff
@@ -96,11 +96,6 @@ func (lookup *peerDiscovery) Run(ctx context.Context) (_ []*pb.Node, err error)
 						)
 					}
 				} else {
-					if next.Id == lookup.target {
-						lookup.cond.L.Lock()
-						allDone = true
-						lookup.cond.L.Unlock()
-					}
```
stefanbenten: Is this above needed here or does it fix something?

egonelbre: It was part of some testing code that I tried to make Jenkins run better, but it got accidentally committed.

PR #2359, opened 2019-06-27
jenkins: avoid using goimports and distribute load better

egonelbre:
Currently golangci-lint ends up using memory/time because of goimports (1min). This replaces goimports with using `gofmt` instead and a custom checker for the import sorting and grouping.

This also makes Build step to build for both race and no-race concurrently, which will speed up linting and distribute the load better.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

Makefile
```diff
@@ -71,6 +71,13 @@ proto: ## Rebuild protobuf files
 	go run scripts/protobuf.go install
 	go run scripts/protobuf.go generate
 
+.PHONY: build-packages
+build-packages: build-packages-race build-packages-normal ## Test docker images locally
```
ifraixedes: Aren't both targets saving the binaries in the same directory, so one overrides the other?

egonelbre: Actually, I don't care about the binaries, it's only to build the packages and cache them, such that compilation and linters run faster later on.

PR #2361, opened 2019-06-27
Add daily node space usage to satellite nodestats endpoint

rikysya:
What: Add daily node space usage to satellite nodestats endpoint

Why: To build disk space usage graph

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/nodestats.proto
```diff
@@ -22,3 +24,18 @@ message UptimeCheckResponse {
...
 
+message DailyStorageUsageRequest {
+    google.protobuf.Timestamp from = 1;
+    google.protobuf.Timestamp to = 2;
+}
+
+message DailyStorageUsageResponse {
+    message StorageUsage {
+        double at_rest_total = 2;
+        google.protobuf.Timestamp time_stamp = 3;
```
VinozzZ: I remember @wthorp mentioned that we can use gogoproto.customtype serialization 'automatically' to time.Time instead 'manually' converting through google.protobuf.Timestamp. Seems like this could be a good place to use that?

rikysya: Didn't know about it

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -195,6 +195,48 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
 	return csv, nil
 }
 
+// QueryNodeDailySpaceUsage returns slice of NodeSpaceUsage for given period
+func (db *StoragenodeAccounting) QueryNodeDailySpaceUsage(ctx context.Context, nodeID storj.NodeID, start time.Time, end time.Time) (_ []accounting.NodeSpaceUsage, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	// as entries are stored on daily basis we don't need
+	// to extract DATE from start_time
```
egonelbre: this is not necessarily true, for older entries these might get condensed into a single entry.

rikysya: So that means we have a limit on how far we can go on the timescale, right? When we start condensing results 1 month, 2 month?

rikysya: It makes more sense to store some stats on the `storagenode` itself, and be able to ask satellite for it's point of view

egonelbre: the accounting rollup is defined from `start_time` to the next entry for a particular node.

PR #2366, opened 2019-06-27
satellite/gc: add gc service and piece tracker to repair checker

navillasa:
Create garbage collection bloom filter as a part of the repair checker loop
https://storjlabs.atlassian.net/browse/V3-1746

Add config options for garbage collection on the satellite
https://storjlabs.atlassian.net/browse/V3-2032

Create a queue on the satellite that is responsible for sending the SNs their bloom filters
https://storjlabs.atlassian.net/browse/V3-2033

What:
- Added a `gc` package at `satellite/gc`, which contains the gc.Service, which runs garbage collection integrated with the repair checker, and the PieceTracker, which stores all of the filters (about which pieces are good) for each node.
- Added a `gc` config located at `satellite/gc/service.go`
- Creates bloom filters with pieces to be retained inside the repair checker.
- Adds a `gc.Service.Send` function which will send all `RetainRequests` (or filters with good piece ids) to all storage nodes.

Why:
- Finishing storage node garbage collection

Please describe the tests:
 - integration test uploads two files to a storagenode. One of the files is deleted from the satellite metainfo (but not from the storagenode). We verify that the storagenode pieceinfo db has pieces for both files. Then, we run garbage collection on the satellite (by running the checker loop). We verify that the file that was deleted from the metainfo is no longer in the storagenode pieceinfo db.
 

satellite/gc/queue.go
```diff
@@ -0,0 +1,102 @@
...
+
+// RetainInfo contains info needed for a storage node to retain important data and delete garbage data
+type RetainInfo struct {
+	Filter       *bloomfilter.Filter
+	CreationDate time.Time
+}
+
+// Queue defines the functions of the satellite's garbage collection "queue"
+// TODO: does this need to be an interface?
+type Queue interface {
```
egonelbre: Do you even need a queue at this point?

satellite/gc/config.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "time"
+
+// Config contains configurable values for garbage collection
+type Config struct {
+	Interval time.Duration `help:"how frequently garbage collection filters should be sent to storage nodes" releaseDefault:"168h" devDefault:"168h"`
+	Active   bool          `help:"set if garbage collection is actively running or not" releaseDefault:"true" devDefault:"true"`
+
+	// TODO: find out what initial number of pieces should be when creating a filter
+	InitialPieces     int64   `help:"the initial number of pieces expected for a storage node to have, used for creating a filter" releaseDefault:"" devDefault:""`
```
egonelbre: This should come from the last run or some other place. This value is different for each storagenode.

pkg/datarepair/checker/checker.go
```diff
@@ -55,10 +56,11 @@ type Checker struct {
 	Loop            sync2.Cycle
 	IrreparableLoop sync2.Cycle
 	monStats        durabilityStats
+	Garbage         *gc.Garbage
```
egonelbre: Calling this Garbage looks really weird, also order of fields

satellite/gc/queue.go
navillasa: do you mean the interface or the data type of queue? if the latter, yeah I don't think so, I think it needs a different name.

satellite/gc/config.go
navillasa: yeah but if garbage collection is starting up for the first time and there is no last run, then it needs this initial starting value for pieces.

pkg/datarepair/checker/checker.go
navillasa: 😄 I agree, it's a placeholder for now. I need a word besides queue since it's not a queue! open to suggestions. I'll fix the order of fields.

satellite/gc/queue.go
egonelbre: I mean the interface in general.

navillasa: I don't think so, I'll remove it.

satellite/gc/config.go
mobyvb: updated to use the config value (average number of pieces per storagenode - approx. 400k) for the first run of garbage collection. The first run tracks the number of pieces _actually_ on each storagenode, and those values are used in the next run instead of the config value

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,126 @@
...
+		// Delete object from metainfo service on satellite
+		err = satellite.Metainfo.Service.Delete(ctx, deletedEncPath)
+		require.NoError(t, err)
+
+		// Check that piece of the deleted object is on the storagenode
+		pieceInfo, err := targetNode.DB.PieceInfo().Get(ctx, satellite.ID(), deletedPieceID)
+		require.NoError(t, err)
+		require.NotNil(t, pieceInfo)
+
+		// Sleep is here because we need at least one full second between the next piece tracker
```
wthorp: _Why_ do we need a second?  Sleep almost always means we're doing it wrong.

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,100 @@
...
+		numPieces = pieceTracker.pieceCounts[nodeID]
+	}
+	if _, ok := pieceTracker.retainInfos[nodeID]; !ok {
+		node, err := pieceTracker.overlay.Get(ctx, nodeID)
+		if err != nil {
+			return Error.Wrap(err)
+		}
+		filter = bloomfilter.NewOptimal(numPieces, pieceTracker.falsePositiveRate)
+		pieceTracker.retainInfos[nodeID] = &RetainInfo{
+			address:      node.GetAddress(),
```
wthorp: nit:  it probably makes more sense to get the node address immediately before sending the bloom filter, rather than when it is created

satellite/gc/gc_test.go
mobyvb: In the `pieceInfo.GetPieceIDs` query, we are converting the piece creation timestamp as well as the filter creation timestamp to `datetime` in sql. This chops off all precision beyond seconds. In our test, the amount of time that elapses between piece uploads and the garbage collection loop is less than a second, meaning `datetime(piece_creation) < datetime(filter_creation)` is false unless we sleep for a second. We went with a sleep because any workaround would have been a lot of effort, and it would be very specific to this test

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,100 @@
...
+			Filter:       filter,
+			CreationDate: pieceTracker.filterCreationDate,
```
wthorp: nit: you could save a trivial amount of memory by using one creating date for all bloom filters

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+		pieceCounts:        service.lastPieceCounts,
+		overlay:            service.overlay,
+	}
+}
+
+// Send sends the piece retain requests to all storage nodes
+func (service *Service) Send(ctx context.Context, pieceTracker PieceTracker, cb func()) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	service.lastSendTime = time.Now().UTC()
```
wthorp: should this be in sendRetainRequests instead?

wthorp: or is it here to avoid using a mutex?

satellite/gc/piecetracker.go
navillasa: I'm not totally sure I understand this comment. I think we are using one creation date (set on the pieceTracker struct) for all bloom filters created during a checker iteration: https://github.com/storj/storj/pull/2366/files/b43086f0f79b71c76ed97380594c77894ce29e0a#diff-f1ff4261b7834d51a195dcfed9762328R53

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+		lastPieceCounts: make(map[storj.NodeID]int),
+	}
+}
+
+// NewPieceTracker instantiates a piece tracker
+func (service *Service) NewPieceTracker() PieceTracker {
+	// Creation date of the gc bloom filter - the storage nodes shouldn't delete any piece newer than this.
+	filterCreationDate := time.Now().UTC()
+
+	if !service.config.Active || filterCreationDate.Before(service.lastSendTime.Add(service.config.Interval)) {
```
wthorp: Isn't the result of NewPieceTracker() persisted (and thereby always be useless)?  I feel like this whole noOpPieceTracker struct is a bad idea and that this check should be elsewhere.

navillasa: It's reset here when an iteration finishes. Do you think you could explain more of your worries about noOpPieceTracker? also, where do you think a better place for the check would be? https://github.com/storj/storj/pull/2366/files/b43086f0f79b71c76ed97380594c77894ce29e0a#diff-8762ff585be3a81707f74a689d803d6cR151

satellite/gc/service.go
navillasa: I don't think it's here to avoid using a mutex. We can move it into sendRetainRequests

pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +141,18 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 
 					// reset durability stats for next iteration
 					checker.monStats = durabilityStats{}
+
+					err := checker.garbageService.Send(ctx, checker.pieceTracker, func() {
```
ifraixedes: Callbacks aren't very common in goland, however there are cases where they are the simpler solution so beforehand I want to say that this isn't about to ban callbacks, hence don't take this comment as callbacks should never be use in Go.

I see that the callback is only used for calls `gcWaitGroup.Done()` so my question is why the `garbageService.Send` has to be a asynchronous  spinning up a new goroutine and it isn't this function who spins up the goroutine?

satellite/gc/piecetracker.go
navillasa: ah, or maybe you mean this date should be set during Send instead of here? if so, I'm kind of in favor of keeping it here since the difference doesn't seem huge.

wthorp: its not important.  :D

satellite/gc/service.go
wthorp: maybe I just need a better explanation of why its necessary

satellite/gc/gc_test.go
wthorp: That's a great reason!  I would include a least a little bit of this info in the comment.

satellite/gc/service.go
navillasa: I think the idea is that we don't want pieceTracker to be tracking pieces (creating retainInfos, etc.) when it's not time for garbage collection to run. This ```if !service.config.Active || filterCreationDate.Before(service.lastSendTime.Add(service.config.Interval)) { ``` is checking if gc is due for an iteration. If not, then we use noOpPieceTracker which does nothing when the PieceTracker methods are called. An alternative is maybe adding this time check within each method and just have one pieceTracker implementation 🤷🏻‍♀️

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+	return &Service{
+		log:             log,
+		config:          config,
+		transport:       transport,
+		overlay:         overlay,
+		lastPieceCounts: make(map[storj.NodeID]int),
+	}
+}
+
+// NewPieceTracker instantiates a piece tracker
```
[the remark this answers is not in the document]
ifraixedes: After having the last scan I've realized that returning a nop tracker looks that we are going to always execute the loop in the line https://github.com/storj/storj/pull/2366/files#diff-8762ff585be3a81707f74a689d803d6cR207 so I think that's better if we don't loop over the pieces when there is no need and we could easily checking that `PieceTracker` is `nil` and this function could return `nil` when it currently returns the nop.

I would prefer to return a `nil` and an `error` with a value that it can be check that the error is because it's deactivated or it doesn't have pass enough time to run it again.

satellite/gc/service.go
navillasa: boop, turned out this did cause a race. moving it back.

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,100 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package gc
```
ifraixedes: Add a `doc.go` file to this new package with a brief description of the package.

satellite/gc/service.go
wthorp: I talked to Moby and a more in favor of allowing PieceTracker to be nil instead of using to Noop one, but this is good enough I suppose.

satellite/gc/service.go
wthorp: k

satellite/gc/service.go
```diff
@@ -0,0 +1,128 @@
...
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/transport"
+	"storj.io/storj/uplink/piecestore"
+)
+
+// Config contains configurable values for garbage collection
+type Config struct {
+	Interval time.Duration `help:"how frequently garbage collection filters should be sent to storage nodes" releaseDefault:"168h" devDefault:"10m"`
+	Active   bool          `help:"set if garbage collection is actively running or not" releaseDefault:"true" devDefault:"true"`
```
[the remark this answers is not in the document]
wthorp: the two options are equivalent

pkg/datarepair/checker/checker.go
```diff
@@ -48,37 +50,43 @@ type durabilityStats struct {
...
 	IrreparableLoop sync2.Cycle
-	monStats        durabilityStats
+	irrdb           irreparable.DB
+	repairQueue     queue.RepairQueue
+	nodestate       *ReliabilityCache
+	metainfo        *metainfo.Service
+	garbageService  *gc.Service
+	pieceTracker    gc.PieceTracker
+	gcWaitGroup     sync.WaitGroup
+	lastChecked     string
```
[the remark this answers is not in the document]
navillasa: that's fair, I think I just changed it to an order that seemed.. less arbitrary than what it was previously. but true it's not required for this PR so I could change it back.

pkg/datarepair/checker/checker.go
```diff
@@ -48,37 +50,43 @@ type durabilityStats struct {
...
 		Loop:            *sync2.NewCycle(config.Interval),
 		IrreparableLoop: *sync2.NewCycle(config.IrreparableInterval),
-		monStats:        durabilityStats{},
+		irrdb:           irrdb,
+		repairQueue:     repairQueue,
+		nodestate:       NewReliabilityCache(overlay, config.ReliabilityCacheStaleness),
+		metainfo:        metainfo,
+		garbageService:  garbageService,
+		pieceTracker:    garbageService.NewPieceTracker(),
+		gcWaitGroup:     sync.WaitGroup{},
```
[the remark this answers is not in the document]
navillasa: changed to gcService

satellite/gc/service.go
thepaul: I find that having fewer conditionals (i.e. checking in several different places whether `pieceTracker` is nil before using it) leads to code that is easier to reason about and easier to change later. The cost in this case is having multiple things that implement the same interface, which is generally an easy thing to reason about, being enforced by the compiler. I like that tradeoff.

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,100 @@
...
+type RetainInfo struct {
+	Filter       *bloomfilter.Filter
+	CreationDate time.Time
+	address      *pb.NodeAddress
+	count        int
+}
+
+// PieceTracker allows access to info about the good pieces that storage nodes need to retain
+type PieceTracker interface {
+	// Add adds a RetainInfo to the PieceTracker
```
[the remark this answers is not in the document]
navillasa: it's not a parameter, but it is created in the method. I'll change the comment though.

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Add adds a pieceID to the relevant node's RetainInfo
+func (pieceTracker *pieceTracker) Add(ctx context.Context, nodeID storj.NodeID, pieceID storj.PieceID) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	var filter *bloomfilter.Filter
+
+	// If we know how many pieces a node should be storing, use that number. Otherwise use default.
+	numPieces := int(pieceTracker.initialPieces)
```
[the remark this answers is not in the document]
navillasa: it's not also needed below in this check?
```
	if pieceTracker.pieceCounts[nodeID] > 0 {
		numPieces = pieceTracker.pieceCounts[nodeID]
	}
```

pkg/datarepair/checker/checker.go
mobyvb: We originally had this function spinning up the goroutine. The problem had to do with setting `checker.pieceTracker = checker.garbageService.NewPieceTracker()` after sending. It is `garbageService.Send` that updates the `lastSendTime` which determines whether or not to create a noop piece tracker or a normal one.
If we create a goroutine in the checker, we must wait for everything to send before we can set the next piece tracker. So if the checker loop starts iterating before all the filters are sent, it will be using an out of date piece tracker. By having the goroutine inside `Send()`, we can update the `lastSendTime`, kick off a goroutine, and return, allowing us to set the new updated piecetracker before the next iteration of the checker.
I am not sure if there is a better way to design this, but the way it is currently seemed like the best solution of the options available (in my opinion)

satellite/gc/piecetracker.go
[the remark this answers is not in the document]
thepaul: I think Ivan might mean taking lines 60 through 64, and moving them all inside the `if` statement that starts at line 65. Because if `ok` is true on line 65, then `numPieces` isn't needed at all, and we can save a microsecond or two by not doing map lookups on `pieceTracker`.

But I don't see any real harm to doing it the way you've got it.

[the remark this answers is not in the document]
navillasa: ah okay, I'll just go ahead and change it.

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+		initialPieces:      service.config.InitialPieces,
+		falsePositiveRate:  service.config.FalsePositiveRate,
+		retainInfos:        make(map[storj.NodeID]*RetainInfo),
+		pieceCounts:        service.lastPieceCounts,
+		overlay:            service.overlay,
+	}
+}
+
+// Send sends the piece retain requests to all storage nodes
+func (service *Service) Send(ctx context.Context, pieceTracker PieceTracker, cb func()) (err error) {
```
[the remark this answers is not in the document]
navillasa: I feel like at some point we decided to not do that because the satellite wouldn't need these errors, but I can't find evidence of that conversation now 😅 so I'll go ahead and add the errs.Combine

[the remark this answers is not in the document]
navillasa: hmm actually, couldn't that cause a race condition on the errList because of the goroutines?

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+	"storj.io/storj/pkg/overlay"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/transport"
+	"storj.io/storj/uplink/piecestore"
+)
+
+// Config contains configurable values for garbage collection
+type Config struct {
+	Interval time.Duration `help:"how frequently garbage collection filters should be sent to storage nodes" releaseDefault:"168h" devDefault:"10m"`
```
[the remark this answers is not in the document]
navillasa: hmm this caused a panic. I guess it's not a thing?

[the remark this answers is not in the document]
thepaul: Yeah, ParseDuration doesn't accept `d` because days aren't all the same length :D

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+		err := service.sendRetainRequests(ctx, pieceTracker, cb)
+		if err != nil {
+			service.log.Error("error sending retain infos", zap.Error(err))
+		}
+	}()
+
+	return nil
+}
+
+func (service *Service) sendRetainRequests(ctx context.Context, pieceTracker PieceTracker, cb func()) (err error) {
```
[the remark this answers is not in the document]
navillasa: The reason this makes sense is because the function is setting up a context in which to call the call back, then it does tearing down afterward (the deferred calls-- because it's important this happens after the callback).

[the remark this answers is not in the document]
navillasa: oh wait, the cb doesn't have anything to do with the context or teardown 😅 fixing now

satellite/gc/service.go
```diff
@@ -0,0 +1,127 @@
...
+			Address: retainInfo.address,
+		}
+		signer := signing.SignerFromFullIdentity(service.transport.Identity())
+
+		ps, err := piecestore.Dial(ctx, service.transport, target, log, signer, piecestore.DefaultConfig)
+		if err != nil {
+			service.log.Error(Error.Wrap(err).Error())
+			continue
+		}
+		defer func() {
```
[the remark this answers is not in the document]
navillasa: that's true! taking it out of the loop

ifraixedes:
I found a data race in the implementation, I wrote a test which detects it.

I'm working on removing it but also trying to do some refactoring for not having the callback because the reason that we have them it makes the implementation very brittle and we should try to avoid it, otherwise it will probably produce problems in the future when we have to change this part.

pkg/datarepair/checker/checker.go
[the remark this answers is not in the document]
ifraixedes: Don't change it back, it's OK, I've just wanted to know if there is some tangible order that we could write it down in the code style conventions.

satellite/gc/service.go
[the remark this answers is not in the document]
ifraixedes: I know that they are equivalent, but I wanted to know if we should go for one to have more consistent style in the code.

satellite/gc/piecetracker.go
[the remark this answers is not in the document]
ifraixedes: I saw it but it's something not visible without checking the implementation and the docs shouldn't require to look to the implementation to make them clear.

I found that it could be confused to mention something that it isn't visible.

satellite/gc/piecetracker.go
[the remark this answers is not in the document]
ifraixedes: Not harm and I didn't comment for the sake of the minor performance grain that it may produce (which is probably negligible and without profiling I would never suggest that change), mostly to keep those variables in the scope that they are used which as a reader of the code allows me to not have to retain them in memory after outside of the scope that they are used when I read the code.

Probably, I comment these things to survive with my :fish: brain

satellite/gc/service.go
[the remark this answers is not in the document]
ifraixedes: oh, my bad, I thought that days were possible.

satellite/gc/service.go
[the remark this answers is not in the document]
ifraixedes: We may return the error in the callback, however let's wait to see if we can avoid to have them.

pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +141,19 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 
 					// reset durability stats for next iteration
 					checker.monStats = durabilityStats{}
+
+					if checker.pieceTracker != nil {
+						err := checker.gcService.Send(ctx, checker.pieceTracker, func() {
+							checker.gcWaitGroup.Done()
+						})
+						if err != nil {
+							checker.logger.Sugar().Errorf("error sending from garbage service: %v", err)
```
kaloyan-raev: It would be better to use structured logging, i.e. 
```go
checker.logger.Error("error sending from garbage service", zap.Error(err))
```

pkg/datarepair/checker/checker.go
```diff
@@ -180,6 +205,16 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
 		return nil
 	}
 
+	if checker.pieceTracker != nil {
+		for _, piece := range pieces {
+			pieceID := remote.RootPieceId.Derive(piece.NodeId, piece.PieceNum)
+			err = checker.pieceTracker.Add(ctx, piece.NodeId, pieceID)
+			if err != nil {
+				checker.logger.Sugar().Debug("error adding (nodeID, pieceID) to pieceTracker (%s, %s)", piece.NodeId, pieceID)
```
kaloyan-raev: Use structured logging and don't forget the error:
```go
checker.logger.Debug("error adding piece to tracker", zap.Stringer("Node ID", piece.NodeId), zap.Stringer("Piece ID", pieceID), zap.Error(err))
```

satellite/gc/service.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// Send sends the piece retain requests to all storage nodes
+func (service *Service) Send(ctx context.Context, pieceTracker *PieceTracker, cb func()) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	service.lastSendTime = time.Now().UTC()
+
+	go func() {
```
egonelbre: Don't make unbounded number goroutines.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,129 @@
...
+		require.NoError(t, err)
+		deletedEncPath, pointerToDelete := getPointer(ctx, t, satellite, upl, "testbucket", "test/path/1")
+		var deletedPieceID storj.PieceID
+		for _, p := range pointerToDelete.GetRemote().GetRemotePieces() {
+			if p.NodeId == targetNode.ID() {
+				deletedPieceID = pointerToDelete.GetRemote().RootPieceId.Derive(p.NodeId, p.PieceNum)
+				break
+			}
+		}
+		require.NotNil(t, deletedPieceID)
```
kaloyan-raev: `storj.PieceID` is defined as a `[32]byte` array. So it can never be `nil`. `require.NotNil(t, deletedPieceID)` will always be satisfied regardless if a value has been set to `deletedPieceID` inside the loop.

Better use `require.NotZero`.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,129 @@
...
+		require.NoError(t, err)
+		_, pointerToKeep := getPointer(ctx, t, satellite, upl, "testbucket", "test/path/2")
+		var keptPieceID storj.PieceID
+		for _, p := range pointerToKeep.GetRemote().GetRemotePieces() {
+			if p.NodeId == targetNode.ID() {
+				keptPieceID = pointerToKeep.GetRemote().RootPieceId.Derive(p.NodeId, p.PieceNum)
+				break
+			}
+		}
+		require.NotNil(t, keptPieceID)
```
kaloyan-raev: `require.NotZero` here too.

satellite/gc/service.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+var (
+	// Error defines the gc service errors class
+	Error = errs.Class("gc service error")
+)
+
+// Config contains configurable values for garbage collection
+type Config struct {
+	Interval time.Duration `help:"how frequently garbage collection filters should be sent to storage nodes" releaseDefault:"168h" devDefault:"10m"`
```
kaloyan-raev: Let's add info what `0` means for this config.

satellite/gc/service.go
```diff
@@ -0,0 +1,136 @@
...
+	Active   bool          `help:"set if garbage collection is actively running or not" releaseDefault:"true" devDefault:"true"`
```
kaloyan-raev: How about `Enabled` instead of `Active`?

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,79 @@
...
+			numPieces = pieceTracker.pieceCounts[nodeID]
+		}
+		node, err := pieceTracker.overlay.Get(ctx, nodeID)
+		if err != nil {
+			return PieceTrackerError.Wrap(err)
+		}
+		address := node.GetAddress()
+		filter = bloomfilter.NewOptimal(numPieces, pieceTracker.falsePositiveRate)
+		pieceTracker.retainInfos[nodeID] = &RetainInfo{
+			address:      address,
```
kaloyan-raev: So we are assigning the node address to RetainInfo at the start of the process of building the bloom filters. This process can take a while and when finished the node's address may have changed.

Wouldn't it be better if we keep track of the node ID instead, and lookup the address from the overlay cache just before sending the bloom filter to the storage node?

satellite/gc/service.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Service implements the garbage collection service
+type Service struct {
+	log             *zap.Logger
+	config          Config
+	transport       transport.Client
+	overlay         overlay.DB
+	lastSendTime    time.Time
+	lastPieceCounts map[storj.NodeID]int
```
kaloyan-raev: Perhaps a topic for a separate PR, but I think we should keep track of the last piece counts in the database, so it can survive satellite restarts.

If one GC round is executed once per week, and the first one always initializes the bloom filters with 400,000 pieces then it would take at least 2 weeks to generate a bloom filter that will execute efficient garbage collection on storage nodes with millions of pieces.

It's very likely that during these 2 weeks the satellite will be restarted due to new deployment or another reason. If we don't keep this info in the DB then the bloom filter generation will start again with 400,000 pieces. This way big storage nodes may never get a good bloom filter.

thepaul:
> for not having the callback because the reason that we have them it makes the implementation very brittle and we should try to avoid it, otherwise it will probably produce problems in the future when we have to change this part.

I strongly disagree that callbacks necessarily make implementations brittle. For a case like this, where there is a context that takes nontrivial setup and closely related teardown, passing in a callback to execute within the created context is often the cleanest and most robust implementation pattern IMO.

satellite/gc/service.go
mobyvb: Could we just add a field to the node dossier or something? I think I agree, but I also think we should prioritize getting this merged first, then do the migration as a separate change

satellite/gc/piecetracker.go
mobyvb: That makes sense. We didn't want to unnecessarily make calls to the db, but I guess if enough time passes it becomes necessary

satellite/gc/service.go
mobyvb: What do you mean? 0 just means we start garbage collection over again as soon as we send rather than waiting.

satellite/gc/service.go
mobyvb: Would it be preferable to block the next iteration of checker until all the bloom filters are sent? Or is there a better way to send the filters in an async way?

satellite/gc/service.go
kaloyan-raev: This is a possible solution. I agree we can do this in a separate PR.

satellite/gc/service.go
kaloyan-raev: The description reads: "how frequently...". So if someone sets `0`, they may think GC filters will never be sent. So better document what you said above.

satellite/gc/service.go
mobyvb: https://storjlabs.atlassian.net/browse/V3-2157

satellite/gc/service.go
mobyvb: By "unbounded number goroutines", do you just mean that we could have an arbitrarily large number of retain requests in our map?

kaloyan-raev: We will have thousands of storage nodes. If we start sending bloom filter to all of them in parallel, this may hog the bandwidth capacity of the satellite and disrupt the operation of other services.

We should keep the GC process low profile. Perhaps, we should upload bloom filter just to a (configurable) handful of storage nodes in parallel.

mobyvb: it should actually be sending the bloom filters one by one, not all in parallel. The goroutine just splits the send loop off from the checker process.

egonelbre: We shouldn't block the next iteration of checker since the sending could take significant amount of time. 

The unbounded means that you shouldn't call `go` unless it's obvious how they are waited for. Currently when you call `Send` there's no way to wait for their completion internally for the service in close.

Also, if it's sending one by one, then it could end up taking a lot of time.

pkg/datarepair/checker/checker.go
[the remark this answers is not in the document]
egonelbre: The current code is really messy.

pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +141,19 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
 
 					// reset durability stats for next iteration
 					checker.monStats = durabilityStats{}
+
+					if checker.pieceTracker != nil {
```
egonelbre: This is called inside defer, which may end the checker loop prematurely with error. In that case we shouldn't send since it could end up wiping all data.

satellite/gc/service.go
mobyvb: In your view, would it be fine to use a `sync2` `Limiter` and add a config value for number of concurrent sends?

pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +141,19 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
...
+						err := checker.gcService.Send(ctx, checker.pieceTracker, func() {
+							checker.gcWaitGroup.Done()
```
egonelbre: This control should be somewhere else and not through callbacks, these are hard to verify for correctness.

satellite/gc/service.go
```diff
@@ -0,0 +1,166 @@
...
+	for id, retainInfo := range pieceTracker.GetRetainInfos() {
+		err := service.sendOneRetainRequest(ctx, id, retainInfo, pieceCounts)
+		if err != nil {
+			errList.Add(err)
+		}
+	}
+	return pieceCounts, errList.Err()
+}
+
+func (service *Service) sendOneRetainRequest(
```
egonelbre: Why is the `pieceCounts` passed in here?

satellite/gc/service.go
egonelbre: There shouldn't be any concurrent sends.

satellite/gc/service.go
```diff
@@ -0,0 +1,166 @@
...
+	mon.IntVal("node_piece_count").Observe(int64(retainInfo.count))
+
+	filterBytes := retainInfo.Filter.Bytes()
+	mon.IntVal("retain_filter_size_bytes").Observe(int64(len(filterBytes)))
+
+	retainReq := &pb.RetainRequest{
+		CreationDate: retainInfo.CreationDate,
+		Filter:       filterBytes,
+	}
+	return ps.Retain(ctx, retainReq)
```
egonelbre: return error not wrapped

satellite/gc/service.go
```diff
@@ -0,0 +1,166 @@
...
+		Address: retainInfo.address,
+	}
+	signer := signing.SignerFromFullIdentity(service.transport.Identity())
+
+	ps, err := piecestore.Dial(ctx, service.transport, target, log, signer, piecestore.DefaultConfig)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = errs.Combine(err, ps.Close())
```
egonelbre: error not wrapped

satellite/gc/service.go
mobyvb: If there aren't concurrent sends, then we will have to be okay with it taking a lot of time because we are sending one by one.

satellite/gc/service.go
```diff
@@ -0,0 +1,166 @@
...
+	}()
+
+	pieceCounts[id] = retainInfo.count // save count for next bloom filter generation
+	mon.IntVal("node_piece_count").Observe(int64(retainInfo.count))
+
+	filterBytes := retainInfo.Filter.Bytes()
+	mon.IntVal("retain_filter_size_bytes").Observe(int64(len(filterBytes)))
```
egonelbre: Let's put the monitoring somewhere else, there's no need for them to be here.

ifraixedes:
> I strongly disagree that callbacks necessarily make implementations brittle. For a case like this, where there is a context that takes nontrivial setup and closely related teardown, passing in a callback to execute within the created context is often the cleanest and most robust implementation pattern IMO.

@thepaul it isn't the callback which makes the implementation brittle it's the way in how they are use to sync things. I should have pointed before the comment where I based my comment https://github.com/storj/storj/commit/498767b1077387863e107dcadee27c37e7ccaae5#r34285090

satellite/gc/service.go
mobyvb: Except in a testing environment, I do not think we will need to worry about creating/sending filters multiple times in the same window, as garbage collection is configured for once a week right now

thepaul:
@ifraixedes OK, fair enough.

satellite/gc/service.go
egonelbre: I mean there shouldn't be two concurrent sending of piecetracker info. The retain requests should still happen in parallel.

egonelbre: We shouldn't use timing for correctness, this causes subtle bugs. E.g. someone in the future tries the configuration with 1h.

satellite/gc/service.go
```diff
@@ -0,0 +1,166 @@
...
+		lastPieceCounts: lastPieceCounts,
+	}
+}
+
+// NewPieceTracker instantiates a piece tracker
+func (service *Service) NewPieceTracker() *PieceTracker {
+	// Creation date of the gc bloom filter - the storage nodes shouldn't delete any piece newer than this.
+	filterCreationDate := time.Now().UTC()
+
+	if !service.config.Enabled || filterCreationDate.Before(service.lastSendTime.Add(service.config.Interval)) {
```
egonelbre: This enabling and disabling is not that obvious.

pkg/datarepair/checker/checker.go
```diff
@@ -127,8 +141,19 @@ func (checker *Checker) IdentifyInjuredSegments(ctx context.Context) (err error)
...
+						})
+						if err != nil {
+							checker.logger.Sugar().Errorf("error sending from garbage service: %v", err)
+						}
+					}
+					checker.pieceTracker = checker.gcService.NewPieceTracker()
 				}
 			}()
+			checker.gcWaitGroup.Add(1)
```
egonelbre: the call to `gcWaitGroup.Wait` will deadlock if there is no piecetracker or pieces

satellite/gc/service.go
ifraixedes: Because it's the way to update map without having mutex, as it's created by its caller https://github.com/storj/storj/pull/2366/files/2d6ce291e8cb8bf8ddc790c14718455539e2a149#diff-f1ff4261b7834d51a195dcfed9762328R104

satellite/gc/service.go
mobyvb: Ok so what about the following implementation?
1. Create `sendLimiter`, which can be configured to do some number of concurrent retain request sends
2. Have some cancellable context attached to the garbage collection service. For every retain request we send, we pass this context into `sendLimiter.Go` along with the function for sending the retain request.
3. When we attempt to send all retain requests, we first cancel the context passed into the limiter (just in case we are currently sending any retain requests from a previous call). We update the service with a new cancellable context before adding extra goroutines to the limiter

satellite/gc/service.go
egonelbre: Why does it have to be updated inside of the sendone in the first place?

ifraixedes: yes, right, it could be in the caller.

I didn't realize when I did the refactoring.

satellite/gc/service.go
egonelbre: The logic for looping should be pulled out of checker, there should be registration, proper observers, joining into looping etc. To clean this up.

satellite/gc/service.go
egonelbre: Why does it have to be updated in this method?

ifraixedes: As I said, the caller can do the update, that was my mistake when I refactored the code.

egonelbre: Sorry, I missed the second comment, github didn't auto-refresh for some reason.

mobyvb:
closed in favor of https://github.com/storj/storj/pull/2577

PR #1773, opened 2019-04-17
Metainfo Database Schema Draft

egonelbre:
This is a metainfo database schema draft for reviewing and discussion.

It is not intended to be merged.

PR #2369, opened 2019-06-28
add bucket metadata table in SA masterDB

JessicaGreben:
What/Why: 
This PR is the first part of the [ticket](https://storjlabs.atlassian.net/browse/V3-1967) to move bucket metadata into the satellite masterDB.  We want to move bucket metadata out of the pointerDB since its been a source of bugs and more complex code treating buckets as objects. This PR simply adds the bucket metadata table in the masterDB. I'm breaking this ticket into many small PRs so they don't get too big.

Please describe the tests:
- added postgres migration test
 
none atm


satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -730,3 +730,38 @@ read count (
...
+
+//--- bucket metadata table ---//
+
+model bucket_metadata (
+	key project_id name
+
+	field project_id		blob
+	field name  			blob
+
+	field segment_size 		int
```
jtolds: can we name this preferred or default segment size? just to indicate that no matter what this value is, not all segments are necessarily that size?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -730,3 +730,38 @@ read count (
...
+
+	field path_cipher 		int
+	field enc_cipher   		int
+	field enc_blocksize		int
+
+	field rs_algorithm 		int
```
mniewrzal: Maybe we can drop DBX here because using `int` for all fields looks to be wasteful. @egonelbre make initial design for those fields and it looks that field types can be much smaller https://github.com/storj/storj/pull/1773/files#diff-ff21df8088ff0b306b1eb76fbbe6b85eR7

JessicaGreben: sounds good. let me see if i can update that

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -730,3 +730,38 @@ read count (
 )
 
 create user_credit ()
+
+//--- bucket metadata table ---//
+
+model bucket_metadata (
```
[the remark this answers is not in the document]
JessicaGreben: let me change to bucket, that is bettere

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -730,3 +730,38 @@ read count (
...
+	key project_id name
+
```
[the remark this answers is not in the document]
JessicaGreben: good idea

satellite/satellitedb/dbx/satellitedb.dbx
jtolds: my opinion is that this is metadata per bucket and this feels like premature optimization. i vote we leave these as individual columns for now out of clarity. if we need to reclaim disk space, i think bucket metadata is the last place we'll look

jtolds: we can always do a migration if we need to

egonelbre: Yeah, this was more about thinking how we will handle this for each object. Since it will be better to keep it consistent with that behavior. However, I'm fine with leaving as separate columns for now.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -730,3 +730,43 @@ read count (
...
+	key    id
+	unique name project_id
+
+	field id             blob
+	field project_id     project.id restrict
+	field name           blob
+
+	field path_cipher    int
+
+	field created_at     utimestamp ( autoinsert )
```
egonelbre: Let's use `timestamp`, because apparently this avoids some weirdness of how databases handle timezones.

PR #2375, opened 2019-06-28
Add os.Env for 'defaults' flag

mniewrzal:
What: Add support for environment variable for `--defaults` flag

Why: Now each `uplink` command needs to set `--defaults=release` for every command

Please describe the tests:
 - Test 1: none
 - Test 2: none
 
### DIFF
--- a/pkg/cfgstruct/bind.go
+++ b/pkg/cfgstruct/bind.go
@@ -264,6 +264,9 @@ func FindIdentityDirParam() string {
 
 // FindDefaultsParam returns '--defaults' param from os.Args (if it exists)
 func FindDefaultsParam() string {
+	if os.Getenv("STORJ_DEFAULTS") != "" {
+		return os.Getenv("STORJ_DEFAULTS")
+	}
 	return FindFlagEarly("defaults")
 }
 


mniewrzal:
Flag `defaults` is a persistent flag and we are reading it directly from command line args. It needs to be known before we are processing config.yaml to be able to determine default values for things not set explicitly in config.yaml. Regarding env var we have support for every all flags from config (e.g. STORJ_SATELLITE_ADDR) so I think the addtional way of setting this flag will be consistent with what we have.

kaloyan-raev:
Can we process the config.yaml in two steps?
- 1. Check if `defaults` exists
- 2. Parse config.yaml based on the `defaults' value found in step 1.

mniewrzal:
maybe but it would require some development, not sure how much. this PR was just to help a little bit. we can write ticket and think about a better solution later.

mniewrzal:
As far I remeber, order is like this (from first to last): env var value, cmd flag value, config value

kaloyan-raev:
I thought we are using viper: https://github.com/spf13/viper

It says flags take precedence to env.

mniewrzal:
Oh, so maybe I was wrong. I will close it for now and think about it later.

PR #2376, opened 2019-06-28
Set correct default erasure share size for libuplink

kaloyan-raev:
What: Sets the default erasure share size in libuplink to 256 bytes.

Why: The value was 1 KiB, which was out of sync with default uplink settings, and the validation config on the satellites. This way bucket created with libuplink had invalid redundancy configuration and all uploads were rejected by the satellite with `provided redundancy scheme parameters not allowed`.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/lib/uplink/project.go
+++ b/lib/uplink/project.go
@@ -83,7 +83,7 @@ func (cfg *BucketConfig) setDefaults() {
 		cfg.Volatile.RedundancyScheme.TotalShares = 130
 	}
 	if cfg.Volatile.RedundancyScheme.ShareSize == 0 {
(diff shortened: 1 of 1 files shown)


kaloyan-raev:
@stefanbenten Yes, we should discuss it.

But we are now damn locked to these RS settings, so at least we should make them consistent.

egonelbre:
Can we create global constants for the defaults and refer to them in the code base rather than have hardcoded values scattered around?

kaloyan-raev:
Will such constants help for the config annotations like this?
https://github.com/storj/storj/blob/master/satellite/metainfo/config.go#L25

egonelbre:
Not really, but we can keep them at least closer to each other.

PR #2380, opened 2019-06-28
Keep some unhealthy pieces in the event of a partial repair

mobyvb:
What: 
If a repair does not repair to more than the success threshold number of pieces, we continue saving unhealthy pieces that are not duplicates after the repair.

Why:
In the event of an incomplete repair, we want to keep unhealthy pieces just in case they come back online. This way we don't risk storing a segment that is too weak and doesn't have a sufficient number of pieces in it, even if the repair doesn't fully succeed 

https://storjlabs.atlassian.net/browse/V3-1885

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/storage/segments/repairer.go
```diff
@@ -67,7 +67,8 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
 	expiration := pointer.GetExpirationDate()
 
 	var excludeNodeIDs storj.NodeIDList
-	var healthyPieces []*pb.RemotePiece
+	var healthyPieces, unhealthyPieces []*pb.RemotePiece
+	healthyMap := make(map[int32]bool)
```
kaloyan-raev: I wonder if we could combine `healthyPieces` and `healthyMap` into a single `map[int32]*pb.RemotePiece`?

pkg/storage/segments/repairer.go
```diff
@@ -160,27 +164,37 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
...
 	healthyRatioAfterRepair := 0.0
 	if pointer.Remote.Redundancy.Total != 0 {
-		healthyRatioAfterRepair = float64(len(healthyPieces)) / float64(pointer.Remote.Redundancy.Total)
+		healthyRatioAfterRepair = float64(healthyLength) / float64(pointer.Remote.Redundancy.Total)
 	}
 	mon.FloatVal("healthy_ratio_after_repair").Observe(healthyRatioAfterRepair)
 
+	// if partial repair, include "unhealthy" pieces that are not duplicates
+	if healthyLength < pointer.Remote.Redundancy.SuccessThreshold {
+		for _, p := range unhealthyPieces {
```
kaloyan-raev: Can we prioritize offline nodes over disqualified nodes? If not easy to do right now, let's at least add a TODO comment.

PR #2383, opened 2019-06-28
add bucket metainfo rpc

JessicaGreben:
What/Why:

This PR is one part of the [ticket](https://storjlabs.atlassian.net/browse/V3-1967) to move bucket metadata into the satellite masterDB. I'm breaking this ticket into many small PRs so they don't get too big. 
We want to move bucket metadata out of the pointerDB since its been a source of bugs and more complex code treating buckets as objects. This PR simply adds the bucket metadata RPCs for interacting with buckets. 

Ref: see design doc here [nv/metainfo-rpc-doc](https://github.com/storj/storj/pull/1957)

Please describe the tests:
- n/a

- none atm


pkg/pb/encryption.proto
```diff
@@ -0,0 +1,24 @@
...
+
+import "gogo.proto";
+
+message EncryptionParameters {
+  CipherSuite cipher_suite = 1;
+  int64 block_size = 2;
+}
+
+message CipherSuite {
+  enum Type {
```
egonelbre: This is Cipher, not CipherSuite

egonelbre:
The protobuf should look something like this:

``` protobuf

service Metainfo {
    rpc CreateBucket(BucketCreateRequest) returns (BucketCreateResponse);
    rpc GetBucket(BucketGetRequest) returns (BucketGetResponse);
    rpc DeleteBucket(BucketDeleteRequest) returns (BucketDeleteResponse);
    rpc ListBuckets(BucketListRequest) returns (BucketListResponse);
    rpc SetBucketAttribution(BucketSetAttributionRequest) returns (BucketSetAttributionResponse);

    ...
}

message Bucket {
    bytes        name = 1;
    CipherSuite  path_cipher = 2;
    bytes        attribution_id = 3;

    google.protobuf.Timestamp created_at = 4;

    int64                default_segment_size = 5;
    RedundancyScheme     default_redundancy_scheme = 6;
    EncryptionParameters default_encryption_parameters = 7;
}

message BucketCreateRequest {
    bytes        name = 1;
    CipherSuite  path_cipher = 2;
    bytes        attribution_id = 3;

    int64                default_segment_size = 4;
    RedundancyScheme     default_redundancy_scheme = 5;
    EncryptionParameters default_encryption_parameters = 6;
}

message BucketCreateResponse {
    Bucket bucket = 1;
}

message BucketGetRequest {
    bytes name = 1;
}

message BucketGetResponse {
    Bucket bucket = 1;
}

message BucketDeleteRequest {
    bytes name = 1;
}
message BucketDeleteResponse {}

message BucketListRequest {
    bytes     cursor    = 1;
    int32     limit     = 2;
}

message BucketListResponse {
    repeated Bucket items = 1;
    bool            more  = 2;
}

message BucketSetAttributionRequest {
    bytes name = 1;
    bytes attribution_id = 2;
}

message BucketSetAttributionResponse {
}
```

JessicaGreben:
@egonelbre for the Bucket message why only a name field and not a project ID or a uuid for the bucket?

egonelbre:
Because bucket uuid and project id are completely internal to how satellite internally implements things.

Also, for requests the project id comes from the api key.

JessicaGreben:
isn't the bucket name just a non-unique string?

egonelbre:
For a given project it must be unique.

JessicaGreben:
oh ok so the RPC Bucket message won't have a project ID field, but in the database on the SA, they will store the project ID in the bucket table?

pkg/pb/encryption.proto
JessicaGreben: ok i changed to Cipher then. Whats the difference? it looks like in our codebase we use both

egonelbre: https://storjlabs.atlassian.net/browse/V3-2006

egonelbre: Also there is a difference in what the enumerated values are.

pkg/pb/metainfo.proto
```diff
@@ -24,6 +31,65 @@ service Metainfo {
     rpc ProjectInfo(ProjectInfoRequest) returns (ProjectInfoResponse);
 }
 
+message BucketItem {
```
JessicaGreben: there is already a `Bucket` so I changed the namee here. Re: https://github.com/storj/storj/blob/master/pkg/pb/inspector.proto#L116

pkg/pb/encryption.proto
jtolds: oh, egon is saying we *want* ciphersuite (and its value definitions) and we *don't want* cipher

pkg/pb/metainfo.proto
```diff
@@ -24,6 +31,72 @@ service Metainfo {
...
+    google.protobuf.Timestamp created_at = 4;
+
+    int64                           default_segment_size = 5;
+    pointerdb.RedundancyScheme      default_redundancy_scheme = 6;
+    encryption.EncryptionParameters default_encryption_parameters = 7;
+}
+
+message BucketListItem {
+    bytes             name = 1;
+    bytes             attribution_id = 2;
```
jtolds: why attribution_id in a bucket list? i don't think we need that in a list

JessicaGreben: when we list buckets i thought we would want to see which ones are attribution buckets but i really don't know. I can go either way.  @egonelbre @mniewrzal thoughts?

pkg/pb/encryption.proto
JessicaGreben: _updated_

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/jtolds/monkit-hw v0.0.0-20190108155550-0f753668cf20
 	github.com/klauspost/cpuid v0.0.0-20180405133222-e7e905edc00e // indirect
 	github.com/klauspost/reedsolomon v0.0.0-20180704173009-925cb01d6510 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
mniewrzal: is this part of this change?

go.sum
```diff
@@ -203,6 +203,8 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kshvakov/clickhouse v1.3.4/go.mod h1:DMzX7FxRymoNkVgizH0DWAL8Cur7wHLgx3MUnGwJqpE=
+github.com/kylelemons/godebug v1.1.0 h1:RPNrshWIDI6G2gRW9EHilWtl7Z6Sb1BR0xunSBf0SNc=
+github.com/kylelemons/godebug v1.1.0/go.mod h1:9/0rRGxNHcop5bhtWyNeEfOS8JIWk580+fNqagV/RAw=
```
mniewrzal: is this part of this change?

pkg/pb/metainfo.proto
```diff
@@ -24,6 +31,72 @@ service Metainfo {
     rpc ProjectInfo(ProjectInfoRequest) returns (ProjectInfoResponse);
 }
 
+message BucketItem {
+    bytes             name = 1;
+    encryption.Cipher path_cipher = 2;
+    bytes             attribution_id = 3;
+
+    google.protobuf.Timestamp created_at = 4;
```
mniewrzal: use gogo to have pure `time.Time` `google.protobuf.Timestamp created_at = 3 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];`

pkg/pb/metainfo.proto
```diff
@@ -24,6 +31,72 @@ service Metainfo {
...
+    int64                           default_segment_size = 5;
+    pointerdb.RedundancyScheme      default_redundancy_scheme = 6;
+    encryption.EncryptionParameters default_encryption_parameters = 7;
+}
+
+message BucketListItem {
+    bytes             name = 1;
+    bytes             attribution_id = 2;
+
+    google.protobuf.Timestamp created_at = 3;
```
mniewrzal: use gogo to have pure `time.Time` `google.protobuf.Timestamp created_at = 3 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];`

pkg/pb/metainfo.proto
mniewrzal: I guess `attribution_id` can be obtained with `GetBucket` if it will be necessary.

go.mod
JessicaGreben: not sure where that came from, but I removed it

pkg/pb/encryption.proto
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+syntax = "proto3";
+option go_package = "pb";
+
+package encryption;
+
+import "gogo.proto";
```
mniewrzal: this import is not used

pkg/pb/encryption.proto
```diff
@@ -0,0 +1,23 @@
...
+option go_package = "pb";
+
+package encryption;
+
+message EncryptionParameters {
+  CipherSuite cipher_suite = 1;
+  int64 block_size = 2;
+}
+
+message CipherSuite {
```
egonelbre: Ah also, this probably could just be `enum CipherSuite {` no need for the extra nesting. I think.

satellite/metainfo/metainfo.go
```diff
@@ -599,3 +599,38 @@ func (endpoint *Endpoint) ProjectInfo(ctx context.Context, req *pb.ProjectInfoRe
 		ProjectSalt: salt[:],
 	}, nil
 }
+
+// CreateBucket creates a bucket
+func (endpoint *Endpoint) CreateBucket(ctx context.Context, req *pb.BucketCreateRequest) (_ *pb.BucketCreateResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+	// TODO: placeholder to implement pb.MetainfoServer interface.
+	return &pb.BucketCreateResponse{}, err
```
egonelbre: Nit: these all should return errors for now. e.g. `status.Error(codes.Unimplemented, "not implemented")`

JessicaGreben: i have another PR going and i can update this

pkg/pb/encryption.proto
JessicaGreben: have another PR going and i can update this

PR #2384, opened 2019-06-28
satellite/rewards: use USD type

egonelbre:
What: Use USD type to avoid conversion mistakes between cents and dollars.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/usercredits.go
```diff
@@ -19,19 +21,19 @@ type UserCredits interface {
...
-	ReferredBy           uuid.UUID
-	CreditsEarnedInCents int
-	CreditsUsedInCents   int
-	ExpiresAt            time.Time
-	CreatedAt            time.Time
+	ID            int
+	UserID        uuid.UUID
+	OfferID       int
+	ReferredBy    uuid.UUID
+	CreditsEarned rewards.USD
```
VinozzZ: This struct is not for display purpose. Should we keep it as `rewards.Cent` ?

satellite/rewards/money.go
```diff
@@ -0,0 +1,34 @@
...
+// See LICENSE for copying information
+
+package rewards
+
+import (
+	"fmt"
+)
+
+// USD describes USD currency in cents.
+type USD struct {
```
VinozzZ: Maybe it's just me but `USD` has the implication that the unit would be dollars? Should we just use `Cents` ?

egonelbre: The comment can be probably improved. Since this is an opaque type (it doesn't show how it stores things internally), it really doesn't matter how it stores things... at least for code outside.

satellite/console/usercredits.go
egonelbre: So `USD` is not for just display, but for representing money such you cannot misconvert to wrong units. However there might be a better place to store `USD` type than package `rewards`.

VinozzZ: A-ha, I think after I read the whole PR I understand the purpose now. Maybe just to me, but `USD` equals `dollars` in terms of unit for money. So it's a little confusing for me when I read this line

satellite/marketingweb/form.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// init safely registers convertStringToTime for the decoder.
+func init() {
+	decoder.RegisterConverter(time.Time{}, convertStringToTime)
+	decoder.RegisterConverter(rewards.USD{}, convertDollarsToUSD)
+}
+
+// convertDollarsToUSD formats form time input as time.Time.
+func convertDollarsToUSD(s string) reflect.Value {
```
VinozzZ: This name feels confusing for me since `Dollars` is already in `USD`

egonelbre: this is converting dollar int to a usd type... and yeah, this similarity is unfortunate

satellite/satellitedb/dbx/satellitedb.dbx.go
```diff
@@ -18,8 +18,9 @@ import (
 
 	"github.com/lib/pq"
 
-	"github.com/mattn/go-sqlite3"
 	"math/rand"
+
+	"github.com/mattn/go-sqlite3"
```
egonelbre: why is this change here?

satellite/satellitedb/pbold/order.go
```diff
@@ -5,10 +5,11 @@ package pbold
 
 import (
 	fmt "fmt"
+	math "math"
+
 	_ "github.com/gogo/protobuf/gogoproto"
 	proto "github.com/gogo/protobuf/proto"
 	_ "github.com/golang/protobuf/ptypes/duration"
```
egonelbre: why is this file changed?

satellite/console/usercredits.go
```diff
@@ -25,15 +23,15 @@ type UserCredit struct {
 	UserID        uuid.UUID
 	OfferID       int
 	ReferredBy    uuid.UUID
-	CreditsEarned rewards.USD
-	CreditsUsed   rewards.USD
+	CreditsEarned int
+	CreditsUsed   int
```
egonelbre: Why was this changed to int? This still should be `USD`

satellite/console/usercredits.go
```diff
@@ -25,15 +23,15 @@ type UserCredit struct {
...
 }
 
 // UserCreditUsage holds information about credit usage information
 type UserCreditUsage struct {
-	Referred         int64
-	AvailableCredits rewards.USD
-	UsedCredits      rewards.USD
+	Referred         int
+	AvailableCredits int
+	UsedCredits      int
```
egonelbre: Why was this changed to int? This still should be USD?

satellite/satellitedb/usercredits.go
```diff
@@ -37,18 +35,18 @@ func (c *usercredits) GetCreditUsage(ctx context.Context, userID uuid.UUID, expi
 	for usageRows.Next() {
 
 		var (
-			usedCredit      sql.NullInt64
-			availableCredit sql.NullInt64
-			referred        sql.NullInt64
+			usedCredit      int
+			availableCredit int
+			referred        int
```
egonelbre: This needs to be null int to work properly with sqlite

satellite/console/usercredits.go
egonelbre: I mean it would be fine to do it in a separate PR, if you feel like this is getting too large.

fuskovic:
@egonelbre I reverted all my changes to your latest commit. When I originally pulled `813cc62` down it wouldn't build due to :

```
satellite/satellitedb/offers.go:63:31: cannot use awardCreditInCents (type int) as type int64 in argument to rewards.Cents
satellite/satellitedb/offers.go:64:33: cannot use inviteeCreditInCents (type int) as type int64 in argument to rewards.Cents
satellite/satellitedb/offers.go:99:51: cannot use o.AwardCredit.Cents() (type int64) as type int in argument to satellitedb.Offer_AwardCreditInCents
satellite/satellitedb/offers.go:100:55: cannot use o.InviteeCredit.Cents() (type int64) as type int in argument to satellitedb.Offer_InviteeCreditInCents
satellite/satellitedb/offers.go:181:52: cannot use offerDbx.AwardCreditInCents (type int) as type int64 in argument to rewards.Cents
satellite/satellitedb/offers.go:182:52: cannot use offerDbx.InviteeCreditInCents (type int) as type int64 in argument to rewards.Cents
satellite/satellitedb/usercredits.go:49:21: invalid operation: usage.UsedCredits += usedCredit.Int64 (mismatched types rewards.USD and int64)
satellite/satellitedb/usercredits.go:51:26: invalid operation: usage.AvailableCredits += availableCredit.Int64 (mismatched types rewards.USD and int64)
satellite/satellitedb/usercredits.go:62:69: cannot use userCredit.CreditsEarned.Cents() (type int64) as type int in argument to satellitedb.UserCredit_CreditsEarnedInCents
satellite/satellitedb/usercredits.go:175:45: cannot use userCreditDBX.CreditsEarnedInCents (type int) as type int64 in argument to rewards.Cents
satellite/satellitedb/usercredits.go:175:45: too many errors
```

So I was trying to use `int` to resolve a lot of that but there's breaking changes either way.

internal/currency/usd.go
```diff
@@ -0,0 +1,37 @@
...
+
+package currency
+
+import (
+	"fmt"
+)
+
+// USD describes USD currency.
+type USD struct {
+	cents int
```
coyle: I do find this odd

internal/currency/usd.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Dollars converts dollars to USD amount.
```
coyle: nit -- converts dollars to cents ?

internal/currency/usd.go
```diff
@@ -0,0 +1,37 @@
...
+func Dollars(dollars int) USD {
+	return USD{dollars * 100}
```
coyle: nit -- I would name this `USD{cents: dollars*100}`

internal/currency/usd_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"storj.io/storj/internal/currency"
+)
+
+func TestCentDollarString(t *testing.T) {
+	type Test struct {
+		Amount   currency.USD
+		Expected string
+	}
+
+	tests := []Test{
```
coyle: love the table driven tests

satellite/marketingweb/form.go
```diff
@@ -0,0 +1,60 @@
...
+	"time"
+
+	"github.com/gorilla/schema"
+
+	"storj.io/storj/internal/currency"
+	"storj.io/storj/satellite/rewards"
+)
+
+// parseOfferForm decodes POST form data into a new offer.
+func parseOfferForm(w http.ResponseWriter, req *http.Request) (o rewards.NewOffer, e error) {
```
coyle: nit -- don't use named return parmams

coyle: nit -- move this down below var and init declarations

satellite/marketingweb/form.go
```diff
@@ -0,0 +1,60 @@
...
+	decoder = schema.NewDecoder()
+)
+
+// init safely registers convertStringToTime for the decoder.
+func init() {
+	decoder.RegisterConverter(time.Time{}, convertStringToTime)
+	decoder.RegisterConverter(currency.USD{}, convertStringToUSD)
+}
+
+// convertStringToUSD formats form time input as time.Time.
```
coyle: I don't think this comment reflects what's happening

satellite/marketingweb/form.go
```diff
@@ -0,0 +1,60 @@
...
+	value, err := strconv.Atoi(s)
+	if err != nil {
+		// invalid decoder value
+		return reflect.Value{}
+	}
+	return reflect.ValueOf(currency.Dollars(value))
+}
+
+// convertStringToTime formats form time input as time.Time.
+func convertStringToTime(value string) reflect.Value {
```
coyle: why not return time.Time type ?

satellite/satellitedb/usercredits.go
```diff
@@ -98,17 +99,17 @@ func (c *usercredits) UpdateAvailableCredits(ctx context.Context, creditsToCharg
 			break
 		}
 
-		creditsForUpdate := credit.CreditsEarnedInCents - credit.CreditsUsedInCents
+		creditsForUpdateInCents := credit.CreditsEarnedInCents - credit.CreditsUsedInCents
```
coyle: Think I've asked this before but can this value be negative ?

internal/currency/usd.go
egonelbre: No, the "cents" is completely implementation detail, it could be changed to "microcents" without changing the public interface or comments.

satellite/satellitedb/usercredits.go
egonelbre: I do not know the answer to this question. @fuskovic / @VinozzZ?

satellite/marketingweb/form.go
egonelbre: parseOfferForm has higher importance than the var and init here.

satellite/marketingweb/form.go
egonelbre: Because that's what gorilla/scheme requires

internal/currency/usd.go
egonelbre: You can take a look at https://golang.org/src/time/time.go?s=6278:7279#L117 for something similar.

satellite/satellitedb/usercredits.go
VinozzZ: No, this value shouldn't be negative here. On line 82, we are only getting credits that has `CreditsEarnedInCents > CreditsUsedInCents`

internal/currency/usd.go
VinozzZ: > The USD (United States Dollar) is the official currency of the United States of America. The United States dollar, or U.S. dollar, is made up of 100 cents. It is represented by the symbol $ or US$ to differentiate it from other dollar-based currencies.

I think this is where the confusion comes from. `USD` is the official term for `Dollars`. If I understand correctly, I think the type here could be `Unit` ?
```golang
type Unit struct {
   cents int
}
```

coyle: I think time is different than currency. Not worth changing but is definitely odd

internal/currency/usd.go
coyle: I think this might be a language thing

coyle: cents are a specific thing microcents are not. I think it will be confusing to most American users but not a huge deal.

satellite/marketingweb/form.go
coyle: ? could you explain. This is new to me.

internal/currency/usd.go
egonelbre: The standard unit for time is seconds, however it's represented as combination of seconds and nanoseconds. So I'm not seeing a large difference in that.

The `type Cents int` has the problem that it can be misconverted easily and miscomputed.

coyle: I think maybe you're confusing the similarity of time and currency USD is a specific type of currency

coyle: It's just an odd naming convention to call something USD but it be Cents

coyle: I see your point it's just odd. I'm not great with names so will leave that to you but it is odd.

satellite/marketingweb/form.go
egonelbre: There are roughly few ideas for ordering package:

1. coherence
2. convention
3. importance
4. (I may have forgotten something)

Coherence says, that try to keep similar things close together. e.g. methods for the same type, this of course can be flipped when there is a large set of types with similar methods, that keep similar methods together. Either way the principle is about "grouping and keeping similar things close to each other".

Convention says, that order things like everything else. e.g. type definition is usually followed by `New` or other constructors, which is followed by `Close` (partly because of convention, partly because it is easier to see what to close, i.e. coherence).

Importance says that, things that matter less go to the bottom. This is about the principle of what you want people to read when they open the file. In this case the important concept is about parsing the form and then comes how it's actually implemented.

coyle: Makes sense, I have never seen this done and thought I remembered reading the convention was to group the variable initialization and  init functions at the top of the file since that was what the compiler initialized first

egonelbre: Yeah, the "initialization" is the convention part... neither is wrong. Here it seemed the init wasn't that important, hence to the bottom it went.

coyle: Thanks for explaining!

internal/currency/usd.go
egonelbre: I think this comes from the difference of the "concept what it is" and "how it is implemented". `cents` is completely an implementation detail to avoid floating point errors with calculations. The concept is some amount of currency in USD.

e.g. it also could be represented like:
```
package money

type Currency struct {
	Name string
	Short string
	Symbol string
}

var USD = Currency{"United States dollar", "USD", "$"}

type Amount struct {
	Value    Fixed100
	Currency Currency
}

type Fixed100 int64
```

PR #2386, opened 2019-06-28
design doc: path component encoding

zeebo:
in talks with a number of people on the team, it seems there is some
opportunity for coordination on how paths should be encoded so that
everyone works works well with each other. in particular, we care
about matching s3 semantics for list for unencrypted paths which
contain empty path components. that makes it hard to increment
and decrement keys for seek operations. additionally, there is
work to reduce our metadata overhead for storage of paths. this
proposal takes most of the ideas from that work and ensures that
it supports the ability to efficiently seek through prefixes of
paths.


egonelbre:
Really nice design.

PR #2387, opened 2019-06-28
Routing table antechamber should expire old nodes that failed connection

jenlij:
What: Calls antechamberRemoveNode within rt.removeNode

Why: Routing table antechamber should expire old nodes that failed connection. Added to removeNode rather than calling separately within ConnectionFailed for simplicity. But I can move it to ConnectionFailed if others think that makes more sense. In that case, I should have rt.removeNode return the error storage.ErrKeyNotFound.Has(err) and check the replacement cache and antechamber for the node in question.

Please describe the tests:
 - Test 1: Updated TestRemoveNode 
 
Please describe the performance impact: Makes the remove node method a little slower while checking the antechamber. Should be negligible. 


pkg/kademlia/antechamber.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+// AntechamberErr is the class for all errors pertaining to antechamber operations
+var AntechamberErr = errs.Class("antechamber error")
+
+// Attempts to add a node the antechamber. Only allowed in if within rt neighborhood.
```
VinozzZ: Should the function starts with `antechamberAddNode` ?

jenlij: I can do that. It's not linter required since it's not an exported method, but if it helps readability, for sure.

PR #2389, opened 2019-06-28
Integration test for garbage collection

mobyvb:
What: 
This test uploads two files to a test sim. It takes one storagenode offline, then deletes one of the files (file 1). Then, the storagenode is brought back online and the garbage collection service is run. We ensure that the storagenode has the piece that was deleted before garbage collection is run, but not after. And we ensure that the storagenode has the piece that was not deleted even after garbage collection is run.

This test is skipped for now, as garbage collection has not been implemented yet.

Why:
https://storjlabs.atlassian.net/browse/V3-2030

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// TestGarbageCollection does the following:
+// * Upload two objects
+// * Put one storagenode offline
+// * Delete one object
+// * Put the storagenode back online
+// * Trigger a bloom filter generation
+// * Check that pieces of the deleted object are deleted on the storagenode
+// * Check that pieces of the kept object are not deleted on the storagenode
```
kaloyan-raev: How about a simpler approach:
- Have only 1 storage node in testplanet instead of 5.
- Upload 2 files.
- Delete the pointer of the first file from the satellite (without using the uplink), e.g. `planet.Satellites[0].Metainfo.Service.Delete(ctx, path)`. This way the pointer is gone from the satellite, but the piece is still on the storage node.
- Trigger a bloom filter generation.
- Check that pieces of the deleted object are deleted on the storagenode.
- Check that pieces of the kept object are not deleted on the storagenode.

mobyvb: Done :)
Good idea, I didn't think about it

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,119 @@
...
+		require.NoError(t, err)
+		deletedEncPath, pointerToDelete := getPointer(ctx, t, satellite, upl, "testbucket", "test/path/1")
+		var deletedPiece *pb.RemotePiece
+		for _, p := range pointerToDelete.GetRemote().GetRemotePieces() {
+			if p.NodeId == targetNode.ID() {
+				deletedPiece = p
+				break
+			}
+		}
+		require.NotNil(t, deletedPiece)
```
kaloyan-raev: Why do we need `deletedPiece` at all? It seems lines 54-61 could be deleted.

kaloyan-raev: Same for `keptPiece`.

mobyvb: ah, didn't realize they weren't used anymore. Removed

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,103 @@
...
+		pieceInfo, err := targetNode.DB.PieceInfo().Get(ctx, satellite.ID(), deletedPieceID)
+		require.NoError(t, err)
+		require.NotNil(t, pieceInfo)
+
+		// Trigger bloom filter generation by running checker
+		err = checker.IdentifyInjuredSegments(ctx)
+		require.NoError(t, err)
+
+		// Check that piece of the deleted object is not on the storagenode
+		pieceInfo, err = targetNode.DB.PieceInfo().Get(ctx, satellite.ID(), deletedPieceID)
```
ifraixedes: should we check that the piece is actually deleted from disk? I don't know if that's possible to check.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,103 @@
...
+		err = satellite.Metainfo.Service.Delete(ctx, deletedEncPath)
+		require.NoError(t, err)
+
+		// Check that piece of the deleted object is on the storagenode
+		pieceInfo, err := targetNode.DB.PieceInfo().Get(ctx, satellite.ID(), deletedPieceID)
+		require.NoError(t, err)
+		require.NotNil(t, pieceInfo)
+
+		// Trigger bloom filter generation by running checker
+		err = checker.IdentifyInjuredSegments(ctx)
```
ifraixedes: I'm just curious, is the injured segments process, the one that will build and send the bloom filter?

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+// TestGarbageCollection does the following:
+// * Set up a network with one storagenode
+// * Upload two objects
+// * Delete one object from the metainfo service on the satellite
+// * Trigger a bloom filter generation
+// * Check that pieces of the deleted object are deleted on the storagenode
+// * Check that pieces of the kept object are not deleted on the storagenode
+func TestGarbageCollection(t *testing.T) {
+	t.Skip("Garbage collection not implemented yet.")
```
ifraixedes: If we know the ticket that it should remove this skip I would add it to the comment, so in case that ticket gets closed without removing this skip, it will be easy to see that it was forgotten the ticket was closed without removing it.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+		// Check that piece of the deleted object is on the storagenode
+		pieceInfo, err := targetNode.DB.PieceInfo().Get(ctx, satellite.ID(), deletedPieceID)
+		require.NoError(t, err)
+		require.NotNil(t, pieceInfo)
+
+		// Trigger bloom filter generation by running checker
+		err = checker.IdentifyInjuredSegments(ctx)
+		require.NoError(t, err)
+
```
aligeti: I am thinking we need to add a sleep/wait timer, why ?because

1. Assuming that the IdentifyInjuredSegments(ctx) handles/implements the bloom filters for each SN. After a bloom filter is created, when does the satellite sends that corresponding bloom filter should be sent to the SNs?
2. After the SN, receives that Bloom Filter thru endpoint, should execute it to remove the pieces accordingly. 

To do all these, my thought is we need to had a wait/sleep timer and then do the PIeceinfo.Get to check if that pieces is removed.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,103 @@
...
+// * Check that pieces of the deleted object are deleted on the storagenode
+// * Check that pieces of the kept object are not deleted on the storagenode
+func TestGarbageCollection(t *testing.T) {
+	t.Skip("Garbage collection not implemented yet.")
+
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 1,
+		Reconfigure: testplanet.Reconfigure{
+			Satellite: func(log *zap.Logger, index int, config *satellite.Config) {
+				// TODO see if we should reconfigure anything
```
navillasa: this will probably be the garbage service interval

satellite/gc/gc_test.go
navillasa: yes! you can check out the changes in pkg/datarepair/checker/checker.go here: https://github.com/storj/storj/pull/2366

mobyvb:
Closed in favor of #2366

PR #2395, opened 2019-06-29
Fix typo in sender.go

nerdatwork:
There was a typo in the file with the word "agreement" spelled wrong.


.clabot
```diff
@@ -47,6 +47,7 @@
     "VitaliiShpital",
     "isaachess",
     "azdagron",
-    "phthano"
+    "phthano",
+    "nerdatwork"
```
stefanbenten: Please remove this change. You need to be in the list before to let the check pass 👍

nerdatwork: Done.

PR #2397, opened 2019-07-01
storagenode: decline uploads when there are too many live requests

egonelbre:
Currently it's easy to overload storage nodes and crash them. This PR adds a maximum concurrent requests parameter which throttles upload requests.

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/piecestore/endpoint.go
```diff
@@ -55,6 +56,7 @@ type OldConfig struct {
 // Config defines parameters for piecestore endpoint.
 type Config struct {
 	ExpirationGracePeriod time.Duration `help:"how soon before expiration date should things be considered expired" default:"48h0m0s"`
+	MaxConcurrentRequests int           `help:"how many concurrent requests are allowed, before uploads are rejected." default:"30"`
```
kaloyan-raev: How would SNOs know what value to configure here?

Is there a way for the storage node to self-diagnose and determine if it is overloaded or not?

egonelbre: Probably with some kind of benchmarking by some other party. The issue is that it's not just about the storage node itself, but also about the network it's able to deliver.

I guess we could try monitoring bandwidth usage and when it isn't increasing anymore or load or memory usage etc. But these all are much more complicated solutions than "this is how much this storage node can serve".

stefanbenten: I think it’s totally fine to have a default, if someone notices a lot of issues he just can tune it down. We had the same mechanic in V2 and that worked fine.
About the default itself, I would recommend something in the neighborhood of 5-10 max.

egonelbre: Yeah, I'm not sure what to put as limits for now. Is there an easy way to test how much we can handle?

stefanbenten: My Server Node was handling 15-20 Requests max per second, when fully cached.
The Pi 3B+ is already overwhelmed with 3 requests.

stefanbenten: I could run a few test uploads on my local Network to see what i can come up with.

storagenode/piecestore/endpoint.go
```diff
@@ -74,6 +76,8 @@ type Endpoint struct {
 	orders      orders.DB
 	usage       bandwidth.DB
 	usedSerials UsedSerials
+
+	liveRequests int32
```
jtolds: we should make this be one of the first struct fields, so arm alignment is better guaranteed (https://golang.org/pkg/sync/atomic/#pkg-note-BUG)

kaloyan-raev: Wasn't this an issue for `int64` only?

jtolds: oh yeah good point, maybe worth testing on an arm device

storagenode/piecestore/endpoint.go
phutchins: I wonder if the storagenodes could potentially obtain data from one or more satellites to help it decide how many requests it can handle and adjust the max.

stefanbenten: With a dynamic scaling that is just adding another layer of potential issues @phutchins 
Lets make a good decision/average and update the docs accordingly

egonelbre: Let me know when you get a number from testing @stefanbenten

stefanbenten: I will test that after lunch 👍

internal/testplanet/storagenode.go
```diff
@@ -109,6 +109,8 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatelliteIDs []strin
 				StaticDir: filepath.Join(developmentRoot, "web/operator/"),
 			},
 			Storage2: piecestore.Config{
+				ExpirationGracePeriod: 0,
+				MaxConcurrentRequests: 100,
```
egonelbre: Leaving testplanet to 100, because having this lower will probably break some tests.

stefanbenten: Worth testing where the limit is, in another PR of course.

PR #2402, opened 2019-07-01
Audit tests cleanup

kaloyan-raev:
What: A few things are cleaned up in audit tests:

- Existing Cursor and Verifier are reused from the satellite instead of always creating new ones. New Verifier is created only if necessary to pass a custom config in the constructor.
- Some test code used `zap.L()` instead of testplanet's logger.
- Some tests used 6 storage nodes where only 4 would be enough
- Some tests upload 1 MiB of test data where only 8 KiB would be enough

Why: Housekeeping, faster test execution and easier maintenance

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/pkg/audit/reporter_test.go
+++ b/pkg/audit/reporter_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/zap"
 
 	"storj.io/storj/internal/memory"
 	"storj.io/storj/internal/testcontext"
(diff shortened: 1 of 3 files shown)


pkg/audit/reporter_test.go
```diff
@@ -22,7 +21,8 @@ func TestReportPendingAudits(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 0,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		err := planet.Satellites[0].Audit.Service.Close()
+		audits := planet.Satellites[0].Audit.Service
```
ifraixedes: Because `audits` isn't use for anything else than `Close()` call, I would leave this one as it was.

ifraixedes:
by the way it's very nice, several line of code got dropped :smile:

pkg/audit/reporter_test.go
kaloyan-raev: It seems I overused it here. Fixed.

PR #2404, opened 2019-07-01
Nodes should not be able to fail the same audit multiple times

wthorp:
What: 
When the node fails an audit on a piece the node should be removed from the pointer so that it is not audited for that piece again.
For containment mode, we needed to add the path to the persisted info.

Why:
Nodes should not be able to fail the same audit multiple times
https://storjlabs.atlassian.net/browse/V3-2000


pkg/audit/verifier.go
```diff
@@ -416,6 +421,11 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
 			err = errs.Combine(err, result.err)
 		}
 	}
+	//remove failed audit pieces from the pointer so as to only penalize once for failed audits
+	err = verifier.RemoveFailedPieces(ctx, stripe, report.Fails)
```
kaloyan-raev: This is not really correct. Reverify is not executed on the selected stripe, but on the stripe in the pending audit of reverified nodes. So, if a node failed reverification on the stripe in the pending audit, this does not mean that it is a bad node for the selected stripe too (which it was not really audited on).

pkg/audit/verifier.go
```diff
@@ -489,6 +499,23 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+	for i, piece := range remoteSegment.GetRemotePieces() {
+		for _, failedPiece := range failedPieces {
+			if piece.NodeId == failedPiece {
+				remoteSegment.RemotePieces = append(remoteSegment.RemotePieces[:i], remoteSegment.RemotePieces[i+1:]...)
+			}
+		}
+	}
+
+	// Update the segment pointer in the metainfo
+	return verifier.metainfo.Put(ctx, stripe.SegmentPath, stripe.Segment)
```
kaloyan-raev: I am a little bit worried about how we update the pieces in the pointer. We do the same in the repair too.

What if we the segment is deleted or repaired just before we call `metainfo.Put` here? This would really mess up the pointer and may lead to false negative audits and unfair disqualification of nodes.

Perhaps, we should introduce a `metainfo.Update` method. Let's add a TODO comment about this here, and think about it in another JIRA ticket / PR.

wthorp: https://storjlabs.atlassian.net/browse/V3-2088

pkg/audit/verifier_test.go
```diff
@@ -523,3 +525,65 @@ func TestVerifierModifiedSegment(t *testing.T) {
...
+		err := planet.Satellites[0].Audit.Service.Close()
+		require.NoError(t, err)
+
+		ul := planet.Uplinks[0]
+		testData := testrand.Bytes(8 * memory.KiB)
+
+		err = ul.Upload(ctx, planet.Satellites[0], "testbucket", "test/path", testData)
+		require.NoError(t, err)
+
+		cursor := audit.NewCursor(planet.Satellites[0].Metainfo.Service)
```
kaloyan-raev: No need to create a new Cursor. Just use the existing one: `audits.Cursor.NextStripe(ctx)`

pkg/audit/verifier_test.go
```diff
@@ -523,3 +525,65 @@ func TestVerifierModifiedSegment(t *testing.T) {
...
+		stripe, _, err := cursor.NextStripe(ctx)
+		require.NoError(t, err)
+		require.NotNil(t, stripe)
+
+		verifier := audit.NewVerifier(zap.L(),
```
kaloyan-raev: No need for a new Verifier too - use the existing: `planet.Satellites[0].Audit.Service.Verifier`

Check #2402 on how audit tests were cleaned up.

pkg/audit/verifier_test.go
```diff
@@ -523,3 +525,65 @@ func TestVerifierModifiedSegment(t *testing.T) {
...
+		require.NoError(t, err)
+
+		report, err := verifier.Verify(ctx, stripe, nil)
+		require.NoError(t, err)
+
+		require.Len(t, report.Successes, origNumPieces-1)
+		require.Len(t, report.Fails, 1)
+		require.Equal(t, report.Fails[0], nodeID)
+		require.Len(t, report.Offlines, 0)
+		require.Len(t, report.PendingAudits, 0)
```
kaloyan-raev: These should be `assert` instead of `require`.

pkg/audit/verifier_test.go
```diff
@@ -523,3 +525,65 @@ func TestVerifierModifiedSegment(t *testing.T) {
...
+		require.NotNil(t, stripe)
+
+		report, err = verifier.Verify(ctx, stripe, nil)
+		require.NoError(t, err)
+
+		//verify no failures because that segment is gone
+		require.Len(t, report.Successes, origNumPieces-1)
+		require.Len(t, report.Fails, 0)
+		require.Len(t, report.Offlines, 0)
+		require.Len(t, report.PendingAudits, 0)
```
kaloyan-raev: These should be `assert` instead of `require`.

In addition, can we also assert that none of the pieces from the `stripe.Segment` pointer is stored on `nodeID`?

pkg/audit/containment.go
```diff
@@ -34,6 +34,7 @@ type PendingAudit struct {
 	ShareSize         int32
 	ExpectedShareHash []byte
 	ReverifyCount     int32
+	Path              string
```
kaloyan-raev: Use `storj.Path` as type.

pkg/audit/containment_test.go
```diff
@@ -32,6 +32,7 @@ func TestContainIncrementAndGet(t *testing.T) {
 			ShareSize:         0,
 			ExpectedShareHash: pkcrypto.SHA256Hash(testrand.Bytes(10)),
 			ReverifyCount:     0,
+			Path:              "",
```
kaloyan-raev: I wonder if it really makes sense to explicitly initialize Path with empty value here. This will be done implicitly anyway. Same is valid of `PieceID`, `StripeIndex`, `ShareSize` and `ReverifyCount`.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -32,6 +32,7 @@ model pending_audits (
 	field share_size          int64
 	field expected_share_hash blob
 	field reverify_count      int64 ( updatable )
+	field path                text
```
kaloyan-raev: How are we going to migrate the existing entry in the `containment` table?

pkg/audit/verifier.go
```diff
@@ -177,6 +177,11 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
 	for _, pieceNum := range pieceNums {
 		failedNodes = append(failedNodes, shares[pieceNum].NodeID)
 	}
+	//remove failed audit pieces from the pointer so as to only penalize once for failed audits
+	err = verifier.RemoveFailedPieces(ctx, stripe.SegmentPath, stripe.Segment, failedNodes)
+	if err != nil {
+		verifier.log.Warn("Verify: failed to delete failed pieces", zap.Error(err))
```
kaloyan-raev: Instead of logging the error, let's return it on line 239, and combine it with the error returned on line 231.

pkg/audit/verifier.go
```diff
@@ -369,9 +374,24 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
 				}
 
 				if errs2.IsRPC(err, codes.NotFound) {
+
+					// Get the original segment pointer in the metainfo
+					oldPtr, err := verifier.metainfo.Get(ctx, pending.Path)
+					remoteSegment := oldPtr.GetRemote()
+					if err != nil || remoteSegment == nil || remoteSegment.RootPieceId != pending.PieceID {
+						ch <- result{nodeID: piece.NodeId, status: skipped}
```
kaloyan-raev: I think we should report `success` status here. The node should exit the containment mode as we cannot reverify it -- the segment it didn't respond to an audit is gone now.

pkg/audit/verifier.go
```diff
@@ -369,9 +374,24 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
 				}
 
 				if errs2.IsRPC(err, codes.NotFound) {
+
+					// Get the original segment pointer in the metainfo
+					oldPtr, err := verifier.metainfo.Get(ctx, pending.Path)
+					remoteSegment := oldPtr.GetRemote()
+					if err != nil || remoteSegment == nil || remoteSegment.RootPieceId != pending.PieceID {
```
kaloyan-raev: Note that we have the `checkIfSegmentDeleted` method for this purpose.

pkg/audit/verifier.go
```diff
@@ -369,9 +374,24 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
...
+						ch <- result{nodeID: piece.NodeId, status: skipped}
+						verifier.log.Debug("Reverify: audit source deleted before reverification", zap.Stringer("Node ID", piece.NodeId), zap.Error(err))
+						return
+					}
+					//remove failed audit pieces from the pointer so as to only penalize once for failed audits
+					err = verifier.RemoveFailedPieces(ctx, pending.Path, oldPtr, storj.NodeIDList{pending.NodeID})
+					if err != nil {
+						verifier.log.Warn("Reverify: failed to delete failed pieces", zap.Error(err))
```
kaloyan-raev: Add the Node ID to the log.

pkg/audit/verifier.go
```diff
@@ -369,9 +374,24 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
...
+					// Get the original segment pointer in the metainfo
+					oldPtr, err := verifier.metainfo.Get(ctx, pending.Path)
+					remoteSegment := oldPtr.GetRemote()
+					if err != nil || remoteSegment == nil || remoteSegment.RootPieceId != pending.PieceID {
+						ch <- result{nodeID: piece.NodeId, status: skipped}
+						verifier.log.Debug("Reverify: audit source deleted before reverification", zap.Stringer("Node ID", piece.NodeId), zap.Error(err))
+						return
+					}
+					//remove failed audit pieces from the pointer so as to only penalize once for failed audits
+					err = verifier.RemoveFailedPieces(ctx, pending.Path, oldPtr, storj.NodeIDList{pending.NodeID})
```
kaloyan-raev: There is another place where reverification fails the audit -- line 416 (hashes mismatch).

littleskunk:
One question (without reading the code)
What happens if an audit and a repair are running at the same time. The repair updates the pointer. Does the audit job overwrites it? How do we avoid concurrent writes?

kaloyan-raev:
@littleskunk We discussed it here: https://github.com/storj/storj/pull/2404#discussion_r299513716

We opened https://storjlabs.atlassian.net/browse/V3-2088 to address this.

pkg/audit/verifier.go
wthorp: will fix

pkg/audit/verifier_test.go
wthorp: Please justify this assertion.

satellite/satellitedb/dbx/satellitedb.dbx
wthorp: we're not.  we're just going to not care.  it's extraordinarily freeing.

wthorp: we will delete them all

pkg/audit/verifier.go
```diff
@@ -396,6 +409,17 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
 				ch <- result{nodeID: piece.NodeId, status: failed}
 				verifier.log.Debug("Reverify: hashes mismatch (audit failed)", zap.Stringer("Node ID", piece.NodeId),
 					zap.Binary("expected hash", pending.ExpectedShareHash), zap.Binary("downloaded hash", downloadedHash))
+				//remove failed audit pieces from the pointer so as to only penalize once for failed audits
```
mobyvb: to keep this consistent with the other "fail" case in `Reverify`, move your changes above the `ch <- result{nodeID: piece.NodeId, status: failed}` on line 409. This will also ensure that your success case (currently on line 415) properly occurs if applicable. Otherwise, you have already marked this node as failed

pkg/audit/verifier.go
```diff
@@ -489,24 +513,42 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
 	}, nil
 }
 
+// RemoveFailedPieces removes lost pieces from a pointer
+func (verifier *Verifier) RemoveFailedPieces(ctx context.Context, path string, pointer *pb.Pointer, failedPieces storj.NodeIDList) (err error) {
```
mobyvb: should we add something like
```
if path == "" {
    return nil
}
```
?

wthorp: yeah, I sort of didn't want to, but I've caved on this and added it around other path logic as well.

pkg/audit/verifier.go
wthorp: I'm not entirely sure I agree

satellite/satellitedb/dbx/satellitedb.dbx
wthorp: nevermind, we will leave the paths blank and our logic will treat things without paths the way it used to.

satellite/satellitedb/migrate.go
```diff
@@ -927,6 +927,13 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					`UPDATE nodes SET disqualified=NULL WHERE disqualified IS NOT NULL AND audit_reputation_alpha / (audit_reputation_alpha + audit_reputation_beta) >= 0.6;`,
 				},
 			},
+			{
+				Description: "Add pending audit path",
+				Version:     40,
+				Action: migrate.SQL{
+					`ALTER TABLE pending_audits ADD COLUMN path text NOT NULL DEFAULT '';`,
```
mobyvb: should we clear the table as part of the migration?

pkg/audit/verifier.go
```diff
@@ -489,24 +517,45 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
```
mobyvb: nit - `failedPieces` should be called `failedPieceNodeIDs`

pkg/audit/verifier.go
```diff
@@ -489,24 +517,45 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+	defer mon.Task()(&ctx)(&err)
+
+	if len(path) == 0 {
+		return nil
+	}
+	remoteSegment := pointer.GetRemote()
+	for i, piece := range remoteSegment.GetRemotePieces() {
+		for _, failedPiece := range failedPieces {
+			if piece.NodeId == failedPiece {
+				remoteSegment.RemotePieces = append(remoteSegment.RemotePieces[:i], remoteSegment.RemotePieces[i+1:]...)
```
mobyvb: I am not sure if we should do it this way since we are iterating over `remoteSegment.RemotePieces` while we are trying to modify it. Maybe it would be better to have a piece list called `goodPieces` that starts out empty, then while we iterate, if the piece is not failed, we append it to `goodPieces`. After the loop ends, we can say `remoteSegment.RemotePieces = goodPieces` and not have to worry about weird slice stuff inside the loop

pkg/audit/verifier.go
```diff
@@ -367,35 +372,58 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
 					verifier.log.Debug("Reverify: unknown transport error (contained)", zap.Stringer("Node ID", piece.NodeId), zap.Error(err))
 					return
 				}
-
 				if errs2.IsRPC(err, codes.NotFound) {
+					if len(pending.Path) > 0 {
```
mobyvb: Since this chunk is repeated, would it make sense to pull it out into a private function?

satellite/satellitedb/migrate.go
wthorp: I don't think we have to now that I added checks for empty paths

pkg/audit/verifier.go
wthorp: fixing

pkg/audit/verifier.go
wthorp: fixing

pkg/audit/verifier.go
wthorp: the calls to checkIfSegmentDeleted are slightly different unfortunately

pkg/audit/verifier.go
kaloyan-raev: Why don't you? Note that in this method it is perfectly OK to return both an error and a report. The caller will process the report despite the error.

pkg/audit/verifier.go
```diff
@@ -489,24 +517,53 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+	for _, piece := range remoteSegment.RemotePieces {
+		isFailed := false
+		for _, failedNode := range failedNodes {
+			if piece.NodeId == failedNode {
+				isFailed = true
+				break
+			}
+		}
+		if !isFailed {
+			newRemotePieces = append(newRemotePieces, piece)
```
kaloyan-raev: Can you replace line 534 (`isFailed = true`) with this line? Then you should be able to get rid of `isFailed` completely.

pkg/audit/verifier.go
```diff
@@ -489,24 +517,53 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+		return newPointer, err
 	}
-
-	if pointer.GetCreationDate().GetSeconds() != stripe.Segment.GetCreationDate().GetSeconds() ||
-		pointer.GetCreationDate().GetNanos() != stripe.Segment.GetCreationDate().GetNanos() {
-		return ErrSegmentDeleted.New(stripe.SegmentPath)
+	if oldPointer != nil {
+		if newPointer.GetCreationDate().GetSeconds() != oldPointer.GetCreationDate().GetSeconds() ||
+			newPointer.GetCreationDate().GetNanos() != oldPointer.GetCreationDate().GetNanos() {
+			return newPointer, ErrSegmentDeleted.New(segmentPath)
```
kaloyan-raev: Why don't we return `nil` instead of `newPointer` on error?

pkg/audit/verifier_test.go
kaloyan-raev: If this `require` fails it does not really prevent the rest of the test to execute, e.g. nothing will panic. Thus, we want it to be `assert` so the test case catches as many issues as possible. This often helps analyzing test failures.

pkg/audit/verifier_test.go
```diff
@@ -514,3 +515,59 @@ func TestVerifierModifiedSegment(t *testing.T) {
...
+		require.NoError(t, err)
+
+		//verify no failures because that segment is gone
+		assert.Len(t, report.Successes, origNumPieces-1)
+		assert.Len(t, report.Fails, 0)
+		assert.Len(t, report.Offlines, 0)
+		require.Len(t, report.PendingAudits, 0)
+
+		for _, newPiece := range stripe.Segment.GetRemote().GetRemotePieces() {
+			require.NotEqual(t, newPiece.NodeId, piece.NodeId)
```
kaloyan-raev: This should be `assert` too. We want the test case to report all mismatches.

satellite/satellitedb/migrate.go
kaloyan-raev: I'd vote for clearing the table and removing the check for empty paths. This way we simplify the production code a little bit, which is a good thing in the long-term. Who will remember why we have this check for empty paths in a few months?

If we go in the direction of clearing the `pending_audits` table, we should also set all `containment` values to `false` in the `nodes` table.

pkg/audit/verifier_test.go
wthorp: I really can't imagine wanting to fix more than one issue at once.  I feel like this is a personal preference, but I'll change it.

pkg/audit/verifier.go
wthorp: What would we do differently if we know that we failed to update the pointer?  Stop the service?  Simply log it somewhere else?  I don't understand the value.

pkg/audit/verifier_test.go
kaloyan-raev: Imagine a flaky test. When such a test fails, any bit of info is valuable. We may get such info from knowing how the rest of the asserts pass.

pkg/audit/verifier.go
wthorp: fixing

pkg/audit/verifier.go
wthorp: I'm not sure if matters, but fixing

pkg/audit/verifier_test.go
wthorp: I'm not sure if matters, but fixing

pkg/audit/verifier.go
kaloyan-raev: OK, you have a point. Let's leave it like this.

pkg/audit/verifier_test.go
kaloyan-raev: Imagine what a massive bug we would find if this check fails more than once.

pkg/audit/verifier_test.go
```diff
@@ -397,7 +398,7 @@ func TestVerifierMissingPiece(t *testing.T) {
 		report, err := audits.Verifier.Verify(ctx, stripe, nil)
 		require.NoError(t, err)
 
-		assert.Len(t, report.Successes, len(stripe.Segment.GetRemote().GetRemotePieces())-1)
+		assert.Len(t, report.Successes, origNumPieces-1)
```
kaloyan-raev: This change should be applied to all relevant tests. I've seen other tests occasionally failing because the pointer was modified.

e.g. https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/PR-2404/33/tests

satellite/satellitedb/migrate.go
wthorp: will fix

pkg/audit/verifier.go
```diff
@@ -487,23 +511,46 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
 }
 
+// RemoveFailedPieces removes lost pieces from a pointer
+func (verifier *Verifier) RemoveFailedPieces(ctx context.Context, path string, pointer *pb.Pointer, failedNodes storj.NodeIDList) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	if len(failedNodes) == 0 {
+		return nil
+	}
+	remoteSegment := pointer.GetRemote()
+	newRemotePieces := remoteSegment.RemotePieces[:0]
```
navillasa: why have `remoteSegment.RemotePieces[:0]` ?

pkg/audit/verifier.go
```diff
@@ -487,23 +511,46 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+OUTER:
```
navillasa: what does this label mean/do?

pkg/audit/verifier.go
```diff
@@ -487,23 +511,46 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
...
+				continue OUTER
+			}
+		}
+		newRemotePieces = append(newRemotePieces, piece)
+	}
+	remoteSegment.RemotePieces = newRemotePieces
+
+	// Update the segment pointer in the metainfo
+	//TODO:  update in a safe manner - https://storjlabs.atlassian.net/browse/V3-2088
+	return verifier.metainfo.Put(ctx, path, pointer)
```
navillasa: we might need to wrap this error but honestly I'm never super sure when wrapping is required lol

pkg/audit/verifier_test.go
wthorp: If you know of a place where it is not, let me know.  I was under the impression I fixed it everywhere I needed to.

pkg/audit/verifier.go
ifraixedes: It's for allowing to continue the first loop without executing the logic after the second loop (currently only one operation which is in the line 529)

pkg/audit/verifier.go
ifraixedes: To save allocations, it reuses the same slice

ifraixedes: In my opinion , modifying passed parameters should be avoided as much as possible, however you need to do so for avoiding copies, here it looks that's needed.

pkg/audit/verifier.go
```diff
@@ -176,6 +176,11 @@ func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[s
 	for _, pieceNum := range pieceNums {
 		failedNodes = append(failedNodes, shares[pieceNum].NodeID)
 	}
+	//remove failed audit pieces from the pointer so as to only penalize once for failed audits
```
ifraixedes: nit: add a space between `//` and `remove`
nit(2): Add an empty line above the comment.

pkg/audit/verifier.go
```diff
@@ -366,35 +371,54 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
...
 				if errs2.IsRPC(err, codes.NotFound) {
+					// Get the original segment pointer in the metainfo
+					oldPtr, err := verifier.checkIfSegmentDeleted(ctx, pending.Path, stripe.Segment)
+					if err != nil {
+						ch <- result{nodeID: piece.NodeId, status: success}
+						verifier.log.Debug("Reverify: audit source deleted before reverification", zap.Stringer("Node ID", piece.NodeId), zap.Error(err))
+						return
+					}
+					//remove failed audit pieces from the pointer so as to only penalize once for failed audits
+					err = verifier.RemoveFailedPieces(ctx, pending.Path, oldPtr, storj.NodeIDList{pending.NodeID})
```
ifraixedes: nit: add a space between `//` and `remove`
nit(2): Add an empty line above the comment.

pkg/audit/verifier.go
```diff
@@ -366,35 +371,54 @@ func (verifier *Verifier) Reverify(ctx context.Context, stripe *Stripe) (report
...
 				return
 			}
-
 			downloadedHash := pkcrypto.SHA256Hash(share.Data)
 			if bytes.Equal(downloadedHash, pending.ExpectedShareHash) {
 				ch <- result{nodeID: piece.NodeId, status: success}
 				verifier.log.Debug("Reverify: hashes match (audit success)", zap.Stringer("Node ID", piece.NodeId))
 			} else {
-				ch <- result{nodeID: piece.NodeId, status: failed}
+				//remove failed audit pieces from the pointer so as to only penalize once for failed audits
```
ifraixedes: Does this comment in the right place? shouldn't it be placed just above the following call to `err = verifier.RemoveFailedPieces(ctx, ....`

nit: add a space between `//` and `remove`

pkg/audit/verifier.go
```diff
@@ -487,23 +511,46 @@ func (verifier *Verifier) GetShare(ctx context.Context, limit *pb.AddressedOrder
 	}, nil
 }
 
+// RemoveFailedPieces removes lost pieces from a pointer
+func (verifier *Verifier) RemoveFailedPieces(ctx context.Context, path string, pointer *pb.Pointer, failedNodes storj.NodeIDList) (err error) {
```
ifraixedes: nit: I would make this method unexported, so far it's only used inside of the package.

pkg/audit/verifier.go
ifraixedes: So far it should be fine not to wrap it, it's used internally and the lines that call this function don't return it, they just log the error, so we'll get the line from the log and we'll know that it comes from this method.

I've seen other places in this package where wrapping the errors is more important than this one.

pkg/audit/verifier_test.go
ifraixedes: > These should be assert instead of require.

was it forgotten?

pkg/audit/verifier.go
wthorp: fixing

pkg/audit/verifier.go
navillasa: cool, thanks for the clarification!

PR #2410, opened 2019-07-01
V3-1989: Storage node database is locked for several minutes while submiting orders

ethanadams:
What: 
Remove infodb locking

Why:
Once per hour the storage node submits orders to the satellite. Meanwhile all audits, uploads and downloads are failing because the database is locked and the storage node is unable to verify the incoming orders (check against used_serials table)

This PR is part 1 of v3-1989 (remove locking).  Splitting up the DBs should be considered next.

Please describe the tests:

 - Test 1: TestFileConcurrency creates a file based infodb, inserts 1000 orders concurrently and then verifies the data. File DB uses _busy_timeout param to control concurrency and timeouts.  Timeout is currently 10 seconds.  Not sure this is the correct value yet.  Probably needs more testing.

 - Test 2: TestInMemoryConcurrency creates an in memory infodb, inserts 1000 orders concurrently and then verifies the data.  InMemory DB uses a unique name, shared cache, and a connection pool with 1 connection.
 


storagenode/storagenodedb/infodb.go
```diff
@@ -43,12 +44,22 @@ func newInfo(path string) (*InfoDB, error) {
...
 	if err != nil {
 		return nil, ErrInfo.Wrap(err)
 	}
 
-	dbutil.Configure(db, mon)
+	// Set max idle and max open to 1 to control concurrent access to the memory DB
+	// Setting max open higher than 1 results in table locked errors
+	db.SetMaxIdleConns(1)
+	db.SetMaxOpenConns(1)
+	db.SetConnMaxLifetime(-1)
```
zeebo: Setting the `MaxOpenConns` to 1 is basically the same thing as having a lock around every transaction.

I think the appropriate thing to do is to, instead of removing the `Mutex` entirely, change it to an `RWMutex` and acquire the appropriate lock. Then we can avoid the 10s busy timeout when the database is backed by a file on disk, as well.

ethanadams: These are for the in memory DB only...Which to my knowledge is only used by tests.  But maybe I missed something

littleskunk:
Test results are looking good. My database is still locked but this time my storage node will answer upload and download requests.

storagenode/storagenodedb/infodb.go
zeebo: Indeed, it is only used in tests, but it'd be nice the tests matched the semantics of what production is as close as possible. We're not actually testing that concurrent access works fine, anymore.

zeebo: I take it back. An `RWMutex` is exclusive to readers while a writer is active, which is not what we want. It's also insufficient to just lock on writes because memory mode does not have a WAL so writes and reads must be exclusive. So either we have different locks on different code paths, accept this as is, or attempt to use a file for the tests. I'd prefer not to have different code paths.

americanbanshee:
This pull request has been mentioned on **Storj Labs Community Forum: Decentralized Cloud Storage**. There might be relevant details there:

https://forum.storj.io/t/storage-node-database-is-locked/299/5

PR #2412, opened 2019-07-01
OrderLimit creation date time limit

aleitner:
What: 
Storage Nodes needs to have a limited window from order creation in which it will accept an order

Why:
We don't want uplinks to create orders and then sit on them for weeks.

The storage node needs time to cash in the order to the satellite. Storage node should only accept orders that provide it enough time to cash in to the satellite, expiration remaining should be greater than the time to submit the order. 

An order created date that the SN can configure when to accept the order by. (Default = order created an hour ago)

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/orders.proto
```diff
@@ -46,6 +46,8 @@ message OrderLimit {
     // satellites aren't necessarily discoverable in kademlia. this allows
     // a storage node to find a satellite and handshake with it to get its key.
     node.NodeAddress satellite_address = 11;
+
+    google.protobuf.Timestamp order_creation = 12;
```
VinozzZ: `google.protobuf.Timestamp order_creation = 12 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];` ?

aleitner: `gogoproto.stdtime` is a `time.Time` but we have been using `google.protobuf.Timestamp`

mniewrzal: We are replacing Timestamp with time.Time in new code. It's doesn't require additional conversation.

aleitner: changed to time.time

proto.lock
```diff
@@ -143,13 +143,7 @@
         ],
         "package": {
           "name": "node"
-        },
```
mniewrzal: you need exact version of protolock to avoid additional changes in proto.lock (`github.com/nilslice/protolock/cmd/protolock@v0.12.0`)

pkg/pb/orders.proto
```diff
@@ -46,6 +46,8 @@ message OrderLimit {
     // satellites aren't necessarily discoverable in kademlia. this allows
     // a storage node to find a satellite and handshake with it to get its key.
     node.NodeAddress satellite_address = 11;
+
+    google.protobuf.Timestamp order_creation = 12 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
egonelbre: Can you make a separate PR that changes this for all fields? Also, maybe sort it near other timestamps.

pkg/storage/ec/client_planet_test.go
```diff
@@ -164,6 +165,7 @@ func newAddressedOrderLimit(ctx context.Context, action pb.PieceAction, satellit
 		Limit:           dataSize.Int64(),
 		PieceExpiration: nil,
 		OrderExpiration: orderExpiration,
+		OrderCreation:   time.Now().UTC(),
```
egonelbre: Move it before expiration.

Same for all other places.

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	github.com/jtolds/monkit-hw v0.0.0-20190108155550-0f753668cf20
 	github.com/klauspost/cpuid v0.0.0-20180405133222-e7e905edc00e // indirect
 	github.com/klauspost/reedsolomon v0.0.0-20180704173009-925cb01d6510 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
egonelbre: remember to clean this up

mniewrzal: I noticed that it's modified when we are regenerating protobufs. Maybe need dependency there?

egonelbre: The file is marked as `+build ignore`, however when running a file it ignores that and still adds it to `go.mod` :/... I tried to fix this in https://github.com/storj/storj/pull/2207 however that confused tools even more.

aleitner: Why don't we add go.mod and go.sum to the .gitignore. Everything already pushed up will stay up and we can update the go.mod and go.sum when necessary with a `git add --force go.mod go.sum`

egonelbre: If you add it to `.gitignore` it will still track changes to `go.mod`. You can use `git update-index --assume-unchanged`, but that is more annoying to use.

storagenode/piecestore/verification.go
```diff
@@ -42,7 +42,8 @@ func (endpoint *Endpoint) VerifyOrderLimit(ctx context.Context, limit *pb.OrderL
 		return ErrProtocol.New("piece expired: %v", limit.PieceExpiration)
 	case endpoint.IsExpired(limit.OrderExpiration):
 		return ErrProtocol.New("order expired: %v", limit.OrderExpiration)
-
+	case time.Now().UTC().Sub(limit.OrderCreation) > endpoint.config.OrderLimitGracePeriod:
```
kaloyan-raev: I don't understand the overuse of `UTC()` in our code base.

AFAIK, for every `t time.Time`, it is true that `t.Equal(t.UTC())`: https://play.golang.org/p/NNLpRtVd7Lm

So this line should work the same way without calling the `UTC()` method.

aleitner: lack of knowledge leads to overuse

storagenode/orders/db_test.go
```diff
@@ -46,7 +47,9 @@ func TestOrders(t *testing.T) {
 		require.NoError(t, err)
 		require.Len(t, emptyArchive, 0)
 
-		now := ptypes.TimestampNow()
+		now := time.Now().UTC()
+		nowTimestamp, err := ptypes.TimestampProto(time.Now().UTC())
```
kaloyan-raev: How is this different than `ptypes.TimestampNow()`?

storagenode/piecestore/endpoint_test.go
```diff
@@ -368,13 +368,15 @@ func GenerateOrderLimit(t *testing.T, satellite storj.NodeID, uplink storj.NodeI
...
 	require.NoError(t, err)
+
 	orderLimit := &pb.OrderLimit{
 		SatelliteId:     satellite,
 		UplinkId:        uplink,
 		StorageNodeId:   storageNode,
 		PieceId:         pieceID,
 		Action:          action,
 		SerialNumber:    serialNumber,
+		OrderCreation:	 time.Now().UTC(),
```
kaloyan-raev: I think `UTC()` here would have no real effect on how the `pb.OrderLimit` is serialized.

storagenode/orders/db_test.go
aleitner: oh oops that's a typo they were suppose to be created from the same time

aleitner: `nowTimestamp, err := ptypes.TimestampProto(now)`

storagenode/orders/db_test.go
```diff
@@ -46,7 +47,9 @@ func TestOrders(t *testing.T) {
 		require.NoError(t, err)
 		require.Len(t, emptyArchive, 0)
 
-		now := ptypes.TimestampNow()
+		now := time.Now().UTC()
```
kaloyan-raev: One last `UTC()` to get rid of :)
Unless you have a good reason why we have it?

PR #2419, opened 2019-07-02
overlay.UpdateStats removes node from containment mode

kaloyan-raev:
What: Calling `overlay.UpdateStats` removes the node from containment mode in the same DB transaction, as if `containment.Delete` was called.

Why: https://storjlabs.atlassian.net/browse/V3-1830

Please describe the tests:
 - Test 1: TestContainUpdateStats - checks that after calling `overlay.UpdateStats`, the `Contained` flag is `false`, and there is no row in the containment table.
 

pkg/audit/containment_test.go
```diff
@@ -12,14 +12,18 @@ import (
...
 	"storj.io/storj/pkg/storj"
 )
 
 func TestContainIncrementAndGet(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 4,
+		SatelliteCount: 1, StorageNodeCount: 2,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		containment := planet.Satellites[0].DB.Containment()
+		cache := planet.Satellites[0].DB.OverlayCache()
```
ifraixedes: I would move these two declarations before the line where they are first used.

pkg/audit/containment_test.go
```diff
@@ -29,30 +33,32 @@ func TestContainIncrementAndGet(t *testing.T) {
...
 		require.True(t, audit.ErrContainedNotFound.Has(err))
 	})
 }
 
 func TestContainIncrementPendingEntryExists(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 4,
+		SatelliteCount: 1, StorageNodeCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		containment := planet.Satellites[0].DB.Containment()
```
ifraixedes: The same than before, I would move it before the first time it's used.

pkg/audit/containment_test.go
```diff
@@ -75,28 +81,31 @@ func TestContainIncrementPendingEntryExists(t *testing.T) {
...
 	})
 }
 
 func TestContainDelete(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 4,
+		SatelliteCount: 1, StorageNodeCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		containment := planet.Satellites[0].DB.Containment()
+		cache := planet.Satellites[0].DB.OverlayCache()
```
ifraixedes: Same about moving them.

pkg/audit/containment_test.go
```diff
@@ -106,31 +115,62 @@ func TestContainDelete(t *testing.T) {
...
 		require.NoError(t, err)
 		require.False(t, isDeleted)
 	})
 }
+
+func TestContainUpdateStats(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		containment := planet.Satellites[0].DB.Containment()
```
ifraixedes: Same than above moving them

pkg/audit/containment_test.go
```diff
@@ -106,31 +115,62 @@ func TestContainDelete(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// update node stats
+		_, err = cache.UpdateStats(ctx, &overlay.UpdateRequest{NodeID: info1.NodeID})
+		require.NoError(t, err)
+
+		// check contained flag set to false
+		node, err := cache.Get(ctx, info1.NodeID)
+		require.NoError(t, err)
+		require.False(t, node.Contained)
```
ifraixedes: We could use an assert because it doesn't stop the test to continue with the next check and at the end of the test we would know in one run if this fails but also the next one `require.True(t, audit.ErrContainedNotFound.Has(err))` too

pkg/audit/containment_test.go
kaloyan-raev: I initialize these vars at the beginning to have a clearer code below. If your preference is to use them inline the main code logic then I would not use them at all.

pkg/audit/containment_test.go
kaloyan-raev: I changed some requires to be asserts in the latest commit.

pkg/audit/containment_test.go
ifraixedes: I've been browsing code, I've found some places where vars were defined before they were needed and it required me to go  up to see what they were.

But I understand your point considering that this is a test.

Another note, I prefer when I have those group initialization to use a `var` block, because it makes clearer the initialization purpose.

ifraixedes:
> Why do you change all the requires to asserts?

They haven't been all of them, only those that the test can continue and if a specific test function fails on one of those asserts, the test doesn't stop and will report error on in all those asserts which fail, otherwise, you may fix the first one but after in a second run find the problem with the second one, and so on.

kaloyan-raev:
The reason is to not terminate the test immediately but to have more checks even if one of them fails. We use `require` to protect mainly from panics, or to terminate the test because the remaining would not make sense if some check fails.

@mobyvb Which assert is worrying you the most?

mobyvb:
@kaloyan-raev none of them in particular. Your explanation made sense. I think I have been writing my tests with a different methodology - I use requires everywhere except when I specifically want to check multiple things without stopping the test early, like checking a bunch of node IDs in a loop.
You can ignore my comment about the requires/asserts

PR #2420, opened 2019-07-02
Fix test-sim debug-addr error

mniewrzal:
What: Use random debug address port for uplink in test-sim. Previously it was using gateway port and it was already used. Also adds `exit 1` to `test-uplink.go` conditions, earlier it was not failing when the condition was not meet.

Error:
```
test:err                    12:40:05.366 | 2019-07-02T12:40:05.366+0200	ERROR	process/exec_conf.go:307	failed to start debug endpoints	{"error": "listen tcp 127.0.0.1:11009: bind: address already in use"}
                                         | storj.io/storj/pkg/process.cleanup.func1
                                         | 	/home/wywrzal/git/storj/pkg/process/exec_conf.go:307
                                         | github.com/spf13/cobra.(*Command).execute

```
Why: Error was annoying.

Please describe the tests:
 - Test 1: Run `make test-sim` and chek if error `bind: address already in use` occurs.
 

scripts/test-uplink.sh
```diff
@@ -25,33 +25,37 @@ random_bytes_file () {
 random_bytes_file 2x1024      "$SRC_DIR/small-upload-testfile" # create 2kb file of random bytes (inline)
 random_bytes_file 5x1024x1024 "$SRC_DIR/big-upload-testfile"   # create 5mb file of random bytes (remote)
 
-uplink --config-dir "$GATEWAY_0_DIR"  mb "sj://$BUCKET/"
+UPLINK_DEBUG_ADDR="127.0.0.1:0"
```
egonelbre: Let's disable debug-addr instead. Since this runs parallel with other tests, it may end up using up some other port that's needed.

mniewrzal: its choosing port when everything is started and its taking it only for one command but I have no problem with disabling it completely

egonelbre: Yeah, but we potentially have two of these running in parallel. although, we haven't yet enabled IPv6 tests.

mniewrzal: done

PR #2421, opened 2019-07-02
Move from allowed range to minimum for Version Control

stefanbenten:
What: Moving from an allowed list to a minimum requirement

Why: Makes releases less error prone.

Please describe the tests:
 - Test 1: already added in storj-sim and testplanet
 
Please describe the performance impact: less work, the better 👍 


internal/version/version.go
```diff
@@ -130,26 +127,9 @@ func (v Info) Proto() (*pb.NodeVersion, error) {
...
-		sVer, err := NewSemVer(subversion)
-		if err != nil {
-			return nil, err
-		}
-		versions = append(versions, *sVer)
-	}
-	return versions, err
+// matchesVersion compares and checks if the passed version is greater/equal than the minimum required version
+func matchesVersion(test SemVer, target SemVer) bool {
+	return test.Major > target.Major || (test.Major == target.Major && test.Minor > target.Minor) || (test.Minor == target.Minor && test.Patch >= target.Patch)
```
littleskunk:
```
test.Major > target.Major || 
(test.Major == target.Major && test.Minor > target.Minor) || 
(test.Minor == target.Minor && test.Patch >= target.Patch)
```

Storage node is running v0.13.5 and our minimum version is v1.13.0. -> false || false || true. The last check is not checking the major version.

internal/version/version.go
```diff
@@ -130,26 +127,9 @@ func (v Info) Proto() (*pb.NodeVersion, error) {
...
-	for _, subversion := range serviceVersions {
-		sVer, err := NewSemVer(subversion)
-		if err != nil {
-			return nil, err
-		}
-		versions = append(versions, *sVer)
-	}
-	return versions, err
+// matchesVersion compares and checks if the passed version is greater/equal than the minimum required version
+func matchesVersion(test SemVer, target SemVer) bool {
```
egonelbre: matchesVersion doesn't seem to be the right name this. Maybe `isNewerThan` or similar?

stefanbenten: Adressed :)

PR #2424, opened 2019-07-02
garbage collection endpoint on storage node

Fadila82:
What: storage node grpc method to handle garbage collection retain request. The storage node deletes pieces not contained in the bloom filter of the request that are older than retain request timestamp

Why: so that the satellite can send garbage collection retain request to the node
https://storjlabs.atlassian.net/browse/V3-1747

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+				return nil, EndpointError.Wrap(err)
+			}
+		}
+	}
+	fmt.Println("size = ", len(infos), " - count = ", count)
+	return nil, nil
+}
+
+// PieceInfo returns pieces info db
+func (endpoint *Endpoint) PieceInfo() pieces.DB {
```
mobyvb: I don't think the `PieceInfo()` or `Store()` endpoints are necessary

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+			if err != nil {
+				return nil, EndpointError.Wrap(err)
+			}
+			err = endpoint.pieceinfo.Delete(ctx, peer.ID, info.PieceID)
+			if err != nil {
+				return nil, EndpointError.Wrap(err)
+			}
+		}
+	}
+	fmt.Println("size = ", len(infos), " - count = ", count)
```
mobyvb: If you keep this log, make sure to change it to use `endpoint.log` rather than `fmt`

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Endpoint implements the garbage collector endpoint
+type Endpoint struct {
+	log       *zap.Logger
+	store     *pieces.Store
+	pieceinfo pieces.DB
+}
+
+// NewEndpoint creates a new endpoint
+func NewEndpoint(log *zap.Logger, store *pieces.Store, pieceinfo pieces.DB) *Endpoint {
```
mobyvb: any reason to make a new endpoint struct rather than just adding this to the struct in `storagenode/piecestore/endpoint.go`?

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+	"go.uber.org/zap"
+
+	"storj.io/storj/pkg/bloomfilter"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/storagenode/pieces"
+)
+
+// EndpointError defines errors class for Endpoint
+var EndpointError = errs.Class("garbage collector endpoint error")
```
kaloyan-raev: Can we keep this shorter?
- var name: `Error`
- class name: `gc error`

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, EndpointError.Wrap(err)
+	}
+
+	filter, err := bloomfilter.NewFromBytes(retainReq.GetFilter())
+	if err != nil {
+		return nil, EndpointError.Wrap(err)
+	}
+
+	infos, err := endpoint.pieceinfo.GetAll(ctx, peer.ID)
```
kaloyan-raev: Loading all pieces into memory will very likely crash the program with "out of memory" error.

storagenode/gc/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, EndpointError.Wrap(err)
+	}
+
+	count := 0
+	for _, info := range infos {
```
kaloyan-raev: We need to find a way to yield the current goroutine after each iteration of the loop. Garbage collection will be a very long process on storage nodes with lots of data. We want to avoid garbage collection to block other goroutines from serving and receiving data.

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -47,6 +47,31 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
 	return ErrInfo.Wrap(err)
 }
 
+// GetAll gets piece information by satellite id.
+func (db *pieceinfo) GetAll(ctx context.Context, satelliteID storj.NodeID) (infos []*pieces.Info, err error) {
```
kaloyan-raev: Instead of this function, it would be better to have `GetPage` function that paginates over the database.

It should take three additional params: `createdBefore time.Time` and `offset, limit int`.

Offset and limit should be use to return paginated result from the DB as described here: https://www.petefreitag.com/item/451.cfm

storagenode/gc/endpoint.go
kaloyan-raev: Perhaps, a call to https://golang.org/pkg/runtime/#Gosched would do the job.

storagenode/gc/endpoint.go
Fadila82: None that I can think of. I'm going to move it :-)

storagenode/pieces/store.go
```diff
@@ -58,6 +58,8 @@ type DB interface {
 	Add(context.Context, *Info) error
 	// Get returns Info about a piece.
 	Get(ctx context.Context, satelliteID storj.NodeID, pieceID storj.PieceID) (*Info, error)
+	// GetAll returns Info about pieces held for a satellite
```
kaloyan-raev: This comment should be fixed to match the name and functionality of the `GetPiecesID`.

storagenode/pieces/store.go
```diff
@@ -58,6 +58,8 @@ type DB interface {
...
+	GetPiecesID(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error)
```
kaloyan-raev: Hmm... Not sure, but shouldn't this be `GetPieceIDs` instead of `GetPiecesID`?

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,50 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	filter, err := bloomfilter.NewFromBytes(retainReq.GetFilter())
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	limit := 20
```
kaloyan-raev: Let's use `1000` for limit. We use this number at other places where we paginate the database. Also, 1000 piece ids are still something that can fit easily into memory (this should be 32 KB memory).

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,50 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+		}
+		for _, pieceID := range pieceIDs {
+			if !filter.Contains(pieceID) {
+				err = errs.Combine(err, endpoint.store.Delete(ctx, peer.ID, pieceID))
+				err = errs.Combine(err, endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID))
+			}
+		}
+		piecesRemaining = (len(pieceIDs) == limit)
+		offset += limit
+		runtime.Gosched()
```
kaloyan-raev: Let's add a comment on why we call `runtime.Gosched()` so it is not removed by accident in the future. The reason is that the GC process is expected to be long, we want to keep it at low priority, so other goroutines can continue serving requests.

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,50 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+			return nil, Error.Wrap(err)
+		}
+		for _, pieceID := range pieceIDs {
+			if !filter.Contains(pieceID) {
+				err = errs.Combine(err, endpoint.store.Delete(ctx, peer.ID, pieceID))
+				err = errs.Combine(err, endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID))
+			}
+		}
+		piecesRemaining = (len(pieceIDs) == limit)
+		offset += limit
```
kaloyan-raev: `offset += len(pieceIDs)` would be more accurate on the last iteration.

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,50 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+		runtime.Gosched()
+	}
+	return &pb.RetainResponse{}, Error.Wrap(err)
```
kaloyan-raev: If we return an error here, I think it will be returned as "internal error" back to the satellite. I am not sure if the satellite is interested if the storage node failed to collect some of the garbage... Perhaps, we should focus on logging for the SNO.

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,50 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+
+	for piecesRemaining {
+		pieceIDs, err := endpoint.pieceinfo.GetPiecesID(ctx, peer.ID, retainReq.GetCreationDate(), limit, offset)
+		if err != nil {
+			return nil, Error.Wrap(err)
+		}
+		for _, pieceID := range pieceIDs {
+			if !filter.Contains(pieceID) {
+				err = errs.Combine(err, endpoint.store.Delete(ctx, peer.ID, pieceID))
+				err = errs.Combine(err, endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID))
```
kaloyan-raev: This combining of errors can be dangerous. As there might be millions of piece IDs to be inspected, if thousands of them must be deleted and they fail with an error, then the combined error may become huge and crash the storage node with out of memory. Perhaps, we should just log the errors as they come.

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -47,6 +47,31 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
...
+// GetPiecesID gets piece information by satellite id.
+func (db *pieceinfo) GetPiecesID(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	rows, err := db.db.QueryContext(ctx, db.Rebind(`
+		SELECT piece_id
+		FROM pieceinfo
+		WHERE satellite_id = ? AND piece_creation < ?
+		LIMIT ? OFFSET ?
+	`), satelliteID, createdBefore, limit, offset)
```
kaloyan-raev: We should subtract an extra hour from `createdBefore` so we leave some room for clock difference between the satellite and the storage node. I believe we had this recommendation in the design doc.

storagenode/piecestore/endpoint.go
navillasa: does this apply only to this line or throughout this function?

storagenode/piecestore/endpoint.go
navillasa: I have these two lines logging instead of combining now, but not sure if the other calls in the function (`GetPieceIDs`, `bloomfilter.NewFromBytes`, and `identity.PeerIdentityFromContext`) should have returned errors or logged errors too.

storagenode/storagenodedb/pieceinfo.go
navillasa: Fadila said this in slack but in case anyone is here wondering, it looks like subtracting one hour isn't necessary because the creation time is fixed by the satellite in order limits

storagenode/piecestore/endpoint.go
```diff
@@ -551,6 +553,56 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+	return &pb.RetainResponse{}, nil
+}
+
+// PieceInfo returns pieces info db
+func (endpoint *Endpoint) PieceInfo() pieces.DB {
+	return endpoint.pieceinfo
+}
+
+// Store returns pieces store
+func (endpoint *Endpoint) Store() *pieces.Store {
```
mobyvb: Since we are not using this function anywhere, let's remove it

storagenode/piecestore/endpoint.go
```diff
@@ -552,6 +554,46 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+	piecesRemaining := true
+
+	for piecesRemaining {
+		pieceIDs, err := endpoint.pieceinfo.GetPieceIDs(ctx, peer.ID, retainReq.GetCreationDate(), limit, offset)
+		if err != nil {
+			return nil, Error.Wrap(err)
+		}
+		for _, pieceID := range pieceIDs {
+			if !filter.Contains(pieceID) {
+				if err = endpoint.store.Delete(ctx, peer.ID, pieceID); err != nil {
```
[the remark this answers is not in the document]
mobyvb: hmm well if we delete from the store successfully and the delete from pieceinfo fails, I think that is fine, because on the next round, the store delete will fail (since it won't exist anymore), but that doesn't block the delete from the pieceinfo from occurring and hopefully working again
However, if we delete from the pieceinfo successfully and deleting from the store fails, that is a problem because we won't ever hit this part of the code again. Maybe we should add a todo or something about this?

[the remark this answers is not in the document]
mobyvb: I think we should do the following (if it's easy):
If `endpoint.store.Delete` succeeds, or it has an error saying the piece doesn't exist (because it was already deleted), we follow through with `endpoint.pieceinfo.Delete`.
If `endpoint.store.Delete` results in some other type of error, we should not delete from pieceinfo. That way, we can attempt again on the next round of garbage collection.
We should only do this if we can easily distinguish between a "doesn't exist/was already deleted" error and other errors for `endpoint.store.Delete`

storagenode/piecestore/endpoint.go
```diff
@@ -552,6 +554,46 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+	}
+
+	filter, err := bloomfilter.NewFromBytes(retainReq.GetFilter())
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	limit := 1000
+	offset := 0
+	piecesRemaining := true
```
wthorp: nit: make name clearer that this is a bool.  EG:  hasPeicesRemaining, hasMorePieces, whatever

storagenode/piecestore/endpoint.go
```diff
@@ -552,6 +554,46 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+			if !filter.Contains(pieceID) {
+				if err = endpoint.store.Delete(ctx, peer.ID, pieceID); err != nil {
+					endpoint.log.Error("failed to delete a piece", zap.Error(Error.Wrap(err)))
+				}
+				if err = endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID); err != nil {
+					endpoint.log.Error("failed to delete piece info", zap.Error(Error.Wrap(err)))
+				}
+			}
+		}
+		piecesRemaining = (len(pieceIDs) == limit)
```
wthorp: not: limit == 0 causes infinite loop

storagenode/piecestore/endpoint.go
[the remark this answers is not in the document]
mobyvb: Ok - I looked into it and it looks like no error is returned when `endpoint.store.Delete` tries to delete something that is not on the filesystem (see `storage/filestore` `Delete()`). So we can just not delete from pieceinfo if deleting from store results in an error, and otherwise delete from pieceinfo.

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -47,6 +47,31 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
...
 
+// GetPieceIDs gets pieceIDs using the satelliteID
+func (db *pieceinfo) GetPieceIDs(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	rows, err := db.db.QueryContext(ctx, db.Rebind(`
+		SELECT piece_id
+		FROM pieceinfo
+		WHERE satellite_id = ? AND piece_creation < ?
+		LIMIT ? OFFSET ?
```
wthorp: ORDER BY is usually required to make sure that LIMIT and OFFSET are applied coherently.

wthorp: Also worth mentioning is that LIMIT and OFFSET doesn't always work all that well outside of a transaction.  Between two calls to GetPieceIDs(), a DELETE could cause you get the same Piece again.  An INSERT could cause you to skip Pieces.

storagenode/piecestore/endpoint_test.go
```diff
@@ -504,3 +513,127 @@ func GenerateOrderLimit(t *testing.T, satellite storj.NodeID, uplink storj.NodeI
...
+	storagenodedbtest.Run(t, func(t *testing.T, db storagenode.DB) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+		pieceInfos := db.PieceInfo()
+		store := pieces.NewStore(zaptest.NewLogger(t), db.Pieces())
+
+		const nbPieces = 1000
+		const nbPiecesToKeep = 990
+		const nbOldPieces = 5 // pieces from nbPiecesToKeep + nbOldPieces to nbPieces will
+		// have a recent timestamp and thus should not be deleted
```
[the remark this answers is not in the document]
navillasa: I'll change this. also, I'm not sure what `nb` stands for- do you know? considering removing it also.

storagenode/piecestore/endpoint.go
navillasa: I think it can't happen because limit is set to 1000, and it never changes, but I just made it a const to make sure it won't change.

storagenode/piecestore/endpoint.go
```diff
@@ -541,6 +543,52 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
...
+		return nil, Error.Wrap(err)
+	}
+
+	const limit = 1000
+	offset := 0
+	hasMorePieces := true
+
+	for hasMorePieces {
+		// subtract one hour to leave room for clock difference between the satellite and storage node
+		createdBefore := retainReq.GetCreationDate().Add(-1 * time.Hour)
```
egonelbre: Let's make this 1 hour a constant.

storagenode/piecestore/endpoint.go
```diff
@@ -541,6 +543,52 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
 	}
 }
 
+// Retain keeps only piece ids specified in the request
+func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainRequest) (*pb.RetainResponse, error) {
+	peer, err := identity.PeerIdentityFromContext(ctx)
+	if err != nil {
+		return nil, Error.Wrap(err)
```
egonelbre: Use correct error codes for grpc.

storagenode/piecestore/endpoint.go
```diff
@@ -541,6 +543,52 @@ func (endpoint *Endpoint) SaveOrder(ctx context.Context, limit *pb.OrderLimit, o
 	}
 }
 
+// Retain keeps only piece ids specified in the request
+func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainRequest) (*pb.RetainResponse, error) {
+	peer, err := identity.PeerIdentityFromContext(ctx)
```
egonelbre: Verify that this is a trusted satellite.

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -57,6 +57,32 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
...
 }
 
+// GetPieceIDs gets pieceIDs using the satelliteID
+func (db *pieceinfo) GetPieceIDs(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	rows, err := db.db.QueryContext(ctx, db.Rebind(`
+		SELECT piece_id
+		FROM pieceinfo
+		WHERE satellite_id = ? AND piece_creation < ?
```
egonelbre: Use `datetime` see https://github.com/storj/storj/pull/2512

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -57,6 +57,32 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
 	return ErrInfo.Wrap(err)
 }
 
+// GetPieceIDs gets pieceIDs using the satelliteID
+func (db *pieceinfo) GetPieceIDs(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error) {
```
egonelbre: If we delete nodes offset will end up wrong.

storagenode/piecestore/endpoint_test.go
```diff
@@ -17,18 +19,25 @@ import (
...
 	"storj.io/storj/internal/testrand"
 	"storj.io/storj/pkg/auth/signing"
+	"storj.io/storj/pkg/bloomfilter"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/pkcrypto"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/storagenode"
 	"storj.io/storj/storagenode/bandwidth"
+	"storj.io/storj/storagenode/pieces"
+	ps "storj.io/storj/storagenode/piecestore"
```
egonelbre: don't use alias

navillasa: it's necessary since "storj.io/storj/uplink/piecestore" is also imported

PR #2430, opened 2019-07-02
Revert "OrderLimit creation date time limit"

aleitner:
Reverts storj/storj#2412


littleskunk:
Why do we need to revert it?

Yesterday I tested master against an old satellite. To do that I had to revert it but I had the feeling that was only needed because the satellite wasn't updated. As long as we update all satellites it should work right?

egonelbre:
@littleskunk this requires also updating all the storage nodes otherwise they are unable to verify the signature.

littleskunk:
Not a big deal. The next release will contain a wipe out. Old nodes will not get any uploads, downloads or audit requests unless they update. Sounds like the perfekt time.

PR #2436, opened 2019-07-02
relax mutex in trust GetSignee

zeebo:
What: Relax a lock to a read-write one

Why: The network request in the case of a cache doesn't need to block others from being able to do reads.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/trust/service.go
+++ b/storagenode/trust/service.go
@@ -34,7 +34,7 @@ type Pool struct {
 
 // satelliteInfoCache caches identity information about a satellite
 type satelliteInfoCache struct {
-	mu       sync.Mutex
+	mu       sync.RWMutex
 	identity *identity.PeerIdentity
 }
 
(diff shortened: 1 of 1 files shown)


storagenode/trust/service.go
```diff
@@ -127,18 +127,26 @@ func (pool *Pool) GetSignee(ctx context.Context, id storj.NodeID) (_ signing.Sig
...
-	if info.identity == nil {
-		identity, err := pool.kademlia.FetchPeerIdentity(ctx, id)
-		if err != nil {
-			return nil, Error.Wrap(err)
+	if identity == nil {
+		info.mu.Lock()
+		defer info.mu.Unlock()
+
+		// Check to see if another goroutine has won
+		identity = info.identity
```
stefanbenten: Do we need to check here again?

storagenode/trust/service.go
```diff
@@ -127,18 +127,26 @@ func (pool *Pool) GetSignee(ctx context.Context, id storj.NodeID) (_ signing.Sig
...
+		if identity == nil {
+			identity, err = pool.kademlia.FetchPeerIdentity(ctx, id)
+			if err != nil {
+				return nil, Error.Wrap(err)
+			}
+			info.identity = identity
```
stefanbenten: Shouldnt the look around the check here be fine?
A la
```
identity, err = pool.kademlia.FetchPeerIdentity(ctx, id)
if info.identity == nil {
    info.mu.Lock()
    info.identity = identity
    info.mu.Unlock()
}
```

stefanbenten: I mean we are running an additional request into the kad network, but that shouldnt be much of an issue right?

zeebo: It would be running potentially N requests if there were N concurrent misses.

storagenode/trust/service.go
zeebo: Yes, in the case that there are two concurrent misses, only one wins the mutex, so the other should just happily use the value that the other stored.

zeebo:
I think this doesn't address any actual problem: the real time spent is in the kademlia lookup, and there's no way a reader could make progress while that's happening.

PR #2439, opened 2019-07-02
Add the ability to set dial and request timeouts from the cmd/uplink CLI to libuplink.

simongui:
# What
This PR increases the transport `dial` and `write` timeouts to be `2 minutes` to allow slower connections to complete.

This PR also adds the ability for libuplink users like mobile applications to configure their own timeouts.

This is a temporary fix until we implement a more elaborate fix for slower performing clients.

# Why
Without this setting we are too aggressive on timing out on slow connections. We don't modify all the defaults so that we don't affect other consumers of libuplink like audits or repairs.

Please describe the tests:
 - Test 1:
 - Test 2:
 
# Please describe the performance impact
Longer timeouts for CLI clients.


cmd/uplink/cmd/root.go
```diff
@@ -86,6 +87,11 @@ func (cliCfg *UplinkFlags) NewUplink(ctx context.Context) (*libuplink.Uplink, er
 		SkipPeerCAWhitelist: !cliCfg.TLS.UsePeerCAWhitelist,
 		PeerCAWhitelistPath: cliCfg.TLS.PeerCAWhitelistPath,
 	}
+
+	// We use longer timeouts for cmd/uplink to account for slower connections.
+	libuplinkCfg.Volatile.DialTimeout = time.Minute * 2
```
mniewrzal: won't this affect all connections, not only slow?

mniewrzal: Shouldn't this be:
```
	libuplinkCfg.Volatile.DialTimeout = cliCfg.Client.DialTimeout
	libuplinkCfg.Volatile.RequestTimeout = cliCfg.Client.RequestTimeout
```

simongui: Yeah but normal/fast connections complete before the timeout whether it's `2m` or the previous `20s`.

mniewrzal: Ok, in the past timeout was set to low as possible (20sec) because in some flows one hanged storage node could increase whole cp/rm time. I'm not sure if its still a case.

simongui: Yeah I think in the future we want to read all the settings from the config file that way and ship different config defaults for audit/repairs than for CLI/mobile clients but we don't have that in place right now. This is a temporary fix to override the defaults for `cmd/uplink` to allow the CLI to have longer timeouts. If there's another way to do this I'm open to suggestions!

mniewrzal: I'm not 100% but satellite transport will have default 20 sec one way or another https://github.com/storj/storj/blob/f56cb1c6123ce752458bc7e29e12326f49c6971c/satellite/peer.go#L249

simongui: Good call! Nobody calls `transport.NewClientWithTimeouts()` so I can make the CLI client do that call specifically and leave the Satellite using the default path that inherits `20s` timeouts. I've pushed changes with your suggestions!

cmd/uplink/cmd/root.go
```diff
@@ -86,6 +86,11 @@ func (cliCfg *UplinkFlags) NewUplink(ctx context.Context) (*libuplink.Uplink, er
 		SkipPeerCAWhitelist: !cliCfg.TLS.UsePeerCAWhitelist,
 		PeerCAWhitelistPath: cliCfg.TLS.PeerCAWhitelistPath,
 	}
+
+	// We use longer timeouts for cmd/uplink to account for slower connections.
```
mniewrzal: comment is outdated now

lib/uplink/uplink.go
```diff
@@ -82,6 +91,12 @@ func (cfg *Config) setDefaults(ctx context.Context) error {
 	if cfg.Volatile.Log == nil {
 		cfg.Volatile.Log = zap.NewNop()
 	}
+	if cfg.Volatile.DialTimeout.Seconds() == 0 {
+		cfg.Volatile.DialTimeout = 20 * time.Second
```
mniewrzal: maybe we can make `transport.defaultDialTimeout` public and use here? also for `RequestTime`. what do you think?

cmd/uplink/cmd/root.go
simongui: Removed!

lib/uplink/uplink.go
simongui: Good idea! Done.

uplink/config.go
```diff
@@ -39,8 +39,8 @@ type ClientConfig struct {
 	SatelliteAddr  string        `releaseDefault:"127.0.0.1:7777" devDefault:"127.0.0.1:10000" help:"the address to use for the satellite" noprefix:"true"`
 	MaxInlineSize  memory.Size   `help:"max inline segment size in bytes" default:"4KiB"`
 	SegmentSize    memory.Size   `help:"the size of a segment in bytes" default:"64MiB"`
-	RequestTimeout time.Duration `help:"timeout for request" default:"0h0m20s"`
-	DialTimeout    time.Duration `help:"timeout for dials" default:"0h0m20s"`
+	RequestTimeout time.Duration `help:"timeout for request" default:"0h2m00s"`
+	DialTimeout    time.Duration `help:"timeout for dials" default:"0h2m00s"`
```
thepaul: These don't match with `DefaultRequestTimeout` and `DefaultDialTimeout`. Without digging through stuff, is there an easy answer to where this would apply versus where the DefaultFoo versions apply?

simongui: It's not very obvious but these defaults get applied via `UplinkFlags` here in [cmd/uplink/cmd/root.go](https://github.com/storj/storj/blob/master/cmd/uplink/cmd/root.go#L30). Since the Satellite goes down a different path it gets a different default applied from `pkg/transport/common.go` above.

thepaul: So these defaults apply to things using the Config struct, while the ones in pkg/transport/common.go only apply for the satellite? Can we move the satellite-only defaults in to the satellite-specific code? Or if it happens by some other mechanism that other code might need to use eventually, can we at least clarify in the comments for `DefaultRequestTimeout` and `DefaultDialTimeout` where they apply as opposed to the default Config values?

simongui: I added the comments you suggested. The config struct only applies to the uplink CLI. `pkg/transport/common.go` applies to every other use case in various packages such as `satellite`, `bootstrap`, `testplanet`, `certificates`, `kademlia`, `peertls`, `storagenode`. I'm not sure pulling it out into the Satellite makes sense yet because all of these still need to inherit a default timeout. Any consumers of libuplink need defaults as well.

Maybe we override it as needed in the future?

thepaul: I don't see the added comments, but a grep for `uplink.Config` shows it being used in `cmd/uplink`, `cmd/gateway`, `testplanet`, and `miniogw` at least. `testplanet` then might use _both_? It seems like making `DefaultDialTimeout` and `DefaultRequestTimeout` public has made the situation a lot more confusing. Or maybe I just haven't identified the right logical separation between these?

simongui: Yeah, it did make it a little bit more confusing than it already was. I was trying to get a fix out. We use the transport in different ways. Sometimes from an uplink client but sometimes direct. For example. It was mentioned above to make `DefaultDialTimeout` public because it may be cleaner to share the transport default instead of just duplicating it somewhere else in the code.

Do you have any other suggestions?

The callstack patterns look as follows.

#### Gateway 
```
lib/uplink.NewUplink()
        lib/uplink.setDefaults()
                  applies `transport.DefaultDialTimeout of 20s
                          passes into `transport.NewClientWithTimeouts()
```

#### Minio
Same as Gateway.
```
lib/uplink.NewUplink()
        lib/uplink.setDefaults()
                  applies `transport.DefaultDialTimeout of 20s
                          passes into `transport.NewClientWithTimeouts()
```

#### Satellite
```
transport.NewClient()
        transport.NewClientWithTimeouts() passed no timeouts
                defaults from transport.DefaultRequestTimeout gets applied.
```

#### Storage node
Same as above Satellite
```
transport.NewClient()
        transport.NewClientWithTimeouts()`passed no timeouts.
                defaults from transport.DefaultRequestTimeout gets applied.
```

#### Test Planet
Explicitly sets up a config and sets the following.
```
config.Client.RequestTimeout = 10 * time.Second
config.Client.DialTimeout = 10 * time.Second

lib/uplink.NewUplink() passed with a timeout of 10s
       passes into `transport.NewClientWithTimeouts()
```

thepaul: When defaultRequestTimeout was private, it was easy to tell that it only applied to that package, and easy to find where specifically it was used. Now it seems like I have to dig through code to find call stacks in order to determine whether we get the default default or the other default.

What if we had `defaultTransportRequestTimeout` in pkg/transport, a separate `defaultUplinkRequestTimeout` in lib/uplink, and the existing config-item default in uplink/config.go, and just let them all be different things?

lib/uplink/uplink.go
```diff
@@ -62,6 +63,14 @@ type Config struct {
...
 		// attribution
 		PartnerID string
+
+		// DialTimeout is the maximum time to wait connecting to another node.
+		// If not set, the library default (20 seconds) will be used.
+		DialTimeout time.Duration
+
+		// RequestTimeout is the maximum time to wait for a request response from another node.
+		// If not set, the library default (20 seconds) will be used.
+		RequestTimeout time.Duration
```
bryanchriswhite: I'm not sure what the expected behavior is but I just want to make sure this is what should be happening. Unless we add these fields to the `UplinkConfig` C struct, a C api user won't have a way to change them. This is fine so long as the default is correct; which, in this case, would be 20 seconds [(`transport.Default...`)](https://github.com/storj/storj/blob/8836fd6e469631e176cef5d05254cd253a9ddb87/pkg/transport/common.go) as per [`Config#setDefaults`](https://github.com/storj/storj/blob/8836fd6e469631e176cef5d05254cd253a9ddb87/lib/uplink/uplink.go#L94).

uplink/config.go
bryanchriswhite: I'm interested in the resolution and/or the answer to @thepaul's last question but I'm not opposed to that going into another ticket/PR.

lib/uplink/uplink.go
simongui: We discussed and decided to put the C changes in another PR.

lib/uplink/uplink.go
```diff
@@ -16,6 +17,9 @@ import (
 	"storj.io/storj/uplink/metainfo"
 )
 
+const defaultUplinkDialTimeout = 20 * time.Second
+const defaultUplinkRequestTimeout = 20 * time.Second
```
egonelbre: You can use a single const block here.

```
const (
    defaultUplinkDialTimeout = 20 * time.Second
    defaultUplinkRequestTimeout = 20 * time.Second
)

```

PR #2440, opened 2019-07-03
lib/uplink: ensure tests are compiled separately

egonelbre:
What: Use `uplink_test` for all tests in `lib/uplink`

Why: The tests depend on testplanet which wants to use `lib/uplink`, and this breaks that cycle.

Please describe the tests:
 - Test 1:
 - Test 2:
 
None


pkg/storj/object.go
```diff
@@ -26,7 +26,7 @@ var (
 // Bucket contains information about a specific bucket
 type Bucket struct {
 	Name                 string
-	Attribution          string
+	Attribution          string // TODO: this should be uuid.UUID
```
jtolds: these should be user agent formatted strings? not uuids

lib/uplink/bucket.go
```diff
@@ -29,6 +29,9 @@ type Bucket struct {
 
 // TODO: move the object related OpenObject to object.go
 
+// Info returns information about the bucket.
+func (b *Bucket) Info() storj.Bucket { return b.bucket }
```
jtolds: any field we need here should probably be on the bucket type directly? there are already a few

pkg/storj/object.go
zeebo: Not according to everywhere else in the code base. For example, see https://github.com/storj/storj/blob/f9ed0dc1a83c0b7e6debef465ac846af56192885/satellite/attribution/db.go#L18-L24

lib/uplink/bucket.go
zeebo: `CreateBucket` on the project type already returns a `storj.Bucket` https://github.com/storj/storj/blob/f9ed0dc1a83c0b7e6debef465ac846af56192885/lib/uplink/project.go#L97

We could promote up just the `Attribution` field, but that seems silly to do just for a test. I'm just going to work around this with some reflect magic, for now.

zeebo:
I hopefully did the minimum required to get this merged.

PR #2441, opened 2019-07-03
storagenode/piecestore: add piece_creation field

Fadila82:
What: Add creation date field in pieceinfo

Why: to be able to compare the creation date and the bloom filter creation timestamp in garbage collection

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/infodb.go
```diff
@@ -217,6 +217,13 @@ func (db *InfoDB) Migration() *migrate.Migration {
 					)`,
 				},
 			},
+			{
+				Description: "Add creation date.",
+				Version:     6,
+				Action: migrate.SQL{
+					`ALTER TABLE pieceinfo ADD COLUMN piece_creation TIMESTAMP`,
```
egonelbre: this probably should be not null

mobyvb: maybe something like `ALTER TABLE pieceinfo ADD COLUMN piece_creation TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT 'epoch';`

storagenode/storagenodedb/infodb.go
```diff
@@ -221,7 +221,7 @@ func (db *InfoDB) Migration() *migrate.Migration {
 				Description: "Add creation date.",
 				Version:     6,
 				Action: migrate.SQL{
-					`ALTER TABLE pieceinfo ADD COLUMN piece_creation TIMESTAMP`,
+					`ALTER TABLE pieceinfo ADD COLUMN piece_creation TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01'`,
```
kaloyan-raev: We define similar columns on the satellite as `TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT 'epoch'`

Perhaps, do the same here for consistency.

storagenode/storagenodedb/testdata/sqlite.v6.sql
```diff
@@ -93,8 +93,8 @@ INSERT INTO certificate VALUES(2,X'2b3a5863a41f25408a8f5348839d7a1361dbd886d7578
...
-INSERT INTO pieceinfo VALUES(X'2b3a5863a41f25408a8f5348839d7a1361dbd886d75786bb139a8ca0bdf41000',X'd5e757fd8d207d1c46583fb58330f803dc961b71147308ff75ff1e72a0df6b0b',123,'2019-05-09 00:00:00.000000+00:00',X'0a20d5e757fd8d207d1c46583fb58330f803dc961b71147308ff75ff1e72a0df6b0b120501020304051a483046022100e623cf4705046e2c04d5b42d5edbecb81f000459713ad460c691b3361817adbf022100993da2a5298bb88de6c35b2e54009d1bf306cda5d441c228aa9eaf981ceb0f3d',2,NULL,NULL);
+INSERT INTO pieceinfo VALUES(X'0ed28abb2813e184a1e98b0f6605c4911ea468c7e8433eb583e0fca7ceac3000',X'd5e757fd8d207d1c46583fb58330f803dc961b71147308ff75ff1e72a0df6b0b',123,'2019-05-09 00:00:00.000000+00:00',X'0a20d5e757fd8d207d1c46583fb58330f803dc961b71147308ff75ff1e72a0df6b0b120501020304051a47304502201c16d76ecd9b208f7ad9f1edf66ce73dce50da6bde6bbd7d278415099a727421022100ca730450e7f6506c2647516f6e20d0641e47c8270f58dde2bb07d1f5a3a45673',1,NULL,'1970-01-01 00:00:01');
```
kaloyan-raev: You can use `epoch` instead of `1970-01-01 00:00:01`.

storagenode/storagenodedb/infodb.go
kaloyan-raev: I think that the `WITH TIME ZONE` declaration now fails the tests. Maybe it was not a good idea for SQLite.

storagenode/piecestore/verification_test.go
```diff
@@ -322,10 +322,12 @@ func setSpace(ctx context.Context, t *testing.T, planet *testplanet.Planet, spac
 		availableSpace, err := storageNode.Storage2.Monitor.AvailableSpace(ctx)
 		require.NoError(t, err)
 		diff := (space - availableSpace) * -1
+		now := time.Now()
```
mobyvb: do we need to use `time.Now().UTC()`?

kaloyan-raev: Why should we?

mobyvb: I thought `time.Now()` used the local timezone, while the dates used for the bloom filters sent from the satellite will be UTC

PR #2445, opened 2019-07-03
Protobufs for the Metainfo RPC service

mniewrzal:
What: Protobuf changes for Metainfo RPC service

Why: We want to refactor current API


mniewrzal:
Messages in conflict with old API have suffix `2`. Satellite-specific messages have prefix `Sat` (SegmentID, StreamID; this one has also conflict with storj.StreamID alias pb.StreamID)

pkg/pb/metainfo.proto
```diff
@@ -15,11 +15,31 @@ import "orders.proto";
...
+    rpc BeginDeleteObject(ObjectBeginDeleteRequest) returns (ObjectBeginDeleteResponse);
+    rpc FinishDeleteObject(ObjectFinishDeleteRequest) returns (ObjectFinishDeleteResponse);
+    // Segment
+    rpc BeginSegment(SegmentBeginRequest) returns (SegmentBeginResponse);
+    rpc CommitSegment2(SegmentCommitRequest2) returns (SegmentCommitResponse2);
+    rpc MakeInlineSegment(SegmentMakeInlineRequest) returns (SegmentMakeInlineResponse);
+    rpc BeginDeleteSegment(SegmentBeginDeleteRequest) returns (SegmentBeginDeleteResponse);
+    rpc FinishDeleteSegment(SegmentFinishDeleteRequest) returns (SegmentFinishDeleteResponse);
+    rpc ListSegments2(SegmentListRequest) returns (SegmentListResponse);
+    rpc DownloadSegment2(SegmentDownloadRequest2) returns (SegmentDownloadResponse2);
```
zeebo: Can't we have the old API in a different service so that we don't have to name this with a `2` forever? It's unclear to me how that would affect implementing it, though, so maybe that's not possible. Maybe there's some other naming scheme we can use?

pkg/pb/metainfo.proto
```diff
@@ -193,3 +213,321 @@ message ProjectInfoRequest {
...
+        SegmentCommitRequest     segment_commit = 12;
+        SegmentMakeInlineRequest segment_make_inline = 13;
+
+        SegmentBeginDeleteRequest  segment_begin_delete = 14;
+        SegmentFinishDeleteRequest segment_finish_delete = 15;
+
+        SegmentListRequest     segment_list = 16;
+        SegmentDownloadRequest segment_download = 17;
+    }
+    //repeated Request requests = 1;
```
zeebo: why are these repeated fields commented out?

pkg/pb/metainfo.proto
mniewrzal: one way or another it will be in `pb` package so moving this wont work. I was trying to find out if `rpc` names can be changed in a backward-compatible way but spec is saying mainly about fields. We can choose a different naming scheme if it will be needed.

pkg/pb/metainfo.proto
mniewrzal: I forgot to fix this. done

pkg/pb/metainfo.proto
egonelbre: Changing name of service methods is not backwards compatible.

mniewrzal: Yeah, I realized that yesterday

mniewrzal: I guess we can rename old one. Otherwise we can end up with something like suffixes (e.g. `_v2`) for all new methods. What do you propose for temp name scheme for old methods and messages? `*_v1`? :) I will prepare separate PR for this.

egonelbre: I would use `Old` prefix, but a particular choice probably doesn't matter that much.

mniewrzal: https://github.com/storj/storj/pull/2462

pkg/pb/metainfo.proto
```diff
@@ -193,3 +213,327 @@ message ProjectInfoRequest {
...
+    int32  version = 3;
+}
+
+message ObjectBeginDeleteResponse {
+    bytes  stream_id = 1 [(gogoproto.customtype) = "StreamID", (gogoproto.nullable) = false];
+
+    // TODO: should this contain a list of segments needing to be deleted?
+}
+
+message ObjectFinishDeleteRequest {
```
mniewrzal: Shouldn't we have here only `streamID` similar to `ObjectCommitRequest`? @egonelbre

egonelbre: Yeah

egonelbre:
Already implemented in an other PR.

PR #2446, opened 2019-07-03
storagenode/consoledb: fix daily bandwidth query

rikysya:
What:  storagenode/consoledb fix daily bandwidth query

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/consoledb.go
```diff
@@ -98,7 +98,7 @@ func (db *consoledb) getDailyBandwidthUsed(ctx context.Context, cond string, arg
 	qb.WriteString("GROUP BY DATE(created_at), action ")
 	qb.WriteString("ORDER BY created_at ASC")
 
-	rows, err := db.db.QueryContext(ctx, db.Rebind(qb.String()), args)
+	rows, err := db.db.QueryContext(ctx, db.Rebind(qb.String()), args...)
 	if err != nil {
```
egonelbre: This seems to be missing tests.

storagenode/storagenodedb/consoledb.go
```diff
@@ -98,7 +98,7 @@ func (db *consoledb) getDailyBandwidthUsed(ctx context.Context, cond string, arg
 	qb.WriteString("GROUP BY DATE(created_at), action ")
 	qb.WriteString("ORDER BY created_at ASC")
```
egonelbre: Using a string addition and multiline string would be completely fine here it doesn't make a big difference.

rikysya: It looks more clear this way and easier to make changes imo

egonelbre: Not just a nit, nothing needs to change. The difference I see:

```
qb := strings.Builder{}
qb.WriteString("SELECT action, SUM(amount), created_at ")
qb.WriteString("FROM bandwidth_usage ")
if cond != "" {
    qb.WriteString(cond + " ")
}
qb.WriteString("GROUP BY DATE(created_at), action ")
qb.WriteString("ORDER BY created_at ASC")

rows, err := db.db.QueryContext(ctx, db.Rebind(qb.String()), args...)
if err != nil {
    return nil, err
}
```

compared to:

```
    query := db.Rebind(`
        SELECT action, SUM(amount), created_at
        FROM bandwidth_usage
        ` + cond + `
        GROUP BY DATE(created_at), action 
        ORDER BY created_at ASC
    `)

    rows, err := db.db.QueryContext(ctx, query, args...)
	if err != nil {
		return nil, err
	}
```

storagenode/storagenodedb/consoledb.go
rikysya: let's make it a separate pull request, so the other part of the team can use the query meanwhile

PR #2452, opened 2019-07-03
change bucket table to fix conflict with pointerDB

JessicaGreben:
What: Why:
This PR changes the name of the new table `buckets` in MasterDB.  Apparently there is a conflict with pointerDB since it has an existing table named `buckets`.  In the future the pointerDB table should be removed/renamed, but for now we will just change the name in masterDB.


satellite/satellitedb/migrate.go
```diff
@@ -905,6 +905,13 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 					`UPDATE nodes SET disqualified=NULL WHERE disqualified IS NOT NULL AND audit_reputation_alpha / (audit_reputation_alpha + audit_reputation_beta) >= 0.6;`,
 				},
 			},
+			{
```
egonelbre: we might need to change the initial create query, because that would still fail during migrations.

JessicaGreben: oh yeah good point.

JessicaGreben: _updated_

PR #2423, opened 2019-07-02
[V3-1927] Repairer uploads to max threshold instead of success threshold

ifraixedes:
**What:**

When a segment needs to be repaired, we should upload only the needed repaired number of pieces to reach success threshold.

Currently, we are uploading to the maximum threshold.

This branch restrict the number of pieces to be repaired to percentage excess over the successful threshold.

**Why:**

Because:

* Uploading more pieces than the success threshold causes that we have to pay more for storage than what's needed.
* Uploading less pieces than the success threshold causes that the segments have to be repaired more frequently, causing that we have to pay more for bandwidth and computation.

We have decided to upload to a percentages excess over the threshold rather than doing the changes for the ideal solution (upload repaired pieces exactly to the success threshold) because the improvement should be good enough to move on and the time required for this improvement should be quite less than the ideal one. The decision was made based in that it's preferable, cost wise, to pay for a bit more storage than having to repair more frequently.

Please describe the tests:
 - Test 1: One existing test has been changed to actually check that the number of repaired pieces doesn't surpass the new threshold (success threshold + the specified excess)
 - Test 2: A new test has been added to make sure that the number of uploaded repaired pieces doesn't surpass the new threshold (success threshold + the specified excess)
 
Please describe the performance impact: Repairers should be faster because it has to repair and upload less pieces, however there isn't any benchmark/profiling in this branch, so this improvement is only an hypothesis.

## Code Review Checklist (to be filled out by reviewer)
 - [x] Does the PR describe what changes are being made?
 - [x] Does the PR describe why the changes are being made?
 - [x] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [x] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [x] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [x] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [x] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [x] Does any documentation need updating?
 - [x] Do the database access patterns make sense?

PR #2454, opened 2019-07-04
pkg/storage/ec: Make minor improvements in client.Repair

ifraixedes:
What:

* Fix some log message & remove unneeded `nil` conditional check

Why:

* Log message was reporting the invalid number of pieces to be repaired for reaching the optimal/success threshold.
* Better to have less logic when there is not needed.


pkg/storage/ec/client.go
```diff
@@ -199,11 +199,11 @@ func (ec *ecClient) Repair(ctx context.Context, limits []*pb.AddressedOrderLimit
 	successfulHashes = make([]*pb.PieceHash, len(limits))
 	var successfulCount int32
 
-	// how many nodes must be repaired to reach the success threshold: o - (n - r)
-	optimalCount := rs.OptimalThreshold() - (rs.TotalCount() - nonNilCount(limits))
-
-	ec.log.Sugar().Infof("Starting a timer for %s for repairing %s to %d nodes to reach the success threshold (%d nodes)...",
-		timeout, path, optimalCount, rs.OptimalThreshold())
+	{ // log num of pieces to be repaired to reach the success threshold
+		optimalCount := rs.OptimalThreshold() - nonNilCount(limits)
```
mniewrzal: Why this formula changed?

ifraixedes: Fro the reason commented in the PR message

> Log message was reporting the invalid number of pieces to be repaired for reaching the optimal/success threshold.

As far as I could understand the previous calculation was wrong, but I may be wrong, so please double check if this change make sense and the previous calculation was not correct.

mniewrzal: hmm isn't that we are trying to upload all `non nil limits` so we should display in log just `nonNilCount(limits)`.

ifraixedes: oh, yes you're right here, because I'm working in https://github.com/storj/storj/pull/2423 and I found this bug in that branch but I wanted to have a separated PR for this fix, I did the change needed by the changes of that PR.

So, yes, here is `nonNilCount(limits)` and I will do this change in the branch https://github.com/storj/storj/pull/2423

ifraixedes: Done https://github.com/storj/storj/pull/2454/commits/1ef9e2a9e8bcc6ea7d92235e81c9c22f23255304

I also rearranged some vars, around that log message, closer where they are used.

mniewrzal:
strange, now I'm seeing changes from different PRs that were merged

ifraixedes:
> strange, now I'm seeing changes from different PRs that were merged

That's because the branch was updated from master from Github, I forgot to pull it to update my local, then I made the commit which changed what you reported and I merged master in this branch in my local; at that point I couldn't push, because I had to pull, then I pull and and push (to be clear without `force`) and it ended up in that way.

PR #2456, opened 2019-07-04
storagenode/dashboard: show console address

rikysya:
What:  storagenode/dashboard show console address

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/inspector.proto
```diff
@@ -184,10 +184,11 @@ message DashboardResponse {
   string bootstrap_address = 3;
   string internal_address = 4;
   string external_address = 5;
-  StatSummaryResponse stats = 6;
-  google.protobuf.Duration uptime = 7;
-  google.protobuf.Timestamp last_pinged = 8;
-  google.protobuf.Timestamp last_queried = 9;
+  string dashboard_address = 6;
```
ifraixedes: I suppose that rearranging fields isn't currently a problem, because we can introduce breaking changes.

pkg/pb/orders.pb.go
```diff
@@ -3,17 +3,15 @@
...
-import fmt "fmt"
-import math "math"
-import _ "github.com/gogo/protobuf/gogoproto"
-import timestamp "github.com/golang/protobuf/ptypes/timestamp"
-
-import time "time"
-
 import (
-	context "golang.org/x/net/context"
+	context "context"
```
egonelbre: Why did this file change? Which version of protoc / gogo are you using or did you manually format the files? (These may have been generated with a different version previously as well.)

storagenode/inspector/inspector.go
```diff
@@ -131,6 +141,7 @@ func (inspector *Endpoint) getDashboardData(ctx context.Context) (_ *pb.Dashboar
 		BootstrapAddress: strings.Join(bsNodes, ", "),
 		InternalAddress:  "",
 		ExternalAddress:  inspector.kademlia.Local().Address.Address,
+		DashboardAddress: inspector.consoleConfig.Address,
```
egonelbre: This is not the address, this is the specification address... which could be `127.0.0.1:0`, you need the actual address.

pkg/pb/orders.pb.go
rikysya: I just use `pb.gen.go`

egonelbre: Yeah, it looks like somebody used older protobuf / protoc or some other version.

egonelbre: Yeah, it looks like somebody used older protobuf / protoc or some other version.

PR #2466, opened 2019-07-05
Move kademlia dialer into separate package

cam-a:
This PR moves the kademlia dialer and its methods into a new package called "kademliaclient".

Why: Circular dependencies in kademlia and trust. We need the ability to dial satellites from the trust package, and we also need to import trust into kademlia to verify vouchers with the trusted satellites list. With the dialer separated into its own package, we don't need to import kademlia into trust


pkg/dial/dialer.go
```diff
@@ -18,20 +19,22 @@ import (
 	"storj.io/storj/pkg/transport"
 )
 
-// Dialer is a kademlia dialer
+var mon = monkit.Package()
+
+// Dialer is a dialer
```
jenlij: 😂love that, can we maybe be a little more descriptive? Not your fault, this is a common occurrence across the codebase.

pkg/kademlia/kademlia_test.go
```diff
@@ -127,30 +149,30 @@ func TestBootstrap(t *testing.T) {
 	ctx := testcontext.New(t)
 	defer ctx.Cleanup()
 
-	bn, s, clean := testNode(ctx, "1", t, []pb.Node{})
+	bn, clean := newTestNode(ctx, "1", t, []pb.Node{})
```
jenlij: does bn stand for bootstrap node?

pkg/kademlia/kademlia_test.go
```diff
@@ -35,6 +34,12 @@ const (
 	defaultAlpha = 5
 )
 
+type testNode struct {
```
jenlij: why is this named testNode?

pkg/dial/dialer.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-package kademlia
+package dial
```
egonelbre: Leave this as a subpackage to `kademlia` because it's kademlia specific.

pkg/dial/dialer_test.go
```diff
@@ -42,7 +42,7 @@ func TestDialer(t *testing.T) {
 	{ // PingNode: storage node pings all other storage nodes
 		self := planet.StorageNodes[0]
 
-		dialer := kademlia.NewDialer(zaptest.NewLogger(t), self.Transport)
+		dialer := dial.NewDialer(zaptest.NewLogger(t), self.Transport)
```
egonelbre: notice the repetition in the naming.

pkg/kademlia/kademlia.go
```diff
@@ -141,7 +142,7 @@ func (k *Kademlia) DumpNodes(ctx context.Context) (_ []*pb.Node, err error) {
 
 // Bootstrap contacts one of a set of pre defined trusted nodes on the network and
 // begins populating the local Kademlia node
-func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
+func (k *Kademlia) Bootstrap(ctx context.Context, transport transport.Client) (err error) {
```
egonelbre: This probably shouldn't need a separate argument for transport client?

cam-a: With the dialer moved to another package, the transport field on it is not reachable from inside `Bootstrap`
https://github.com/storj/storj/pull/2466/files#diff-f6eb0f0cefab0c21ae61c0e1022a854eL186

pkg/kademlia/kademlia_test.go
cam-a: yes

pkg/kademlia/kademlia_test.go
cam-a: because it is returned by the testNode function

cam-a: I thought it would be better to make a struct to hold the variables rather than return 4 variables out of testNode

jenlij: gotcha

pkg/kademlia/kademlia.go
```diff
@@ -183,7 +184,7 @@ func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
 			// The way FetchPeerIdentityUnverified does is is to do a basic ping request, which
 			// we have now done. Let's tell all the transport observers now.
 			// TODO: remove the explicit transport observer notification
-			k.dialer.transport.AlertSuccess(ctx, &pb.Node{
+			transport.AlertSuccess(ctx, &pb.Node{
```
egonelbre: Was there a reason dialer isn't alerting itself?

pkg/kademlia/dial/dialer.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
-package kademlia
+package dial
```
egonelbre: Try to see whether you can find a better name for this.

pkg/kademlia/kademlia.go
cam-a: Apparently it has something to do with the comment above it, but I don't see why we couldn't just call this inside `FetchPeerIdentityUnverified`. Waiting for @jtolds to have some free time to discuss

cam-a: I added a wrapper to call this from dial package

pkg/kademlia/dial/dialer.go
cam-a: Any suggestions? Perhaps package `dialer`, but change the name of the variable that the dialer is assigned to, ie 
```
type Foo struct {
    dial *dialer.Dialer
}
```
Perhaps permutations of the word "dial" isn't the best way to go

PR #2467, opened 2019-07-05
Rename voucher service to endpoint

cam-a:
This PR renames the satellite voucher Service to Endpoint because it is an endpoint, not a service


satellite/vouchers/service.go
```diff
@@ -18,13 +18,13 @@ import (
 	"storj.io/storj/pkg/pb"
 )
```
jenlij: there's not really a good place to add this comment, but should this file be named endpoint.go rather than service.go now?

cam-a: oh thanks for catching that!

PR #2474, opened 2019-07-08
satellite/satellitedb : Set-Default-Offers

fuskovic:
What: Adds a couple of insert statements to the migrate file.

Why: The offers table needs a default offer type set for each offer type.

Please describe the tests:
 - Test 1: No unit test. If the default offers populate when starting the service we'll know the change was successfully implemented.
 - Test 2:
 

fuskovic:
After adding the insert statements I ran a `go generate` inside `dbx/` but the offers don't render to the UI and are not being retrieved from the table. Not sure why. Going to ask for help. Its worth mentioning that I confirmed both query statements are valid because I tested executing each one in my DB browser for SQLite client.

satellite/satellitedb/dbx/satellitedb.dbx.go
```diff
@@ -485,7 +484,7 @@ CREATE TABLE buckets (
 	project_id bytea NOT NULL REFERENCES projects( id ),
 	name bytea NOT NULL,
 	path_cipher integer NOT NULL,
-	created_at timestamp NOT NULL,
+	created_at timestamp with time zone NOT NULL,
```
VinozzZ: Why is this change?

fuskovic: This must have happened on the go-generate because I didn't change this manually.

satellite/satellitedb/migrate.go
```diff
@@ -904,6 +904,67 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 				Action: migrate.SQL{
 					`UPDATE nodes SET disqualified=NULL WHERE disqualified IS NOT NULL AND audit_reputation_alpha / (audit_reputation_alpha + audit_reputation_beta) >= 0.6;`,
 				},
+			},{
+				Description: "Set default offer for each offer type in offers table",
```
VinozzZ: Add a new test data file for version 40

VinozzZ:
@fuskovic I have tested the migration with pointing storj-sim to a local postgres instance and it's working as expected. You should be able to add a test data file and then it will make jenkins happy again

jenlij:
Did you pull the changes from master and they've showed up as new changes on your branch? 125 files changed looks unusual.

PR #2476, opened 2019-07-08
adding expiration date check on satellite's CreateSegment

aligeti:
What:
Bug fix
https://storjlabs.atlassian.net/browse/V3-2001

Why:
Root cause:
The expiration time when set, was checked at the user’s uplink or libuplink end. If the user changes the code to accept it, the satellite should also check the validity of the time before segments are created. Per the problem stated and upon looking at the satellite’s metainfo createsegment() I see satellite accepts ANY expiration date. The root cause is the satellite should filter invalid expiration times before it goes any further. Changing just the error is not the root cause.
Please describe the tests:

Test 1:
TestExpirationTimeSegment()
Test 2:

### DIFF
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -135,9 +135,8 @@ func (endpoint *Endpoint) CreateSegmentOld(ctx context.Context, req *pb.SegmentW
 			return nil, status.Errorf(codes.InvalidArgument, err.Error())
 		}
 
-		err = validateExpiration(ctx, exp)
-		if err != nil {
-			return nil, status.Errorf(codes.InvalidArgument, err.Error())
+		if !exp.After(time.Now()) {
+			return nil, errs.New("Invalid expiration time")
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


satellite/metainfo/metainfo.go
```diff
@@ -128,6 +129,18 @@ func (endpoint *Endpoint) SegmentInfoOld(ctx context.Context, req *pb.SegmentInf
...
 	defer mon.Task()(&ctx)(&err)
 
+	if req.Expiration != nil {
+		exp, err := ptypes.Timestamp(req.Expiration)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, err.Error())
+		}
+
+		err = validateExpiration(ctx, exp)
+		if err != nil {
```
egonelbre: `if expiration.Less(time.Now().Add(piece expiration thershold)) {`

aligeti: updated using exp.After(u Time) and cleaned up the code

PR #2477, opened 2019-07-08
Removing trace of encodedPiece.Read() because it's fast.

simongui:
What: Removing monkit tracing of `encodedPiece.Read()` because it's really fast (less than `1ms`) and it's exploding tracing files to nearly `100MB`.

Why: The generated SVG is too large to open in Chrome and search effectively.

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/eestream/encode.go
```diff
@@ -191,8 +191,6 @@ type encodedPiece struct {
 }
 
 func (ep *encodedPiece) Read(p []byte) (n int, err error) {
```
egonelbre: Are there other places that might have this problem?

simongui: The problem of making the trace file too big? Potentially yeah because recently some additional tracing was added that is tracing really fast functions but it's pretty difficult to open the trace file right now with these reads. Maybe after this PR we can sweep and do a cleanup PR once this PR trims the trace file size down to a size we can actually open and search.

thepaul: Would it be worth adding a comment like "tracing deemed unnecessary here" so that some future cleanup operation doesn't add it back again?

egonelbre: maybe not unnecessary, but "Tracing here is too noisy"?

simongui: Added a comment explaining why tracing was removed.

PR #2478, opened 2019-07-08
Update vouchers to use time.Time instead of timestamp

aleitner:
What: 
Update vouchers to use time.Time instead of timestamp

Why: timestamp is an unnecessary conversion

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/vouchers.go
```diff
@@ -58,7 +54,7 @@ func (db *vouchersdb) Put(ctx context.Context, voucher *pb.Voucher) (err error)
 func (db *vouchersdb) NeedVoucher(ctx context.Context, satelliteID storj.NodeID, expirationBuffer time.Duration) (need bool, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	expiresBefore := time.Now().UTC().Add(expirationBuffer)
```
jenlij: why can we remove UTC()?

aleitner: time zones are irrelevant except for when printing

cam-a: Not always. I dealt with some issues in accounting where the dbx type was `utimestamp` which expects the time passed in to be UTC, but local time was being passed in. Thus, 12pm EST was being passed in, but when pulling it out, it was 12pm UTC. But, this is just SQLite so it should be okay

aleitner: The date is essentially printed into the db so the comment still applies

aleitner: That is a good catch for writing to database as that would have been a bug because i was assuming the database stored as time.Time serialized data

PR #2484, opened 2019-07-08
satellite/console: add redeem reward method

VinozzZ:
What: 

- add `RedeemRewards` method in console
- refactor `Create` method in usercredits
- remove `Redeem` from offers db

Why:
- in order to prevent concurrency issue when checking redeemable capacity of current offer, we check current offer redeemed number in user credit table on new entry creation.
- instead of storing the redeemed number in offers table, we can check the number directly in user credits table

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/satellitedb/usercredits.go
```diff
@@ -55,21 +55,41 @@ func (c *usercredits) GetCreditUsage(ctx context.Context, userID uuid.UUID, expi
...
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		return errs.Wrap(err)
 	}
 
-	return convertDBCredit(credit)
+	if rows != 1 {
+		return errs.New("Create user credit failed")
```
thepaul: Should this error message say anything about why, like "this offer was already redeemed the maximum number of times"? Or is that not something we want to expose

satellite/satellitedb/offers.go
```diff
@@ -118,24 +118,6 @@ func (db *offersDB) Create(ctx context.Context, o *rewards.NewOffer) (*rewards.O
...
 }
 
-// Redeem adds 1 to the amount of offers redeemed based on offer id
-func (db *offersDB) Redeem(ctx context.Context, oID int, isDefault bool) error {
-	if isDefault {
-		return nil
-	}
-
-	statement := db.db.Rebind(
-		`UPDATE offers SET num_redeemed = num_redeemed + 1 where id = ? AND status = ? AND num_redeemed < redeemable_cap`,
```
thepaul: It looks like this change makes `num_redeemed` an unused column. Is it worth removing it from the schema so people don't get confused and think that it has meaning?

VinozzZ: Do you think it would be better to do it in another PR? In order to remove this column, I will need to change the logic of getting current offer. That seems to make this PR too big

VinozzZ: I did it in another PR #2510

satellite/console/service.go
```diff
@@ -486,6 +486,31 @@ func (s *Service) GetUserCreditUsage(ctx context.Context) (usage *UserCreditUsag
...
+
+	if err != nil {
+		return errs.Wrap(err)
+	}
+
+	newCredit := UserCredit{
+		UserID:        auth.User.ID,
+		OfferID:       offer.ID,
+		ReferredBy:    referrerID,
+		CreditsEarned: offer.AwardCredit,
```
VinozzZ: check reward type to get correct credit amount

satellite/satellitedb/usercredits.go
```diff
@@ -55,21 +56,41 @@ func (c *usercredits) GetCreditUsage(ctx context.Context, userID uuid.UUID, expi
...
-	)
+func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit, redeemableCap int) error {
+	var statement string
+
+	switch t := c.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		statement = `
+			INSERT INTO user_credits (user_id, offer_id, credits_earned_in_cents, credits_used_in_cents, expires_at, referred_by, created_at)
+				SELECT * FROM (VALUES (?, ?, ?, 0, ?, ?, time('now'))) AS v
+					WHERE (SELECT COUNT(offer_id) FROM user_credits WHERE offer_id = ? ) < ?;
```
VinozzZ: Need to check if redeemableCap is 0

PR #2487, opened 2019-07-08
Add checker metrics

mobyvb:
What: 
Add metrics
`checker_segment_total_count` - Number of total segments in pointer during checker iteration
`checker_segment_healthy_count` - Number of healthy segments in pointer during checker iterationn
`time_since_checker_queue` - Seconds elapsed between checker queue and beginning repair
`time_for_repair` - Seconds elapsed between beginning repair and ending repair/dequeueing

Why:
https://storjlabs.atlassian.net/browse/V3-2095
More data for DS
https://storjlabs.atlassian.net/browse/V3-2105
We want to track mean time to repair

Please describe the tests: NA
 
Please describe the performance impact: NA


pkg/datarepair/repairer/repairer.go
```diff
@@ -140,6 +140,12 @@ func (service *Service) process(ctx context.Context) (err error) {
 
 func (service *Service) worker(ctx context.Context, seg *pb.InjuredSegment) (err error) {
 	defer mon.Task()(&ctx)(&err)
+
+	insertedTime := seg.GetInsertedTime()
```
egonelbre: What if an item in the queue doesn't have this time.

mobyvb: Oh good point. Do we also need to worry about this?
```
seg *pb.InjuredSegment
var path string
err = tx.Tx.QueryRowContext(ctx, r.db.Rebind(`
	SELECT path, data FROM injuredsegments
	WHERE attempted IS NULL
	OR attempted < datetime('now','-1 hours')
	ORDER BY path LIMIT 1`)).Scan(&path, &seg)
```
If the `pb.InjuredSegment` struct has changed and the `data` stored in the `injuredsegments` table uses the old struct? I am not sure if it would still scan properly. Do you know?

pkg/datarepair/checker/checker.go
```diff
@@ -193,6 +193,9 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
 
 	numHealthy := int32(len(pieces) - len(missingPieces))
 	redundancy := pointer.Remote.Redundancy
+	mon.IntVal("checker_segment_total_count").Observe(int64(len(pieces)))
```
jenlij: how would you confirm that these are actually being logged?

pkg/datarepair/repairer/repairer.go
```diff
@@ -150,5 +153,16 @@ func (service *Service) worker(ctx context.Context, seg *pb.InjuredSegment) (err
 	if err != nil {
 		return Error.New("repair delete failed: %v", err)
 	}
+
+	insertedTime := seg.GetInsertedTime()
+	// do not send metrics if segment was added before the InsertedTime field was added
```
jenlij: why not?

pkg/datarepair/repairer/repairer.go
```diff
@@ -140,6 +140,9 @@ func (service *Service) process(ctx context.Context) (err error) {
 
 func (service *Service) worker(ctx context.Context, seg *pb.InjuredSegment) (err error) {
```
jenlij: is this method tested anywhere?

pkg/datarepair/checker/checker.go
mobyvb: It is not a log, it is actually a monkit statistic. So by going to `debugaddress/mon/stats`, you would see a bunch of details such as 
```
checker_segment_total_count sum 43919
checker_segment_total_count min	40
checker_segment_total_count avg	80
checker_segment_total_count max 100
```

pkg/datarepair/repairer/repairer.go
mobyvb: we would end up sending a bunch of inaccurate times to datascience about how long segments are repaired. Specifically for segments added before this PR. They would have the 0 time which would throw off `time_since_checker_queue`

pkg/datarepair/repairer/repairer.go
mobyvb: The datarepair integration test (`pkg/datarepair/datarepair_test.go`) ends uses this method indirectly

PR #2490, opened 2019-07-09
overlay cache: slight modification of node-is-online rules

jtolds:
the proposition is that we aren't always consistent at pinging nodes in a regular fashion. either the satellite was not running, or the discovery refresh was not operating at the regular schedule. we will consider a node online if it was successfully contacted in the last hour, even if it has failed most recently (and perhaps temporarily). we will also consider a node offline if our most recent interaction was successful, regardless of if the most recent interaction was recent.

 

internal/testplanet/reconfigure.go
```diff
@@ -37,3 +39,11 @@ var DisablePeerCAWhitelist = Reconfigure{
 		config.Server.UsePeerCAWhitelist = false
 	},
 }
+
+// ShortenOnlineWindow returns a `Reconfigure` that sets the NodeSelection
```
mobyvb: after the new test changes I am not sure whether we need this still

satellite/satellitedb/overlaycache.go
```diff
@@ -99,8 +99,8 @@ func (cache *overlaycache) SelectNewStorageNodes(ctx context.Context, count int,
 		AND free_bandwidth >= ?
 		AND free_disk >= ?
 		AND (total_audit_count < ? OR total_uptime_count < ?)
-		AND last_contact_success > ?
-		AND last_contact_success > last_contact_failure`
+		AND (last_contact_success > ?
+				 OR last_contact_success > last_contact_failure)`
```
stefanbenten: Identation is off here

internal/testplanet/satellite.go
```diff
@@ -104,7 +104,7 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 					UptimeCount:       0,
 					AuditCount:        0,
 					NewNodePercentage: 0,
-					OnlineWindow:      time.Hour,
+					OnlineWindow:      0,
```
stefanbenten: @mobyvb I am not sure if its a good idea to disable the window entirely for all tests per default.
For that reason i created the above  "Reconfigure".

internal/testplanet/reconfigure.go
stefanbenten: I would prefer to use this method to have the ability to test the full "grace" period of 1hr per default.
If we set the value to 0 for all tests, we are basically adding the immediate check back, right?

internal/testplanet/satellite.go
littleskunk: Can we just merge and release it. Update the test later (if that is needed)

PR #2501, opened 2019-07-09
cmd/uplink: use scopes to open

zeebo:
What: Change cmd/uplink to use scopes

It moves the fields that will be subsumed by scopes into an explicit legacy section and hides their configuration flags.

Why: So that it can read scopes in from files and stuff

Please describe the tests:
 - Test 1: 
 - Test 2:
 
Please describe the performance impact: None


uplink/config.go
```diff
@@ -4,13 +4,21 @@
...
 import (
+	"io/ioutil"
+	"strings"
 	"time"
 
+	"github.com/zeebo/errs"
+	"gopkg.in/spacemonkeygo/monkit.v2"
+
 	"storj.io/storj/internal/memory"
+	libuplink "storj.io/storj/lib/uplink"
```
egonelbre: Package uplink probably shouldn't depend on libuplink. `uplink` is the lower-level with all the bits and pieces for custom streaming/metainfo/encryption etc. making it depend on `libuplink` seems weird.

cmd/storj-sim/network.go
```diff
@@ -24,7 +24,9 @@ import (
 	"storj.io/storj/internal/dbutil/pgutil"
 	"storj.io/storj/internal/fpath"
 	"storj.io/storj/internal/processgroup"
+	"storj.io/storj/lib/uplink"
 	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
```
egonelbre: storj-sim was intentionally trying to keep dependencies minimal, so it would be relatively independent of the binary versions, since we need to test multiple versions of things at some point

egonelbre: i.e. Use the binaries to achieve what you want to do.

uplink/config.go
zeebo: Right now all `uplink` has is a config struct and some helpers to create some objects from it.

I don't mind moving the helper somewhere else, though, but something has to create a Scope from an `uplink.Config`.

cmd/storj-sim/network.go
zeebo: This is used to create a serialized scope to pass to the gateway for configuration. Do you have any good suggestions on how to do that? I see no reason for any command to need to do that.

uplink/config.go
egonelbre: Some of the config part shouldn't be part of it as well, it's conflating the uplink and the needs of the binary.

I suspect we are missing a uplinkflags package under `cmd/uplink/`, or `cmd/internal`, for these things.

cmd/storj-sim/network.go
egonelbre: Here we should have something that creates a uplink ready for action that has the necessary rights. Currently we've been hacking gateway to do that, even our tests use `storj-sim network env GATEWAY_0_DIR` to find the necessary information.

However the better approach would be to have the capability to add/remove users on the command-line, create scopes etc. this also comes handy for testing. The storj-sim in some terms is also a test of the tooling or how easily we can interact with the API without resorting to code.

Unfortunately we have already cut corners in quite a few places (making graphql requests).

We can leave for now, but yeah, not happy about it.

uplink/config.go
zeebo: Sure, I'd love to see a refactoring of the `uplink` package from someone who has a vision of what it's fore lol

cmd/storj-sim/network.go
zeebo: Alright, I'll keep this in mind when I'm building out the `uplink` command's features for creating/viewing/importing scopes.

uplink/config.go
egonelbre: Ok, let's leave as is, once we finish metainfo refactor, I'll clean it up. Otherwise there are too many refactorings in-flight at the same time.

jtolds:
documentation will be added during our big doc push after scopes are complete

pkg/process/config.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// readBoolAnnotation is a helper to see if a boolean annotation is set to true on the flag.
+func readBoolAnnotation(flag *pflag.Flag, key string) bool {
+	annotation := flag.Annotations[key]
+	return len(annotation) > 0 && annotation[0] == "true"
+}
+
+// atomicWrite is a helper to atomically write the data to the outfile.
+func atomicWrite(outfile string, mode os.FileMode, data []byte) (err error) {
```
jtolds: does this work right on windows? do we need atomic writes?

zeebo: yes, assuming it’s not replacing a directory.

jtolds: i think that's the Linux rules? on phone but Linux you can't rename over a directory iirc, and windows you need to remove the target first. what do we gain over just opening the destination directly for this?

zeebo: i went over this with andrew when we were pairing and we agreed it would work fine on windows. here’s the relevant code from the stdlib.  https://golang.org/src/internal/syscall/windows/syscall_windows.go#L259

indeed, our other code that does renames does the same thing except with `MOVEFILE_WRITE_THROUGH` (https://github.com/storj/storj/blob/master/storage/filestore/dir_windows.go#L124) which as per the docs at https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa should only matter if the move is performed as a copy, which this one cannot. 

in general, i prefer atomic file writing, especially for important values like secrets that may not be easily replicated.

jtolds: i am placated

PR #2504, opened 2019-07-09
storagenode: use minimum time in the order for expiration

zeebo:
What: Use the smaller of the two expiration times in the piecestore for the order limits 

Why: So that the database is much smaller.

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact: Should help with database load.

### DIFF
--- a/internal/testplanet/storagenode.go
+++ b/internal/testplanet/storagenode.go
@@ -115,7 +115,7 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatellites storj.Nod
 			Storage2: piecestore.Config{
 				ExpirationGracePeriod: 0,
 				MaxConcurrentRequests: 100,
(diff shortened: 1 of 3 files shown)


thepaul:
I can see that the existing entries will all fall off eventually, but maybe it's worth having a migration still that updates their expiration dates so that they can fall off sooner?

45 days is a long time to wait for the full benefit.

zeebo:
I wonder if we can just ask node operators to run `delete from used_serial` once if they want it to be smaller?

egonelbre:
We can just delete the used_serials, the creation date will mostly protect against replays.

zeebo:
Alright, should I add a migration to this that deletes from the table? might as well drop that index while I'm there.

zeebo:
https://github.com/storj/storj/pull/2508 is to do the migration/drop.

PR #2507, opened 2019-07-09
storagenode: batch archiving unsent_orders

zeebo:
What: Batch up archiving unsent_orders

Why: This avoids doing a transaction/fsync per order. Now it's 1000x less transactions.

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact: 1000x less transactions.


storagenode/orders/sender.go
```diff
@@ -120,15 +143,24 @@ func (sender *Sender) runOnce(ctx context.Context) (err error) {
 		ctx, cancel := context.WithTimeout(ctx, sender.config.Timeout)
 		defer cancel()
 
+		ch := make(chan archiveRequest, batchSize)
+		done := make(chan struct{})
```
egonelbre: Let's use errgroup here to propagate the error and make it clearer the goroutine is managed and waited for.

storagenode/orders/sender.go
```diff
@@ -137,15 +169,69 @@ func (sender *Sender) runOnce(ctx context.Context) (err error) {
...
+	defer func() {
+		for range ch {
+		}
+	}()
+
+	buffer := make([]archiveRequest, 0, cap(ch))
+
+	// Flush performs an archive on all of the requests in the buffer.
+	flush := func() (err error) {
+		txn, err := sender.orders.BeginArchive(ctx)
```
egonelbre: This implementation would be clearer if you passed in `[]ArchiveRequest` to the database.

wthorp: I agree wholeheartedly.  I think the channels, etc., needlessly hard to follow.

zeebo: I don’t think the channels are going away. This is about exposing a transaction handle as compared to exposing an api that can perform multiple actions against the database.

storagenode/orders/sender.go
```diff
@@ -201,20 +254,21 @@ func (sender *Sender) settle(ctx context.Context, log *zap.Logger, satelliteID s
...
-			err = sender.orders.Archive(ctx, satelliteID, response.SerialNumber, StatusRejected)
-			if err != nil {
-				errHandle(OrderError, "failed to archive order as rejected: serial: %v, %v", response.SerialNumber, err)
-			}
+			status = StatusRejected
 		default:
 			errHandle(OrderError, "unexpected response: %v", response.Status)
 		}
+
+		ch <- ArchiveRequest{
```
mniewrzal: won't order be archived even with `unexpected response`?

zeebo: indeed. fixing.

storagenode/storagenodedb/orders.go
```diff
@@ -145,10 +145,36 @@ func (db *ordersdb) ListUnsentBySatellite(ctx context.Context) (_ map[storj.Node
...
+			return ErrInfo.Wrap(err)
+		}
+	}
+
+	return nil
+}
+
+// archiveOne marks order as being handled.
+func (db *ordersdb) archiveOne(ctx context.Context, txn *sql.Tx, req orders.ArchiveRequest) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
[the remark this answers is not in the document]
zeebo: in my experience i've never been upset about a monitoring call existing, but i have frequently been upset about monitoring calls not existing.

it's probably not going to be useful data, but it's also not really harmful.

storagenode/storagenodedb/orders.go
```diff
@@ -146,10 +146,36 @@ func (db *ordersdb) ListUnsentBySatellite(ctx context.Context) (_ map[storj.Node
...
+	defer func() {
+		if err == nil {
+			err = txn.Commit()
+		} else {
+			err = errs.Combine(err, txn.Rollback())
+		}
+	}()
+
+	for _, req := range requests {
+		err := db.archiveOne(ctx, txn, req)
```
thepaul: Have you considered putting all the necessary statements together in one longish query? There isn't any real need for an explicit transaction block here, and this look like an awful lot of back-and-forth time. That might not matter-  I don't think anything much cares whether this is fast- but it could reduce database load.

The query could just be the concatenation of all the queries from archiveOne, or that query could be modified to have a clause like

    WHERE (satellite_id = ? AND serial_number = ANY(?, ?, ?, ?, ...))
        OR (satellite_id = ? AND serial_number = ANY(?, ?, ?, ...))

thepaul: sorry, clarification- that would only work on postgresql. I imagine this needs to work against sqlite still.

zeebo: There are probably better ways to accomplish this, but I just want to do the easiest thing that stops doing a full transaction/fsync per order. I also don't know if sqlite's query engine is capable enough to handle one huge query cheaper than many small queries.

Also, I think it only has to work against sqlite.

thepaul: Oh right this is storagenode. Oops. Then yeah, the extra transaction block and round trip time are non-issues. Sorry for noise

PR #2515, opened 2019-07-10
Satellite console auth message

crawter:
What: 
console.Users interface extended with new method to make an additional check
Why:
we should change error message when registered, but not verified user tries to login

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/service.go
```diff
@@ -466,6 +467,17 @@ func (s *Service) Token(ctx context.Context, email, password string) (token stri
 
 	user, err := s.store.Users().GetByEmail(ctx, email)
 	if err != nil {
+		users, err := s.store.Users().GetAllByEmail(ctx, email)
```
egonelbre: Why would there be multiple users with the same password?

crawter: @egonelbre we don't have unique constraint in users table for email column

at the moment we are able to create multiple users with same email until one of them confirm their email address.

also, in theory, there could be users with same email and password, because we can't prevent it

stefanbenten: We should add the constraint asap then.

satellite/console/users_test.go
```diff
@@ -54,7 +54,7 @@ func TestUserRepository(t *testing.T) {
 			ID:           testrand.UUID(),
 			FullName:     name,
 			ShortName:    lastName,
-			Email:        email,
+			Email:        "new" + email,
```
stefanbenten: Why do we need this change?

satellite/console/service.go
crawter: @stefanbenten 
imagine situation when someone will use ( accidentally or not ) your email to create own account.
To prevent it we should send confirmation email, to be sure that its email really yours.
But if we will have unique constraint on email column we will have unactivated account with such email.
And real owner of this email won't be able to create own account with his email.

satellite/console/users_test.go
crawter: @stefanbenten 
because we run tests twice. First time we add user with partnerID, second without. 
One of the tests is created to count users with specified email. 
after first call of testUsers we have 2 users with "email@mail.test". after second call - 3
so i should dynamically change expected number of users in my test. 
or change email of user that will be inserted in second run of testUsers function

satellite/console/service.go
```diff
@@ -475,6 +476,17 @@ func (s *Service) Token(ctx context.Context, email, password string) (token stri
 
 	user, err := s.store.Users().GetByEmail(ctx, email)
 	if err != nil {
```
mobyvb: should we keep track of this error for any of the cases below?

satellite/console/service.go
```diff
@@ -475,6 +476,17 @@ func (s *Service) Token(ctx context.Context, email, password string) (token stri
...
+		users, err := s.store.Users().GetAllByEmail(ctx, email)
+		if err != nil {
+			return "", errs.New(internalErrMsg)
+		}
+
+		for _, userToCheck := range users {
+			if err = bcrypt.CompareHashAndPassword(userToCheck.PasswordHash, []byte(password)); err == nil {
+				return "", errs.New(verifyEmailErrMsg)
```
mobyvb: If we return here, we skip checking all the rest of the users in the loop. I am assuming if any of the items pass the `CompareHashAndPassword` check, you do not want to return the error. That does not happen right now.

mobyvb: Also, why are we comparing passwords to determine whether to send `verifyEmailErrMsg`? If `bcrypt.CompareHashAndPassword(userToCheck.PasswordHash, []byte(password))` fails, does that definitely mean the user is not verified? Is there a different/better way to check?

PR #2522, opened 2019-07-10
satellite/satellitedb: add default offer for offers table

VinozzZ:
What: 
marketing team needs two default offer for the referral program
Why:
add insert into migration so the two default offer is auto generated

Please describe the tests:
 - Test 1:
 - Test 2:
 

VinozzZ:
> This isn't returning any offers for the `GET` handler

What do you mean?

fuskovic:
@VinozzZ when I run this in my local environment and log out the result of `db.ListAll()` it doesn't return any offers. I was expecting to get two offers - one for each offer type.

VinozzZ:
>  when I run this in my local environment and log out the result of `db.ListAll()` it doesn't return any offers. I was expecting to get two offers - one default offer for each offer type.
@fuskovic are you running storj-sim with postgres? if not, you won't see the change locally since storj-sim is using sqlite

satellite/rewards/rewards_test.go
```diff
@@ -59,27 +59,26 @@ func TestOffer_Database(t *testing.T) {
...
-				ID:        new.ID,
-				Status:    rewards.Done,
-				ExpiresAt: time.Now(),
-			}
+			//TODO: delete this test after removing Redeem method
 
-			isDefault := update.Status == rewards.Default
-			err = planet.Satellites[0].DB.Rewards().Redeem(ctx, update.ID, isDefault)
-			require.NoError(t, err)
+			//isDefault := new.Status == rewards.Default
```
coyle: why not just delete now?

satellite/rewards/rewards_test.go
```diff
@@ -59,27 +59,26 @@ func TestOffer_Database(t *testing.T) {
...
+			//current, err := planet.Satellites[0].DB.Rewards().ListAll(ctx)
+			//require.NoError(t, err)
+			//if current[i].Status == rewards.Default {
+			//	require.Equal(t, new.NumRedeemed, current[i].NumRedeemed)
+			//} else {
+			//	require.Equal(t, new.NumRedeemed+1, current[i].NumRedeemed)
+			//}
 
-			err = planet.Satellites[0].DB.Rewards().Finish(ctx, update.ID)
+			err = planet.Satellites[0].DB.Rewards().Finish(ctx, all[i].ID)
```
coyle: is it possible for `i` to be greater than the length of `all` ?

thepaul: Looks like no- we insert one offer in each iteration, so the length of "all offers" will be at least i+1

PR #2524, opened 2019-07-10
Stylistic Go Cleanup

jenlij:
What: I went searching for dead code and also ended up changing a bunch of things including variable name collisions with package names, unused variables and parameters, and other linty things. I didn't update all the files, but did reach a lot of them.

Why: Go style (as determined by goland) Let me know if there are objections.


pkg/overlay/cache_test.go
```diff
@@ -285,12 +285,12 @@ func TestIsVetted(t *testing.T) {
 		},
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
 		var err error
-		satellite := planet.Satellites[0]
-		satellite.Audit.Service.Loop.Pause()
-		satellite.Repair.Checker.Loop.Pause()
-		service := satellite.Overlay.Service
+		peer := planet.Satellites[0]
```
wthorp: why peer instead of satellite?  satellite is more specific

pkg/kademlia/utils_test.go
```diff
@@ -26,7 +26,7 @@ func TestSortByXOR(t *testing.T) {
 }
 
 func BenchmarkSortByXOR(b *testing.B) {
-	nodes := []storj.NodeID{}
+	var nodes []storj.NodeID
```
wthorp: nit: since k=1000 is known, `nodes := make([]storj.NodeID, 1000)` would be better

pkg/audit/cursor_test.go
```diff
@@ -36,7 +36,7 @@ func TestAuditSegment(t *testing.T) {
 		// change limit in library to 5 in
 		// list api call, default is  0 == 1000 listing
 		//populate metainfo with 10 non-expired pointers of test data
-		tests, cursor, metainfo := populateTestData(t, planet, time.Now().Add(3*time.Second))
+		tests, cursor, mi := populateTestData(t, planet, time.Now().Add(3*time.Second))
```
wthorp: why mi instead of metainfo?  metainfo is clearer to me.

pkg/overlay/cache_test.go
jenlij: the name satellite conflicts with the package satellite name and it has caused issues in the past... Goland recommended the name `peer` but I'm open to other options

pkg/audit/cursor_test.go
jenlij: same here about the variable/package name conflict

bootstrap/bootstrapdb/database.go
```diff
@@ -39,9 +39,9 @@ func New(config Config) (*DB, error) {
 	}, nil
 }
 
-// NewInMemory creates new inmemory master database for storage node
-// TODO: still stores data on disk
-func NewInMemory(storageDir string) (*DB, error) {
+// NewInMemory creates new in-memory master database for storage node
+// TODO: still stores data on disk, pass in storageDir string when updated
```
ifraixedes: Is this comment updated? the parameter `storageDir` isn't present anymore.

pkg/overlay/cache_test.go
wthorp: I feel like peer is overly generic.  _sat_ works for me; I'm not sure if others would find that to be clear enough.

pkg/audit/cursor_test.go
ifraixedes: it collides with an imported package identifier

pkg/overlay/cache_test.go
jenlij: Ok I like _sat_

pkg/kademlia/kademlia_test.go
```diff
@@ -150,7 +149,7 @@ func TestBootstrap(t *testing.T) {
 	assert.Len(t, nodeIDs, 3)
 }
 
-func testNode(ctx *testcontext.Context, name string, t *testing.T, bn []pb.Node) (*Kademlia, *grpc.Server, func()) {
+func testNode(ctx *testcontext.Context, t *testing.T, bn []pb.Node) (*Kademlia, *grpc.Server, func()) {
```
ifraixedes: I think that the `name` parameter shouldn't be removed and it should be used to enhance the assertion error message.

For example:

```go
lis, err := net.Listen("tcp", "127.0.0.1:0")
require.NoErrorf(t, err, "node: %s", name)
```

pkg/audit/containment.go
```diff
@@ -19,9 +19,6 @@ var (
 	// ErrContainedNotFound is the errs class for when a pending audit isn't found
 	ErrContainedNotFound = errs.Class("pending audit not found")
 
-	// ErrContainDelete is the errs class for when a pending audit can't be deleted
```
ifraixedes: Perhaps the error should exist and the real issues is that it has not been used where it should be; have you checked that?

pkg/kademlia/kademlia.go
```diff
@@ -27,8 +27,6 @@ import (
 var (
 	// NodeErr is the class for all errors pertaining to node operations
 	NodeErr = errs.Class("node error")
-	// BootstrapErr is the class for all errors pertaining to bootstrapping a node
-	BootstrapErr = errs.Class("bootstrap node error")
```
ifraixedes: Same than before, have you checked if the issues is that the error has not been used where it should be rather than it exists without reason?

pkg/kademlia/kademlia_test.go
```diff
@@ -258,7 +257,7 @@ func TestFindNear(t *testing.T) {
 		require.NoError(t, err)
 		return *n
 	}
-	nodeIDA := newNode("AAAAA", 1, 4)
+	var nodeIDA = newNode("AAAAA", 1, 4)
```
ifraixedes: I'm just wondering why this is much better than using `:=`?

pkg/kademlia/peer_discovery.go
```diff
@@ -29,9 +28,6 @@ type peerDiscovery struct {
 	queue discoveryQueue
 }
 
-// ErrMaxRetries is used when a lookup has been retried the max number of times
-var ErrMaxRetries = errs.Class("max retries exceeded for id:")
```
ifraixedes: Same than before, have you checked if the issues is that the error has not been used where it should be?

pkg/storj/nodeurl.go
```diff
@@ -116,8 +116,8 @@ func ParseNodeURLs(s string) (NodeURLs, error) {
 // String converts NodeURLs to a string
 func (urls NodeURLs) String() string {
 	var xs []string
-	for _, url := range urls {
-		xs = append(xs, url.String())
+	for _, nodeURL := range urls {
```
ifraixedes: nit: because it's one line in the for loop instruction I would use just `u` as a var name.

pkg/kademlia/kademlia_test.go
jenlij: Goland recommends the change but I don't know which is objectively better. It may be that Goland defaults to var since := has some exceptions but don't apply here.

jenlij:
@ifraixedes RE `could you add these linting rules to our golangci-lint configuration file so everybody could have those checks independently if they use GoLand or not and ultimately the CI will check those?` Great idea! I will see if I can do that! If not, I can add them to the style guide at least.

pkg/kademlia/kademlia_test.go
jenlij: Turns out @egonelbre says the original is more idomatic actually... I can change them back

egonelbre:
PS: avoid merging anything here that conflicts with https://github.com/storj/storj/pull/2525

jenlij:
https://github.com/storj/docs/pull/93

pkg/audit/containment.go
jenlij: good catch, I added this back and used it in satellitedb/containment.go

pkg/kademlia/peer_discovery.go
jenlij: I'm not sure where this would be used

pkg/audit/disqualification_test.go
```diff
@@ -49,14 +49,14 @@ func TestDisqualificationTooManyFailedAudits(t *testing.T) {
 		require.NoError(t, err)
 
 		var (
-			sat    = planet.Satellites[0]
-			nodeID = planet.StorageNodes[0].ID()
-			report = &audit.Report{
+			satellitePeer = planet.Satellites[0]
```
egonelbre: For the `sat`, `satellite` rename to `satellitePeer`. Either seemed better, because I don't think adding the `Peer` adds any new information to the variable name. `sat` was unique enough, I don't think we have conflicts with other things... I would keep it as `satellite`.

pkg/audit/disqualification_test.go
```diff
@@ -118,42 +118,42 @@ func TestDisqualifiedNodesGetNoDownload(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		satellite := planet.Satellites[0]
-		upl := planet.Uplinks[0]
+		satellitePeer := planet.Satellites[0]
+		uplinkPeer := planet.Uplinks[0]
```
egonelbre: `upl` yeah doesn't pass the telephone test, so there should be a better variable. `uplinkPeer` seems long similiarly to `satellitePeer`, but I don't have a better suggestion for it yet.

pkg/kademlia/peer_discovery.go
ifraixedes: Ok, then fine, it can go away.

pkg/audit/disqualification_test.go
ifraixedes: in this case `satellite` masquerades a package identifier.
In my opinion `sat` is fine.

We could think to add our business domain abbreviations into a document (kind of a glossary), not all at once but have the document and start to add the ones that we start to use and are clear for all of use without that much discussion.

pkg/audit/disqualification_test.go
ifraixedes: No idea of a better name, but I have this in mind `ulnk` however I don't call it for being better than the ones used.

pkg/audit/cursor_test.go
```diff
@@ -112,9 +112,9 @@ func TestDeleteExpired(t *testing.T) {
 		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
 		//populate metainfo with 10 expired pointers of test data
-		_, cursor, metainfo := populateTestData(t, planet, time.Now().Add(-time.Second*1))
+		_, cursor, metainfoService := populateTestData(t, planet, time.Now().Add(-time.Second*1))
```
ifraixedes: `svc` always sounded to me a good abbreviation for `service`.

Not asking for a change here, just commenting to find out what others think about.

pkg/audit/disqualification_test.go
egonelbre: Yeah, the naming conflict in terms of package identifier will be inevitable, since there are so many packages. Of course, having that naming conflict is not ideal. I haven't yet gotten confused with `satellite` package and variable name... however with `uplink` I have. So not quite sure what the best balance is.

ifraixedes: There are many packages, but normally only a few collide, I don't think that it's very hard to find another name without clashing with a package identifier.

I think that it's hard to convey to conventions where some people see that a specific thing confuse them while not to others, straight and tangible conventions are easier to follow and contribute to be more respected.

For this specific case, `satellitePeer` looks better to me than `satellite` because of the last added convention in our style guide despite it's a long name because it turns that we have quite a few long named variables in our sources and they have never been a problem.

pkg/audit/cursor_test.go
jenlij: metainfoSvc sounds good to me

PR #2525, opened 2019-07-10
protocol: implement new piece signing and verification

egonelbre:
Previous protocol relies on uplink identity and certificates to sign orders and piece hashes, however this means we need to keep track of all uplink certificates. This is problematic since uplink frequently regenerate their identity, so there will be too many of them.

More efficient approach is that the:

1. Satellite generates a new key-pair for each piece.
2. The piece public key is included in the order limit and signed by the satellite
3. The piece private key is included in the create segment, download segment or delete a segment.
4. The uplink uses piece private key to sign the orders.
5. The storage node verifies order limit signature as usual and then uses the piece public key to verify the order and piece hash.
6. Storage node sends both the order limit and order to the satellite during settlement.
7. Storage node sends data, order limit and piece hash to the satellite, if it needs to prove data-consistency to the satellite.

#2526 contains just the PiecePrivateKey and PiecePublicKey implementation, to reduce size of this PR.

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/orders.proto
```diff
@@ -28,8 +28,10 @@ message OrderLimit {
     // satellite who issued this order limit allowing orderer to do the specified action
     bytes satellite_id = 2 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
     // uplink who requested or whom behalf the order limit to do an action
-    bytes uplink_id = 3 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
-    // storage node who can reclaim the order limit specified by serial
+    bytes uplink_id = 3 [(gogoproto.customtype) = "NodeID"]; // deprecated
```
zeebo: Would it be a huge diff to change the name to `deprecated_uplink_id`? It may also help with being sure that we've stopped using it in the appropriate spots.

pkg/pb/orders.proto
```diff
@@ -64,15 +65,14 @@ message PieceHash {
...
-    bytes signature = 3;
     // size of uploaded piece
-    int64 piece_size = 4;
-    // timestamp when upload occur
-    google.protobuf.Timestamp timestamp = 5 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    int64 piece_size = 3;
+    // timestamp when upload occurred
+    google.protobuf.Timestamp timestamp = 4 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    // signature either satellite or storage node
+    bytes signature = 5;
```
zeebo: Why make these changes?

go.sum
```diff
@@ -404,6 +404,7 @@ golang.org/x/crypto v0.0.0-20180910181607-0e37d006457b/go.mod h1:6SG95UA2DQfeDnf
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8 h1:1wopBVtVdWnn03fZelqdXTqk7U7zPQCb+T4rbU9ZEoU=
 golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
+golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4 h1:HuIa8hRrWRSrqYzx1qI49NNxhdi2PrY7gxVSq1JjLDc=
```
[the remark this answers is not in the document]
egonelbre: it's downloading a newer version

pkg/pb/orders.proto
egonelbre: done

pkg/pb/orders.proto
egonelbre: Just to make it look pretty, and these ones we still can change.

satellite/orders/service.go
```diff
@@ -475,36 +484,43 @@ func (service *Service) CreateAuditOrderLimit(ctx context.Context, auditor *iden
...
 	err = service.saveSerial(ctx, serialNumber, bucketID, orderExpiration)
 	if err != nil {
-		return nil, Error.Wrap(err)
+		return nil, storj.PiecePrivateKey{}, Error.Wrap(err)
 	}
 
 	projectID, bucketName, err := SplitBucketID(bucketID)
 	if err != nil {
-		return limit, err
+		return limit, piecePrivateKey, Error.Wrap(err)
```
thepaul: Why do we return `piecePrivateKey` in this error case, but not in the other error cases? Is there something special going on here?

storagenode/piecestore/verification.go
```diff
@@ -104,18 +92,18 @@ func (endpoint *Endpoint) VerifyOrder(ctx context.Context, peer *identity.PeerId
 		return ErrProtocol.New("order exceeded allowed amount=%v, limit=%v", order.Amount, limit.Limit) // TODO: report grpc status bad message
 	}
 
-	if err := signing.VerifyOrderSignature(ctx, signing.SigneeFromPeerIdentity(peer), order); err != nil {
-		return ErrVerifyUntrusted.New("invalid order signature") // TODO: report grpc status bad message
+	if err := signing.VerifyUplinkOrderSignature(ctx, limit.UplinkPublicKey, order); err != nil {
+		return ErrProtocol.Wrap(err)
```
thepaul: Why does this error case change from ErrVerifyUntrusted to ErrProtocol?

satellite/orders/service.go
egonelbre: yeah, I'll make it empty

storagenode/piecestore/verification.go
egonelbre: fixed

pkg/auth/signing/sign_test.go
```diff
@@ -0,0 +1,37 @@
...
+	require.Equal(t, orderLimitBytes, orderLimitMarshaled, "order limit marshaling changed")
+
+	order := pb.Order{}
+	err = proto.Unmarshal(orderBytes, &order)
+	require.NoError(t, err)
+
+	orderBytesMarshaled, err := proto.Marshal(&order)
+	require.NoError(t, err)
+	require.Equal(t, orderBytes, orderBytesMarshaled, "order marshaling changed")
+}
```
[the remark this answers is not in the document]
egonelbre: fixed

pkg/auth/signing/verify.go
```diff
@@ -38,6 +38,17 @@ func VerifyOrderSignature(ctx context.Context, uplink Signee, signed *pb.Order)
 	return uplink.HashAndVerifySignature(ctx, bytes, signed.UplinkSignature)
 }
 
+// VerifyUplinkOrderSignature verifies that the signature inside order belongs to the uplink.
```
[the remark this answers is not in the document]
egonelbre: fixed

pkg/auth/signing/marshaling_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/pkg/pb"
+)
+
+func TestOrderSerialization(t *testing.T) {
+	t.Skip("broken")
```
thepaul: 😢

egonelbre: Yeah, it needs separate fixes to the serialization to make it compatible. It was already broken before the current PR.

pkg/auth/signing/verify.go
```diff
@@ -38,6 +38,17 @@ func VerifyOrderSignature(ctx context.Context, uplink Signee, signed *pb.Order)
...
+func VerifyUplinkOrderSignature(ctx context.Context, publicKey storj.PiecePublicKey, signed *pb.Order) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	bytes, err := EncodeOrder(ctx, signed)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	return Error.Wrap(publicKey.Verify(bytes, signed.UplinkSignature))
```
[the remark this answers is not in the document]
thepaul: For posterity: we have decided that we can probably count on marshaled messages remaining unchanged for the near future, as long as all participating endpoints are using Gogo stable marshaling. This is decidedly non-ideal, but the risk/reward ratio is expected to be lower this way than if we try to make all the necessary changes to use a more signed-envelope-type protobuf structure.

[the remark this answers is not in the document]
thepaul: If the encoding suddenly changes, we'll know it essentially right away, and we'll have to decide whether to fork protobufs or gogo or write a custom serializer for these few specific types or what.

pkg/auth/signing/marshaling_test.go
thepaul: But does that mean the serialization has already changed and it might change again? That would be not good.

PR #2533, opened 2019-07-11
Order by attempted time in injured segments select

mobyvb:
What: 
Order by `attempted` rather than `path` in the satellitedb repair queue. In postgres, we want to make sure null values are first. Add an index on `attempted`

Why:
There is no logic to ordering in repair queue currently. We want to give all injured segments a fair chance

Please describe the tests:
 - Test 1:
 - Test 2:
 

navillasa:
I feel like this needs a test to make sure the queries are getting the injured segments in the order they want. worst case you could move the query changes + tests in a new pr.

mobyvb:
@navillasa yeah I'd like that too. I don't think that would pass with sqlite3 though. But this at least adds coverage to ensure that every segment is selected exactly once when a bunch of fresh pieces are added

PR #2534, opened 2019-07-11
Metainfo RPC objects methods

mniewrzal:
What: Part of Metainfo RPC refactoring for objects manipulation and listing.

Why: 

Please describe the tests:
 - Test 1: automatic test added
 

satellite/metainfo/metainfo.go
```diff
@@ -847,3 +852,239 @@ func convertBucketToProto(ctx context.Context, bucket storj.Bucket) (pbBucket *p
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	// take bucket RS values if not set in request
+	pbRS := req.RedundancyScheme
+	if pbRS.ErasureShareSize == 0 {
+		pbRS.ErasureShareSize = bucket.DefaultRedundancyScheme.ShareSize
+	}
+	if pbRS.MinReq == 0 {
```
egonelbre: I wonder whether this check should be: `if pbRS.MinReq == 0 && pbRS.RepairThreshold == 0 ... { use default }` and then `if pbRS.MinReq == 0 || pbRS.RepairThreshold == 0 { return invalid arguments error }`. That way if you only partially supply the values it will fail, rather than default them to something. But we can figure out this in a later PR.

satellite/metainfo/metainfo.go
```diff
@@ -847,3 +852,239 @@ func convertBucketToProto(ctx context.Context, bucket storj.Bucket) (pbBucket *p
...
+		Op:            macaroon.ActionWrite,
+		Bucket:        streamID.Bucket,
+		EncryptedPath: streamID.EncryptedPath,
+		Time:          time.Now(),
+	})
+	if err != nil {
+		return nil, status.Errorf(codes.Unauthenticated, err.Error())
+	}
+
+	// no logic for shim implementation
```
egonelbre: add `// TODO: `

satellite/metainfo/metainfo.go
mniewrzal: yeah, I was not sure if we should take all or nothing. for now I follow approach from libuplink https://github.com/storj/storj/blob/ae8b9698f936f65972b83f2887a91bc1d93384ff/lib/uplink/bucket.go#L137

satellite/metainfo/metainfo.go
```diff
@@ -847,3 +852,239 @@ func convertBucketToProto(ctx context.Context, bucket storj.Bucket) (pbBucket *p
...
+
+	prefix, err := CreatePath(ctx, keyInfo.ProjectID, -1, req.Bucket, req.EncryptedPrefix)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	metaflags := meta.All
+	// TODO use flags
+	// TODO find out how EncryptedCursor -> startAfter/endAfter
+	segments, more, err := endpoint.metainfo.List(ctx, prefix, "", "", false, req.Limit, metaflags)
```
egonelbre: metainfo service internally has a better way to list things, https://github.com/storj/storj/blob/master/satellite/metainfo/service.go#L86.

satellite/metainfo/metainfo.go
mniewrzal: its just comment that with shim implementation we won't do anything specific here

egonelbre: ah, maybe make it more explicit. `// we don't need to do anything for shim implementation`

satellite/metainfo/metainfo.go
mniewrzal: hmm I don't see big difference, why it's better? because it combines params in a struct?

pkg/pb/metainfo.proto
```diff
@@ -196,3 +203,129 @@ message ProjectInfoRequest {
...
+    bytes  bucket = 1;
+    bytes  encrypted_path = 2;
+    int32  version = 3;
+
+    pointerdb.RedundancyScheme redundancy = 4;
+
+    google.protobuf.Timestamp creation_date = 5  [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    google.protobuf.Timestamp expiration_date = 6  [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+
+    bytes satellite_signature = 7;
```
egonelbre: maybe let's add a stream id expiration here as well, to avoid them living for too long?

satellite/metainfo/metainfo.go
```diff
@@ -847,3 +852,239 @@ func convertBucketToProto(ctx context.Context, bucket storj.Bucket) (pbBucket *p
...
+	encodedStreamID, err := proto.Marshal(satStreamID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, err.Error())
+	}
+
+	streamID, err := storj.StreamIDFromBytes(encodedStreamID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, err.Error())
+	}
+
```
egonelbre: Will the deleting segments come in an other PR?

pkg/pb/metainfo.proto
mniewrzal: make sense

satellite/metainfo/metainfo.go
mniewrzal: yes, to avoid too large PR

uplink/metainfo/client.go
```diff
@@ -341,3 +341,100 @@ func convertProtoToBucket(pbBucket *pb.Bucket) storj.Bucket {
...
+		Bucket:          bucket,
+		EncryptedPrefix: encryptedPrefix,
+		EncryptedCursor: encryptedCursor,
+		Limit:           limit,
+	})
+	if err != nil {
+		return []storj.Object{}, false, Error.Wrap(err)
+	}
+
+	// TODO maybe we should just return pb.ObjectListItem
```
egonelbre: You can make a separate `ObjectListItem` into `storj` package.

satellite/metainfo/metainfo.go
egonelbre: It does some less guessing in how to interpret the arguments. But, either is fine for now.

pkg/pb/metainfo.proto
mniewrzal: or we can just use `creation_date` and add e.g. 2h check?

egonelbre: Large uploads might fail with it.

mniewrzal: ok it can be 24h, one way or another we need to choose time for the expiration

pkg/auth/signing/encode.go
```diff
@@ -88,3 +88,13 @@ func EncodeVoucher(ctx context.Context, voucher *pb.Voucher) (_ []byte, err erro
 	voucher.SatelliteSignature = signature
 	return out, err
 }
+
+// EncodeStreamID encodes stream ID into bytes for signing.
+func EncodeStreamID(ctx context.Context, streamID *pb.SatStreamID) (_ []byte, err error) {
+	defer mon.Task()(&ctx)(&err)
+	signature := streamID.SatelliteSignature
+	streamID.SatelliteSignature = nil
```
kaloyan-raev: I wonder if this could cause race issues. What if another goroutine is trying to access the SatelliteSignature while we are marshaling the `streamID`? Wouldn't it be safer if we copy `streamID` to a new struct and then marshal it?

pkg/pb/metainfo.proto
```diff
@@ -196,3 +203,129 @@ message ProjectInfoRequest {
 message ProjectInfoResponse {
     bytes project_salt = 1;
 }
+
+message Object {
```
kaloyan-raev: Is this a replacement for the `Pointer` protobuf? If yes, where is the info for remote pieces stored?

mniewrzal: `Segment` is part of `Object` and it contains `nodes`

pkg/storj/streamid_test.go
```diff
@@ -6,28 +6,28 @@ package storj_test
...
-
 	for i := 0; i < 10; i++ {
-		streamID := testrand.StreamID()
+		expectedSize := testrand.Intn(255)
+		streamID := testrand.StreamID(expectedSize)
 
 		fromString, err := storj.StreamIDFromString(streamID.String())
-		assert.NoError(t, err)
+		require.NoError(t, err)
+		require.Equal(t, streamID.String(), fromString.String())
```
kaloyan-raev: Aren't we overusing `require` over `assert`?

`require.NoError` is usually good as it prevents the test from panicking. But `require.Equal` interrupts the tests too early. If we have a failing test (especially flaky one) the result from the rest of the checks might be very valuable in analyzing the issue.

pkg/pb/metainfo.proto
kaloyan-raev: Where can I see the `Segment` protobuf? I also don't see any reference to segments here in the `Object` protobuf. Is this `stream_id`?

pkg/pb/metainfo.proto
```diff
@@ -196,3 +203,129 @@ message ProjectInfoRequest {
...
+    bytes  encrypted_metadata_nonce = 9 [(gogoproto.customtype) = "Nonce", (gogoproto.nullable) = false];
+    bytes  encrypted_metadata       = 10;
+
+    int64                           fixed_segment_size    = 11;
+    pointerdb.RedundancyScheme      redundancy_scheme     = 12;
+    encryption.EncryptionParameters encryption_parameters = 13;
+
+    int64 total_size  = 14;
+    int64 inline_size = 15;
+    int64 remote_size = 16;
```
kaloyan-raev: These 3 are quite confusing. Can we have a comment which of them is what?

pkg/storj/streamid_test.go
mniewrzal: personally, I prefer to fail-fast but I agree that with more complex test it might be useful to have results from multiple assertions. In this case rerunning test is a matter of seconds so having only first error is good enough for me.

pkg/pb/metainfo.proto
mniewrzal: its in design doc https://github.com/storj/storj/blob/6aefa1ae809d9df083f36f58a812bce91e68380c/docs/design/metainfo-rpc.md#L222

pkg/pb/metainfo.proto
mniewrzal: :+1:

pkg/pb/metainfo.proto
kaloyan-raev: Are we going to introduce it with a separate PR?

mniewrzal: yes, I'm working on that but its hard to push it without changes from this PR

kaloyan-raev: :+1:

pkg/auth/signing/encode.go
mniewrzal: not sure tbh, I guess will have a problem always when we will try to use `streamID` in different gouroutines.

PR #2542, opened 2019-07-12
signing: ensure we don't break signatures

egonelbre:
What: protobuf serialization changed accidentally ensure that we always can verify older order limits signatures

Why:

Please describe the tests:
 - Test 1: Added regression tests for order limit, order and piece hash
 - Test 2:
 

pkg/pb/orders.proto
```diff
@@ -50,6 +50,36 @@ message OrderLimit {
     node.NodeAddress satellite_address = 11;
 }
 
+// OrderLimitSigning provides OrderLimit signing serialization
+message OrderLimitSigning {
```
thepaul: Could we have a little bit more explanation here also? For example, maybe something like "OrderLimitSigning messages are never directly stored or sent across the network; they are only used when calculating a signature over an OrderLimit. Its purpose is to solidify the format of how we serialize OrderLimits for signing, to handle some backwards compatibility considerations."?

pkg/pb/compatibility_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+		tag := f.Tag.Get("protobuf")
+		if tag == "" {
+			continue
+		}
+		tags := strings.Split(tag, ",")
+		fields = append(fields, Field{
+			Name:  f.Name,
+			Type:  tags[0],
+			Index: tags[1],
```
thepaul: Is it necessary that the index numbers of the fields in X and XSigning stay in sync? As far as I can tell, the only thing that matters is that XSigning continues to match the format of signed X messages from before this change (and that appears to be sufficiently covered by sign_test.go).

egonelbre: Not necessarily, however it's quite likely that someone adds a field to one and forgets to update the other.

thepaul: Definitely it's worth checking that no new fields are added and that the types don't change- otherwise we have a security flaw. But indexes are just 🤷‍♂. Oh well I'm fine with checking them.

PR #2543, opened 2019-07-12
storagenode/nodestats: cache node stats

rikysya:
What:  Cache node_stats and disk space usage from the satellite

Why: To reduce amount of calls to satellites

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/pb/nodestats.proto
```diff
@@ -36,6 +36,7 @@ message DailyStorageUsageRequest {
 
 message DailyStorageUsageResponse {
     message StorageUsage {
+        int64 rollup_id = 1;
```
egonelbre: Isn't timestamp sufficient for this?

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,160 @@ func (c *Client) Close() error {
...
 
 // NewService creates new instance of service
-func NewService(log *zap.Logger, transport transport.Client, kademlia *kademlia.Kademlia) *Service {
+func NewService(log *zap.Logger, transport transport.Client, consoleDB console.DB, kademlia *kademlia.Kademlia) *Service {
 	return &Service{
-		log:       log,
-		transport: transport,
-		kademlia:  kademlia,
+		log:         log,
+		statsTicker: time.NewTicker(time.Duration(time.Second * 30)),
```
egonelbre: no need for the cast

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,160 @@ func (c *Client) Close() error {
...
+		spaceTicker: time.NewTicker(time.Duration(time.Second * 60)),
+		transport:   transport,
+		consoleDB:   consoleDB,
+		kademlia:    kademlia,
+	}
+}
+
+// RunStatsLoop continuously queries satellite for stats info with some interval
+func (s *Service) RunStatsLoop(ctx context.Context) (err error) {
```
egonelbre: use the same approach as other services

pkg/pb/nodestats.proto
```diff
@@ -36,6 +36,7 @@ message DailyStorageUsageRequest {
...
         double at_rest_total = 2;
         google.protobuf.Timestamp time_stamp = 3 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
egonelbre: timestamp is one word

storagenode/console/service.go
```diff
@@ -189,27 +200,17 @@ func (s *Service) GetUptime(ctx context.Context) time.Duration {
 }
 
 // GetStatsFromSatellite returns storagenode stats from the satellite
-func (s *Service) GetStatsFromSatellite(ctx context.Context, satelliteID storj.NodeID) (_ *nodestats.Stats, err error) {
+func (s *Service) GetStatsFromSatellite(ctx context.Context, satelliteID storj.NodeID) (err error) {
```
egonelbre: should these be empty?

storagenode/peer.go
```diff
@@ -364,6 +364,13 @@ func (peer *Peer) Run(ctx context.Context) (err error) {
 		return errs2.IgnoreCanceled(peer.Server.Run(ctx))
 	})
 
+	group.Go(func() error {
+		return errs2.IgnoreCanceled(peer.NodeStats.RunStatsLoop(ctx))
```
egonelbre: `return errs2.IgnoreCanceled(peer.NodeStats.Run(ctx))`

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,160 @@ func (c *Client) Close() error {
...
+			return ctx.Err()
+		// wait for the next interval to happen
+		case <-s.statsTicker.C:
+		}
 	}
 }
 
+// CacheSpaceUsageFromSatellites queries disk space usage from all the satellites
+// known to the storagenode and stores this information into db
+func (s *Service) CacheSpaceUsageFromSatellites(ctx context.Context) (err error) {
```
egonelbre: Why isn't there only a single ticker loop?

storagenode/storagenodedb/consoledb.go
```diff
@@ -57,6 +57,155 @@ func (db *consoledb) GetSatelliteIDs(ctx context.Context, from, to time.Time) (_
...
+	}
+
+	return &stats, nil
+}
+
+// UpdateStats updates stored stats
+func (db *consoledb) UpdateStats(ctx context.Context, stats console.Stats) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	stmt := db.Rebind(`UPDATE node_stats
```
egonelbre: No need for rebind, because this is sqlite only.

storagenode/console/service.go
```diff
@@ -32,12 +31,26 @@ type DB interface {
 	// GetSatelliteIDs returns list of satelliteIDs that storagenode has interacted with
 	// at least once
 	GetSatelliteIDs(ctx context.Context, from, to time.Time) (storj.NodeIDList, error)
+	// CreateStats inserts new stats into the db
```
egonelbre: the `DB` interface is getting really large without clear separation

storagenode/storagenodedb/consoledb.go
```diff
@@ -153,6 +302,113 @@ func (db *consoledb) getDailyBandwidthUsed(ctx context.Context, cond string, arg
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err != nil {
+			err = errs.Combine(err, tx.Rollback())
+		}
+
+		err = tx.Commit()
```
egonelbre: this overwrites the previous non-nil error

pkg/pb/nodestats.proto
rikysya: When I tested I had 170 roll up entries for 10 storagenodes just for one day. So I am not sure if storagenode can trust satellite here. But yeah, maybe it’s sufficient just to replace old value

storagenode/nodestats/service.go
rikysya: We discussed that we should have to separate intervals, as nodestats is updated more frequently and is less costly than rollups query

storagenode/storagenodedb/consoledb.go
rikysya: Yeah, forgot to put `return` here

pkg/pb/nodestats.proto
stefanbenten: I agree with Egon, i dont think we need a rollup_id

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -195,18 +195,23 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
-// QueryNodeDailySpaceUsage returns slice of NodeSpaceUsage for given period
-func (db *StoragenodeAccounting) QueryNodeDailySpaceUsage(ctx context.Context, nodeID storj.NodeID, start time.Time, end time.Time) (_ []accounting.NodeSpaceUsage, err error) {
+// QueryNodeStorageUsage returns slice of NodeStorageUsage for given period
+func (db *StoragenodeAccounting) QueryNodeStorageUsage(ctx context.Context, nodeID storj.NodeID, start time.Time, end time.Time) (_ []accounting.NodeStorageUsage, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	// as entries are stored on daily basis we don't need
-	// to extract DATE from start_time
-	query := `SELECT at_rest_total, start_time 
+	query := `SELECT id, at_rest_total, start_time,
```
stefanbenten: this query needs to change if id is removed.

satellite/satellitedb/utils.go
```diff
@@ -24,6 +25,26 @@ func bytesToUUID(data []byte) (uuid.UUID, error) {
 	return id, nil
 }
 
+// null time helps convert nil to time.Time
+type nullTime struct {
```
stefanbenten: This and its helper functions should go into our date helper package in internal

storagenode/console/satellites.go
```diff
@@ -0,0 +1,18 @@
...
+	"time"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Satellites queries satellite related info from db
+type Satellites interface {
+	// GetIDs returns list of satelliteIDs that storagenode has interacted with
+	// at least once
+	GetIDs(ctx context.Context, from, to time.Time) (storj.NodeIDList, error)
```
stefanbenten: Why do we need a time range for this?

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,133 @@ func (c *Client) Close() error {
...
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get disk space usage query failed: %v", err))
+	}
+
+	for {
+		select {
+		// handle cancellation signal first
+		case <-ctx.Done():
+			return ctx.Err()
+		// wait for the next cache interval to happen
```
stefanbenten: _next stats interval_

storagenode/storagenodedb/consoledb.go
```diff
@@ -26,17 +26,37 @@ func (db *InfoDB) Console() console.DB { return &consoledb{db} }
...
 
 	var satellites storj.NodeIDList
 
-	rows, err := db.db.QueryContext(ctx, db.Rebind(`
-		SELECT DISTINCT satellite_id
+	rows, err := db.db.QueryContext(ctx,
+		`SELECT DISTINCT satellite_id
 		FROM bandwidth_usage
-		WHERE ? <= created_at AND created_at <= ?`), from.UTC(), to.UTC())
+		WHERE ? <= created_at AND created_at <= ?`, from.UTC(), to.UTC())
```
stefanbenten: I dont think we need that here. Query all and ensure we get stats for all of them.
Besides that the boundries are both selecting if the time is lower or equal then X. One needs to be greater or equal.

satellite/satellitedb/storagenodeaccounting.go
rikysya: `ID` is no longer a primary key here. It is stored to be able to reference appropriate `rollup` entry if needed

satellite/satellitedb/utils.go
rikysya: `internal/dbutil` is it appropriate place to move this struct?

storagenode/console/satellites.go
rikysya: Because ids is gathered from bandwidth table, selecting only distinct ones. If you change time window, your result list will also change

storagenode/storagenodedb/consoledb.go
rikysya: `before <= created_at <= after`. After is greater and before is lower, no? Or I didn't understand what you mean

rikysya: Also If I want monthly stats, do I still need all of the satellites on Dashboard?

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -195,18 +195,23 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
+			SELECT MAX(id)
+			FROM accounting_rollups
+			WHERE node_id = ?
+			AND ? <= start_time AND start_time <= ?
+			GROUP BY start_time
+			ORDER BY start_time ASC
+		)
+		WINDOW win AS (ORDER BY ID)`
 
 	rows, err := db.db.QueryContext(ctx, db.db.Rebind(query), nodeID, start, end)
```
simongui: Is this using UTC here?

storagenode/console/satellites.go
stefanbenten: Doesnt that mean that if i store data for satellite X and it didnt download or upload new data to me, it wont get selected? I think that is very sensitive to failures and/or miscalculations

satellite/satellitedb/storagenodeaccounting.go
stefanbenten: I think we should remove the ID here and in the protobuf as mentioned above.
I cannot find a line where we use the ID, so it does not have any benefit (yet).

storagenode/nodestats/service.go
stefanbenten: Yeah, i agree having two is better in this workflow!

storagenode/storagenodedb/consoledb.go
stefanbenten: Oh, my fault. I misread the order.
The other question, why to we only focus on satellites that you have interacted with? If you just store data and dont get a single audit, upload or download, the satellite would be missing, I am voting for querying all satellites here.

rikysya: Done, the question was, show all or only those that were active during suggested time

storagenode/console/satellites.go
rikysya: Yeah, let’s agree on all satellites, I have changed code to it already

satellite/satellitedb/storagenodeaccounting.go
rikysya: I was wondering if by anyway you have two rollups, for the same time, and resulting one has wrong data, then we can use this id to reference where it came from. But yes, it is not used anywhere now. So I am going to change that

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,133 @@ func (c *Client) Close() error {
...
+	defer mon.Task()(&ctx)(&err)
+
+	err = s.CacheStatsFromSatellites(ctx)
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get stats query failed: %v", err))
+	}
+
+	err = s.CacheSpaceUsageFromSatellites(ctx)
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get disk space usage query failed: %v", err))
```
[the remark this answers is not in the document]
rikysya: Resulting log is more readable this way

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,133 @@ func (c *Client) Close() error {
...
 	}
 }
 
+// Run runs loop
+func (s *Service) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	err = s.CacheStatsFromSatellites(ctx)
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get stats query failed: %v", err))
```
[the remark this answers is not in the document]
rikysya: Resulting log is more readable this way

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,133 @@ func (c *Client) Close() error {
...
+	for {
+		select {
+		// handle cancellation signal first
+		case <-ctx.Done():
+			return ctx.Err()
+		// wait for the next stats interval to happen
+		case <-s.statsTicker.C:
+			err = s.CacheStatsFromSatellites(ctx)
+			if err != nil {
+				s.log.Error(fmt.Sprintf("Get stats query failed: %v", err))
```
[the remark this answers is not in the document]
rikysya: Resulting log is more readable this way

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return stamps, nil
+}
+
+// GetDailyTotal returns daily disk usage summed across all known satellites
+// for provided time range
+func (db *diskSpaceUsage) GetDailyTotal(ctx context.Context, from, to time.Time) (_ []console.DiskSpaceUsage, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	query := `SELECT SUM(at_rest_total), timestamp
```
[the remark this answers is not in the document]
rikysya: no, as it is SQLite only

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return db.withTx(ctx, cb)
+}
+
+// GetDaily returns daily disk usage for particular satellite
+// for provided time range
+func (db *diskSpaceUsage) GetDaily(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) (_ []console.DiskSpaceUsage, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	query := `SELECT *
```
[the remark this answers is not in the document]
rikysya: no, as it is SQLite only

storagenode/nodestats/service.go
```diff
@@ -162,17 +253,33 @@ func (s *Service) DialNodeStats(ctx context.Context, satelliteID storj.NodeID) (
...
+func (s *Service) Close() error {
+	defer mon.Task()(nil)(nil)
+	s.statsTicker.Stop()
+	s.spaceTicker.Stop()
+	return nil
+}
+
+// getMonthEdges extract month from the provided date and returns its edges
+func getMonthEdges(t time.Time) (time.Time, time.Time) {
+	startDate := time.Date(t.Year(), t.Month(), 1, 0, 0, 0, 0, t.Location())
```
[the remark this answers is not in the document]
rikysya: It is a Date, no?

storagenode/storagenodedb/consoledb.go
```diff
@@ -60,44 +79,42 @@ func (db *consoledb) GetSatelliteIDs(ctx context.Context, from, to time.Time) (_
...
 	defer mon.Task()(&ctx)(&err)
 
-	query := db.Rebind(`
-		SELECT action, SUM(amount), created_at
-		FROM bandwidth_usage
-		` + cond + `
-		GROUP BY DATE(created_at), action
-		ORDER BY created_at ASC
-	`)
+	query := `SELECT action, SUM(amount), created_at
```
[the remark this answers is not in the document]
rikysya: See Egon comments below

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Store stores disk space usage stamps to db
+func (db *diskSpaceUsage) Store(ctx context.Context, stamps []console.DiskSpaceUsage) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if len(stamps) == 0 {
+		return nil
+	}
+
+	stmt := `INSERT OR REPLACE INTO rollup_disk_storage_usages(rollup_id, satellite_id, at_rest_total, timestamp)
```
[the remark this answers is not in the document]
rikysya: No, as it is SQLite only

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,141 @@
...
+	defer mon.Task()(&ctx)(&err)
+
+	query := `SELECT *
+				FROM rollup_disk_storage_usages
+				WHERE timestamp IN (
+					SELECT MAX(timestamp) 
+					FROM rollup_disk_storage_usages
+					WHERE satellite_id = ?
+					AND ? <= timestamp AND timestamp <= ?
+					GROUP BY DATE(timestamp)
```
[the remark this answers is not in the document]
rikysya: `WHERE satellite_id = ?` is sufficient to receive entries only for one satellite, no?

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,141 @@
...
+func (db *diskSpaceUsage) GetDailyTotal(ctx context.Context, from, to time.Time) (_ []console.DiskSpaceUsage, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	query := `SELECT SUM(at_rest_total), timestamp 
+				FROM rollup_disk_storage_usages
+				WHERE timestamp IN (
+					SELECT MAX(timestamp)
+					FROM rollup_disk_storage_usages
+					WHERE ? <= timestamp AND timestamp <= ?
+					GROUP BY DATE(timestamp), satellite_id
```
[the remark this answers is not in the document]
rikysya: Why? we want max per day for `satelliteID`

storagenode/nodestats/service.go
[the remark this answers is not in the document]
ifraixedes: I asked if we should use structured logger in storagenode and the answer was yes, for that reason I reported.

storagenode/nodestats/service.go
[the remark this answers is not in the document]
ifraixedes: :thinking: I don't understand your question.

I'm curious to know why the third parameter (which is hours) is 1 and not 0 for example.

storagenode/storagenodedb/diskspaceusages.go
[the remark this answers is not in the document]
ifraixedes: for that reason, with `GROUP BY satellite_id` should be enough as we are only selecting the `MAX`

storagenode/storagenodedb/diskspaceusages.go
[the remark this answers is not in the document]
ifraixedes: Oh, yes, my fault, I missed it.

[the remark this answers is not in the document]
ifraixedes: But then the `GROUP BY` isn't needed.

The PK of the table is `satellite_id, timestamp`, hence if we filter by `satellite_id` cannot be any duplicated `timestamp` to group moreover we are taking the maximum, so it doesn't look that's needed for anything.

storagenode/storagenodedb/diskspaceusages.go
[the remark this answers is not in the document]
rikysya: I also want entries to be grouped by day, so I receive list from 1 to 31 July for example

storagenode/storagenodedb/diskspaceusages.go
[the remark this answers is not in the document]
rikysya: If I have two entries per day, and 30 days. I want result as 30 entries, 1 per day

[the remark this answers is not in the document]
rikysya: Where time stamp is the most recent one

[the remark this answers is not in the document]
rikysya: With `MAX` I will receive only the most recent one, but I want entries per day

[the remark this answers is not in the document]
ifraixedes: I'm not entirely sure that the query is doing so. I'll try to reproduce it to see if the query is doing what you expect, unless that there is a test in this branch or somewhere that is verifying, I need it for the sake of my inquietude.

storagenode/nodestats/service.go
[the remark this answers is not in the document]
rikysya: Can you please send me an example because I don't understand what it is

storagenode/nodestats/service.go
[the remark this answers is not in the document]
rikysya: I think it's date, the order should be year month date hours mins sec nanosec
But I will double check

storagenode/nodestats/service.go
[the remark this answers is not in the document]
ifraixedes: Mostly the change would be

`s.log.Erorr("Get disk space usage query failed", zap.Error(err))`

However I'm not sure if that then produce a good error message in the storage node logs despite that it should be better for being ingested by some centralized logging platform.

@kaloyan-raev may have a better assessment than me

storagenode/nodestats/service.go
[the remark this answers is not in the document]
ifraixedes: I'll also do because I've never used before.
Thanks for reply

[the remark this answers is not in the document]
ifraixedes: never mind, that's now when I've realized that the 1 is day not an hour, I misread the position of the parameters when I read the function and I set in my mind that 1 was located in the hour position (the 4th) not the day (the 3rd).

storagenode/nodestats/service.go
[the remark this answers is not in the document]
kaloyan-raev: @ifraixedes I think your assessment is good.

storagenode/storagenodedb/testdata/sqlite.v15.sql
```diff
@@ -0,0 +1,167 @@
...
+INSERT INTO bandwidth_usage_rollups VALUES('2019-07-12 20:00:00+00:00',X'2b3a5863a41f25408a8f5348839d7a1361dbd886d75786bb139a8ca0bdf41000',4,4);
+INSERT INTO bandwidth_usage_rollups VALUES('2019-07-12 18:00:00+00:00',X'0ed28abb2813e184a1e98b0f6605c4911ea468c7e8433eb583e0fca7ceac3000',5,5);
+INSERT INTO bandwidth_usage_rollups VALUES('2019-07-12 20:00:00+00:00',X'2b3a5863a41f25408a8f5348839d7a1361dbd886d75786bb139a8ca0bdf41000',5,5);
+INSERT INTO bandwidth_usage_rollups VALUES('2019-07-12 18:00:00+00:00',X'0ed28abb2813e184a1e98b0f6605c4911ea468c7e8433eb583e0fca7ceac3000',6,6);
+INSERT INTO bandwidth_usage_rollups VALUES('2019-07-12 20:00:00+00:00',X'2b3a5863a41f25408a8f5348839d7a1361dbd886d75786bb139a8ca0bdf41000',6,6);
+
+-- NEW DATA --
+insert INTO node_stats VALUES(X'0ed28abb2813e184a1e98b0f6605c4911ea468c7e8433eb583e0fca7ceac3000',1,1,1.0,1.0,1.0,1,1,1.0,1.0,1.0,'2019-07-19 20:00:00+00:00');
+
+insert INTO disk_storage_usages VALUES(X'0ed28abb2813e184a1e98b0f6605c4911ea468c7e8433eb583e0fca7ceac3000',5.0,'2019-07-19 20:00:00+00:00');
```
ifraixedes: This table should be `rollup_disk_storage_usages`

ifraixedes: It's also missing one column value

pkg/accounting/db.go
```diff
@@ -69,8 +69,8 @@ type StoragenodeAccounting interface {
 	LastTimestamp(ctx context.Context, timestampType string) (time.Time, error)
 	// QueryPaymentInfo queries Nodes and Accounting_Rollup on nodeID
 	QueryPaymentInfo(ctx context.Context, start time.Time, end time.Time) ([]*CSVRow, error)
-	// QueryNodeDailySpaceUsage returns slice of NodeSpaceUsage for given period
-	QueryNodeDailySpaceUsage(ctx context.Context, nodeID storj.NodeID, start time.Time, end time.Time) ([]NodeSpaceUsage, error)
+	// QueryNodeStorageUsage returns slice of NodeStorageUsage for given period
+	QueryNodeStorageUsage(ctx context.Context, nodeID storj.NodeID, start time.Time, end time.Time) ([]NodeStorageUsage, error)
```
egonelbre: `QueryStorageNodeUsage` and `StorageNodeUsage`

pkg/pb/nodestats.proto
```diff
@@ -36,8 +36,8 @@ message DailyStorageUsageRequest {
 
 message DailyStorageUsageResponse {
     message StorageUsage {
-        double at_rest_total = 2;
-        google.protobuf.Timestamp time_stamp = 3 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+        double at_rest_total = 1;
+        google.protobuf.Timestamp timestamp = 2 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
egonelbre: Was the previous version pb deployed to production or is this safe to do?

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -217,24 +223,39 @@ func (db *StoragenodeAccounting) QueryNodeDailySpaceUsage(ctx context.Context, n
...
 
-		nodeSpaceUsages = append(nodeSpaceUsages, accounting.NodeSpaceUsage{
+		// skip first entry as we can not extract hours
+		// properly without storagenode storage tallies
+		// which formed this value
+		if !prevStartTime.Valid {
+			continue
+		}
+
+		atRest := atRestTotal - prevAtRestTotal.Float64
```
egonelbre: Why not calculate this directly in SQL query?

storagenode/console/consoleserver/server.go
```diff
@@ -49,18 +48,18 @@ type DashboardResponse struct {
...
+	Bandwidth          console.BandwidthInfo    `json:"bandwidth"`
+	DiskSpace          console.DiskSpaceInfo    `json:"diskSpace"`
+	WalletAddress      string                   `json:"walletAddress"`
+	VersionInfo        version.Info             `json:"versionInfo"`
+	IsLastVersion      bool                     `json:"isLastVersion"`
+	Uptime             time.Duration            `json:"uptime"`
+	NodeID             string                   `json:"nodeId"`
+	Satellites         storj.NodeIDList         `json:"satellites"`
+	UptimeCheck        console.ReputationStats  `json:"uptimeCheck"`
+	AuditCheck         console.ReputationStats  `json:"auditCheck"`
```
egonelbre: `console` package seems to be accumulating concerns from all the things.

storagenode/console/stats.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"time"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Stats is interface for working with node stats db
+type Stats interface {
```
egonelbre: The fully qualified name is `console.Stats` so I would expect it to contain statistics about the console.

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,131 @@ func (c *Client) Close() error {
...
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get stats query failed: %v", err))
+	}
+
+	err = s.CacheSpaceUsageFromSatellites(ctx)
+	if err != nil {
+		s.log.Error(fmt.Sprintf("Get disk space usage query failed: %v", err))
+	}
+
+	for {
```
egonelbre: Use sync2.Cycle as everywhere else.

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,131 @@ func (c *Client) Close() error {
...
+func (s *Service) CacheSpaceUsageFromSatellites(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	satellites, err := s.consoleDB.Satellites().GetIDs(ctx)
+	if err != nil {
+		return NodeStatsServiceErr.Wrap(err)
+	}
+
+	// get current month edges
+	startDate, endDate := getMonthEdges(time.Now().UTC())
```
egonelbre: We have a package for such utilities.

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,131 @@ func (c *Client) Close() error {
...
+			cacheStatsErr.Add(err)
+			continue
+		}
+
+		stats.UpdatedAt = time.Now()
+
+		// try to update stats from satellite
+		if err = s.consoleDB.Stats().Update(ctx, *stats); err != nil {
+			// if stats doesn't exists - create new one
+			if err == sql.ErrNoRows {
```
egonelbre: This part of the code shouldn't know anything about sql. It might not be stored in sql, but in plain files or somewhere else.

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,131 @@ func (c *Client) Close() error {
...
+				_, err = s.consoleDB.Stats().Create(ctx, *stats)
```
egonelbre: This should happen as a single update not separately. This kind of code is inherently racy and will cause problems.

storagenode/nodestats/service.go
```diff
@@ -162,17 +251,32 @@ func (s *Service) DialNodeStats(ctx context.Context, satelliteID storj.NodeID) (
...
 	}
 
 	return stamps
 }
+
+// Close clear time.Tickers
+func (s *Service) Close() error {
+	defer mon.Task()(nil)(nil)
+	s.statsTicker.Stop()
+	s.spaceTicker.Stop()
```
egonelbre: This must wait for completion `Run` completion.

storagenode/nodestats/service.go
```diff
@@ -66,21 +43,131 @@ func (c *Client) Close() error {
...
+func (s *Service) CacheStatsFromSatellites(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	satellites, err := s.consoleDB.Satellites().GetIDs(ctx)
+	if err != nil {
+		return NodeStatsServiceErr.Wrap(err)
+	}
+
+	var cacheStatsErr errs.Group
+	for _, satellite := range satellites {
```
egonelbre: This loop is not checking for ctx cancellation.

storagenode/nodestats/service.go
```diff
@@ -145,7 +232,9 @@ func (s *Service) GetDailyStorageUsedForSatellite(ctx context.Context, satellite
 }
 
 // DialNodeStats dials GRPC NodeStats client for the satellite by id
-func (s *Service) DialNodeStats(ctx context.Context, satelliteID storj.NodeID) (*Client, error) {
+func (s *Service) DialNodeStats(ctx context.Context, satelliteID storj.NodeID) (_ *Client, err error) {
+	defer mon.Task()(&ctx)(&err)
+
 	satellite, err := s.kademlia.FindNode(ctx, satelliteID)
```
egonelbre: We should have the necessary information in trust pool, we shouldn't search for satellites on the kademlia.

storagenode/storagenodedb/diskspaceusages.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if len(stamps) == 0 {
+		return nil
+	}
+
+	stmt := `INSERT OR REPLACE INTO disk_storage_usages(satellite_id, at_rest_total, timestamp) 
+			VALUES(?,?,?)`
+
+	cb := func(tx *sql.Tx) error {
+		txStmt, err := tx.PrepareContext(ctx, stmt)
```
egonelbre: Why bother with the statements? AFAIK, it doesn't have any significant performance benefits and usually ends up making the code more complicated.

storagenode/storagenodedb/infodb.go
```diff
@@ -405,6 +405,52 @@ func (db *InfoDB) Migration() *migrate.Migration {
 					return nil
 				}),
 			},
+			{
+				Description: "Add stats and space usage cache tables",
+				Version:     15,
+				Action: migrate.SQL{
+					`CREATE TABLE node_stats (
```
egonelbre: any reason not to call this `reputation`?

storagenode/storagenodedb/infodb.go
```diff
@@ -405,6 +405,52 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+						uptime_reputation_score REAL NOT NULL,
+						audit_success_count INTEGER NOT NULL,
+						audit_total_count INTEGER NOT NULL,
+						audit_reputation_alpha REAL NOT NULL,
+						audit_reputation_beta REAL NOT NULL,
+						audit_reputation_score REAL NOT NULL,
+						updated_at TIMESTAMP NOT NULL,
+						PRIMARY KEY (satellite_id)
+					)`,
+					`CREATE TABLE disk_storage_usages (
```
egonelbre: Is there a better name for the table?

storagenode/storagenodedb/nodestats.go
```diff
@@ -0,0 +1,135 @@
...
+// nodeStats works with node stats db
+type nodeStats struct {
+	*InfoDB
+}
+
+// Create inserts new stats into the db
+func (db *nodeStats) Create(ctx context.Context, stats console.NodeStats) (_ *console.NodeStats, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	stmt := `INSERT INTO node_stats (
```
egonelbre: this is not a statement, it's a query

storagenode/storagenodedb/nodestats.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return &stats, nil
+}
+
+// Update updates stored stats
+func (db *nodeStats) Update(ctx context.Context, stats console.NodeStats) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	stmt := `UPDATE node_stats
```
egonelbre: not a statement, the variable name is misleading

egonelbre:
Also PR-s should be smaller. This is too large for a proper review.

storagenode/storagenodedb/diskspaceusages.go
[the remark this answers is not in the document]
ifraixedes: The key point here which was key for me to understanding it was the `DATE` function which I'd totally missed it.

My apologies for all this back and forth, I'm blind :sweat_smile:

storagenode/storagenodedb/testdata/sqlite.v15.sql
rikysya: I removed rollup_id, so it's no longer needed

pkg/pb/nodestats.proto
rikysya: It is in master, but that service is not used there. So I am not sure

storagenode/console/consoleserver/server.go
rikysya: I agree, but may be it's better to fix already pushed `console` code in separate PR

satellite/satellitedb/storagenodeaccounting.go
rikysya: Because I am not sure how do a query that will work with both `postgres` and `sqlite`

storagenode/storagenodedb/reputation.go
```diff
@@ -0,0 +1,90 @@
...
+// reputation works with node reputation DB
+type reputationDB struct {
+	*InfoDB
+}
+
+// Store inserts or updates reputation stats into the db
+func (db *reputationDB) Store(ctx context.Context, stats reputation.Stats) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	stmt := `INSERT OR REPLACE INTO reputation (
```
stefanbenten: How can this every be an Update/Replace?

storagenode/storagenodedb/reputation.go
```diff
@@ -0,0 +1,90 @@
...
+		stats.Audit.Beta,
+		stats.Audit.Score,
+		stats.UpdatedAt.UTC(),
+	)
+
+	return nil
+}
+
+// Get retrieves stats for specific satellite
+func (db *reputationDB) Get(ctx context.Context, satelliteID storj.NodeID) (_ *reputation.Stats, err error) {
```
stefanbenten: Wouldnt it make sense to limit this query? instead of sending bulk data forth and back?

storagenode/storagenodedb/storageusage.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Store stores storage usage stamps to db replacing conflicting entries
+func (db *storageusageDB) Store(ctx context.Context, stamps []storageusage.Stamp) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if len(stamps) == 0 {
+		return nil
+	}
+
+	query := `INSERT OR REPLACE INTO storage_usage(satellite_id, at_rest_total, timestamp)
```
stefanbenten: Same here, i dont think there should at any time be an replace/update happening..

storagenode/console/service.go
```diff
@@ -29,27 +31,22 @@ var (
...
-	bandwidthDB bandwidth.DB
-	pieceInfoDB pieces.DB
-	kademlia    *kademlia.Kademlia
-	version     *version.Service
-	nodestats   *nodestats.Service
+	trust          *trust.Pool
+	consoleDB      DB
+	bandwidthDB    bandwidth.DB
+	reputationDB   reputation.DB
+	storageusageDB storageusage.DB
```
stefanbenten: Would it be possible make reputationDB and storageusageDB a combination?
something like
```
type StatsCache struct {
    bandwidth  bandwidth.DB
    reputation reputation.DB
}
```

storagenode/storagenodedb/infodb.go
```diff
@@ -405,6 +405,52 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+						uptime_reputation_alpha REAL NOT NULL,
+						uptime_reputation_beta REAL NOT NULL,
+						uptime_reputation_score REAL NOT NULL,
+						audit_success_count INTEGER NOT NULL,
+						audit_total_count INTEGER NOT NULL,
+						audit_reputation_alpha REAL NOT NULL,
+						audit_reputation_beta REAL NOT NULL,
+						audit_reputation_score REAL NOT NULL,
+						updated_at TIMESTAMP NOT NULL,
+						PRIMARY KEY (satellite_id)
```
stefanbenten: wouldnt it make sense to have a primary key as (satellite_id, updated_at) here

storagenode/storagenodedb/reputation.go
rikysya: Same timestamp/satellite

storagenode/console/service.go
rikysya: We discussed to have separate calls with different intervals for reputation and storage usage

storagenode/storagenodedb/infodb.go
rikysya: We don’t care about past entries, if I understand correctly

storagenode/console/service.go
stefanbenten: They are still seperate calls, but reduce the amount of entries in the peer and personally make the structure more clear to me.
Both of them can be called and accessed independently. Its just grouping them, as they are used "together"

storagenode/storagenodedb/reputation.go
stefanbenten: That should never happen. Else we have a bug in our software!

storagenode/storagenodedb/infodb.go
stefanbenten: okay, so its basically just a single entry per satellite?

rikysya: Yeah

storagenode/console/service.go
rikysya: Yeah, makes sense

storagenode/storagenodedb/reputation.go
rikysya: Oh, sorry, that's reputation, so I am just replacing satellite old value

storagenode/storagenodedb/storageusage.go
rikysya: I am not sure here, I can't not be sure that third party satellite has the same policy and intervals for rollups. We have no protocolo enforcement for that. And that's storagenodes side, where it can interact with basically any satellite on the network

satellite/accounting/db_test.go
```diff
@@ -5,10 +5,12 @@ package accounting_test
 
 import (
 	"fmt"
+	"math/rand"
```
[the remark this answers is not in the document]
egonelbre: We have a `testrand` package.

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,131 @@
...
+
+// NewCache creates new caching service instance
+func NewCache(log *zap.Logger, service *Service, trust *trust.Pool, reputationDB reputation.DB, storageusageDB storageusage.DB) *Cache {
+	return &Cache{
+		log:             log,
+		service:         service,
+		trust:           trust,
+		reputationDB:    reputationDB,
+		storageusageDB:  storageusageDB,
+		reputationCycle: sync2.NewCycle(time.Hour * 4),
```
mobyvb: these times should be configurable. Add a struct like
```
type Config struct {
	ReputationSync time.Duration `help:"how often to sync reputation" releaseDefault:"4h" devDefault:"5m"`
	StorageSync time.Duration `help:"how often to sync storage" releaseDefault:"12h" devDefault:"5m"`
}
```

Then, take `config Config` as an argument to `NewCache`, and add it to the `Config` struct in `peer.go`.

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,131 @@
...
+
+	cache.reputationCycle.Start(ctx, &group, func(ctx context.Context) error {
+		err := cache.CacheReputationStats(ctx)
+		if err != nil {
+			cache.log.Error("Get stats query failed", zap.Error(err))
+		}
+
+		return nil
+	})
+	cache.storageCycle.Start(ctx, &group, func(ctx context.Context) error {
```
mobyvb: why use `cycle.Start()` instead of `cycle.Run()`?

rikysya: Because `cycle.Start` has `errgroup` as input, as I have 2 cycles that should run in parallel under control of single `cache.Run` I thought that it is more appropriate to do it this way

storagenode/nodestats/cache.go
rikysya: Yeah, it feels like it should be configurable, but the main reason this caching service exits is to reduce satellite load, so if intervals are configurable it is more likely that some `storagenodes` will query `satellite` to often

storagenode/console/consoleserver/server.go
```diff
@@ -175,14 +176,11 @@ func (server *Server) dashboardHandler(writer http.ResponseWriter, request *http
 func (server *Server) getDashboardData(ctx context.Context, satelliteID *storj.NodeID) (DashboardData, error) {
 	var response = DashboardData{}
 
-	satellites, err := server.service.GetSatellites(ctx)
-	if err != nil {
-		return response, err
-	}
+	satellites := server.service.GetSatellites(ctx)
```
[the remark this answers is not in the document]
rikysya: In scope of this PR I just changed some logic cause I started to use `trustpool`. But you are right, I will change in next PR that will update `storagenode/console.Server` logic

pkg/pb/nodestats.proto
egonelbre: Then we need to test this to ensure that we can deploy this without crashing things.

internal/date/utils.go
```diff
@@ -6,16 +6,11 @@ package date
...
-	now := time.Now()
-	currentYear, currentMonth, _ := now.Date()
-	currentLocation := now.Location()
-
-	firstDay = time.Date(currentYear, currentMonth, 1, 0, 0, 0, 0, currentLocation)
-	lastDay = firstDay.AddDate(0, 1, -1)
-
-	return
+// MonthBoundary extract month from the provided date and returns its edges
+func MonthBoundary(t time.Time) (time.Time, time.Time) {
```
egonelbre: I cannot find a test being updated. If a test is missing we should add one.

satellite/nodestats/endpoint.go
```diff
@@ -104,14 +104,14 @@ func (e *Endpoint) DailyStorageUsage(ctx context.Context, req *pb.DailyStorageUs
 	}, nil
 }
 
-// toPBDailyStorageUsage converts NodeSpaceUsage to PB DailyStorageUsageResponse_StorageUsage
-func toPBDailyStorageUsage(usages []accounting.NodeSpaceUsage) []*pb.DailyStorageUsageResponse_StorageUsage {
+// toPBDailyStorageUsage converts StorageNodeUsage to PB DailyStorageUsageResponse_StorageUsage
+func toPBDailyStorageUsage(usages []accounting.StorageNodeUsage) []*pb.DailyStorageUsageResponse_StorageUsage {
```
egonelbre: `toProtoDailyStorageUsage`

satellite/satellitedb/storagenodeaccounting.go
egonelbre: Are there separate queries that we can write?

storagenode/console/consoleserver/server.go
```diff
@@ -49,18 +50,18 @@ type DashboardResponse struct {
...
+	Bandwidth          console.BandwidthInfo   `json:"bandwidth"`
+	DiskSpace          console.DiskSpaceInfo   `json:"diskSpace"`
+	WalletAddress      string                  `json:"walletAddress"`
+	VersionInfo        version.Info            `json:"versionInfo"`
+	IsLastVersion      bool                    `json:"isLastVersion"`
+	Uptime             time.Duration           `json:"uptime"`
+	NodeID             string                  `json:"nodeId"`
+	Satellites         storj.NodeIDList        `json:"satellites"`
+	UptimeCheck        reputation.Metric       `json:"uptimeCheck"`
+	AuditCheck         reputation.Metric       `json:"auditCheck"`
```
egonelbre: This reads much nicer.

storagenode/console/service.go
```diff
@@ -29,27 +31,22 @@ var (
...
-	// at least once
-	GetSatelliteIDs(ctx context.Context, from, to time.Time) (storj.NodeIDList, error)
-	// GetDailyBandwidthUsed returns slice of daily bandwidth usage for provided time range,
-	// sorted in ascending order
-	GetDailyTotalBandwidthUsed(ctx context.Context, from, to time.Time) ([]BandwidthUsed, error)
-	// GetDailyBandwidthUsed returns slice of daily bandwidth usage for provided time range,
-	// sorted in ascending order for particular satellite
-	GetDailyBandwidthUsed(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) ([]BandwidthUsed, error)
+	// Bandwidth is a getter for Bandwidth db
+	Bandwidth() Bandwidth
```
egonelbre: Why not supply `Bandwidth` directly it seems the extra indirection isn't doing much here anymore.

storagenode/console/service.go
```diff
@@ -60,7 +57,7 @@ type Service struct {
 
 // NewService returns new instance of Service
 func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceInfo pieces.DB, kademlia *kademlia.Kademlia, version *version.Service,
-	nodestats *nodestats.Service, allocatedBandwidth, allocatedDiskSpace memory.Size, walletAddress string, versionInfo version.Info) (*Service, error) {
+	allocatedBandwidth, allocatedDiskSpace memory.Size, walletAddress string, versionInfo version.Info) (*Service, error) {
```
egonelbre: There are a lot of arguments here.

storagenode/nodestats/cache.go
egonelbre: People can modify the parameters anyways whether we put the flags in place or not. We should instead have rate-limiting on the satellite.

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,131 @@
...
+		reputationCycle: sync2.NewCycle(time.Hour * 4),
+		storageCycle:    sync2.NewCycle(time.Hour * 12),
+	}
+}
+
+// Run runs loop
+func (cache *Cache) Run(ctx context.Context) error {
+	var group errgroup.Group
+
+	cache.reputationCycle.Start(ctx, &group, func(ctx context.Context) error {
```
egonelbre: There's a thundering herd concern here.

storagenode/nodestats/cache.go
egonelbre: There's a thundering herd concern here.

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,131 @@
...
+type Cache struct {
+	log *zap.Logger
+
+	service        *Service
+	trust          *trust.Pool
+	reputationDB   reputation.DB
+	storageusageDB storageusage.DB
+
+	reputationCycle *sync2.Cycle
+	storageCycle    *sync2.Cycle
```
egonelbre: The cycles don't have to be pointers here.

storagenode/peer.go
```diff
@@ -123,7 +127,10 @@ type Peer struct {
 
 	Collector *collector.Service
 
-	NodeStats *nodestats.Service
+	NodeStats struct {
+		Service *nodestats.Service
+		Loop    *nodestats.Cache
```
egonelbre: Why is this called `Loop`?

storagenode/reputation/reputation.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"time"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// DB works with reputation database
+type DB interface {
+	// Store inserts or updates reputation stats into the DB
+	Store(ctx context.Context, stats Stats) error
```
egonelbre: Currently there's a mismatch in naming. Usually there are names paired together `Store`/`Load` and `Set`/`Get` or `Put`/`Get`.

In this case probably `Set` makes more sense.

storagenode/storagenodedb/storageusage.go
```diff
@@ -0,0 +1,139 @@
...
+	defer mon.Task()(&ctx)(&err)
+
+	if len(stamps) == 0 {
+		return nil
+	}
+
+	query := `INSERT OR REPLACE INTO storage_usage(satellite_id, at_rest_total, timestamp) 
+			VALUES(?,?,?)`
+
+	cb := func(tx *sql.Tx) error {
```
egonelbre: Write the callback inline `withTx` it'll endup slightly nicer.

storagenode/storagenodedb/storageusage.go
```diff
@@ -0,0 +1,139 @@
...
+	return db.withTx(ctx, cb)
+}
+
+// GetDaily returns daily storage usage stamps for particular satellite
+// for provided time range
+func (db *storageusageDB) GetDaily(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) (_ []storageusage.Stamp, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	query := `SELECT *
+				FROM storage_usage
```
egonelbre: Indentation looks weird here.

storagenode/storagenodedb/infodb.go
```diff
@@ -405,6 +405,52 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+						PRIMARY KEY (satellite_id, timestamp)
+					)`,
+				},
+			},
 		},
 	}
 }
+
+// withTx is a helper method which executes callback in transaction scope
+func (db *InfoDB) withTx(ctx context.Context, cb func(tx *sql.Tx) error) error {
```
egonelbre: Coordinate this change with @simongui since he has a PR about splitting up the database into multiple.

storagenode/nodestats/cache.go
rikysya: Can you pls tell more about what is causing other `gouroutines` to wait here?

egonelbre: I'm not quite sure I understand the question.

storagenode/nodestats/cache.go
rikysya: In other places like `sender` pointer is used for cycle, also constructor returns pointer

egonelbre: Mostly it should be non-pointers. You can initialize it with:

```
reputationCycle: *sync2.NewCycle ...
```

rikysya: Ok, just thought that returning it as a pointer is a hint of how it should be used

internal/dbutil/nulltime.go
```diff
@@ -0,0 +1,50 @@
...
+
+package dbutil
+
+import (
+	"database/sql/driver"
+	"time"
+)
+
+const (
+	sqliteTimeLayout = "2006-01-02 15:04:05-07:00"
```
egonelbre: sqlite uses multiple layouts - https://www.sqlite.org/lang_datefunc.html

internal/testplanet/storagenode.go
```diff
@@ -87,6 +88,11 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatellites storj.Nod
 			Collector: collector.Config{
 				Interval: time.Minute,
 			},
+			Nodestats: nodestats.Config{
+				MaxSleepDuration: 1,
```
egonelbre: `MaxSleep: time.Second,`

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,161 @@
...
+		storageCycle:     *sync2.NewCycle(config.StorageSync),
+	}
+}
+
+// Run runs loop
+func (cache *Cache) Run(ctx context.Context) error {
+	var group errgroup.Group
+
+	cache.reputationCycle.Start(ctx, &group, func(ctx context.Context) error {
+		sync2.Sleep(ctx, time.Duration(rand.Intn(cache.maxSleepDuration))*time.Second)
```
egonelbre:
```
if !sync2.Sleep(ctx, time.Duration(rand.Intn(int(cache.maxSleepDuration))) {
    return ctx.Error()
}
```

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,161 @@
...
+// Cache runs cache loop and stores reputation stats
+// and storage usage into db
+type Cache struct {
+	log *zap.Logger
+
+	db      CacheStorage
+	service *Service
+	trust   *trust.Pool
+
+	maxSleepDuration int
```
egonelbre: This should be a `time.Duration`

storagenode/nodestats/cache.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+var (
+	// NodeStatsCacheErr defines node stats cache loop error
+	NodeStatsCacheErr = errs.Class("node stats cache error")
+)
+
+// Config defines nodestats cache configuration
+type Config struct {
+	MaxSleepDuration int           `help:"max number of seconds cache waits before performing each sync" releaseDefault:"300" devDefault:"1"`
```
egonelbre: This should be a `time.Duration` then you don't have to mention the units

`"maximum duration to wait before requesting data"`

internal/dbutil/nulltime.go
rikysya: Should we expand functionality when needed or try to cover all layouts at once?

egonelbre: Yeah, not sure at the moment. Is there a reason we are not using a `var someTime *time.Time` and then scan into `&someTime`?

PR #2547, opened 2019-07-12
Jg/add tests

JessicaGreben:
What: Why:
Adds a test to test that pointerDB only stores objects and not buckets.


satellite/metainfo/db_test.go
```diff
@@ -35,7 +35,7 @@ func newTestBucket(name string, projectID uuid.UUID) storj.Bucket {
 		},
 		DefaultEncryptionParameters: storj.EncryptionParameters{
 			CipherSuite: storj.EncAESGCM,
-			BlockSize:   32,
+			BlockSize:   9 * 10,
```
jenlij: what was wrong with 32?

JessicaGreben: i hit this error https://github.com/storj/storj/blob/master/pkg/metainfo/kvmetainfo/buckets.go#L81

satellite/metainfo/service_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+func TestIterate(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 6, UplinkCount: 1,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		saPeer := planet.Satellites[0]
```
jenlij: nice naming conventions

satellite/metainfo/db_test.go
thepaul: Do you think it might be worth using named constants for 9 and 10, or a named `storj.RedundancyScheme` structure above, so it can be more clear that the intent here is `rs.ShareSize * rs.RequiredShares`? Or just a comment maybe.

PR #2551, opened 2019-07-12
Add test to ensure correct order of repair queue selection

mobyvb:
What: 
Add satellitedb test to make sure we remove from the repair queue in the following order:
1. attempted = null
2. attempted in ascending order (older timestamps first)

Why:
We recently merged a change (#2533) to update the order used for returning items in the repair queue, but the test didn't cover some details

Please describe the tests:
 - Test 1:
Add four paths to the repair queue (nullPath, recentRepairPath, oldRepairPath, and olderRepairPath)
Update recentRepairPath to have a last attempted time of now
Update oldRepairPath to have a last attempted time of 2 hours ago
Update olderRepairPath to have a last attempted time of 3 hours ago
Select from the repair queue 4 times. Expect nullPath first, olderRepairPath second, oldRepairPath third, and an empty queue error the fourth time.
 

satellite/satellitedb/repairqueue_test.go
```diff
@@ -48,3 +53,78 @@ func TestRepairQueue(t *testing.T) {
 		}
 	})
 }
+
+func TestRepairQueueOrder(t *testing.T) {
```
ifraixedes: I think that the description of the test that you wrote in the issue could be added here.

satellite/satellitedb/repairqueue_test.go
```diff
@@ -48,3 +53,78 @@ func TestRepairQueue(t *testing.T) {
...
+			injuredSeg := &pb.InjuredSegment{Path: path}
+			err := repairQueue.Insert(ctx, injuredSeg)
+			require.NoError(t, err)
+		}
+
+		dbAccess := db.(interface{ TestDBAccess() *dbx.DB }).TestDBAccess()
+		var timeConvertPrefix string
+		switch d := dbAccess.DB.Driver().(type) {
+		case *sqlite3.SQLiteDriver:
+			timeConvertPrefix = "datetime("
```
ifraixedes: I've seen in some PRs that we're moving away from using `datetime` in sqlite due the precision that it has. Shouldn't this test be updated accordingly?

mobyvb: I was not aware that we were moving away from datetime, but the repair queue select query itself uses datetime, so I wanted to be consistent here

aligeti: Moby the date time is UTC based? make sure the timing is handled correctly. Other than that test does what is explained in the PR.

mobyvb: yeah, it should be utc based after I did some searching. These are also the same strings we use for postgres vs. sqlite elsewhere in the codebase

PR #2554, opened 2019-07-14
storagenode/pieces: remove buffering from reading/writing and fix io.EOF bug

thepaul:
What: 

* Remove `bufio.Reader` and `bufio.Writer` doesn't seem to have significant impact on reading/writing for large chunk sizes.
* Special case `io.EOF` since most Go code assumes them not to be wrapped.
* Fix bug in endpoint, which assumed `Read` always returns sufficient amount of data.

Please describe the tests:
    (No new functionality; existing tests should cover the changed behavior sufficiently.)
 
    Most likely a performance improvement, due to using a read buffer (assuming we do some small or unaligned reads sometimes). Could potentially cause a slight slowdown for read operations in pathological cases, though.


thepaul:
wow ok I guess it does cause problems. should we take out the bufio.Reader entirely? or is there something dumb i'm missing?

egonelbre:
Two things:
1. we probably should test that the bufio is useful
2. we should figure out what's causing the failure, since adding a buffer somewhere shouldn't break things.

egonelbre:
PS: One common thing that breaks with adding buffering is wrong code that looks like:
```
n, err := buf.Read(data)
if err != nil {
	return err
}

process(data[:n])
```

Whereas, it should be:

```
n, err := buf.Read(data)
if n > 0 {
	process(data[:n])
}
if err != nil {
	return err
}
```

thepaul:
Made a ticket to investigate further at V3-2158.

egonelbre:
So I found a few suspicious read handling:

* mobile/bucket.go:333: resets n to -1 when reader returns io.EOF
* pkg/ranger/file.go:70: resetting of the error code looks suspicious
* storagenode/piecestore/endpoint.go:451: piecereader.Read isn't handling `n`

egonelbre:
It seems the buffering doesn't matter when writing/reading 512KB at a time:

```
BEFORE:
$ go test -bench . -run XXXX ./storagenode/pieces                       
pkg: storj.io/storj/storagenode/pieces                                  
BenchmarkReadWrite/Write-32                    9         121431378 ns/op
BenchmarkReadWrite/Read-32                   184           6444910 ns/op
PASS                                                                    
ok      storj.io/storj/storagenode/pieces       3.423s                  

AFTER:
$ go test -bench . -run XXXX ./storagenode/pieces                       
goos: windows                                                           
goarch: amd64                                                           
pkg: storj.io/storj/storagenode/pieces                                  
BenchmarkReadWrite/Write-32                    9         126349222 ns/op
BenchmarkReadWrite/Read-32                   193           6145245 ns/op
```

storagenode/pieces/readwrite_test.go
```diff
@@ -0,0 +1,85 @@
...
+	defer ctx.Check(blobs.Close)
+
+	store := pieces.NewStore(zap.NewNop(), blobs)
+
+	// setup test parameters
+	const blockSize = int(256 * memory.KiB)
+	satelliteID := testrand.NodeID()
+	source := testrand.Bytes(30 * memory.MB)
+
+	testPieceID := storj.PieceID{1}
```
mniewrzal: can you move writing a test piece after `Write` benchmark to be clear that its used only to test `Read`?

egonelbre: ah, just moved it before... but sure.

egonelbre:
The grpc buffer sizes seem appropriate.

mniewrzal:
hmm its 32KB according to godoc and our buffer was 256KiB before removing.

egonelbre:
The grpc buffer size is different from the chunk size we transfer.

PR #2555, opened 2019-07-15
ensure uplink is sending correct size with PieceHash

thepaul:
What: 

ensure uplink is sending correct size with PieceHash

Why:

If we verify that the size matches reality, we can then expect to use the actual piece _file_ size to store the piece size as given in the signed `PieceHash` from the uplink. Otherwise, the uplink might send a garbage size value, leaving the storagenode with no good way to verify the uplink signature on the PieceHash at a later date, unless we explicitly track what the uplink *sent* as the piece size, which seems like just a waste of bytes and a potential source of confusion when we have multiple sources of truth.

This was motivated by ongoing work to move pieceinfo data into piece files and out of the sqlite database.

Also fix the code in `uplink/piecestore/` so that it sends a valid size, because it was being rude and sending `0`.

Please describe the tests:

 Existing tests cover this behavior ok (I know because after I added the extra check in storagenode code, many tests started failing until I also fixed the uplink code)
 
  insignificant

### DIFF
--- a/storagenode/piecestore/endpoint.go
+++ b/storagenode/piecestore/endpoint.go
@@ -282,6 +282,10 @@ func (endpoint *Endpoint) Upload(stream pb.Piecestore_UploadServer) (err error) 
 			if err := endpoint.VerifyPieceHash(ctx, limit, message.Done, expectedHash); err != nil {
 				return err // TODO: report grpc status internal server error
 			}
+			if message.Done.PieceSize != pieceWriter.Size() {
+				return ErrProtocol.New("Size of finished piece does not match size declared by uplink! %d != %d",
+					message.Done.GetPieceSize(), pieceWriter.Size())
+			}
 
 			if err := pieceWriter.Commit(ctx); err != nil {
 				return ErrInternal.Wrap(err) // TODO: report grpc status internal server error
--- a/uplink/piecestore/upload.go
+++ b/uplink/piecestore/upload.go
@@ -181,8 +181,10 @@ func (client *Upload) Commit(ctx context.Context) (_ *pb.PieceHash, err error) {
 
 	// sign the hash for storage node
 	uplinkHash, err := signing.SignUplinkPieceHash(ctx, client.privateKey, &pb.PieceHash{
-		PieceId: client.limit.PieceId,
-		Hash:    client.hash.Sum(nil),
+		PieceId:   client.limit.PieceId,
+		PieceSize: client.offset,
+		Hash:      client.hash.Sum(nil),
+		Timestamp: client.limit.OrderCreation,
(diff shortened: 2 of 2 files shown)


storagenode/piecestore/endpoint.go
```diff
@@ -282,6 +282,10 @@ func (endpoint *Endpoint) Upload(stream pb.Piecestore_UploadServer) (err error)
 			if err := endpoint.VerifyPieceHash(ctx, limit, message.Done, expectedHash); err != nil {
 				return err // TODO: report grpc status internal server error
 			}
+			if message.Done.PieceSize != pieceWriter.Size() {
+				return ErrProtocol.New("Size of finished piece does not match size declared by uplink! %d != %d",
+					message.Done.GetPieceSize(), pieceWriter.Size())
+			}
```
egonelbre: Let's also check that the timestamp is reasonable.

uplink/piecestore/upload.go
```diff
@@ -181,8 +181,10 @@ func (client *Upload) Commit(ctx context.Context) (_ *pb.PieceHash, err error) {
 
 	// sign the hash for storage node
 	uplinkHash, err := signing.SignUplinkPieceHash(ctx, client.privateKey, &pb.PieceHash{
-		PieceId: client.limit.PieceId,
-		Hash:    client.hash.Sum(nil),
+		PieceId:   client.limit.PieceId,
+		PieceSize: client.offset,
+		Hash:      client.hash.Sum(nil),
+		Timestamp: client.limit.OrderCreation,
```
mniewrzal: I think this should be just `time.Now()`. Its just timestamp when PieceHash was created.

thepaul: I really tend to dislike adding `time.Now()` deep in code like this, because it makes things nondeterministic and hard to test. This can change in the future without breaking backwards compat (because the storagenodes have to keep track of the exact timestamp sent whatever it is), so I would like to put off that decision if possible, and get the size change in before the wipe.

storagenode/piecestore/endpoint.go
thepaul: I would agree, but I'm not sure what reasonable necessarily means here. I think the storagenode will already have to keep track of the exact sent timestamp in pieceinfo (the piece file timestamp on the filesystem probably doesn't have enough resolution and may not be entirely reliable under all possible operating conditions), so I don't much care that the uplink sends any timestamp in particular. If it wants to send 0 or something in 1989, the storagenode can be fine with that if we want it to be. The size is more important- we _can_ use the size of the piece file to keep track of what goes in the PieceHash, as long as we make sure it was right to begin with.

PR #2557, opened 2019-07-15
storagenode: remove datetime calls in favor of UTC

zeebo:
What: pass UTC timestamps for everything in the storagenodedb.
 
Why: datetime only has second level granularity whereas string
comparisons don't. Since we're wiping everything anyway, it's
easier to just use UTC everywhere rather than migrate to
datetime calls.

Please describe the tests:
 - Test 1:
 - Test 2:
 

egonelbre:
Can we add wrapper to sqlite calls to ensure that all times are definitely UTC?

zeebo:
I'm not sure how. Do you mean wrap the `*sql.DB` in something that intercepts the methods and checks `[]interface{}` values?

thepaul:
LGTM.

I'm also worried about having missed some timestamp usage somewhere though. Hard to have any definite thing/code/pattern to search for. I like @egonelbre 's wrapper idea, even if we only end up using it during testing.

zeebo:
I have some quick test commits coming.

zeebo:
Please take another look.

storagenode/storagenodedb/infodb.go
```diff
@@ -88,10 +88,10 @@ func (db *InfoDB) CreateTables(log *zap.Logger) error {
 }
 
 // RawDB returns access to the raw database, only for migration tests.
-func (db *InfoDB) RawDB() *sql.DB { return db.db }
+func (db *InfoDB) RawDB() *sql.DB { return db.db.db }
 
 // Begin begins transaction
-func (db *InfoDB) Begin() (*sql.Tx, error) { return db.db.Begin() }
+func (db *InfoDB) Begin() (*sql.Tx, error) { return db.db.db.Begin() }
```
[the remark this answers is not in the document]
zeebo:
```go
type db struct { *db }
```

storagenode/storagenodedb/bandwidthdb_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/pkg/pb"
+)
+
+func TestBandwidthdb_Trivial(t *testing.T) {
```
egonelbre: These tests are near the interface not the implementation.

storagenode/storagenodedb/consoledb_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testrand"
+)
+
+func TestConsoledb_Trivial(t *testing.T) {
+	Run(t, func(t *testing.T, ctx context.Context, db *DB) {
```
egonelbre: These are near the interface not the implementation.

storagenode/storagenodedb/bandwidthdb_test.go
zeebo: I don't understand what this means. What's the desired change?

storagenode/storagenodedb/infodb.go
```diff
@@ -263,14 +263,10 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
-					`DROP INDEX idx_used_serial`,
 					`DROP INDEX idx_pieceinfo_expiration`,
-					`DROP INDEX idx_bandwidth_usage_created`,
 					`DROP INDEX idx_order_archive_satellite`,
 					`DROP INDEX idx_order_archive_status`,
-					`CREATE INDEX idx_used_serial ON used_serial(datetime(expiration))`,
-					`CREATE INDEX idx_pieceinfo_expiration ON pieceinfo(datetime(piece_expiration)) WHERE piece_expiration IS NOT NULL`,
-					`CREATE INDEX idx_bandwidth_usage_created ON bandwidth_usage(datetime(created_at))`,
+					`CREATE INDEX idx_pieceinfo_expiration ON pieceinfo(piece_expiration) WHERE piece_expiration IS NOT NULL`,
 				},
```
egonelbre: shouldn't we have a migration to convert all timestamps to UTC, in case they already weren't?

storagenode/storagenodedb/pieceinfo_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/storagenode/pieces"
+)
+
+func TestPieceinfo_Trivial(t *testing.T) {
```
egonelbre: The tests are near the interface.

storagenode/storagenodedb/infodb.go
zeebo: We're dropping the databases during the wipe.

PR #2559, opened 2019-07-15
add vouchers to QueryRequest

cam-a:
add vouchers field to QueryRequest to allow sending vouchers during kademlia lookup
Related to https://storjlabs.atlassian.net/browse/V3-1833


pkg/pb/overlay.proto
```diff
@@ -20,6 +21,7 @@ message QueryRequest {
     node.Node target = 2;
     int64 limit = 3;
     bool pingback = 4;
+    repeated vouchers.Voucher vouchers = 5;
```
JessicaGreben: why add this to query versus other overlay RPC like Ping or RequestInfo?

egonelbre: Also since Satellite won't be on kademlia in the future, it wouldn't also have Nodes service

cam-a: The Query endpoint is where `pingback` is requested, which will put us into either the node's routing table or antechamber

PR #2560, opened 2019-07-15
satellite/gc: Fix data race accessing Service.lastSendTime

ifraixedes:
What: Fix the data race with accessing to the `Service.lastSendTime` for the branch `green/gc-queue` (__NOTE__ this branch is not intended to be merged to __MASTER__ it's going to be merged into branch under an open PR)

Why: Because we may avoid to get those errors in production when the service runs concurrently.

Please describe the tests: Add a test case which showed up the data race when calling the `Service.Send` and `Service.NewPieceTracker` concurrently.
 
Please describe the performance impact: Probably slower than without controlling the access, but there is no choice without having a full refactoring and even with that, it isn't clear if it won't need such lock and it will perform better.

### DIFF
--- a/satellite/gc/gc_test.go
+++ b/satellite/gc/gc_test.go
@@ -4,6 +4,8 @@
 package gc_test
 
 import (
+	"context"
+	"sync"
 	"testing"
 	"time"
 
@@ -19,6 +21,7 @@ import (
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/gc"
 )
 
 // TestGarbageCollection does the following:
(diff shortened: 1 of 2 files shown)


mobyvb:
can you update this to be merging into `green/gc-queue` instead of master?

ifraixedes:
Sorry, I didn't remember to change the base branch.

satellite/gc/gc_test.go
```diff
@@ -109,6 +112,32 @@ func TestGarbageCollection(t *testing.T) {
 	})
 }
 
+func TestService_NewPieceTracker_and_Send(t *testing.T) {
+	config := gc.Config{
+		Interval:          time.Minute,
+		Active:            true,
```
navillasa: should update to Enabled

satellite/gc/service.go
```diff
@@ -164,3 +168,9 @@ func (service *Service) lastPieceCountsNumNodes() int {
 
 	return len(m.(map[storj.NodeID]int))
 }
+
+func (service *Service) isActiveFrom(from time.Time) bool {
```
navillasa: should change to isEnabledFrom

ifraixedes: I named this method intentionally like this because I believed that `isEnabled` was going to look like only checking the `Enabled` config field.

However, the method evolved to be `isActiveFrom` hence `isEnabledFrom` may not cause the confusion anymore.

Would you like that we rename it despite of the reason that I've written down (I don't mind to rename it if that's looks clear)?

satellite/gc/gc_test.go
ifraixedes: Done by the merge of the last update of the base branch.

satellite/gc/service.go
navillasa: okay I think the justification makes sense, isActiveFrom is fine!

ifraixedes:
The CI is failing because the base branch is failing; it's basically that the satellite config lock must be updated.

PR #2563, opened 2019-07-15
Add metainfo loop service

mobyvb:
What: 
Add a metainfo loop service on the satellite that can be subscribed to by various services that need to make use of metainfo information

Why:
Checker and garbage collection both need to make use of metainfo data, but might all be configured with different intervals and might have slightly different requirements. It is getting increasingly complex to try to combine everything into the checker loop.

https://storjlabs.atlassian.net/browse/V3-2159

Please describe the tests:
**TestMetainfoLoop** does the following
* upload 5 remote files with 1 segment
* upload 2 inline files
* connect two observers to the metainfo loop
* run the metainfo loop
* expect that each observer has seen
    - 5 remote files
    - 5 remote segments
    - 2 inline files/segments
    - 7 unique path items

**TestMetainfoLoopObserverCancel** does the following:
* upload 3 remote segments
* hook three observers up to metainfo loop
* let observer 1 run normally
* let observer 2 return an error from one of its handlers
* let observer 3's context be canceled
* expect observer 1 to see all segments
* expect observers 2 and 3 to finish with errors

**TestMetainfoLoopCancel** does the following:
* upload 3 remote segments
* hook two observers up to metainfo loop
* cancel loop context partway through
* expect both observers to exit with an error and see fewer than 3 remote segments
* expect that a new observer attempting to join at this point receives a loop closed error
 

pkg/datarepair/checker/checker.go
```diff
@@ -37,41 +34,30 @@ type Config struct {
...
-	remoteSegmentsChecked       int64
-	remoteSegmentsNeedingRepair int64
-	remoteSegmentsLost          int64
-	remoteSegmentInfo           []string
-}
-
 // Checker contains the information needed to do checks for missing pieces
 type Checker struct {
-	metainfo        *metainfo.Service
+	metainfoloop    *metainfoloop.Service
```
egonelbre: Don't change the checker in this PR, only implement the metainfoloop.

satellite/metainfoloop/metainfoloop.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package metainfoloop
```
egonelbre: under `satellite/metainfo/metainfoloop`

satellite/peer.go
```diff
@@ -651,6 +662,9 @@ func (peer *Peer) Run(ctx context.Context) (err error) {
 	group.Go(func() error {
 		return errs2.IgnoreCanceled(peer.Repair.Checker.Run(ctx))
 	})
+	group.Go(func() error {
+		return errs2.IgnoreCanceled(peer.Metainfoloop.Service.Run(ctx))
```
mobyvb: probably should comment this out for this PR

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+// Run starts the looping service.
+func (service *LoopService) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) error {
+		if len(service.waitingObservers) == 0 {
+			return nil
+		}
+		// wait for observers
+		// coalesce incoming observers, within 5s
```
mobyvb: Not sure the best way to implement this

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+						return LoopError.New("error unmarshalling pointer %s", err)
+					}
+
+					path := storj.Path(item.Key.String())
+
+					// send segment info to every observer
+					for _, o := range service.observers {
+						remote := pointer.GetRemote()
+						if remote != nil {
+							_ = o.RemoteSegment(ctx, path, pointer)
```
mobyvb: how should we handle errors? Just log them?

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Close halts the metainfo loop
+func (service *LoopService) Close() error {
+	service.Loop.Close()
+	return nil
+}
+
+// Join will join the looper for one full cycle until completion and then returns.
+// On ctx cancel the observer will return without completely finishing.
```
ifraixedes: I guess that you're working on it, however in the current implementation I'm not clear how the context can be used here to cancel the loop, anyway, I'll see it, mostly I commented to remind you about this.

satellite/peer.go
mobyvb: and disable test, since commenting it out will break the test

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+						if remote != nil {
+							_ = o.RemoteSegment(ctx, path, pointer)
+
+							pathElements := storj.SplitPath(path)
+							if len(pathElements) >= 2 && pathElements[1] == "l" {
+								_ = o.RemoteObject(ctx, path, pointer)
+							}
+
+						} else {
+							_ = o.InlineSegment(ctx, path, pointer)
```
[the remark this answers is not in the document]
mobyvb: I think an inline segment is always an object, so having a separate function would be unnecessary. We could probably call it `InlineObject` if you think that's better

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+type LoopService struct {
+	waitingObservers  []Observer
+	observers         []Observer
+	Loop              *sync2.Cycle
+	metainfo          *Service
+	observersCombined chan bool
+	loopEnded         chan error
+}
+
+// NewLoop creates a new metainfo loop service
```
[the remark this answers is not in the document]
mobyvb: I prefer without `.` so I will remove it everywhere

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,132 @@
...
+// Close halts the metainfo loop
+func (service *LoopService) Close() error {
+	service.Loop.Close()
+	return nil
+}
+
+// Join will join the looper for one full cycle until completion and then returns.
+// On ctx cancel the observer will return without completely finishing.
+// Only on full complete iteration it will return nil.
+func (service *LoopService) Join(ctx context.Context, observer Observer) (err error) {
```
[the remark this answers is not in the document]
mobyvb: I don't know if it needs to, but I will add it just in case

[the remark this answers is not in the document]
ifraixedes: OK to me, but I let you know that's the Go standard library is doing the opposite.

satellite/metainfo/metainfoloop.go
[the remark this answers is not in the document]
ifraixedes: Fine to me, but I let you know that's the Go standard library is doing the opposite.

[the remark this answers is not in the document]
mobyvb: Oh ok I'll change it back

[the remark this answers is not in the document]
ifraixedes: I will add it to our code style guide

satellite/metainfo/metainfoloop.go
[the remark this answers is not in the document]
ifraixedes: Oh, that's what I thought but I thought that in that case we should have the call to `_ = o.RemoteObject(ctx, path, pointer)` but now reading the function name it doesn't make sense because it's called __Remote...__

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+type Observer interface {
+	RemoteSegment(context.Context, storj.Path, *pb.Pointer) error
+	RemoteObject(context.Context, storj.Path, *pb.Pointer) error
+	InlineSegment(context.Context, storj.Path, *pb.Pointer) error
+}
+
+// LoopConfig contains configurable values for the metainfo loop
+type LoopConfig struct {
+	CoalesceDuration time.Duration `help:"how long to wait for new observers before starting iteration" releaseDefault:"5s" devDefault:"5s"`
+	Interval         time.Duration `help:"how long to wait between metainfo iterations" releaseDefault:"30s" devDefault:"30s"`
```
egonelbre: There shouldn't be an interval here.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// LoopService is a metainfo loop service
+type LoopService struct {
+	waitingObservers []Observer
+	observers        []Observer
+	Loop             *sync2.Cycle
```
egonelbre: It shouldn't have a loop, the loop is triggered by the observers not anything else.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package metainfo
```
egonelbre: I initially meant the package should be `metainfo/metainfoloop/loop.go`, however having it be `metainfo.Loop` would be cool as well. Let's rename the file to `loop.go`.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// LoopConfig contains configurable values for the metainfo loop
+type LoopConfig struct {
+	CoalesceDuration time.Duration `help:"how long to wait for new observers before starting iteration" releaseDefault:"5s" devDefault:"5s"`
+	Interval         time.Duration `help:"how long to wait between metainfo iterations" releaseDefault:"30s" devDefault:"30s"`
+}
+
+// LoopService is a metainfo loop service
+type LoopService struct {
```
egonelbre: Just `Loop` is sufficient.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+	waitingObservers []Observer
+	observers        []Observer
```
egonelbre: There shouldn't be a need to keep a list of observers on the struct.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+	Loop             *sync2.Cycle
+	metainfo         *Service
+	loopStartChans   map[Observer]chan struct{}
+	loopEndChans     map[Observer]chan error
```
egonelbre: You shouldn't need this map here.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+	mux              sync.Mutex
```
egonelbre: `mux` usually refers to a server muxer, not a mutex. Mutex is usually referred by `mu` or something more specific. https://github.com/storj/storj/wiki/Style#mutexes

However I don't think you should need a mutex on it in the first place.

egonelbre: Also mutexes should stay above the things it protects.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+	defer mon.Task()(&ctx)(&err)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) error {
+		// wait for observers
+		if len(service.waitingObservers) == 0 {
+			return nil
+		}
+
+		// coalesce incoming observers, within 5s
+		time.Sleep(service.config.CoalesceDuration)
```
egonelbre: This sleep is not cancelable, however it should be.

satellite/metainfo/metainfoloop.go
```diff
@@ -0,0 +1,157 @@
...
+		service.mux.Unlock()
+
+		// notify observers of loop start
+		for _, obs := range service.observers {
+			service.loopStartChans[obs] <- struct{}{}
+		}
+
+		err = service.metainfo.Iterate(ctx, "", "", true, false,
+			func(ctx context.Context, it storage.Iterator) error {
+				var item storage.ListItem
```
egonelbre: this inner loop is not checking for observer cancellation

satellite/peer.go
```diff
@@ -169,6 +170,10 @@ type Peer struct {
 		Endpoint2 *metainfo.Endpoint
 	}
 
+	Metainfoloop struct {
+		Service *metainfo.LoopService
```
egonelbre: Nest this under metainfo as `Loop`.

satellite/peer.go
```diff
@@ -111,8 +111,9 @@ type Config struct {
 	Overlay   overlay.Config
 	Discovery discovery.Config
 
-	Metainfo metainfo.Config
-	Orders   orders.Config
+	Metainfo     metainfo.Config
+	Metainfoloop metainfo.LoopConfig
```
egonelbre: Nest this into metainfo rather than separate.

satellite/metainfo/metainfoloop.go
mobyvb: how can it be implemented without a list of observers? We will have observers in different threads trying to join the same iteration of the metainfo loop. We will also have observers wanting to join while a loop is already in progress, and we need to track these somehow

satellite/metainfo/metainfoloop.go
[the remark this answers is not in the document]
egonelbre: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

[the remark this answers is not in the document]
ifraixedes: Thanks, you anticipated my first task; I wasn't sure if it was there, so I was going to check and if not, then added to our style doc.

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -205,6 +205,12 @@ kademlia.operator.wallet: ""
 # validate redundancy scheme configuration
 # metainfo.rs.validate: true
 
+# how long to wait for new observers before starting iteration
+# metainfoloop.coalesce-duration: 5s
+
+# how long to wait between metainfo iterations
+# metainfoloop.interval: 30s
```
ifraixedes: Is this config still available after the last changes?

satellite/metainfo/loop_test.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+func TestMetainfoLoop(t *testing.T) {
+	// upload 5 remote files with 1 segment
+	// (TODO) upload 3 remote files with 2 segments
+	// upload 2 inline files
+	// connect two observers to the metainfo loop
+	// run the metainfo loop
+	// expect that each observer has seen
+	//     5 remote files
```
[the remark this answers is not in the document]
mobyvb: I think we should actually lower the numbers to be the fewest we need to properly test. Uploading more will just make jenkins slower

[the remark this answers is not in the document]
ifraixedes: It's mostly that I didn't get why is 5 remote files but now I see that it's because `upload 3 remote files with 2 segments` is a `TODO`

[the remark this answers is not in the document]
mobyvb: well I was originally trying to upload 5 remote with 1 segment, 3 remote with 2 segments. But the numbers are arbitrary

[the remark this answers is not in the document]
mobyvb: Oh yeah, there does not currently seem to be a good way to upload a 2 segment file with testplanet. At least not from my brief experimentation. So I left the todo and don't plan to fix that in this PR

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,214 @@
...
+	return &Loop{
+		metainfo: metainfo,
+		config:   config,
+		join:     make(chan *observerContext),
+		done:     make(chan struct{}),
+	}
+}
+
+// Run starts the looping service
+func (service *Loop) Run(ctx context.Context) (err error) {
```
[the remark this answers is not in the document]
mobyvb: I will change it to `loop`

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,214 @@
...
+				if len(observers) == 0 {
+					return nil
+				}
+
+				select {
+				case <-ctx.Done():
+					for _, observer := range observers {
+						observer.HandleError(ctx.Err())
+					}
+					observers = nil
```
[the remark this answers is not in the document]
mobyvb: I am not sure. It could have been added for better cleanup. @egonelbre probably knows better why it's here

satellite/metainfo/loop_test.go
```diff
@@ -0,0 +1,301 @@
...
+		// for i := 0; i < 3; i++ {
+		// 	testData := testrand.Bytes(2 * segmentSize)
+		// 	path := "/some/other/remote/path/" + string(i)
+		// 	err := ul.Upload(ctx, satellite, "bucket", path, testData)
+		// 	require.NoError(t, err)
+		// }
+
+		// upload 2 inline files
+		for i := 0; i < 2; i++ {
+			testData := testrand.Bytes(1 * memory.KiB)
```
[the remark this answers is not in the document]
mobyvb: I think it will be fine to do `segmentSize / 8`, and we can update it in the future if it fails for some reason

scripts/testdata/satellite-config.yaml.lock
mobyvb: will update

satellite/metainfo/loop.go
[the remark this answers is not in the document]
ifraixedes: I thought that was for helping the GC to release the slice but rather than that I didn't find any other reason.

satellite/metainfo/loop_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/metainfo"
+)
+
+func TestMetainfoLoop(t *testing.T) {
+	// upload 5 remote files with 1 segment
```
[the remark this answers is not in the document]
mobyvb: I will update them before the header. I think you are right

satellite/peer.go
mobyvb: no need to uncomment now that loop only runs when observers join

satellite/metainfo/loop.go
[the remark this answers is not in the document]
mobyvb: yes that's what I thought too. There is no functional necessity it seems

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,214 @@
...
+
+	var observers []*observerContext
+
+	defer func() {
+		for _, observer := range observers {
+			observer.Finish()
+		}
+	}()
+
+	select {
```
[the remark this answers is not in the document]
mobyvb: It will make more sense once we add comments, but this `select` is needed to accept the first observer. The `for` loop that follows is on a time limit (`CoalesceDuration`), so it will only accept new observers within a time window after we get the first observer. If we take this `select` out, we will end up doing work that we don't need to following the `for` loop if no observers join in the time limit

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,214 @@
...
+				default:
+				}
+			}
+			return nil
+		})
+}
+
+// Close halts the metainfo loop
+func (service *Loop) Close() error {
+	return nil
```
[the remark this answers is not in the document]
mobyvb: I tried to implement it but it didn't end up being so simple. I am not sure how important implementing this would be, since I don't think anyone would use it

satellite/metainfo/loop.go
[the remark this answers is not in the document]
egonelbre: This is here to avoid double close of observer channels.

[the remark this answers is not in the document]
egonelbre: This is here to avoid double close of observer channels.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,214 @@
...
+
+	obsContext := &observerContext{
+		Observer: observer,
+		ctx:      ctx,
+		done:     make(chan error),
+	}
+
+	select {
+	case <-loop.done:
+		return LoopError.New("loop closed")
```
ifraixedes: I think that it will be preferable to be able to identify this error over others in the way that you the caller can inspect it using the [`Is` function](https://godoc.org/github.com/zeebo/errs#Is), however it would depend in how `Close` gets implemented.

satellite/metainfo/loop.go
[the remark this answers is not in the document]
egonelbre: There’s actually a worse downside, which is that the coalescing time won’t end up being the time specified.

satellite/metainfo/loop.go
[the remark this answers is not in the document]
ifraixedes: Oh, I see, it's because of the `defer`

satellite/metainfo/loop.go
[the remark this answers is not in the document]
ifraixedes: I think that `Join` method should specify this constraint, so one can call `Join` but it will have to wait to some `Run` call that can take it, it may be the first one but it may not be.

I'm not clear about this behavior reading the description of the `CoalesceDuration`, `new observers`, in my opinion, doesn't mean that they are all except the first one in the queue.

satellite/metainfo/loop.go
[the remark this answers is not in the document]
mobyvb: implemented now

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// NewLoop creates a new metainfo loop service.
+func NewLoop(config LoopConfig, metainfo *Service) *Loop {
+	return &Loop{
+		metainfo: metainfo,
+		config:   config,
+		join:     make(chan *observerContext),
+		done:     make(chan struct{}),
+		close:    make(chan struct{}),
```
egonelbre: this extra close channel shouldn't be necessary

mobyvb: sorry didn't have my latest changes pushed

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,247 @@
...
+				}
+
+				observers = nextObservers
+				if len(observers) == 0 {
+					return nil
+				}
+
+				// if context has been canceled, send the error to observers and exit. Otherwise, continue
+				select {
+				case <-loop.someCtx.Done():
```
egonelbre: This shouldn't be here, the cancellation is propagated.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	return obsContext.Wait()
+}
+
+// Run starts the looping service.
+func (loop *Loop) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	loop.someCtx = ctx
```
egonelbre: you shouldn't need to hang on to this context

satellite/metainfo/loop.go
mobyvb: I added it because I couldn't figure out why `<-ctx.Done()` was not happening. It was to experiment and see if the parent context was canceled. Which it was. But the check below never happens, even when the parent context is canceled. This is why I am confused.

satellite/metainfo/loop.go
mobyvb: Purely for experimentation purposes. Not intended to be a permanent change. I added it when the child context was not behaving as expected

satellite/metainfo/loop.go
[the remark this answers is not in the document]
egonelbre: I suspect that it would make the Join/Run more complicated to follow because now you have to coordinate multiple goroutines to get the work, cancellation and errors.

e.g. let's say the first observer starts the Run, then 10 others join it, now the first observer errors and it should exit... this means it would now need to select another observer to take over the work.

satellite/metainfo/loop.go
egonelbre: Things in satellite are implemented such that they don't call close before run finishes. I suspect that's what is causing you problems.

mobyvb: well in the test, we are creating a new loop, separate from the satellite, that way we can pass in a cancelable context we create within the test. The test in question isn't even calling `Close()`. It is just canceling the context passed into `Run()`

mobyvb: When we run the test with the code as is right now, we do see "the original context is done" printed out. When we get rid of `loop.someCtx` and its associated `case` here, and re run the test, we just see "continuing iteration" and the loop never exits due to a context cancelation

mobyvb: If we remove the two lines in `Run` that create a new `ctx` with cancel based on the old context, the test passes as expected. It is only when we create a new context with cancel that it fails

satellite/metainfo/loop.go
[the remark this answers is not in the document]
ifraixedes: I think that you misunderstood my comment.
I was only saying that the `Join` should comment the behavior in its doc comment and `CoalesceDuration` help message doesn't suggest to me saying `new observers` that all the observers are new except the first one in the queue.

I didn't suggest any change in the current implementation, I only asked for clarification in the docs.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,235 @@
...
+
+	return obsContext.Wait()
+}
+
+// Run starts the looping service.
+func (loop *Loop) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	ctx, cancel := context.WithCancel(ctx)
+	loop.cancel = cancel
```
ifraixedes: This cancel leaks.
Do we need it? it looks that it isn't used for anything.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,235 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-loop.done:
+		return LoopClosedError
+	}
+
+	return obsContext.Wait()
+}
+
+// Run starts the looping service.
```
ifraixedes: I would add a comment that mentions that `Run` can only be called once otherwise it panics.
Good to mention in the type or that `Wait` and `Join` are safe to be called concurrently.

satellite/metainfo/loop_test.go
```diff
@@ -0,0 +1,334 @@
...
+		loopCtx, cancel := context.WithCancel(ctx)
+
+		// create 1 normal observer
+		obs1 := newTestObserver(nil)
+
+		var once int64
+		// create another normal observer that will wait before returning during RemoteSegment so we can sync with context cancelation
+		obs2 := newTestObserver(func(ctx context.Context) error {
+			// cancel context during call to obs2.RemoteSegment inside loop
+			fmt.Println("WE ARE CANCELING THE Context")
```
ifraixedes: remove this `Println`

satellite/metainfo/loop_test.go
```diff
@@ -0,0 +1,334 @@
...
+func (obs *testObserver) RemoteObject(ctx context.Context, path storj.Path, pointer *pb.Pointer) error {
+	obs.remoteFileCount++
+	return nil
+}
+
+func (obs *testObserver) InlineSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) error {
+	obs.inlineSegCount++
+	if _, ok := obs.uniquePaths[path]; ok {
+		// TODO: collect the errors and check in test
+		panic("Expected unique path in observer.InlineSegment")
```
ifraixedes: Out of curiosity, the usage of `testing.T` caused any problem? (before it was using `T` now it calls `panic`)

satellite/metainfo/loop.go
egonelbre: Yeah, this `context.WithCancel` shouldn't be necessary anymore.

satellite/metainfo/loop_test.go
egonelbre: Calling `t` concurrently is not thread-safe, it all should happen in the test goroutine.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,235 @@
...
+				}
+
+				path := item.Key.String()
+				pathElements := storj.SplitPath(path)
+				isLastSeg := len(pathElements) >= 2 && pathElements[1] == "l"
+
+				nextObservers := observers[:0]
+
+				for _, observer := range observers {
+					// TODO: move single observer handling into a separate func
```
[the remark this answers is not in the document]
mobyvb: I am removing because since handling a single observer involves loop logic and it would probably make things more confusing if we had another function for this

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,235 @@
...
+					// clear observers slice so they aren't double closed
+					observers = nil
+					return ctx.Err()
+				default:
+				}
+			}
+			return nil
+		})
+
+	if err != nil {
```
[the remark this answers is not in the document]
mobyvb: yeah, looks like you are right. I will keep the logic outside the iterator since it reduces duplicate code

satellite/metainfo/loop.go
```diff
@@ -174,7 +172,6 @@ waitformore:
 				nextObservers := observers[:0]
 
 				for _, observer := range observers {
-					// TODO: move single observer handling into a separate func
```
egonelbre: this todo was not copleted

egonelbre:
This is also still missing per-observer metrics.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,231 @@
...
+				case <-ctx.Done():
+					return ctx.Err()
+				default:
+				}
+			}
+			return nil
+		})
+
+	// if there is an error, send it to all observers before returning
+	if err != nil {
```
egonelbre: You can make this part of the defer above.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,231 @@
...
+	timer := time.NewTimer(loop.config.CoalesceDuration)
+waitformore:
+	for {
+		select {
+		case observer := <-loop.join:
+			observers = append(observers, observer)
+		case <-timer.C:
+			break waitformore
+		case <-ctx.Done():
+			for _, observer := range observers {
```
egonelbre: When the final err handling is part of the defer above, then you can return the error directly here.

satellite/metainfo/loop.go
```diff
@@ -0,0 +1,231 @@
...
+					return LoopError.New("unexpected error unmarshalling pointer %s", err)
+				}
+
+				path := item.Key.String()
+				pathElements := storj.SplitPath(path)
+				isLastSeg := len(pathElements) >= 2 && pathElements[1] == "l"
+
+				nextObservers := observers[:0]
+
+				for _, observer := range observers {
```
egonelbre: Put single observer handling in a separate func, then this part should become clearer.

satellite/metainfo/loop.go
mobyvb: I addressed it in another comment, but since a lot of the observer logic is directly tied to this for loop, I thought it would have been clearer to just leave that logic here rather than moving it out to another function. If you still think I should move it out to another function, I will

PR #2566, opened 2019-07-15
Network Wipe (Pre Beta)

stefanbenten:
What: Deleting all data of the Storagenodes that are tied to our production satellites.

Deletes all blob folders of our satellites and renames the storage folders to prevent further problems

Why: Incompatible Changes.


storagenode/storagenodedb/infodb.go
```diff
@@ -269,6 +272,31 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+				Action: migrate.SQL{
+					`DELETE FROM pieceinfo`,
+					`DELETE FROM used_serial`,
+					`DELETE FROM order_archive`,
+				},
+			},
+			{
+				Description: "Free Storagenodes from trash data",
+				Version:     12,
+				Action: migrate.Func(func(log *zap.Logger, mgdb migrate.DB, tx *sql.Tx) error {
```
zeebo: Uh oh, the memory db has no location but still runs migrations. Doesn’t this remove stuff under `/blob`, etc.?

storagenode/storagenodedb/infodb.go
```diff
@@ -269,6 +272,31 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+					deletecmd := exec.Command("rm", "-r",
+						fmt.Sprintf("%s/blob/%s", db.location, "ukfu6bhbboxilvt7jrwlqk7y2tapb5d2r2tsmj2sjxvw5qaaaaaa"), // us-central1
```
egonelbre: Don't use `fmt.Sprintf`, only `filepath.Join` for path handling.

storagenode/storagenodedb/infodb.go
```diff
@@ -269,6 +272,31 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+					`DELETE FROM pieceinfo`,
+					`DELETE FROM used_serial`,
+					`DELETE FROM order_archive`,
+				},
+			},
+			{
+				Description: "Free Storagenodes from trash data",
+				Version:     12,
+				Action: migrate.Func(func(log *zap.Logger, mgdb migrate.DB, tx *sql.Tx) error {
+					deletecmd := exec.Command("rm", "-r",
```
egonelbre: I think Windows doesn't have `rm` command by default.

storage/filestore/dir.go
```diff
@@ -40,18 +40,18 @@ func NewDir(path string) (*Dir, error) {
...
 
 // Path returns the directory path
 func (dir *Dir) Path() string { return dir.path }
 
-func (dir *Dir) blobdir() string  { return filepath.Join(dir.path, "blob") }
-func (dir *Dir) tempdir() string  { return filepath.Join(dir.path, "tmp") }
-func (dir *Dir) trashdir() string { return filepath.Join(dir.path, "trash") }
+func (dir *Dir) blobsdir() string   { return filepath.Join(dir.path, "blobs") }
+func (dir *Dir) tempdir() string    { return filepath.Join(dir.path, "temp") }
+func (dir *Dir) garbagedir() string { return filepath.Join(dir.path, "garbage") }
```
egonelbre: Why the rename?

stefanbenten: Jeff and JT said, that we should try to take other paths to ensure we are not writing in the "old" directories.
More of a overeager sanity check.

storagenode/storagenodedb/infodb.go
stefanbenten: I add a check for empty location path, that should fix that.

storagenode/storagenodedb/infodb.go
stefanbenten: The Docker container all have that. Thats why we can use it.

storagenode/storagenodedb/infodb.go
stefanbenten: good idea, let me do that.

storagenode/storagenodedb/infodb.go
zeebo: I think we can also just call `os.RemoveAll`

storage/filestore/dir.go
zeebo: In this way we don't have to worry about losing track of migrations, running them again, and deleting all of the live data.

storagenode/storagenodedb/infodb.go
stefanbenten: RemoveAll on the specific folder?

zeebo: Yeah, instead of `rm -r <dir>`, you can do `os.RemoveAll(<dir>)`.

storagenode/storagenodedb/infodb.go
```diff
@@ -269,6 +271,45 @@ func (db *InfoDB) Migration() *migrate.Migration {
...
+					}
+					err = os.RemoveAll(filepath.Join(filepath.Dir(db.location), "blob/qstuylguhrn2ozjv4h2c6xpxykd622gtgurhql2k7k75wqaaaaaa")) // asia-east1
+					if err != nil {
+						log.Sugar().Debug(err)
+					}
+					err = os.RemoveAll(filepath.Join(filepath.Dir(db.location), "blob/abforhuxbzyd35blusvrifvdwmfx4hmocsva4vmpp3rgqaaaaaaa")) // "tothemoon (stefan)"
+					if err != nil {
+						log.Sugar().Debug(err)
+					}
+					// If a folder is not found, the above command is returning an err.
```
ifraixedes: According the GoDoc https://golang.org/pkg/os/#RemoveAll the function will return a `nil` error if the folder path does not exists, otherwise it will return an error.

Not sure if we want to skip all the errors.

stefanbenten: Yes, from what i heard, we dont want to block the SN from starting if it fails to delete data.
As we moved into a new blobs directory, we can just delete it later on manually if users have issues with that.

ifraixedes: OK, then remove that comment because it's incorrect and cause confusion

storagenode/storagenodedb/infodb.go
```diff
@@ -289,9 +289,47 @@ func (db *InfoDB) Migration() *migrate.Migration {
 				Description: "Clear Tables from Alpha data",
 				Version:     12,
 				Action: migrate.SQL{
-					`DELETE FROM pieceinfo`,
-					`DELETE FROM used_serial`,
-					`DELETE FROM order_archive`,
+					`DROP TABLE pieceinfo`,
+					`DROP TABLE used_serial`,
+					`DROP TABLE order_archive`,
+					`CREATE TABLE pieceinfo_ (
```
egonelbre: I am not sure how the rename reduces the risk here since we are dropping the table?

PR #2567, opened 2019-07-16
Metainfo RPC segment methods (part 1)

mniewrzal:
What: Part of Metainfo RPC refactoring related to segments. First part contains:
* custom object for protobuf
* metainfo protobuf changes
* metainfo endpoint method stubs

Full logic for methods will be added in next PR.

Why: 

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/auth/signing/encode.go
```diff
@@ -99,3 +99,14 @@ func EncodeStreamID(ctx context.Context, streamID *pb.SatStreamID) (_ []byte, er
 	streamID.SatelliteSignature = signature
 	return out, err
 }
+
+// EncodeSegmentID encodes segment ID into bytes for signing.
+func EncodeSegmentID(ctx context.Context, segmentID *pb.SatSegmentID) (_ []byte, err error) {
+	defer mon.Task()(&ctx)(&err)
+	signature := segmentID.SatelliteSignature
+	// TODO verify if that can cause race
+	segmentID.SatelliteSignature = nil
```
coyle: could you / would you want to make a copy to avoid a race condition ?

pkg/storj/segmentid.go
```diff
@@ -0,0 +1,84 @@
...
+		return SegmentID{}, ErrSegmentID.Wrap(err)
+	}
+	return SegmentIDFromBytes(idBytes)
+}
+
+// SegmentIDFromBytes converts a byte slice into a segment ID
+func SegmentIDFromBytes(b []byte) (SegmentID, error) {
+	id := make([]byte, len(b))
+	copy(id, b)
+	return id, nil
```
coyle: is there any reason to have the signature return an error ?

pkg/storj/segmentid.go
```diff
@@ -0,0 +1,84 @@
...
+func (id SegmentID) Bytes() []byte { return id[:] }
+
+// Marshal serializes a segment ID
+func (id SegmentID) Marshal() ([]byte, error) {
+	return id.Bytes(), nil
+}
+
+// MarshalTo serializes a segment ID into the passed byte slice
+func (id *SegmentID) MarshalTo(data []byte) (n int, err error) {
+	n = copy(data, id.Bytes())
```
coyle: same comment about error also you can `return copy(data, id.Bytes()), nil`

pkg/storj/segmentid.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// UnmarshalJSON deserializes a json string (as bytes) to a segment ID
+func (id *SegmentID) UnmarshalJSON(data []byte) error {
+	var err error
+	*id, err = SegmentIDFromString(string(data))
+	if err != nil {
+		return err
+	}
+	return nil
```
coyle: nit -- you can just `return err`

coyle: unless it should be wrapped

satellite/metainfo/metainfo.go
```diff
@@ -1100,3 +1100,213 @@ func (endpoint *Endpoint) FinishDeleteObject(ctx context.Context, req *pb.Object
...
+	streamID, err := endpoint.unmarshalSatStreamID(ctx, req.StreamId)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	_, err = endpoint.validateAuth(ctx, macaroon.Action{
+		Op:            macaroon.ActionWrite,
+		Bucket:        streamID.Bucket,
+		EncryptedPath: streamID.EncryptedPath,
+		Time:          time.Now(),
```
coyle: should we convert to UTC ?

pkg/storj/segmentid.go
mniewrzal: this is a temporary version of method to support Metainfo refactoring transition. In future this method will be checking is `b` length is correct. I can remove `error` for now if you think its important.

coyle: maybe just a comment to explain but I'm fine leaving it in if it was intentional

pkg/pb/metainfo.proto
```diff
@@ -328,4 +340,139 @@ message SatStreamID {
...
+    int32               piece_num = 1;
+    bytes               node_id = 2 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
+    orders.PieceHash    hash = 3;
+}
+
+// only for satellite use
+message SatSegmentID {
+    SatStreamID stream_id = 1;
+    int32    part_number = 2;
+    int32    index = 3;
```
aligeti: Does the part_number in message SatSegmentID and the part_number in message SegmentPosition refer to segment's position in the stream?

pkg/pb/metainfo.proto
```diff
@@ -28,6 +28,14 @@ service Metainfo {
     rpc BeginDeleteObject(ObjectBeginDeleteRequest) returns (ObjectBeginDeleteResponse);
     rpc FinishDeleteObject(ObjectFinishDeleteRequest) returns (ObjectFinishDeleteResponse);
 
+    rpc BeginSegment(SegmentBeginRequest) returns (SegmentBeginResponse);
+    rpc CommitSegment(SegmentCommitRequest) returns (SegmentCommitResponse);
+    rpc MakeInlineSegment(SegmentMakeInlineRequest) returns (SegmentMakeInlineResponse);
```
aligeti: **Just a thought.** Why do we need to separately tell satellite that store this segment ( which apparently happened to be a small size) as inline. can we(uplink) just encode like it does with other pieces(bigger than inline size) and let satellite decide based on the size of the piece, to store it as inline or remote. In short eliminate the uplink making the decision and telling satellite to store it has inline or remote.

pkg/pb/metainfo.proto
```diff
@@ -328,4 +340,139 @@ message SatStreamID {
...
+    repeated AddressedOrderLimit addressed_limits = 2;
+}
+
+message SegmentFinishDeleteRequest {
+    bytes segment_id = 1 [(gogoproto.customtype) = "SegmentID", (gogoproto.nullable) = false];
+    repeated SegmentPieceDeleteResult results = 2;
+}
+
+message SegmentPieceDeleteResult {
+    int32               piece_num = 1;
```
aligeti: Is this message sent by satellite to uplink,? what causes this message to be generated? how is this message handled? is this a request or response message?

satellite/metainfo/metainfo.go
aligeti: I always had that question. And still confused with time.Now() and time.Now().UTC().

satellite/metainfo/metainfo.go
```diff
@@ -1100,3 +1100,213 @@ func (endpoint *Endpoint) FinishDeleteObject(ctx context.Context, req *pb.Object
...
+	// TODO implement logic
+
+	return &pb.SegmentDownloadResponse{}, status.Error(codes.Unimplemented, "not implemented")
+}
+
+func (endpoint *Endpoint) unmarshalSatStreamID(ctx context.Context, streamID storj.StreamID) (*pb.SatStreamID, error) {
+	satStreamID := &pb.SatStreamID{}
+	err := proto.Unmarshal(streamID, satStreamID)
+	if err != nil {
+		return nil, err
```
aligeti: can err returns stauts.Error() to be consistent? and other places where 'err' is returned

pkg/storj/segmentid.go
mniewrzal: done

pkg/storj/segmentid.go
mniewrzal: I added `(implements gogo's custom type interface)` to all necessary methods

pkg/auth/signing/encode.go
```diff
@@ -93,9 +93,18 @@ func EncodeVoucher(ctx context.Context, voucher *pb.Voucher) (_ []byte, err erro
...
 	return out, err
 }
+
+// EncodeSegmentID encodes segment ID into bytes for signing.
+func EncodeSegmentID(ctx context.Context, segmentID *pb.SatSegmentID) (_ []byte, err error) {
+	defer mon.Task()(&ctx)(&err)
+	signature := segmentID.SatelliteSignature
+	segmentID.SatelliteSignature = nil
+	out, err := proto.Marshal(segmentID)
+	segmentID.SatelliteSignature = signature
```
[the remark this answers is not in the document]
mniewrzal: value is set back because `segmentID` is a pointer to struct and we are modifying object used also outside this method

pkg/auth/signing/sign.go
```diff
@@ -127,3 +127,21 @@ func SignStreamID(ctx context.Context, signer Signer, unsigned *pb.SatStreamID)
 
 	return &signed, nil
 }
+
+// SignSegmentID signs the segment ID using the specified signer
+// Signer is a satellite
+func SignSegmentID(ctx context.Context, signer Signer, unsigned *pb.SatSegmentID) (_ *pb.SatSegmentID, err error) {
```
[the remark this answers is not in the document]
mniewrzal: what kind of `datatype/keyword` is that? such names are used in all methods in this file. its convinient to have it like that because in case of making another one it will have the same structure and naming. it's consistent and easy to modify copied (new) method.

pkg/auth/signing/verify.go
```diff
@@ -93,3 +93,14 @@ func VerifyStreamID(ctx context.Context, satellite Signee, signed *pb.SatStreamI
 
 	return satellite.HashAndVerifySignature(ctx, bytes, signed.SatelliteSignature)
 }
+
+// VerifySegmentID verifies that the signature inside segment ID belongs to the satellite
+func VerifySegmentID(ctx context.Context, satellite Signee, signed *pb.SatSegmentID) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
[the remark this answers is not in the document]
mniewrzal: ^

pkg/pb/metainfo.proto
mniewrzal: Field `part_number` in general will be used for multipart upload in the future.

pkg/pb/metainfo.proto
mniewrzal: its part of `SegmentFinishDeleteRequest`, it will be sent from uplink to satellite after deleting pieces from storage nodes

satellite/metainfo/metainfo.go
mniewrzal: I'm not sure I understand your comment

pkg/pb/metainfo.proto
mniewrzal: the inline segment size is an uplink decision, can be larger than our current default but price for it will be much higher than remote stuff

pkg/storj/segmentid.go
mniewrzal: comment added

satellite/metainfo/metainfo.go
mniewrzal: I don't think we need to convert it to UTC. At this level `time.Time` will handle timezones.

satellite/metainfo/metainfo.go
aligeti: Instead of returning 'err' in that file, the errors are returning status codes as part of the error returning. The comment is about keeping it consistent.

aligeti: see line 1273 something like "status.Error(.....)"

mniewrzal: its private method, `status.Error(.....)` is returned one level above, here just simple error is returned

PR #2571, opened 2019-07-16
v3-2156: Add partner_id on user creation

ethanadams:
What: 
Add partner ID to the user table on user creation, if the partner ID is provided.

Add partner ID to api key if the user creating the API key has a partner ID associated with it.

Why:
Associate a user to a partner ID for value attribution
Associate a api key to a partner ID

Please describe the tests:
 - Test 1: Updated existing users_test
 - Test 2: Updated existing mutation_test
 

satellite/satellitedb/users.go
```diff
@@ -78,6 +82,7 @@ func (users *users) Delete(ctx context.Context, id uuid.UUID) (err error) {
 // Update is a method for updating user entity
 func (users *users) Update(ctx context.Context, user *console.User) (err error) {
```
coyle: We don't want to allow for the partner_id to be set after the account is created

satellite/console/consoleweb/consoleql/user.go
```diff
@@ -75,6 +77,9 @@ func graphqlUserInput() *graphql.InputObject {
 			FieldPassword: &graphql.InputObjectFieldConfig{
 				Type: graphql.String,
```
VinozzZ: Just curious on why we need to add the `PartnerID` in the `UserInput` type but not in the `User` type?

satellite/satellitedb/users.go
VinozzZ: I think the diff here is misleading. The `Update` method is not changed. It's the `userFromDBX` function that's modified

satellite/console/consoleweb/consoleql/user.go
ethanadams: updated

PR #2572, opened 2019-07-16
Add UpdatePieces method to Metainfo service

kaloyan-raev:
What: Introduces UpdatePieces method to the Metainfo Service.

Why: We need to atomically of add and remove pieces from pointers. See https://storjlabs.atlassian.net/browse/V3-2088 for details.

Please describe the tests:
 - Test 1:
 - Test 2:
 

kaloyan-raev:
@egonelbre Could you have a look at how I've done synchronization?

satellite/metainfo/service.go
```diff
@@ -40,31 +41,121 @@ func (s *Service) Put(ctx context.Context, path string, pointer *pb.Pointer) (er
...
+// It will first check if the pointer has been deleted or replaced. Then it
+// will remove the toRemove pieces and then it will add the toAdd pieces.
+// Replacing the node ID and the hash of a piece can be done by adding the
+// piece to both toAdd and toRemove.
+func (s *Service) UpdatePieces(ctx context.Context, path string, ref *pb.Pointer, toAdd, toRemove []*pb.RemotePiece) (pointer *pb.Pointer, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	// lock the mutex to prevent Put and Delete to mess up with the pointer
+	// while it is updated here
+	s.pointerDBWriteLock.Lock()
```
egonelbre: This won't be sufficient since we will break the satellite into separate servers.

kaloyan-raev: So what is the correct way to do it?

egonelbre: There are a few ways to techincally implement them, but all of them require database support:

1. builtin transactions
2. compare-and-swap
3. compare-and-swap with versioning.

All of them roughly translate to code:

```
if !compare "old-value" and swap "new-value" {
    handle change conflict
}
```

Or in other words, you check whether the old-value has changed only update when it hasn't... otherwise, you need to handle the conflict.

Depending on the situation it might be that:
1. you load the value that is now in the database and redo your operation,
2. cancel the database change and clean-up (e.g. by deleting the newly uploaded piece)
3. reject the request

kaloyan-raev: So, we need to add a new method to the `KeyValueStore` interface:
```go
CompareAndSwap(ctx context.Context, key Key, oldValue, newValue Value) error
```
and implement it correctly in all database implementations.

Is this correct?

egonelbre: Yeah

kaloyan-raev: How would this affect `Put` and `Delete`?

I guess we'd want to `Put` only if there is no such key yet -- to prevent overwriting changes that were just added by another process/goroutune. Perhaps, `Put` should be implemented as `CompareAndSwap(ctx, key, nil, value)`.

How about `Delete`? Should we be worried about anything there?

egonelbre: Yeah, put should fail if something already exists, so yes.

For the delete yeah, it's somewhat equivalent to `CompareAndSwap(ctx, key, oldvalue, nil)`. But, sometimes nil value has meaning. I don't think there's much more to delete.

kaloyan-raev: I created https://github.com/storj/storj/pull/2602 for adding `CompareAndSwap`.

kaloyan-raev: This now uses the new `CompareAndSwap`.

pkg/storage/segments/repairer.go
```diff
@@ -186,57 +186,61 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
...
-				revisedPointerPieces = append(revisedPointerPieces, p)
+	// if partial repair, leave "unhealthy" pieces in the pointer that were not repaired
+	unhealthyPiecesToRemove := unhealthyPieces
+	if healthyAfterRepair < pointer.Remote.Redundancy.SuccessThreshold {
+		for _, p := range unhealthyPiecesToRemove {
+			num := p.GetPieceNum()
+			if !repairedMap[num] {
+				// leave only repaired pieces in the slice, unrepaired
+				// unhealthy pieces are not removed from the pointer
+				unhealthyPiecesToRemove = append(unhealthyPiecesToRemove[:num], unhealthyPiecesToRemove[num+1:]...)
```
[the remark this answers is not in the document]
kaloyan-raev: Oh, you are right! We should use the loop index when removing the item from `unhealthyPiecesToRemove`. Fixed.

pkg/storage/segments/repairer.go
```diff
@@ -186,57 +186,61 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
 		return Error.Wrap(err)
 	}
 
-	// Add the successfully uploaded pieces to the healthyPieces
+	// Add the successfully uploaded pieces to repairedPieces
+	var repairedPieces []*pb.RemotePiece
+	repairedMap := make(map[int32]bool)
```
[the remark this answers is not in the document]
kaloyan-raev: Because `PieceNum` is `int32`. I can switch to `int`, but then we should typecast `p.GetPieceNum()` on line 225. So I don't see a benefit.

pkg/storage/segments/repairer.go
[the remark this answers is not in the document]
mobyvb: Well  I am still not sure if that is sufficient. Suppose you have a `list` of 10 elements, and you are iterating over it with 
`for i, item := range list {`
If you are iterating and you are on index 4, and you say `list = append(list[:i], list[i+1:])`, what do you expect in the next iteration? The next item we *want* is at index 4 now (because we removed the current item). But will the `for` loop start the next iteration with `i=4` again, or will it continue to `i=5` and skip an item?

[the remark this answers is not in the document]
mobyvb: see this: https://play.golang.org/p/jFZJRWoaZp5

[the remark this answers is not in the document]
kaloyan-raev: omg, I completely screwed it. Let me think about a better solution.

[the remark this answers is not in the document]
mobyvb: I think we could just append the pieces to a new list and keep the one we are iterating over static. Then, after the loop, we can set the one we looped over to the new one. This is similar to what was happening before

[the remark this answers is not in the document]
kaloyan-raev: I think I fixed it now, following your suggestion.

pkg/datarepair/checker/checker_test.go
```diff
@@ -150,8 +150,9 @@ func TestIdentifyIrreparableSegments(t *testing.T) {
 				RemotePieces: pieces,
 			},
 		}
-		// put test pointer to db
-		metainfo = planet.Satellites[0].Metainfo.Service
+		// update test pointer in db
+		err = metainfo.Delete(ctx, "fake-piece-id")
```
ifraixedes: Out of curiosity, why wasn't it needed before?

pkg/storage/segments/repairer.go
```diff
@@ -186,57 +186,63 @@ func (repairer *Repairer) Repair(ctx context.Context, path storj.Path) (err erro
...
 	// Update the segment pointer in the metainfo
-	return Error.Wrap(repairer.metainfo.Put(ctx, path, pointer))
+	_, err = repairer.metainfo.UpdatePieces(ctx, path, pointer, repairedPieces, toRemove)
+	return err
 }
 
 // sliceToSet converts the given slice to a set
-func sliceToSet(slice []int32) map[int32]struct{} {
-	set := make(map[int32]struct{}, len(slice))
+func sliceToSet(slice []int32) map[int32]bool {
```
ifraixedes: Which is the benefit of this change?

satellite/metainfo/service.go
```diff
@@ -43,28 +42,110 @@ func (s *Service) Put(ctx context.Context, path string, pointer *pb.Pointer) (er
...
 	}
 
-	// TODO(kaloyan): make sure that we know we are overwriting the pointer!
-	// In such case we should delete the pieces of the old segment if it was
-	// a remote one.
-	if err = s.DB.Put(ctx, []byte(path), pointerBytes); err != nil {
-		return Error.Wrap(err)
-	}
+	// CompareAndSwap is used instead of Put to avoid overwriting existing pointers
+	err = s.DB.CompareAndSwap(ctx, []byte(path), nil, pointerBytes)
```
ifraixedes: With this change this `Put` method behaves like an `Add` isn't it?

pkg/datarepair/checker/checker_test.go
kaloyan-raev: Because the test wants to replace the pointer. Now `metainfo.Put` fails if a pointer for this path already exists. Hence, the test needs to execute Delete and Put.

pkg/storage/segments/repairer.go
kaloyan-raev: The code for testing if a key exists in the test is simpler:
```go
if mySet[key] { ... }
```
vs
```go
if _, ok := mySet[key]; ok { ... }
```

satellite/metainfo/service.go
kaloyan-raev: It acts like the `Put` before, but it will fail with `storage.ErrValueChanged` if such a path already exists. This way we avoid overriding of pointers, which can lead to serious issues.

pkg/datarepair/checker/checker_test.go
ifraixedes: oh, yes silly question, I've read that later in the review and I didn't think to review my previous comments before submitting it.

pkg/storage/segments/repairer.go
ifraixedes: the downside of it is that you have to understand that the `bool` doesn't have any meaning (i.e. if the key exists it always be `true`)

satellite/metainfo/service.go
ifraixedes: Then it's always inserted, so it's like an `Add`.
I don't see `Put` as a good name after this has changed, but obviously renaming the method will break the public API of the package requiring to update many other packages.

PR #2573, opened 2019-07-16
Re-implement libstorj API (V2) using libuplink (V3)

bryanchriswhite:
What (**[Jira](https://storjlabs.atlassian.net/browse/V3-1181?atlOrigin=eyJpIjoiMWM3NGM4ZDcwMmE5NDIyZmFlMzc3YjkxYzU1ZjUwZWYiLCJwIjoiaiJ9))**: An "as backwards-compatible as possible" re-implementation of the libstorj API using libuplink.

This PR includes libstorj as a submodule at `lib/libstorj` and only includes changes required to support that work.

Why: In order to provide V2/libstorj API consumers (e.g. FileZilla) a convenient way to interact with the V3 nework.

Please describe the tests: This adds a test that compiles and executes the libstorj C tests similar to how we run the C test for libuplink. See [libstorj#459](https://github.com/storj/libstorj/pull/459)

Please describe the performance impact: see [libstorj#459](https://github.com/storj/libstorj/pull/459)


lib/libstorj/src/storj.c
```diff
@@ -0,0 +1,782 @@
...
+
+            file->created = strdup(created_str);
+            file->mimetype = strdup(object_item.content_type);
+            file->id = strdup(object_item.path);
+            file->bucket_id = strdup(object_item.bucket.name);
+            file->filename = strdup(object_item.path);
+            file->decrypted = true;
+
+            // TODO: if we want to populate size we need to
+            //  get object meta for each file.
```
kaloyan-raev: Why is this? In the Java bindings, we have a `Size` field in the `ObjectInfo` class. I don't see no reason to be the same for the C bindings. Perhaps, it was just missed to be added.

lib/libstorj/src/storj.h
```diff
@@ -0,0 +1,905 @@
...
+    uv_async_t async;
+    storj_log_levels_t *log;
+
+    /* New in V3 */
+    UplinkRef uplink_ref;
+    ProjectRef project_ref;
+
+    // TODO: delete?
+    /* unused in V3 */
+    storj_http_options_t *http_options;
```
kaloyan-raev: Perhaps, leave it as is here, but ignore it in the implementation.

lib/libstorj/src/uploader.c
```diff
@@ -0,0 +1,250 @@
...
+        size_t remaining_size = state->file_size - state->uploaded_bytes;
+        if (remaining_size >= state->buffer_size) {
+            buf_len = state->buffer_size;
+        } else {
+            buf_len = remaining_size;
+        }
+
+        buf = malloc(buf_len);
+        size_t read_size = fread(buf, sizeof(char), buf_len, state->original_file);
+        // TODO: what if read_size != buf_len?
```
kaloyan-raev: According to the documentation (http://www.cplusplus.com/reference/cstdio/fread/) this can happen either on EOF or error. I think EOF is excluded here because we request reading of no more than the remaining bytes in the file. So, the only case remaining is a read error.

lib/libstorj/src/uploader.c
```diff
@@ -0,0 +1,250 @@
...
+
+        int written_size = upload_write(state->uploader_ref, buf, buf_len, STORJ_LAST_ERROR);
+        STORJ_RETURN_SET_STATE_ERROR_IF_LAST_ERROR;
+        if (written_size != buf_len) {
+            free(buf);
+            return;
+        }
+
+        // TODO: use uv_async_init/uv_async_send instead of calling cb directly?
```
kaloyan-raev: I believe we should be good enough with calling the callback directly.

lib/libstorj/src/uploader.h
```diff
@@ -0,0 +1,23 @@
...
+
+#define STORJ_NULL -1
+#define STORJ_MAX_REPORT_TRIES 2
+#define STORJ_MAX_PUSH_FRAME_COUNT 6
+
+static uv_work_t *uv_work_new();
+
+static int check_in_progress(storj_upload_state_t *state, int status);
+// TODO: is this still used?
+char *create_tmp_name(storj_upload_state_t *state, char *extension);
```
kaloyan-raev: Perhaps not needed anymore, any temp files should be created in the Go space.

lib/uplinkc/encryption_access.go
```diff
@@ -0,0 +1,65 @@
...
+// set_default_key sets the default key for the encryption access context.
+func set_default_key(encAccessRef C.EncryptionAccessRef, key *C.uint8_t, cerr **C.char) {
+	encAccess, ok := universe.Get(encAccessRef._handle).(*uplink.EncryptionAccess)
+	if !ok {
+		*cerr = C.CString("invalid encryption access")
+		return
+	}
+
+	goKey, cKey := storj.Key{}, (*storj.Key)(unsafe.Pointer(key))
+	copy(goKey[:], cKey[:])
```
kaloyan-raev: We now use `SaltedKeyFromPassphrase` instead of plain `copy` to convert the human-readable passphrase to an encryption key.

bryanchriswhite: usage updated :+1:

lib/libstorj/src/storj.h
bryanchriswhite: :+1:

lib/uplinkc/testdata/object_test.c
```diff
@@ -138,6 +153,15 @@ void handle_project(ProjectRef project) {
         free_list_objects(&objects_list);
     }
 
+    { // Delete objects
```
mobyvb: Imo, the delete test should verify that the object is no longer in existence after the delete. Either by doing `get` and expecting an error or by doing a `list` and expecting that the object is not in the list

lib/uplinkc/object.go
```diff
@@ -95,11 +96,11 @@ type Upload struct {
 
 // upload uploads a new object, if authorized.
 //export upload
-func upload(cBucket C.BucketRef, path *C.char, cOpts *C.UploadOptions, cErr **C.char) (downloader C.UploaderRef) {
+func upload(cBucket C.BucketRef, path *C.char, cOpts *C.UploadOptions, cErr **C.char) C.UploaderRef {
```
wthorp: Does this apply here?  "Cgo translates C types into equivalent unexported Go types. Because the translations are unexported, a Go package should not expose C types in its exported API: a C type used in one Go package is different from the same C type used in another. "

bryanchriswhite: This change is just a superficial one. The unused, named return argument was named the wrong thing; hence, just un-naming it. 

That line from the cgo is referring to a situation like this: 
```
// uplink_definitions.h
typedef struct Bucket { long _handle; } BucketRef; 
```
```
package one
// #include "uplink_definitions.h"
import "C"
type BucketRef =  C.BucketRef
func AssertEqualZeroValue() {
    assert(BucketRef{} == C.BucketRef{})
}
```
```
package two
// #include "uplink_definitions.h"
import "C"
type BucketRef =  C.BucketRef
func AssertEqualZeroValue() {
    assert(BucketRef{} == C.BucketRef{})
}
```
```
package main

import "reflect"
import "one"
import "two"

assert(one.BucketRef{} == two.BucketRef{}) // should fail
assert(C.BucketRef{} == one.BucketRef{}) // should fail
assert(C.BucketRef{} == two.BucketRef{}) // should fail
```
**Disclaimer**: I haven't run any of this code :/

Here's a related SO question: [Why are Go C types different when they come from different modules?](https://stackoverflow.com/questions/47117064/why-are-go-c-types-different-when-they-come-from-different-modules) and here's a github issue discussing changing this: [cmd/cgo: make identical C types identical Go types across packages](https://github.com/golang/go/issues/13467).

lib/uplinkc/testdata/object_test.c
bryanchriswhite: Absolutely, there is a [todo in the test](https://github.com/bryanchriswhite/libstorj/blob/V3/test/tests.c#L381) for this. Maybe we should track that in the review process for the libstorj code?

lib/uplinkc/testdata/object_test.c
```diff
@@ -40,7 +48,9 @@ void handle_project(ProjectRef project) {
 
 
     for(int i = 0; i < num_of_objects; i++) {
-        size_t data_len = 1024 * (i + 1) * (i + 1);
+//        size_t data_len = 1024 * (i + 1) * (i + 1);
```
wthorp: comment this

lib/uplinkc/object.go
```diff
@@ -95,11 +96,11 @@ type Upload struct {
 
 // upload uploads a new object, if authorized.
 //export upload
```
ifraixedes: nit: I've mostly seen `//export` being above the function doc comment, let's move this one to above in this same PR

lib/uplinkc/testdata_test.go
```diff
@@ -86,3 +92,65 @@ func TestC(t *testing.T) {
 		}
 	})
 }
+
+func TestLibstorj(t *testing.T) {
+	ctx := testcontext.NewWithTimeout(t, 5*time.Minute)
+	defer ctx.Cleanup()
+
+	libuplink := ctx.CompileShared(t, "uplink", "storj.io/storj/lib/uplinkc")
```
ifraixedes: Lets rename this variable for avoiding the collision with the `libuplink` package, rather than define a specific package identifier.
We recently added such convention https://github.com/storj/docs/blob/master/code/Style.md#variable-naming

bryanchriswhite: Good catch. I think we are trying to use the `libuplink` alias for the package at storj.io/storj/lib/uplink because there is a different `upliink` package at storj.io/storj/uplink. I will rename this variable, it's not a great name in addition to colliding.

lib/uplinkc/object.go
```diff
@@ -135,11 +136,17 @@ func upload_write(uploader C.UploaderRef, bytes *C.uint8_t, length C.size_t, cEr
 		return C.size_t(0)
 	}
 
+	select {
+	case <-upload.ctx.Done():
+		return C.size_t(0)
```
mobyvb: should you set `cErr` here?

lib/uplinkc/object.go
```diff
@@ -252,11 +270,17 @@ func download_read(downloader C.DownloaderRef, bytes *C.uint8_t, length C.size_t
 		return C.size_t(0)
 	}
 
+	select {
+	case <-download.ctx.Done():
+		return C.size_t(0)
```
mobyvb: should you set `cErr`? You could use `download.ctx.Err()` perhaps

lib/uplinkc/testdata/object_test.c
```diff
@@ -19,7 +19,15 @@ void handle_project(ProjectRef project) {
     char **err = &_err;
 
     char *bucket_name = "test-bucket";
-    char *enc_ctx = "12VtN2sbbn9PvaEvNbNUBiSKnRcSUNxBADwDWGsPY7UV85e82tT6u";
+
+    uint8_t *salted_key = project_salted_key_from_passphrase(project,
```
mobyvb: `salted_key` was `malloc`ed, so we should free it, I think

lib/uplinkc/testdata/object_test.c
mobyvb: ok. Could you explain how this section of the test (`lib/uplinkc/testdata/object_test.c`) relates to `libstorj/test/tests.c`, specifically `check_delete_file()`, the function you linked above?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libstorj"]
+	path = lib/libstorj
+	url = https://github.com/bryanchriswhite/libstorj.git
```
mobyvb: are we planning to have this submodule in master ultimately?

lib/uplinkc/testdata/object_test.c
bryanchriswhite: I wasn't paying attention before when you first commented; this is a C test for libuplink and the tests.c file in libstorj is for testing the libstorj api, they are unrelated.

lib/uplinkc/testdata/object_test.c
bryanchriswhite: :+1:

lib/uplinkc/object.go
bryanchriswhite: Haha yes, that's perfect!

lib/uplinkc/object.go
bryanchriswhite: :+1:

.gitmodules
bryanchriswhite: Good question; I don't know. I like having it as a submodule and I think this is getting removed again at some point.

lib/uplinkc/testdata/object_test.c
bryanchriswhite: 7d1ddc182

bryanchriswhite: :roll_eyes: before you say anything... adding another commit to free that thing

bryanchriswhite: 18e507129

lib/uplinkc/testdata/object_test.c
```diff
@@ -40,7 +49,7 @@ void handle_project(ProjectRef project) {
 
 
     for(int i = 0; i < num_of_objects; i++) {
-        size_t data_len = 1024 * (i + 1) * (i + 1);
+        size_t data_len = pow(10, (double)i) * 1024 * 5;
```
mobyvb: a comment explaining these numbers would be nice

.gitmodules
jtolds: after many long battles with submodules, i'm not opposed to them, but i think i'd rather not see submodules in our go v3 repo. can we have the new libstorj submodule in storj/storj instead of vice versa?

e.g., storj/libstorj has the submodule, instead of this

Dockerfile.jenkins
```diff
@@ -5,7 +5,7 @@ FROM golang:1.12.7
 RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
 RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list
 
-RUN apt-get update && apt-get install -y -qq postgresql-11 redis-server unzip
+RUN apt-get update && apt-get install -y -qq postgresql-11 redis-server unzip libuv1-dev libjson-c-dev nettle-dev
```
kaloyan-raev: Why is nettle still required as dependency?

bryanchriswhite: The cli uses it to encrypt the apikey and encryption key and save it to a file

lib/uplinkc/testdata/object_test.c
```diff
@@ -68,6 +83,8 @@ void handle_project(ProjectRef project) {
 
             upload_commit(uploader, err);
             require_noerror(*err);
+
+            free_uploader_ref(uploader);
```
egonelbre: Why doesn't commit/cancel already free it?

lib/uplinkc/testdata/object_test.c
```diff
@@ -112,6 +129,8 @@ void handle_project(ProjectRef project) {
             download_close(downloader, err);
             require_noerror(*err);
             require(memcmp(data, downloaded_data, data_len) == 0);
+
+            free_downloader_ref(downloader);
```
egonelbre: Why doesn't commit/cancel already free it?

egonelbre: Also no need for the `_ref`

lib/uplinkc/testdata/object_test.c
egonelbre: Also no need for the `_ref`

PR #2574, opened 2019-07-16
wip: support empty path components

isaachess:
What: 

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/lib/uplink/list_test.go
+++ b/lib/uplink/list_test.go
@@ -1,0 +1,192 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package uplink_test
+
+import (
+	"context"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/lib/uplink"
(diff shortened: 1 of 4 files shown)


lib/uplink/list_test.go
```diff
@@ -0,0 +1,233 @@
...
+				},
+			}
+
+			for _, test := range tests {
+				runTest(ctx, t, apiKey, satelliteAddr, test)
+			}
+		})
+}
+
+func runTest(ctx context.Context, t *testing.T, apiKey, satelliteAddr string,
```
ifraixedes: I would rename this function to be `runTestPutGetList` to easy see that it's the companion of `TestPutGetList` and if we start to write more tests in this way in this package this function won't be the one with too generic name.

lib/uplink/list_test.go
```diff
@@ -0,0 +1,233 @@
...
+						"a/b-file",
+						"a/b/slash-file",
+						"a/b////",
+						"a/b//",
+						"a/b//c-file",
+						"a/b//c/",
+						"//bob",
+						"/",
+					},
+					listChallenges: []listChallenge{
```
zeebo: A couple of these challenges only have the same results for all the commands for encrypted paths, right? Specifically, `[]string{"a", "a/"}`, and `[]string{"a/b", "a/b/"}`.

PR #2577, opened 2019-07-16
Implement garbage collection on satellite

navillasa:
Create garbage collection bloom filter as a part of the repair checker loop
https://storjlabs.atlassian.net/browse/V3-1746

Add config options for garbage collection on the satellite
https://storjlabs.atlassian.net/browse/V3-2032

Create a queue on the satellite that is responsible for sending the SNs their bloom filters
https://storjlabs.atlassian.net/browse/V3-2033

Add garbage collection integration tests
https://storjlabs.atlassian.net/brows/V3-2030

What:
- Added a `gc` package at `satellite/gc`, which contains the gc.Service, which runs garbage collection integrated with the metainfoloop, and the gc Observer, which implements the metainfo loop Observer interface and stores all of the filters (about which pieces are good) for each node.
- Added a `gc` config located at `satellite/gc/service.go`
- Creates bloom filters with pieces to be retained inside the metainfo loop
- Adds a `gc.Service.Send` function which will send all `RetainRequests` (or filters with good piece ids) to all storage nodes.

Why:
- Finishing storage node garbage collection

Please describe the tests:
TestGarbageCollection does the following:
* Set up a network with one storagenode
* Upload two objects
* Delete one object from the metainfo service on the satellite
* Trigger a bloom filter generation
* Check that pieces of the deleted object are deleted on the storagenode
* Check that pieces of the kept object are not deleted on the storagenode
 
Since this is the first service that uses the metainfo loop, and tally and checker have not yet been integrated with the metainfo loop, this PR will create an additional case where metainfo could be iterated over concurrently with other metainfo iterations. Once all the services mentioned implement the metainfo loop, the same iteration of metainfo will be used for services waiting to get data


satellite/gc/observer.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"go.uber.org/zap"
+
+	"storj.io/storj/pkg/bloomfilter"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+// Observer implements the observer interface for gc
```
wthorp: This name and comment do little to help me understand what this things does.

mobyvb: How is "Observer implements the metainfo loop Observer interface, allowing us to get segment information"

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+	newPieceCounts := make(map[storj.NodeID]int)
+
+	limiter := sync2.NewLimiter(int(service.config.ConcurrentSends))
+	var errList errs.Group
+	for id, retainInfo := range obs.retainInfos {
+		service.sendRetainFromLimiter(ctx, id, retainInfo, newPieceCounts, limiter, errList)
+	}
+	limiter.Wait()
+	if errList.Err() != nil {
+		service.log.Error("error sending retain infos", zap.Error(errList.Err()))
```
wthorp: Part of me says "if we're just going to log these, why did we do all this fancy errGroup with mutex junk"

satellite/gc/observer.go
wthorp: great!

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+type Service struct {
+	log              *zap.Logger
+	Loop             *sync2.Cycle
+	metainfoloop     *metainfo.Loop
+	transport        transport.Client
+	overlay          overlay.DB
+	lastSendTime     time.Time
+	lastPieceCounts  atomic.Value
+	config           Config
+	errGroupMutex    *sync.Mutex
```
egonelbre: no need for the pointer, show what the mutex protects. See style/testing guides.

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+func (service *Service) Send(ctx context.Context, obs *Observer) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	service.lastSendTime = time.Now().UTC()
+	newPieceCounts := make(map[storj.NodeID]int)
+
+	limiter := sync2.NewLimiter(int(service.config.ConcurrentSends))
+	var errList errs.Group
+	for id, retainInfo := range obs.retainInfos {
+		service.sendRetainFromLimiter(ctx, id, retainInfo, newPieceCounts, limiter, errList)
```
egonelbre: don't pass in the control as an argument, it usually complicates the code.

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+	if errList.Err() != nil {
+		service.log.Error("error sending retain infos", zap.Error(errList.Err()))
+	}
+	service.lastPieceCounts.Store(newPieceCounts)
+
+	return nil
+}
+
+func (service *Service) sendRetainFromLimiter(ctx context.Context, id storj.NodeID, retainInfo *RetainInfo, newPieceCounts map[storj.NodeID]int, limiter *sync2.Limiter, errList errs.Group) {
+	limiter.Go(ctx, func() {
```
egonelbre: inline this func

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+		err := service.sendRetainRequest(ctx, id, retainInfo, newPieceCounts)
+		if err != nil {
+			service.errGroupMutex.Lock()
+			errList.Add(err)
```
egonelbre: log, no need to gather

satellite/gc/observer.go
```diff
@@ -0,0 +1,82 @@
...
+func (observer *Observer) add(ctx context.Context, nodeID storj.NodeID, pieceID storj.PieceID) {
+	var filter *bloomfilter.Filter
+
+	if _, ok := observer.retainInfos[nodeID]; !ok {
+		// If we know how many pieces a node should be storing, use that number. Otherwise use default.
+		numPieces := int(observer.config.InitialPieces)
+		if observer.pieceCounts[nodeID] > 0 {
+			numPieces = observer.pieceCounts[nodeID]
+		}
+		filter = bloomfilter.NewOptimal(numPieces, observer.config.FalsePositiveRate)
```
egonelbre: this must have a size limit, otherwise the grpc call can fail

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Service implements the garbage collection service
+type Service struct {
+	log              *zap.Logger
+	Loop             *sync2.Cycle
+	metainfoloop     *metainfo.Loop
+	transport        transport.Client
+	overlay          overlay.DB
+	lastSendTime     time.Time
+	lastPieceCounts  atomic.Value
```
egonelbre: why do you need the atomic value?

satellite/gc/service.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Service implements the garbage collection service
+type Service struct {
+	log              *zap.Logger
+	Loop             *sync2.Cycle
+	metainfoloop     *metainfo.Loop
+	transport        transport.Client
+	overlay          overlay.DB
+	lastSendTime     time.Time
```
egonelbre: why do you need the last send time

satellite/gc/service.go
mobyvb: I don't think we do anymore. It is leftover from the checker implementation

satellite/gc/service.go
mobyvb: also leftover from checker implementation

satellite/gc/observer.go
```diff
@@ -0,0 +1,82 @@
...
+	return nil
+}
+
+// adds a pieceID to the relevant node's RetainInfo
+func (observer *Observer) add(ctx context.Context, nodeID storj.NodeID, pieceID storj.PieceID) {
+	var filter *bloomfilter.Filter
+
+	if _, ok := observer.retainInfos[nodeID]; !ok {
+		// If we know how many pieces a node should be storing, use that number. Otherwise use default.
+		numPieces := int(observer.config.InitialPieces)
```
kaloyan-raev: Aren't we losing precision by casting `int64` to `int`? If we believe this won't happen then why `IntialPieces` is of type `int64`?

satellite/gc/service.go
mobyvb: by doing it as an inline func instead, I need to take `controlFunc func(context.Context, func()) bool` as an argument into `sendRetainFromLimiter`. This to me seems more confusing. Am I thinking about this wrong?

satellite/gc/observer.go
mobyvb: I usually see int64 used for config variables even if we will never use numbers that big. I can change it to an int

mobyvb: Or maybe I am confusing config variables with protobuf values.

satellite/gc/service.go
mobyvb: True

satellite/gc/observer.go
kaloyan-raev: Yes, protobufs can be either `int32` or `int64`, but not `int`. I don't think this applies to config. I'd vote to change the config to `int`.

satellite/gc/observer.go
kaloyan-raev: can we use streaming?

mobyvb: I don't think the bloom filters should be large enough that we will need streaming. But I am not sure how to calculate bloom filter size

satellite/gc/service.go
mobyvb: nevermind, I misunderstood. Fixing.

satellite/gc/observer.go
mobyvb: either way, what should the size limit be?

egonelbre: max 2MB, default grpc limit is 4MB

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return &Service{
+		log:              log,
+		Loop:             sync2.NewCycle(config.Interval),
+		metainfoloop:     loop,
+		transport:        transport,
+		overlay:          overlay,
+		lastPieceCounts:  lastPieceCounts,
+		config:           config,
+		pieceCountsMutex: sync.Mutex{},
```
kaloyan-raev: Wouldn't `pieceCountsMutex` be initialized even without this statement?

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+	InitialPieces     int     `help:"the initial number of pieces expected for a storage node to have, used for creating a filter" releaseDefault:"400000" devDefault:"10"`
+	FalsePositiveRate float64 `help:"the false positive rate used for creating a filter" releaseDefault:"0.1" devDefault:"0.1"`
+	ConcurrentSends   int     `help:"the number of nodes to concurrently send bloom filters to" releaseDefault:"1" devDefault:"1"`
+}
+
+// Service implements the garbage collection service
+type Service struct {
+	log          *zap.Logger
+	Loop         *sync2.Cycle
+	metainfoloop *metainfo.Loop
```
kaloyan-raev: nit: `metainfoLoop`.

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Send sends the piece retain requests to all storage nodes
+func (service *Service) Send(ctx context.Context, obs *Observer) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	service.pieceCounts = make(map[storj.NodeID]int)
+
+	limiter := sync2.NewLimiter(service.config.ConcurrentSends)
+	for id, retainInfo := range obs.retainInfos {
+		func(id storj.NodeID, retainInfo *RetainInfo) {
```
kaloyan-raev: This is the first time I see such `func` statement without preceding it with `go`. What does it exactly do?

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+				err := service.sendRetainRequest(ctx, id, retainInfo)
+				if err != nil {
+					service.log.Error("error sending retain info", zap.Error(err))
+				}
+			})
+		}(id, retainInfo)
+	}
+	limiter.Wait()
+
+	service.lastPieceCounts = service.pieceCounts
```
kaloyan-raev: Why don't we do this before dealing with the limiter?

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+	defer mon.Task()(&ctx)(&err)
+
+	log := service.log.Named(id.String())
+
+	dossier, err := service.overlay.Get(ctx, id)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	target := &pb.Node{
```
kaloyan-raev: You can just use `dossier.Node` instead of creating `target`.

satellite/gc/service.go
```diff
@@ -0,0 +1,171 @@
...
+		Address: dossier.Address,
+	}
+
+	ps, err := piecestore.Dial(ctx, service.transport, target, log, piecestore.DefaultConfig)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	defer func() {
+		err2 := ps.Close()
+		err = errs.Combine(err, Error.Wrap(err2))
```
kaloyan-raev: `err = errs.Combine(err, Error.Wrap(ps.Close()))` is shorter and less errorprone.

satellite/gc/service.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		// send retain requests
+		limiter := sync2.NewLimiter(service.config.ConcurrentSends)
+		for id, info := range obs.retainInfos {
+			id, info := id, info
+			limiter.Go(ctx, func() {
+				err := service.sendRetainRequest(ctx, id, info)
+				if err != nil {
+					service.log.Error("error sending retain info", zap.Error(err))
```
egonelbre: I just realized that this probably should also use the storage node id for logging.

satellite/gc/service.go
mobyvb: it is an anonymous function that runs immediately in the same thread, rather than starting a separate thread as a goroutine. It allows us to pass in `id` and `retainInfo` as arguments into a goroutine without having to worry about the for loop iterating before the goroutine starts

satellite/gc/service.go
mobyvb: mutex unnecessary now. Also, I think you are right

satellite/gc/service.go
mobyvb: Also, this has been removed in favor of a simpler solution where we just copy `id` and `retainInfo` to new variables before starting `limiter.Go`

satellite/gc/service.go
mobyvb: the order shouldn't really matter. We just need update it after `metainfoLoop.Join`

mobyvb:
@simongui what happens in the case of a false positive is that the storagenode will keep a piece that is supposed to be deleted. In other words, the satellite did not add piece ID x to storagenode y's bloom filter. When storagenode y plugs piece ID x into the bloom filter sent from the satellite, it will think that piece x is in the bloom filter (false positive) and therefore will not delete the piece.

There are already some tests for this in `pkg/bloomfilter/filter_test.go`, so not sure if we need to add more here

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,86 @@
...
+		creationDate: time.Now().UTC(),
+		pieceCounts:  pieceCounts,
+
+		retainInfos: make(map[storj.NodeID]*RetainInfo),
+	}
+}
+
+// RemoteSegment takes a remote segment found in metainfo and adds pieces to bloom filters
+func (pieceTracker *PieceTracker) RemoteSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: Wouldn't be good to trace the `path` passing it to the Task function (i.e. `defer mon.Task()(&ctx, path)(&err)`)

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,86 @@
...
+
+// InlineSegment returns nil because we're only doing gc for storage nodes for now
+func (pieceTracker *PieceTracker) InlineSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	return nil
+}
+
+// adds a pieceID to the relevant node's RetainInfo
+func (pieceTracker *PieceTracker) add(ctx context.Context, nodeID storj.NodeID, pieceID storj.PieceID) {
+	var filter *bloomfilter.Filter
```
ifraixedes: Remove the declaration of this var and just declared in the line 77

satellite/gc/piecetracker.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// InlineSegment returns nil because we're only doing gc for storage nodes for now
+func (pieceTracker *PieceTracker) InlineSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	return nil
+}
+
+// adds a pieceID to the relevant node's RetainInfo
+func (pieceTracker *PieceTracker) add(ctx context.Context, nodeID storj.NodeID, pieceID storj.PieceID) {
```
ifraixedes: `ctx` isn't used and it's an unexported method, so I will remove it; if at some point we need it we could always add it without breaking the public contract.

satellite/gc/service.go
```diff
@@ -0,0 +1,144 @@
...
+
+		transport:    transport,
+		overlay:      overlay,
+		metainfoLoop: loop,
+	}
+}
+
+// Run starts the gc loop service
+func (service *Service) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: If I understand correctly what kind of traces `mon.Task` do I'm wondering how useful is to trace the call to this function because it runs a loop so the trace will be so long and we mostly know when it stops.

I think that it will more useful to trace each iteration of the loop.

satellite/gc/service.go
```diff
@@ -0,0 +1,144 @@
...
+	// TODO retrieve piece counts from overlay (when there is a column for them)
+	lastPieceCounts := make(map[storj.NodeID]int)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) error {
+		pieceTracker := NewPieceTracker(service.log.Named("gc observer"), service.config, lastPieceCounts)
+
+		// collect things to retain
+		err := service.metainfoLoop.Join(ctx, pieceTracker)
+		if err != nil {
+			return Error.Wrap(err)
```
ifraixedes: If I'm not mistaken, this will stop the service which is run by the satellite, however the satellite won't be stopped, I'm wondering if that's OK that this service stops and all the rest of services which the satellite run, continue running and basically there won't be garbage collection until we realize and we restart the satellite server.

satellite/gc/service.go
```diff
@@ -0,0 +1,144 @@
...
+		for _, info := range pieceTracker.retainInfos {
+			mon.IntVal("node_piece_count").Observe(int64(info.Count))
+			mon.IntVal("retain_filter_size_bytes").Observe(info.Filter.Size())
+		}
+		return nil
+	})
+}
+
+func (service *Service) sendRetainRequest(ctx context.Context, id storj.NodeID, info *RetainInfo) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: Wouldn't be worth to trace the `id`? (i.e. `defer mon.Task()(&ctx, id)(&err)`)

satellite/gc/service.go
mobyvb: Not sure what others think, but I think it is okay for a first implementation. If we get data that it is exiting here, we can probably think more about gracefully restarting

satellite/gc/service.go
mobyvb: so do you think I should remove it from here? I am not sure how to trace each iteration of the loop. We do have some stats that we send at the end of each successful iteration

ifraixedes: I'm not entirely sure about monkit but in other systems you get a span on each function call which is traced which extends through the time from the start to the end; if that's the case, I believe that's more useful to know the time on each iteration rather than the `Run` function.

I think that somebody with more knowledge about monkit should take a look to this.

satellite/gc/service.go
ifraixedes: My question is if it's better for a first implementation to just log the error an abort the loop without returning the error rather than returning the error.

I agree that for a first implementation we shouldn't do anything else.

mobyvb: Oh yeah that's a good point

satellite/gc/service.go
mobyvb: updated to send bloom filters last

pkg/bloomfilter/filter.go
```diff
@@ -144,3 +162,8 @@ func (filter *Filter) Bytes() []byte {
 	copy(bytes[3:], filter.table)
 	return bytes
 }
+
+// Size returns the size of Bytes call.
+func (filter *Filter) Size() int64 {
+	return int64(1 + 1 + 1 + len(filter.table))
```
kaloyan-raev: Can we add a comment what `1 + 1 + 1` represents?

satellite/gc/service.go
kaloyan-raev: I think it makes sense to extract the anonymous function passed to `service.Loop.Run()` as a separate method and call `mon.Task` there.

satellite/gc/service.go
```diff
@@ -0,0 +1,146 @@
...
+	// TODO retrieve piece counts from overlay (when there is a column for them)
+	lastPieceCounts := make(map[storj.NodeID]int)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) error {
+		pieceTracker := NewPieceTracker(service.log.Named("gc observer"), service.config, lastPieceCounts)
+
+		// collect things to retain
+		err := service.metainfoLoop.Join(ctx, pieceTracker)
+		if err != nil {
+			service.log.Sugar().Errorf("error joining metainfoloop: %v", Error.Wrap(err))
```
kaloyan-raev: Better use structured logging than sugared logging, i.e.
```go
service.log.Error("error joining metainfoloop", zap.Error(err))
```

satellite/gc/service.go
mobyvb: and should we remove `mon.Task` from here? If the data wouldn't be useful maybe we should take it out

satellite/gc/service.go
mobyvb: I should still do `Error.Wrap`, shouldn't I?

satellite/gc/service.go
simongui: yeah don't keep `mon.Task` we aren't finding valuable because they can make the trace files really big and hard to open at times.

satellite/gc/service.go
kaloyan-raev: There are other places like this.

mobyvb: yes, I will remove all the sugared loggers

ifraixedes: @kaloyan-raev you raised a good point here. I know the benefits of using the structure logger, however I've seen quite a few places (not just in this branch, I'm more talking about others) that the sugared logger is used and I haven't commented because I mostly tried to see if that function is call many times or not before suggesting to change to structure logger.

Is there any kind of hint when to use the sugared or not? (this will facilitate my reviews and improve them)

Thanks in advance.

satellite/gc/service.go
mobyvb: I didn't extract the anonymous function, but I moved the `defer mon.Task` to inside that function

satellite/gc/service.go
kaloyan-raev: I think sugared logging makes sense when you want to format a log message to be nicely readable by end-users. For example, logs that will be displayed as the output of a CLI app.

For the backend, and especially for the satellite, structured logging is much better, because it would enable fine-grained search in tons of logs. Imagine you want to select only the logs related to a specific Node ID.

ifraixedes: ok, I was mostly assessing for performance reasons!

Should we use structured logger for storagenode too?

kaloyan-raev: I believe so.

satellite/gc/service.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return service.Loop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+
+		pieceTracker := NewPieceTracker(service.log.Named("gc observer"), service.config, lastPieceCounts)
+
+		// collect things to retain
+		err = service.metainfoLoop.Join(ctx, pieceTracker)
+		if err != nil {
+			service.log.Error("error joining metainfoloop", zap.Error(Error.Wrap(err)))
```
kaloyan-raev: No need to wrap the error before passing it to `zap.Error`. It makes sense to wrap errors only if you return them up to the call stack.

satellite/gc/service.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		// send retain requests
+		limiter := sync2.NewLimiter(service.config.ConcurrentSends)
+		for id, info := range pieceTracker.retainInfos {
+			id, info := id, info
+			limiter.Go(ctx, func() {
+				err := service.sendRetainRequest(ctx, id, info)
+				if err != nil {
+					service.log.Error("error sending retain info to node", zap.String("node ID", id.String()), zap.Error(err))
```
kaloyan-raev: Use `zap.Stringer("Node ID", id)` instead of `zap.String("node ID", id.String())`.

This is better for performance. `zap.Stringer` will call `id`'s `String()` method only if this message is going to be logged, i.e. log severity is at least set to Error.

satellite/gc/gc_test.go
```diff
@@ -0,0 +1,122 @@
...
+// * Delete one object from the metainfo service on the satellite
+// * Wait for bloom filter generation
+// * Check that pieces of the deleted object are deleted on the storagenode
+// * Check that pieces of the kept object are not deleted on the storagenode
+func TestGarbageCollection(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 1,
+		Reconfigure: testplanet.Reconfigure{
+			Satellite: func(log *zap.Logger, index int, config *satellite.Config) {
+				config.GarbageCollection.FalsePositiveRate = 0.001
```
kaloyan-raev: Using `0.001` means that once in 1000 times we will have a false positive and this test will fail, right? 

Can we use the smallest possible number here?

mobyvb: In theory, yeah. The smaller the number, the larger the bloom filter I think. I don't know what a reasonable value is, but the test has not failed as a result of this yet.

PR #2581, opened 2019-07-17
Log fatal errors with respective severity

kaloyan-raev:
What: Log fatal errors with FATAL severity instead of DEBUG.

Why: Fatal errors should be immediately visible for the user/operator. This happened to me when I tried to start a storage node with a non-existing storage location. The process just quietly stopped with no error info. Only when I switched to debug log I could see the error.

Please describe the tests:
 - Test 1:
 - Test 2:
 

coyle:
Doesn't Fatal Panic? I don't think we should ever panic

kaloyan-raev:
@coyle It doesn't panic, but it exists the process gracefully.

We also used this in many other places. Just search for `zap.S().Fatal` and `.Sugar().Fatal` in the code.

I don't think there is an issue in this case because there is an `os.Exit(1)` call below, so we are exiting the process anyway.

PR #2584, opened 2019-07-17
storagenode/trust: implement fetching peer identity without kademlia and endpoint

egonelbre:
Ensure that we directly dial the whitelisted satellites instead of going through kademlia.

To implement, it sets up a custom credentials which captures the auth info during the initial handshake, this avoids the necessity for a separate Endpoint for fetching information.

As a consequence it also removes ability to trust all satellites, which was a bad idea in the first place.

Why: Currently kademlia lookups keep failing, however it's much faster and reliable to dial directly.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/pkg/transport/fetchidentity.go
+++ b/pkg/transport/fetchidentity.go
@@ -23,7 +23,7 @@ type handshakeCapture struct {
 }
 
 // ClientHandshake does the authentication handshake specified by the corresponding
-// authentication protocol on rawConn for clients. It returns the authenticated
+// authentication protocol on conn for clients. It returns the authenticated
 // connection and the corresponding auth information about the connection.
 func (capture *handshakeCapture) ClientHandshake(ctx context.Context, s string, conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
 	conn, auth, err := capture.TransportCredentials.ClientHandshake(ctx, s, conn)
--- a/pkg/transport/fetchidentity.go
+++ b/pkg/transport/fetchidentity.go
@@ -18,7 +18,6 @@ import (
 )
 
 // handshakeCapture implements a credentials.TransportCredentials for capturing handshake information.
-// TODO: check whether this would be problematic with concurrency or should we thread the authinfo through context?
 type handshakeCapture struct {
 	credentials.TransportCredentials
 
--- a/pkg/transport/fetchidentity.go
+++ b/pkg/transport/fetchidentity.go
@@ -1,0 +1,101 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package transport
+
+import (
+	"context"
+	"net"
+
+	"github.com/zeebo/errs"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/peertls"
+)
+
+// TODO: check whether this would be problematic with concurrency or should we thread the authinfo through context?
+type handshakeCapture struct {
+	credentials.TransportCredentials
+	AuthInfo credentials.AuthInfo
+}
+
+// ClientHandshake does the authentication handshake specified by the corresponding
+// authentication protocol on rawConn for clients. It returns the authenticated
+// connection and the corresponding auth information about the connection.
+func (capture *handshakeCapture) ClientHandshake(ctx context.Context, s string, conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, auth, err := capture.TransportCredentials.ClientHandshake(ctx, s, conn)
+	capture.AuthInfo = auth
+	return conn, auth, err
+}
+
+// ServerHandshake does the authentication handshake for servers. It returns
+// the authenticated connection and the corresponding auth information about
+// the connection.
+func (capture *handshakeCapture) ServerHandshake(conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, auth, err := capture.TransportCredentials.ServerHandshake(conn)
+	capture.AuthInfo = auth
+	return conn, auth, err
+}
+
+// FetchPeerIdentity dials the node and fetches the identity
+func (transport *Transport) FetchPeerIdentity(ctx context.Context, node *pb.Node, opts ...grpc.DialOption) (_ *identity.PeerIdentity, err error) {
+	defer mon.Task()(&ctx, "node: "+node.Id.String()[0:8])(&err)
+
+	if node.Address == nil || node.Address.Address == "" {
+		return nil, Error.New("no address")
+	}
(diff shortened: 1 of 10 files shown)


storagenode/trust/service.go
```diff
@@ -23,55 +24,39 @@ var mon = monkit.Package()
...
 
 // satelliteInfoCache caches identity information about a satellite
 type satelliteInfoCache struct {
 	mu       sync.Mutex
+	url      storj.NodeURL
 	identity *identity.PeerIdentity
-	nodeURL  storj.NodeURL
 }
 
 // NewPool creates a new trust pool using kademlia to find certificates and with the specified list of trusted satellites.
```
ifraixedes: Update this doc comment

storagenode/trust/service.go
```diff
@@ -93,29 +78,15 @@ func (pool *Pool) GetSignee(ctx context.Context, id storj.NodeID) (_ signing.Sig
...
-				pool.trustedSatellites[id] = info
-			}
-			pool.mu.Unlock()
-		}
-	} else {
-		if !ok {
-			return nil, fmt.Errorf("signee %q is untrusted", id)
-		}
+	if !ok {
+		return nil, fmt.Errorf("signee %q is untrusted", id)
```
ifraixedes: should we have a class for this error? (it's probably to do it in another PR)

pkg/transport/fetchidentity.go
```diff
@@ -0,0 +1,101 @@
...
+	AuthInfo credentials.AuthInfo
+}
+
+// ClientHandshake does the authentication handshake specified by the corresponding
+// authentication protocol on rawConn for clients. It returns the authenticated
+// connection and the corresponding auth information about the connection.
+func (capture *handshakeCapture) ClientHandshake(ctx context.Context, s string, conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, auth, err := capture.TransportCredentials.ClientHandshake(ctx, s, conn)
+	capture.AuthInfo = auth
+	return conn, auth, err
```
navillasa: nit - does the error need to be wrapped?

pkg/transport/fetchidentity.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// ServerHandshake does the authentication handshake for servers. It returns
+// the authenticated connection and the corresponding auth information about
+// the connection.
+func (capture *handshakeCapture) ServerHandshake(conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, auth, err := capture.TransportCredentials.ServerHandshake(conn)
+	capture.AuthInfo = auth
+	return conn, auth, err
```
navillasa: nit - does the error need to be wrapped?

pkg/transport/fetchidentity.go
egonelbre: here, no, because it's wrapping and we don't know how the error is handled internally, by wrapping we might lose important interface methods.

pkg/transport/fetchidentity.go
egonelbre: here, no, because it's wrapping and we don't know how the error is handled internally, by wrapping we might lose important interface methods.

pkg/transport/fetchidentity.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	options := append([]grpc.DialOption{
+		grpc.WithTransportCredentials(capture),
+		grpc.WithBlock(),
+		grpc.FailOnNonTempDialError(true),
+		grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", addr)
+			if err != nil {
+				return nil, err
```
navillasa: nit - error wrap?

pkg/transport/fetchidentity.go
ifraixedes: We have to find out a way to let the callers know with errors (mostly methods that return errors) are fine to be inspected like this one and which ones not.

pkg/transport/fetchidentity.go
egonelbre: no, because we are here, no, because it is wrapping internal behavior of grpc and we don't know how the error is handled internally, by wrapping we might lose important interface methods.

ifraixedes: this one bubbles up in the call `conn, err := grpc.DialContext(timedCtx, node.GetAddress().Address, options...)`, no?

storagenode/trust/service.go
navillasa: agree, I think this could be its own err class vs fmt. but I think that would be fine to just add at the top of this file in this pr like `var ErrUntrusted = errs.Class("signee is untrusted:")` so then you could do `return nil, ErrUntrusted.New(id.String())`. but also I don't think this is a critical change.

egonelbre: I currently made it use `Error` directly, since we don't need the distinction at the moment, I didn't add a new class. When we do, we can add more.

pkg/transport/fetchidentity.go
egonelbre: Not sure what you mean?

pkg/transport/fetchidentity.go
egonelbre: I suspect that the error warpping might hide essential interfaces, such as `interface { Temporary() bool }`. https://godoc.org/google.golang.org/grpc#FailOnNonTempDialError, however I haven't checked the errs implementation for it.

pkg/transport/fetchidentity.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/zeebo/errs"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/peertls"
+)
+
+// TODO: check whether this would be problematic with concurrency or should we thread the authinfo through context?
```
zeebo: I'm worried about this TODO. I don't want data races causing weird memory corruption because `AuthInfo` is an interface and so you can tear the writes. It's not too hard to just add a mutex.

egonelbre: Yeah, it's less about concurrent writes, I'm not sure how grpc internally calls the handshakes, i.e. does it call it once, in parallel, does it cache or something else.

egonelbre: Either way, I'll add a mutex.

pkg/transport/fetchidentity.go
ifraixedes: sure, I was only saying that we should find away that the callers can know which methods return original errors so they can do type assertions.

pkg/transport/fetchidentity.go
ifraixedes: When this function (which is an option) is called and due that its result is returned?

I think that the function gets called by `grpc.DialContext(...` as its receiving the options.

PR #2585, opened 2019-07-17
update irreparableDB.GetLimited query to use where instead of offset

JessicaGreben:
**What:**
This PR makes 3 main changes:
1. changes the irreparable `GetLimited()` db query to use `where` clause instead of `offset`
2. increases the `limit` from 1 to 1000 in the `GetLimited()` db query
3. updates the inspector cli irreparable code with these changes

**Why:**
We want to reduce the load on the Satellite master DB.  We calculated which queries the database spent most of its time performing and `irreparableDB.GetLimited` was in the top ten. This goal of this PR is to reduce that.

**Please describe the tests:**
- update the existing tests with the above changes
 
**Please describe the performance impact:**
Overall the performance is improved by 2 things. 
1. previously when using the `offset` the offset is applied to the result of the query. This means that every time we executed the db query in `GetLimited()`, we would retrieve all of the rows from the table then do a sequential scan to the offset. As the offset got larger, the performance was increasingly worse. Now with the change to using `where`, we can seek to that position using the index scan.
2. Previously we were always using a limit of 1 and making a db connection each time. Now we increased the limit to 1000 so this will be reduced. 


pkg/datarepair/checker/checker.go
```diff
@@ -251,30 +251,32 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
-// IrreparableProcess picks items from irreparabledb and add them to the repair
-// worker queue if they, now, can be repaired.
+// IrreparableProcess iterates over all items in the irreparabledb. If an item can
+// now be repaired then it is added to a worker queue.
 func (checker *Checker) IrreparableProcess(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
-
-	limit := 1
-	var offset int64
+	const limit = 1000
```
coyle: why 1000 ?

pkg/datarepair/checker/checker.go
```diff
@@ -251,30 +251,32 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
 	return nil
 }
 
-// IrreparableProcess picks items from irreparabledb and add them to the repair
-// worker queue if they, now, can be repaired.
+// IrreparableProcess iterates over all items in the irreparabledb. If an item can
+// now be repaired then it is added to a worker queue.
```
coyle: nit -- sp/now/not

pkg/datarepair/checker/checker.go
JessicaGreben: i just picked a number that isn't too big or too small. i was thinking either 1000 or 10,000. you have opinions?

coyle: should we make it configurable ? I typically find magic numbers to be a bad pattern.

satellite/satellitedb/irreparabledb.go
```diff
@@ -87,10 +87,14 @@ func (db *irreparableDB) Get(ctx context.Context, segmentPath []byte) (resp *pb.
 	}, nil
 }
 
-// Getlimited number of irreparable segments by offset
-func (db *irreparableDB) GetLimited(ctx context.Context, limit int, offset int64) (resp []*pb.IrreparableSegment, err error) {
+// GetLimited returns a list of irreparable segment info starting after the last segment info we retrieved
+func (db *irreparableDB) GetLimited(ctx context.Context, limit int, lastSeenSegmentPath []byte) (resp []*pb.IrreparableSegment, err error) {
 	defer mon.Task()(&ctx)(&err)
-	rows, err := db.db.Limited_Irreparabledb_OrderBy_Asc_Segmentpath(ctx, limit, offset)
+	const offset = 0
```
coyle: same comment about magic numbers here. maybe a comment explaining why

pkg/datarepair/checker/checker.go
JessicaGreben: i dont think it needs to be configurable. i think we should pick what is most performant.

coyle: agreed but how are we going to determine most performant ?

satellite/satellitedb/irreparabledb.go
JessicaGreben: will do, i can add a comment. but also

pkg/datarepair/checker/checker.go
JessicaGreben: are you saying to replace "now" with "not"?

pkg/datarepair/checker/checker.go
JessicaGreben: im trying to test that right now. I have a dump of the irreparable table with 200k rows that im testing queries on. i will post the details once i have some findings.

pkg/datarepair/checker/checker.go
coyle: depends, I read it as `If an item cannot be repaired then it is added to a worker queue` but looking at the original comment you can ignore me

pkg/datarepair/checker/checker.go
coyle: 👍 I assumed it would be easier to determine the most performant choice if it was a configurable number

JessicaGreben: im just running queries against the db table directly so its easy

JessicaGreben: i did some testing by running queries with `explain analyze` against the irreparable table with 230k rows (a dump from stefan's prod SA). This showed that that a limit of 1k is four times faster than 10k. The query plan with 1k limit does an index scan, while the query plan of 10k changes to do a bitmap index/heap scan. While these results are specific to this table and the size, I think its fine to hard code 1k limit for now.  If the performance is still a bottleneck in the future we can always modify downt the road.

cmd/inspector/main.go
```diff
@@ -553,22 +553,25 @@ func getSegments(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return ErrInspectorDial.Wrap(err)
 	}
-
-	length := irreparableLimit
-	var offset int32
+	var lastSeenSegmentPath = []byte{}
```
egonelbre: why isn't this a `storj.Path`, `string` or something similar?

pkg/datarepair/irreparable/irreparable.go
```diff
@@ -15,8 +15,8 @@ type DB interface {
 	IncrementRepairAttempts(ctx context.Context, segmentInfo *pb.IrreparableSegment) error
 	// Get returns irreparable segment info based on segmentPath.
 	Get(ctx context.Context, segmentPath []byte) (*pb.IrreparableSegment, error)
-	// GetLimited number of segments from offset
-	GetLimited(ctx context.Context, limit int, offset int64) ([]*pb.IrreparableSegment, error)
+	// GetLimited returns a list of irreparable segment info starting after the last segment info we retrieved
+	GetLimited(ctx context.Context, limit int, lastSeenSegmentPath []byte) ([]*pb.IrreparableSegment, error)
```
egonelbre: Usually this is called `cursor` in our database code.

JessicaGreben: good point.

cmd/inspector/main.go
JessicaGreben: yeah i guess i could have. I didn't even consider that though since it would just add a bunch of code to convert between string and []byte. Maybe that is worth it? im no sure.

PR #2586, opened 2019-07-17
satellite/console: add lowercase to reset password email input

Barterio:
What: 
Implemented normalized_email field to users table to have opportunity to compare emails without changing user email representation
Why:

User is unable to reset password if enter email with any UpperCase letter
example@storj.io
not the same as
Example@storj.io

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/consoleweb/consoleql/query.go
```diff
@@ -137,6 +138,8 @@ func rootQuery(service *console.Service, mailService *mailservice.Service, types
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 					email, _ := p.Args[FieldEmail].(string)
 
+					email = strings.ToLower(email)
+
 					user, err := service.GetUserByEmail(p.Context, email)
 					if err != nil {
 						return false, fmt.Errorf("%s is not found", email)
```
rikysya: use `errs.New`

satellite/console/consoleweb/consoleql/query.go
```diff
@@ -137,6 +138,8 @@ func rootQuery(service *console.Service, mailService *mailservice.Service, types
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
 					email, _ := p.Args[FieldEmail].(string)
 
+					email = strings.ToLower(email)
```
rikysya: May be we should move it to `service` lvl

stefanbenten: Lets move that into the service and also use it across the board as normalized entries in the database.

crawter: but in our CreateUser method in console.Service we "normalize" email making in lowercase

crawter: and its the only place where we can set an email

stefanbenten: Take a look at the current database and you will see that there are entries in it with "non" lowercase emails.

egonelbre: The email casing must be preserved for presentation.

stefanbenten: why @egonelbre ?

egonelbre: Let's say my name is written as "AbCdE", I would expect it to be displayed that way. I.e. I expect to enter my email as I use it and see as I use it. There are of course issues with unicode with regards to case transformation, but, don't change the name or email of a person.

We can internally have a canonical representation, but it shouldn't be showed to the user.

Barterio: In this case we should disable email normalizing during user registration and login as well

egonelbre: Just to clarify, logging in should still work with differently cased email.

bryanchriswhite: Can someone explain why we're changing the case here?

Barterio: @stefanbenten @egonelbre I'm going to rewrite `satellitedb/user GetUserByEmail` to be case insensitive.
Also I'll Add case insensitive check to projectMembers and reset password queries as well.

What do you think about this solution?

egonelbre: Yes, making `GetUserByEmail` can be case-insensitive.

However, just to clarify:

1. We need two fields `email` and `normalized_email`.
2. When we query by email, we'll use `normalized_email`.
3. When we show to the user, we'll use `email`.

`normalized_email` would be in this case lower-cased. `email` whatever the user used to sign-up.

egonelbre: Also `normalized_email` might remove dots, sub-addressing etc.

Barterio: @egonelbre We can make normalization by querying like `` select * from users where UPPER(email) = UPPER(input).
Or this is bad approach?

egonelbre: That approach might be bad performance wise, since I think that would cause a full scan of the users table. If we want some more complex canonicalization, then it would need some more complex things.

But, it could work for the short term.

satellite/satellitedb/users.go
```diff
@@ -143,3 +145,7 @@ func userFromDBX(ctx context.Context, user *dbx.User) (_ *console.User, err erro
 
 	return &result, nil
 }
+
+func normalizeEmail(email string) string {
+	return strings.ToUpper(email)
```
egonelbre: Please verify that the SQL `UPPER` behaves the same as `strings.ToUpper`

stefanbenten:
Also please update the PR description @Barterio

satellite/satellitedb/users.go
Barterio: Added few tests

PR #2590, opened 2019-07-18
Use trusted pool to get satellite addr in storagenode orders send

mobyvb:
What: 
Instead of using kad to lookup the satellite address when sending order limits on the storagenode, get the address directly from the trust pool

Why:
We are seeing weird kad issues on some storagenodes, which are unable to properly lookup satellites in their routing tables. This does not fix the root issue, but should at least allow nodes to send their orders to satellites properly

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/peer.go
```diff
@@ -258,6 +258,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, ver
 			peer.Transport,
 			peer.Kademlia.Service,
```
egonelbre: Is kademlia still used here somewhere, maybe it can be removed?

mobyvb: it is not used anymore. I removed it

PR #2601, opened 2019-07-19
V3-1839 web satellite auth cookie subdomain differentiation

crawter:
What: tokenManager improved and now stores cookie with specified subdomain + domain prefix.

Why: If two satellites share the same domain but have different subdomains (mars.tardigrade.io, saturn.tardigrade.io), we want to make sure that logging into one does not share access cookies and secrets with another.

Please describe the tests:
 - Test 1:
 - Test 2:
 

web/satellite/src/utils/authToken.ts
```diff
@@ -0,0 +1,46 @@
...
+        document.cookie = AuthToken.tokenKey + '=' + tokenValue + '; path=/';
+    }
+
+    public static remove(): void {
+        document.cookie = AuthToken.tokenKey + '=; path=/';
+    }
+
+    private static getCookie(cname: string): string {
+        let name: string = cname + '=';
+        let decodedCookie: string = decodeURIComponent(document.cookie);
```
VinozzZ: Do we also need to cast `document` as `any` here?

crawter: we didn't set --strictNullChecks flag, so document or document.location both could be null

to prevent this error i can cast document to any or use smth like this: document!.location!.hostname + AuthToken.tokenKeySuffix;

crawter: @VinozzZ maybe we should think about modifying our tsconfig, so if u will have some thoughts about it - please, share them with us. Thanks!

PR #2602, opened 2019-07-19
CompareAndSwap in KeyValueStore

kaloyan-raev:
What: Adds `CompareAndSwap` method to the `KeyValueStore` interface and implements it in all of the TestStore, BoltDB, Redis and PostgreSQL implementations.

Why: We need to be able to atomically update the pieces in metainfo pointers. See https://storjlabs.atlassian.net/browse/V3-2088 for details.

Please describe the tests:
 - All existing tests are updated to use `CompareAndSwap` instead of `Put` for updating the key.
 - New constraints tests added:
   - `CompareAndSwap Empty Key`
   - `CompareAndSwap Empty Old Value`
   - `CompareAndSwap Empty New Value`
   - `CompareAndSwap Empty Both Empty Values`
   - `CompareAndSwap Missing Key`
   - `CompareAndSwap Value Changed`
   - `CompareAndSwap Concurrent`
 

storage/postgreskv/client.go
```diff
@@ -317,3 +318,67 @@ func (client *Client) Iterate(ctx context.Context, opts storage.IterateOptions,
...
+}
+
+// CompareAndSwapPath atomically compares and swaps oldValue with newValue in the given bucket
+func (client *Client) CompareAndSwapPath(ctx context.Context, bucket, key storage.Key, oldValue, newValue storage.Value) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	if key.IsZero() {
+		return storage.ErrEmptyKey.New("")
+	}
+
+	tx, err := client.pgConn.BeginTx(ctx, nil)
```
egonelbre: This should be possible without transactions.

`UPDATE pathdata(bucket, fullpath, metadata) VALUES ..... WHERE metadata = old-value`, this will fail when there is no such old value/

storage/testsuite/test.go
```diff
@@ -85,4 +86,68 @@ func testConstraints(t *testing.T, store storage.KeyValueStore) {
...
+			t.Fatalf("failed to update %q: %v -> %v: %v", key, val, nil, err)
+		}
+
+		value, err := store.Get(ctx, key)
+		if err == nil {
+			t.Fatalf("got deleted value %q = %v", key, value)
+		}
+	})
+
+	t.Run("CompareAndSwap Empty Both Empty Values", func(t *testing.T) {
```
egonelbre: Also write concurrent compare-and-swap.

storage/postgreskv/client.go
kaloyan-raev: Unfortunately, it won't fail, but it will be successful with 0 rows affected.

I can check if 0 rows are affected, but I cannot determine if this is due to a missing key (where we should return `ErrKeyNotFound`), or due to changed value (where we should return `ErrValueChanged`).

Any idea how to implement this without transaction?

egonelbre: @thepaul any ideas?

storage/testsuite/test.go
kaloyan-raev: I added a test with concurrent compare-and-swaps. It works perfectly on my system, but in Jenkins tests fail "i/o timeout". I tried increasing the timeouts, but still the same...

kaloyan-raev: I resolve the issue by installing `redis-server` on Jenkins. If `redis-server` is not available the Redis implementation falls back to `miniredis`. The latter seems to have an issue with high concurrency.

storage/postgreskv/client.go
thepaul: The best thing I can think of here would be a CTE, looking something like

```
WITH matching_key AS (
    SELECT * FROM pathdata WHERE bucket = $1 AND fullpath = $2
), updated AS (
    UPDATE pathdata
        SET metadata = $3
        FROM matching_key mk
        WHERE pathdata.metadata = $4
            AND pathdata.bucket = mk.bucket
            AND pathdata.fullpath = mk.fullpath
        RETURNING 1
)
SELECT EXISTS(SELECT 1 FROM matching_key) AS key_present, EXISTS(SELECT 1 FROM updated) AS value_set;
```

If it returns `t, t` then the key was present and the value was updated.
If it returns `t, f` then the key was present but the value did not match expectations, so it was not updated.
If it returns `f, f` then the key was not present.
If it returns `f, t` then your database system is on fire.

thepaul: When value is nil and you want a delete instead, then it would be mostly the same, but the `updated` CTE would look like this instead:

```
    DELETE FROM pathdata
        USING matching_key mk
        WHERE pathdata.metadata = $4
            AND pathdata.bucket = mk.bucket
            AND pathdata.fullpath = mk.fullpath
        RETURNING 1
```

I see that this function also falls back on an `INSERT` sometimes, but the semantics there seem confusing to me in the context of a compare-and-swap operation, so I'm not entirely sure how to translate that.

wthorp: You could self join and use a returning clause ala https://stackoverflow.com/questions/7923237/return-pre-update-column-values-using-sql-only-postgresql-version

wthorp: ^^ this starts to enter the realm of hard-to-understand code though

thepaul: @wthorp That's sort of what I'm doing above, but in a more readable way imo. Or does that approach buy us something else I'm not seeing yet?

wthorp: Mostly I just didn't refresh / see your post before I posted.  I like readable.  :D

kaloyan-raev: @thepaul 
> I see that this function also falls back on an INSERT sometimes, but the semantics there seem confusing to me in the context of a compare-and-swap operation, so I'm not entirely sure how to translate that.

This is the case of `CompareAndSwap(ctx, "some-key", nil, "some-value")`, i.e. the key is set to the value only if the key did not exist yet.

kaloyan-raev: @thepaul thanks for your suggestions! I believe I applied them correctly in https://github.com/storj/storj/pull/2602/commits/e60a89a7f72f3349adbe5214762cae0325e83ea4. All tests are fine.

storage/postgreskv/client.go
```diff
@@ -332,58 +331,85 @@ func (client *Client) CompareAndSwapPath(ctx context.Context, bucket, key storag
...
+		), updated AS (
+			DELETE FROM pathdata
+				USING matching_key mk
+				WHERE pathdata.metadata = $3::BYTEA
+					AND pathdata.bucket = mk.bucket
+					AND pathdata.fullpath = mk.fullpath
+				RETURNING 1
+		)
+		SELECT EXISTS(SELECT 1 FROM matching_key) AS key_present, EXISTS(SELECT 1 FROM updated) AS value_updated
+		`
```
thepaul: From this point on, the logic ought to be identical to the other branch, starting at line 402. The only difference between these cases should be the query string. (This case doesn't need any argument for `$4`, but having the extra argument passed to `QueryRow()` shouldn't hurt anything.) Would it make sense to combine these two cases?

storage/postgreskv/client.go
```diff
@@ -317,3 +317,99 @@ func (client *Client) Iterate(ctx context.Context, opts storage.IterateOptions,
...
+			return Error.Wrap(err)
+		}
+		return storage.ErrValueChanged.New(key.String())
+	}
+
+	if oldValue == nil {
+		q := `INSERT INTO pathdata (bucket, fullpath, metadata) VALUES ($1::BYTEA, $2::BYTEA, $3::BYTEA)`
+		_, err = client.pgConn.Exec(q, []byte(bucket), []byte(key), []byte(newValue))
+		if err, ok := err.(*pq.Error); ok {
+			if err.Code == "23505" { // unique_violation
```
thepaul: This could be

    if err.Code.Name() == "unique_violation" {

I think. That would be slightly safer during future maintenance, but would also be very slightly slower due to a map lookup and a longer string compare. I guess I don't care much which approach you use, but if we use `"23505"` as a constant, I would very much prefer for it to be a named constant.

thepaul: Another approach to this which would avoid inspecting error codes entirely would look like

```sql
    INSERT INTO pathdata (bucket, fullpath, metadata) VALUES ($1::BYTEA, $2::BYTEA, $3::BYTEA)
        ON CONFLICT DO NOTHING
        RETURNING 1
```

You'd run that as though it was a select query. If it returned any rows, then the insert succeeded. If it returned zero rows, then there was a uniqueness violation.

storage/postgreskv/client.go
kaloyan-raev: What you say makes sense, but `QueryRow()` actually fails with ` pq: got 4 parameters but the statement requires 3` due to the extra argument.

thepaul: Oh! I guess they do check that after all. Well, you could put a `$4` into the query in a way that doesn't have any actual effect (example: change the last line to have `EXISTS(SELECT $4 FROM matching_key)` but that would be confusing. Or maybe you could do the `QueryRow` calls in the two branches separately, but still combine the remainder of the logic?

storage/postgreskv/client.go
kaloyan-raev: You are a wizard! This query actually worked :)

storage/postgreskv/client.go
kaloyan-raev: I went with including `QueryRow` in both branches. Chech the latest commit.

storage/redis/client.go
```diff
@@ -229,3 +212,82 @@ func (client *Client) allPrefixedItems(prefix, first, last storage.Key) (storage
...
+			}
+
+			if newValue == nil {
+				return nil
+			}
+
+			// runs only if the watched keys remain unchanged
+			_, err = tx.Pipelined(func(pipe redis.Pipeliner) error {
+				return put(ctx, pipe, key, newValue, client.TTL)
+			})
```
zeebo: Isn't this missing a return here?

thepaul: It looks like it was intended to fall down to the `err != nil` check on line 239. Otherwise `err` would probably get a `:=` operator here.

zeebo: Yes, but if there's no error in this, then we fall even further into logic that we shouldn't hit.

thepaul: Oh dang I totally missed that. I'll add that because it's really late for Kaloyan and he might not be around atm

kaloyan-raev: Oh, this is a good catch @zeebo !

PR #2609, opened 2019-07-20
satellite/satellitedb: add updateEarnedCredits method for user_credits table

VinozzZ:
Will add migration test data once the implementation is reviewed and approved
What: 
Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/service.go
```diff
@@ -121,6 +123,15 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
 	if err != nil {
 		return nil, errs.New(internalErrMsg)
 	}
+	if user.PartnerID != "" {
+		offerType = rewards.Partner
+	}
+
+	//TODO: Create a current offer cache to replace database call
+	currentReward, err := s.rewards.GetCurrentByType(ctx, offerType)
+	if err != nil && err.Error() != rewards.NoCurrentOfferErr {
```
thepaul: This is checking whether `err` is a `rewards.NoCurrentOfferErr` by checking whether the stringified version of `err` is equal to the `rewards.NoCurrentOfferErr` string. This can be a fragile way to do things, because another developer might unknowingly change the error instance before it gets to this point (maybe they are adding helpful explanatory information about what's happening now, or they are `.Wrap()`ping the error in another error class, etc). If the error instance changes at all, this check will no longer work as intended.

But! There is a more idiomatic way to do this (well, idiomatic for our codebase, anyway) which also solves that problem. First, you would make `rewards.NoCurrentOfferErr` be an error class instead of a string. In satellite/rewards/rewards.go:

```go
var (
    MaxRedemptionErr  = errs.Class("This offer redemption has reached its capacity")
    NoCurrentOfferErr = errs.Class("no current offer")
)
```

Then, when returning one of those errors, you make a new err instance from the class. For example, in offers.go:

```go
    if err == sql.ErrNoRows {
        return nil, rewards.NoCurrentOfferErr.New("offerType=%d", offerType)
    }
```

(You don't have to use that format string; that's just an example of something that might be useful in a NoCurrentOfferErr instance. The empty string would work too.) Finally, when we want to check if the error is not from `NoCurrentOfferErr`, on this line of the code, you could do something like

```go
    if err != nil && !rewards.NoCurrentOfferErr.Has(err) {
```

That will still work if extra information is added to the err string or if the err instance gets `.Wrap()`ped in other errror classes too.

satellite/console/service.go
```diff
@@ -405,12 +431,16 @@ func (s *Service) ActivateAccount(ctx context.Context, activationToken string) (
...
 
 	user.Status = Active
-
 	err = s.store.Users().Update(ctx, user)
 	if err != nil {
 		return errs.New(internalErrMsg)
 	}
 
+	err = s.store.UserCredits().UpdateEarnedCredits(ctx, user.ID)
+	if err != nil && err.Error() != NoCreditForUpdateErr {
```
thepaul: `NoCreditForUpdateErr` should also probably be an `errs.Class`.

satellite/satellitedb/migrate.go
```diff
@@ -1044,6 +1044,14 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 						WHERE type=1 AND status=1 AND id=2;`,
 				},
 			},
+			{
+				Description: "Create partial index for user_credits table",
```
thepaul: This ought to have a comment saying why, since I think it's not obvious. Perhaps "This partial unique index enforces uniqueness among `(id, offer_id)` pairs for users that have signed up but are not yet activated (`credits_earned_in_cents`=0). Among users that _are_ activated, uniqueness of `(id, offer_id)` pairs is not required or desirable."

satellite/console/consoleweb/consoleql/mutation.go
```diff
@@ -83,60 +79,26 @@ func rootMutation(log *zap.Logger, service *console.Service, mailService *mailse
...
 
-					offerType := rewards.FreeCredit
-
 					createUser := fromMapCreateUser(input)
 
 					secret, err := console.RegistrationSecretFromBase64(secretInput)
 					if err != nil {
-						log.Error("register: failed to parse secret",
-							zap.String("rawSecret", secretInput),
+						log.Error("register: failed to create account",
```
VinozzZ: @ifraixedes Do you think this is a good way to do it?

ifraixedes: I think so according our conventions.

I would add to the `Debug` message something like `"register: failed to create an account with: "

It will be easier to related the this error message  with the debug one

satellite/console/consoleweb/consoleql/mutation.go
```diff
@@ -83,60 +79,26 @@ func rootMutation(log *zap.Logger, service *console.Service, mailService *mailse
...
 							zap.Error(err))
+						log.Debug("register: ", zap.String("rawSecret", secretInput))
```
coyle: I don't think we should log the secret

satellite/console/service.go
```diff
@@ -155,6 +166,21 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
+				UserID:        u.ID,
+				OfferID:       currentReward.ID,
+				ReferredBy:    nil,
+				CreditsEarned: currency.Cents(0),
+				ExpiresAt:     time.Now().UTC().AddDate(0, 0, currentReward.InviteeCreditDurationDays),
+			}
+
+			err = tx.UserCredits().Create(ctx, newCredit)
+			if err != nil {
+				return err
```
coyle: should this error be wrapped?

thepaul: Looks like the `Create` method there always returns errs-created error instances anyway, so no need to wrap them for the purpose of attaching a traceback. But I'm not entirely sure what our policy is. Do we always wrap anyway, unless there are error values that break if wrapped? It's getting confusing to me

satellite/satellitedb/testdata/postgres.v48.sql
```diff
@@ -0,0 +1,351 @@
...
+	project_id bytea NOT NULL REFERENCES projects( id ) ON DELETE CASCADE,
+	payer_id bytea NOT NULL REFERENCES user_payments( user_id ) ON DELETE CASCADE,
+	payment_method_id bytea NOT NULL,
+	is_default boolean NOT NULL,
+	created_at timestamp with time zone NOT NULL,
+	PRIMARY KEY ( id )
+);
+CREATE INDEX bucket_name_project_id_interval_start_interval_seconds ON bucket_bandwidth_rollups ( bucket_name, project_id, interval_start, interval_seconds );
+CREATE UNIQUE INDEX bucket_id_rollup ON bucket_usages ( bucket_id, rollup_end_time );
+CREATE INDEX injuredsegments_attempted_index ON injuredsegments ( attempted );
```
thepaul: This line is the one that differs from postgres.v47.sql. But I _think_ this is an old version of what we want to add. Should this match the `credits_earned_user_id_offer_id` index definition in migrate.go?

VinozzZ: Oh, I just copied from the `satellitedb.dbx.postgres.sql` file. Maybe my branch is not up-to-date? I will double check

satellite/satellitedb/usercredits.go
```diff
@@ -85,7 +87,13 @@ func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit)
 		referrerID = userCredit.ReferredBy[:]
 	}
 
-	result, err := c.db.DB.ExecContext(ctx, c.db.Rebind(statement), userCredit.UserID[:], userCredit.OfferID, userCredit.CreditsEarned.Cents(), userCredit.ExpiresAt, referrerID, new([]byte), userCredit.OfferID, userCredit.OfferID)
+	var result sql.Result
+	var err error
+	if c.tx != nil {
+		result, err = c.tx.Tx.ExecContext(ctx, c.db.Rebind(statement), userCredit.UserID[:], userCredit.OfferID, userCredit.CreditsEarned.Cents(), userCredit.ExpiresAt, referrerID, new([]byte), userCredit.OfferID, userCredit.OfferID)
+	} else {
+		result, err = c.db.DB.ExecContext(ctx, c.db.Rebind(statement), userCredit.UserID[:], userCredit.OfferID, userCredit.CreditsEarned.Cents(), userCredit.ExpiresAt, referrerID, new([]byte), userCredit.OfferID, userCredit.OfferID)
```
thepaul: Ugh, DBX makes this ugly, doesn't it. There isn't a common-interface way to get the `DB` member from `c.tx` and `c.db`, so you have to have an if-else.

One thing you could do- just to reduce the risk of one of these calls changing and not matching the other- is use a temporary interface type to unify the `ExecContext()` calls, if not the part about getting the DB handle. I think it would look something like

```go
    var dbExec interface {
        ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
    }
    if c.tx != nil {
        dbExec = c.tx.Tx
    } else {
        dbExec = c.db.DB
    }
    result, err = dbExec.ExecContext(ctx, .....)
```

It's not a _lot_ better, though, granted. Only a suggestion.

satellite/console/service.go
VinozzZ: I'm not sure what our policy exactly is for wrapping errors either. I found this styling guide in our doc https://github.com/storj/docs/blob/master/code/Style.md#error-handling

jtolds: the real policy is that if an error isn't already always a zeebo/errs generated error, it should be wrapped. so the answer is contextual. the main point of zeebo/errs for us is to add stack traces, so if there isn't a stack trace attached to the error yet, wrapping adds one. there's one exception to this as egon recently pointed out which is errors that have very specific behavior associated such as io.EOF. but that won't happen here, and i believe we've extended dbx to use zeebo/errs, so i don't think you need to wrap here. on the flip side, because this won't be io.EOF, it also won't hurt to wrap here.

satellite/satellitedb/testdata/postgres.v48.sql
VinozzZ: I feel a little confused on the index creation for migrations. Should we add the `CREATE INDEX` manually into the test data file?

thepaul: I think so. If I remember right, we're supposed to copy the most recent `postgres.v*.sql` file to `postgres.v${YOUR_MIGRATION_NUMBER}.sql`, and then modify it so that it creates the same schema that would result from applying the migration with the same number.

thepaul: But I also think there is an automated test that checks this, so if that didn't fail, then maybe everything is fine?

satellite/console/service.go
zeebo: Additionally, if you don’t care about adding a class to the error, using `errs.Wrap` only adds a stack trace if one does not already exist. (Class wrapping also has this property, and `errs.Wrap` can be thought of wrapping with no class.) The intent of that function is to make the decision easy for readers and writers: always call `errs.Wrap` unless you care about io.EOF as described above.

ifraixedes: I will add and don't wrap if the error is a GRPC one and it has to be sent through the wire

PR #2611, opened 2019-07-22
Add GetObject method to Metainfo

mniewrzal:
What: Adds missing `GetObject` method to new Metainfo API

Why: Without this we are not able to do some operations like a segment listing.

Please describe the tests:
 - Test 1: added automatic test
 

pkg/pb/metainfo.proto
```diff
@@ -24,6 +24,7 @@ service Metainfo {
     // Object
     rpc BeginObject(ObjectBeginRequest) returns (ObjectBeginResponse);
     rpc CommitObject(ObjectCommitRequest) returns (ObjectCommitResponse);
+    rpc GetObject(ObjectGetRequest) returns (ObjectGetResponse);
```
coyle: I don't recall this from the original discussion re RPC methods. What is the reasoning here?

satellite/metainfo/metainfo.go
```diff
@@ -1032,6 +1032,75 @@ func (endpoint *Endpoint) CommitObject(ctx context.Context, req *pb.ObjectCommit
...
 
+// GetObject gets single object
+func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetRequest) (resp *pb.ObjectGetResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	keyInfo, err := endpoint.validateAuth(ctx, macaroon.Action{
+		Op:            macaroon.ActionRead,
+		Bucket:        req.Bucket,
+		EncryptedPath: req.EncryptedPath,
+		Time:          time.Now(),
```
coyle: should we do UTC() ?

pkg/pb/metainfo.proto
mniewrzal: yeah, we missed that one. We need a method to access an object in general and to have object `stream_id` for segment methods like e.g. `ListSegments`. See https://storj.slack.com/archives/CCXRMC54M/p1563789833001000

satellite/metainfo/metainfo.go
mniewrzal: no, it should be handled by apikey/macroons validation

uplink/metainfo/client.go
```diff
@@ -408,6 +408,32 @@ func (client *Client) CommitObject(ctx context.Context, streamID storj.StreamID)
...
+	})
+	if err != nil {
+		return storj.Object{}, storj.StreamID{}, Error.Wrap(err)
+	}
+
+	object := storj.Object{
+		Bucket: storj.Bucket{
+			Name: string(response.Object.Bucket),
+		},
+		Path:    string(response.Object.EncryptedPath),
```
kaloyan-raev: nit: typecast to `storj.Path` instead of `string`.

uplink/metainfo/client.go
```diff
@@ -408,6 +408,32 @@ func (client *Client) CommitObject(ctx context.Context, streamID storj.StreamID)
...
+		Created: response.Object.CreatedAt,
+		Expires: response.Object.ExpiresAt,
+		// TODO custom type for response object or modify storj.Object
```
kaloyan-raev: How about size, metadata and other fields?

mniewrzal: Currently, we are not storing `Object` anywhere so only some information can be extracted from segment. `storj.Object` doesn't fit well into new Metainfo structs e.g. metainfo is `map[string]string` instead of `[]byte`. I believe before we will finish whole Metainfo refactoring we still need to adjust client-side objects. That's why I left TODO.

uplink/metainfo/client.go
mniewrzal: done

uplink/metainfo/client.go
kaloyan-raev: OK. I see.

PR #2614, opened 2019-07-22
satellite/console: referral link for individual users

VinozzZ:
What: 
- add logic for individual user's referral link in satellite GUI
- change invitee_credits column in database to be the one that stores free credits configuration
Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

web/satellite/src/views/register/Register.vue
```diff
@@ -48,7 +49,7 @@
             let referralIds = ids ? JSON.parse(atob(ids)) : undefined;
```
rikysya: Is there any reason we use `route params` to pass ids? Also, I am curious what is the difference between `refUserID` and `partnerID`?

VinozzZ: The route params is just a preference of how marketing want the referral link look like. I will add a comment to explain the difference between `refUserID` and `partnerID`. PartnerID is part of the `User` struct if it exists on the URL, it means that a partner has referred a user. RefUserID is a existing user's ID, it means that the new user is referred by an existing user instead of a partner

PR #2541, opened 2019-07-12
SN DB Optimization: Add rollups to bandwidth usage

ethanadams:
What: 
Add hourly rollups for bandwidth usage

Why:
Improve query performance and reduce disk usage

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact:

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?
 - [ ] Do the database access patterns make sense?

PR #2617, opened 2019-07-22
Re-enable SN bandwidth rollups. Fix SN bandwidth rollup unique constraint issue. Re-organize service code

ethanadams:
What: 
Add a date range for rollups to avoid duplicate data issue.
Re-organized service code based on suggestions in the original PR (https://github.com/storj/storj/pull/2541#pullrequestreview-264083750)

Why:
Some SNs reported fatal errors when the rollup returns a unique constraint issue.  This appeared to be caused by a failed delete of the last rollup.

Please describe the tests:
 - Test 1: Added test to add data back to bandwidth_usage after the hours rollup occurs.
 - Test 2:
 

storagenode/bandwidth/service_test.go
```diff
@@ -0,0 +1,43 @@
...
+// See LICENSE for copying information.
+
+package bandwidth_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"storj.io/storj/internal/testcontext"
```
ReneSmeekes:
```suggestion

	"storj.io/storj/internal/testcontext"
```

Fix linting error for import style

storagenode/storagenodedb/bandwidthdb.go
```diff
@@ -197,6 +182,7 @@ func (db *bandwidthdb) Rollup(ctx context.Context) (err error) {
 		SELECT datetime(strftime('%Y-%m-%dT%H:00:00', created_at)) created_hr, satellite_id, action, SUM(amount)
```
egonelbre: 1. The date-range probably should come as an argument instead of internally computed.
2. The coalescing interval as well.
3. This code should assume that there can be bandwidths in bandwidth_usage older than 1 day, and they should be properly handled.

storagenode/storagenodedb/bandwidthdb.go
```diff
@@ -197,6 +182,7 @@ func (db *bandwidthdb) Rollup(ctx context.Context) (err error) {
...
 			FROM bandwidth_usage
 		WHERE datetime(created_at) < datetime(?)
+		AND datetime(created_at) >= coalesce((SELECT datetime(MAX(interval_start), '+1 hour') FROM bandwidth_usage_rollups), datetime(0, 'unixepoch'))
 		GROUP BY created_hr, satellite_id, action;
 
 		DELETE FROM bandwidth_usage WHERE datetime(created_at) < datetime(?);
```
egonelbre: This would discard some values that weren't accounted before.

storagenode/storagenodedb/bandwidthdb.go
```diff
@@ -197,6 +182,7 @@ func (db *bandwidthdb) Rollup(ctx context.Context) (err error) {
 		SELECT datetime(strftime('%Y-%m-%dT%H:00:00', created_at)) created_hr, satellite_id, action, SUM(amount)
 			FROM bandwidth_usage
 		WHERE datetime(created_at) < datetime(?)
+		AND datetime(created_at) >= coalesce((SELECT datetime(MAX(interval_start), '+1 hour') FROM bandwidth_usage_rollups), datetime(0, 'unixepoch'))
```
egonelbre: This probably shouldn't be necessary.

storagenode/storagenodedb/bandwidthdb.go
ethanadams: 1. This select groups by hour, satellite, action.  So it's necessary to have it done in the select so we can handle multiple hours of followups at one time
2. This could be passed in, but it would require 2 trips to the DB
3. Added a test to show the rollup handles data older than 1 day

storagenode/bandwidth/db_test.go
```diff
@@ -145,6 +148,120 @@ func TestCachedBandwidthMonthRollover(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer ctx.Check(db.Close)
+
+	t.Run("Sqlite", func(t *testing.T) {
+		err := db.CreateTables()
+		if err != nil {
+			t.Fatal(err)
+		}
+		testID1 := teststorj.NodeIDFromString("testId1")
```
egonelbre:
```
storj.NodeID{1}
storj.NodeID{2}
storj.NodeID{3}
```

storagenode/bandwidth/db_test.go
```diff
@@ -145,6 +148,120 @@ func TestCachedBandwidthMonthRollover(t *testing.T) {
 	})
 }
 
+func TestBandwidthRollup(t *testing.T) {
+	ctx := testcontext.New(t)
+	defer ctx.Cleanup()
+
+	log := zaptest.NewLogger(t)
+
+	db, err := storagenodedb.NewTest(log, ctx.Dir("storage"))
```
egonelbre: `storagenodedbtest.Run`

storagenode/storagenodedb/bandwidthdb.go
egonelbre: Based on the code I'm not able to understand that it handles it correctly. Let's say I put data older than 1 year into bandwidth_usage table. The rollup won't touch that, since the `created_at` never matches the query. Then the `DELETE FROM` part will delete it.

So this suggest to me the tests have a bug somewhere.

storagenode/storagenodedb/bandwidthdb.go
coyle: so your concern is data from the past is added and then not rolled up? I think that is the expectation of rolling up, otherwise why roll the data?

egonelbre: Yes, data from past is added and then discarded without rolling them up.

storagenode/storagenodedb/bandwidthdb.go
ethanadams: First time the rollup occurs it rolls up all data up until the last hour.  Correct, all subsequent rollups will ignore data that was added not with the time range.  But that would suggest some process added more bw usage over an hour after the upload occurred.  Is that possible? BW usage is added after each piece is uploaded.

egonelbre: From the failure we got, we can assume that it already happened.

storagenode/storagenodedb/bandwidthdb.go
coyle: correct I think that's intentional with this system

storagenode/storagenodedb/bandwidthdb.go
egonelbre: Also since we have turned off rollups for a while there are many entries that need to be rolled up.

ethanadams: From the 2 databases I looked at that had this issue, the data was rolled up, but not deleted.

I verified by running a query and the amounts in the BW usage table, matched the rollup values for that hour.

If we had code that did an insert/on conflict update, the rollup would have been double the actual amount used, in this situation.

Maybe a solution would be to put a date range on the delete so we don't delete the data....It just gets left behind.  Then we can observe the frequency of the phenomenon. However, we could still be reporting bandwidth incorrectly.

egonelbre: Why wasn't it deleted?

ethanadams: Unfortunately I don't have enough info to answer that.  How about we add the date range to the delete so we don't delete the data that causes the constraint issue?

Then we can use that data to help us debug the issue and frequency of the problem.

storagenode/bandwidth/db_test.go
ethanadams: updated

storagenode/bandwidth/db_test.go
ethanadams: updated

storagenode/storagenodedb/bandwidthdb.go
egonelbre: Let's try to recreate the situation before trying to fix it.

I just started wondering, what would happen if we set the rollup interval to every 5s.

storagenode/bandwidth/service.go
```diff
@@ -0,0 +1,45 @@
...
+	db   DB
+	Loop sync2.Cycle
+}
+
+// NewService creates a new bandwidth service.
+func NewService(log *zap.Logger, db DB) *Service {
+	return &Service{
+		log:  log,
+		db:   db,
+		Loop: *sync2.NewCycle(time.Hour * 1),
```
egonelbre: This should be configurable.

storagenode/storagenodedb/bandwidthdb.go
ReneSmeekes: I can confirm that the database I saw had the same amounts in bandwidth_usage as in bandwidth_usage_rollup for the duplicate keys. This suggests the process got interrupted between the insert and the delete statement. You could consider wrapping the two into a transaction to prevent this from happening.

egonelbre: Yes, both are possible causes, and both need fixes.

ethanadams: An alternative:
```
	tx, err := db.Begin()
	if err != nil {
		return ErrInfo.Wrap(err)
	}

	defer func() {
		if err == nil {
			err = tx.Commit()
		} else {
			err = errs.Combine(err, tx.Rollback())
		}
	}()

	result, err := tx.Exec(`
		INSERT INTO bandwidth_usage_rollups (interval_start, satellite_id,  action, amount)
		SELECT datetime(strftime('%Y-%m-%dT%H:00:00', created_at)) created_hr, satellite_id, action, SUM(amount)
			FROM bandwidth_usage
		WHERE datetime(created_at) < datetime(?)
		GROUP BY created_hr, satellite_id, action
		ON CONFLICT(interval_start, satellite_id,  action)
		DO UPDATE SET amount = bandwidth_usage_rollups.amount + excluded.amount;

		DELETE FROM bandwidth_usage WHERE datetime(created_at) < datetime(?);
	`, hour, hour)
```
This uses a transaction and also allows for any straggling data after the rollup.

egonelbre: You can include the BEGIN and COMMIT in the exec query.

ethanadams: updated. leaving tx code as is, per our conversation

storagenode/storagenodedb/bandwidthdb.go
ethanadams: removed in favor of new insert/on conflict/transaction code

storagenode/storagenodedb/bandwidthdb.go
ethanadams: see above

storagenode/bandwidth/db_test.go
ethanadams: updated

storagenode/bandwidth/service.go
ethanadams: updated

storagenode/bandwidth/service.go
```diff
@@ -0,0 +1,62 @@
...
+	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+
+	"storj.io/storj/internal/sync2"
+)
+
+var mon = monkit.Package()
+
+// Config defines parameters for storage node Collector.
+type Config struct {
+	Interval time.Duration `help:"how frequently bandwidth usage rollups are calculated" default:"1h0m0s"`
```
aligeti: just to be consistent add "devDefault:"xxx" releaseDefault:":xxx""

aligeti: example take a look at /pkg/accounting/rollup/rollup.go line 19

PR #2618, opened 2019-07-22
update testplanet with libuplink

JessicaGreben:
What: 
Update testplanet's uplink code to use libuplink.

Why:
So that our testplanet tests are using the same code as production.

Please describe the tests:
- updated any broken tests
 
n/a


internal/testplanet/uplink.go
```diff
@@ -342,75 +332,94 @@ func atLeastOne(value int) int {
...
-		return nil, nil, cleanup, err
+// NewLibuplink creates a libuplink.Uplink object with the testplanet Uplink config
+func (planetUplink *Uplink) NewLibuplink(ctx context.Context) (*libuplink.Uplink, error) {
+	config := getDefaultConfig()
+	libuplinkCfg := &libuplink.Config{}
+	libuplinkCfg.Volatile.MaxInlineSize = config.Client.MaxInlineSize
+	libuplinkCfg.Volatile.MaxMemory = config.RS.MaxBufferMem
+	libuplinkCfg.Volatile.PeerIDVersion = config.TLS.PeerIDVersions
+	libuplinkCfg.Volatile.TLS = struct {
+		SkipPeerCAWhitelist bool
```
isaachess: This is a nitpick, but this pattern seems less clear to me than simply continuing what you're doing above, such as:

```
libuplinkCfg.Volatile.TLS.SkipPeerCAWhitelist = !config.TLS.UsePeerCAWhitelist
libuplinkCfg.Volatile.TLS.PeerCAWhitelistPath = config.TLS.PeerCAWhitelistPath
```

internal/testplanet/uplink.go
```diff
@@ -198,128 +192,124 @@ func (uplink *Uplink) UploadWithExpirationAndConfig(ctx context.Context, satelli
...
 
-	metainfo, streams, cleanup, err := DialMetainfo(ctx, uplink.Log.Named("metainfo"), config, uplink.Identity)
+	project, bucket, err := planetUplink.GetProjectAndBucket(ctx, satellite, bucketName, config)
 	if err != nil {
 		return err
 	}
-	defer func() {
-		err = errs.Combine(err, cleanup())
-	}()
+	defer func() { err = errs.Combine(err, project.Close(), bucket.Close()) }()
```
mniewrzal: nit: not sure if that will make a difference but currently project will be closed before bucket and we should close things in reverse open order.

internal/testplanet/uplink.go
```diff
@@ -342,75 +332,94 @@ func atLeastOne(value int) int {
...
-			// close metainfo if any of the setup fails
-			err = errs.Combine(err, m.Close())
-		}
-	}()
 
-	project, err := kvmetainfo.SetupProject(m)
+	project, err := testLibuplink.OpenProject(ctx, satellite.Addr(), key)
 	if err != nil {
-		return nil, nil, cleanup, errs.New("failed to create project: %v", err)
+		return nil, errs.Combine(err, testLibuplink.Close())
```
mniewrzal: `testLibuplink.Close()` will be also closed in defer

internal/testplanet/uplink.go
```diff
@@ -342,75 +332,94 @@ func atLeastOne(value int) int {
...
 	if err != nil {
-		return nil, nil, cleanup, errs.New("failed to create erasure coding client: %v", err)
+		return nil, nil, err
 	}
-	rs, err := eestream.NewRedundancyStrategy(eestream.NewRSScheme(fc, config.RS.ErasureShareSize.Int()), config.RS.RepairThreshold, config.RS.SuccessThreshold)
+
+	access, err := setup.LoadEncryptionAccess(ctx, planetUplinkCfg.Enc)
 	if err != nil {
-		return nil, nil, cleanup, errs.New("failed to create redundancy strategy: %v", err)
+		return nil, nil, errs.Combine(err, project.Close())
```
mniewrzal: maybe use defer to close `project` instead of manual closing with each error?

satellite/metainfo/metainfo_test.go
```diff
@@ -233,17 +233,15 @@ func TestServiceList(t *testing.T) {
...
-	defer ctx.Check(cleanup)
-
 	type Test struct {
 		Request  storj.ListOptions
 		Expected storj.ObjectList // objects are partial
 	}
 
-	list, err := metainfo.ListObjects(ctx, "testbucket", storj.ListOptions{Recursive: true, Direction: storj.After})
+	config := planet.Uplinks[0].GetConfig(planet.Satellites[0])
+	_, bucket, err := planet.Uplinks[0].GetProjectAndBucket(ctx, planet.Satellites[0], "testbucket", config)
```
mniewrzal: shouldn't `project` and `bucket` be closed somewhere?

internal/testplanet/uplink.go
```diff
@@ -140,46 +134,46 @@ func (planet *Planet) newUplink(name string, storageNodeCount int) (*Uplink, err
 }
 
 // ID returns uplink id
-func (uplink *Uplink) ID() storj.NodeID { return uplink.Info.Id }
+func (planetUplink *Uplink) ID() storj.NodeID { return planetUplink.Info.Id }
```
egonelbre: This receiver name is worse than it was before.

JessicaGreben: we have too many uplinks. i was trying to find a way to differentiate between testplanet Uplink vs libuplink vs uplink pkg. this is what i came up with.  do you have any better suggestions?

internal/testplanet/uplink.go
JessicaGreben: good idea, i will fix

internal/testplanet/uplink.go
egonelbre: best I can come up with: `client`, `user`, `peer`

mniewrzal: I feel ok with `client` or `peer` :)

JessicaGreben: jusst `client` or `uplinkClient`?

internal/testplanet/uplink.go
JessicaGreben: sounds good. yeah i really dislike instantiating this object like this. its still bad, but your way is prob a tad better. will update

internal/testplanet/uplink.go
egonelbre: just `client`

internal/testplanet/uplink.go
JessicaGreben: i only want to close if there is an error, otherwise don't close.

mniewrzal: add named `error` result and then you can do just this:
```
	defer func() {
		if err != nil {
			err = errs.Combine(err, project.Close())
		}
	}()
```

PR #2620, opened 2019-07-22
Implement checker observer

mobyvb:
What: 
Integrate checker into metainfo loop

Why:
The metainfo loop got merged, so the checker should implement it to reduce concurrent iterations over metainfo in the future.

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/datarepair/checker/checker_test.go
```diff
@@ -205,106 +202,3 @@ func makePointer(t *testing.T, planet *testplanet.Planet, pieceID string, create
 	err := pointerdb.Put(ctx, pieceID, pointer)
 	require.NoError(t, err)
 }
-
-func TestCheckerResume(t *testing.T) {
```
aligeti: curious why this resume test case is removed?

pkg/datarepair/checker/checker.go
```diff
@@ -56,11 +54,11 @@ type Checker struct {
 	logger          *zap.Logger
 	Loop            sync2.Cycle
 	IrreparableLoop sync2.Cycle
-	monStats        durabilityStats
+	metaLoop        *metainfo.Loop
```
aligeti: I like the monStats is not stored. Didn't feel like we needed to store. :

satellite/repair/checker/checker.go
```diff
@@ -71,7 +69,7 @@ func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overl
 		logger:          logger,
 		Loop:            *sync2.NewCycle(config.Interval),
 		IrreparableLoop: *sync2.NewCycle(config.IrreparableInterval),
-		monStats:        durabilityStats{},
```
jenlij: We don't need monStats any longer?

satellite/repair/checker/checker.go
```diff
@@ -107,53 +105,24 @@ func (checker *Checker) Close() error {
...
+		repairQueue: checker.repairQueue,
+		irrdb:       checker.irrdb,
+		nodestate:   checker.nodestate,
+		monStats:    durabilityStats{},
+		log:         checker.logger,
+	}
+	err = checker.metaLoop.Join(ctx, observer)
 	if err != nil {
-		return err
+		return Error.New("IdentifyInjuredSegments err %v", err)
```
ifraixedes: For now, there is no way to conserve the `err` stack adding some additional message unless we use `Combine`, however, I have the feeling that `Combine` is not for this kind of things, so we should log `err` before returning.

Assess if `Combine` is fine, if you think that it isn't appropriated then log `err`

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
 		// we need one additional piece for error correction. If only the minimum is remaining the file can't be repaired and is lost.
 		// except for the case when minimum and repair thresholds are the same (a case usually seen during testing)
 	} else if numHealthy <= redundancy.MinReq && numHealthy < redundancy.RepairThreshold {
+
+		// make an entry in to the irreparable table
```
ifraixedes: `in to` => `into` ??

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+			SegmentDetail:      pointer,
+			LostPieces:         int32(len(missingPieces)),
+			LastRepairAttempt:  time.Now().Unix(),
+			RepairAttemptCount: int64(1),
+		}
+
+		// add the entry if new or update attempt count if already exists
+		err := checker.irrdb.IncrementRepairAttempts(ctx, segmentInfo)
+		if err != nil {
+			return Error.New("error handling irreparable segment to queue %s", err)
```
ifraixedes: Same problem here with the `err` stack, use `Combine` or log `err`

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+
+	pieces := remote.GetRemotePieces()
+	if pieces == nil {
+		obs.log.Debug("no pieces on remote segment")
+		return nil
+	}
+
+	missingPieces, err := obs.nodestate.MissingPieces(ctx, pointer.CreationDate, pieces)
+	if err != nil {
+		return Error.New("error getting missing pieces %s", err)
```
ifraixedes: Same problem here than before, use `Combine` or log `err`

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+	}
+
+	numHealthy := int32(len(pieces) - len(missingPieces))
+	redundancy := pointer.Remote.Redundancy
```
ifraixedes: move this declaration above where it's the variable first used.

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+	mon.IntVal("checker_segment_total_count").Observe(int64(len(pieces)))
+	mon.IntVal("checker_segment_healthy_count").Observe(int64(numHealthy))
+
+	// we repair when the number of healthy pieces is less than or equal to the repair threshold
+	// except for the case when the repair and success thresholds are the same (a case usually seen during testing)
+	if numHealthy > redundancy.MinReq && numHealthy <= redundancy.RepairThreshold && numHealthy < redundancy.SuccessThreshold {
+		if len(missingPieces) == 0 {
+			obs.log.Warn("Missing pieces is zero in checker, but this should be impossible -- bad redundancy scheme.")
```
ifraixedes: Why not report an error rather than a warning?
In any case, it would be good if we could add the redundancy of this segment and any other information to identify it.

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+			return nil
+		}
+		obs.monStats.remoteSegmentsNeedingRepair++
+		err = obs.repairQueue.Insert(ctx, &pb.InjuredSegment{
+			Path:         []byte(path),
+			LostPieces:   missingPieces,
+			InsertedTime: time.Now().UTC(),
+		})
+		if err != nil {
+			obs.log.Sugar().Errorf("error adding injured segment to queue %s", err)
```
ifraixedes: Change this line to be `obs.log.Error("error adding injured segment to queue", zap.Error(err))`

satellite/repair/checker/checker.go
```diff
@@ -221,18 +182,97 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
+			return nil
+		}
+
+		// delete always returns nil when something was deleted and also when element didn't exists
+		err = obs.irrdb.Delete(ctx, []byte(path))
+		if err != nil {
+			obs.log.Sugar().Errorf("error deleting entry from irreparable db: ", zap.Error(err))
```
ifraixedes: There is no need to use `Sugar` here, you can do the same with `obs.log.Error` and it's mostly what we want.

satellite/repair/checker/checker.go
```diff
@@ -243,11 +283,26 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
 		}
 
 		// add the entry if new or update attempt count if already exists
-		err := checker.irrdb.IncrementRepairAttempts(ctx, segmentInfo)
+		err := obs.irrdb.IncrementRepairAttempts(ctx, segmentInfo)
 		if err != nil {
-			return Error.New("error handling irreparable segment to queue %s", err)
+			obs.log.Sugar().Errorf("error handling irreparable segment to queue %s", err)
```
ifraixedes: Change this line to be `obs.log.Error("error handling irreparable segment to queue", zap.Error(err))`

satellite/repair/checker/checker.go
```diff
@@ -243,11 +283,26 @@ func (checker *Checker) updateSegmentStatus(ctx context.Context, pointer *pb.Poi
...
 		}
 	}
+
+	return nil
+}
+
+func (obs *checkerObserver) RemoteObject(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	obs.monStats.remoteFilesChecked++
```
ifraixedes: How this metric works.

Some segments can be an object but not all of them, so if we repair them and they are objects then this method is called and we increment the metric, is that correct?

satellite/repair/checker/checker.go
```diff
@@ -56,11 +54,11 @@ type Checker struct {
 	logger          *zap.Logger
 	Loop            sync2.Cycle
 	IrreparableLoop sync2.Cycle
-	monStats        durabilityStats
+	metaLoop        *metainfo.Loop
 }
 
 // NewChecker creates a new instance of checker
-func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overlay *overlay.Cache, irrdb irreparable.DB, limit int, logger *zap.Logger, config Config) *Checker {
+func NewChecker(metainfo *metainfo.Service, repairQueue queue.RepairQueue, overlay *overlay.Cache, irrdb irreparable.DB, limit int, metaLoop *metainfo.Loop, logger *zap.Logger, config Config) *Checker {
```
ifraixedes: Because you have changed the signature of this function is a good chance to align it with our _argument order_ conventions https://github.com/storj/docs/blob/master/code/Style.md#argument-order

satellite/repair/checker/checker.go
ifraixedes: On the other hand, I'm not entirely sure that if there is an error we want to terminate these process because of the loop in https://github.com/storj/storj/pull/2620/files#diff-dc8ddcff2c5de4611c4f5d3eda6fd766L85 will terminate and the process will only run once the satellite is tear down an started again.

I'm not saying that this is a problem just here, it's also happening in `checker.IrreparableProcess` and likely in other processes that the satellite run in `satellite/peer.Run`.

We could open a ticket to find out where this is happening and adopt a strategy, for example:

1. The person on call get paged and it has to restart the satellite as soon as possible in order of having again these process running.
2. We don't stop the service but we report, the person on call is paged, then she/he can check the errors and decide if to restart or keep it in observation to see if the same error is happening again in the short time and make the decision of restarting.

satellite/metainfo/loop.go
```diff
@@ -148,8 +148,6 @@ waitformore:
 			observers = append(observers, observer)
 		case <-timer.C:
 			break waitformore
-		case <-ctx.Done():
```
ifraixedes: Why? what was the problem with it?

egonelbre: Yeah, this MUST be here.

satellite/repair/checker/checker.go
mobyvb: Yes, if a remote segment is an object, it will be passed into both `RemoteObject` and `RemoteSegment`. Since we already do the checker things in `RemoteSegment`, we just keep track of one of the metrics to send datascience in `RemoteObject`. Before we were trying to calculate it ourselves from this file, but now it can be done more easily with the metainfo loop.

satellite/metainfo/loop.go
mobyvb: I am not sure why, but this was causing a lot of test failures. I plan to put it back in and try to figure out why after I get the tests to pass.

satellite/peer.go
```diff
@@ -670,10 +673,10 @@ func (peer *Peer) Run(ctx context.Context) (err error) {
 		return errs2.IgnoreCanceled(peer.Discovery.Service.Run(ctx))
 	})
 	group.Go(func() error {
-		return errs2.IgnoreCanceled(peer.Repair.Checker.Run(ctx))
+		return errs2.IgnoreCanceled(peer.Metainfo.Loop.Run(ctx))
```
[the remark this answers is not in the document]
mobyvb: we want to start the checker after the metainfo loop has been started, otherwise there will be no loop for the checker to join.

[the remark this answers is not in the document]
ifraixedes: This change doesn't guarantee that, because of `group.Go` spins up a new goroutine and it's kind indeterministic the execution order of goroutines.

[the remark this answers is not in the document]
mobyvb: Ok I'll switch it back since it doesn't seem like the order is super important

pkg/datarepair/checker/checker_test.go
mobyvb: the metainfo iteration part of the checker loop is no longer something that needs to be tested, since that iteration has been made into a separate service

satellite/repair/checker/checker.go
mobyvb: I don't think the additional messaging is important enough that we need to wrap it. Also, this error can't be wrapped anymore because it was causing a context canceled error to fail a lot of tests, since the satellite would not end nicely.

Regarding your other point, it seems to me like most services on the satellite will return an error that will crash the satellite, and it will need to be restarted by whoever is on call. I like the idea of reporting the issue when it occurs, but not stopping the service by returning an error, I just don't know how we would do that right now. And if we just log the error out, we might not notice it, especially if it does not happen often.

satellite/repair/checker/checker.go
mobyvb: I will use Combine here

satellite/repair/checker/checker.go
mobyvb: I will use Combine here

satellite/repair/checker/checker.go
ifraixedes: > I don't think the additional messaging is important enough that we need to wrap it. Also, this error can't be wrapped anymore because it was causing a context canceled error to fail a lot of tests, since the satellite would not end nicely.

I don't get how that can fails, the original error is lost so if it's `context canceled` there is no possibility to check it.
No saying that it has to be wrapped, mostly I'm trying to understand the problem.

By the way, before I missed, that it could be `Error.Wrap(err)`, is that also failing?

> Regarding your other point, it seems to me like most services on the satellite will return an error that will crash the satellite, and it will need to be restarted by whoever is on call.

Yes, whatever option I've described above imply so.

> I like the idea of reporting the issue when it occurs,

That's the thing, if I'm not mistaken, right now it won't report it until the satellite finish due to the `errgroup.Wait()`, so we won't notice it.

> but not stopping the service by returning an error,

I haven't suggested that; it could be in that way if we would run more than one instance of the same satellite, but that isn't the case right now.

> I just don't know how we would do that right now. And if we just log the error out, we might not notice it, especially if it does not happen often.

We will notice less if there is nothing in the log reporting it (per my above comment).

We can return the error and stop the process as it's happening now but before returning it log the error.

mobyvb: What I meant is that when we do `Error.New....` instead of just returning `err`, the satellite would not close nicely because of this in `satellite/peer.go`:
```
group.Go(func() error {
	return errs2.IgnoreCanceled(peer.Repair.Checker.Run(ctx))
})
```
The context canceled error is not properly ignored when we return `Error.New(...` which is why I just returned `err` directly.

mobyvb: I can log out the errors before we return them, but I do not know when we should return, or log and then return.

ifraixedes: > The context canceled error is not properly ignored when we return Error.New(... which is why I just returned err directly.

Oh man! I was checking the changes where I found this conversation in "files changed" but it turns that they weren't the last changes so I was reading the same than before and I thought that you didn't do any change. I had to manually go to the line to find the new change.

It makes sense now what you were saying.

>  can log out the errors before we return them, but I do not know when we should return, or log and then return.

Currently, the easiest is logging it before returning it, but we have not to log it if it's context canceled using [`errs2.IsCanceled`](https://godoc.org/storj.io/storj/internal/errs2#IsCanceled)

mobyvb: I checked if `!errs2.IsCanceled` and logged if it is not canceled. I plan to merge shortly unless anyone brings up other concerns

PR #2622, opened 2019-07-23
satellite/satellitedb: Use var block for single variable declaration

ifraixedes:
What: Use var block and remove some spaces at the end of lines.

Why: To have more consistent godoc format

Before this change the godoc format is rendered as

2019-07-23-135743_813x252_scrot

While in other packages we have used blocks for a single var declaration at package level. For example

2019-07-23-135804_1090x345_scrot

I consider that when there is more than block of error variables, it's easier to read that all the different variables declared in different files of the same package be defined in blocs.

### DIFF
--- a/satellite/satellitedb/migrate.go
+++ b/satellite/satellitedb/migrate.go
@@ -18,8 +18,10 @@ import (
 	"storj.io/storj/satellite/satellitedb/pbold"
 )
 
-// ErrMigrate is for tracking migration errors
-var ErrMigrate = errs.Class("migrate")
+var (
+	// ErrMigrate is for tracking migration errors
+	ErrMigrate = errs.Class("migrate")
+)
 
 // CreateTables is a method for creating all tables for database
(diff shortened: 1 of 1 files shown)


satellite/satellitedb/migrate.go
```diff
@@ -18,8 +18,10 @@ import (
 	"storj.io/storj/satellite/satellitedb/pbold"
 )
 
-// ErrMigrate is for tracking migration errors
-var ErrMigrate = errs.Class("migrate")
+var (
```
wthorp: I will always disagree with more lines of code

ifraixedes: It's about the godoc otherwise I would never have found it, not changed it

wthorp: ah, thanks for teaching me that.

ifraixedes: you can see the difference in the PR comment

PR #2623, opened 2019-07-23
[v3 2137] - Add more info to find out repair failures

ifraixedes:
What:  Add more information to the reported errors to find out how/where the repair process fails

Why: Because we are having troubles with repair and we want to find which is the part which causes the most of them

Please describe the tests: There is not change in the logic rather than, in majority, improving the error messages or wrapping errors, hence the existing tests are enough.
 
Please describe the performance impact: probably a bit, however we prefer to have more contextualized errors rather than having the minor improvement in performance.

### DIFF
--- a/pkg/storage/ec/client.go
+++ b/pkg/storage/ec/client.go
@@ -343,10 +343,15 @@ func (ec *ecClient) Get(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 
 	rr, err = eestream.Decode(rrs, es, ec.memoryLimit, ec.forceErrorDetection)
 	if err != nil {
-		return nil, err
+		return nil, Error.Wrap(err)
 	}
 
-	return eestream.Unpad(rr, int(paddedSize-size))
+	ranger, err := eestream.Unpad(rr, int(paddedSize-size))
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	return ranger, err
 }
 
 func (ec *ecClient) Delete(ctx context.Context, limits []*pb.AddressedOrderLimit, privateKey storj.PiecePrivateKey) (err error) {
--- a/pkg/storage/ec/client.go
+++ b/pkg/storage/ec/client.go
@@ -347,11 +347,7 @@ func (ec *ecClient) Get(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 	}
 
 	ranger, err := eestream.Unpad(rr, int(paddedSize-size))
-	if err != nil {
-		return nil, Error.Wrap(err)
-	}
-
-	return ranger, err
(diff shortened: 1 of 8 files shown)


ifraixedes:
After the inspection of the code and with the minimal instructions that we have in the [ticket](https://storjlabs.atlassian.net/browse/V3-2137) that these changes should close, I don't see what else, except of abusing more of monkit tracing we can add, so please have a look and provide me feedback on what else should be added before this PR is moved out from draft.

pkg/storage/ec/client.go
```diff
@@ -343,10 +343,15 @@ func (ec *ecClient) Get(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 
 	rr, err = eestream.Decode(rrs, es, ec.memoryLimit, ec.forceErrorDetection)
 	if err != nil {
-		return nil, err
+		return nil, Error.Wrap(err)
+	}
+
+	ranger, err := eestream.Unpad(rr, int(paddedSize-size))
+	if err != nil {
```
wthorp: Nit: Error.Wrap returns nil if the input is nil.  I don't think you need this 'if' check.

satellite/metainfo/config.go
```diff
@@ -57,6 +57,10 @@ func NewStore(logger *zap.Logger, dbURLString string) (db storage.KeyValueStore,
 		err = Error.New("unsupported db scheme: %s", driver)
 	}
 
+	if err != nil {
```
wthorp: Again, this check is probably just extra code / against Go style guidelines.

storage/postgreskv/client.go
```diff
@@ -89,12 +89,14 @@ func (client *Client) GetPath(ctx context.Context, bucket, key storage.Key) (_ s
 	row := client.pgConn.QueryRow(q, []byte(bucket), []byte(key))
 	var val []byte
 	err = row.Scan(&val)
-	if err == sql.ErrNoRows {
-		return nil, storage.ErrKeyNotFound.New(key.String())
-	}
 	if err != nil {
```
wthorp: Nit: Again, this check is probably just extra code / against Go style guidelines.

wthorp: Nit: Again, this check is probably just extra code / against Go style guidelines.

pkg/storage/ec/client.go
ifraixedes: I do not like to abuse of it, but I think that's clear when nil is returned when no error than having to know that `Wrap` returns nil.

However I will change them

satellite/metainfo/config.go
ifraixedes: This one it's needed as the `log.Debug` should only happen if there is no error.

PR #2624, opened 2019-07-23
Improve Node Selection Query

stefanbenten:
What: Remove unnecessary UNION and Selects

Why: Improves Speed
 
Please describe the performance impact: less queries, thus better throughput


stefanbenten:
@BlackDuck888 There you go 👍

stefanbenten:
Query Plan:
image

BlackDuck888:
Let this one open, I step in tomorrow. I thought futher, to reduce one step more.

stefanbenten:
@BlackDuck888 I'll change that now.

stefanbenten:
@BlackDuck888 We cannot further reduce the query, if we merge the last two Select Statements, we are squashing/removing the randomness of nodes within the last_net segment

egonelbre:
It also would be nice to have a benchmark for the improvement.

egonelbre:
@thepaul I think that is fine, since during initial connect, discovery or refresh we should get an IP... so it should rarely happen that a node doesn't have `last_net`.

stefanbenten:
@thepaul I agree with Egon, we definitely have the last_net with any connection now. The only exception are nodes that are on versions before we implemented that change, but they wont be selected anyway.

thepaul:
Oh ok! I missed that we had made that change. Yes, this is great then.

stefanbenten:
As discussed with @BlackDuck888 , we will merge this for now and he opens a nother PR to further improve it!

PR #2625, opened 2019-07-23
make sure to return after put()

thepaul:
What: 

so that we don't get down into the wrong logic when the put is
successful, as pointed out by jeff

Why:

cause I missed seeing this during the review of #2602 

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storage/redis/client.go
+++ b/storage/redis/client.go
@@ -235,7 +235,7 @@ func (client *Client) CompareAndSwap(ctx context.Context, key storage.Key, oldVa
 			_, err = tx.Pipelined(func(pipe redis.Pipeliner) error {
 				return put(ctx, pipe, key, newValue, client.TTL)
 			})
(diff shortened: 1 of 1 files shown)


storage/redis/client.go
```diff
@@ -235,6 +235,7 @@ func (client *Client) CompareAndSwap(ctx context.Context, key storage.Key, oldVa
 			_, err = tx.Pipelined(func(pipe redis.Pipeliner) error {
 				return put(ctx, pipe, key, newValue, client.TTL)
 			})
+			return Error.Wrap(err)
```
kaloyan-raev: We should not wrap the error here, as wrapping makes checking for `redis.TxFailedErr` unreliable.

kaloyan-raev: I pushed the change. This should fix the tests.

PR #2626, opened 2019-07-24
Add ability for satellite to discover antechamber nodes

cam-a:
https://storjlabs.atlassian.net/browse/V3-1729
What: Modify the kademlia `FindNear` method to return antechamber nodes as well as routing table nodes. This PR also modifies most of the functions along the kademlia lookup code path to be able to pass along an optional maximum number of antechamber nodes starting from the call to `FindNodeWithAntechamber` in `discovery.discover` and terminating in the modified `FindNear` in `endpoint.Query`

Why: this will allow antechamber nodes to be found during the satellite's node discovery

*** TODO: TESTS! ***
Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/discovery/service.go
```diff
@@ -56,6 +58,7 @@ func New(logger *zap.Logger, ol *overlay.Cache, kad *kademlia.Kademlia, config C
 
 		refreshLimit:       config.RefreshLimit,
 		refreshConcurrency: config.RefreshConcurrency,
+		antechamberLimit:   config.AntechamberLimit,
```
jenlij: why not have this as part of the request, like the original findNear limit?

pkg/kademlia/inspector.go
```diff
@@ -64,7 +64,7 @@ func (srv *Inspector) FindNear(ctx context.Context, req *pb.FindNearRequest) (_
 	defer mon.Task()(&ctx)(&err)
 	start := req.Start
 	limit := req.Limit
-	nodes, err := srv.kademlia.FindNear(ctx, start, int(limit))
+	nodes, err := srv.kademlia.FindNear(ctx, start, int(limit), 0)
```
jenlij: Maybe we should have a flag to view the antechamber nodes via the inspector?

pkg/kademlia/kademlia.go
```diff
@@ -109,11 +109,19 @@ func (k *Kademlia) Queried() {
 	k.lastQueried = time.Now()
 }
 
-// FindNear returns all nodes from a starting node up to a maximum limit
-// stored in the local routing table.
-func (k *Kademlia) FindNear(ctx context.Context, start storj.NodeID, limit int) (_ []*pb.Node, err error) {
+// FindNear returns a limited number of XOR closest nodes from the routing table
+// and the antechmaber respectively.
+func (k *Kademlia) FindNear(ctx context.Context, start storj.NodeID, routingTableLimit int, antechamberLimit int) (_ []*pb.Node, err error) {
```
jenlij: I think we actually want to return the RT nodes and antechamber nodes in separate structs

pkg/discovery/service.go
cam-a: Not sure I understand what you mean

cam-a: It's part of the request here
https://github.com/storj/storj/pull/2626/files#diff-288a6a2c91f0ea201adc03ad9dbeb188R65

pkg/kademlia/kademlia.go
```diff
@@ -275,16 +283,23 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (_ *pb.InfoRespo
 	return info, nil
 }
 
-// FindNode looks up the provided NodeID first in the local Node, and if it is not found
-// begins searching the network for the NodeID. Returns and error if node was not found
+// FindNode wraps a call to FindNodeWithAntechamber with 0 antechamber nodes
```
jenlij: why are we just wrapping FindNodeWithAntechamber?

pkg/discovery/service.go
jenlij: Ok never mind

pkg/kademlia/kademlia.go
cam-a: Every instance of FindNode, we never want to be given any antechamber nodes. I could have just added the `antechamberLimit` argument to FindNode, but it seemed weird to me to have to specify an argument that we know we don't want except for one place: node discovery. With FindNode as a wrapper, this allows us to leave all the existing FindNode calls alone

PR #2632, opened 2019-07-25
V3-1987: Optimize audits stats persistence

ethanadams:
What: 
Update reporter and overlay to update audit stats in batched transactions.

Why:
Traces show significant time being spend looping through individual stats updates.

https://storjlabs.atlassian.net/browse/V3-1987

Simple benchmark tests show a significant improvement between single update/transaction model vs batching within a transaction.
```
##
## sqlite (file)
##

# 100 updates
BenchmarkBulk/100_updates_and_commits         	     300	   4066550 ns/op	  125079 B/op	    4100 allocs/op
BenchmarkBulk/100_updates_1_commit           	    2000	    760509 ns/op	   45066 B/op	    1527 allocs/op
BenchmarkBulk/100_in_one_db_call             	    2000	    652836 ns/op	  403846 B/op	    1329 allocs/op


# 500 updates
BenchmarkBulk/500_updates_and_commits         	      50	  20456170 ns/op	  627740 B/op	   20516 allocs/op
BenchmarkBulk/500_updates_1_commit            	     500	   3689338 ns/op	  221219 B/op	    7527 allocs/op
BenchmarkBulk/500_in_one_db_call              	     300	   5042053 ns/op	 9825272 B/op	    6546 allocs/op

##
## postgres (localhost)
##

# 100 updates
BenchmarkBulk/100_updates_and_commits         	      30	  41845984 ns/op	   92011 B/op	    1907 allocs/op
BenchmarkBulk/100_updates_1_commit            	     100	  13995990 ns/op	   38376 B/op	    1010 allocs/op
BenchmarkBulk/100_in_one_db_call              	     200	   8571144 ns/op	  210926 B/op	     616 allocs/op

# 500 updates
BenchmarkBulk/500_updates_and_commits         	       5	 211364240 ns/op	  455345 B/op	    9516 allocs/op
BenchmarkBulk/500_updates_1_commit            	      20	  71476219 ns/op	  188978 B/op	    5011 allocs/op
BenchmarkBulk/500_in_one_db_call              	      30	  45730926 ns/op	 4931225 B/op	    3030 allocs/op
```

Still considering the single SQL call, but that will require a lot of business logic in the SQL, which would duplicate logic in Go code.

Please describe the tests:
 - Test 1: Updated existing tests to use BatchUpdateStats
 - Test 2:
 

pkg/audit/reporter.go
```diff
@@ -150,26 +150,26 @@ func (reporter *Reporter) recordOfflineStatus(ctx context.Context, offlineNodeID
...
 
 // recordAuditSuccessStatus updates nodeIDs in overlay with isup=true, auditsuccess=true
 func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successNodeIDs storj.NodeIDList) (failed storj.NodeIDList, err error) {
 	defer mon.Task()(&ctx)(&err)
-	var errlist errs.Group
-	for _, nodeID := range successNodeIDs {
-		_, err := reporter.overlay.UpdateStats(ctx, &overlay.UpdateRequest{
+
+	updateRequests := make([]*overlay.UpdateRequest, len(successNodeIDs))
+	for i, nodeID := range successNodeIDs {
```
egonelbre: Is there a reason we cannot do just single query for the whole batch?

egonelbre: Most of these updates should be possible to condense into a single update.

satellite/satellitedb/overlaycache.go
```diff
@@ -671,23 +671,84 @@ func (cache *overlaycache) UpdateAddress(ctx context.Context, info *pb.Node, def
...
+	nodeStats, err := updateStats(ctx, updateReq, tx)
+	if err != nil {
+		return nodeStats, errs.Combine(err, tx.Rollback())
+	}
+
+	return nodeStats, errs.Combine(err, tx.Commit())
+}
+
+// UpdateStats a single storagenode's stats in the db
+func updateStats(ctx context.Context, updateReq *overlay.UpdateRequest, tx *dbx.Tx) (stats *overlay.NodeStats, err error) {
```
egonelbre: I wonder whether this would end up problematic with multiple servers updating at the same time.

pkg/audit/reporter.go
ethanadams: Yes, I agree.  I was looking for opinions as to whether I should put all the business logic in SQL vs using the existing Go logic.  

I will work on the SQL model next.

egonelbre: You shouldn't need a SQL model either. An add/subtract with set to DQ should do it.

ethanadams: updated to execute updates/deletes in on call to DB

satellite/audit/reporter.go
```diff
@@ -118,21 +118,21 @@ func (reporter *Reporter) RecordAudits(ctx context.Context, req *Report) (_ *Rep
...
-		if err != nil {
-			failed = append(failed, nodeID)
-			errlist.Add(err)
 		}
 	}
-	if len(failed) > 0 {
-		return failed, errs.Combine(Error.New("failed to record some audit fail statuses in overlay"), errlist.Err())
+
+	failed, err = reporter.overlay.BatchUpdateStats(ctx, updateRequests)
+	if err != nil && len(failed) > 0 {
```
coyle: should this be an `or` not an `and` ?

satellite/audit/reporter.go
```diff
@@ -187,19 +189,29 @@ func (reporter *Reporter) recordPendingAudits(ctx context.Context, pendingAudits
...
+		}
+	}
+
+	failedBatch, err := reporter.overlay.BatchUpdateStats(ctx, updateRequests)
+	if err != nil {
+		errlist.Add(err)
+	}
+	for _, nodeID := range failedBatch {
+		for _, pendingAudit := range pendingAudits {
+			if nodeID == pendingAudit.NodeID {
```
coyle: maybe a better algorithm here would be to build a map of [NodeID]pendingAudit then check if any nodeIDs from failedBatch are present in the map, would decrease this in time complexity

satellite/audit/reporter.go
ethanadams: I agree. Updated

satellite/audit/reporter.go
ethanadams: Added map for lookup

satellite/audit/reporter.go
```diff
@@ -152,27 +152,27 @@ func (reporter *Reporter) recordOfflineStatus(ctx context.Context, offlineNodeID
...
 	return nil, nil
 }
 
 // recordAuditSuccessStatus updates nodeIDs in overlay with isup=true, auditsuccess=true
 func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successNodeIDs storj.NodeIDList) (failed storj.NodeIDList, err error) {
 	defer mon.Task()(&ctx)(&err)
-	var errlist errs.Group
-	for _, nodeID := range successNodeIDs {
-		_, err := reporter.overlay.UpdateStats(ctx, &overlay.UpdateRequest{
+
```
aligeti: shouldn't you add  the length of successNodeIDs > 0 check? else return nil with a warning?

satellite/audit/reporter.go
```diff
@@ -181,6 +181,8 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
 func (reporter *Reporter) recordPendingAudits(ctx context.Context, pendingAudits []*PendingAudit) (failed []*PendingAudit, err error) {
 	defer mon.Task()(&ctx)(&err)
 	var errlist errs.Group
+
+	var updateRequests []*overlay.UpdateRequest
```
aligeti: shouldn't you add the length of pendingAudits > 0 check? else return nil with a warning? and at other similar usages

satellite/audit/reporter.go
aligeti: I think the conditions should be separated. BatchupdateStats returns  nil for error, if the passed slice length is zero, nothing to update and returning nil is ok. but changing the condition to OR will return err.  As suggested above, checking the length of the slice before calling BatchupdateStats will work with using `if err != nil || len(failed) > 0 `

satellite/satellitedb/overlaycache.go
simongui: LWW (last write wins) would be the guarantee this provides wouldn't it?

PR #2633, opened 2019-07-25
Add RetainStatus to storagenode config

mobyvb:
What: 
Make it possible for SNOs to configure retain requests to be enabled, disabled, or in debug mode.

The default setting for RetainStatus is "disabled", which will ignore retain requests from the satellite.

Why:
In the future, we will want to enable garbage collection on the satellite. But we may want to be more careful before we fully enable garbage collection everywhere. With this change, we will be able to set retain requests on the storagenode to "debug", which will log the pieces to be deleted and the number of pieces to be deleted, without actually deleting them. It should allow us to verify that garbage collection works as intended before doing it for real.

https://storjlabs.atlassian.net/browse/V3-2232

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/piecestore/endpoint.go
+++ b/storagenode/piecestore/endpoint.go
@@ -70,7 +70,7 @@ type RetainStatus uint32
 
 const (
 	// RetainDisabled means we do not do anything with retain requests
-	RetainDisabled RetainStatus = iota
+	RetainDisabled RetainStatus = iota + 1
 	// RetainEnabled means we fully enable retain requests and delete data not defined by bloom filter
 	RetainEnabled
 	// RetainDebug means we partially enable retain requests, and print out pieces we should delete, without actually deleting them
--- a/storagenode/piecestore/endpoint.go
+++ b/storagenode/piecestore/endpoint.go
@@ -59,14 +59,14 @@ type Config struct {
 	MaxConcurrentRequests int           `help:"how many concurrent requests are allowed, before uploads are rejected." default:"6"`
 	OrderLimitGracePeriod time.Duration `help:"how long after OrderLimit creation date are OrderLimits no longer accepted" default:"1h0m0s"`
 	RetainTimeBuffer      time.Duration `help:"allows for small differences in the satellite and storagenode clocks" default:"1h0m0s"`
-	RetainStatus          RetainStatus  `help:"allows configuration to enable, disable, or test retain requests from the satellite" default:"disabled"`
+	RetainStatus          RetainStatus  `help:"allows configuration to enable, disable, or test retain requests from the satellite. Options: (disabled/enabled/debug)" default:"disabled"`
 
 	Monitor monitor.Config
 	Sender  orders.SenderConfig
 }
 
 // RetainStatus is a type defining the enabled/disabled status of retain requests
-type RetainStatus int
+type RetainStatus uint32
 
 const (
 	// RetainDisabled means we do not do anything with retain requests
@@ -588,7 +588,7 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
 	defer mon.Task()(&ctx)(&err)
 
 	// if retain status is disabled, quit immediately
-	if endpoint.config.RetainStatus == DISABLED {
+	if endpoint.config.RetainStatus == RetainDisabled {
 		return &pb.RetainResponse{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


storagenode/piecestore/endpoint.go
```diff
@@ -59,11 +59,56 @@ type Config struct {
...
 	Monitor monitor.Config
 	Sender  orders.SenderConfig
 }
 
+// RetainStatus is a type defining the enabled/disabled status of retain requests
+type RetainStatus int
+
+const (
+	// RetainDisabled means we do not do anything with retain requests
+	RetainDisabled RetainStatus = iota
```
ifraixedes: I would start the value with 1 because if during a development somebody forget to set the value then it will result on an error due to the zero value of an `int` is 0

storagenode/piecestore/endpoint.go
```diff
@@ -59,11 +59,56 @@ type Config struct {
...
 	OrderLimitGracePeriod time.Duration `help:"how long after OrderLimit creation date are OrderLimits no longer accepted" default:"1h0m0s"`
 	RetainTimeBuffer      time.Duration `help:"allows for small differences in the satellite and storagenode clocks" default:"1h0m0s"`
+	RetainStatus          RetainStatus  `help:"allows configuration to enable, disable, or test retain requests from the satellite" default:"disabled"`
 
 	Monitor monitor.Config
 	Sender  orders.SenderConfig
 }
 
+// RetainStatus is a type defining the enabled/disabled status of retain requests
+type RetainStatus int
```
ifraixedes: This is a status, we don't expect to do any math operations with it, so I think that `uint32` would be better.

storagenode/piecestore/endpoint.go
```diff
@@ -59,11 +59,56 @@ type Config struct {
 	MaxConcurrentRequests int           `help:"how many concurrent requests are allowed, before uploads are rejected." default:"6"`
 	OrderLimitGracePeriod time.Duration `help:"how long after OrderLimit creation date are OrderLimits no longer accepted" default:"1h0m0s"`
 	RetainTimeBuffer      time.Duration `help:"allows for small differences in the satellite and storagenode clocks" default:"1h0m0s"`
+	RetainStatus          RetainStatus  `help:"allows configuration to enable, disable, or test retain requests from the satellite" default:"disabled"`
```
ifraixedes: Is the CLI help message showing the values that this flag can take?

storagenode/piecestore/endpoint.go
```diff
@@ -572,25 +622,35 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
-					continue
-				}
-				if err = endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID); err != nil {
-					endpoint.log.Error("failed to delete piece info", zap.Error(Error.Wrap(err)))
+				endpoint.log.Sugar().Debugf("About to delete piece id (%s) from satellite (%s). RetainStatus: %s", pieceID.String(), peer.ID.String(), endpoint.config.RetainStatus.String())
+
+				// if retain status is enabled, delete pieceid
+				if endpoint.config.RetainStatus == ENABLED {
+					if err = endpoint.store.Delete(ctx, peer.ID, pieceID); err != nil {
+						endpoint.log.Error("failed to delete a piece", zap.Error(Error.Wrap(err)))
```
ifraixedes: There is no need to wrap only for the purpose of logging it.

storagenode/piecestore/endpoint.go
```diff
@@ -572,25 +622,35 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
+						// continue because if we fail to delete from file system,
+						// we need to keep the pieceinfo so we can delete next time
+						continue
+					}
+					if err = endpoint.pieceinfo.Delete(ctx, peer.ID, pieceID); err != nil {
+						endpoint.log.Error("failed to delete piece info", zap.Error(Error.Wrap(err)))
```
ifraixedes: There is no need to wrap only for the purpose of logging it.

storagenode/piecestore/endpoint.go
mobyvb: Could you clarify? I am not sure how someone could forget to set the value since there is a default set. If it is set to an invalid value, the user will get an error at the cli because of `func (v *RetainStatus) Set(s string) error {`

mobyvb: I also don't understand how starting at 1 instead of 0 fixes any issues

ifraixedes: for example

```golang
var rstatus RetainStatus

// Do whatever logic here and the rstatus should be set to some specific
// value, if the value doesn't get set because of a bug, then the value
// will have a non valid value.

```

I prefer to have specific values rather than defaults, except when they are for end users (e.g. CLI, etc.).

When I'm reading code, if I see that var declaration and it isn't set to any value for knowing the value I have to find out which is the default while if the value has to be always set the code is easy to read.

By the way, this is a suggestion, not claiming to do this for approving it.

storagenode/piecestore/endpoint.go
mobyvb: I'll add it to the message, I don't think it does.

storagenode/piecestore/endpoint.go
mobyvb: Ok. The main reason I did the `iota` thing is because that seems standard for enums in golang. Do you think it would be better to just do this?
```
const (
	// RetainDisabled means we do not do anything with retain requests
	RetainDisabled = 1
	// RetainEnabled means we fully enable retain requests and delete data not defined by bloom filter
	RetainEnabled = 2
	// RetainDebug means we partially enable retain requests, and print out pieces we should delete, without actually deleting them
	RetainDebug = 3
)
```

ifraixedes: you can use iota and start with different values
E.g. https://play.golang.org/p/S1GQPCt16YR

mobyvb: Ok, that makes sense. Updated.

storagenode/piecestore/endpoint.go
```diff
@@ -59,11 +59,56 @@ type Config struct {
...
+func (v *RetainStatus) Set(s string) error {
+	switch s {
+	case "disabled":
+		*v = RetainDisabled
+	case "enabled":
+		*v = RetainEnabled
+	case "debug":
+		*v = RetainDebug
+	default:
+		return Error.New("invalid option %q", s)
```
wthorp: nit: maybe "invalid RetainStatus"

mobyvb: done

zeebo:
The only thing is that it might be nice to run some tests under the different retain modes.

mobyvb:
> The only thing is that it might be nice to run some tests under the different retain modes.

I will add tests before merging

PR #2634, opened 2019-07-25
Swaps NO OP logger for Global Logger

coyle:
What: 

If cfg.Volatile.Log is nil we pass a global Logger instead of a no op logger

Why:

If the satellite doesn't provide a logger, we still want to be able to debug the system 

Please describe the tests:
- We see logs again
 
### DIFF
--- a/lib/uplink/uplink.go
+++ b/lib/uplink/uplink.go
@@ -92,7 +92,7 @@ func (cfg *Config) setDefaults(ctx context.Context) error {
 		cfg.Volatile.MaxMemory = 0
 	}
 	if cfg.Volatile.Log == nil {
-		cfg.Volatile.Log = zap.NewNop()
+		cfg.Volatile.Log = zap.L()
 	}
 	if cfg.Volatile.DialTimeout.Seconds() == 0 {
 		cfg.Volatile.DialTimeout = defaultUplinkDialTimeout


stefanbenten:
@coyle Just tested it and it logs all things again properly! **Get it merged!**

lib/uplink/uplink.go
```diff
@@ -92,7 +92,7 @@ func (cfg *Config) setDefaults(ctx context.Context) error {
 		cfg.Volatile.MaxMemory = 0
 	}
 	if cfg.Volatile.Log == nil {
-		cfg.Volatile.Log = zap.NewNop()
+		cfg.Volatile.Log = zap.L()
```
egonelbre: The reason that this was not the default was because of C bindings that shouldn't use stdout or stderr.

coyle: I'm not sure I follow but we should probably solve that is a way that doesn't shut off logging for the rest of the system

egonelbre: When something needs logging, then it would need to explicitly enable it by passing in the appropriate logger. The default should probably still be nop.

coyle: I think logging should be on by default, if something wants to shut off logging, it can pass in a no op logger

egonelbre: I disagree with that, our things are intended to be used as libraries not as a program. Every program has their own logging and concerns to worry about, the log of our library code shouldn't be their concern.

Imagine Go runtime or GRPC or HTTP2 all logging by default, it doesn't make sense, at least for me.

coyle: Should "libraries" log at all ?

egonelbre: I think they should be quiet by default. Vital information is communicated through errors.

coyle: I think uplink is a service not a library and to have it not show errors by default will be confusing and lead to issues, like we experienced.

egonelbre: `uplink`, yes, but not `libuplink`...

coyle: I think the default behavior of no logger is counter intuitive for a "library" like libuplink. It caused us to not log anything and took time to track down what happened. There are well regarded open source projects that log by default and I think we should keep it like this until there is good practical reasons for not doing so.

thepaul: I'd have to disagree. If we're calling the thing "libuplink", then people will treat it as a library and have library expectations for it. Libraries should never log without being explicitly configured to do so.

But I'm all for making it easy to see how that should be done in our docs, sample code, etc.

egonelbre: 1. You may not always have stdout/stderr available
2. C handles buffering differently than Go
3. Our code is concurrent and uses different logging framework which will interrupt stdout output
4. as far as I know, S3, GCS packages don't log

stefanbenten: @egonelbre Can you provide the correct solution that resolves the problem of not logging at all in uplink and gateway?

coyle: >  Libraries should never log without being explicitly configured to do so.
Agreed 100% -- I guess my argument is uplink feels like a service not a library (we aren't calling it libuplink it's just int he lib folder, maybe it shouldn't be) but I lack any good reason for that other than having a config file for a library seems unusual.

egonelbre: Specify:

```
config.Volatile.Log = zap.L()
libuplink.NewUplink(ctx, config)

or 

config.Volatile.Log = parentLogger
libuplink.NewUplink(ctx, config)
```

coyle: the issue is the default when there is no parent

coyle: Talked with @thepaul offline. I'll make another PR adding loggers to calls to NewUplink and swap the default back

PR #2635, opened 2019-07-25
bash script to upload/download/compare between releases

aligeti:
What: 
automating the upload/download/compare  files between alpha releases
The bash script **storjsimscript.sh** is placed in /scripts folder 
Be in the scripts folder to run the script. To run the script type **./storjsimscript.sh** at the command prompt. You will see below options, upon running the scripts

===========================
Tool for quick storj test 
===========================
**1) Download & build Storj Release
2) Run Storj Sim
3) Uplink Upload
4) Uplink Download
5) Compare Files
6) Quit
Choose an option:**

Why:
Automation need

Please describe the tests:
 - Test 1:
 - Test 2:

No impact to the existing storj code base. 


scripts/storjsimscript.sh
```diff
@@ -0,0 +1,151 @@
...
+    echo "Downloading $storjRelease ...."
+    #git clone https://github.com/storj/storj -b $storjRelease
+    cd $HOME/storj
+    echo "build & install storj ...."
+    go install -v ./...
+}
+
+function runStorjSim {
+    echo "Running storj-sim...."
+    osascript -e 'tell app "Terminal" to do script "storj-sim network run"'
```
coyle: this only works on macs

coyle: this can be an executable step

scripts/storjsimscript.sh
```diff
@@ -0,0 +1,151 @@
...
+    then
+        echo "directory exists, deleting....."
+        #rm -rf storj
+    else
+        echo "directory doesn't exists"
+    fi
+
+    read -p 'Enter the Release :' storjRelease
+    echo "Downloading $storjRelease ...."
+    #git clone https://github.com/storj/storj -b $storjRelease
```
coyle: I think we should:
1)  make a temp directory
2) make previous release directory in temp directory
3) make current release directory in temp directory
4) download/clone the release into the current release temp directory
5) download/clone the release into the previous release temp directory

scripts/storjsimscript.sh
coyle: as discussed in person, make this a background process outputting to a log file

scripts/storjsimscript.sh
```diff
@@ -0,0 +1,151 @@
...
+    go install -v ./...
+}
+
+function runStorjSim {
+    echo "Running storj-sim...."
+    osascript -e 'tell app "Terminal" to do script "storj-sim network run"'
+    sleep 2
+}
+
+function uplinkUpload() {
```
coyle: I'm thinking if the goal is to compare upload/download speeds between releases, we should have a standard size we use here for comparison and then output the difference in time

scripts/storjsimscript.sh
```diff
@@ -0,0 +1,151 @@
...
+        if [ -f "$uplinkUploadFile" ]; then
+            uplink cp $uplinkUploadFile sj://$uplinkBucket/
+            UPLOADFILE=$uplinkUploadFile
+        else
+            echo "Invalid file path"
+        fi
+    fi
+}
+
+function uplinkDownload() {
```
coyle: same as above

scripts/storjsimscript.sh
coyle: For the goal of making sure the new releases uploads can download the old release's uploads, we should automate that with junk.

coyle: upload:

1) make random bytes file
2) using previous release, upload to bucket
3) using new release, download from bucket
4) compare downloaded file to uploaded file for equivalency

aligeti: based on slack conversation with @littleskunk my understanding was to very that a file uploaded using a release (say 14) should be able to be downloaded with the next release(say 15). we can for add lots of features this script to do some metric analysis. I can look into the metrics extension. First I wanted to get @littleskunk opinion as well.

coyle: That's was what I intended to describe above. Major takeaway is we don't need the user to specify the file/bucket etc. we can use random bytes and just cleanup.

coyle: we can also combine the download and upload into a single equivalency check

scripts/storjsimscript.sh
```diff
@@ -0,0 +1,151 @@
...
+            echo "**************************************************"
+            echo "  Upload & Download files are *** NOT SAME *** "
+            echo "**************************************************"
+            echo
+            echo
+        fi
+    fi
+}
+
+function shutdown {
```
egonelbre: why is this needed?

egonelbre:
The whole script should be roughly:

```
//TODO: setup temp dir
STORJ_NETWORK_DIR=$TMP

PATH=old-releases-bin
storj-sim network setup
storj-sim network test upload.sh input.bin
PATH=new-releases-bin
storj-sim network test download.sh output.bin
compare input.bin output.bin
```

stefanbenten:
@aligeti You made commits with a different account/setup, which uses a wrong username. you wont be able to get this merged until you change that. Take a look at the commit history of your PR, to find which commits are the culprit.

PR #2644, opened 2019-07-26
CSP implementation for satellite console

crawter:
What: 
Content Security Policies added to satellite console. To implement it i removed some inline styles from markup, updated js bundle generation. Due to this fixes i removed webpack.config.js as redundant file, because we use vue-cli tool to generate js bundle.

Why:
To detect and mitigate certain types of attacks, including Cross Site Scripting (XSS) and data injection attacks

Please describe the tests:
 - Test 1:
 - Test 2:
 
none


satellite/console/consoleweb/server.go
```diff
@@ -117,6 +117,20 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
...
+		"default-src 'self'",
+		"script-src 'self' *.stripe.com cdn.segment.com",
+		"frame-src 'self' *.stripe.com",
+		"img-src 'self' data:",
+	}
+
+	header.Set("Content-Type", "text/html; charset=UTF-8")
+	header.Set("Content-Security-Policy", strings.Join(cspValues, "; "))
+
+	w.WriteHeader(200)
```
stefanbenten: i would prefer the actual html variable here, instead of the integer.

web/satellite/vue.config.js
```diff
@@ -1,15 +1,21 @@
 // Copyright (C) 2019 Storj Labs, Inc.
 // See LICENSE for copying information.
 
+const path = require('path');
+
 module.exports = {
+    publicPath: "/static/dist",
```
stefanbenten: Not sure if thats ok to be an absolute path?

satellite/console/consoleweb/server.go
egonelbre: I think, Go should automatically write it anyways, so need to explicitly add it.

stefanbenten: You are right

web/satellite/vue.config.js
rikysya: That's route to our file server

PR #2652, opened 2019-07-29
Add ability to send vouchers over kademlia

cam-a:
https://storjlabs.atlassian.net/browse/V3-1833
What: This PR adds functionality to send vouchers to other nodes during kademlia refresh. It modifies multiple functions to be able to pass vouchers along the kademlia lookup code path. Also, `self` is passed into lookup, so that not all lookups will request pingback

Why: To require nodes to be vetted by satellites in order to fully enter the kademlia network

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/kademlia/kademlia.go
```diff
@@ -275,16 +277,23 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (_ *pb.InfoRespo
 	return info, nil
 }
 
-// FindNode looks up the provided NodeID first in the local Node, and if it is not found
-// begins searching the network for the NodeID. Returns and error if node was not found
+// FindNode wraps a call to FindNodeWithVouchers with no vouchers
 func (k *Kademlia) FindNode(ctx context.Context, nodeID storj.NodeID) (_ pb.Node, err error) {
+	defer mon.Task()(&ctx)(&err)
+	return k.FindNodeWithVouchers(ctx, nodeID, nil, nil)
```
aligeti: If vouchers are nil, how are you going to send vouchers to other nodes. Is something missing here?

pkg/kademlia/kademlia.go
```diff
@@ -275,16 +277,23 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (_ *pb.InfoRespo
...
+// and request to be put into their routing table. Returns and error if node was not found
+func (k *Kademlia) FindNodeWithVouchers(ctx context.Context, nodeID storj.NodeID, vrs []*pb.Voucher, self *pb.Node) (_ pb.Node, err error) {
 	defer mon.Task()(&ctx)(&err)
 	if !k.lookups.Start() {
 		return pb.Node{}, context.Canceled
 	}
 	defer k.lookups.Done()
 
-	results, err := k.lookup(ctx, nodeID)
+	results, err := k.lookup(ctx, nodeID, vrs, self)
```
aligeti: nodeID of self , can be obtained selfnode := self.Local().Node, then why send self as an argument. What do you think?

pkg/kademlia/kademlia.go
aligeti: do you have to check if 'nodeID' is not nil before calling 
`k.FindNodeWithVouchers(ctx, nodeID, nil, nil)`

pkg/kademlia/kademlia.go
```diff
@@ -275,16 +277,23 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (_ *pb.InfoRespo
...
+// FindNode wraps a call to FindNodeWithVouchers with no vouchers
 func (k *Kademlia) FindNode(ctx context.Context, nodeID storj.NodeID) (_ pb.Node, err error) {
+	defer mon.Task()(&ctx)(&err)
+	return k.FindNodeWithVouchers(ctx, nodeID, nil, nil)
+}
+
+// FindNodeWithVouchers looks up the provided NodeID first in the local Node, and if it is not found
+// begins searching the network for the NodeID. When it contacts nodes, it will send vouchers to them
+// and request to be put into their routing table. Returns and error if node was not found
+func (k *Kademlia) FindNodeWithVouchers(ctx context.Context, nodeID storj.NodeID, vrs []*pb.Voucher, self *pb.Node) (_ pb.Node, err error) {
```
aligeti: do you have to handle any differently, if the node has voucher vs no voucher and trying to find other nodes?

satellite/peer.go
```diff
@@ -331,7 +331,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 			peer.Transport = peer.Transport.WithObservers(peer.Kademlia.RoutingTable)
 		}
 
-		peer.Kademlia.Service, err = kademlia.NewService(peer.Log.Named("kademlia"), peer.Transport, peer.Kademlia.RoutingTable, config)
+		peer.Kademlia.Service, err = kademlia.NewService(peer.Log.Named("kademlia"), peer.Transport, peer.Kademlia.RoutingTable, nil, config)
```
aligeti: Always curious, when satellite is not a Storagenode, then why run kademlia process on satellite and made known to  outside world. This is just a thought .

pkg/kademlia/kademlia.go
cam-a: The only time we want to send vouchers is during kademlia refresh, where we call `FindNodeWithVouchers`
```
func (k *Kademlia) refresh(ctx context.Context, threshold time.Duration) (err error) {
	...		
       _, _ = k.FindNodeWithVouchers(ctx, rID, vrs, &self) // ignore node not found
	...	
}
```

cam-a: self is checked here
```
func (dialer *Dialer) Lookup(ctx context.Context, self *pb.Node, ask pb.Node, find storj.NodeID, limit int, vouchers []*pb.Voucher) (_ []*pb.Node, err error) {
	...
	if self != nil {
		req.Pingback = true
		req.Sender = self	
		req.Vouchers = vouchers
	}
        ...
}
```

satellite/peer.go
cam-a: We're going to remove the satellite from kademlia soon

pkg/kademlia/kademlia.go
cam-a: sometimes when we call lookup, we want to pass self, other times we don't
related: https://github.com/storj/storj/pull/2652#discussion_r308390951

pkg/kademlia/kademlia.go
cam-a: Whether it has a voucher or not will determine if the node goes into another node's antechamber. We expect them to have no vouchers if they are brand new nodes

PR #2654, opened 2019-07-29
Adds more detailed error logging information V3-2239

aligeti:
What: 
Add more info about the nodes when they fail audit
Why:
more error logging detailed info needed for debugging 

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/satellite/audit/reporter.go
+++ b/satellite/audit/reporter.go
@@ -131,6 +131,7 @@ func (reporter *Reporter) recordAuditFailStatus(ctx context.Context, failedAudit
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit Failed Nodes ", zap.Strings("Node IDs", failed.Strings()))
 		return failed, errs.Combine(Error.New("failed to record some audit fail statuses in overlay"), errlist.Err())
 	}
 	return nil, nil
@@ -148,6 +149,7 @@ func (reporter *Reporter) recordOfflineStatus(ctx context.Context, offlineNodeID
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit Offline Nodes ", zap.Strings("Node IDs", failed.Strings()))
 		return failed, errs.Combine(Error.New("failed to record some audit offline statuses in overlay"), errlist.Err())
 	}
 	return nil, nil
@@ -169,6 +171,7 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit success Nodes ", zap.Strings("Node IDs", failed.Strings()))
 		return failed, errs.Combine(Error.New("failed to record some audit success statuses in overlay"), errlist.Err())
 	}
 	return nil, nil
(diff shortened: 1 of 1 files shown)


satellite/audit/reporter.go
```diff
@@ -131,6 +131,7 @@ func (reporter *Reporter) recordAuditFailStatus(ctx context.Context, failedAudit
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit Failed Nodes ", zap.Strings("Node IDs", failed.Strings()))
```
cam-a: It might be worth changing these statements to `log.Error`

satellite/audit/reporter.go
```diff
@@ -148,6 +149,7 @@ func (reporter *Reporter) recordOfflineStatus(ctx context.Context, offlineNodeID
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit Offline Nodes ", zap.Strings("Node IDs", failed.Strings()))
```
cam-a: This logger is named `audit:reporter`, so having the word "Audit" in the message might be unnecessary

satellite/audit/reporter.go
```diff
@@ -199,6 +202,9 @@ func (reporter *Reporter) recordPendingAudits(ctx context.Context, pendingAudits
 		}
 	}
 	if len(failed) > 0 {
+		for _, v := range failed {
+			reporter.log.Debug("Error recording Audit Pending Nodes ", zap.Stringer("Node ID", v.NodeID), zap.String("Path", v.Path))
```
cam-a: I think in general a good message for these logs would be `failed to record x y z`.

satellite/audit/reporter.go
cam-a: Oh, actually, I see that the errors themselves are being logged as `Error`. Debug is okay

cam-a: Nevermind, debug is good

coyle: There is a possibility log is nil. I would do a check before calling.

satellite/audit/reporter.go
```diff
@@ -169,6 +171,7 @@ func (reporter *Reporter) recordAuditSuccessStatus(ctx context.Context, successN
 		}
 	}
 	if len(failed) > 0 {
+		reporter.log.Debug("Error recording Audit success Nodes ", zap.Strings("Node IDs", failed.Strings()))
```
coyle: Node IDs -> NodeIDs

satellite/audit/reporter.go
coyle: Node IDs -> NodeIDs

satellite/audit/reporter.go
aligeti: updated by initializing the log if it is nil

satellite/audit/reporter.go
aligeti: updated

aligeti: updated

satellite/audit/reporter.go
aligeti: updated

satellite/audit/reporter.go
aligeti: updated

satellite/audit/reporter.go
```diff
@@ -36,6 +36,10 @@ type Report struct {
 
 // NewReporter instantiates a reporter
 func NewReporter(log *zap.Logger, overlay *overlay.Cache, containment Containment, maxRetries int, maxReverifyCount int32) *Reporter {
+	if log == nil {
+		log = log.Named("audit:reporter")
```
egonelbre: This shouldn't be done, the logger must come from outside, if it does not then it is a bug.

This check also does not make sense.

coyle: I think it should at minimum be a noop ... I get it's a bug if a logger isn't passed but we shouldn't allow the system to panic

egonelbre: No, this is completely a programmer bug that should never pass tests.

aligeti: it was a comment/suggestion I received from peer review.  I will reopen the branch to update it.

PR #2655, opened 2019-07-29
satellite/rewards: add generate referral link logic

VinozzZ:
What: 
- create a `referralInfo` struct for necessary info for a referral link
- use base64 encoding to encode referralInfo json
- attach encoded string to the referral link
- reorganize `rewards` package to separate logic for UI displaying from data

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/rewards/display.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// OpenSourcePartner contains all data for an Open Source Partner.
+type OpenSourcePartner struct {
+	PartnerInfo
+	PartnerOffers OrganizedOffers
+}
+
+// PartnerSet contains a list of Open Source Partners.
+type PartnerSet []OpenSourcePartner
```
mobyvb: I think `PartnerList` makes more sense

satellite/rewards/display.go
```diff
@@ -0,0 +1,121 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package rewards
+
+// OrganizedOffers contains a list of offers organized by status.
+type OrganizedOffers struct {
+	Active  Offer
```
mobyvb: what are the meanings of `Active` and `Default`?

satellite/rewards/display.go
```diff
@@ -0,0 +1,121 @@
...
+	PartnerID string
+}
+
+// OrganizeOffersByStatus organizes offers by OfferStatus.
+func (offers Offers) OrganizeOffersByStatus() OrganizedOffers {
+	var oo OrganizedOffers
+
+	for _, offer := range offers {
+		switch offer.Status {
+		case Active:
```
mobyvb: do we need to worry about whether there might be multiple offers where `offer.Status` is `Active` or `Default`?

satellite/rewards/display.go
```diff
@@ -0,0 +1,121 @@
...
+			oo.Done = append(oo.Done, offer)
+		}
+	}
+	return oo
+}
+
+// OrganizeOffersByType organizes offers by OfferType.
+func (offers Offers) OrganizeOffersByType() OfferSet {
+	var (
+		fc, ro, p Offers
```
mobyvb: I would prefer longer/more descriptive variable names here, but it's not a big deal since their scope is so small

satellite/rewards/display.go
```diff
@@ -0,0 +1,121 @@
...
+// organizePartnerData returns a list of Open Source Partners
+// whose offers have been organized by status, type, and
+// assigned to the correct partner.
+func organizePartnerData(offers Offers) PartnerSet {
+	partnerData := matchOffersToPartnerSet(offers, createPartnerSet())
+	return partnerData
+}
+
+// getTardigradeDomains returns domain names for tardigrade satellites
+func getTardigradeDomains() []string {
```
mobyvb: is this the only file where these domain names will need to be accessible? If they will be needed elsewhere, it might make sense to push this function to a higher level so you don't have to worry about multiple sources for this information. Or - is this file the place that makes the most sense for this function in this package?

satellite/rewards/partners.go
```diff
@@ -65,3 +66,30 @@ func LoadPartnerInfos() Partners {
...
+
+// GeneratePartnerLink returns base64 encoded partner referral link
+func GeneratePartnerLink(offerName string) ([]string, error) {
+	referralInfo := &referralInfo{UserID: "", PartnerID: GetPartnerIDByName(offerName)}
+	refJSON, err := json.Marshal(referralInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	domians := getTardigradeDomains()
```
mobyvb: `domians` -> `domains`

satellite/rewards/partners.go
```diff
@@ -65,3 +66,30 @@ func LoadPartnerInfos() Partners {
...
+	referralLinks := make([]string, len(domians))
+
+	for i, url := range domians {
+		referralLinks[i] = url + base64.StdEncoding.EncodeToString(refJSON)
```
mobyvb: this makes an assumption that the url has a `/` at the end, which is not always safe to do. I would recommend using `path.Join` from this package: https://golang.org/pkg/path/

mobyvb: Also, you can do `base64.StdEncoding.EncodeToString(refJSON)` outside of the loop, so it does not need to be evaluated separately for each iteration

satellite/rewards/display.go
VinozzZ: There should be only one Active or Default offer at a given time

satellite/rewards/display.go
VinozzZ: I wanted to get the satellite URL dynamically but didn't find a good easy way to do it so I came up with this temporary solution

mobyvb: there might be a way to pass it down from somewhere in `satellite/peer.go`, but I don't know. Maybe that is functionality that can be added in the future.

PR #2656, opened 2019-07-29
Jg/1622 add backwards compatibility test

JessicaGreben:
What: 

Add an integration test that checks upload/downloads work between the current branch and the last release branch.  Currently this PR will only add the scripts containing the tests. We need to fix a backwards compatibility issue with satellite config files (see jira ticket 2340) before we can add the test to jenkins. See jira ticket 2342 that describes the still needed jenkins work.

Why:
So that we have a test that checks that we don't introduce backwards incompatible changes.

Please describe the tests:
 - Test 1:  upload/download test between current branch and last release
 
This will make jenkins slower if we run with every build. We may only want to run this on master branch.


scripts/test-sim-backwards.sh
```diff
@@ -0,0 +1,44 @@
...
+
+TMP=$(mktemp -d -t tmp.XXXXXXXXXX)
+export STORJ_NETWORK_DIR=$TMP
+cleanup(){
+    rm -rf "$STORJ_NETWORK_DIR"
+    echo "cleaned up test successfully"
+}
+trap cleanup EXIT
+
+BRANCH_DIR="$STORJ_NETWORK_DIR/branch"
```
aligeti: shouldnt you be using `git describe --tags `git rev-list --tags --max-count=2`` this command? to get the current (latest) and the previous release?

JessicaGreben: no, we should be testing the latest release and the current branch (or master branch in the future)

isaachess:
Hmmmm, idk if I can approve this, you don't have any tests for your tests .... :wink:

PR #2657, opened 2019-07-29
satellite confirms success threshold reached

navillasa:
What:
- adds an error if the number of remote pieces to upload to is less than the success threshold. (I'm pretty sure this wasn't already happening in CommitSegment, but would welcome feedback. Maybe `validateRedundancy` should've been used differently?)
- adds a test to ensure that the satellite will send an error if unable to upload to success threshold
- also looking for feedback on the use of the pointer's redundancy scheme vs. the metainfo endpoint's RS config (see todos).

Why:
- Satellite should confirm that enough pieces (o) were uploaded - https://storjlabs.atlassian.net/browse/V3-1884

Please describe the tests:
TestCommitSegment:
 - ensures that CommitSegment will error if the number of remote pieces is less than the success threshold
 

satellite/metainfo/metainfo.go
```diff
@@ -523,12 +524,11 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
 			return Error.New("all pieces needs to have the same size")
 		}
 
-		// we repair when the number of healthy files is less than or equal to the repair threshold
-		// except for the case when the repair and success thresholds are the same (a case usually seen during testing)
-		if int32(len(remotePieces)) <= remote.Redundancy.RepairThreshold && int32(len(remotePieces)) < remote.Redundancy.SuccessThreshold {
```
mobyvb: do we still want to return a distinct error if the number of pieces is less than the repair threshold? Or do we want to return the same error in either case as you are doing now?

satellite/metainfo/metainfo_test.go
```diff
@@ -313,11 +313,60 @@ func TestCommitSegment(t *testing.T) {
 				ErasureShareSize: 256,
 			}
 			expirationDate := time.Now().Add(time.Hour)
-			addresedLimits, rootPieceID, _, err := metainfo.CreateSegment(ctx, "bucket", "path", -1, redundancy, 1000, expirationDate)
+			addressedLimits, rootPieceID, _, err := metainfo.CreateSegment(ctx, "bucket", "path", -1, redundancy, 1000, expirationDate)
 			require.NoError(t, err)
 
 			// create number of pieces below repair threshold
-			usedForPieces := addresedLimits[:redundancy.RepairThreshold-1]
+			usedForPieces := addressedLimits[:redundancy.RepairThreshold-1]
```
mobyvb: you could test `redundancy.SuccessThreshold-1` instead to get the edge case

wthorp:
Should this ticket also implement SegmentCommitRequest hash and signature validation? [Probably in another PR?]  Otherwise we could still have bad / rogue / wishful segment?  https://github.com/storj/storj/blob/2f15c0c5e46ac58058aa531d7e9a2457dde9f18e/satellite/metainfo/metainfo.go#L1363

satellite/metainfo/metainfo.go
```diff
@@ -523,12 +524,11 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
-			return Error.New("Number of valid pieces (%d) is less than or equal to the repair threshold (%d)",
+		// todo: is it okay that this is using endpoint.requiredRSConfig instead of pointer rs values?
+		if len(remotePieces) < endpoint.requiredRSConfig.SuccessThreshold {
```
kaloyan-raev: I think we want to make this check against the actual pointer.

kaloyan-raev:
@wthorp we have a ticket for this: https://storjlabs.atlassian.net/browse/V3-1583

satellite/metainfo/metainfo.go
navillasa: yeah I think it would make more sense to add a distinct error for less than repair threshold. I'll go ahead and add that.

satellite/metainfo/metainfo_test.go
navillasa: that's a good idea!

satellite/metainfo/metainfo.go
navillasa: why the actual pointer and not the endpoint settings?

kaloyan-raev: Because we want to check that the uplink reached the success threshold for the specific object it is committing.

It's true the satellite currently checks that redundancy scheme of the pointer matches exact values, but there is a plan to change this in future with a more relaxed check - that the pointer's redundancy falls in some ranges.

navillasa: ah okay. do you know if I should leave this whole `and` statement: 
```if int32(len(remotePieces)) <= remote.Redundancy.RepairThreshold && int32(len(remotePieces)) < remote.Redundancy.SuccessThreshold {
```
or just the latter half?

satellite/metainfo/metainfo.go
wthorp: I think we can remove the entire block above this.  Repair threshold _should_ always be less than success threshold.  If we need a sanity check for this, that sanity check should probably be elsewhere.

satellite/metainfo/metainfo.go
```diff
@@ -523,15 +523,22 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
 			return Error.New("all pieces needs to have the same size")
 		}
 
-		// we repair when the number of healthy files is less than or equal to the repair threshold
-		// except for the case when the repair and success thresholds are the same (a case usually seen during testing)
+		// We repair when the number of healthy files is less than or equal to the repair threshold
```
wthorp: This comment is bogus because this code isn't called by repair (anymore?)

satellite/metainfo/metainfo_test.go
```diff
@@ -343,13 +343,62 @@ func TestCommitSegment(t *testing.T) {
...
+			for i, addressedLimit := range addressedLimits {
+				limits[i] = addressedLimit.Limit
 			}
 			_, err = metainfo.CommitSegment(ctx, "bucket", "path", -1, pointer, limits)
 			require.Error(t, err)
-			require.Contains(t, err.Error(), "less than or equal to the repair threshold")
+			require.Contains(t, err.Error(), "is less than or equal to the repair threshold")
+		}
+
+		{
```
wthorp: I still sort of have the vague feeling that repair threshold (the previous test) shouldn't matter because it _should_ always less than the success threshold.  Not a deal breaker, but more tests = slower testing.

satellite/metainfo/metainfo.go
```diff
@@ -523,15 +523,22 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
+		// except for the case when the repair and success thresholds are the same (a case usually seen during testing).
 		if int32(len(remotePieces)) <= remote.Redundancy.RepairThreshold && int32(len(remotePieces)) < remote.Redundancy.SuccessThreshold {
 			return Error.New("Number of valid pieces (%d) is less than or equal to the repair threshold (%d)",
 				len(remotePieces),
 				remote.Redundancy.RepairThreshold,
 			)
 		}
 
+		if int32(len(remotePieces)) < remote.Redundancy.SuccessThreshold {
```
aligeti: This is preventing from committing into the pointerdb, the info about pieces that are already uploaded and happening to be less than threshold pieces. So satellite won't make a commit segment info into pointerdb. What about the pieces that are already uploaded onto SNOs and now SNOs will be sending those orders to satellite for payment, which satellite will reject. Is my understanding right that this change will potentially use up SNOs bandwidth and space?

mobyvb: Yeah I think you are correct. If `CommitSegment` fails, the nodes that data was uploaded to don't seem to be notified in anyway. But I think garbage collection should help with this in the future. Either way, we do not want the satellite to say that a segment is successful if it was not even uploaded to the correct threshold.

satellite/metainfo/metainfo.go
mobyvb: This code was never called by repair afaik. The comment seems to clarify why there is a `&&` for the error check

satellite/metainfo/metainfo_test.go
mobyvb: I don't think these tests are particularly intense, and having the two separate cases allows us to have more coverage. The first test is to make sure that the error we get when we are less than the repair threshold is the "less than the repair threshold" error. The new test is to test that the error we get when we are greater than the repair threshold but less than the success threshold is the "less than the success threshold" error. These tests are both useful regardless of the issue regarding other tests where the two thresholds are equal.

navillasa: yeah I think I agree with Moby

navillasa:
> is my understanding right about the side effect of this change?

@aligeti I think that problem was already existing, but I opened this ticket: https://storjlabs.atlassian.net/browse/V3-2306

PR #2664, opened 2019-07-30
satellite/satellitedb: get active offer for partners

VinozzZ:
What: 
- add `GetActiveOffer` method on `Offers` struct
- rename `GetCurrentByType` to `GetActiveOffersByType`
- modify `GetActiveOffersByType` from returning one offer to returning offers
Why:
- each partner will have its own active offer which requires `GetActiveOffersByType` to return multiple offers

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/service.go
```diff
@@ -125,10 +125,16 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
 		offerType = rewards.Partner
+	} else if refUserID != "" {
+		offerType = rewards.Referral
 	}
 
 	//TODO: Create a current offer cache to replace database call
-	currentReward, err := s.rewards.GetCurrentByType(ctx, offerType)
+	offers, err := s.rewards.GetActiveOffersByType(ctx, offerType)
+	if err != nil && !rewards.NoCurrentOfferErr.Has(err) {
+		return nil, errs.New(internalErrMsg)
```
ifraixedes: this is an internal error we have to log it for having a clue what could be wrong.

Add this line before returning `s.log.Error("internal error", zap.Error(err))`

I would say that's worth to do in this line 139 too, but that line has not been modified in this branch so it could be done another time.

satellite/console/service.go
```diff
@@ -657,15 +663,14 @@ func (s *Service) GetUsersProjects(ctx context.Context) (ps []Project, err error
...
 
 // GetCurrentRewardByType is a method for querying current active reward offer based on its type
-func (s *Service) GetCurrentRewardByType(ctx context.Context, offerType rewards.OfferType) (reward *rewards.Offer, err error) {
+func (s *Service) GetCurrentRewardByType(ctx context.Context, offerType rewards.OfferType) (offer *rewards.Offer, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	reward, err = s.rewards.GetCurrentByType(ctx, offerType)
+	offers, err := s.rewards.GetActiveOffersByType(ctx, offerType)
 	if err != nil {
 		return nil, errs.New(internalErrMsg)
```
ifraixedes: We would need to log this error too, however, this could be done in another time.

PR #2668, opened 2019-07-30
Adding more info the error logs in bootstrap

aligeti:
What: 
Adding more info the error logs in bootstrap
Why:
Need to know when, how and what caused the failure

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -145,6 +145,7 @@ func (k *Kademlia) DumpNodes(ctx context.Context) (_ []*pb.Node, err error) {
 func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 	defer k.bootstrapFinished.Release()
+	defer k.timeTrack(time.Now(), "Bootstrap")
 
 	if !k.lookups.Start() {
 		return context.Canceled
@@ -152,7 +153,7 @@ func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
 	defer k.lookups.Done()
 
 	if len(k.bootstrapNodes) == 0 {
-		k.log.Warn("No bootstrap address specified.")
+		k.log.Warn("No bootsrap address specified", zap.Stringer(k.routingTable.self.Type.String(), k.routingTable.self.Id))
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/kademlia/kademlia.go
```diff
@@ -145,14 +145,15 @@ func (k *Kademlia) DumpNodes(ctx context.Context) (_ []*pb.Node, err error) {
...
+	defer k.timeTrack(time.Now(), "Bootstrap")
 
 	if !k.lookups.Start() {
 		return context.Canceled
 	}
 	defer k.lookups.Done()
 
 	if len(k.bootstrapNodes) == 0 {
-		k.log.Warn("No bootstrap address specified.")
+		k.log.Warn("No bootsrap address specified", zap.Stringer(k.routingTable.self.Type.String(), k.routingTable.self.Id))
```
jenlij: why do we choose to warn here and debug on 172?

aligeti: I had to add debug info where needed as minimal and mostly beef up the existing logging info

pkg/kademlia/kademlia.go
```diff
@@ -145,14 +145,15 @@ func (k *Kademlia) DumpNodes(ctx context.Context) (_ []*pb.Node, err error) {
 func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 	defer k.bootstrapFinished.Release()
+	defer k.timeTrack(time.Now(), "Bootstrap")
```
egonelbre: There's no need to add the extra "Bootstrap" argument since it's used only once here.

pkg/kademlia/kademlia.go
```diff
@@ -423,3 +426,9 @@ func randomIDInRange(start, end bucketID) (storj.NodeID, error) {
 	}
 	return randID, nil
 }
+
+// timeTrack tracks how long a function ran for
+func (k *Kademlia) timeTrack(start time.Time, name string) {
+	elapsed := time.Since(start)
+	k.log.Debug("", zap.Stringer(k.routingTable.self.Type.String(), k.routingTable.self.Id), zap.Duration(name, elapsed))
```
egonelbre: There's no need to log self.Type and self.Id... this is obvious from the log that you are looking at. It only can be one thing.

pkg/kademlia/kademlia.go
egonelbre: No need to log `self`.

pkg/kademlia/kademlia.go
```diff
@@ -168,13 +169,14 @@ func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
 		var foundOnlineBootstrap bool
 		for i, node := range k.bootstrapNodes {
 			if ctx.Err() != nil {
+				k.log.Debug("Context Error received while Boostraping ", zap.Stringer(k.routingTable.self.Type.String(), k.routingTable.self.Id), zap.Stringer("Bootstrap Node", node.Id))
```
egonelbre: No need to log self.

pkg/kademlia/kademlia.go
```diff
@@ -168,13 +169,14 @@ func (k *Kademlia) Bootstrap(ctx context.Context) (err error) {
...
 				errGroup.Add(ctx.Err())
 				return errGroup.Err()
 			}
 
 			ident, err := k.dialer.FetchPeerIdentityUnverified(ctx, node.Address.Address)
 			if err != nil {
-				errGroup.Add(BootstrapErr.Wrap(err))
+				errGroup.Add(BootstrapErr.Wrap(BootstrapErr.New("%s : %s unable to fetch unverified peer identity node address %s: %s", k.routingTable.self.Type.String(), k.routingTable.self.Id.String(), node.Address.Address, err)))
```
egonelbre: BootstrapErr.New already wraps internally, so no need to do that explicitly. Also there's no need to log self.

pkg/kademlia/kademlia.go
kaloyan-raev: Also no need to call `String()` when using `zap.Stringer`. This is the sole purpose of `zap.Stringer` -- to avoid calling `String()` prematurely.

PR #2671, opened 2019-07-31
don't use global loggers

egonelbre:
What: multiple places were using zap.L(), pass in the properly scoped logger.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/certificates/certificates_test.go
```diff
@@ -609,7 +609,7 @@ func TestCertificateSigner_Sign_E2E(t *testing.T) {
 				require.NoError(t, err)
 				require.NotNil(t, serverOpts)
 
-				service, err := server.New(serverOpts, sc.Address, sc.PrivateAddress, nil, config)
+				service, err := server.New(zap.NewNop(), serverOpts, sc.Address, sc.PrivateAddress, nil, config)
```
stefanbenten: Why do we have the NewNop() Logger here?

egonelbre: fixed

pkg/server/interceptors.go
```diff
@@ -29,20 +29,19 @@ func logOnErrorStreamInterceptor(srv interface{}, ss grpc.ServerStream, info *gr
 			err == io.EOF {
 			return err
 		}
-		zap.S().Errorf("%+v", err)
+		server.log.Sugar().Errorf("%+v", err)
```
ifraixedes: Not to change in this branch, but should we use the normal log rather than the sugared one?

egonelbre: Tried to avoid figuring this out, but probably yes.

ifraixedes: It was a chance to ask, I saw it on Monday while working on something and it popped up here.

PR #2676, opened 2019-07-31
lib/uplink: ensure it's silent by default

egonelbre:
lib/uplink is intended to be used as a library. The default behavior for libraries is to be silent unless explicitly requested.

Please describe the tests:
 - Test 1:
 - Test 2:
 

internal/testplanet/uplink.go
```diff
@@ -330,6 +330,7 @@ func atLeastOne(value int) int {
 func (client *Uplink) NewLibuplink(ctx context.Context) (*libuplink.Uplink, error) {
 	config := getDefaultConfig()
 	libuplinkCfg := &libuplink.Config{}
+	libuplinkCfg.Volatile.Log = client.Log
```
stefanbenten: Can we name this log? From the build logs it looks like it is not named and thus printed with the global logger.

lib/uplinkc/uplink.go
```diff
@@ -27,6 +27,7 @@ func new_uplink(cfg C.UplinkConfig, cerr **C.char) C.UplinkRef {
 	scope := rootScope("") // TODO: pass in as argument
 
 	libcfg := &uplink.Config{} // TODO: figure out a better name
+	// TODO: add a way to support logging
```
stefanbenten: Would be good to add a ticket for this. I assume, this will be highly requested once we are in public Beta

internal/testplanet/uplink.go
egonelbre: Hmm, it should be using client.Log, which is named.

stefanbenten: Let me take a look at the logs again, maybe i mixed something up

stefanbenten: Oh seems like my fault: https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/PR-2676/1/pipeline/43#step-59-log-839
Thats the output of the actual stdout and stderr of the test using the gateway/uplink

stefanbenten:
I am going to check this PR out and test it.

lib/uplinkc/uplink.go
egonelbre: https://storjlabs.atlassian.net/browse/V3-2302

ifraixedes: Could you add the ticket number to the `TODO` comment?

mobile/uplink.go
```diff
@@ -46,6 +46,7 @@ func NewUplink(config *Config, tempDir string) (*Uplink, error) {
 
 	cfg := &libuplink.Config{}
 	if config != nil {
+		// TODO: support logging somehow
```
ifraixedes: Maybe another ticket for this one too?

lib/uplinkc/uplink.go
egonelbre: added

mobile/uplink.go
egonelbre: already done

stefanbenten:
Just checked the Funcionality and works flawlessly!

PR #2678, opened 2019-07-31
docs/design: multipart upload

egonelbre:
This design document describes what problems multipart upload solves and how it can be implemented. Currently there is no direct support for multipart upload on the Satellite side. The feature is simulated completely on the gateway side, which requires keeping a lot of data in memory and doesn't achieve the full usefulness of metainfo upload.

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/design/multipart-upload.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Design
+
+We change the path handling to support listing parts uploaded in arbitrary order. When we start an upload we will create an object at the following path:
+
+```
+<project-id>/<bucket-id>/objects/<path ...>_<version> => object information (partial)
+```
+
+The object will contain a `<stream-id>` that defines the immutable content. We suffix the path with the version number to distinguish between different versions.
```
kaloyan-raev: Where exactly the object will contain the `<stream-id>`? In the object's content/body, or as a key in the UserMetadata?

docs/design/multipart-upload.md
```diff
@@ -0,0 +1,82 @@
...
+* the parts can be uploaded from different computers,
+* the parts can be uploaded out of order, and
+* the parts can be reuploaded, in case of a failed upload.
+
+## Design
+
+We change the path handling to support listing parts uploaded in arbitrary order. When we start an upload we will create an object at the following path:
+
+```
+<project-id>/<bucket-id>/objects/<path ...>_<version> => object information (partial)
```
kaloyan-raev: Can we avoid adding new special characters (`_` in this case) to the path grammar?

It can be `<project-id>/<bucket-id>/objects/<path ...>/<version>`. The rule can be that the last path element is the version.

docs/design/multipart-upload.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+The object will contain a `<stream-id>` that defines the immutable content. We suffix the path with the version number to distinguish between different versions.
+
+We don't know the part sizes, and segments cannot be arbitrarily large, which means we need to split each part into multiple segments. To uniquely find segments we'll assign them a unique number `<part-number>_<segment-number>`, which we call the _segment position_.
```
kaloyan-raev: Same here regarding the underscore. Can we define the segment position as `<part-number>/<segment-number>` instead of `<part-number>_<segment-number>`?

docs/design/multipart-upload.md
egonelbre: I think that would make distinguishing folders more difficult.

docs/design/multipart-upload.md
egonelbre: The segment position would end up as a single `int64`, the `_` is a way of delimiting numbers.

docs/design/multipart-upload.md
egonelbre: Object content since it's essential part of the structure. UserMetadata contains things that the satellite doesn't need.

docs/design/multipart-upload.md
egonelbre: I'm open to other suggestions, but I failed to come up with a solution. We could use some special character based on path encoding.

docs/design/multipart-upload.md
kaloyan-raev: OK

docs/design/multipart-upload.md
kaloyan-raev: How does this make a difference?

If we use `/` instead of `_` then we can list and iterate parts and segments like we do for object and folders.

For example:
- If we want to list all segment of a part, we could list with prefix `<project-id>/<bucket-id>/streams/<stream-id>/<part-number>`
- If we want to count all parts of an object, we could list non-recursively with prefix `<project-id>/<bucket-id>/streams/<stream-id>`

egonelbre: We can still do the same by using the stream-id as the prefix. So, list everything that has `<project-id>/<bucket-id>/streams/<stream-id>/<part-number>` as the prefix.

kaloyan-raev: Oh, you are saying that the KeyValueStore does not really know anything about path delimiters and the prefix does not need to consist of complete path elements.

In such case, yeah, no need to change anything.

kaloyan-raev: Hmm, but what would be the list prefix to list count the number of parts in an object? If we list everything with `<project-id>/<bucket-id>/streams/<stream-id>` as prefix it will return all segments, not just the parts.

egonelbre: You would get that by listing all the segments and finding the unique parts... however, true, it could be beneficial to list all the parts. Listing all parts would be important only for uploads, so I'm not too worried about it either. I'll think about it a bit.

docs/design/multipart-upload.md
kaloyan-raev: I am thinking how would this affect listing folders...

Let's assume we have this in the metainfo db:
```
<project-id>/<bucket-id>/objects/a/b/c/x_0
<project-id>/<bucket-id>/objects/a/b/c/x_1
<project-id>/<bucket-id>/objects/a/b/c/x_2
<project-id>/<bucket-id>/objects/a/b/c/y_0
<project-id>/<bucket-id>/objects/a/b/c/z_0
```
If we list with prefix `<project-id>/<bucket-id>/objects/a/b/c/`, this would return all files (x, y and z) with all their versions, right?

What if we would want to list only the latest version? There should be something that should post-process the result list, right?

What if we want to list only the latest version with limit = 2, i.e. to have x_2 and y_0 in the result? Will it be possible to do this efficiently?

kaloyan-raev: Also, just getting the latest version of an object would require listing all its versions to find the latest one, right?

We need to list with prefix like `<project-id>/<bucket-id>/objects/a/b/c/x_` and then find the greatest version. If there are thousands of versions then this would require several iterations of listing. Meaning the performance to access the latest version of an object would degrade with more versions added over time.

egonelbre:
Moved to https://review.dev.storj.io/c/storj/storj/+/282

PR #2680, opened 2019-07-31
Split storage node sqlite3 databases into individual databases

simongui:
**What:**  This PR migrates the current SQLite3 database into separate SQLite3 database files for each table. The process uses the SQLite3 backup API because of its additional safety. I felt this approach was worth while because it eliminates certain types of bugs that could be introduced by someone unaware of the migration code if we just did a file copy. Even in my own testing the SQLite3 backup API has informed me when something failed and was corrupted or incomplete.

For simplification of our database code I left `info.db` present but it only contains the `versions` table.

This is how the new database and table layout looks.

```
DATABASE FILE                   | TABLES
--------------------------------------------------
bandwidth_usage.db            bandwidth_usage, bandwidth_usage_rollup
certificate.db                certificate
info.db                       versions
order_archive.db              order_archive_
pieceinfo.db                  pieceinfo_
unsent_order.db               unsent_order
used_serial.db                used_serial_
vouchers.db                   vouchers
```

**Why:** We are suffering from a lot of lock contention in the storage node SQLite3 database. We've decided to split each table into their own SQLite3 database files.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/internal/dbutil/sqlite3_backup.go
+++ b/internal/dbutil/sqlite3_backup.go
@@ -1,0 +1,149 @@
+package dbutil
+
+import (
+	"context"
+	"database/sql"
+	"path/filepath"
+	"strings"
+
+	"github.com/mattn/go-sqlite3"
+	"github.com/zeebo/errs"
+)
+
+// ErrSqlite3Backup is the default error class for sqlite3_backup.
+var ErrSqlite3Backup = errs.Class("sqlite3_backup")
+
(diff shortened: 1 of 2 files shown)


internal/dbutil/sqlite3_backup.go
```diff
@@ -0,0 +1,149 @@
+package dbutil
```
egonelbre: move into sqliteutil

PR #2681, opened 2019-07-31
satellite/overlay: reduce calls to UpdateAddress

isaachess:
What: We frequently call UpdateAddress in the overlaycache. This is causing issues with database performance. This is a quick fix that reduces those calls.

It adds an in-memory cache that checks whether the address has changed from before in the UpdateAddress call.

Why: Database performance is suffering due to so many UpdateAddress calls.

Please describe the tests:
 - Test 1: New unit tests confirm that the caching mechanism works as designed.
 
Please describe the performance impact: This should great improve database performance by greatly reducing UpdateAddress calls.

### DIFF
--- a/satellite/overlay/combinedcache.go
+++ b/satellite/overlay/combinedcache.go
@@ -1,0 +1,90 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package overlay
+
+import (
+	"context"
+	"sync"
+
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+type addressInfo struct {
+	address   string
+	lastIP    string
+	transport pb.NodeTransport
+}
+
+// CombinedCache is a simple caching mechanism for overlaycache updates. It
+// provdes methods to help reduce calls to UpdateAddress and UpdateTime, but can
+// be extended for other calls in the future.
+type CombinedCache struct {
+	DB
+	addressLock  sync.RWMutex
+	addressCache map[storj.NodeID]*addressInfo
+
+	keyLock *KeyLock
+}
+
+// NewCombinedCache instantiates a new CombinedCache
+func NewCombinedCache(db DB) *CombinedCache {
+	return &CombinedCache{
+		DB:           db,
+		addressCache: make(map[storj.NodeID]*addressInfo),
+		keyLock:      NewKeyLock(),
+	}
+}
+
+// UpdateAddress overrides the underlying db.UpdateAddress and provides a simple
+// caching layer to reduce calls to the underlying db. The cache is guaranteed
+// to match the values held in the database; however this code does not
+// guarantee that concurrent UpdateAddress calls will be handled in any
+// particular order.
+func (c *CombinedCache) UpdateAddress(ctx context.Context, info *pb.Node, defaults NodeSelectionConfig) (err error) {
+	// Update internal cache and check if this call requires a db call
+
+	if info == nil {
+		return ErrEmptyNode
+	}
+
+	address := info.Address
+	if address == nil {
+		address = &pb.NodeAddress{}
+	}
+
+	c.addressLock.RLock()
+	cached, ok := c.addressCache[info.Id]
+	c.addressLock.RUnlock()
+
+	if ok &&
+		address.Address == cached.address &&
+		address.Transport == cached.transport &&
+		info.LastIp == cached.lastIP {
+
+		return nil
+	}
+
+	// Acquire lock for this node ID. This prevents a concurrent db update to
+	// this same node ID and guarantees the cache and database stay in sync.
+	// This solution works so long as calls to this code are occurring within a
+	// single process.
+	unlockFunc := c.keyLock.Lock(info.Id)
+	defer unlockFunc()
+
+	err = c.DB.UpdateAddress(ctx, info, defaults)
+	if err != nil {
+		return err
+	}
+
+	c.addressLock.Lock()
+	c.addressCache[info.Id] = &addressInfo{
+		address:   address.Address,
+		lastIP:    info.LastIp,
+		transport: address.Transport,
+	}
+	c.addressLock.Unlock()
+
+	return nil
+}
--- a/satellite/overlay/keylock.go
+++ b/satellite/overlay/keylock.go
@@ -1,0 +1,64 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package overlay
+
+import (
+	"sync"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// KeyLock provides per-key RW locking. Locking is key-specific, meaning lock
+// contention only exists for locking calls using the same key parameter. As
+// with all locks, do not call Unlock() or RUnlock() before a corresponding
+// Lock() or RLock() call.
+//
+// Note on memory usage: internally KeyLock lazily and atomically creates a
+// separate sync.RWMutex for each key. To maintain synchronization guarantees,
+// these interal mutexes are not freed until the entire KeyLock instance is
+// freed.
+type KeyLock struct {
+	locksMu sync.Mutex
+	locks   map[storj.NodeID]*sync.RWMutex
+}
+
+// UnlockFunc is the function to unlock the associated successful lock
+type UnlockFunc func()
+
+// NewKeyLock create a new KeyLock
+func NewKeyLock() *KeyLock {
+	return &KeyLock{
+		locks: make(map[storj.NodeID]*sync.RWMutex),
+	}
+}
+
+// Lock the provided key. Returns the unlock function.
+func (l *KeyLock) Lock(nodeID storj.NodeID) UnlockFunc {
+	lock := l.getLock(nodeID)
+	lock.Lock()
+	return lock.Unlock
+}
+
+// RLock the provided key. Returns the unlock function.
+func (l *KeyLock) RLock(nodeID storj.NodeID) UnlockFunc {
+	lock := l.getLock(nodeID)
+	lock.RLock()
+	return lock.RUnlock
+}
+
+// getLock will atomically load the RWMutex for this key. If one does not yet
+// exist, it will be lazily and atomically created. The resulting RWMutex is
+// returned.
+func (l *KeyLock) getLock(nodeID storj.NodeID) *sync.RWMutex {
+	l.locksMu.Lock()
+	defer l.locksMu.Unlock()
+
+	lo, ok := l.locks[nodeID]
+	if ok {
+		return lo
+	}
+	mu := &sync.RWMutex{}
+	l.locks[nodeID] = mu
+	return mu
+}
--- a/satellite/overlay/keylock_test.go
+++ b/satellite/overlay/keylock_test.go
@@ -1,0 +1,52 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information
+
+package overlay_test
+
+import (
+	"testing"
+
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/internal/teststorj"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/overlay"
+)
+
+func TestKeyLock(t *testing.T) {
+	ml := overlay.NewKeyLock()
+	key := teststorj.NodeIDFromString("hi")
+	unlockFunc := ml.Lock(key)
+	unlockFunc()
+	unlockFunc = ml.RLock(key)
+	unlockFunc()
+}
+
+func BenchmarkKeyLock(b *testing.B) {
+	b.ReportAllocs()
+	ml := overlay.NewKeyLock()
+	numNodes := 100
+	nodeIDs := make([]storj.NodeID, numNodes)
+	for i := 0; i < numNodes; i++ {
+		nodeIDs[i] = testrand.NodeID()
+	}
+	b.Run("lock all new nodes", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			unlockFunc := ml.Lock(testrand.NodeID())
+			unlockFunc()
+		}
+	})
+	b.Run("lock existing nodes", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			idx := i % numNodes
+			unlockFunc := ml.Lock(nodeIDs[idx])
+			unlockFunc()
+		}
+	})
+	b.Run("rlock existing nodes", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			idx := i % numNodes
+			unlockFunc := ml.RLock(nodeIDs[idx])
+			unlockFunc()
+		}
+	})
+}
--- a/satellite/peer.go
+++ b/satellite/peer.go
@@ -158,6 +158,7 @@ type Peer struct {
 	}
 
 	Overlay struct {
+		DB        overlay.DB
 		Service   *overlay.Service
 		Inspector *overlay.Inspector
 	}
@@ -271,7 +272,8 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 	{ // setup overlay
 		log.Debug("Starting overlay")
 
-		peer.Overlay.Service = overlay.NewService(peer.Log.Named("overlay"), peer.DB.OverlayCache(), config.Overlay)
+		peer.Overlay.DB = overlay.NewCombinedCache(peer.DB.OverlayCache())
+		peer.Overlay.Service = overlay.NewService(peer.Log.Named("overlay"), peer.Overlay.DB, config.Overlay)
 		peer.Transport = peer.Transport.WithObservers(peer.Overlay.Service)
 
 		peer.Overlay.Inspector = overlay.NewInspector(peer.Overlay.Service)
@@ -485,7 +487,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 			peer.Log.Named("garbage collection"),
 			config.GarbageCollection,
 			peer.Transport,
-			peer.DB.OverlayCache(),
+			peer.Overlay.DB,
 			peer.Metainfo.Loop,
 		)
 	}
@@ -643,7 +645,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 
 		peer.NodeStats.Endpoint = nodestats.NewEndpoint(
 			peer.Log.Named("nodestats:endpoint"),
-			peer.DB.OverlayCache(),
+			peer.Overlay.DB,
 			peer.DB.StoragenodeAccounting())
 
 		pb.RegisterNodeStatsServer(peer.Server.GRPC(), peer.NodeStats.Endpoint)
--- a/satellite/overlay/service.go
+++ b/satellite/overlay/service.go
@@ -305,6 +305,7 @@ func (service *Service) Put(ctx context.Context, nodeID storj.NodeID, value pb.N
 	if value.Address == nil {
 		return errors.New("node has no address")
 	}
+
 	// Resolve IP Address Network to ensure it is set
 	value.LastIp, err = GetNetwork(ctx, value.Address.Address)
 	if err != nil {
--- a/satellite/overlay/statdb_test.go
+++ b/satellite/overlay/statdb_test.go
@@ -26,6 +26,12 @@ func TestStatDB(t *testing.T) {
 
 		testDatabase(ctx, t, db.OverlayCache())
 	})
+	satellitedbtest.Run(t, func(t *testing.T, db satellite.DB) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+
+		testDatabase(ctx, t, overlay.NewCombinedCache(db.OverlayCache()))
+	})
 }
 
 func testDatabase(ctx context.Context, t *testing.T, cache overlay.DB) {


satellite/peer.go
```diff
@@ -80,7 +81,7 @@ type DB interface {
 	// CertDB returns database for storing uplink's public key & ID
 	CertDB() certdb.DB
 	// OverlayCache returns database for caching overlay information
-	OverlayCache() overlay.DB
+	OverlayCache(isUpDur time.Duration) overlay.DB
```
JessicaGreben: `isUpDur` seems like an odd var name. what does it mean?

satellite/overlay/config.go
```diff
@@ -20,7 +20,8 @@ var (
 // Overlay cache responsibility.
 type Config struct {
 	Node                 NodeSelectionConfig
-	UpdateStatsBatchSize int `help:"number of update requests to process per transaction" default:"100"`
+	UpdateStatsBatchSize int           `help:"number of update requests to process per transaction" default:"100"`
+	UptimeCacheDuration  time.Duration `help:"the length of time to cache uptime before updating the overlay cache again" releaseDefault:"1m" devDefault:"1s"`
```
navillasa: when this is set to 0 does it just constantly update or never update?

satellite/satellitedb/updatechecker.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return cached.stats
+}
+
+func (c *CachedChecker) SetNodeStats(nodeId storj.NodeID, isUp bool, stats *overlay.NodeStats) {
+	c.uptimeLock.Lock()
+	c.uptimeCache[nodeId] = &uptimeInfo{
+		isUp:       isUp,
+		lastUptime: time.Now(),
```
navillasa: maybe this should be time.Now().UTC()?

satellite/overlay/config.go
isaachess: Constantly update.

satellite/satellitedb/updatechecker.go
isaachess: Can you help me understand why?

simongui: I think since it's stored in memory there's no functional issue with it currently but it opens up a potential bug in the future if someone hooks this up to persistence and doesn't notice we aren't using UTC. Maybe just use UTC to protect against accidental bugs in the future?

navillasa: oh wait, maybe timezones aren't relevant here, but I thought it was sort of a rule of thumb to generally use `time.Now().UTC()` instead of `time.Now()` for consistency.

isaachess:
@simongui I don't think there would be much benefit, and there would be additional cost due to the additional lock contention. Imo it's nicer to have them separate so that a call to UpdateAddress is never blocked by a call to UpdateUptime, and vice-versa.

satellite/satellitedb/updatechecker.go
isaachess: We already are persisting time.Now() in the overcache db. If you look here https://github.com/storj/storj/blob/bd42ab28d671120e777f479f37aa0af3c4c88bc7/satellite/satellitedb/overlaycache.go#L883 you'll see that what we actually persist is time.Now() in the db, so this keeps the cache using the same times that we have in the db.

simongui:
I don't think I saw that we ever `delete()` from the 

> @simongui I don't think there would be much benefit, and there would be additional cost due to the additional lock contention. Imo it's nicer to have them separate so that a call to UpdateAddress is never blocked by a call to UpdateUptime, and vice-versa.

Good point about the lock contention.

isaachess:
@simongui we don't delete from the cache (is that what you were typing?). It will grow at most to have a cache for each node, which should be easily manageable given the number of nodes we have.

simongui:
> @simongui we don't delete from the cache (is that what you were typing?). It will grow at most to have a cache for each node, which should be easily manageable given the number of nodes we have.

LOL you're replying faster than I'm fixing my typos and edits hahaha.

I'm worried that the satellite is a very long running process and what if someone spins up a bunch of malicious nodes and floods our cache we would never remove them. Is this possible? If so do we need to clean up the caches periodically? I'm not clear if we protect from this elsewhere in the system.

isaachess:
> what if someone spins up a bunch of malicious nodes and floods our cache we would never remove them. Is this possible?

@simongui That is a good question. I imagine (though only guessing) that if we aren't protected against this, it would cause a large number of other problems in the system, and my cache would be only one of many worries.

I do know that we require a proof-of-work node ID, which I think makes this scenario unlikely. I believe it can take several minutes to generate a node ID with the proper proof-of-work?

cmd/satellite/usage.go
```diff
@@ -54,7 +54,7 @@ func generateCSV(ctx context.Context, start time.Time, end time.Time, output io.
 	for _, row := range rows {
 		nid := row.NodeID
 
-		node, err := db.OverlayCache().Get(ctx, nid)
+		node, err := db.OverlayCache(0).Get(ctx, nid)
```
egonelbre: This seems like a weird change when reading this code. Maybe instead add a method that changes the properties.

`db.OverlayCache().WithCacheDuration()`?

egonelbre: Alternatively, maybe let's set some default cache and then have an explicit `Flush`/`Sync` call?

satellite/overlay/config.go
egonelbre: Wondering whether there's a better name... maybe `CoalesceDuration`, `UptimeFlushInterval`, `UptimeFlushDelay`?

satellite/satellitedb/database.go
```diff
@@ -110,8 +112,8 @@ func (db *DB) Attribution() attribution.DB {
 }
 
 // OverlayCache is a getter for overlay cache repository
-func (db *DB) OverlayCache() overlay.DB {
-	return &overlaycache{db: db.db}
+func (db *DB) OverlayCache(isUpDur time.Duration) overlay.DB {
```
egonelbre: Won't this end up creating multiple caches?

satellite/satellitedb/updatechecker.go
```diff
@@ -0,0 +1,115 @@
...
+	return &CachedChecker{
+		addressCache: make(map[storj.NodeID]*addressInfo),
+		uptimeCache:  make(map[storj.NodeID]*uptimeInfo),
+		isUpDur:      isUpDur,
+	}
+}
+
+// AddressRequiresUpdate returns true if the address should be updated in the
+// underlying db, and false if not. It also keeps the internal cache up-to-date.
+func (c *CachedChecker) AddressRequiresUpdate(node *pb.Node) bool {
```
egonelbre: This is also modifies the state, however the method name implies it doesn't change it.

satellite/satellitedb/overlaycache.go
```diff
@@ -899,7 +916,15 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
 	}
 
-	return getNodeStats(dbNode), Error.Wrap(tx.Commit())
+	err = tx.Commit()
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	stats = getNodeStats(dbNode)
+	cache.updateChecker.SetNodeStats(nodeID, isUp, stats)
```
egonelbre: What if this races with some other update uptime?

satellite/satellitedb/overlaycache.go
```diff
@@ -824,6 +836,11 @@ func (cache *overlaycache) UpdateNodeInfo(ctx context.Context, nodeID storj.Node
 func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID, isUp bool, lambda, weight, uptimeDQ float64) (stats *overlay.NodeStats, err error) {
 	defer mon.Task()(&ctx)(&err)
 
+	stats = cache.updateChecker.GetNodeStats(nodeID, isUp)
```
egonelbre: what if this races with some other update uptime?

satellite/satellitedb/database.go
isaachess: Yes, I'm fixing that now. Good catch, I wondered if anyone would notice before I had a chance to fix it and change how it works. :)

egonelbre:
Do we have performance benchmarks to demonstrate this helps?

satellite/satellitedb/overlaycache.go
isaachess: It is possible to race, but I left it because the original code is also susceptible to race conditions. That is, if two `UpdateUptime` calls are made from two different routines, with different information, there's no guarantee which will be applied last (and persisted) in the db. Because of that, it seemed better to keep the solution simpler and allow for a short-term race issue.

If you feel the potential race is unacceptable, let me know and I'll happily change it, though it will complicate the solution.

satellite/satellitedb/overlaycache.go
isaachess: See comment above.

satellite/satellitedb/overlaycache.go
egonelbre: Yeah, but they have a tx boundary to serialize the changes. This could end up caching a DQ-ed node in the updateChecker as being valid. But, yeah, I'm not sure how much damage it would do.

isaachess: It's true, but only for the duration specified, if that makes sense. If we hold the cache for each node for a limited time, it reduces the risk of being out-of-date in a way that would actually affect the system ... I think. :)

I'd appreciate any more thoughts on this, I did give it some thought while developing but thought that the risk of problems was low enough that it didn't justify the added complexity. I'll give it more thought, see if I can think of a different way to do it.

satellite/satellitedb/updatechecker_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/teststorj"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/satellite/overlay"
+	"storj.io/storj/satellite/satellitedb"
+)
+
+func TestCachedChecker(t *testing.T) {
+	cachedChecker := satellitedb.NewCachedChecker(30 * time.Minute)
```
egonelbre: Let's do parallel tests for these.

satellite/satellitedb/updatechecker.go
```diff
@@ -0,0 +1,115 @@
...
+		return true
+	}
+
+	return false
+}
+
+// GetNodeStats returns cached NodeStats for the supplied nodeId. If the
+// returned stats are not nil the caller should use them as up-to-date stats. If
+// nil, the called should re-cache NodeStats for this nodeId.
+func (c *CachedChecker) GetNodeStats(nodeId storj.NodeID, isUp bool) (stats *overlay.NodeStats) {
```
egonelbre: This also needs a better name.

satellite/satellitedb/overlaycache.go
egonelbre: I'm trying to figure out the potential problem cases... so here's the table:

```
db   time        1s     30s   1m
--------------------------------
up   no-cache    up!    up    up
up   no-cache    do!    up    up
do   no-cache    do!    up    up
do   no-cache    up!    up    up

up   cached      up     up    up!
up   cached      do!    up!   up
do   cached      do     up!   up
do   cached      up!    up!   up
```
up - up, do - down, ! - flush to database.

The only concerning case seems to be:
```
up   cached      up     up    up!
up   cached      up!    up!   up!
```

It seems to be that the result of these will end up with a different uptime count.

cmd/satellite/usage.go
isaachess: I implemented your suggestion, with a method called `WithUpdateCache(updateCache UpdateCache) DB`. This solves two problems: allows me to not change the signature of the `OverlayCache()` method, and allows me to reuse the same cache between multiple `OverlayCache` calls, both of which were desirable.

satellite/satellitedb/database.go
isaachess: This is now fixed.

satellite/satellitedb/updatechecker.go
isaachess: Fixed.

satellite/overlay/cache.go
```diff
@@ -69,6 +69,8 @@ type DB interface {
 	UpdateNodeInfo(ctx context.Context, node storj.NodeID, nodeInfo *pb.InfoResponse) (stats *NodeDossier, err error)
 	// UpdateUptime updates a single storagenode's uptime stats.
 	UpdateUptime(ctx context.Context, nodeID storj.NodeID, isUp bool, lambda, weight, uptimeDQ float64) (stats *NodeStats, err error)
+	// WithUpdateCache allows you to add an update cache to the DB
+	WithUpdateCache(updateCache *UpdateCache) DB
```
egonelbre: Any reason to flip the dependency, such that UpdateCache calls into overlay.Cache rather than the other way around?

isaachess: Can you explain more? I'm not sure what you mean here.

egonelbre: `UpdateCache` could also be written as a:

```
type CoalescedCache struct {
    overlay.Cache
    ...
}

func (cache *CoalescedCache) UpdateUptime(ctx context.Context ....) error {
```

Just wondering whether the resulting code would end up clearer.

isaachess: I can't telling if you're "wondering" in a Socratic sense and are trying to help me grasp what you already see (well, maybe more "rhetorical" than Socratic, as I'm sure Socrates would deny he actually saw anything), if you're genuinely curious of my opinion. :smile: 

Personally I think the current solution, after your other great suggestions, is fairly clean code-wise. But if you'd prefer I rework it I can do that.

egonelbre: Currently, I'm really wondering :D. My gut feeling tells me that this would keep the overlay.Cache itself unmodified, and keeping the caching logic outside of the direct implementation there's less guessing in which path code takes. But there also might be some parts that I'm forgetting and it could end up more complicated.

isaachess: I went ahead and implemented this, thanks for the suggestion.

satellite/peer.go
```diff
@@ -268,10 +268,12 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 		}
 	}
 
+	overlayCache := overlay.NewInternalDBCache(peer.DB.OverlayCache(), config.Overlay.UptimeFlushInterval)
```
egonelbre: Add as peer.Overlay.XYZ.

satellite/overlay/internaldbcache.go
```diff
@@ -0,0 +1,141 @@
...
+type uptimeInfo struct {
+	isUp       bool
+	lastUptime time.Time
+	stats      *NodeStats
+}
+
+// InternalDBCache is a simple caching mechanism for overlaycache updates. It
+// provdes methods to help reduce calls to UpdateAddress and UpdateTime, but can
+// be extended for other calls in the future.
+type InternalDBCache struct {
```
egonelbre: Having `Cache` and `InternalDBCache` will be confusing.

satellite/peer.go
VinozzZ: Do you mind to explain why we should add as above?

satellite/overlay/internaldbcache.go
isaachess: I renamed it to `overlay.CombinedCache` I'm hoping this passes muster.

scripts/testdata/satellite-config.yaml.lock
```diff
@@ -307,6 +307,9 @@ kademlia.operator.wallet: ""
 # number of update requests to process per transaction
 # overlay.update-stats-batch-size: 100
 
+# the length of time we will cache uptime status before updating the overlay cache again
+# overlay.uptime-flush-interval: 1m0s
```
jenlij: how did we decide on 1 min?

isaachess: Mostly as a guess, to be honest. I assumed that it would be fairly easy to change if we need to tweak it. The real question is how long we are ok going without updating reputation metrics. During this internal, if the uptime changes, we will not change the reputation of the node. So it depends on what interval we're comfortable waiting.

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,141 @@
...
+
+type uptimeInfo struct {
+	isUp       bool
+	lastUptime time.Time
+	stats      *NodeStats
+}
+
+// CombinedCache is a simple caching mechanism for overlaycache updates. It
+// provdes methods to help reduce calls to UpdateAddress and UpdateTime, but can
+// be extended for other calls in the future.
```
jenlij: great comment

jenlij:
- [X ] Does the PR describe what changes are being made?
- [X ] Does the PR describe why the changes are being made?
- [X ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
- [X ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
- [X ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
- [X ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
- [X ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
- [X ] Does any documentation need updating? --- Perhaps we could update wherever the cache is described
- [X ] Do the database access patterns make sense?

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+// UpdateUptime overrides the underlying db.UpdateUptime and provides a simple
+// caching layer to reduce calls to the underlying db.
+func (c *CombinedCache) UpdateUptime(ctx context.Context, nodeID storj.NodeID, isUp bool, lambda, weight, uptimeDQ float64) (stats *NodeStats, err error) {
+	// First check the internal cache. If it returns stats, use them.
+	stats = c.GetNodeStats(nodeID, isUp)
+	if stats != nil {
+		return stats, nil
```
thepaul: I'm still worrying about potential harm from throwing away the new information here. With the underlying overlayCache as it is, we definitely want to pass on the uptime to `c.DB.UpdateUptime()` so it can adjust the node stats, which includes a whole bunch of things other than last contact success and last contact failure. I know team green is changing all of how that works, but I expect we'll still want as much data about when a node was up and when it was down as we can get.

It is a valid option just to merge what we have here and then change it later if needed, since we don't really know exactly what team green plans, but if we do that I recommend adding to the docstring here to make it very explicitly clear that we will throw away the values of `lambda`, `weight`, and `uptimeDQ` if there was already an update recently.

satellite/satellitedb/updatechecker.go
thepaul: That won't work right in SQLite. SQLite does preserve the time zone when it converts the value to a string, but then if you try to compare other timestamps against it you won't quite get the right thing (because it's just doing string comparisons). If this isn't causing problems in tests yet, it probably will at some point. You may want to fix both locations.

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+	// Refresh internal stats
+	c.SetNodeStats(nodeID, isUp, stats)
+	return stats, nil
+}
+
+// UpdateAddress overrides the underlying db.UpdateAddress and provides a simple
+// caching layer to reduce calls to the underlying db.
+func (c *CombinedCache) UpdateAddress(ctx context.Context, info *pb.Node, defaults NodeSelectionConfig) (err error) {
```
thepaul: There seems to be a race condition in calls to UpdateAddress- observations that we make about a node might end up getting passed to UpdateAddress in a different order from the order they happened in. Much of the time that won't matter, but I haven't been able to prove to myself that it will _never_ matter.

Picture this scenario-

* Node N contacts satellite asking for pingback at address A1
* Satellite goroutine G1 gets the request, pings node N, makes a successful connection to A1
* Node N switches to using a different interface with a different IP
* Node N contacts satellite asking for pingback at address A2
* Satellite goroutine G2 gets the request, pings node N, makes a successful connection to A2
* Goroutine G2 calls UpdateAddress, gets the addressCache lock, writes address A2 to cache and db
* Goroutine G1 calls UpdateAddress, gets the addressCache lock, writes address A1 to cache and db

So the addressCache and the db would not actually have the latest contacted IP. Your class invariants would be violated. Does that matter even then? I'm not sure. Maybe not. But if a race condition like this exists, and there is a way to break the described behavior of a software component, it _will_ totally happen because computers are the worst and they hate us. Also it sounds possible for malicious node operators to cause this situation on purpose fairly reliably.

If you're sure that all of that is harmless and acceptable, then that's ok too, but we just need to change some comments and docstrings and possibly method names to make it clear that addressCache might not have the absolute latest address and IP, and the db might not have the absolute latest address and IP, and that instead it's a "best effort" situation in both cases.

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	address := node.Address
+	if address == nil {
+		address = &pb.NodeAddress{}
+	}
+
+	c.addressLock.RLock()
+	cached, ok := c.addressCache[node.Id]
+	c.addressLock.RUnlock()
```
thepaul: This gap between RUnlock and Lock is another race, wherein another goroutine could change what's in the map and cause results to be applied out of order. If the race I mentioned above is deemed not a problem, then maybe this one isn't either.

Still, this one has an easier fix: just use Mutex instead of RWMutex, and you won't have to reacquire the lock in order to do a write. I know how attractive read-write mutexes are, believe me, but every time I use them it seems to end up being overcomplication or premature optimization. Read-write mutexes are quite a bit slower than plain mutexes—with a good implementation it may be on the order of 3x as long to do a `(*RWMutex).RLock()` than a `(*Mutex).Lock()`. And that is still fine if there are lots of readers and lots of lock contention and read locks tend to be held for a nontrivial amount of time, because the additional work allows many more readers to progress in parallel. But in this case, there is only a single map lookup going on with the read-lock held. I suspect that the map lookup might well take less time than the RLock or the RUnlock next to it. So if there is so much lock contention that an RWMutex is justified, then there is a very very large number of calls going on to RLock and RUnlock, and the cpu will end up taking a lot more total time in locking and unlocking than it does in moving the program forward.

satellite/overlay/combinedcache.go
isaachess: I believe I have fixed this with my latest update.

satellite/overlay/combinedcache.go
isaachess: I believe this is also fixed in my newest implementation.

I think in this case the RW Mutex is worthwhile, since in the majority of cases the UpdateAddress calls should end up being returned from the cache, making an actual Lock call unnecessary. But let me know if you disagree after seeing the final code update.

internal/sync2/keylock.go
```diff
@@ -0,0 +1,65 @@
...
+// contention only exists for locking calls using the same key parameter. As
+// with all locks, do not call Unlock() or RUnlock() before a corresponding
+// Lock() or RLock() call.
+//
+// Note on memory usage: internally KeyLock lazily and atomically creates a
+// separate sync.RWMutex for each key. To maintain synchronization guarantees,
+// these interal mutexes are not freed until the entire KeyLock instance is
+// freed.
+type KeyLock struct {
+	locks sync.Map
```
egonelbre: I'm not quite sure whether using sync.Map here is the best approach here. I think it fits the regular optimization criteria, but I'm not certain. Using a regular map would allow using `storj.NodeID` as the key, rather than interface{}.

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,86 @@
...
+	transport pb.NodeTransport
+}
+
+// CombinedCache is a simple caching mechanism for overlaycache updates. It
+// provdes methods to help reduce calls to UpdateAddress and UpdateTime, but can
+// be extended for other calls in the future.
+type CombinedCache struct {
+	DB
+	addressLock  sync.RWMutex
+	addressCache map[storj.NodeID]*addressInfo
```
egonelbre: any reason not to add the lock directly into addressInfo?

internal/sync2/keylock.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// RLock the provided key
+func (l *KeyLock) RLock(key interface{}) {
+	lock := l.getNewLock(key)
+	lock.RLock()
+}
+
+// RUnlock the provided key
+func (l *KeyLock) RUnlock(key interface{}) {
```
egonelbre: When you return the unlocking func from the locking then you can avoid the double lookup.

internal/sync2/keylock.go
isaachess: From the sync.map documentation:

```The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow```

This fits that case.

I chose to attempt a general solution here, which is why I am not using `storj.NodeID` explicitly as the key.

File: satellite/overlay/combinedcache.go (hunk 18 above)
Comment (thepaul) [reply satellite/overlay/combinedcache.go]: I probably haven’t explained very well. The RLock() operation itself is the slow thing. It has to perform more than one synchronization primitive under the covers in order to enforce the read/write semantics (I don’t know exactly which approach Go takes here, but I don’t know of any way to do it without multiple steps). Calling Lock() on a plain Mutex should be significantly faster. RWMutex generally probably isn’t worth it unless readers tend to hold their read locks for nontrivial amounts of time, so that it helps for them to be able to overlap ownership. Of course, I could be wrong about all of this and maybe Go will surprise me again, so if you have timing data that says RWMutex and RLock is better, then great! But I’m sufficiently sure about it that I think it’s worth checking.

File: internal/sync2/keylock.go (hunk 19 above)
Comment (egonelbre) [reply internal/sync2/keylock.go]: Yeah, but using sync.Map happens together with using interfaces, which may negate the performance benefit and/or end up using more memory.

File: satellite/overlay/combinedcache.go (hunk 17 above)
Comment (thepaul) [reply satellite/overlay/combinedcache.go]: To really avoid the race between UpdateUptime calls with time-separated connections to the same node, I want to suggest using a channel and a dedicated goroutine to govern writes into the cache. Depending on how reads get managed, you might be able to do away with all the mutexes and locks in this code altogether.

File: internal/sync2/keylock.go (hunk 21 above)
Comment (isaachess) [reply internal/sync2/keylock.go]: Done.

File: satellite/overlay/combinedcache.go [hunk 22]
```diff
@@ -0,0 +1,86 @@
...
+	if ok &&
+		address.Address == cached.address &&
+		address.Transport == cached.transport &&
+		info.LastIp == cached.lastIP {
+
+		return nil
+	}
+
+	// Acquire lock for this node ID. This prevents a concurrent db update to
+	// this same node ID and guarantees the cache and database stay in sync
```
egonelbre: This wouldn't work with separate processes.

satellite/overlay/combinedcache.go
isaachess: I personally feel the code is simpler with the lock existing in a separate abstraction. It's mainly handling the lazy-creation of the nodeID-based locks that would complicated the code here.

satellite/overlay/combinedcache.go
isaachess: We discussed this in person, just recording a few thoughts here. As we discussed, it appears this race already exists -- that is, a badly-behaving node can already request two ping-backs at different addresses such that we can't guarantee the order in which they arrive int he database. My update should not change this behavior, but should guarantee that the db and cache remain in sync (which I believe it does).

It does appear, as we discussed, that these calls all happen synchronously. I can add a monotonic clock check for each UpdateAddress request and toss out any request that is older than a previous request we've seen, though I still believe that this should be handled in a separate ticket and seems outside the scope of my ticket. :smile_cat:

satellite/overlay/combinedcache.go
isaachess: As we discussed in slack some, I ran some benchmark tests that suggested that, while RLock may be slightly slower than a regular Lock, the difference is a few nanoseconds. By keeping the RLock, we optimize the case (which I hope is the common case) where eventually we have cached most node addresses, and the cache (with an RLock) is able to concurrently handle almost all `UpdateAddress` calls.

satellite/overlay/combinedcache.go
isaachess: Done.

satellite/overlay/combinedcache.go
zeebo: Random seagull flyby knowledge poop: in the case of no pending writers, an `RLock` is just an atomic add. If there is a pending writer, it’s just that add and a semaphore wait. It’s implemented by a counter of pending read locks that the `Lock` call atomically “negates” so that readers can semaphore.

The main performance problem with an rw-mutex comes from 1. too many write locks, 2. acquiring a read lock still is a write to a shared bit of memory, and so doesn’t scale very well to high core counts. It’s bottlenecked by processor cache level coordination.

satellite/overlay/combinedcache.go
zeebo: Also, it’s possible addressInfo has the appropriate granularity. There may not yet be an addressInfo to hold the mutex, for example. It also makes updating the addressInfo harder as you can’t just overwrite the map key, possibly leading to subtle races.

satellite/overlay/combinedcache.go
thepaul: Well, the atomic add is still being used there as a synchronization primitive, with its own associated costs (at least one forced cache miss, possibly more, plus a memory fence, plus potential retries if it’s implemented with Intel CAS).

I’ve timed the Go implementation before, under moderate contention, and the 3x cost I’m used to from pthreads world was still about the same. Although, I don’t remember if that was on ARM or x86_64.

satellite/overlay/combinedcache.go
thepaul: And again, I am fine with leaving the race as it is, and filing a ticket or another PR or whatever to address later, but until that happens it is imperative that the weaker semantics be documented. The class can not claim to cache the latest address always- it can just say “probably the latest” or “best effort” or something like that. Our components just need to be able to rely on each other absolutely for us to build a robust and secure distributed system.

isaachess: I think I understand better now. When I say "latest address always" I'm referring to the "last address updated in the database." I see what you mean by the fact that it may not actually be the latest in a real-world-time sense. I will definitely document that, thank you for pointing it out and helping me understand what you mean. Sorry that I've been a little slow on the uptake there to fully understand.

internal/sync2/keylock.go
isaachess: I have changed the implementation to use a direct map with a sync.Mutex and placed it in the overlay package. Performance was slightly better, and Paul was right that there wasn't a clear need for a generalized solution.

satellite/overlay/keylock_test.go
```diff
@@ -0,0 +1,46 @@
...
+	unlockFunc()
+}
+
+func BenchmarkKeyLock(b *testing.B) {
+	b.ReportAllocs()
+	ml := overlay.NewKeyLock()
+	numNodes := 100000
+	nodeIDs := make([]storj.NodeID, numNodes)
+	for i := 0; i < numNodes; i++ {
+		nodeIDs[i] = teststorj.NodeIDFromString(strconv.Itoa(i))
```
egonelbre: Use `testrand.NodeID()` rather than a sequence of items.

satellite/overlay/keylock_test.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	b.Run("lock", func(b *testing.B) {
```
egonelbre: Separate the measurement of the first lock and relocking.

satellite/overlay/combinedcache.go
egonelbre: You should still be able to overwrite the map key.

zeebo: Maybe I misunderstand the suggestion and how the logic would work with the mutex on the addressInfo itself, then. Specifically, I don’t see how
1. You could avoid a lock around mutating the map any better
2. When you would want to acquire the mutex on the addressInfo

egonelbre: I meant something like this https://github.com/storj/storj/pull/2716/files, since we are racing with multiple processes anyways.

egonelbre: And if you want to have better ordering you could add a information version number.

zeebo: I think that’s closer to what the original PR did, and I objected (sorry, privately) because it makes it so that the cache can become out of sync with the database. While that’s apparently not a problem, the fact that it’s not a problem is not necessarily obvious. One must understand subtle whole system details to prove to themselves that it’s ok, and I can imagine future changes that lose that property. Additionally, every reader of the code in the future hits that problem. Indeed, I think it has been discussed multiple times even in this PR.

For those reasons, I strongly prefer if there is synchronization to prevent the cache from having different state from the database.

satellite/overlay/keylock_test.go
isaachess: I was intentionally trying to simulate the case where it has to lazily create the locks, but then is able to re-use existing locks. That is why I didn't create a random nodeID for each test iteration. Is there a better way to do this? Or should I not worry about simulating that?

satellite/overlay/keylock_test.go
isaachess: I'm not sure what you mean here, can you explain further?

satellite/overlay/keylock_test.go
egonelbre: I just mean having the node-ids themselves be random rather than `teststorj.NodeIDFromString`. We have an `internal/testrand` package that can generate random node id. The slice of id-s is fine.

satellite/overlay/keylock_test.go
egonelbre: I suspect there's a performance difference for adding a new node and locking/unlocking an already existing node.

Most of the time we will be locking/unlocking existing node-ids rather than adding, so we care about the latter number more. It's nicer if these benchmarks show up separately.

satellite/overlay/combinedcache.go
isaachess: @egonelbre yes your PR is basically exactly what I had in the first iteration of this. The reason I have changed it is because there is a risk that the cached data is out of sync with the underlying DB. As Jeff says, it is difficult to reason about precisely where this could cause problems (or not), and this is why I introduced the `KeyLock` concept to guarantee that the cache and the db are always in sync.

satellite/overlay/keylock_test.go
isaachess: I made that change. Just FYI, here's the results:

```
BenchmarkKeyLock/lock_all_new_nodes-8            2000000               805 ns/op
BenchmarkKeyLock/lock_existing_nodes-8          10000000               177 ns/op
BenchmarkKeyLock/rlock_existing_nodes-8         10000000               166 ns/op
```

egonelbre: nice :)

satellite/overlay/combinedcache.go
egonelbre: Ok, let's keep keylock.

satellite/overlay/keylock.go
```diff
@@ -0,0 +1,64 @@
...
+// KeyLock provides per-key RW locking. Locking is key-specific, meaning lock
+// contention only exists for locking calls using the same key parameter. As
+// with all locks, do not call Unlock() or RUnlock() before a corresponding
+// Lock() or RLock() call.
+//
+// Note on memory usage: internally KeyLock lazily and atomically creates a
+// separate sync.RWMutex for each key. To maintain synchronization guarantees,
+// these interal mutexes are not freed until the entire KeyLock instance is
+// freed.
+type KeyLock struct {
```
egonelbre: I wonder whether there's a nicer name for this. Something like `NodeLocker` or something :D?

satellite/overlay/combinedcache.go
```diff
@@ -0,0 +1,90 @@
...
+		info.LastIp == cached.lastIP {
+
+		return nil
+	}
+
+	// Acquire lock for this node ID. This prevents a concurrent db update to
+	// this same node ID and guarantees the cache and database stay in sync.
+	// This solution works so long as calls to this code are occurring within a
+	// single process.
+	unlockFunc := c.keyLock.Lock(info.Id)
```
egonelbre: The `func` part in `unlockFunc` is slightly redundant. `unlock` would suffice...

Also this could be written as `defer c.keyLock.Lock(info.Id)()`.

PR #2640, opened 2019-07-26
uplink: integrate new Metainfo calls

mniewrzal:
What: Client side is adjusted to use new Metainfo methods.

Why: This is part of Metainfo refactoring. 
https://storjlabs.atlassian.net/browse/V3-2060
https://storjlabs.atlassian.net/browse/V3-2061

Please describe the tests:
 - Test 1: current tests should pass without significant changes
 - Test 2: some tests were removed because make no sense with the current design
 
Please describe the performance impact: TBD

## Code Review Checklist (to be filled out by reviewer)
 - [x] Does the PR describe what changes are being made?
 - [x] Does the PR describe why the changes are being made?
 - [x] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [x] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [x] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [x] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [x] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [x] Does any documentation need updating?
 - [x] Do the database access patterns make sense?

PR #2682, opened 2019-08-01
Integrate new Metainfo calls (server side)

mniewrzal:
What: New Metainfo API integration but only server side is included in this PR. 

I divided my previous PR #2640 because this PR can be merged without concerns that it will break current upload/download flow. When this part will be deployed we can test if second part of #2640 is not breaking or decreasing performance.

Why: This is part of Metainfo refactoring.

https://storjlabs.atlassian.net/browse/V3-2060
https://storjlabs.atlassian.net/browse/V3-2061

Please describe the tests:
 - Test 1: added 


mniewrzal:
@ifraixedes @egonelbre I made this PR from #2640 because of reasons from description. Also, this will be smaller and easier to review.

pkg/pb/metainfo.proto
```diff
@@ -401,12 +401,12 @@ message SegmentBeginResponse {
 message SegmentCommitRequest {
     bytes segment_id = 1 [(gogoproto.customtype) = "SegmentID", (gogoproto.nullable) = false];
 
-    bytes encrypted_key_nonce = 2 [(gogoproto.customtype) = "Nonce", (gogoproto.nullable) = false];
-    bytes encrypted_key = 3;
+    bytes encrypted_key_nonce = 3 [(gogoproto.customtype) = "Nonce", (gogoproto.nullable) = false];
```
egonelbre: why the number change here?

mniewrzal: oh, I moved `position` back to `SegmentBeginRequest` and I forgot to adjust numbers

pkg/storj/segment.go
```diff
@@ -17,6 +17,16 @@ type SegmentListItem struct {
 // SegmentDownloadInfo represents segment download information inline/remote
 type SegmentDownloadInfo struct {
```
egonelbre: I'm not quite sure how much this makes sense since you also need the addressed order limits for this to work... but I guess it's fine for now.

mniewrzal: Because addressed order limits are from `pb` I'm returning them as a separate value in `client.DownloadSegment`

uplink/metainfo/client.go
```diff
@@ -685,6 +734,12 @@ func (client *Client) DownloadSegment(ctx context.Context, params DownloadSegmen
 		}
 	}
 
+	for i := range response.AddressedLimits {
```
egonelbre: why is it clearing the limits?

mniewrzal: because we don't have pointer on client side anymore when downloading I need to associate `order limit` with `piece num` and because protbuf is not allowing nils on top level I'm sending empty addressed limit which is converted to nil on client side before it will be returned. Earlier we were sorting limits with `sortLimits` method.

PR #2683, opened 2019-08-01
 Make one implementation for SetAttribution/SetBucketAttribution

mniewrzal:
What: We are moving to new Metainfo API and attribution method has a different name. For some time we will keep both methods and that's why it would be good to have one implementation for both.

Why: 

Please describe the tests:
 - Test 1: no changes in existing tests
 

satellite/metainfo/metainfo.go
```diff
@@ -829,8 +781,68 @@ func getAllowedBuckets(ctx context.Context, action macaroon.Action) (_ macaroon.
...
-	return resp, status.Error(codes.Unimplemented, "not implemented")
+func (endpoint *Endpoint) SetBucketAttribution(ctx context.Context, req *pb.BucketSetAttributionRequest) (resp *pb.BucketSetAttributionResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	err = endpoint.setBucketAttribution(ctx, req.Name, req.AttributionId)
+
+	return &pb.BucketSetAttributionResponse{}, err
+}
+
+func (endpoint *Endpoint) setBucketAttribution(ctx context.Context, bucketName []byte, parterID []byte) error {
```
aligeti: curious why have this 'setBucketAttribution'. I believe the other function 'Old' will be removed. So can we make one function with 'SetBucketAttribution`.

mniewrzal: It will be removed but I'm not sure when. For now we need to keep `Old` version to be backward compatible.

PR #2686, opened 2019-08-01
docs/design: Kademlia Removal Design Proposal

jenlij:
What & Why: This design document outlines the communication protocol between satellites and 
storage nodes, network refreshes, and kademlia removal given the satellite opt-in 
capability for storage nodes. 
https://storj.io/blog/2019/08/so-youre-a-storage-node-operator.-which-satellites-do-you-trust/


docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+2. Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their 
+    trusted list every hour. 
+    - Satellites will ping the nodes back
+        - If is it successful, the satellite will insert or update the node in the overlay cache and 
+        notify the node of success
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
```
wthorp: A response from the server after updating the cache would mean a second communication from the server back to storage node (third overall).  Do you think there's value in rolling this feature into the initial ping of the satellite from storage node, where that ping would return a 'last status'?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted 
+    satellites list.
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+2. Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their 
+    trusted list every hour. 
+    - Satellites will ping the nodes back
```
wthorp: Clarify that the initial connection from the storage node stays open, and that status responses to the storage node are returned via the initial connection.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+        - If is it successful, the satellite will insert or update the node in the overlay cache and 
+        notify the node of success
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
+    - This will use grpc for now, but future development may use a non-ssl connection from node to satellite 
+    to initiate the refresh connection
+    - Update the overlay cache refresh method to communicate with nodes directly, without relying on 
+    kademlia lookups. Iterate through cache and dial each node.
```
wthorp: I'm still skeptical that we need the server to reach out to the storage nodes.  I feel at minimum we need a sentence that justifies keeping this functionality.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+
+3. Disintegrate Kademlia from the network, storj sim and testplanet setups
+    - Remove the “discovery” package
```
wthorp: Also remove transport/observer I think.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
+# Kademlia Removal
+
+## Abstract
+
+This design document outlines the communication protocol between satellites and 
+storage nodes, network refreshes, and kademlia removal given the satellite opt-in 
+capability for storage nodes. 
+
+## Background
+User research on how Storage Node Operators would connect to satellites resulted in a
```
egonelbre: Whitespace above.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Abstract
+
+This design document outlines the communication protocol between satellites and 
+storage nodes, network refreshes, and kademlia removal given the satellite opt-in 
+capability for storage nodes. 
+
+## Background
+User research on how Storage Node Operators would connect to satellites resulted in a 
+simple solution called Opt-in SNO select. Satellites wait for storage nodes to work
```
egonelbre: Storage Nodes select which Satellites to work with, hence the Storage Nodes do not need to discover Satellites on the network and Satellites have no benefit from finding Storage Nodes on the network.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+
+This design document outlines the communication protocol between satellites and 
+storage nodes, network refreshes, and kademlia removal given the satellite opt-in 
+capability for storage nodes. 
+
+## Background
+User research on how Storage Node Operators would connect to satellites resulted in a 
+simple solution called Opt-in SNO select. Satellites wait for storage nodes to work 
+with them, and storage nodes manually select the satellites they want to work with. 
+The initial implementation of this allows SNOs to update their trusted satellite list
```
egonelbre: Separate paragraph maybe?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Background
+User research on how Storage Node Operators would connect to satellites resulted in a 
+simple solution called Opt-in SNO select. Satellites wait for storage nodes to work 
+with them, and storage nodes manually select the satellites they want to work with. 
+The initial implementation of this allows SNOs to update their trusted satellite list 
+in a hardcoded configuration file, but future improvements will enable users to manage 
+this list through a web console. Due to the nature of this change, the Storj network 
+at this point in time no longer requires the Kademlia DHT or related entities. 
+We will replace this protocol with direct communication between satellites and
```
egonelbre: Separate paragraph.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+at this point in time no longer requires the Kademlia DHT or related entities. 
+We will replace this protocol with direct communication between satellites and 
+storage nodes, and keep the network fresh without kademlia node discovery and random 
+lookups.
+
+## Design
+
+1. Nodes reach out to satellites that they want to work with
+    
+    - At the time of writing, a storage node operator can input a list of trusted satellite
```
egonelbre: Can we give this information in Background?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+in a hardcoded configuration file, but future improvements will enable users to manage 
+this list through a web console. Due to the nature of this change, the Storj network 
+at this point in time no longer requires the Kademlia DHT or related entities. 
+We will replace this protocol with direct communication between satellites and 
+storage nodes, and keep the network fresh without kademlia node discovery and random 
+lookups.
+
+## Design
+
+1. Nodes reach out to satellites that they want to work with
```
egonelbre: Can we separate these into subheadings rather than the counted list?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    - Remove the “discovery” package
+    - Remove the bootstrap node - work with Ops
+    - Work with QA to make sure storage nodes don’t crash on errors related to the elimination of Kademlia 
+    if they don’t update immediately
+
+4.Update whitepaper to address kademlia removal and the addition of satellite opt-in
+    - Remove the audit gating design doc
+    - Update the wiki
+
+## Rationale
```
egonelbre: The rationale is about why this exact design... the reason for removal of kademlia is already described above.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    IDs and addresses into their configuration file on setup. There are several tardigrade-level 
+    satellites already listed as a preset. This list is imported into a Go map data structure 
+    within the “trust” package.
+    - Eventually during the design and implementation of trusted satellite management, 
+    this configured list may be imported into a on-disk data store, eg SQL table. Users will 
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted 
+    satellites list.
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
```
wthorp: It might be worth having a section on NodeIDs themselves.  Without Kad, the value proposition of NodeIDs changes.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+2. [Nodes keep up to date with satellites by periodically pinging and updating their trusted list](https://storjlabs.atlassian.net/browse/V3-2275)
+
+3. [Remove kad random lookups for discovery & update overlay cache refresh](https://storjlabs.atlassian.net/browse/V3-2305])
+
+4. [Disintegrate kademlia from the network, storj sim and testplanet setups](https://storjlabs.atlassian.net/browse/V3-2276)
+
+5. Update Documentation
+
+## Open issues (if applicable)
+
```
littleskunk: How about the storage nodeID? It is generated with a minimum requirement and signed by our CA server to make an sybil attack expensive. On the other side the sybil attack is kind of useless if the incentive (kademlia) gets removed. What now?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+2. Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their 
+    trusted list every hour. 
+    - Satellites will ping the nodes back
+        - If is it successful, the satellite will insert or update the node in the overlay cache and 
+        notify the node of success
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
+    - This will use grpc for now, but future development may use a non-ssl connection from node to satellite
```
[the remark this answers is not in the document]
jtolds: SSL also has a significant resource and performance impact. any place we can avoid SSL will make us faster. in particular, this specific place is safe to not have over SSL. it's basically an anonymous tip box for satellites to find out about new nodes. satellites must still validate that they can talk to the node over ssl, but the tip box part does not need to be over ssl

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
+    - This will use grpc for now, but future development may use a non-ssl connection from node to satellite 
+    to initiate the refresh connection
+    - Update the overlay cache refresh method to communicate with nodes directly, without relying on 
+    kademlia lookups. Iterate through cache and dial each node.
+
+3. Disintegrate Kademlia from the network, storj sim and testplanet setups
+    - Remove the “discovery” package
+    - Remove the bootstrap node - work with Ops
+    - Work with QA to make sure storage nodes don’t crash on errors related to the elimination of Kademlia
```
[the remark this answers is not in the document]
jtolds: a storage node that upgrades straight to C will just appear offline to A. that's fine, right? I think we can skip B.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,84 @@
...
+    - Eventually during the design and implementation of trusted satellite management, 
+    this configured list may be imported into a on-disk data store, eg SQL table. Users will 
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted 
+    satellites list.
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+2. Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their
```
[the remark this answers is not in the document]
jtolds: that's why it makes sense to avoid SSL earlier. nonetheless, with kademlia satellites are doing this much work if not more

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: > it's basically an anonymous tip box for satellites to find out about new nodes
Can I claim that you IP has changed? Without ssl I don't need a private key to send that message. The ssl connection is preventing this very good.

> satellites must still validate that they can talk to the node over ssl
Why do we need 2 steps for that? Can't we add/update and validate the information in one step?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: The question is what happens with the storage nodes that are not upgrading? The satellite is pining the nodes once per hour. That is a kademlia ping. -> Upgraded satellite will be unable to talk to not upgraded storage nodes. -> Backwards incompatible.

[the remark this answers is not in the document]
egonelbre: Why wouldn't it be able to talk to them? We have a method to fetch peer identity from a node without having a special endpoint.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: The satellite can loadbalance the pings. With 1M storage nodes the satellite might take a long time for one ping itteration. We might get issues because the interval is too high but the satellite will not get DDoS.

If we turn that arround that means all storage nodes will ping the satellite. If we add a nice little bug that all storage nodes are crashing at the 29th feb they will restart all at the same time. They will take the satellite offline and because they failed to contact the satellite they will retry. How do we protect the satellite in that situation?

docs/design/kademlia-removal.md
wthorp: Yeah, mine was a bad idea.  The connection should probably be closed.  The storage node will need to create a timer as it pings the satellite.  If the satellite doesn't ping it back, it should error and/or alert.

docs/design/kademlia-removal.md
wthorp: NodeIDs may still be useful as a public key check to prevent MITM attacks.  See ParseNodeURL().

docs/design/kademlia-removal.md
cam-a: I think the rationale was that if the satellite only relies on the nodes pinging _it_, then a node could wake up once every hour to contact the satellite, then go offline. However, they're going to fail audits if they do that, so I'm not sure that's a viable strategy anyway.

There must be a reason that I'm forgetting; Couldn't the cache be updated when SNs contact the satellite every hour?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: Because from the satellite point of view the nodes are offline because they can't be pinged.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
cam-a: In this scenario, is the SN going to use a non ssl connection for every uptime ping or just the initial contact?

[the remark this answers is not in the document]
jtolds: > Can I claim that you IP has changed? Without ssl I don't need a private key to send that message. The ssl connection is preventing this very good.

Sure, but the satellite won't trust that until it connects to the storage node to validate.

> Why do we need 2 connections for that? Can't we add/update and validate the information with one ssl connection?

We need two connections because the Satellite needs to confirm that the given ip/port will actually reach the node in question. The IP/port provided could be bad. We need to test with an *outgoing* connection from the satellite.

> In this scenario, is the SN going to use a non ssl connection for every uptime ping or just the initial contact?

 just the initial suggestion to the satellite of where to find the storage node

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: we should still have a ping method.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: they won't take the satellite offline if they just send single udp packets. the satellite can ignore them. the satellite will only go down if it chooses to engage

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
egonelbre: Why do we need a ping? When we can establish a grpc connection that should be sufficient to get the same information.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
cam-a: What if the packet is dropped and the SN's uptime suffers?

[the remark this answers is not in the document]
jtolds: yeah that would stink. i think team green is planning on some kind of reaper service that checks in with storage nodes we haven't heard from recently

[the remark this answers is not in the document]
cam-a: If we're using udp, are we going to use DTLS?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: How about a man in the middle attack? I tell the satellite that the ip and port for your node has changed. The satellite will try to contact me and I will redirect that to your real ip and port.
The connection is encrypted? Which keypair is used for the encryption? Can I get a copy of that keypair and read all the packages? Even if that is not possible I could wait for the initial handshake and than take your node offline.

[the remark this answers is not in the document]
jtolds: Okay, you're right, the UDP messages need to be signed by the node's identity. Good catch.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: littleskunk and i realized on a call today that if we keep just the overlay.Ping rpc method, it will be much easier for a new satellite to work with old and new storage nodes.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: no, but littleskunk points out the udp message should at least be signed. let's avoid DTLS though

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
littleskunk: If a storage node is offline we will not select it for uploads and downloads. If the satellite is updated but the storage nodes are not updated and the satellite can't ping the storage nodes, all downloads should fail.

docs/design/kademlia-removal.md
wthorp: Clarifying, I understand we need to make sure the node is reachable by the server.  But, I feel like "Update the overlay cache refresh method" implies that this is happening on an interval and initiated by the server.  Still not sure about that part.  Maybe a process which reaches out to SNs not contacted for other reasons in the last hour.  That would add some randomness to the timings.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
wthorp: I thought that this was basically the reaper service!

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted 
+    satellites list.
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+### Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their 
+    trusted list every hour. 
+    - Satellites will ping the nodes back
```
wthorp: Can we specify that this is rate limited.  Like if we haven't heard from them in X amount of time?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
wthorp: Should explicitly state that they do this on startup (to reduce the amount of time they appear offline)?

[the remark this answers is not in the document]
jtolds: that's a good idea, yeah

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Storage Nodes reach out to satellites that they want to work with
+    - Eventually during the design and implementation of trusted satellite management, 
+    this configured list may be imported into a on-disk data store, eg SQL table. Users will 
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted 
+    satellites list.
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
```
egonelbre: How do storage nodes end up in the overlay cache?

docs/design/kademlia-removal.md
jenlij: Could you elaborate on what you mean here @wthorp ? I suppose we may not need the node dossier any longer either since we won't be exposing node info to other nodes on the network

docs/design/kademlia-removal.md
jenlij: they're inserted by the transport observer

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jenlij: @wthorp @jtolds what do I need to add here?

docs/design/kademlia-removal.md
jenlij: If the satellites haven't heard from the nodes in x amount of time?

docs/design/kademlia-removal.md
egonelbre: Yes, but when? When a storage node makes the initial query against satellite, we shouldn't add them immediately to overlay, since that would cause problems. This means there needs to be a process when they get added.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: 1) storage nodes should notify satellites when they start up, wait a random amount of time (to add jitter http://highscalability.com/blog/2012/4/17/youtube-strategy-adding-jitter-isnt-a-bug.html), then start reporting in roughly on the hour
2) when they report in to the satellite's new-node-notification tipbox, it can be over ssl, but it doesn't have to be. if it isn't over ssl it should at least be signed.
3) there should be a reaper service that checks out what's going on with nodes that haven't reported in in a while

docs/design/kademlia-removal.md
jtolds: i think he's saying we should reconsider if everything about node ids is still necessary (such as difficulty, construction based on certificate authority public key, etc).

that said, i think that's worth another design document. i also don't thiiiiink we'll want to make any node id changes anyways. the difficulty and the construction are still useful for other reasons.

docs/design/kademlia-removal.md
jtolds: the storage nodes reach out to the satellite when they start up and then once an hour, telling the satellite about themselves. in response, the satellite will verify and confirm that the storage node is who it says it is at the address it says it has. that second communication will insert nodes into the overlay cache (either via the transport observer or directly, doesn't matter)

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: that's just my summary of the comment thread, not where it should go in the design doc

docs/design/kademlia-removal.md
jtolds: satellites should ping the node back if the node has asked the satellite to ping it, however often the node asks

docs/design/kademlia-removal.md
jtolds: okay, this is the reaper part - we should iterate through and only dial nodes we haven't heard from recently (in the last hour)

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+        - If is it successful, the satellite will insert or update the node in the overlay cache and 
+        notify the node of success
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
+    - This will use grpc for now, but future development may use a non-ssl connection from node to satellite 
+    to initiate the refresh connection
+    - Update the overlay cache refresh method to communicate with nodes directly, without relying on 
+    kademlia lookups. Iterate through cache and dial each node.
+
+### Disintegrate Kademlia from the network, storj sim and testplanet setups
+    - Remove the “discovery” package
```
egonelbre: discovery package also keeps nodes in overlay updated, so may want to specify that "remove kademlia from discovery package"... alternatively we would need to create a new package for refreshing nodes.

docs/design/kademlia-removal.md
egonelbre: So there is a deduplication queue somewhere for the verifications? Or a regular queue? Or a database table? Or is it kept all in memory?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. [Nodes should find satellites through the trust packages rather than kademlia](https://storjlabs.atlassian.net/browse/V3-2274)
+
+2. [Nodes keep up to date with satellites by periodically pinging and updating their trusted list](https://storjlabs.atlassian.net/browse/V3-2275)
+
+3. [Remove kad random lookups for discovery & update overlay cache refresh](https://storjlabs.atlassian.net/browse/V3-2305])
+
+4. [Disintegrate kademlia from the network, storj sim and testplanet setups](https://storjlabs.atlassian.net/browse/V3-2276)
+
+5. Update Documentation
```
egonelbre: Should we also have an issue about removing vouchers?

jenlij: Ah yes definitely

docs/design/kademlia-removal.md
jenlij: sounds good! I will make some notes about subsequent design docs

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+storage nodes, and keep the network fresh without kademlia node discovery and random 
+lookups.
+
+## Design
+
+### Storage Nodes reach out to satellites that they want to work with
+    - Eventually during the design and implementation of trusted satellite management, 
+    this configured list may be imported into a on-disk data store, eg SQL table. Users will 
+    be able to modify this list through the storage node operator console. However, this is a 
+    separate discussion from this design document. This design is satisfied with a static trusted
```
isaachess: By "static" here, I assume you mean it must be edited manually?

docs/design/kademlia-removal.md
isaachess: > i also don't thiiiiink we'll want to make any node id changes anyways

Whether we make changes or not, having a degree of difficulty in generating a valid nodeID is a useful feature, as it prevents users from flooding the network with bogus nodes just for the lulz. There are other ways (other than "proof of work") that this could be done, but POW is likely the easiest decentralized way to gatekeep noes coming onto the network.

docs/design/kademlia-removal.md
isaachess: +1 for removing the transport observer, and explicitly updating the overlay cache in the code, in those places where it makes sense to do so. In fact, if the nodes ping the satellite once/hour, this (and audits) could be the only code that maintains uptime checks.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
isaachess: Is this design doc intended to apply to just tardigrade satellites, or others? If others, I suggest that satellites can determine how often nodes are supposed to check in for uptime checks. Perhaps there is an "introduction ping" that occurs the first time the node comes online, and when the satellite responds, it includes how often nodes are expected to check back in to maintain good reputation.

docs/design/kademlia-removal.md
isaachess: > however often the node asks

Really? Why? If a satellite has a set policy that uptime checks are once/hour, it's reasonable that if a node keeps asking for uptime pings over and over again, you can ignore them. And potentially even DQ the node.

I suggested above that the satellite can tell the node its policy for how often the node should check in. Based on that, if a node is asking for ping-backs too frequently, I could see that as something to protect against.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,86 @@
...
+    - Nodes will be able to communicate with satellites through the “transport” and “trust” packages 
+    directly rather than using kademlia to traverse the network to find the address of a given ID.
+
+### Network refreshes at a regular interval
+    - Nodes will keep themselves up to date in the network by pinging all the satellites in their 
+    trusted list every hour. 
+    - Satellites will ping the nodes back
+        - If is it successful, the satellite will insert or update the node in the overlay cache and 
+        notify the node of success
+        - If not, it does not proceed with updating the overlay cache, and the node receives an error message
```
isaachess: How does the node "receive" an error message? From the satellite? If the node is using UDP, and if the satellite cannot ping the node back, I don't see how the node gets anything from the satellite.

Unless you mean that there is some type of timer on the node that times out if the pingback hasn't occurred within a period of time. If so, I think being more explicit here would be good.

jenlij: Good call... I'm wondering if we even need to tell the node anything here.

docs/design/kademlia-removal.md
jenlij: Yes, I can clarify

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: this is supposed to apply to others, but i'd like to scope configurable windows for a later feature and we don't tackle that now

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
isaachess: What about a DDOS attack here (or something similar -- I don't know all the formal names for attack vectors)? For instance, if I maliciously drop 20 million "tips" into the tip-box, the satellite now has too much work to do, it'll never get around to pinging real nodes.

It seems you should only be able to request a ping-back for yourself, and that rate-limiting it is a good idea.

docs/design/kademlia-removal.md
jtolds: okay, it's not a hard requirement that we ping back every time the node asks, but i also don't think it's an "attack" worth defending against. the storage node operator has to pay the same resources the satellite does. it would be really hard to DoS a satellite this way. a misconfigured node could be annoying though, i agree.

docs/design/kademlia-removal.md
isaachess: I understand this except for the "dial nodes" part. If they haven't pinged us, why should we ping them back? The node is violating the protocol at that point, and perhaps its reputation could suffer, but I don't see a reason to dial it explicitly ... ?

Or rather, I only really see a need for one or the other. If the node is in charge of pinging the satellite every hour, the node is responsible for keeping its "uptime" good. Or, if the satellite is in charge of pinging the node periodically, the satellite is in charge of knowing each node's uptime. But I don't see the need for both the node to ping the satellite and the satellite to ping offline nodes.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: that's a reason why they should be signed. there's not much more we can do. flooding the tip box with signed udp packets is pretty much the same as DoSing the satellite with ssl handshakes. but yeah we should make sure they're signed so we can at least throw out nonsense requests

docs/design/kademlia-removal.md
jtolds: it may be the case that for whatever reason, the satellite missed a bunch of udp packets. we don't want to consider storage nodes offline unless they're really offline, especially if they have lots of data that we care about.

jtolds: in my view, pinging nodes we haven't heard from is a robustness thing. less single points/systems of failure

isaachess: In that case, could we not simply use TCP (without TLS) and then know that we haven't missed any udp packets? I see your point, but I do worry that the small win we get from using UDP creates other complications that may not seem worth it. (Remember our place of former employment, and how they used udp so much, and how it was a complete nightmare ... ? :smile_cat:)

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
isaachess: Signed with what? If we have a signed UDP packet for the very first ping from a node, what do we use to verify the signature?

docs/design/kademlia-removal.md
jtolds: udp is great, maaaan!

okay, let's say we know for sure storage nodes will continue to send us packets until we get them (what switching to unauthenticated TCP would essentially provide). then the satellite goes offline for 2 hours. it's still more robust to make sure the satellite doesn't penalize anyone until it's double checked, right?

isaachess: Yes, but there are other ways to be sure you aren't prematurely penalizing. For example, if the satellite knows its policy is that nodes check in every hour, when it comes online, it knows to wait at least that long before penalizing nodes. Seems simple.

I love UDP as much as the next guy, but sometimes delivery guarantees are pretty nice, ya know? For server and client.

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jtolds: ooohh hot dang

so okay i guess the packet needs to include
1) the address of the node
2) the certificate chain to be expected once talking to the node that identifies the node
3) a signature of the above things with the leaf private key of that certificate chain.

this does mean you can just generate certificates willy nilly and submit whatever address you want to the tip box, but they will be ignored if
1) the difficulty of the computed id isn't high enough
2) the node you end up talking to doesn't have the same node id as the one computed from the tipster certificate chain.

docs/design/kademlia-removal.md
jtolds: but what if the satellite doesn't know it went offline (network connection link dipped for a bit, etc)? the ping double checking isn't only because of delivery guarantees. i agree with you that we could require that nodes check in with satellites once an hour and we penalize any node that we don't hear from. the udp part is a red herring - if we use UDP, a node gets a delivery guarantee if the satellite pings it back (okay, it got the udp packet). the question is, do we want to double check before we penalize?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
isaachess: > ooohh hot dang

Making you say that is what I live for.

> they will be ignored

Ignored in the long run, but only after the satellite has done a significant amount of work (ie pinging the node to verify everything).

That said, I think the POW in the nodeID saves us here. If we (1) de-dupe so that we're not processing the same nodeID over and over, (2) check the POW first thing, then I _think_ it becomes not worth it for a troll to do this just for fun.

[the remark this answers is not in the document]
jtolds: > Making you say that is what I live for.

lol oh no, we should get you some additional life goals

but yeah, great catch

docs/design/kademlia-removal.md
isaachess: ok, I'm convinced. Dialing before penalizing is good, you're right it could have network difficulties it doesn't know about.

UDP is fine. I just feel like, in general, when using UDP for packets that you actual do care about, it can lead to long-term issues that are hard to anticipate now and we end up engineering around for years to come. But I'll admit I can't foresee them now, I just have warning bells that go off when we use udp for packets are, actually, important.

jtolds: well the Achilles heel in my argument i don't know what to do about is... what if the network interruption is still ongoing when we try to ping back? what if the satellite was disconnected from the internet via tractor? what if only cogent/level3/comcast links were cut off via tractor? i don't think we've figured out a general solution to that problem yet

jtolds: https://www.theguardian.com/technology/2014/aug/14/google-undersea-fibre-optic-cables-shark-attacks

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jenlij: Would we like to create this UDP tip box in this iteration, or save it for future improvements?

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jenlij: what's the plan here?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+
+We will replace our Kademlia DHT and related entities with direct communication between satellites and storage nodes, 
+and keep the network fresh without kademlia node discovery and random lookups.
+
+## Design
+
+### Nodes reach out to satellites that they want to work with
+- The satellites are listed in the trust package
+- Nodes should communicate with satellites directly rather than using kademlia to traverse the network to find the address of a given ID.
+- Storage nodes should notify satellites when they start up, wait a random amount of time (to add jitter
```
wthorp: What is the scenario where storage node startup becomes a thundering herd?  Just in testing?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+### Network refreshes at a regular interval
+- Nodes will keep themselves up to date in the network by pinging all the satellites in their
+   trusted list every hour.
+- Satellites will ping the nodes back to confirm their addresses
+    - If is it successful, the satellite will insert or update the node in the overlay cache and
+       notify the node of success. Make sure to close the connection. Don’t use the transport observer to update the cache.
+    - If the the satellite does not confirm the node address, it does not proceed with updating the overlay cache. The node 
+    receives an error message and closes the connection when it times out.
+- Reaper service: Satellites check node connections on nodes that they haven’t heard from in a certain amount of time.
+
```
jenlij: We're not using this communication to determine reputation

jenlij: We can collect this info in log files if needed

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+- Nodes should communicate with satellites directly rather than using kademlia to traverse the network to find the address of a given ID.
+- Storage nodes should notify satellites when they start up, wait a random amount of time (to add jitter 
+http://highscalability.com/blog/2012/4/17/youtube-strategy-adding-jitter-isnt-a-bug.html), then start reporting in roughly on the hour
+
+### Network refreshes at a regular interval
+- Nodes will keep themselves up to date in the network by pinging all the satellites in their
+   trusted list every hour.
+- Satellites will ping the nodes back to confirm their addresses
+    - If is it successful, the satellite will insert or update the node in the overlay cache and
+       notify the node of success. Make sure to close the connection. Don’t use the transport observer to update the cache.
```
jenlij: update ip and uptime directly

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+    - If the the satellite does not confirm the node address, it does not proceed with updating the overlay cache. The node 
+    receives an error message and closes the connection when it times out.
+- Reaper service: Satellites check node connections on nodes that they haven’t heard from in a certain amount of time.
```
jenlij: uptime auditor

docs/design/kademlia-removal.md
jenlij: @jtolds ?

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+http://highscalability.com/blog/2012/4/17/youtube-strategy-adding-jitter-isnt-a-bug.html), then start reporting in roughly on the hour
+
+### Network refreshes at a regular interval
+- Nodes will keep themselves up to date in the network by pinging all the satellites in their
+   trusted list every hour.
+- Satellites will ping the nodes back to confirm their addresses
+    - If is it successful, the satellite will insert or update the node in the overlay cache and
+       notify the node of success. Make sure to close the connection. Don’t use the transport observer to update the cache.
+    - If the the satellite does not confirm the node address, it does not proceed with updating the overlay cache. The node 
+    receives an error message and closes the connection when it times out.
```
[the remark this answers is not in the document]
jenlij: I think we just want to send them a log message of sorts to let them know that they were not pinged back before their timeout

docs/design/kademlia-removal.md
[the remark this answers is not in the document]
jenlij: Ok I decided that we will use UDP in a future improvement

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,96 @@
...
+update the overlay cache, we can remove the transport observer. This would simplify uptime checks.
+
+### Node -> satellite communication initiation
+- To save resources and improve performance, satellites can have a “tip box” to receive UDP messages about new nodes
+- UDP messages need the address of the node, the certificate chain to be expected once talking to the node that identifies 
+the node, and a signature of the above things with the leaf private key of that certificate chain.
+- Messages will ultimately be ignored if the difficulty of the computed ID isn't high enough or the node you end up talking 
+to doesn't have the same node id as the one computed from the tipster certificate chain.
+
+### Uptime checking
```
ifraixedes: I will delete this section, all those concerns are related to uptime tracking/calculation design doc.
Perhaps, the only thing to mention is that this design doc will remove the transport observers and the node uptime column will be updated less frequently, but the uptime tracking/calculation design doc should cover how to measure an approximated time that an SN is offline.

jenlij: cool cool will do

docs/design/kademlia-removal.md
jenlij: @wthorp Ok so it turns out watchtower updates nodes all at once. This means that all those nodes will try to ping the satellite at once when they come online, and therefore every hour. We need to add some jitter to the autoupdate system, but for now the jitter on start up pings should help prevent a thundering herd.

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,89 @@
...
+- The satellites are listed in the trust package
+- Nodes should communicate with satellites directly rather than using kademlia to traverse the network to find the address of a given ID.
+- Storage nodes should notify satellites when they start up, wait a random amount of time (to add jitter 
+http://highscalability.com/blog/2012/4/17/youtube-strategy-adding-jitter-isnt-a-bug.html), then start reporting in roughly on the hour
+
+### Network refreshes at a regular interval
+- Nodes will keep themselves up to date in the network by pinging all the satellites in their
+   trusted list every hour.
+- Satellites will ping the nodes back to confirm their addresses
+    - If is it successful, the satellite will insert or update the node in the overlay cache and
```
thepaul: (I know this is already merged, and I'm late! But) Might want to note some requirements for a 'successful' pingback:

If a node with nodeID `N` has contacted satellite `S` and claimed its address is `A`,

1. `S` _must_ initiate a network connection `C` to address `A`
2. `S` _must_ verify that the remote endpoint on `C` has a private key corresponding to public key/node ID `N`. (It would be sufficient to complete negotiation of an SSL session over `C` and then verify that the remote end is using the same certificate used by `N` in the initial incoming ping.)
3. `S` _should_ verify that the remote endpoint on `C` agrees that its address is `A`. (This doesn't seem strictly necessary for security, but could prevent misconfigurations where a SNO runs multiple nodes with the same identity.)
4. `S` _should_ respond to the initial incoming ping from `N` with the result of the pingback, so that the dashboard on that node can report whether the node can receive incoming connections.

egonelbre: I added that information in https://github.com/storj/storj/pull/2837

docs/design/kademlia-removal.md
```diff
@@ -0,0 +1,89 @@
...
+### Disintegrate Kademlia from the network, storj sim and testplanet setups
+- Remove kademlia from the discovery package
+- Remove the bootstrap node - work with Ops
+  - Remove the vouchers service and related tables
+  - Work with QA to make sure storage nodes don’t crash on errors related to the elimination of Kademlia
+- if they don’t update immediately ->  keep just the overlay.Ping rpc method, it will be much easier for a new satellite 
+to work with old and new storage nodes.
+
+### Update whitepaper to address kademlia removal and the addition of satellite opt-in
+  - Delete the audit gating design doc
```
JessicaGreben: sorry im late to the CR party, buy I'm wondering if we want to create an `docs/archive` directory and move old not-no-longer-relevant design docs there instead of deleting?

jenlij: we could do that!

PR #2690, opened 2019-08-01
satellite/metainfo: validate piece hash

aleitner:
What: 
Verify piece hash from storage node on satellite

Why:
We need to verify the piece hash proto received from the storage node is signed correctly.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/satellite/metainfo/validation.go
+++ b/satellite/metainfo/validation.go
@@ -20,6 +20,7 @@ import (
 	"storj.io/storj/pkg/encryption"
 	"storj.io/storj/pkg/macaroon"
 	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/signing"
 	"storj.io/storj/pkg/storj"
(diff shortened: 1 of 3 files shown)


satellite/metainfo/validation.go
```diff
@@ -369,5 +370,20 @@ func (endpoint *Endpoint) validatePieceHash(ctx context.Context, piece *pb.Remot
 			return errs.New("piece hash PieceSize is larger than order limit, removing from pointer (%v > %v)", piece.Hash.PieceSize, limit.Limit)
 		}
 	}
+
+	publicKey, err := endpoint.certdb.GetPublicKey(ctx, piece.NodeId)
```
egonelbre: Note, we always load it from the database otherwise it'll add massive amount of traffic to the database.

Similarly there should be one DB call per verification rather than ~100.

PR #2694, opened 2019-08-02
metainfo: Batch request

mniewrzal:
What: Additional `Batch` request to send multiple Metainfo requests as one call. e.g. multiple delete segment requests in one call.

Why: To optimize uplink -> satellite communication and reduce round trips. 
https://storjlabs.atlassian.net/browse/V3-2304

Please describe the tests:
 - Test 1: basic automatic test added
 
Please describe the performance impact: none, its just implementation and its not used yet


mniewrzal:
Current implementation fail fast on first error but also returns list of responses for completed requests so we can still track where error occur and what was not finished.
tbh I'm not sure which approach will be better. I choose to abort on first error to make it simpler and avoid complexity with cases where something from the middle of the list will fail and we will continue executing. I believe that we will use `Batch` to combine requests for common groups and not things completely unrelated to each other. Of course, I might be wrong :)

satellite/metainfo/batch.go
```diff
@@ -0,0 +1,215 @@
...
+func (endpoint *Endpoint) Batch(ctx context.Context, req *pb.BatchRequest) (resp *pb.BatchResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	resp = &pb.BatchResponse{}
+
+	resp.Responses = make([]*pb.BatchResponseItem, 0, len(req.Requests))
+
+	// TODO find a way to pass some parameters between request -> response > request
+	// TODO maybe use reflection to shrink code
+	for _, request := range req.Requests {
```
coyle: would it make sense to do some actions concurrently ?

wthorp:
ah, gotcha

satellite/metainfo/batch.go
mniewrzal: maybe at some point. We would need logic to determine what we can do in parallel. I think it's out of scope for now.

uplink/metainfo/batch.go
```diff
@@ -0,0 +1,327 @@
...
+)
+
+// Batch represents sending requests in batch
+type Batch struct {
+	client   pb.MetainfoClient
+	requests []*pb.BatchRequestItem
+}
+
+// AddCreateBucket adds request to batch
+func (batch *Batch) AddCreateBucket(params CreateBucketParams) {
```
egonelbre: I would flip the assignments and accept an []Request in the batch client call, by making a method like:

```
func (params *CreateBucketParams) BatchItem() *pb.BatchRequestItem {
```

This would allow to define an interface `type BatchItem interface { BatchItem() *pb.BatchRequestItem }`. Then client would accept an argument called `...BatchItem`.

mniewrzal: I like it, thanks!

I guess with one PR in future it would be good to rename `*Params` types into `*Request` to be more consistent.

PR #2696, opened 2019-08-02
web/satellite common search component created

VitaliiShpital:
What: web/satellite common search component created

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

web/satellite/src/components/common/SearchComponent.vue
```diff
@@ -0,0 +1,78 @@
...
+</template>
+
+<script lang="ts">
+    import { Component, Prop, Vue } from 'vue-property-decorator';
+
+    declare type searchCallback = (search: string) => Promise<void>;
+
+    @Component
+    export default class SearchComponent extends Vue {
+        @Prop({default: ''})
```
VinozzZ: Is line 24 not enough for declaring the `default` prop?

web/satellite/tests/unit/common/__snapshots__/SearchComponent.spec.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
VinozzZ: Is this file expected to be in here?

web/satellite/src/components/common/SearchComponent.vue
VitaliiShpital: i'm not sure if i understand you right

there is '@Prop' decorator for each prop we have. Each prop has it's own default value so default value should be declared in every '@Prop' decorator.

here 24 line default value stands for 25 line prop and 22 line default value stands for 23 line prop.

Did you mean something else?

web/satellite/tests/unit/common/__snapshots__/SearchComponent.spec.ts.snap
VitaliiShpital: this file was generated automatically during testing. Should i remove this line?

web/satellite/src/components/common/SearchComponent.vue
VinozzZ: A-ha, I see. You got it just right! Thanks for clarifying for me

web/satellite/tests/unit/common/__snapshots__/SearchComponent.spec.ts.snap
VinozzZ: IMO, I don't think we need to have source control for snapshots, but maybe we can see what others think?

crawter: We need this snapshots to be able to compare that our markup has been changed.

The Jest [docs](https://jestjs.io/docs/en/snapshot-testing.html) also say that "the snapshot artifact should be committed alongside code changes"

Also, this means that the tests won't have a baseline for comparison when run in a fresh environment or during CI

VinozzZ: Oh, I didn't know we are going to set up running the tests on CI now. That makes sense!

crawter: @VinozzZ 
Yeah, we are going to make some frontend improvements at the beginnig of september. Adding linter/unit test/auto tests (yes, we have them :D ) to our CI also part of it

web/satellite/tests/unit/common/SearchComponent.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        expect(wrapper.find('input').props('placeHolder')).toMatch('');
+        expect(wrapper.find('input').props('search')).toMatch('');
+    });
+
+    it('functions onMouseEnter/onMouseLeave work correctly', () => {
+        const wrapper = mount(SearchComponent);
+
+        wrapper.vm.onMouseEnter();
+
+        expect(wrapper.vm.inputWidth).toMatch('602px');
```
VinozzZ: Seems like this test broke the build @VitaliiShpital . If we want to access private field, we need to use `vm.['inputWidth']`

crawter: Do u mean that npm run build command fails?
its strange, because npm run test passes

PR #2699, opened 2019-08-02
storagenode: more live request tracking

jtolds:
What: show grpc errors in monkit data and add more tracking around storage node live requests

Why: when an upload is rejected we return the Unavailable status code over grpc, but monkit doesn't know how to show grpc status codes, so we don't know how many unavailable requests are getting returned. first, add grpc status codes to monkit so monkit knows and can show us. second - we can see stats about individual types of live requests, but add tracking for all of them

Please describe the tests: no functionality change, just additional instrumentation

Please describe the performance impact: very, very slight - a couple new monkit compare and swaps for upload/download and a couple of new function calls during request errors


pkg/process/error.go
```diff
@@ -0,0 +1,18 @@
...
+
+package process
+
+import (
+	"google.golang.org/grpc/status"
+	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+)
+
+func init() {
+	monkit.AddErrorNameHandler(func(err error) (string, bool) {
```
stefanbenten: Not entirely sure what this error handler adds as a benefit.
Would you explain that for me?

stefanbenten: Do you want to add additional metrics what grpc error is being returned?

jtolds: i tried to explain this in the PR description - this part:
> when an upload is rejected we return the Unavailable status code over grpc, but monkit doesn't know how to show grpc status codes, so we don't know how many unavailable requests are getting returned. first, add grpc status codes to monkit so monkit knows and can show us.

stefanbenten: Is that function not run for every err that is being returned by any function?
Not sure about the overhead there.

jtolds: you're absolutely right, but errors (hopefully! ha) don't happen that often.

PR #2701, opened 2019-08-02
Add more monkit for segment piece info

bryanchriswhite:
What ([Jira](https://storjlabs.atlassian.net/browse/V3-2096)): Adding the following monkit metrics:
- Observing `segment_pieces`, an integer of the number of pieces the segment consists of.
- Observing `segment_piece_percent_success`, a percentage of successful piece puts divided by the number of optimal/success threshold.
- Observing `segment_piece_percent_failure`, a percentage of failed piece puts divided by the optimal/success threshold.

Why: More data is better.

Please describe the tests: No tests are added in this PR because it only adds monkit metrics.
 

uplink/ecclient/client.go
```diff
@@ -148,6 +149,18 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 	}()
 
 	successes := int(atomic.LoadInt32(&successfulCount))
+	mon.IntVal("segment_pieces").Observe(int64(len(limits)))
```
navillasa: could this be pieceCount instead of len(limits)?

uplink/ecclient/client.go
```diff
@@ -148,6 +149,18 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, p
...
+	mon.FloatVal("segment_pieces_percent_success").Observe(float64(successes) / float64(rs.OptimalThreshold()))
+
+	var failures int
+	for range limits {
+		info := <-infos
+		if info.err != nil && err != context.Canceled {
```
navillasa: just curious what context canceled errors are reported as if not failures? are they expected/normal so don't need to be reported?

uplink/ecclient/client.go
bryanchriswhite: It should be, thanks!

uplink/ecclient/client.go
bryanchriswhite: Yea that was the thought. We're really only interested in stats on the pieces that either failed for some other reason or were successful. We have the total number of pieces so we can always calculate how many were cancelled if we need.

mobyvb: We should probably use `errs2.IsCanceled(err)` from `storj.io/storj/internal/errs2`

uplink/ecclient/client.go
```diff
@@ -148,6 +149,18 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 	}()
 
 	successes := int(atomic.LoadInt32(&successfulCount))
+	mon.IntVal("segment_pieces").Observe(int64(pieceCount))
+	mon.FloatVal("segment_pieces_percent_success").Observe(float64(successes) / float64(rs.OptimalThreshold()))
```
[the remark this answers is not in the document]
bryanchriswhite: :+1:

uplink/ecclient/client.go
```diff
@@ -148,6 +149,18 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, p
...
+
+	var failures int
+	for range limits {
+		info := <-infos
```
mobyvb: I don't think you can listen on the channel like this. We are already piping exactly one `info` into `infos` per `ec.putPiece` above, and those were already consumed. At this point in execution, all the `info`s should have already been received by the for loop above.

bryanchriswhite: Ah you're right. It's been a minute since I've used the channels. How embarrassing. I'll fix this.

internal/errs2/ignore.go
```diff
@@ -9,14 +9,19 @@ import (
...
 	"github.com/zeebo/errs"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // IsCanceled returns true, when the error is a cancellation.
 func IsCanceled(err error) bool {
 	return errs.IsFunc(err, func(err error) bool {
+		status, ok := status.FromError(err)
```
mobyvb: this won't fail if `err` is not a grpc error, correct?

uplink/ecclient/client.go
```diff
@@ -117,6 +120,9 @@ func (ec *ecClient) Put(ctx context.Context, limits []*pb.AddressedOrderLimit, p
 		}
 
 		if info.err != nil {
+			if !errs2.IsCanceled(info.err) {
```
mobyvb: do you think it would be beneficial for us to report the number of nodes that were canceled? `segment_pieces_canceled` or something?

internal/errs2/ignore.go
bryanchriswhite: Ok will be false in that case, hence `(ok && status.Code()...)`.

uplink/ecclient/client.go
bryanchriswhite: Well you did say you can never have enough data

internal/errs2/ignore.go
kaloyan-raev: @egonelbre could you review the changes in this function?

kaloyan-raev: @bryanchriswhite wouldn't `status.Code(err) == codes.Canceled` be a simpler check?

bryanchriswhite: It would seem so. This is the first time i've used `grpc/status` & `grpc/codes`.

PR #2702, opened 2019-08-02
satellite/repair: add shouldDelete

wthorp:
What: 
Adds a 'shouldDelete' flag to repair worker responses

Why:
There are some circumstances where an item may get stuck in the repair queue forever.  This is an attempt to fix that.
https://storjlabs.atlassian.net/browse/V3-2139

Please describe the tests:
Should use existing tests.
 
Should prevent some items from being stuck in the repair queue, which should make the queue more robust, and perhaps faster in some sense.


wthorp:
> While this looks right to me, I feel like this might need some tests just to confirm that repair jobs are consistently getting deleted when they should, and outcomes don't change in the future.

I felt the same way... but there are 14 different returns changed in this function alone, not to mention the amount of potential logic behind those functions.  Making 14+ unit tests doesn't seem to be the right answer either.  Thoughts?

satellite/repair/repairer/segments.go
```diff
@@ -95,21 +95,21 @@ func (repairer *SegmentRepairer) Repair(ctx context.Context, path storj.Path) (e
...
-		return Error.New("error getting missing pieces %s", err)
+		return false, Error.New("error getting missing pieces %s", err)
 	}
 
 	numHealthy := len(pieces) - len(missingPieces)
 	// irreparable piece, we need k+1 to detect corrupted pieces
 	if int32(numHealthy) < pointer.Remote.Redundancy.MinReq+1 {
 		mon.Meter("repair_nodes_unavailable").Mark(1)
-		return Error.Wrap(IrreparableError.New("segment %v cannot be repaired: only %d healthy pieces, %d required", path, numHealthy, pointer.Remote.Redundancy.MinReq+1))
+		return false, Error.Wrap(IrreparableError.New("segment %v cannot be repaired: only %d healthy pieces, %d required", path, numHealthy, pointer.Remote.Redundancy.MinReq+1))
```
mobyvb: We should delete in the case of an `IrreparableError`. In fact, we have an explicit case for it in `repairer.go`.

satellite/repair/repairer/repairer.go
```diff
@@ -110,7 +110,7 @@ func (service *Service) worker(ctx context.Context, seg *pb.InjuredSegment) (err
 	workerStartTime := time.Now().UTC()
 
 	service.log.Info("Limiter running repair on segment", zap.Binary("segment", seg.GetPath()))
-	err = service.repairer.Repair(ctx, string(seg.GetPath()))
+	shouldDelete, err := service.repairer.Repair(ctx, string(seg.GetPath()))
```
mobyvb: Make sure you still check `shouldDelete` if `err != nil`

jtolds: wait, really? normal convention is all guarantees about other return values are off if err != nil. that's surprising

jtolds: maybe to make it clear we're bucking convention, it should be `err, shouldDelete := service.repairer.Repair(...`?

mobyvb: oh, I guess there are certain errors that we will want to delete for, and other errors that we don't want to delete for. But we can keep it as is and explicitly delete if the error is a certain type rather than relying on `shouldDelete`. I don't know the best practice in this situation.

satellite/repair/repairer/segments.go
```diff
@@ -159,31 +159,31 @@ func (repairer *SegmentRepairer) Repair(ctx context.Context, path storj.Path) (e
...
 	if err != nil {
-		return Error.Wrap(err)
+		return false, Error.Wrap(err)
 	}
 
 	// Download the segment using just the healthy pieces
 	rr, err := repairer.ec.Get(ctx, getOrderLimits, getPrivateKey, redundancy, pointer.GetSegmentSize())
 	if err != nil {
-		return Error.Wrap(err)
+		return true, Error.Wrap(err)
```
bryanchriswhite: Same question about why delete in this case?

satellite/repair/repairer/repairer.go
kaloyan-raev: I guess this should be an exception from the normal convention similar to `Read`: https://golang.org/pkg/io/#Reader
In `Read` the caller is supposed to check the `n` return value even if `err != nil`.

satellite/repair/repairer/segments.go
```diff
@@ -133,13 +133,13 @@ func (repairer *SegmentRepairer) Repair(ctx context.Context, path storj.Path) (e
 
 	bucketID, err := createBucketID(path)
 	if err != nil {
-		return Error.Wrap(err)
+		return true, Error.Wrap(err)
```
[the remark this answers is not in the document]
kaloyan-raev: The only error this function can return is `no bucket component in path`. This should only happen if we are repairing a segment with an invalid path. In such case, it makes sense to delete it from the queue, because it will fail the same way on the next iteration.

satellite/repair/repairer/segments.go
wthorp: fixing

satellite/repair/repairer/segments.go
wthorp: I should figure out how to check for database connectivity errors.  All of the other logic in Get is input validation or that the pointer has been deleted.

wthorp: Will add a comment to code, clarifying.  From what I can tell, .Get() only sets up the connection and can only fail from input validation.

bryanchriswhite: :+1:

satellite/repair/repairer/repairer.go
wthorp: I added a comment clarifying the use of shouldDelete

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,28 @@
+[This is a template for Storj v3 design docs]
```
mobyvb: Should this file be in this PR?

wthorp: rm'ed

PR #2703, opened 2019-08-02
docs/design: auditing 2.0

navillasa:
What: 
Audit by node design doc.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+After selection, the rest of the auditing process will occur the same way (picking a segment, picking a stripe, downloading all erasure shares associated with that stripe).
+The chances of selecting the same stripe are rare, but should be accounted for.
+
+With both loops, we should have auditing that occurs statistically uniform across both nodes and bytes.
+
+### **Selection via Reservoir Sampling**
+
+Reservoir sampling: a family of algorithms for randomly choosing a sample of items with uniform probability from a stream of data (of unknown size).
+
+During uploads or during garbage collection bloom filter generation, we're going through all segments and creating filters per node.
```
wthorp: Could we do reservoir sampling during the segment-based audit loop? [And thus actually only have one loop?]

navillasa: oh yeah I guess the gc observer is just going through the pointerdb same as the existing audit loop. I feel like there's some difference I'm missing too but I'm not sure.. 🤔

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+
+This design doc outlines how we will implement a Version 2 of the auditing service.
+With our current auditing service, we are currently auditing per segment of data.
+We need to keep this existing method of auditing, but also add an additional audit selection process that selects random nodes to audit.
+
+## Background
+
+As our network grows, it will take longer for nodes to get vetted.
+This is because every time an upload happens, we send 5% of the uploaded data to unvetted nodes and 95% to vetted nodes.
+When auditing occurs, we currently select a random stripe within a segment.
```
mobyvb: *within a random segment

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+Currently, there's not a way to select a stripe based on a storage node.
+
+## Design
+
+Two different loops will select audits:
+- One loop will select based on stripes.
+- The second loop will select based on nodes.
+
+After selection, the rest of the auditing process will occur the same way (picking a segment, picking a stripe, downloading all erasure shares associated with that stripe).
+The chances of selecting the same stripe are rare, but should be accounted for.
```
mobyvb: We could use a bloom filter to determine whether we have audited a stripe before lol

docs/design/audit-v2.md
mobyvb: The segment based audit loop uses random selection so it will not uniformly hit all nodes/segments. The GC observer (or any observer on the metainfo loop) will always hit every segment on the satellite exactly once, allowing us to get the most accurate possible sample for each node. In garbage collection, we also keep track of the current number of pieces seen for each node ID, which would allow us to easily determine the probability that we should replace something in that node's reservoir when we find a new piece for that node

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+This is a good opportunity to insert reservoir sampling for auditing purposes.
+
+We will have little reservoirs of segments for all of the nodes.
+We could generate 5 random segments per node, then randomly select some segments per node.
+By choosing nodes out of the reservoirs, we would receive a random sample.
+Unvetted nodes would get the same number of audits as vetted nodes.
+
+However, one problem with piggybacking off of existing garbage collection bloom filter generation is that bloom filter generation occurs every seven days.
+
+The gc observer subscribes to the metainfo loop every week, but we could make an observer that subscribes more often and gets a stream as often as we want, so it doesn't have to be part of gc.
```
mobyvb: The main issue with integrating it into the gc observer is it means we will always be forcing the gc interval and the node audit reservoir sampling interval to be exactly the same. I don't think the performance gain of combining the two is necessarily worth the limitations created. Plus, this is the entire reason we created the metainfo loop/observer architecture.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+However, this could potentially create a lot more db transactions, which we want to minimize.
+[See Open Issue 1.]
+
+## Rationale
+
+An initial idea for implementation was to sort the nodes table for nodes with least amount of audits, then select one node randomly within that low amount of audits.
+However, we decided it may not be necessary to keep track of how many audits per storage node if we're able to randomly select across nodes.
+
+We were also considering implementing a reverse way of looking up segments or pieces by node IDs.
+e.g. a table where each row is a node ID and an encrypted metainfo path.
```
mobyvb: Based on the conversations we've had, I think we want to avoid adding additional tables if possible

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+Every time a segment is committed or deleted, that table (and every node) gets updated.
+The advantage is that it can super simplify the garbage collection process, but complexify upload and download.
+
+## Implementation
```
mobyvb: I think there are only like 2 or 3 possible implementations based on what we know so far, so I think it would be good to write down ticket-level descriptions of how they'd be implemented. Here's my idea for one:

1. Add reservoir sampling struct for node auditing
2. Create node audit observer that connects to metainfo loop and updates the reservoir sampling struct
3. Add audit loop to randomly select a node, biased in favor of unvetted nodes, and select a random segment to audit based on the populated reservoir sampling struct for that node
4. Audit the segment selected from part 3, as we are doing already

Another example would be almost the same, but using the garbage collection observer instead of creating a new one for step 2. Some of those in the list also seem like they should be part of the same ticket. idk :)

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+3. Add new methods to this loop that allow for selection from reservoir.
+4. Coordinate both selection loops to use same "audit path" for verification.
+
+## Open issues
+
+1. Location of reservoir sampling.
+    - Where do we have this happen?
+    - We want uploads to be performant with minimal db transactions, but we know that audits need to happen very frequently.
+
+2. How to coordinate both loops running simultaneously/using same "audit path" once data is selected for audit?
```
mobyvb: I think this could be reworded to something like "Should we run both audit selection processes within the same loop or in separate loops?"

docs/design/audit-v2.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. Nail down open issue 1 and identify where reservoir sampling should occur.
+2. Create new part of audit service that loops over nodes.
+3. Add new methods to this loop that allow for selection from reservoir.
+4. Coordinate both selection loops to use same "audit path" for verification.
+
+## Open issues
+
+1. Location of reservoir sampling.
+    - Where do we have this happen?
```
mobyvb: Specify the options (garbage collection observer or a new observer)

docs/design/audit-v2.md
navillasa: tenor

docs/design/audit-v2.md
navillasa: I think that's asking a different question though, because I'm definitely anticipating two loops, but I guess I was thinking they wouldn't be able to do the regular verifying functions at the same time. But actually now that I think about it, that shouldn't be an issue. I might get rid of this question!

navillasa: Or I guess I'll just add the question you suggested.

mobyvb: Yeah, I think if there are two separate loops, there is no coordination to be done, except maaaybe avoiding re-audits of the same segment

docs/design/audit-v2.md
kaloyan-raev: What's wrong if we occasionally audit a stripe more than once?

docs/design/audit-v2.md
```diff
@@ -0,0 +1,85 @@
...
+With both loops, we should have auditing that occurs statistically uniform across both nodes and bytes.
+
+### **Selection via Reservoir Sampling**
+
+Reservoir sampling: a family of algorithms for randomly choosing a sample of items with uniform probability from a stream of data (of unknown size).
+
+During uploads or during garbage collection bloom filter generation, we're going through all segments and creating filters per node.
+This is a good opportunity to insert reservoir sampling for auditing purposes.
+
+We will have little reservoirs of segments for all of the nodes.
```
kaloyan-raev: Why do we need reservoirs for all of the nodes? Wouldn't it be enough to have reservoirs only for unvetted nodes? Once a node becomes vetted, we can delete its reservoir.

docs/design/audit-v2.md
mobyvb: I think it should be fine to audit a stripe more than once sometimes.

docs/design/audit-v2.md
mobyvb: We still want to audit vetted nodes through this process, right? So we should have reservoirs for them. If we were only auditing unvetted nodes, I would agree.

kaloyan-raev: My understanding is that we introduce this additional auditing process to give a chance for unvetted nodes to get more audit requests. Vetted nodes are already audited often enough through the existing audit process based on the amount the data they store. The more data a node stores, the more often would be audited.

docs/design/audit-v2.md
wthorp: As discussed in standup, we'll consider using a single audit loop.  That means we'll have to update the main audit to use the metainfo observer loop, and use a reservoir sample to get random segments for the main loop.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,78 @@
...
+As more nodes join the network, it will become exponentially less likely that an unvetted node will be audited since most data will be stored on vetted nodes.
+With a satellite with one petabyte of data currently, new nodes will take one month to get vetted, which is way too long.
+
+We want a way to select segments to audit based on statistically randomly selected storage nodes.
+However, the way we select nodes should also be biased toward unvetted nodes.
+Currently, there's not a way to select a stripe based on a storage node.
+
+## Design
+
+We will create an audit observer that uses the metainfo loop, and this observer will use both methods of selection. It will create reservoir samples for segments following the current selection method (randomly across bytes), and also create a small reservoir for each unvetted node.
```
egonelbre: Why not have reservoirs for all the nodes?

docs/design/audit-v2.md
```diff
@@ -0,0 +1,78 @@
...
+
+After selection, the rest of the auditing process will occur the same way (picking a segment, picking a stripe, downloading all erasure shares associated with that stripe).
+The chances of selecting the same stripe are rare, but it's normal/expected behavior if occasionally a stripe is audited more than once.
+
+With both loops, we should have auditing that occurs statistically uniform across both nodes and bytes.
+
+### **Selection via Reservoir Sampling**
```
egonelbre: Remove the bold, it's inconsistent with the default styling.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,78 @@
...
+
+An audit observer on the metainfo loop, like every observer on the loop, will always hit every segment on the satellite exactly once, allowing us to get the most accurate possible sample for each node.
+
+While we initially considered integrating the audit system's random node selection process with the existing garbage collection observer, we decided not to do this because the difference in required interval for each observer would mean either too many bloom filters being created unnecessarily or audits occurring too slowly.
+
+An initial idea for implementation was to sort the nodes table for nodes with least amount of audits, then select one node randomly within that low amount of audits.
+However, we decided it may not be necessary to keep track of how many audits per storage node if we're able to randomly select across nodes.
+
+We were also considering implementing a reverse way of looking up segments or pieces by node IDs.
+e.g. a table where each row is a node ID and an encrypted metainfo path.
```
egonelbre: This would increase the database size significantly.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. Location of reservoir sampling. (resolved)
+    - Where do we have this happen? In a garbage collection observer or a new observer?
+
+From Moby: "The main issue with integrating it into the gc observer is it means we will always be forcing the gc interval and the node audit reservoir sampling interval to be exactly the same. I don't think the performance gain of combining the two is necessarily worth the limitations created. Plus, this is the entire reason we created the metainfo loop/observer architecture."
+
+    - We want uploads to be performant with minimal db transactions, but we know that audits need to happen very frequently.
+    - We'll create a new audit observer.
+
+2. Should we run both audit selection processes within the same loop or in separate loops?
```
egonelbre: Joining into a loop doesn't create a new iteration unless the timings are different.

We could add a join together that ensures that multiple observers always. But let's keep the logic separate to avoid conflating different approaches.

docs/design/audit-v2.md
mobyvb: We discussed this a little today, but we want to minimize memory consumption, and the primary reason we are adding this method of auditing is to give unvetted nodes more audits so they can get vetted. Once they are vetted, they should be audited as normal based on how much data they are storing, but as far as I understand, it is not super important to give already-vetted nodes more audits.

docs/design/audit-v2.md
mobyvb: Yes, we wanted to include this as something we discussed, but the actual implementation we have decided upon does not include this reverse lookup.

docs/design/audit-v2.md
navillasa: it's more visible.

docs/design/audit-v2.md
navillasa: I'll clarify in the doc that this is something we're not actually going to do

docs/design/audit-v2.md
egonelbre: Let's add this information to the documentation.

docs/design/audit-v2.md
egonelbre: Why not make it a higher level heading?

navillasa: because it should be under the design heading.

docs/design/audit-v2.md
navillasa: does this mean I need to add further clarification to the doc? if so, what do I need to add?

egonelbre: Let's keep the observers separate, it makes the logic easier to read.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Implementation
+
+1. Add reservoir sampling struct for node auditing.
+2. Create an audit observer that connects to metainfo loop but otherwise uses existing audit code for selection.
+3. Have the audit observer update the reservoir sampling structs.
+3. Make selection biased in favor of unvetted nodes, and select a random segment to audit based on the populated reservoir sampling struct for that node.
+4. Audit the segment selected from part 3, as we are doing already.
+
+## Open issues
```
littleskunk: Do we need to fill the reservoir samples every time or can we run it once per day and use the results for multiple audits?

What do we do if the satellite gets restarted? The audit process will start befor the reservoir sampling is finished. How do we handle that?

Do we get problems with the containment mode? Containment mode can deal with only 1 audit at a time (primary key on nodeID). If we run the new audit in parallel the containment mode will fail.

Same interval for both audit jobs? That would allow us to send out up to 2880 audits with up to 80 storage nodes per audit. That is about 230 audits per storage node. Split that halve on both audit jobs? New nodes would get vetted in only 1 day. How do we balance that? (We can increase the vetting requirement to get back to 1 month. However I am not sure if a 50/50 split is the best approach. We could split it 90/10 or what ever. I have no opinion on that. I just want to point it out.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,79 @@
...
+We need to keep this existing method of auditing, but also add an additional audit selection process that selects random nodes to audit.
+
+## Background
+
+As our network grows, it will take longer for nodes to get vetted.
+This is because every time an upload happens, we send 5% of the uploaded data to unvetted nodes and 95% to vetted nodes.
+When auditing occurs, we currently select a random stripe within a random segment.
+If we're selecting audits at random per byte, every segment has some percentage that went to vetted nodes.
+As more nodes join the network, it will become exponentially less likely that an unvetted node will be audited since most data will be stored on vetted nodes.
+With a satellite with one petabyte of data currently, new nodes will take one month to get vetted, which is way too long.
```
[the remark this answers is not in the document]
navillasa: ah okay I understand, will change

docs/design/audit-v2.md
```diff
@@ -0,0 +1,79 @@
...
+
+This design doc outlines how we will implement a Version 2 of the auditing service.
+With our current auditing service, we are currently auditing per segment of data.
+We need to keep this existing method of auditing, but also add an additional audit selection process that selects random nodes to audit.
+
+## Background
+
+As our network grows, it will take longer for nodes to get vetted.
+This is because every time an upload happens, we send 5% of the uploaded data to unvetted nodes and 95% to vetted nodes.
+When auditing occurs, we currently select a random stripe within a random segment.
```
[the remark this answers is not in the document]
jtolds: i'm pretty sure it's per segment. the satellite doesn't know about files, just segments

docs/design/audit-v2.md
jtolds: maybe we need three audit loops - per segment, per node, and per unvetted node?

navillasa: when you say loops do you mean observers on the metainfo loop? we were only planning on using one observer

jtolds: i guess i should have said "audit selection strategies" instead of "audit loops"

mobyvb: I don't think we need to separate "per node" and "per unvetted node". If we are going to be auditing on a per-node basis, we might as well keep them together. We keep a reservoir for every node, then we just have a query to pick some number of vetted and unvetted nodes to audit.

navillasa: or could we do what we talked about on Friday with reservoir sampling but doing it for all nodes instead of only unvetted nodes?

mobyvb: Oh, I guess we could have a single loop to build the reservoirs for all the nodes as well as the "primary" segment selection reservoir, then three separate services like you say that can be configured independently.

docs/design/audit-v2.md
mobyvb: > Do we need to fill the reservoir samples every time or can we run it once per day and use the results for multiple audits?

We should be able to use the same reservoir sample for all the segments in the reservoir. So if our reservoir size is 10, we should be able to use that for 10 separate audits for any node without updating it.

> What do we do if the satellite gets restarted? The audit process will start befor the reservoir sampling is finished. How do we handle that?

Not sure if there is a way to get around this issue at the moment. The audit job will just have to wait for the reservoirs to be properly populated. Otherwise all segments don't get an equal chance to be placed in the reservoir.

> Do we get problems with the containment mode? Containment mode can deal with only 1 audit at a time (primary key on nodeID). If we run the new audit in parallel the containment mode will fail.

We can do the same thing that we are doing right now, which is if a node is in containment mode, we re-audit the segment it was put into containment mode for, and do not audit it for the "currently selected" segment.

As far as the intervals for the different audit jobs, I imagine we will make them configurable, so we can adjust them based on our network size and number of vetted/unvetted nodes.

docs/design/audit-v2.md
kaloyan-raev: > maybe we need three audit loops - per segment, per node, and per unvetted node?

Aren't we overcomplicating everything?

I'd step back and ask "Why do we need auditing?". From past discussions, we agree that it is not feasible to audit all data. This is why we were selecting random stripes from random segments. So we can catch unreliable nodes and disqualify them. This randomness approach gives more pressure (i.e. audit requests) to nodes storing more data.

Now, with the above approach, we see that there are some issues. In a large network, unvetted nodes would get very few audit requests, and there are also cheater strategies that vetted nodes with a small amount of stored data could apply.

What if, we implement only a single loop with reservoir samples for all (vetted and unvetted) nodes? And ditch the old loop that select random stripes from random segments. This way we give all nodes equal chance to get audited. We will still catch if they are reliable storing data. And, if a node is failing too many audits, does it really matter how much data it stores? We would disqualify it regardless of the amount of data, right? So, sending more audit requests to nodes storing more data should not make any sense.

littleskunk: We need both.
The per node selection is good to get vetted and to catch cheater. We want to detect cheater nodes within a few das.
The per byte selection is good to detect hardware failures in big nodes as quickly as possible. This time we don't want to wait a few days. The bigger the node the faster we want to detect the failure.

docs/design/audit-v2.md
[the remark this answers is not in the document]
littleskunk: The satellite knows about the last segment which is the same as per file. Let me run a quick test. Just because I want to know it^^

docs/design/audit-v2.md
littleskunk: > We can do the same thing that we are doing right now, which is if a node is in containment mode, we re-audit the segment it was put into containment mode for, and do not audit it for the "currently selected" segment.

That doesn't work. Lets say both audit jobs are running in parallel. At the point they are starting your node is not in containment mode. At the point they both finish they want to add you to containment mode. The containment mode has a primary key. You can't insert 2 entries. The containment mode is not designed to deal with audit jobs running in parallel. See https://storjlabs.atlassian.net/browse/V3-2071

docs/design/audit-v2.md
kaloyan-raev: Note that even with per node detection only, nodes that store more data will be audited more frequently. This is due to the fact that we don't audit just a single node at the same time (like the reverification of contained node), but we download the erasure shares from all nodes in the pointer.

The reservoir sample for each node will include pointers that for sure have a piece for this specific node. But when a stripe from this pointer is audited, the erasure shares will be downloaded from all nodes in this pointer. If a node stores more data, there is a higher chance that it has a piece from this pointer too.

docs/design/audit-v2.md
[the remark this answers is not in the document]
littleskunk:
```
head -c 50M </dev/urandom | uplink put s3://littleskunk/file1
head -c 100M </dev/urandom | uplink put s3://littleskunk/file2
```

A total of 2 files and 3 segments in the database. After 134 audits I can see 51/43/40 times the same pieceIDs. Good that we figured out that I am wrong...

docs/design/audit-v2.md
navillasa: wouldn't the primary key make it so that just one pending audit entry would be inserted? if not, do you know of a way that we could prevent this potential problem?

littleskunk: The latest information was that we can replace the existing audit job instead of running two audit jobs in parallel. This means we don't need to worry about the containment mode for the moment and can fix that issue later.

littleskunk:
@aligeti 
> Why not let the Storagenodes do the auditing based on info sent by satellite and report back the result with checksum, which satellite can verify with it stored info.

The satellite has to download one random stripe from all storage nodes. Thanks to our reed solomon encoding the satellite can verify the pieces by comparing them against each other. On the storage node side this validation wouldn't be possible and the satellite doesn't store additional informations that could be used for an audit.

> That way we can let Storagenodes decide how fast they wanna get vetted.

I am a dirty little cheater. Can I decide to get vetted within a few hours? -> Our vetting process needs to be long enough to detect bad nodes. As long as they are unvetted they are no risk for us.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,93 @@
...
+If we're selecting audits at random per byte, every segment has some percentage that went to vetted nodes.
+As more nodes join the network, it will become exponentially less likely that an unvetted node will be audited since most data will be stored on vetted nodes.
+With a satellite with one petabyte of data, new nodes will take one month to get vetted. However, with 12PB of data on the network, this vetting process time would take 12 months, which is much too long. We need a scalable approach.
+
+We want a way to select segments to audit based on statistically randomly selected storage nodes.
+Currently, there's not a way to select a random storage node and audit on that basis.
+
+## Design
+
+We will create an audit observer that uses the metainfo loop, and this observer will create a reservoir sample of segments for every node. This audit observer will replace the existing method of auditing per byte. The audit will pick a random node, then a random segment from the node's reservoir.
```
wthorp: "the audit will pick a random node" should probably be "a configurable number of random nodes"

docs/design/audit-v2.md
```diff
@@ -0,0 +1,93 @@
...
+
+After selection, the rest of the auditing process will occur the same way as it does currently: picking a random segment, picking a random stripe, downloading all erasure shares associated with that stripe and using Berlekamp-Welch algorithm (via the Infectious library) to verify that they haven't been altered.
+The chances of selecting the same stripe are rare, but it's normal/expected behavior if occasionally a stripe is audited more than once.
+
+We plan to run a simulation for this new method of auditing so we can estimate appropriate settings around reservoir sampling. If we decide that we want to prefer nodes with less audits, then we will implement the power of two choices, in which we randomly select two nodes, then choose the one with less audits.
+This would still require tracking number of audits, but it would prevent having to sort and query all nodes by audit count, which could cause undesirable behavior. For example, when new nodes join the network, the audit system could become stuck auditing only new nodes, and ignoring more established nodes.
```
wthorp: or we could half-heartedly track audit counts in-memory

docs/design/audit-v2.md
navillasa: okay- so "The audit will pick a configurable number of random nodes, then a random segment from one of the nodes' reservoirs"?

wthorp: I'm just saying that the wording is vague.  Do we loop through all of metainfo db and build a reservoir cache for each node only to audit a single one of them?  Probably not.  We probably want to audit (a random segment each) from _MANY_ of them.

navillasa: ahh that totally makes sense, will clarify.

navillasa:
went ahead and merged since @thepaul and I reviewed yesterday. @jtolds if you have any outstanding feedback, I can still go ahead and add it!

docs/design/audit-v2.md
```diff
@@ -0,0 +1,93 @@
...
+n001 + + + + + + +
+n002 + + + +
+n003 + + + +
+n004 + + + + + + +
+
+Unvetted nodes should get 25,000 pieces per month. On a good day, there will be 1000 pieces added to an unvetted node, which should quickly fill the reservoir sample.
+
+• We have a reservoir of k items and a stream of n items, where n is an unknown number.
+• Fill the reservoir from [0...k-1] with the first k items in the stream
+• For every item in the stream from index i=k..n-1, pick a random number j=rand(0..i), and if j<k, replace reservoir[j] with stream[i]
```
kaloyan-raev: It seems this bullet list does not format well in rendered mode.

docs/design/audit-v2.md
```diff
@@ -0,0 +1,93 @@
...
+We are expecting close to 3 audits per day for unvetted nodes. The satellite currently issues one audit every 30 seconds, the current interval, which gives us close to 3,000 audits per day. There are about 1,000 nodes on the network currently. This could mean that the default size of a reservoir should be three, if we assume a full iteration of reservoirs in one day.
+
+### Selection via Reservoir Sampling
+
+Reservoir sampling: a family of algorithms for randomly choosing a sample of items with uniform probability from a stream of data (of unknown size).
+
+As the audit observer uses the metainfo loop to iterate through the pointerdb, we're going through all segments and creating reservoirs per node, and filling the reservoirs with segments to audit.
+
+In order to increase the amount of audits for unvetted nodes, we can build larger reservoir samples for unvetted nodes.
+Two configurations for reservoir sampling: number of segments per unvetted nodes, and for vetted.
```
kaloyan-raev: Reading more about reservoir sampling, I think we can actually have it working with only one single slot per reservior.

The audit observer will constantly update the reservoirs as the metainfo loop iterate over the metainfo db. Due to the nature of reservior sampling, this should give an equal probability for each relevant pointer to be in the reservior at the time the audit service makes a selection.

When the audit service selects a node for auditing, it will just pick the pointer that is currently in the node's reservoir, then select a random stripe from it, etc.

Having just one slot in the reservoir will save some memory on the satellite and will simplify the implementation.

mobyvb: We can configure the reservoir in the current implementation to just have a size of 1. Then we can just adjust that if we need to.

kaloyan-raev: The question should we go with a more complex implementation if a single slot reservior is sufficient?

mobyvb: Do you think the implementation would be significantly simpler if we limited ourselves to a sample size of 1?
I think since most of it is already written, we would just have to adjust a config value to do a reservoir size of 1. Then if we ever decided that needs to change, it will be easy. I like the more flexible implementation.

kaloyan-raev: If we go with a single slot reservoirs then a simple `map[storj.NodeID]storj.Path` can hold all reservoirs.

mobyvb: We still need to keep track of the number of elements seen for each reservoir. And the calculation of the random index would be the same as it is the other way.

navillasa: it's already not a huge amount of code so I would also be in favor of keeping the implementation adjustable to increase the reservoir size. we also still need to do the simulation thing to make more decisions about config.

kaloyan-raev: OK. You are right about this. Perhaps, it makes sense to have a generic Reservoir type and use it with size 1 for auditing.

PR #2704, opened 2019-08-02
 store storage node public cert for future PieceHash verification during upload on satellite

aligeti:
What: 
Adds peer identity info  of the nodes(uplinks and Storagenodes)  to certRecorddb. Uplink's peer identity is added at the time of order limit creation and Storagenode's peer identity is created during kademlia discovery. ( I understand kademlia is under refactor, but got the most of the framework of storing the peer identity into the certRecord DB in place). At the time of implementing this ticket the only place where the satellite can extract Storagenode's peer  identity is in Kademlia. When that changes only this part changes and rest of the frame work doesn't need any major changes.

Why:
V3-1992 ticket  (Part - II)

Please describe the tests:
 - Test 1: 
 - Test 2:
 

pkg/kademlia/kademliaclient/kademliaclient.go
```diff
@@ -147,21 +147,27 @@ func (dialer *Dialer) FetchPeerIdentityUnverified(ctx context.Context, address s
...
 	}
 	defer dialer.limit.Unlock()
 
 	conn, err := dialer.dialNode(ctx, target)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
+	}
+	p := &peer.Peer{}
+	_, err = conn.client.Ping(ctx, &pb.PingRequest{}, grpc.Peer(p))
```
egonelbre: Since we are removing kademlia, there may not be a ping available.

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
+	chain = append(chain, pi.CA.Raw...)
+	for _, cert := range pi.RestChain {
+		chain = append(chain, cert.Raw...)
 	}
-	return pubkey, nil
+	return chain
 }
 
-// GetPublicKeys gets the public keys of a storagenode corresponding to storagenode id
-func (certs *certDB) GetPublicKeys(ctx context.Context, nodeID storj.NodeID) (pubkeys []crypto.PublicKey, err error) {
```
aligeti: **Rational for removing `GetPublicKeys()`:**
Didn't see a scenario of needing more than one peer-identity for validating. This certRecord DB does store all previous peer-identities for a particular Storagenode, if it decides to have a new identity.

satellite/discovery/service.go
```diff
@@ -131,6 +134,11 @@ func (discovery *Discovery) refresh(ctx context.Context) (err error) {
 				if _, err = discovery.cache.UpdateNodeInfo(ctx, node.Id, info); err != nil {
 					discovery.log.Warn("could not update node info", zap.Stringer("ID", node.GetAddress()))
 				}
+
+				// update/add the certdb with nodeID's public key
+				if err = discovery.certdb.SavePublicKey(ctx, node.Id, ident); err != nil {
```
aligeti: **`Rational for calling a SavePublicKey()`**
At the time of implementing this ticket the only place where the satellite can extract Storagenode's peer identity is in Kademlia. When Kad implementation changes, only this part changes and rest of the frame work doesn't need any major changes.

pkg/kademlia/kademliaclient/kademliaclient.go
egonelbre: `transport.Client` `FetchPeerIdentity` can also fetch the peer identity.

Even, if it didn't exist, we would need to implement it.

aligeti: updated to use FetchPeerIdentity

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
-	_, err = certs.db.Create_CertRecord(ctx,
-		dbx.CertRecord_Publickey(pubbytes),
-		dbx.CertRecord_Id(nodeID.Bytes()),
-	)
-	// another goroutine might race to create the cert record, let's ignore that error
-	if pgutil.IsConstraintError(err) || sqliteutil.IsConstraintError(err) {
-		return nil
-	} else if err != nil {
+	var id int64
+	query := `SELECT id FROM certRecords WHERE peer_identity = ?;`
```
egonelbre: Let's use the serial number to verify whether it needs updating otherwise this will end up really slow.

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
 		return Error.Wrap(err)
 	}
 
+	// already public key exists, just return
 	return nil
 }
 
 // GetPublicKey gets the public key of uplink corresponding to uplink id
-func (certs *certDB) GetPublicKey(ctx context.Context, nodeID storj.NodeID) (_ crypto.PublicKey, err error) {
+func (certs *certDB) GetPublicKey(ctx context.Context, nodeID storj.NodeID) (_ *identity.PeerIdentity, err error) {
```
egonelbre: See notes in https://github.com/storj/storj/pull/2690#discussion_r310639785. This is sufficient for this, however we should design it such that we can avoid these large requests.

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
 	defer mon.Task()(&ctx)(&err)
-	dbxInfo, err := certs.db.All_CertRecord_By_Id_OrderBy_Desc_UpdateAt(ctx, dbx.CertRecord_Id(nodeID.Bytes()))
+	dbxInfo, err := certs.db.All_CertRecord_By_NodeId_OrderBy_Desc_UpdateAt(ctx, dbx.CertRecord_NodeId(nodeID.Bytes()))
 	if err != nil {
 		return nil, err
```
egonelbre: Wrapping missing.

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
 	}
 
+	if len(dbxInfo) == 0 {
+		return nil, Error.New("Invalid nodeID : %+v: %+v ", nodeID.String(), err)
```
egonelbre: `return nil, Error.New("Unknown NodeID %v: %v", err)`

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
+	// already public key exists, just return
 	return nil
 }
 
 // GetPublicKey gets the public key of uplink corresponding to uplink id
-func (certs *certDB) GetPublicKey(ctx context.Context, nodeID storj.NodeID) (_ crypto.PublicKey, err error) {
+func (certs *certDB) GetPublicKey(ctx context.Context, nodeID storj.NodeID) (_ *identity.PeerIdentity, err error) {
 	defer mon.Task()(&ctx)(&err)
-	dbxInfo, err := certs.db.All_CertRecord_By_Id_OrderBy_Desc_UpdateAt(ctx, dbx.CertRecord_Id(nodeID.Bytes()))
+	dbxInfo, err := certs.db.All_CertRecord_By_NodeId_OrderBy_Desc_UpdateAt(ctx, dbx.CertRecord_NodeId(nodeID.Bytes()))
```
egonelbre: Why use `All` when there can only be one answer?

satellite/certdb/certdb_test.go
```diff
@@ -31,75 +32,63 @@ func testDatabase(ctx context.Context, t *testing.T, upldb certdb.DB) {
...
-
-		{ // add another key for the same storagenode ID, this the latest key
-			err := upldb.SavePublicKey(ctx, sn1ID.ID, sn2ID.Leaf.PublicKey)
-			assert.NoError(t, err)
+		{ // adding two different pubkeys for same storagnode
+			{ // add a key for to storagenode ID
+				err := upldb.SavePublicKey(ctx, sn2PI.ID, sn1PI)
+				assert.NoError(t, err)
+			}
+			time.Sleep(5)
```
egonelbre: Remove the sleep.

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +22,96 @@ type certDB struct {
...
+	for _, cert := range pi.RestChain {
+		chain = append(chain, cert.Raw...)
 	}
-	return pubkey, nil
+	return chain
 }
 
-// GetPublicKeys gets the public keys of a storagenode corresponding to storagenode id
-func (certs *certDB) GetPublicKeys(ctx context.Context, nodeID storj.NodeID) (pubkeys []crypto.PublicKey, err error) {
+func decodePeerIdentity(ctx context.Context, chain []byte) (_ *identity.PeerIdentity, err error) {
```
egonelbre: These should be probably in `pkg/identity` with tests.

satellite/certdb/certdb.go
```diff
@@ -5,17 +5,15 @@ package certdb
...
 
+	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/storj"
 )
 
 // DB stores uplink public keys.
 type DB interface {
 	// SavePublicKey adds a new bandwidth agreement.
-	SavePublicKey(context.Context, storj.NodeID, crypto.PublicKey) error
+	SavePublicKey(context.Context, storj.NodeID, *identity.PeerIdentity) error
```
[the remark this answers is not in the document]
aligeti: Changed to Set and Get

satellite/satellitedb/migrate.go
```diff
@@ -1076,9 +1076,13 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 				Description: "Changing the primary key constraint",
 				Version:     50,
 				Action: migrate.SQL{
-					`ALTER TABLE certRecords DROP CONSTRAINT certrecords_pkey;
-					ALTER TABLE certRecords ADD CONSTRAINT certrecords_pkey PRIMARY KEY (publickey);
-					CREATE INDEX certrecord_id_update_at ON certRecords ( id, update_at );`,
+					`ALTER TABLE certRecords RENAME COLUMN id TO node_id;
+					ALTER TABLE certRecords RENAME COLUMN publickey TO peer_identity;
```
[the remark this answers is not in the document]
aligeti: I am bad in picking names... :-) If you have an appropriate name please suggest, I am open and will rename.the  names that comes to my thoughts are 'identityRecords'? or 'peerRecords' or peerIdentities. Based on you comment/suggestion, I will drop the table and call it peerIdentities table and will have the following fields. 

This is what the new table looks like. Please let me know what your thoughts are....
//--- peerIdentities ---//

model peerIdentities (
    key peer_identity

	index (
		name peerIdentities_node_id_update_at
		fields node_id update_at 
	)
    
    field serial_number  blob
	field peer_identity  blob
	field node_id        blob
	field update_at timestamp ( autoinsert, autoupdate )
)

create peerIdentities ( )
delete peerIdentities ( where peerIdentities.node_id = ? )

read one (
	select peerIdentities
	where  peerIdentities.serial_number = ?
	orderby desc peerIdentities.update_at
)

read one (
	select peerIdentities
	where  peerIdentities.peer_identity = ?
)

read all (
	select peerIdentities
	where  peerIdentities.node_id = ?
	orderby desc peerIdentities.update_at
)

satellite/satellitedb/certdb.go
aligeti: are you saying that there will be only one entry per nodeID? if already a nodeID has an peerIdenitty entry in the table and a new one comes in, will it get updated? or should I make a new entry, because we need that nodeID's previous peer identities as well? what's is the requirement. 

or I am not completely savoy in identities concept, but based on leaf concept and my crude 10000 ft level understanding, if the node(SNO) happened to change its identity, then it just gets added to it previous ones as leaf and keeps the NodeID same. If this is my correct understanding than , there will be only one entry per nodeID and if the peeridentiy column is different , then I just update it. Is my understanding right?

satellite/satellitedb/certdb.go
aligeti: I looked at the comment 
"Note, we always load it from the database otherwise it'll add massive amount of traffic to the database.

Similarly there should be one DB call per verification rather than ~100."

am lost on your first comment. What I see is @aleitner is calling GetPublicKey() with parameters and it calls the DB's related function to make a query and get the appropriate info. I am bit lost here. I am at the DB level which makes call directly to DB and gets the information to other higher calling functions.. .. am like an API call. I must certainly be missing something ...

satellite/satellitedb/certdb.go
aligeti: will add the wrapper...

satellite/certdb/certdb_test.go
aligeti: will do .. I added to create a little time different in two entries with same nodeID but different peer identity. But again might change based on my above comment, if only one nodeID exists or can there be more than one entry with same nodeIDs but different peer identities...

satellite/satellitedb/certdb.go
aligeti: will move to pkg/identity and add tests

satellite/satellitedb/certdb.go
aligeti: fixed

satellite/certdb/certdb.go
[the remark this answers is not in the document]
egonelbre: I cannot see the changes.

go.sum
```diff
@@ -366,9 +365,7 @@ github.com/spf13/pflag v1.0.3 h1:zPAT6CGy6wXeQ7NtTnaTerfKOsV6V6F8agHXFiazDkg=
 github.com/spf13/pflag v1.0.3/go.mod h1:DYY7MBk1bdzusC3SYhjObp+wFpr4gzcvqqNjLnInEg4=
 github.com/spf13/viper v1.4.0 h1:yXHLWeravcrgGyFSyCgdYpXQ9dR9c/WED3pg1RhxqEU=
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
-github.com/storj/grpc-go v1.22.1-0.20190808142608-8c6988074762/go.mod h1:iGXtpppuCBUYpHRAS+WAD5sZH3LsRjHdqV0iDN+I0nY=
-github.com/storj/grpc-go v1.22.1-notrace h1:cLzaU3QNAHmweqqfXl0cew09ifL4CC2ewXjzmGm8xak=
-github.com/storj/grpc-go v1.22.1-notrace/go.mod h1:iGXtpppuCBUYpHRAS+WAD5sZH3LsRjHdqV0iDN+I0nY=
+github.com/storj/grpc-go v1.22.2-0.20190808142608-8c6988074762 h1:zwA+hIzyPvRd6GExwri4BM8aIGxu28TVFhwnmZrrK+4=
```
aligeti: I am had problems building .. I will fix this... I am aware ...

go.sum
```diff
@@ -536,6 +515,5 @@ gopkg.in/yaml.v2 v2.2.1 h1:mUhvW9EsL+naU5Q3cakzfE91YhliOondGd6ZrsDBHQE=
 gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
-honnef.co/go/tools v0.0.0-20180728063816-88497007e858/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
```
aligeti: I am had problems building .. I will fix this... I am aware ...

satellite/satellitedb/migrate.go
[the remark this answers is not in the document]
mniewrzal: for DB names use `_`, so it will be more `peer_identity`

satellite/satellitedb/certdb.go
mniewrzal: Less requests to DB is better for satellite performance. The comment is about making method to get identities for multiple peers with one DB request (sql).  E.g.
`Get(ctx context.Context, nodeIDs []storj.NodeID) (_ []*identity.PeerIdentity, err error) {`

For example then you can ask for all nodes identities from segment piece hashes in one call.

aligeti: Thanks for the clarification... I got it.. updated the code support batch Get of all peeridentites like a transaction.

satellite/satellitedb/certdb.go
aligeti: updated

pkg/kademlia/kademliaclient/kademliaclient.go
```diff
@@ -158,9 +158,7 @@ func (dialer *Dialer) FetchInfo(ctx context.Context, target pb.Node) (_ *pb.Info
 	if err != nil {
 		return nil, err
 	}
-
 	resp, err := conn.client.RequestInfo(ctx, &pb.InfoRequest{})
-
```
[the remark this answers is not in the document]
aligeti: updated

satellite/certdb/certdb.go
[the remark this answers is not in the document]
aligeti: I am not sure why it didn't show here.. but the changes are made and they are in. Please see  ^^ "files Changed" link and you should see the requested changes.

pkg/kademlia/kademlia.go
```diff
@@ -272,7 +272,7 @@ func (k *Kademlia) FetchInfo(ctx context.Context, node pb.Node) (_ *pb.InfoRespo
 
 	info, err := k.dialer.FetchInfo(ctx, node)
 	if err != nil {
-		return nil, NodeErr.Wrap(err)
+		return nil, NodeErr.Wrap(NodeErr.New("%s : %s failed to fetch info from node ID %s: %s", k.routingTable.self.Type.String(), k.routingTable.self.Id.String(), node.Id.String(), err))
```
egonelbre: See comments in https://github.com/storj/storj/pull/2668

satellite/satellitedb/certdb.go
```diff
@@ -19,82 +19,85 @@ type certDB struct {
...
-	dbxInfo, err := certs.db.All_CertRecord_By_Id_OrderBy_Desc_UpdateAt(ctx, dbx.CertRecord_Id(nodeID.Bytes()))
-	if err != nil {
-		return nil, err
+	if len(nodeIDs) == 0 {
+		return nil, nil
 	}
 
-	// the first indext always holds the lastest of the keys
-	pubkey, err := pkcrypto.PublicKeyFromPKIX(dbxInfo[0].Publickey)
+	tx, err := certs.db.Open(ctx)
```
egonelbre: Don't use a transaction here, there's no need for it, query all at the same time instead.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -2,164 +2,164 @@
 
 //--- Value Attribution ---//
 model value_attribution (
-	key project_id bucket_name
+    key project_id bucket_name
```
egonelbre: Don't make other changes in the same PR.

egonelbre:
Also this PR is getting really large.

satellite/satellitedb/certdb.go
aligeti: made change to request all nodeIDs peer identities at the one shot

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: that was unintentional . I will revert it.

PR #2705, opened 2019-08-03
satellite/marketingweb: make display order consistent

VinozzZ:
What: 
- display for partner offers needs to be consistent

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/rewards/display.go
```diff
@@ -78,9 +80,18 @@ func (offers Offers) OrganizeOffersByType() OfferSet {
...
+	keys := make([]string, len(partners))
+	i := 0
+	for k := range partners {
+		keys[i] = k
+		i++
+	}
+
+	sort.Strings(keys)
+
+	for _, key := range keys {
```
coyle: why not range partners here and just set to the key

satellite/rewards/display.go
```diff
@@ -78,9 +80,18 @@ func (offers Offers) OrganizeOffersByType() OfferSet {
 func createPartnerSet() PartnerSet {
 	partners := LoadPartnerInfos()
 	var ps PartnerSet
-	for _, partner := range partners {
+	keys := make([]string, len(partners))
+	i := 0
+	for k := range partners {
```
wthorp: It looks like you could have written the more typical `for i := 0; k := range partners; i++ {`

satellite/rewards/display.go
wthorp: because the sort is needed, I think.  another option would be https://gobyexample.com/sorting-by-functions

satellite/rewards/display.go
wthorp: I think this would be easier to read as: `for i := 0; i < len(partners); i++{ keys[i] = partners[i] }`

VinozzZ: `partners` is a map so I can't use index as the key to access the value

PR #2707, opened 2019-08-03
storagenode: keep live sum of piecefiles in memory

JessicaGreben:
**What/Why:**

When a piece is uploaded to a storagenode, it first calculates how much free space it has available by iteratively walking through all piecefiles (pieces are stored as files on the storagenode, so we call them piecefiles) stored on disk, stats each piecefile to get size, then sums up those sizes. This is very slow. 

The goal of this PR is to improve the performance by storing live used-space totals in-memory. To accomplish that this PR does makes the following changes:
- adds a data structure to the storagenode that contains the live used-space totals
- when the storagenode starts, the initial live used-space totals are calculated 
- when pieces are created or deleted, the live used-space totals are updated
- the storagenodes services that used the previous iterative approach to calculate totals are now updated to use the in-memory live used-space values instead. This includes `storagenode/monitor`, `storagenode/inspector`, and `storagenode/console`
- adds a loop to the `storagenode/monitor` service that recalculates the in-memory used-space totals once an hour

**Describe the tests:**
WIP
 
**Describe the performance impact:**
The performance should be improved now that we are using a value in memory instead of iterating over every piecefile on disk.

**Open questions:**
- currently the monitor service will run once an hour and recalculate and reset the in-memory used space values.  Is this what we want? One scenario I'm worried about happening is if it takes a very long time to recalculate and over that time there are many piece additions/deletions that are not accounted for in the recalculation, potentially when the in-memory values are reset the new values could be more inaccurate than the original value.


storagenode/pieces/store.go
```diff
@@ -250,6 +276,28 @@ func (store *Store) Delete(ctx context.Context, satellite storj.NodeID, pieceID
...
+	}
+	return pieceAccess.ContentSize(ctx)
+}
+
+// UpdateLiveUsedSpaceTotals updates the live used space totals
+// with a pieceSize that was either created or deleted where the pieceSize is
+// only the content size and does not include header bytes
+func (store *Store) UpdateLiveUsedSpaceTotals(ctx context.Context, satelliteID storj.NodeID, pieceSize int64) {
+	store.liveUsedSpace.mu.Lock()
+	defer store.liveUsedSpace.mu.Unlock()
```
VinozzZ: Just curious what's the benefit of using defer than putting the `Unlock` after line 298

JessicaGreben: in this case, i don't think it matters either way. I guess I just kept it like this for consistency since its how i use it in other places.

VinozzZ: How about in general? Is it because we want to keep the `Lock` and `Unlcok` close so we know for sure the `Unlock` is called?

thepaul: One reason to use `defer` in cases like this is that the defered function is still run when something down the call stack calls `panic()`. If we called `Unlock` normally at the end of the function, and there was a panic, the lock might never be unlocked. That hopefully isn't an issue here, because our code doesn't do much recovering from panics, but it could be elsewhere.

JessicaGreben: I do think using defer to unlock makes that pretty clear. Also, for example, here https://github.com/storj/storj/pull/2707/files#diff-67139f1d1f58f7706b076d08baba9ec4R464 we return a value and want to unlock after so the defer helps with that. However, its best to minimize the time of locking, so you wouldn't want to use the defer to unlock if you don't need the lock for the rest of the function

VinozzZ: Thank you for explaining it! These are great info to know

thepaul:
Yeah, that's a good thing to be concerned about. The iteration might indeed take a long time. Longer than an hour in really bad cases, I would guess. (Maybe we want to have it run less often, or wait for the specified interval between the _end_ of one run and the start of the next, instead of having it be the interval between starts?)

We won't be able to make the iteration give the exact guaranteed correct value, ever. But maybe we can get pretty close. What about an algorithm like this-

* when the iteration operation starts, tell the live-summing system to save a copy of its current count as `totalSpaceCountWhenIterationStarted`
* the live-summing continues as usual while the iteration proceeds
* when iteration is done, we ask the live-summing system to tell us the delta between its current count and `totalSpaceCountWhenIterationStarted`. This `spaceDeltaDuringIteration` is (we assume) how much change has been applied to the filesystem while we were iterating.
* statistically speaking, we can expect that half of the blob writes and half of the blob deletes that happened during iteration were done to blobs that were not yet traversed by the iteration, and half of the writes and deletes were done to blobs that _were_ traversed by the iteration. So `resultOfIteration` has already counted half of the written blob sizes and already doesn't include half of the deleted blob sizes. With this expectation, we will cut `spaceDeltaDuringIteration` in half.
* we add `spaceDeltaDuringIteration / 2` to `resultOfIteration` and that becomes our new baseline.

Does that seem like a sane thing to do? I haven't thought through it very well.

JessicaGreben:
@thepaul yeah that sounds like it will be a good estimate. I work on implementing that.

storagenode/pieces/store.go
```diff
@@ -132,20 +136,35 @@ type Store struct {
 	reservedSpace int64
 }
 
+// SpaceUsedLive stores the live totals of used space for all
+// piece content (not including headers) and also total used space
+// by satellites
+type SpaceUsedLive struct {
```
egonelbre: Let's implement this caching as wrapper instead of complicating the existing code.

storagenode/pieces/store.go
```diff
@@ -132,20 +136,35 @@ type Store struct {
...
-	return &Store{
+func NewStore(log *zap.Logger, blobs storage.Blobs, v0PieceInfo V0PieceInfoDB, expirationInfo PieceExpirationDB) (*Store, error) {
+	newStore := Store{
 		log:            log,
 		blobs:          blobs,
 		v0PieceInfo:    v0PieceInfo,
 		expirationInfo: expirationInfo,
+		spaceUsedLive:  SpaceUsedLive{},
+	}
+	err := newStore.InitSpaceUsedLive()
```
egonelbre: This is not handling context cancellation, since this may take a long time.

storagenode/pieces/store.go
JessicaGreben: a wrapper around what?

storagenode/pieces/store.go
JessicaGreben: what migration? this function is iterating over all the piecefiles so i don't think there is a migration for that.

storagenode/pieces/store.go
JessicaGreben: added commit to address this comment and add a wrapper: https://github.com/storj/storj/pull/2707/commits/296ccc47ff0e8cf56f49d541cfe6d56ea595bbda

storagenode/pieces/store.go
JessicaGreben: added commit to address this comment and have initialization occur after db migration: https://github.com/storj/storj/pull/2707/commits/296ccc47ff0e8cf56f49d541cfe6d56ea595bbda

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+	"sync"
+
+	"go.uber.org/zap"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+// StoreWithCache has access to the pieces store and also keeps track of real time
+// totals of space used for all pieces and for all pieces by satellite
+type StoreWithCache struct {
```
thepaul: Do you think it's necessary to have this be a new type, as opposed to adding this functionality to the main piece store? It seems like most services shouldn't need to care whether their piece stores have space-used caches inside or not.

thepaul: Maybe we could have some kind of SpaceUsedGetter interface, and `pieces.Store` has one of those, and it abstracts how the space-used calculation happens, whether it's cached, how often the cache is updated, etc?

JessicaGreben: originally i had it added to the main piece store, but I changed it this way based on Egon's comment here https://github.com/storj/storj/pull/2707#discussion_r310641344

im open to either way.

thepaul: Oh. Did he clarify to you what he meant outside of the PR? If he wants a wrapper around all of pieces.Store, maybe he has a rationale that hasn't occurred to me.

thepaul: I.e., I wonder if maybe he meant a wrapper around `SpaceUsed()` and `SpaceUsedInNamespace()` themselves?

JessicaGreben: yeah i asked him what he meant by "wrapper" and he linked an example that isaac did that used embedding. So I believe he was saying to embed the Store in a wrapper that has caching.  Her is the example: https://github.com/storj/storj/pull/2681/files#diff-2196cdf1c7f5a0b80f2b17bf167406b5R23

thepaul: Huh. That seems wrong to me in this case. Maybe a wrapper around `filestore.Store` would make more sense? Or perhaps we should wait for @egonelbre to clarify once he's back online.

cmd/storagenode/main.go
```diff
@@ -161,6 +161,11 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
 		return errs.New("Error creating tables for master database on storagenode: %+v", err)
 	}
 
+	err = peer.Storage2.Store.InitCache()
```
thepaul: I think we may want to avoid blocking startup here while we wait for a full dir iteration. Maybe this call should just kick off a fresh calculation instead of blocking, or (my vote) we should persist the most recent full count to disk and load it here as our space-used starting point.

JessicaGreben: i changed it to this based on a comment from egon https://github.com/storj/storj/pull/2707#discussion_r310642467.

But yeah blocking on startup could take a long time if there is a lot stored, so maybe this isn't ideal.  A fresh calculation could work except that we would either need to halt access to the cache value until the calculation is complete or be ok with is being a zero value and potentially very inaccurate.

So maybe persisting the most recent count is ideal. I will look into adding that.

storagenode/pieces/cache.go
JessicaGreben: yeah maybe i misunderstood something.  would be nice to get egon's take on this.

storagenode/monitor/monitor.go
```diff
@@ -127,6 +124,11 @@ func (service *Service) Run(ctx context.Context) (err error) {
 		if err != nil {
 			service.log.Error("error during updating node information: ", zap.Error(err))
 		}
+
+		err = service.RecalculateSpaceUsedLive(ctx)
```
thepaul: It looks like this is the only place that calls `store.RecalculateSpaceUsedCache()` (by way of `service.RecalculateSpaceUsedLive()`). I don't think the monitor service should be the one responsible for managing the freshness of the cache inside the piece store. For one thing, the piece store service should continue working correctly even if there isn't a monitor service running at all. Maybe we should have a new `sync2.Cycle` object owned by the store-wrapper object itself (or maybe owned by whatever object ends up owning the cache object, if that structure changes).

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+	Store
+
+	cacheMutex sync.RWMutex
```
thepaul: I may have missed it, but I don't think we use `RLock()` on `cacheMutex` anywhere. This should be a plain `sync.Mutex` instead. Partially to avoid confusion ("but why is it an RWMutex?") but also because Mutex.Lock is faster than RWMutex.Lock; it doesn't need to make the extra check for existing read-locks.

You could also change some places to use `RLock()` instead of `Lock()`, but I don't think the expected level of lock contention or the expected hold durations justify it.

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+
+func estimateRecalculation(newSpaceUsedTotal, spaceUsedWhenIterationStarted, spaceUsedWhenIterationEnded int64) int64 {
+	if newSpaceUsedTotal == spaceUsedWhenIterationEnded {
+		return newSpaceUsedTotal
+	}
+
+	// If we missed writes/deletes while iterating, we will assume that half of those missed occurred before
+	// the iteration and half occurred after. So here we add half of the delta to the result space used totals
+	// from the iteration to account for those missed.
+	spaceUsedDeltaDuringIteration := spaceUsedWhenIterationStarted - spaceUsedWhenIterationEnded
```
thepaul: It's late and I'm totally not sure, but is this backward? I think we want `spaceUsedWhenIterationEnded - spaceUsedWhenIterationStarted`, because a positive delta should mean an increase in space used (since we add half of it to `newSpaceUsedTotal`)

JessicaGreben: hmm good catch. let me double check which way it should be

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+
+	store.cache.totalBySatellite = estimatedTotalsBySatellites
+	store.cache.total = estimateRecalculation(newSpaceUsed.total,
+		oldSpaceUsed.total,
+		store.cache.total,
+	)
+}
+
+func estimateRecalculation(newSpaceUsedTotal, spaceUsedWhenIterationStarted, spaceUsedWhenIterationEnded int64) int64 {
+	if newSpaceUsedTotal == spaceUsedWhenIterationEnded {
```
thepaul: This shortcut might actually be wrong. Check me on this:

1. spaceUsedWhenIterationStarted = 10
2. meanwhile on disk there's actually 30 bytes being used. 20 of bytes just didn't get counted for whatever reason
3. while iterating over blob directories, 4 blobs get written at 10 bytes each
4. half of those blobs are noticed by the directory iteration. it returns 50 (30 + 2 * 10)
5. spaceUsedWhenIterationEnded = 50 (10 + 4 * 10)
6. estimateRecalculation sees 50 == 50, shortcuts, returns 50
7. but the right answer would have been 70 (30 + 4 * 10), which we would have gotten if we had followed through the half-delta adding below (50 + (50 - 10) / 2)

cmd/storagenode/main.go
egonelbre: Yeah, we want to avoid blocking startup as much as possible. The question is what is the implication of not having up to date information about this.

storagenode/pieces/cache.go
egonelbre: Usually you want to wrap the smallest thing possible to add the functionality. This ensures that you don't have to forward calls that much.

The reason I suggested trying wrapping is that it might preserve the rest of the implementation clear of this concern. I.e. you would be able to remove / add / replace the caching with some other approach when needed. Similarly you can change the store without worrying breaking caching. In principle it keeps clarity between "this is what we cache", "this is how we store" and "this is how we do X / Z".

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+// StoreWithCache has access to the pieces store and also keeps track of real time
+// totals of space used for all pieces and for all pieces by satellite
+type StoreWithCache struct {
+	Store
+
+	cacheMutex sync.RWMutex
+	cache      spaceUsedCache
+}
+
+type spaceUsedCache struct {
```
egonelbre: You can inline these above since there is a separate struct for representing the idea of a cache.

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+// get initial values for the storage cache totals of space used
+func (store *StoreWithCache) InitCache() error {
+	newCacheTotals, err := store.spaceUsedForPiecesAndBySatellitesSlow()
+	if err != nil {
+		return err
+	}
+
+	store.cacheMutex.Lock()
+	defer store.cacheMutex.Unlock()
+	store.cache.total = newCacheTotals.total
```
egonelbre: You can assign these by struct as well `store.cache = newCacheTotals`.

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,163 @@
...
+	return nil
+}
+
+// spaceUsedForPiecesAndBySatellitesSlow iterates over all the pieces stored on disk
+// and sums the bytes for all pieces (not including headers) currently stored. These
+// new summed up values are used for the live in memory values.
+// The size of each piece is measuered sequentially and writes/deletes still occur while iterating.
+// Any writes/deletes that occur late in the process are likely to be missed and therefore the result of this
+// method is an estimate.
+func (store *StoreWithCache) spaceUsedForPiecesAndBySatellitesSlow() (spaceUsedCache, error) {
```
egonelbre: `recalculateSpaceUsed`

JessicaGreben:
@wthorp this PR is still WIP and i have a bunch of changes to make. I just closed the PR in the meantime, will reopen once i have stuff in working order.

PR #2711, opened 2019-08-05
uplink/ecclient: read concurrently with dials during download

cam-a:
What: This PR implements a new lazy io.ReadCloser `lazyPieceReader` which dials a node and reads the download immediately

Why: Previously, we would wait until all dials were complete before reading. In the event of a long tail the process would be blocked

Please describe the tests:
 - Test 1:
 - Test 2:
 

uplink/ecclient/client.go
```diff
@@ -459,19 +459,61 @@ func (lr *lazyPieceRanger) Size() int64 {
...
-	if err != nil {
-		return nil, errs.Combine(err, ps.Close())
+	if lr.dialed && lr.Downloader == nil {
+		return 0, io.EOF
 	}
-	return &clientCloser{download, ps}, nil
+
+	return lr.Downloader.Read(data)
+}
+func (lr *lazyPieceReader) Close() error {
```
egonelbre: We may want to add a protection against calling read after close.

uplink/ecclient/client.go
```diff
@@ -479,9 +479,9 @@ type lazyPieceReader struct {
 	client *piecestore.Client
 }
 
-func (lr *lazyPieceReader) Read(data []byte) (int, error) {
+func (lr *lazyPieceReader) Read(data []byte) (_ int, err error) {
+	mon.Task()(&lr.ctx)(&err)
```
egonelbre: Add separate tracing for the dial rather than the whole func.

uplink/ecclient/client.go
cam-a: What about a bool on the struct indicating whether it has been closed?

egonelbre: Yeah, that would be fine.

uplink/ecclient/client.go
```diff
@@ -473,30 +474,73 @@ func (lr *lazyPieceRanger) Size() int64 {
...
-func (client *clientCloser) Close() error {
+func (lr *lazyPieceReader) Read(data []byte) (_ int, err error) {
+	lr.mu.RLock()
+	defer lr.mu.RUnlock()
+
+	if lr.closed {
+		return 0, io.EOF
+	}
+	if !lr.dialed {
+		mon.Task()(&lr.ctx)(&err)
```
kaloyan-raev: How about extracting the dialing logic in a separate method?

egonelbre: Also defer is missing.

uplink/ecclient/client.go
```diff
@@ -473,30 +474,81 @@ func (lr *lazyPieceRanger) Size() int64 {
...
-type clientCloser struct {
-	piecestore.Downloader
-	client *piecestore.Client
-}
+func (lr *lazyPieceReader) Close() error {
+	lr.mu.Lock()
+	defer lr.mu.Unlock()
+	lr.closed = true
+
+	if lr.Downloader == nil {
```
egonelbre: We can check here lr.closed now instead, rather than `Downloader`.

uplink/ecclient/client.go
```diff
@@ -473,30 +474,81 @@ func (lr *lazyPieceRanger) Size() int64 {
...
+	lr.dialed = true
+	ps, err := lr.dialPiecestore(lr.ctx, &pb.Node{
 		Id:      lr.limit.GetLimit().StorageNodeId,
 		Address: lr.limit.GetStorageNodeAddress(),
 	})
 	if err != nil {
-		return nil, err
+		return err
 	}
+	lr.client = ps
```
isaachess: It looks like the only reason we're holding on to lr.client is to close it on a close call. I suggest we add a field on `lr` called `closers`, like:

`closers []io.Closer`

Then, you add each:

```
lr.closers = append(lr.closers, downloader)
lr.cloers = append(lr.closers, client)
```

Then in the close function:

```
for _, closer := range lr.closers {
    err = errs.Combine(err, closer.Close())
}
return err
```

You know, now that I've written it out I feel more ambivalent about it. I'm going to leave the comment so you can see the idea, the nice thing is you're not holding onto specific member variables that only exist to be closed (at the risk of forgetting later!). But I'm sixes on it.

uplink/ecclient/client.go
```diff
@@ -473,30 +474,81 @@ func (lr *lazyPieceRanger) Size() int64 {
...
+	return nil
 }
 
-type clientCloser struct {
-	piecestore.Downloader
-	client *piecestore.Client
-}
+func (lr *lazyPieceReader) Close() error {
+	lr.mu.Lock()
+	defer lr.mu.Unlock()
```
isaachess: Is the purpose of mutex to handle a concurrent call to `Close` with `Read`? It seems like that's its benefit. To me, a simple comment at the top of the struct stating that "lazyPieceReader does not support concurrent calls to `Read` or `Close`. Only make sequential calls to Read, and only call Close after all calls to Read are finished."

That seems like would make it possible to get rid of the mutex, but I'm ok either way.

uplink/ecclient/client.go
```diff
@@ -473,30 +474,81 @@ func (lr *lazyPieceRanger) Size() int64 {
 // Range implements Ranger.Range to be lazily connected
 func (lr *lazyPieceRanger) Range(ctx context.Context, offset, length int64) (_ io.ReadCloser, err error) {
 	defer mon.Task()(&ctx)(&err)
-	ps, err := lr.dialPiecestore(ctx, &pb.Node{
+
+	reader := &lazyPieceReader{}
+	reader.lazyPieceRanger = lr
+	reader.ctx = ctx
+	reader.offset = offset
+	reader.length = length
```
isaachess: I don't know what our style guide is, but I personally prefer:

```
reader := &lazyPieceReader{
    ranger: lr,
    ctx: ctx,
    etc....
}
```

rather than setting each variable individually after the fact.

uplink/ecclient/client.go
cam-a: Not sure why, because it seems like we close everything at the very end of reading from a download, but I was running into a data race on reading and writing to `lazyPieceReader.closed`, so it seemed there was some concurrent reads and closes?

PR #2713, opened 2019-08-05
pkg/process: support json trace output

jtolds:
What: support writing out json

Why: svgs are hard to parse!

Please describe the tests: no tests for this debug feature
 - Test 1:
 - Test 2:
 

pkg/process/exec_conf.go
```diff
@@ -326,7 +326,11 @@ func cleanup(cmd *cobra.Command) {
 			if err != nil {
 				return err
 			}
-			err = present.SpansToSVG(fh, collect.CollectSpans(ctx, work))
+			if strings.HasSuffix(*traceOut, ".json") {
```
egonelbre: maybe switch on the `filepath.Ext`?

jtolds: i believe we have some folks already who are writing svgs out without an svg extension, so i didn't want to break those use cases

PR #2723, opened 2019-08-06
docs/design: use piece hashes for repair

wthorp:
What: 
Design doc for using piece hashes during repair.

Why:
Using hashes should be more efficient than extra Reed Solomon decoding.
For piece signing for background look here: https://github.com/storj/storj/pull/2525

Please describe the tests:
N/A
 
Implementation would reduce satellite load.


docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Implementation
+
+1. Add an optional `PieceHash` field to the `PieceDownloadResponse` protocol buffer for returning hashes.
+
+2. Alter the storage node code to populate `PieceHash` when responding to the owning satellite downloads.
+
+3. Implement changes to `SegmentRepairer.Repair()`, `ecClient.Repair()`, `decodedRanger.Range()` and any other methods to allow hash checking.
+
+4. Implement changes to `ecClient.Get()` and any other methods to allow a minimum number of nodes to optionally be used, rather than the complete set of nodes in an OrderLimit.
```
jtolds: it will probably be simpler to have two independent implementations of this. one cares deeply about long tail canceling and speed. the other does not care about speed at all and wants to download the entire piece before doing anything.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+5. Implement changes to `SegmentRepairer.Repair()` and any other methods to retry with another node when a single download fails, ensuring that previously downloaded pieces can be reused without downloading again.
+
+## Open issues
+
+Instead of implementation #3, can we somehow build a SHARanger type instead, which somehow aligns to Pieces?
+
+Are extending the existing methods ala `decodedRanger` the best option for repair?
```
jtolds: no, lol

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+3. Implement changes to `SegmentRepairer.Repair()`, `ecClient.Repair()`, `decodedRanger.Range()` and any other methods to allow hash checking.
+
+4. Implement changes to `ecClient.Get()` and any other methods to allow a minimum number of nodes to optionally be used, rather than the complete set of nodes in an OrderLimit.
+
+5. Implement changes to `SegmentRepairer.Repair()` and any other methods to retry with another node when a single download fails, ensuring that previously downloaded pieces can be reused without downloading again.
+
+## Open issues
+
+Instead of implementation #3, can we somehow build a SHARanger type instead, which somehow aligns to Pieces?
```
jtolds: we absolutely could build a SHARanger, but i don't think it should be shoehorned into the other logic

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred to each.  
+
+## Implementation
+
+1. Add an optional `PieceHash` field to the `PieceDownloadResponse` protocol buffer for returning hashes.
+
+2. Alter the storage node code to populate `PieceHash` when responding to the owning satellite downloads.
```
mobyvb: Where will the storagenode get the piece hash from? Is it something that is already stored?

docs/design/repair-with-hashes.md
mobyvb: If we are already partway through a download when we realize that one of the pieces is wrong, we will need to pick another node to download a piece from the order limit. But will we have to re download pieces from the other nodes once we have the new piece? Or can it just be added onto the same stream without interrupting the others?

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+3. Implement changes to `SegmentRepairer.Repair()`, `ecClient.Repair()`, `decodedRanger.Range()` and any other methods to allow hash checking.
+
+4. Implement changes to `ecClient.Get()` and any other methods to allow a minimum number of nodes to optionally be used, rather than the complete set of nodes in an OrderLimit.
+
+5. Implement changes to `SegmentRepairer.Repair()` and any other methods to retry with another node when a single download fails, ensuring that previously downloaded pieces can be reused without downloading again.
```
mobyvb: if I recall correctly, `SegmentRepairer.Repair()` just calls `ecClient.Repair()`, so if we can implement this change inside the ecclient, we shouldn't need to change the segment repairer code.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+Currently, each piece is uploaded to a storage node and validated according to its SHA256 hash. The uplink-signed `uplink_piece_hash` is already stored in the `pieceinfo_` table as a serialized protocol buffer and includes the `PieceID`, `Hash`, and `Signature`. Future work includes sending this uplink-signed hash to the satellite along with each piece during repair.
+
+Only them minimum number of pieces should be used for Reed-Solomon decoding. The satellite should check the hash and signature of each piece it downloads. `CertDB` is already be available on the Satellite for obtaining uplink public keys to validate signatures. If a piece fails the hash check, a different piece should be downloaded. Hash check failures may indicate cheating. Reporting and handling cheaters is outside the scope of this document.
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred to each.  
+
+## Implementation
+
```
mobyvb: I think it this is missing a step for adding piece hashes to the satellite. Or is that hash retrieved from the storagenode on download?

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+## Design
+
+Currently, each piece is uploaded to a storage node and validated according to its SHA256 hash. The uplink-signed `uplink_piece_hash` is already stored in the `pieceinfo_` table as a serialized protocol buffer and includes the `PieceID`, `Hash`, and `Signature`. Future work includes sending this uplink-signed hash to the satellite along with each piece during repair.
+
+Only them minimum number of pieces should be used for Reed-Solomon decoding. The satellite should check the hash and signature of each piece it downloads. `CertDB` is already be available on the Satellite for obtaining uplink public keys to validate signatures. If a piece fails the hash check, a different piece should be downloaded. Hash check failures may indicate cheating. Reporting and handling cheaters is outside the scope of this document.
```
mniewrzal: We are not storing uplink cert in `CertDB` anymore because uplink identity is generated on fly and size of `CertDB` was exploding. Uplink is receiving `piece private key` with some calls to metainfo and its used to sign piece hash and orders. Uplink public key (necessary for verification) is stored in `OrderLimit`. To verify piece hash signature Satellite needs to download piece `OrderLimit` with uplink public key, download piece hash and verify OrderLimit with own cert and verify piece hash with public key from OrderLimit.

docs/design/repair-with-hashes.md
mniewrzal: We are not storing piece hashes on satellite to save disk space. Piece hash should be retrieved on-demand from the storage node.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred to each.  
+
+## Implementation
+
+1. Add an optional `PieceHash` field to the `PieceDownloadResponse` protocol buffer for returning hashes.
```
mniewrzal: I'm wondering if we need separate rpc method for repair version of `PieceDownloadResponse`. We will need additionally `PieceHash` and `OrderLimit` and one way or another we need to distinguish if its satellite repair requests (to include it in response) or its standard download (not include). It might be cleaner to have just separate method like `Repair` or something.

docs/design/repair-with-hashes.md
mniewrzal: uplink `Piece Hash` and satellite `Order Limit` are stored along with piece

docs/design/repair-with-hashes.md
kaloyan-raev: The storage node can check the Action field of the OrderLimit. If the Action is GET_REPAIR, it can return the piece hash in the response.

mniewrzal: ah yes, true

docs/design/repair-with-hashes.md
kaloyan-raev: Streaming won't work at all with Repair with Hashes. The satellite should do the following:
1. Download all pieces entirely from stroage nodes to local storage.
2. Calculate the hashes (this can be done while downloading) and compare them to the expected piece hashes.
3. Erasure decode the segment from the downloaded pieces.
4. Erasure encode the new repair pieces from the reconstructed segment.
5. Upload the repair pieces to new storage nodes.

Steps 1 and 2 will repeat until we have enough pieces to execute Step 3.

docs/design/repair-with-hashes.md
kaloyan-raev: The Repair logic should be moved away from both Segment Store and ECClient. We currently have an issue that the uplink drags a lot of the satellite dependencies. This affects the size of libuplink libraries too. One reason for this is that Segment Store and ECClient has the repair logic.

mobyvb: The segment repairer is now in the `repairer` package, not in the segment store.
It does make sense to separate that logic away from the ecclient.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+> To solve this problem, hashes of every piece will be stored alongside each piece on each storage node. A validation hash that the set of hashes is correct will be stored in the pointer. During repair, the hashes of every piece can be retrieved and validated for correctness against the pointer, thus allowing each piece to be validated in its entirety. This allows the repair system to correctly assess whether or not repair has been completed successfully without using extra redundancy for the same task.
+
+Hash verification on the satellite requires understanding the current piece signing and verification workflow:
+
+1. Satellite generates a new key-pair for each piece.
+2. The piece public key is included in the order limit and signed by the satellite
+3. The piece private key is included in the create segment, download segment or delete a segment.
+4. The uplink uses piece private key to sign the orders.
```
mniewrzal: to sign orders and uplink piece hash

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+5. The storage node verifies order limit signature as usual and then uses the piece public key to verify the order and piece hash.
+6. Storage node sends both the order limit and order to the satellite during settlement.
+7. Storage node sends data, order limit and piece hash to the satellite, if it needs to prove data-consistency to the satellite.
```
mniewrzal: this point is not part of current workflow

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred over download even a single more piece.
+
+Downloading for repair is significantly different enough from streaming as to warrant a new, custom implementation.
+
+Using only the minimum number of pieces means that Reed-Solomon does not act as a check during repair. Hence hashing is used instead. While and uplink could potentially send signed bogus data to a storage node, the storage node would not be penalized by these actions. This requires that Audit implements a similar piece hash check instead of relying solely on Reed-Solomon encoding.
+
+The size of all piece hashes downloaded should be roughly equal to a segment size : 64mB.  It seems preferable to keep this in memory over dealing with persistance to disk.
```
mniewrzal: 64MiB is current default max segment size, it doesn't have to be equal or close to 64MiB

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. Add an optional `PieceHash` and `OrderLimit` fields to the `PieceDownloadResponse` protocol buffer for returning hashes.
+
+2. Alter the storage node code to populate `PieceHash` and `OrderLimit` when the `OrderLimit` `Action` is `GET_REPAIR`.
+
+3. Write code which can validate a piece using a `PieceHash` and `OrderLimit` as described in design.
+
+4. Write code which ensures that enough pieces share the same `PiecePublicKey` as described in design.
+
+5. Impelement custom repair download code which uses a Reed-Solomon minimum number of pieces, persisting the downloaded pieces only in memory. If statistics are available, prefer to download from faster storage nodes using the power of two choices.
```
mniewrzal: typo `Implement`

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Implementation
+
+1. Add an optional `PieceHash` and `OrderLimit` fields to the `PieceDownloadResponse` protocol buffer for returning hashes.
+
+2. Alter the storage node code to populate `PieceHash` and `OrderLimit` when the `OrderLimit` `Action` is `GET_REPAIR`.
+
+3. Write code which can validate a piece using a `PieceHash` and `OrderLimit` as described in design.
+
+4. Write code which ensures that enough pieces share the same `PiecePublicKey` as described in design.
```
mniewrzal: Is this from a satellite perspective?

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Background
+
+The white-paper states:
+
+> Data repair is an ongoing, costly operation that will use significant bandwidth, memory, and processing power, often impacting a single operator. As a result, repair resource usage should be aggressively minimized as much as possible.
+> For repairing a segment to be effective at minimizing bandwidth usage, only as few pieces as needed for reconstruction should be downloaded. Unfortunately, Reed-Solomon is insufficient on its own for correcting errors when only a few redundant pieces are provided. Instead, piece hashes provide a better way to be confident that we’re repairing the data correctly.
+> To solve this problem, hashes of every piece will be stored alongside each piece on each storage node. A validation hash that the set of hashes is correct will be stored in the pointer. During repair, the hashes of every piece can be retrieved and validated for correctness against the pointer, thus allowing each piece to be validated in its entirety. This allows the repair system to correctly assess whether or not repair has been completed successfully without using extra redundancy for the same task.
```
egonelbre: Maybe split such that these end up as separate paragraphs.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+4. The uplink uses piece private key to sign the orders.
+5. The storage node verifies order limit signature as usual and then uses the piece public key to verify the order and piece hash.
+6. Storage node sends both the order limit and order to the satellite during settlement.
+7. Storage node sends data, order limit and piece hash to the satellite, if it needs to prove data-consistency to the satellite.
+
+Thus the uplink-signed `uplink_piece_hash` is already stored in the storage node's `pieceinfo_` table as a serialized protocol buffer and includes the `PieceID`, `Hash`, and uplink `Signature`. The `pieceinfo_` table also includes the `order_limit` which contains the `PiecePublicKey` necessary for validating the `PieceHash`, as well as the `SatelliteSignature` to validate the `OrderLimit` itself.
+
+## Design
+
+When a satellite is download pieces for repair, the storage node should also include the `PieceHash` and `OrderLimit` to the satellite along with each piece.
```
egonelbre: typo "When a satellite downloads pieces for repair,"

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+Downloading for repair is significantly different enough from streaming as to warrant a new, custom implementation.
+
+Using only the minimum number of pieces means that Reed-Solomon does not act as a check during repair. Hence hashing is used instead. While and uplink could potentially send signed bogus data to a storage node, the storage node would not be penalized by these actions. This requires that Audit implements a similar piece hash check instead of relying solely on Reed-Solomon encoding.
+
+The size of all piece hashes downloaded should be roughly equal to a segment size : 64mB.  It seems preferable to keep this in memory over dealing with persistance to disk.
+
+## Implementation
+
+1. Add an optional `PieceHash` and `OrderLimit` fields to the `PieceDownloadResponse` protocol buffer for returning hashes.
```
egonelbre: `PieceDownload` is streaming RPC, so it would need to either include it at the start or at the end of the download.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+2. Alter the storage node code to populate `PieceHash` and `OrderLimit` when the `OrderLimit` `Action` is `GET_REPAIR`.
```
egonelbre: AFAIU, we want to get rid of distinguishing between repair_gets and gets... I wonder whether we should always return them on a full piece download.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Design
+
+When a satellite is download pieces for repair, the storage node should also include the `PieceHash` and `OrderLimit` to the satellite along with each piece.
+
+The satellite should attempt to download exactly the Reed-Solomon minimum number of pieces during a repair, storing them in memory. It should validate the `OrderLimit` using its own private key.  It should validate that the `PieceHash` `Signature` corresponds to the `PublicKey` contained in the `OrderLimit`.  The satellite should calculate the size and hash of the piece as it downloads them, and it should be equal to the `PieceHash` `Hash` and `PieceSize`. If a piece fails any of these checks, an error should be logged and a different piece should be downloaded.  Once all pieces are downloaded, the `PiecePublicKey` of all pieces should be compared and equal.  If they are not equal, an error should be logged and more pieces should be downloaded until a minimum number of pieces share the same `PiecePublicKey`.  When enough pieces have been downloaded and validated, Reed-Solomon decoding should be used to reassemble the segment, and repair should continue as previously designed.
```
navillasa: just because there's a lot of vocab and things moving around here, it could be helpful to include what a struct would look like for a `PiecePublicKey` or even `OrderLimit` here even though that isn't new.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+Failed hash, signature, or public keys checks may indicate cheating. Therefore audit must implement this same piece hash check logic. Changes to audit and the reporting of cheaters are otherwise outside the scope of this document.
```
navillasa: maybe "public key" instead of "keys"

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+Failed hash, signature, or public keys checks may indicate cheating. Therefore audit must implement this same piece hash check logic. Changes to audit and the reporting of cheaters are otherwise outside the scope of this document.  
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred over download even a single more piece.
```
navillasa: maybe should be "preferred over downloading even a single piece."

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+Downloading for repair is significantly different enough from streaming as to warrant a new, custom implementation.
+
+Using only the minimum number of pieces means that Reed-Solomon does not act as a check during repair. Hence hashing is used instead. While and uplink could potentially send signed bogus data to a storage node, the storage node would not be penalized by these actions. This requires that Audit implements a similar piece hash check instead of relying solely on Reed-Solomon encoding.
```
navillasa: While *an uplink

Also, does a ticket already exist for implementing piece hash checks in the audit system? We could make one that's like- make a mini design doc for piece hash checking for the audit system

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+3. Write code which can validate a piece using a `PieceHash` and `OrderLimit` as described in design.
+
+4. Write code which ensures that enough pieces share the same `PiecePublicKey` as described in design.
+
+5. Impelement custom repair download code which uses a Reed-Solomon minimum number of pieces, persisting the downloaded pieces only in memory. If statistics are available, prefer to download from faster storage nodes using the power of two choices.
+
+6. Wire in validation logic to repair loop.  If a piece fails these checks, download another.
+
+7. Reassemble the segment using Reed-Solomon decoding and continue with repair as previously implemented.
```
navillasa: now that we're using piece hash checks to determine if pieces are correct, at what point do we know to reassemble the segment? do we update the pointer when we find that piece hashes don't match?

docs/design/repair-with-hashes.md
mobyvb: If the key pair is generated on the satellite, why is it the uplink that does the signing here? Maybe I am just misunderstanding.

docs/design/repair-with-hashes.md
mobyvb: Should we try downloading k+1 to allow us to do rs error detection? Or just stick with exactly the minimum?

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+2. Alter the storage node code to populate `PieceHash` and `OrderLimit` when the `OrderLimit` `Action` is `GET_REPAIR`.
+
+3. Write code which can validate a piece using a `PieceHash` and `OrderLimit` as described in design.
+
+4. Write code which ensures that enough pieces share the same `PiecePublicKey` as described in design.
+
+5. Impelement custom repair download code which uses a Reed-Solomon minimum number of pieces, persisting the downloaded pieces only in memory. If statistics are available, prefer to download from faster storage nodes using the power of two choices.
+
+6. Wire in validation logic to repair loop.  If a piece fails these checks, download another.
```
mobyvb: And also make sure the piece that failed the check is removed from the pointer

docs/design/repair-with-hashes.md
mobyvb: We can reassemble the segment as soon as we have the minimum Reed-Solomon number of pieces that match their piece hashes.  We should update the pointer when we find a piece hash does not match, but that doesn't appear to be mentioned in the implementation yet.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,63 @@
...
+
+7. Reassemble the segment using Reed-Solomon decoding and continue with repair as previously implemented.
+
+## Open issues
+
+With all this hashing and checking of signatures, are we still sure this method is preferred?
```
mobyvb: The amount of data transferred is still reduced with this method for repair, even with the extra hashing/checking logic. And since bandwidth is the most expensive resource, it makes sense to prioritize reducing that over having a "simpler" solution.

docs/design/repair-with-hashes.md
mniewrzal: piece private key is used by uplink to sign piece hash which contains hash of uploaded data and order (not order limit) which informs how much bandwidth was used (put/get). Sending such data to satellite to sign make no sense.

mobyvb: Step #1 says "Satellite generates a new key-pair for each piece". Why does the satellite generate the pair if the uplink is the one using it?

docs/design/repair-with-hashes.md
wthorp: The entire point of this is to avoid downloading the +1.  [That and if there are 2 cheaters, downloading +1 extra might not be enough, if there are 3 cheaters, downloading +2 might not be enough, etc.]

docs/design/repair-with-hashes.md
wthorp: The satellite generates the key pair so that the public key can be signed by the satellite, so that the satellite can trust the signing done with it.

I confess this text is stolen from https://github.com/storj/storj/pull/2525; maybe looking at that PR would help.

docs/design/repair-with-hashes.md
wthorp: I couldn't figure out how without my markdown linter complaining.  It seems my linter was incorrect. :/

egonelbre: I think it should be:

```
> paragraph 1
>
> paragraph 2
>
> paragraph 3
```

docs/design/repair-with-hashes.md
mobyvb: Thanks for the link. That clarified things.

docs/design/repair-with-hashes.md
wthorp: yes!  https://storjlabs.atlassian.net/browse/V3-1322  It turns out that this ticket came out of design / document and this fact was never explained to us.  https://docs.google.com/document/d/1Vcu7lLPSMa9OTemGxdlLbhptGqKu1FyY4ltNHczP2lw/edit  :sigh:  I discovered this too late in the process for it to be of much use.

docs/design/repair-with-hashes.md
wthorp: I think I'll leave them as optional, we can possible include them only if the PeerIdentity is that of the satellite which owns that piece.  The payload size difference is much greater including them than making them optional, so this is risk management I suppose.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,79 @@
...
+> 
+> To solve this problem, hashes of every piece will be stored alongside each piece on each storage node. A validation hash that the set of hashes is correct will be stored in the pointer. During repair, the hashes of every piece can be retrieved and validated for correctness against the pointer, thus allowing each piece to be validated in its entirety. This allows the repair system to correctly assess whether or not repair has been completed successfully without using extra redundancy for the same task.
+
+Hash verification on the satellite requires understanding the current piece signing and verification workflow:
+
+1. Satellite generates a new key-pair for each piece.
+2. The piece public key is included in the order limit and signed by the satellite
+3. The piece private key is included in the create segment, download segment or delete a segment.
```
kaloyan-raev: I think this sentence misses a word or two. It's not clear where exactly the private key is included.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Rationale
+
+Both download and Reed Solomon decoding create more load for a satellite than hashing, so hashing should be preferred over downloading even a single piece.
+
+Downloading for repair is significantly different enough from streaming as to warrant a new, custom implementation.
+
+Using only the minimum number of pieces means that Reed-Solomon does not act as a check during repair. Hence hashing is used instead. While an uplink could potentially send signed bogus data to a storage node, the storage node would not be penalized by these actions. This requires that Audit implements a similar piece hash check instead of relying solely on Reed-Solomon encoding.
+
+The size of all piece hashes downloaded should be roughly equal to a default maximum segment size : 64MiB.  It seems preferable to keep this in memory over dealing with persistance to disk.
```
kaloyan-raev: Shouldn't the "size of all downloaded pieces", not the "piece hashes downloaded" be roughly equal to the segment size?

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,79 @@
...
+1. Add an optional `PieceHash` and `OrderLimit` fields to the `PieceDownloadResponse` protocol buffer for returning hashes.
+
+```message PieceDownloadResponse {
+    // Chunk response for download request
+    message Chunk {
+        int64 offset = 1;
+        bytes data = 2;
+    }
+    Chunk chunk = 1;
+    optional PieceHash hash = 2;
```
kaloyan-raev: The `optional` keyword is a thing of proto2. It was removed in proto3.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,79 @@
...
+    optional OrderLimit limit = 3;
+}```
+
+2. Alter the storage node code to populate `PieceHash` and `OrderLimit` when the `OrderLimit` `Action` is `GET_REPAIR`.
```
kaloyan-raev: It does not make sense to populate the `PieceHash` and `OrderLimit` with every chunk. Perhaps, populating them in the first-chunk message only would be enough.

docs/design/repair-with-hashes.md
```diff
@@ -0,0 +1,79 @@
...
+
+7. If any pieces failed any of the above checks, update the pointer in metainfo, removing those pieces.
+
+7. Reassemble the segment using Reed-Solomon decoding and continue with repair as previously implemented.
+
+## Open issues
+
+We may want to get rid of distinguishing between repair_gets and gets... should we always return them on a full piece download?
+
+64MiB is current default max segment size, it doesn't have to be equal or close to 64MiB.  Should we have an option to persist to disk?
```
kaloyan-raev: There is a check on the satellite on `CommitSegment` that limits the segment size requested by the uplink. Thus, I think it is safe to have an in-memory only implementation.

PR #2724, opened 2019-08-06
Fix retain request iteration in debug mode

mobyvb:
What:
Update `pieceInfo.GetPieceIDs` to use a cursor rather than a numeric offset to allow us to more easily iterate when deleting pieces during retain requests.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/piecestore/endpoint.go
+++ b/storagenode/piecestore/endpoint.go
@@ -614,7 +614,7 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
 	}
 
 	const limit = 1000
-	lastSeen := storj.PieceID{}
+	cursor := storj.PieceID{}
 	numDeleted := 0
 	hasMorePieces := true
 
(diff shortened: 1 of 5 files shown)


storagenode/piecestore/endpoint.go
```diff
@@ -649,7 +649,9 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
 
 		hasMorePieces = (len(pieceIDs) == limit)
 		offset += len(pieceIDs)
-		offset -= numDeleted
+		if endpoint.config.RetainStatus == RetainEnabled {
+			offset -= numDeleted
```
egonelbre: This would still would be problematic when the deletion fails, so we need to track the pieceid where we left off, somehow... rather than using a numberic offset.

storagenode/storagenodedb/pieceinfo.go
```diff
@@ -67,16 +67,16 @@ func (db *pieceinfo) Add(ctx context.Context, info *pieces.Info) (err error) {
 }
 
 // GetPieceIDs gets pieceIDs using the satelliteID
-func (db *pieceinfo) GetPieceIDs(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit, offset int) (pieceIDs []storj.PieceID, err error) {
+func (db *pieceinfo) GetPieceIDs(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, limit int, lastSeen storj.PieceID) (pieceIDs []storj.PieceID, err error) {
```
egonelbre: `lastSeen` is usually called `cursor`.

mobyvb:
@egonelbre added a test and renamed to `cursor`

PR #2725, opened 2019-08-06
libuplink upload/download err handling improvements

bryanchriswhite:
What: Use `errs2.IsCanceled` instead of comparing the error against `context.Canceled`.

Why: Because combined and/or nested errors cause false negatives.

Please describe the tests:
  - `lib/uplinkc/object_test.c`: added sections that exercise upload and download cancellation.
  - `internal/errs2/ignore_test.go`: new test to cover `errs2.IsCanceled`.


lib/uplinkc/object.go
```diff
@@ -148,7 +148,9 @@ func upload_write(uploader C.UploaderRef, bytes *C.uint8_t, length C.size_t, cEr
 
 	n, err := upload.wc.Write(buf)
 	if err != nil {
-		*cErr = C.CString(fmt.Sprintf("%+v", err))
```
mobyvb: We don't want to set the error if it's not cancelled here? There was no check before.

lib/uplinkc/object.go
```diff
@@ -320,7 +324,9 @@ func download_close(downloader C.DownloaderRef, cErr **C.char) {
 
 	err := download.rc.Close()
 	if err != nil {
-		*cErr = C.CString(fmt.Sprintf("%+v", err))
```
mobyvb: Same question here. We weren't checking before, but now we are.

lib/uplinkc/testdata/object_test.c
```diff
@@ -141,6 +196,53 @@ void handle_project(ProjectRef project) {
...
+            // closing canceled download isn't an error
+            download_close(downloader, err);
+            require_noerror(*err);
+
+            // writing canceled download isn't an error
+            int read_size = download_read(downloader, (uint8_t *)downloaded_data, data_len, err);
+            require_noerror(*err);
+            require(read_size == 0);
+
+            // TODO: test with combined/nested context cancellation error
```
mobyvb: Isn't that more of a test for `errs2.IsCanceled`? I don't think we need to test it here.

bryanchriswhite: good point

lib/uplinkc/object.go
bryanchriswhite: We don't set the error if it **is** canceled. These changes make `upload_cancel`, `upload_commit`, `upload_write`, `download_close`, `download_cancel`, and `download_read` all behave the same way in that they return early if they encounter errors but they only set the error string if that error is not a context cancellation.

I'm not certain that this is the best approach but the inconsistency was causing issues. If we keep it this way I think it might make sense to be able to know from C whether your uploader/downloader is closed/comitted/canceled. I think you and I have actually talked about that before.

lib/uplinkc/object.go
bryanchriswhite: (discussion above)

PR #2726, opened 2019-08-06
Remove the use of in memory SQLite3 tables for storage nodes.

simongui:
**What:** Removed the use of SQLite3 in-memory databases when tests run using `testplanet`. This PR enables `testplanet` to run a normal `SQLite3` database file in a temporary location during tests so that we more closely match production.

**Why:** As part of [PR 2680](https://github.com/storj/storj/pull/2680) it became increasingly difficult to work around the way we run `testplanet` differently than production and migration started to get ugly. I've heard this difference from others has caused other issues in the past. This PR makes our test and production code more consistent.

**Please describe the tests:**
 - Test 1: I ran all tests several times to ensure no failures were introduced.
 - Test 2:
 
**Please describe the performance impact:**
On my local MacBook Pro.
`make test` on `master` using SQLite3 in-memory took `161s`.
`make test` on this PR using a SQLite3 database file on disk took `262s`.

I tested with `PRAGMA schema.synchronous = 0` set and it did not affect much.


internal/testplanet/storagenode.go
```diff
@@ -91,14 +71,14 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatellites storj.Nod
 				BootstrapBackoffBase: 500 * time.Millisecond,
 				BootstrapBackoffMax:  2 * time.Second,
 				Alpha:                5,
-				DBPath:               storageDir, // TODO: replace with master db
+				DBPath:               filepath.Join(storageDir, "kademlia/"),
```
egonelbre: Would this break existing setups?

internal/testplanet/storagenode.go
```diff
@@ -91,14 +71,14 @@ func (planet *Planet) newStorageNodes(count int, whitelistedSatellites storj.Nod
...
 				Operator: kademlia.OperatorConfig{
 					Email:  prefix + "@mail.test",
 					Wallet: "0x" + strings.Repeat("00", 20),
 				},
 			},
 			Storage: piecestore.OldConfig{
-				Path:                   "", // TODO: this argument won't be needed with master storagenodedb
+				Path:                   filepath.Join(storageDir, "pieces/"),
```
egonelbre: This argument isn't used AFAIK.

PR #2730, opened 2019-08-07
lib/uplinkc: make .so smaller

aleitner:
What: Make the libuplink shared object smaller. Omit the DWARF symbol table and omit the symbol table and debug information when building the .so

Why: Our shared object is many megabytes too big.


JessicaGreben:
how much does this change the size?

aleitner:
> how much does this change the size?

22mb -> 16mb

JessicaGreben:
nice!

mniewrzal:
Linux (fedora): 24.8mb -> 19.8mb

PR #2731, opened 2019-08-07
satellite/metainfo: update partner id from request

aligeti:
What: 
Adds  the partnerid to the bucketsDB, when a new bucket is created or updats an already existing bucket if partner id exists in the APIkey info, used to access(create) a new bucket.
Why:
OSPP integration testing

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/metainfo/metainfo.go
```diff
@@ -645,15 +645,10 @@ func (endpoint *Endpoint) CreateBucket(ctx context.Context, req *pb.BucketCreate
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
-	// checks if bucket exists before updates it or makes a new entry
+	// checks if bucket exists before updates it with partner id if it exits or makes a new entry
+	partnerID := keyInfo.PartnerID
 	bucket, err := endpoint.metainfo.GetBucket(ctx, req.GetName(), keyInfo.ProjectID)
 	if err == nil {
-		var partnerID uuid.UUID
```
VinozzZ: Do we still need this logic for the app-connector?

satellite/metainfo/metainfo.go
```diff
@@ -645,15 +645,10 @@ func (endpoint *Endpoint) CreateBucket(ctx context.Context, req *pb.BucketCreate
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
-	// checks if bucket exists before updates it or makes a new entry
+	// checks if bucket exists before updates it with partner id if it exits or makes a new entry
+	partnerID := keyInfo.PartnerID
 	bucket, err := endpoint.metainfo.GetBucket(ctx, req.GetName(), keyInfo.ProjectID)
 	if err == nil {
```
coyle: This is actually what @egonelbre 's comments came from a few weeks ago. We should be pulling the partnerID from a user-agent string not the request object.

satellite/metainfo/metainfo.go
aligeti: updated the logic

satellite/metainfo/metainfo.go
aligeti: got it..

aligeti:
> LGTM, other than the comment from Dennis. Also is there tests for this?

there are test cases written in place... I will double check it again.

satellite/metainfo/metainfo.go
mniewrzal: [How to get UserAgent on server side](https://github.com/grpc/grpc-go/issues/1100)

aligeti:
> LGTM, other than the comment from Dennis. Also is there tests for this?

Line #649 keyInfo structure has partnerID that is extracted thru ctx.Value(ctx). so the partner ID is extracted thru user-agent.

satellite/metainfo/metainfo.go
coyle: Let's do this in this PR

satellite/metainfo/validation.go
```diff
@@ -371,3 +381,29 @@ func (endpoint *Endpoint) validatePieceHash(ctx context.Context, piece *pb.Remot
...
+func (endpoint *Endpoint) getUserAgent(ctx context.Context) (_ *uuid.UUID, err error) {
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		if userAgent, ok := md["user-agent"]; ok {
+			if len(userAgent) == 1 {
+				ua := useragent.Parse(userAgent[0])
+				if ua == nil {
+					endpoint.log.Sugar().Info("invalid user-agent found %s", userAgent[0])
+					return nil, status.Errorf(codes.InvalidArgument, err.Error())
+				}
+				partnerID, err := rewards.GetPartnerID(ua.Name)
```
VinozzZ: Should we also check if the `ua.PartnerID` is equal to the `partnerID` that we get from the satellite?

VinozzZ: After the meeting today, we will not validate the partner ID here and only validate it during billing

PR #2733, opened 2019-08-07
docs/design: Uptime disqualification

ifraixedes:
What: First draft of the uptime disqualification

Why: Because we had to disable the uptime disqualification after the initial approach disqualified several nodes unfairly and/or unclear.

### DIFF
--- a/docs/design/uptime-reputation-calculation.md
+++ b/docs/design/uptime-reputation-calculation.md
@@ -118,7 +118,6 @@ Once an _SN_ is selected for an uptime check, the current implemented uptime che
 1. Check the _SN_ as it's currently done; if it succeeds, then end, otherwise follow with the next step.
 1. Insert a new row in the `failed_uptime_checks` using its node ID and setting the current timestamp to `when` and `last_check`.
 
-
 #### Uptime Recheck Loop
 
 This process should run independently from any current satellite process and it should run in a configurable time interval.
@@ -133,14 +132,26 @@ The algorithm for each time interval iteration is the following:
 1. Calculate its total offline time of the uptime check period, accumulating the offline time interval of each row corresponding to this _SN_ with `back_online` not being `NULL` and the current offline interval time calculated with the current timestamp and the `when` value of this row. If the total offline time doesn't exceed the established network limits go to 1, otherwise continue.
 1. Update the row setting `disqualified` and `last_check` to the current timestamp and increment `count`; then disqualifies the _SN_<sup>2</sup>.
 
-
-#### Configurable parameters
+### Configurable parameters
 
 The following parameters will be configurable:
 
 1. Uptime check period: It's the number of days where uptime check failures are accumulated for calculating the total offline time of _SNs_. Default 30 days.
 1. Maximum allowed offline time. It's the percentage of the uptime check period that _SNs_ can be offline for not being disqualified. Default 0.05%.
 1. Uptime recheck interval. It's the time interval that _SNs_ which failed the initial uptime check are rechecked until they are back online, disqualified or we stop to check them because they are not in the network anymore. Default 1 hour.
+
+### Metrics
+
+This section mentions the specific metrics that the [Uptime Recheck Loop](#uptime-recheck-loop) (algorithm part 2) must take; the section doesn't mention any metric regarding to [Uptime Check _SN_ selection](uptime-check-sn-selection) (algorithm part 1) because the current implementation is used so it already has metrics and it's out of the scope of this design document.
+
+The following metrics must be tracked:
+
+* Number of failed uptime checks per _SN_.
+* Number of succeeded uptime checks per _SN_.
+* Subtotals that each_SN_ has been offline per uptime check period.
+* The total number of _SNs_ are disqualified due to uptime check failures per uptime check period.
+
+This metrics will be tracked through [monkit](https://github.com/spacemonkeygo/monkit) which is the system currently used for such purpose.
 
 ## Open issues
 
--- a/docs/design/uptime-reputation-calculation.md
+++ b/docs/design/uptime-reputation-calculation.md
@@ -15,22 +15,22 @@ The problem with the current uptime _SN_ reputation score calculation is that it
 While we understand that an _SN_ that has remained in the network with a low uptime check failures could be considered a more reliable _SN_ than new ones which recently joined, we see, at the same time, that new _SN_ shouldn't be harsher penalized than veteran nodes due to uptime check failures because:
 
 1. Any _SN_ (veteran and new ones) requires some time to be offline for maintenance and both should have the same maximum amount of time for it.
-1. The end goal of the uptime checks is to ensure that the network has certain availability and retrievability and uptime check failures affects to them no matter on how long the _SN_ has remained in the network and how much data an _SN_ has, hence node with more or less data must be treated equally.
+1. The end goal of the uptime checks is to ensure that the network has certain availability and retrievability and uptime check failures affect to them no matter on how long the _SN_ has remained in the network and how much data an _SN_ has, hence node with more or less data must be treated equally.
 
 This design doc is focused on how to design and implement a fair, simple and understandable uptime reputation calculation after our initial approach didn't behave as expected, causing the disqualification of several nodes without having a proof that they have failed enough uptime checks that the _SNs_ were offline long time enough which put in danger the minimum level of the network availability and retrievability.
 
 ## Design
 
-The new process for doing the uptime checks is based on once an _SN_ is selected for having the check, if the _SN_ fails the initial check, then it's tracked to periodically receive uptime checks until (a) it successfully passes an uptime check or (b) gets disqualified due to many uptime check failures in one month.
+The new process for doing the uptime checks is based on once an _SN_ is selected for having the check, if the _SN_ fails the initial check, then it's tracked to periodically receive uptime checks until (a) it successfully passes an uptime check or (b) gets disqualified due to be offline the maximum allowed by the established number of days (henceforh uptime check period) or it isn't needed to be checked anymore because it has been excluded from the network by any reason.
 
-Once an _SN_ has failed the first uptime check, the _SN_ is considered offline and the subtotal time, which the _SN_ is considered offline, is the time elapsed between the first and the last failed uptime check. The total offline time of an _SN_ is the accumulation of the subtotals that the _SN_ has had during the ongoing month and when such total surpasses the maximum permitted percentage time, the _SN_ will get disqualified.
+Once an _SN_ has failed the first uptime check, the _SN_ is considered offline and the subtotal time, which the _SN_ is considered offline, is the time elapsed between the first and the last failed uptime check. The total offline time of an _SN_ is the accumulation of the subtotals that the _SN_ has had during the ongoing uptime check period and when such total surpasses the maximum permitted percentage offline time, the _SN_ will get disqualified.
 
 The maximum percentage of time that an _SN_ can be offline will be established with the network commitment in terms of availability and retrievability during one month.
 
 The process is divided in 2 parts:
 
 1. Selecting _SNs_ for receiving uptime checks.
-1. Periodically check the _SNs_ which has failed the first uptime check until they pass one or get disqualified.
+1. Periodically check the _SNs_ which have failed the first uptime check until they pass one or get disqualified or get excluded from the network.
 
 For the first part, we foresee to use the current implemented mechanism with some minor modifications for excluding the _SNs_ which are being checked by the second part.
 
@@ -50,7 +50,7 @@ And the following disadvantages
 1. It requires the satellite to store, for some period of time, historical data for nodes which have failed at least one uptime check.
 1. It requires more resources than the current implemented approach.
 1. It doesn't calculate the exact time that an _SN_ is offline; once an _SN_ has a failed uptime check the time elapsed between the last failed uptime check and the next succeeded uptime check causes not to have the accurate offline time.
-1. The _"score"_ obtained from uptime check failures, without any further modification, doesn't integrate with the current _SN reputation system_.
+1. There isn't a _"score"_ obtained from uptime check failures which can be integrate with the current _SN reputation system_.
 
 While one of our goals is to reduce the number of resources used by the satellite we see that the advantages beat the disadvantages considering that the current system got disabled after some nodes got disqualified unfairly and/or unclear.
 
@@ -73,7 +73,7 @@ CREATE TABLE failed_uptime_checks (
 )
 ```
 
-The table will be filled with _SNs_ which have failed at least one uptime check during the ongoing month.
+The table will be filled with _SNs_ which have failed at least one uptime check during the ongoing uptime check period.
 
 Each row in the table indicates when an _SN_ failed the first uptime check (`when`), while the _SN_ not succeed any future uptime check (`back_online` is `NULL`); when the _SN_ is back online any future uptime check failure will insert a new row to mark a new period of time that the node is offline again.
 
@@ -112,11 +112,19 @@ The algorithm for each time interval iteration is the following:
         1. Update the `back_online` column of the row with the current timestamp.
 
 
+#### Configurable parameters
+
+The following parameters will be configurable:
+
+1. Uptime check period: It's the number of days where uptime check failures are accumulated for calculating the total offline time of _SNs_. Default 30 days.
+1. Maximum allowed offline time. It's the percentage of the uptime check period that _SNs_ can be offline for not being disqualified. Default 0.05%.
+1. Uptime recheck interval. It's the time interval that _SNs_ which failed the initial uptime check are rechecked until they are back online, disqualified or we stop to check them because they are not in the network anymore. Default 1 hour.
+
 ## Open issues
 
 Currently we have the following issues and concerns which can affects somehow the implementation of this system:
 
-1. Currently the uptime checks are combined with the audit check for calculating the total _SN_ reputation; after this implementation, the uptime check _"score"_ cannot be combined anymore as before.
+1. Currently the uptime checks are combined with the audit check for calculating the total _SN_ reputation; after this implementation, the uptime check doesn't present a _"score"_ which can directly combined with.
 1. The implementation of this system requires to interact with other satellite processes, <sup>1</sup> getting the _SN_ address and <sup>2</sup> disqualify a _SN_; for keeping this system uncoupled of those processes and being able to scale up the satellite, it should be a way to perform those operations through an interface which will remain once the satellite will be broken into different distributed system.
 1. The implementation of the _uptime recheck loop_ mentions to retrieve a list of _SNs_ in order of being able to maintain the order, however, it may be a problem with it in order of holding the result set open too long, impacting the perfomance of database.
-1. The implementation of _uptime recheck loop_ section should contain a protocol buffers definition in case that the process will run on a different machine; this part hasn't been elaborated because we may decide on not doing it for the first version but we should consider it during the implementation for being able to adapt it with the minimal modifications.
+1. The implementation of _uptime recheck loop_ section should contain a protocol buffers definition in case the process run on a different machine; this part hasn't been elaborated because we may decide on not doing it for the first version but we should consider it during the implementation for being able to adapt it with the minimal modifications.
--- a/docs/design/uptime-reputation-calculation.md
+++ b/docs/design/uptime-reputation-calculation.md
@@ -1,0 +1,11 @@
+# Uptime reputation caculation
+
+## Abstract
+
+Haveing a fair, simple and understandable way to calculate the uptime of _SNs_ is crucial for not unfairly disqualifying them, makes clear how _SNOs_ should administrate and maintain their _SN_ for being align with the network needs and have a clear proof of disqualification for _SN_ getting disqualified due to too many uptime checks failures.
+
+## Background
+
+Currently our disqualification system is described in the [disqualification design doc](disqualification.md) which is based in the cacluation of different _SN_ reputation scores as described in [node-selection design doc](node-selection.md).
+
+This design doc is focused in how to design and implement a fair, simple and undestandble uptime reputation score calcuation after our initial approach din't behave as expected causing the undisqualication of several nodes without having a clear way to proof it and not being able o show that such disqualification was fair.
--- a/docs/design/uptime-reputation-calculation.md
+++ b/docs/design/uptime-reputation-calculation.md
@@ -2,50 +2,121 @@
 
 ## Abstract
 
-Having a fair, simple and understandable way to calculate the uptime of _storagenodes (henceforth SNs)_ is crucial for not causing unfair disqualifications, clarifying how _storagenode operators (henceforth SNOs)_ should administrate and maintain their _SN_ for being align with the network requirements and last but not least have a clear proof of disqualification for _SN_ getting disqualified due to too many uptime checks failures.
+Having a fair, simple and understandable way to calculate the uptime of _storagenodes (henceforth SNs)_ is crucial for not causing unfair disqualifications, clarifying how _storagenode operators (henceforth SNOs)_ should administrate and maintain their _SN_ for being align with the network requirements and last but not least have a clear proof of disqualification for _SN_ getting disqualified due to too many uptime check failures.
 
-This design doc propose a different way for calculating uptime check scores.
+This design doc proposes a different way for calculating uptime check scores.
 
 ## Background
 
 Currently our disqualification system is described in the [disqualification design doc](disqualification.md) which is based in the calculation of different _SN_ reputation scores as described in [node-selection design doc](node-selection.md).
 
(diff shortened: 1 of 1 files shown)


docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The algorithm for each time interval iteration is the following:
+
+1. Select all the rows from `failed_uptime_checks` table which has `back_online` and `disqualification` columns set to `NULL` sorted by `when` in ascending order.
+1. For each selected _SN_, retrieve the _SN_ address<sup>1</sup> and performs the uptime check.
+    * If it fails
+        1. Calculate its total offline time of the month, accumulating the offline time interval of each row corresponding to this _SN_ with `back_online` not being `NULL` and the current offline interval time calculated with the current timestamp and the `when` value of this row. If the total offline time doesn't exceed the established network limits end, otherwise follow with the next step.
+        2. Update the `disqualified` column of the row with the current timestamp and disqualifies the _SN_<sup>2</sup>
+    * If it succeeds
+        1. Update the `back_online` column of the row with the current timestamp.
```
littleskunk: I see 2 edge case. Lets say we allow an offline time of 24 hours before we disqualify a storage node.

1. A storage node can be offline for 5 minutes every day (dynamic IP change). -> It fails the inital check and will be marked as offline. Now the loop will select the storage node after 1 hour and notice that it is not offline. back_online will be set. At the end of the month the storage node gets disqualified. In our database we have 30 entries with 1 hour each. The storage node was offline only 5 minutes each but we didn't catch that. The problem is that we will disqualify storage nodes because the loop takes to long for one interation. We could ignore the first or last uptime check. If the storage node is back online than set back_online not with the current timestamp. Instead use the last when timestamp. In this example that would mean the storage node gets 0 penalty and we would be force to reduce the loop interval in order to catch him.


2. The satellite might be offline for several hours. We need to make sure satellite offline time doesn't count. How about a job that is running once on restart and sets back_online for all nodes. Same trick. If we set it to the last when timestamp we can make sure the storage nodes are getting 0 penalties for our mistakes.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Open issues
+
+Currently we have the following issues and concerns which can affects somehow the implementation of this system:
+
+1. Currently the uptime checks are combined with the audit check for calculating the total _SN_ reputation; after this implementation, the uptime check _"score"_ cannot be combined anymore as before.
```
littleskunk: I would say combined. Uptime checks and audit checks are working independent. Storage nodes can get disqualified for beeing offline too long or for failing too many audits. They can't get disqualified for a combination of both. One of them has to hit the magic line.

docs/design/uptime-reputation-calculation.md
zeebo: Another option is to fill any `back_online` columns necessary when the node connects with the satellite as well. Then, the node coming online could reach out during startup.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+A database is required to store information about _SNs_ uptime check failures. This section presents the new required database schema and a brief explanation what it will store.
+
+Only one database table is required and its schema is the following (pseudo code):
+
+```sql
+CREATE TABLE failed_uptime_checks (
+    node_id       BYTEA NOT NULL,
+    when          TIMESTAMP WITH TIMEZONE NOT NULL, -- When the SN failed the an uptime check until it succeeds a future uptime check
+    back_online   TIMESTAMP WITH TIMEZONE, -- When SN has succeeded an uptime check after the last one failed
+    disqualified  TIMESTAMP WITH TIMEZONE  -- When the SN has been disqualified by any disqualification system (uptimes, audits, etc.)
```
zeebo: Should this be denormalized with some other table? Is this supposed to be the canonical location for if a node is disqualified?

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+
+This design doc is focused on how to design and implement a fair, simple and understandable uptime reputation calculation after our initial approach didn't behave as expected, causing the disqualification of several nodes without having a proof that they have failed enough uptime checks that the _SNs_ were offline long time enough which put in danger the minimum level of the network availability and retrievability.
+
+## Design
+
+The new process for doing the uptime checks is based on once an _SN_ is selected for having the check, if the _SN_ fails the initial check, then it's tracked to periodically receive uptime checks until (a) it successfully passes an uptime check or (b) gets disqualified due to many uptime check failures in one month.
+
+Once an _SN_ has failed the first uptime check, the _SN_ is considered offline and the subtotal time, which the _SN_ is considered offline, is the time elapsed between the first and the last failed uptime check. The total offline time of an _SN_ is the accumulation of the subtotals that the _SN_ has had during the ongoing month and when such total surpasses the maximum permitted percentage time, the _SN_ will get disqualified.
```
zeebo: Is "ongoing month" calculated based on the proceeding 30 days (whatever they were) or is it from the beginning of the calendar month?

If it's the latter, a node can decide, near the end of the month, to get some free "offline time". In other words, if there's only 8 hours left in the calendar month, and you need 10 hours to be disqualified, you can just go offline.

docs/design/uptime-reputation-calculation.md
ifraixedes: @littleskunk I thought about the (1) and let's discuss in the meeting how we can detect those cheaters because the document was trying to define that the if the interval between the last failed uptime check and the next one was ignored for not unfairly penalizing.

Let's discuss the scenario (2) too.

docs/design/uptime-reputation-calculation.md
ifraixedes: For me that's OK but we'll have to consult data science team about it because the initial implementation and the one that we have now (although it's disabled) is mentioning to combine them.

docs/design/uptime-reputation-calculation.md
ifraixedes: No, it isn't the canonical location.

It exists here for being able to run this process separately and not using the DB for sharing state between different services. Perhaps we don't nee this level of independence in such we can query the value from the table which holds it.

zeebo: I expect there will have to be a canonical location for features like manually disqualifying or ... qualifying a node. Indeed, perhaps there are other reasons or factors that go into the decision of disqualification.

Maybe rather than call it "disqualified", call it "stopped_checking"?

docs/design/uptime-reputation-calculation.md
ifraixedes: > Is "ongoing month" calculated based on the proceeding 30 days (whatever they were) or is it from the beginning of the calendar month?

I had the same doubt, but after some thoughts, I thought that a calendar month is clearer from the _SN_ point of view and when I found metrics on some services they speak in months, which I interpreted as calendar months.

> If it's the latter, a node can decide, near the end of the month, to get some free "offline time". In other words, if there's only 8 hours left in the calendar month, and you need 10 hours to be disqualified, you can just go offline.

~Sure, but doesn't apply the same for any SLA commitment?~

As far as I understand the problem here is that a bunch of SNs which have 8 hours left at the end of the month may decide to go offline all at once and they won't get disqualified but the network may be broken.

We may randomize the period of time for each one as a solution.

docs/design/uptime-reputation-calculation.md
ifraixedes: > Maybe rather than call it "disqualified", call it "stopped_checking"?

I like that.

docs/design/uptime-reputation-calculation.md
zeebo: I think trailing 30 days is less gameable, and I personally think users would equally understand metrics like "You had 1.6 hours of downtime in the last 30 days" and "You had 1.6 hours of downtime this month", but I guess that's a product question.

docs/design/uptime-reputation-calculation.md
wthorp: we combine them when creating an overall reputation score by which we prioritize node selection using "the power of two choice"

docs/design/uptime-reputation-calculation.md
wthorp: I believe the intention was a 30 day window.

ifraixedes: As commented in the meeting I will update the document to calculated in days.
It will also be mentioned that his value is configurable.

docs/design/uptime-reputation-calculation.md
ifraixedes: As commented in the meeting, I will update the document to reflect that  the uptime check that detects the _SN_ is back online, it will request if the _SN_ has contacted the satellite before this last check and after the previous one and if it did, then use such time stamp to fill the `back_online` otherwise use the current timestamp.

I will add a new column (`last_check`) to register the timestamp of the previous check because it will solve the scenario (2) and will allow to querying the database one row a time rather than having to have an open result set for too long.

Also, I will add a column to count the number of checks despite that it could be roughly calculated knowing the used time interval it's less straightforward and we'll need to track if the time interval has been changed during the period of time involved for the calculation.

--- 

__I've got a question when writing this__.

Once we use the last _SN_ contact timestamp with the satellite or the current timestamp, do we need to discard the elapsed time between the succeeded uptime check and the previous failed one?

If we assume that when an _SN_ goes offline, it has always to contact the satellite at the moment that it gets online again then it looks that we can precisely measure the offline time; if the _SN_ didn't do so we could consider that it's cheating.

docs/design/uptime-reputation-calculation.md
ifraixedes: As commented A part of renaming the column, I will mention in the algorithm to check not for _SN_ disqualification but for anything else that we should stop to check the _SN_ because it isn't considered part of the network anymore.

docs/design/uptime-reputation-calculation.md
ifraixedes: That's a good point; I didn't remember so.

Then it looks that we have to look for a way to get a value that we can combine them.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The approach described in this document present the following advantages:
+
+1. It's clear why an _SN_ get disqualified due to failed uptime checks.
+1. It has a clear proof for an _SNOs_ why their _SNs_ get disqualified due to failed uptime check in case that they request it before the designed time to clean up the historical failed uptime checks data.
+1. It allows to report to _SN_ at any how far is to be disqualified by failing uptime checks.
+
+And the following disadvantages
+
+1. It requires to store, for some period of time, historical for nodes which have failed at least one uptime check.
```
benjaminsirb: In DS, we are hoping to have SN uptime as a time series (i.e. timestamped 1 for success and 0 for failure) to have richer data with which we may be able to do forecasting. Thus, I'm not sure how big (1) is of a disadvantage, given that DS would find this information useful and we are hoping to get it either way.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+The second part requires a totally new implementation which should be totally independent of any current satellite service for being able in the near future to run it separately for scalability purpose.
+
+
+## Rationale
+
+The approach described in this document present the following advantages:
+
+1. It's clear why an _SN_ get disqualified due to failed uptime checks.
+1. It has a clear proof for an _SNOs_ why their _SNs_ get disqualified due to failed uptime check in case that they request it before the designed time to clean up the historical failed uptime checks data.
+1. It allows to report to _SN_ at any how far is to be disqualified by failing uptime checks.
```
wthorp: Maybe `It allows reporting to a _SN_, at any point, how far it is from disqualified by failing uptime checks.`

docs/design/uptime-reputation-calculation.md
wthorp: Maybe `It requires the satellite to store, for some period of time, historical data for nodes which have failed at least one uptime check`

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+
+And the following disadvantages
+
+1. It requires to store, for some period of time, historical for nodes which have failed at least one uptime check.
+1. It requires more resources than the current implemented approach.
+1. It doesn't calculate the exact time that an _SN_ is offline; once an _SN_ has a failed uptime check the time elapsed between the last failed uptime check and the next succeeded uptime check causes not to have the accurate offline time.
```
wthorp: If we coordinate with the Kademlia Removal (https://github.com/storj/storj/pull/2686) we can have a plan for the storage node to notify the node when it comes back online.  This should reduce this concern.

ifraixedes: Yes, I thought about it and I asked for confirmation in https://github.com/storj/storj/pull/2733#discussion_r312189487

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+    back_online   TIMESTAMP WITH TIMEZONE, -- When SN has succeeded an uptime check after the last one failed
+    disqualified  TIMESTAMP WITH TIMEZONE  -- When the SN has been disqualified by any disqualification system (uptimes, audits, etc.)
+)
+```
+
+The table will be filled with _SNs_ which have failed at least one uptime check during the ongoing month.
+
+Each row in the table indicates when an _SN_ failed the first uptime check (`when`), while the _SN_ not succeed any future uptime check (`back_online` is `NULL`); when the _SN_ is back online any future uptime check failure will insert a new row to mark a new period of time that the node is offline again.
+
+When the _SN_ is disqualified by any disqualification system and the _SN_ isn't back online, the `disqualified` column of the row will be updated for not receiving more uptime checks because we don't do business with disqualified _SNs_.
```
wthorp: Could we gain some efficiency by removing them from this table completely instead?  Or do we think that won't be worth the effort?

ifraixedes: :thinking: we actually could.

I will present that in today's meeting.

docs/design/uptime-reputation-calculation.md
ifraixedes: I'm holding this until we assess in today's meeting https://github.com/storj/storj/pull/2733#discussion_r312436739

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+1. Select which _SNs_ will be checked for uptime, henceforth uptime check _SN_ selection.
+1. Recheck the uptime of the _SNs_ which has failed the initial uptime check (part 1) during an interval period of time, henceforth uptime recheck loop.
+
+#### Uptime Check _SN_ selection
+
+This design doc doesn't change the _SN_ selection process for uptime checks therefore the current implemented mechanism with some minor modifications to classify those nodes which fail the uptime check in order of being recheck by the second part of the algorithm.
+
+Once an _SN_ is selected for an uptime check, the current implemented uptime check mechanism must be modified to achieve the following algorithm:
+
+1. If there is a row in the `failed_uptime_checks` table for the selected _SN_ with `back_online` set to `NULL` do nothing, otherwise follow with the next step.
```
wthorp: Isaac is currently adding a cache of this information, I believe?  Can the document reflect the fact that this information should be cached?  My ideal would be that the majority of uptime / address reporting doesn't even do a SELECT from the database.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+This process should run independently from any current satellite process and it should run in a configurable time interval.
+
+The algorithm for each time interval iteration is the following:
+
+1. Select all the rows from `failed_uptime_checks` table which has `back_online` and `disqualification` columns set to `NULL` sorted by `when` in ascending order.
+1. For each selected _SN_, retrieve the _SN_ address<sup>1</sup> and performs the uptime check.
+    * If it fails
+        1. Calculate its total offline time of the month, accumulating the offline time interval of each row corresponding to this _SN_ with `back_online` not being `NULL` and the current offline interval time calculated with the current timestamp and the `when` value of this row. If the total offline time doesn't exceed the established network limits end, otherwise follow with the next step.
+        2. Update the `disqualified` column of the row with the current timestamp and disqualifies the _SN_<sup>2</sup>
+    * If it succeeds
```
wthorp: If it succeeds, it could still be `disqualified`

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,122 @@
...
+Once an _SN_ is selected for an uptime check, the current implemented uptime check mechanism must be modified to achieve the following algorithm:
+
+1. If there is a row in the `failed_uptime_checks` table for the selected _SN_ with `back_online` set to `NULL` do nothing, otherwise follow with the next step.
+1. Check the _SN_ as it's currently done; if it succeeds, then end, otherwise follow with the next step.
+1. Insert a new row in the `failed_uptime_checks` using its node ID and setting the current timestamp to `when`.
+
+
+#### Uptime Recheck Loop
+
+This process should run independently from any current satellite process and it should run in a configurable time interval.
```
wthorp: If a SN alerts the satellite when its back online, do we need a service loop for rechecks?  Could we just check for disqualification when a node comes back online? [maybe also audit or something?]

docs/design/uptime-reputation-calculation.md
ifraixedes: The columns are added.
I also found that we needed a new table to know the current uptime check period in case the satellite restarts.

docs/design/uptime-reputation-calculation.md
ifraixedes: This has been rephrased in terms of that this operations aren't directly with the DB are through a service interface and such database should only be accessible by the service.

Isaac is adding a cache system for the existing system, not for this one.

docs/design/uptime-reputation-calculation.md
ifraixedes: This part has been changed considerably, please check it again and make the comment on it because right now I cannot understand the comment probably because it has changed.

docs/design/uptime-reputation-calculation.md
ifraixedes: This has been discussed with @benjaminsirb and the document has been updated.

docs/design/uptime-reputation-calculation.md
ifraixedes: I will think about this next week and we'll discuss it in a meeting that I will convoke.

docs/design/uptime-reputation-calculation.md
ifraixedes: This also will be discussed in the next meeting that I'll convoke.

ifraixedes:
> When do we clear old entries out of the failed_uptime_checks table? Would we'd have to leave the last status for each one in there?

I added a section for the historical data clean up.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,177 @@
...
+
+Having a fair, simple and understandable way to calculate the uptime of _storagenodes (henceforth SNs)_ is crucial for not causing unfair disqualifications, clarifying how _storagenode operators (henceforth SNOs)_ should administrate and maintain their _SN_ for being align with the network requirements and last but not least have a clear proof of disqualification for _SN_ getting disqualified due to too many uptime check failures.
+
+This design doc proposes a different way for calculating uptime check scores.
+
+## Background
+
+Currently our disqualification system is described in the [disqualification design doc](disqualification.md) which is based in the calculation of different _SN_ reputation scores as described in [node-selection design doc](node-selection.md).
+
+The problem with the current uptime _SN_ reputation score calculation is that it isn't based on the actual time that a node is being offline, by cause of to the nature of the chosen model, causing that an _SN_ will get disqualified due its number of uptime checks success and failure which implies that new nodes can get disqualified having less uptime check failures than veteran nodes.
```
egonelbre: `vetted nodes`

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,177 @@
...
+
+While we understand that an _SN_ that has remained in the network with a low uptime check failures could be considered a more reliable _SN_ than new ones which recently joined, we see, at the same time, that new _SN_ shouldn't be harsher penalized than veteran nodes due to uptime check failures because:
+
+1. Any _SN_ (veteran and new ones) requires some time to be offline for maintenance and both should have the same maximum amount of time for it.
```
egonelbre: `vetted`

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,177 @@
+# Uptime reputation caculation
+
+## Abstract
+
+Having a fair, simple and understandable way to calculate the uptime of _storagenodes (henceforth SNs)_ is crucial for not causing unfair disqualifications, clarifying how _storagenode operators (henceforth SNOs)_ should administrate and maintain their _SN_ for being align with the network requirements and last but not least have a clear proof of disqualification for _SN_ getting disqualified due to too many uptime check failures.
```
egonelbre: Let's not use abbreviations that aren't in the common vocabulary of software development. These design documents are intended to be read by other people, not just the development team.

ifraixedes: All the abbreviations are mentioned what they are before using them despite that they belong to the Storj business domain.
I've seen other documents using the abbreviations without explaining what they are.

docs/design/uptime-reputation-calculation.md
jenlij: so it's based on percent failures/total right?

egonelbre: No, I think this was my misunderstanding. The veteran here mean't nodes that have been in the network for a long time.

egonelbre: But, yes, currently it's based roughly on failures/total, but with forgetting.

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,177 @@
...
+
+While we understand that an _SN_ that has remained in the network with a low uptime check failures could be considered a more reliable _SN_ than new ones which recently joined, we see, at the same time, that new _SN_ shouldn't be harsher penalized than veteran nodes due to uptime check failures because:
+
+1. Any _SN_ (veteran and new ones) requires some time to be offline for maintenance and both should have the same maximum amount of time for it.
+1. The end goal of the uptime checks is to ensure that the network has certain availability and retrievability and uptime check failures affect to them no matter on how long the _SN_ has remained in the network and how much data an _SN_ has, hence node with more or less data must be treated equally.
+
+This design doc is focused on how to design and implement a fair, simple and understandable uptime reputation calculation after our initial approach didn't behave as expected, causing the disqualification of several nodes without having a proof that they have failed enough uptime checks that the _SNs_ were offline long time enough which put in danger the minimum level of the network availability and retrievability.
```
jenlij: This is a nit pick, but to make this sentence a little easier to read, you could break it up like this...
"This design doc is focused on how to design and implement a fair, simple and understandable uptime reputation calculation after our initial approach didn't behave as expected. Our original design caused the disqualification of several nodes without having a proof that they have failed enough uptime checks. This flaw put the minimum level of the network availability and retrievability at risk."

docs/design/uptime-reputation-calculation.md
```diff
@@ -0,0 +1,177 @@
...
+
+The new process for doing the uptime checks is based on once an _SN_ is selected for having the check, if the _SN_ fails the initial check, then it's tracked to periodically receive uptime checks until (a) it successfully passes an uptime check or (b) gets disqualified due to be offline the maximum allowed by the established number of days (henceforh uptime check period) or it isn't needed to be checked anymore because it has been excluded from the network by any reason.
+
+Once an _SN_ has failed the first uptime check, the _SN_ is considered offline and the subtotal time, which the _SN_ is considered offline, is the time elapsed between the first and the last failed uptime check. The total offline time of an _SN_ is the accumulation of the subtotals that the _SN_ has had during the ongoing uptime check period and when such total surpasses the maximum permitted percentage offline time, the _SN_ will get disqualified.
+
+The maximum percentage of time that an _SN_ can be offline will be established with the network commitment in terms of availability and retrievability during one month.
+
+The process is divided in 2 parts:
+
+1. Selecting _SNs_ for receiving uptime checks.
```
jenlij: Maybe we can add the uptime auditor here: satellite tries to reach nodes we haven't heard from in an hour and considers uptime audit failures for reputation. If this is out of scope, maybe we should create another design doc

ifraixedes: I would address the audit uptime failures in another doc, considering that this one will be renamed to be only uptime tracking/calculation.
Nonetheless, I will take a look at the audit code.

ifraixedes:
As discussed in last week architecture meeting this design document:

1. Had to be aligned with the changes outlined in the [Kademlia removal design document](https://github.com/storj/storj/blob/master/docs/design/kademlia-removal.md).
2. Other discussions in several meetings ended up in the agreement that first of all we have to measure and register the Storage Node offline time. In the near future, we'll decide how offline time affects Storage Node reputation and disqualification, hence the design document limits the scope to only measuring and tracking offline time.

Those decisions changed the direction of the document, so I have mostly written a new one.

I close this PR in favor of https://github.com/storj/storj/pull/2857

PR #2734, opened 2019-08-07
docs/design: Storage Node Graceful Exit

ethanadams:
What:  When a Storage Node wants to leave the network but does not want to lose their escrow we need to have a mechanism for them to exit the network “gracefully”.

Why:  Give Storage Nodes a mechanism to leave the network while receiving their escrows, and reduce repair caused by node churn on satellites. 

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+	}
+	```
+- Update DBX - Add table exit_pieceinfos
+-  ```
+	model exit_pieceinfo (
+		key node_id path
+
+		field node_id           blob
+		field path              blob
+		field peice_info        blob
```
aligeti: nit pick .. piece.. also instead of dt mention 'date'

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+  - Add GetExitingNodeIds method to overlaycache. Returns nodes IDs where `exit_initiated_dt` is not null and `exit_completed_dt` is null.
+- Create GracefulExit endpoint
+  - Endpoints should be secured using the peer Identity provided in context
+  - Initiates the exit by setting `nodes.exit_initiated` to current time
+  - ``` 
+	service GracefulExit {
+		rpc Initiate(stream InitiateRequest) returns (stream InitiateResponse) {}
+	}
+
+	message InitiateRequest {
```
aligeti: This is not a comment, but a thought... when I first read 'initiaterequest' didn't think it was referring to graceful exit , so what do you think, of naming it 'GracefulExitInitiateRequest'

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+		field node_id           blob
+		field path              blob
+		field peice_info        blob
+		field durability_ratio  float64
+		field queued_dt         timestamp ( autoinsert )
+		field sent_dt           timestamp ( updateable )
+		field completed_dt      timestamp ( updateable )
+	)
+   ```
+- Add `PieceAction` field to, `cache.FindStorageNodesRequest`. Update `cache.FindStorageNodesWithPreferences` to ignore exiting nodes for uploads and repairs.
```
aligeti: instead of piece action, can u query to find nodes that have only 'exit_initiated_dt' null?, that way no need to add a new piece action.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+	```
+- Update bandwidth usage monitors to ignore `PieceAction_PUT_EXIT` bandwidth actions
+- Add GracefulExit service
+  - Iterates over `exit_order` where `completed_dt` is null
+    - Pushes the pieces to the storage node identified in the order using `ecclient`
+    - Sends the signed new node response to the satellite via a new `CommitPiece` method (uses `metainfo.UpdatePieces`).
+    - Updates `bytes_deleted` with the number of bytes deleted and sets `completed_dt` to current time
+  - Execution intervals and batch sizes should be configurable
 
 ## Open Questions
```
aligeti: "What happens if a piece is deleted when the piece is in the process of being transferred from the exiting node to other node in the network?"

I believe if I understood this correctly, the exiting storage node shall only honor download request and not delete requests or new upload request, Infact exiting SN should internal be made to accept only the download and reject accepting new uploads and/or delete requests

docs/design/storagenode-graceful-exit.md
```diff
@@ -36,31 +36,32 @@ This process including the Storage nodes transferring their pieces to other node
...
-- When a Storage Node is gracefully exiting the network I want all of the pieces they are storing to be deleted from their hard drive as they exit so that their computers hard drive space is no longer used.
-- When a Storage Node is exiting the network gracefully I want the satellite to have the ability to track how much egress they used for exiting so that we do not pay them for that bandwidth.
+- When a Storage Node is in the process of gracefully exiting the network, it shall continue to participate in the requested audits and uptimes checks.
+- When a Storage Node is in the process of gracefully exiting the network, it shall continue to honor to download requests.
+	- The Storage Node keep a separate and detailed metric of network bandwidth used to serve the data to clients vs bandwidth used for graceful exit. The Storage Node shall NOT get paid for bandwidth used for graceful exit but shall get paid for the bandwidth used to serve data to clients(downloads, audits, uptime checks etc...)
```
mniewrzal: I don't think we are paying for `audits, uptime checks etc...`

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+
+#### Satellite
+- Update DBX - Add updateable `exit_initiated_dt` and `exit_completed_dt` timestamps to nodes table with indexes
+  - Add GetExitingNodeIds method to overlaycache. Returns nodes IDs where `exit_initiated_dt` is not null and `exit_completed_dt` is null.
+- Create GracefulExit endpoint
+  - Endpoints should be secured using the peer Identity provided in context
+  - Initiates the exit by setting `nodes.exit_initiated` to current time
+  - ``` 
+	service GracefulExit {
+		rpc Initiate(stream InitiateRequest) returns (stream InitiateResponse) {}
```
mniewrzal: Why `stream`? Do we plan to keep it open during the whole exit process?

ethanadams: Oversight.  Fixed

docs/design/storagenode-graceful-exit.md
ethanadams: Updated to GracefulExit rpc InitiateExit(InitiateExitRequest) returns (InitiateExitResponse)

docs/design/storagenode-graceful-exit.md
kaloyan-raev: For sure, we are not paying for uptime checks. We should double-check if we are paying for audits.

docs/design/storagenode-graceful-exit.md
```diff
@@ -36,31 +36,32 @@ This process including the Storage nodes transferring their pieces to other node
...
+- When a Storage Node is exiting the network gracefully, it shall keep a detailed record of the network bandwidth used for the purpose of Graceful Exit and shall provide the metrics information to satellite upon request. 
+- The Satellite shall have a mechanism in place to make sure that the Storage Node reporting is correct. This shall eliminate Storage Node from reporting incorrect metrics interms of bandwidth usage. (TBD, is this needed on the satellite side??)
 
 ### When Exited 
-- When a Storage Node leaves the network I want the ability to run a report on the satellite to get information about what Storage Nodes exited so that I can pay them their escrow amounts. 
-	- Create a report on the satellite to display which nodes have exited gracefully during a specified time frame. The report must include:
+- When a Storage Node left the network, the satellite shall have the capabilty to run a report to get information exited and/or exiting Storage Nodes, so that it shall pay them their escrow amounts accordingly. The report shall support the number of Storage Nodes left in a configurable specified time frame and shall include other information as shown below (TBD):
```
kaloyan-raev: The bullet list below does not render properly in markdown mode after this change.

kaloyan-raev: "...to get information exited...". It looks like the word "about" is missing between "information" and "exited".

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
 	- Receiving storage nodes should be uploaded to as if they were receiving any other upload (orders, etc).
 - We need to have the storage node keep track how its progress during graceful exit by calculating how much data it had when it started the graceful exit and keeping track of how much data it has deleted (data is deleted from the exiting node as it sends pieces to new nodes). This is an estimation since the storage node is likely holding more data than it will gracefully exit (garbage data). 
 
+## Implementation (MVP)
+
+#### Satellite
+- Update DBX - Add updateable `exit_initiated_dt` and `exit_completed_dt` timestamps to nodes table with indexes
```
kaloyan-raev: Is the `_dt` suffix intended to be `_at`? Similar to the `created_at` and `updated_at` fields in the `nodes` table?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+  - Add GetExitingNodeIds method to overlaycache. Returns nodes IDs where `exit_initiated_dt` is not null and `exit_completed_dt` is null.
```
kaloyan-raev: Don't we want to distinguish between those nodes which initiated the graceful exit process and those which already completed it?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,133 @@ This process including the Storage nodes transferring their pieces to other node
...
+- Create GracefulExit endpoint
```
kaloyan-raev: The endpoint is on the satellite, right?

docs/design/storagenode-graceful-exit.md
ethanadams: changed _dt to _at to be consistent

docs/design/storagenode-graceful-exit.md
ethanadams: updated to downloads and audit egress

docs/design/storagenode-graceful-exit.md
ethanadams: updated

docs/design/storagenode-graceful-exit.md
ethanadams: updated

docs/design/storagenode-graceful-exit.md
ethanadams: added another lookup method for exited nodes

docs/design/storagenode-graceful-exit.md
ethanadams: correct

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
 
+## Implementation (MVP)
+
+#### Satellite
+- Update DBX - Add updateable `exit_initiated_at` and `exit_completed_at` timestamps to nodes table with indexes
+  - Add GetExitingNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is null.
+  - Add GetExitedNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is not null. 
+- Create GracefulExit endpoint
+  - Endpoints should be secured using the peer Identity provided in context
+  - Initiates the exit by setting `nodes.exit_initiated` to current time
```
kaloyan-raev: `nodes.exit_initiated` --> `nodes.exit_initiated_at`

docs/design/storagenode-graceful-exit.md
kaloyan-raev: The new PieceAction field indeed seems unnecessary. The caller can take advantage of the existing `ExcludedNodes` field to list the current nodes from the pointer.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
+		field piece_info        blob
+		field durability_ratio  float64
+		field queued_at         timestamp ( autoinsert )
+		field sent_at           timestamp ( updateable )
+		field completed_at      timestamp ( updateable )
+	)
+   ```
+- Add `PieceAction` field to, `cache.FindStorageNodesRequest`. Update `cache.FindStorageNodesWithPreferences` to ignore exiting nodes for uploads and repairs.
+- Update Repairer service
+  - Modify `checker` to check segments for pieces associated with a storage node that is exiting. Add to `exit_pieceinfo` table if matches criteria.
```
kaloyan-raev: This should be implemented as a new observer to the metainfo loop.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
+	```
+- Update DBX - Add table exit_pieceinfos
+-  ```
+	model exit_pieceinfo (
+		key node_id path
+
+		field node_id           blob
+		field path              blob
+		field piece_info        blob
+		field durability_ratio  float64
```
kaloyan-raev: What's `durability_ratio` and why do we need it in this table?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
+- Update Repairer service
+  - Modify `checker` to check segments for pieces associated with a storage node that is exiting. Add to `exit_pieceinfo` table if matches criteria.
+- Add `PieceAction_PUT_EXIT` to orders protobuf. This is used to differentiate exiting bandwidth from other bandwidth usage.
+- Create GracefulExit service
+  - SendOrders loop
+    - Iterates over `exit_pieceinfo`, creates signed order limits with action type `PieceAction_PUT_EXIT`
+    - Batches orders and sends them to the exiting storagenode `GracefulExit.ProcessOrders` endpoint
+    - Execution intervals and batch sizes should be configurable
+  - CheckStatus loop
+    - Queries `exit_order` grouping by node ID where all records are marked completed.  The MAX(completed_at) should be used to determine if the last order was completed within a reasonable (???) threshold to ensure the repairer/checker was able to make enough passes to capture all pieces for the exiting node.
```
kaloyan-raev: > ... to ensure the repairer/checker was able to make enough passes to capture all pieces for the exiting node.

One complete cycle of the metainfo loop should be enough to capture all pieces of exiting nodes, assuming that:
- the satellites stops issuing PUT orders to the node after receiving the `InitiateExitRequest`
- the observer of the metainfo loop waits for the beginning of the next cycle of the metainfo loop to start capturing the pieces for the exiting node.

docs/design/storagenode-graceful-exit.md
ethanadams: updated

docs/design/storagenode-graceful-exit.md
ethanadams: updated

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
+    - Execution intervals should be configurable
+- Create gracefulexitreport command in satellite cli
+  - Accepts 2 parameter: start date and end date
+  - Generates a list of all completed exits with NodeID, Wallet address, Date joined, Date exited, GB Transferred (calculated using `PieceAction_PUT_EXIT` bandwidth actions), ordered by date exited descending.
+
+#### Storagenode
+- Add `gracefulexit` command to storagenode CLI
```
kaloyan-raev: How about just `exit` as the name of the command. We don't offer commands for other kinds of exit.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,6 +92,134 @@ This process including the Storage nodes transferring their pieces to other node
...
+  - When executed, the user should be prompted with a numbered list of satellites (what would we display as a name?)
+  - After selecting a satellite, the user should be prompted for a confirmation
+  - Once confirmed, the command should call the `GracefulExit.Initiate` endpoint for that satellite
+  - Records starting disk usage in `gexit.exit_status`
+- Add `gexit` DB implementation with `exit_order`  and `exit_order` tables
```
kaloyan-raev: The second `exit_order` should be `exit_status`.

docs/design/storagenode-graceful-exit.md
ethanadams: updated doc. it's to support this requirement:

"pieces should be prioritized based on how many other pieces are still on the network. Pieces for files that are closer to needing to be repaired should be transferred off the exiting before pieces for files that are still considered healthy."

docs/design/storagenode-graceful-exit.md
ethanadams: Agreed.  Not sure where I was going with this one.  Document updated.

docs/design/storagenode-graceful-exit.md
AlexeyALeonov: > For sure, we are not paying for uptime checks. We should double-check if we are paying for audits.

We are paying for audits traffic.

simongui:
> model nodes (
>  	...
>  	field exit_loop_count           int
>  	field exit_initiated_at         timestamp ( updateable )
>  	field exit_completed_at         timestamp ( updateable )
>
>  }

Since `exit_loop_count` is stored in the database and I think this is incremented and stored every loop cycle per node? How often would this happen? Do we need to persist this so it survives restarts or can it just be in-memory?

docs/design/storagenode-graceful-exit.md
```diff
@@ -36,31 +36,32 @@ This process including the Storage nodes transferring their pieces to other node
...
-- When a Storage Node is exiting the network gracefully I want the satellite to have the ability to track how much egress they used for exiting so that we do not pay them for that bandwidth.
+- When a Storage Node is in the process of gracefully exiting the network, it shall continue to participate in the requested audits and uptimes checks.
+- When a Storage Node is in the process of gracefully exiting the network, it shall continue to honor to download requests.
+	- The Storage Node keep a separate and detailed metric of network bandwidth used to serve the data to clients vs bandwidth used for graceful exit. The Storage Node shall NOT get paid for bandwidth used for graceful exit but shall get paid for the bandwidth used to serve data to clients(downloads, audit egress)
+- When a Storage Node is gracefully exiting the network, it shall delete the piece from its storage after it is successfully transferred to other peer Storage Nodes as directed by the Satellite. The Satellite shall support the mechanism to verify that the transferred piece is correct and complete.
```
egonelbre: It should only delete when it has gotten a confirmation from the satellite for changing segment.

docs/design/storagenode-graceful-exit.md
```diff
@@ -36,31 +36,32 @@ This process including the Storage nodes transferring their pieces to other node
...
+- When a Storage Node is exiting the network gracefully, it shall keep a detailed record of the network bandwidth used for the purpose of Graceful Exit and shall provide the metrics information to satellite upon request. 
+- The Satellite shall have a mechanism in place to make sure that the Storage Node reporting is correct. This shall eliminate Storage Node from reporting incorrect metrics interms of bandwidth usage. (TBD, is this needed on the satellite side??)
```
egonelbre: Yes, this should go through similar approach of verifying piece hashes as uplink.

egonelbre:
I cannot comment on some lines:

> The exiting node will send the new node the hash it has for the piece. (that hash contains the uplinks signature). The new node will sign it and send it back to the exiting node.

The storage node signs the hash it calculated, not the one it was sent.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+## Implementation (MVP)
+
+#### Satellite
+- Update DBX - Add fields to `nodes` table
+  ```
+    model nodes (
+		...
+		field exit_loop_count           int
+		field exit_initiated_at         timestamp ( updateable )
+		field exit_completed_at         timestamp ( updateable )
```
egonelbre:
```
exit_started_at
exit_finished_at
```

The `loop_count` seems weird.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+
+	}
+  ```
+- Add GetExitingNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is null.
+- Add GetExitedNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is not null.
```
egonelbre: `GetExitedNodeIDs`, this would require offsets and limits. It will grow over time.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+  ```
+    model nodes (
+		...
+		field exit_loop_count           int
+		field exit_initiated_at         timestamp ( updateable )
+		field exit_completed_at         timestamp ( updateable )
+
+	}
+  ```
+- Add GetExitingNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is null.
```
egonelbre: `GetExitingNodeIDs`

egonelbre: Maybe one method instead of two? `GetNodesWithStatus(`

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+- Add GetExitedNodeIds method to overlaycache. Returns node IDs where `exit_initiated_at` is not null and `exit_completed_at` is not null. 
+- Update DBX - Add table exit_pieceinfos
+   ```
+	model exit_pieceinfo (
```
egonelbre: `graceful_exit_transfer_queue`

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+		key node_id path
+
+		field node_id             blob
+		field path                blob
+		field piece_num           blob
```
egonelbre: Why is this a blob?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+		field durability_ratio    float64
+		field queued_at           timestamp ( autoinsert )
+		field requested_at        timestamp ( updateable )
```
egonelbre: What is `requested_at`?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+		field failed_at           timestamp ( updateable )
+		field failed_status_code  int
+		field completed_at        timestamp ( updateable )
```
egonelbre: Why do we keep completed pieces in the table? It's going to waste a lot of memory.

docs/design/storagenode-graceful-exit.md
egonelbre: All references to overlaycache should be fixed to `overlay.DB` or `overlay.Service`

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+	)
+   ```
+- Update `cache.FindStorageNodesWithPreferences` to ignore exiting nodes for uploads and repairs.
+- Add `PieceAction_PUT_EXIT` to orders protobuf. This is used to differentiate exiting bandwidth from other bandwidth usage.
+- Create GracefulExit service
+  - Add MetainfoObserver loop
```
egonelbre: Previous design described repair doing the checking, however here it's different.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+    - The service queries the `nodes` to get a list of exiting node IDs where `exit_loop_count` <= `requiredMetainfoLoops`, then
+      - Joins the metainfo loop `observers` using the exiting node IDs and looks for segments that contain remote pieces that the exiting node stores. It adds `node_id`, `path`, `piece_num`, `durability_ratio`, and `queued_at` in `exit_pieceinfo` if a record does not exist.
```
egonelbre: Why wouldn't a record exist if it's running the second time?

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+	  - Join returns after one full metainfo loop, or on err. On success, the service should increment `nodes.exit_loop_count` for the nodeIDs that were processed in the full loop.
+    - Execution intervals and requiredMetainfoLoops should be configurable
+  - Add CheckStatus loop
+    - Queries `exit_pieceinfo` grouping by node ID and counts completed, and incomplete.  If all records are complete and `nodes.exit_loop_count` <= `requiredMetainfoLoops`
+      - Set `nodes.exit_completed` to current time.
+      - Remove all `exit_pieceinfos` for the completed node ID
```
egonelbre: This would mean that it could finish before transfers are completed.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+    - Execution intervals should be configurable
+- Create GracefulExit endpoint
+  - Endpoints should be secured using the peer Identity provided in context
+  - InitiateExit initiates the exit process by setting `nodes.exit_initiated_at` to current time
+  - GetPutOrders
+    - Queries `nodes` by `id` and `exit_completed`.
```
egonelbre: This should already happen in the begging on what the status is.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+      - If completed it returns `PutOrdersResponse` with `exit_completed` set to true and `exit_completed_signature` set to a satellite signed value of satellite id, storage, node id, "gracefuleexit-successfully-complete", completed timestamp.
```
egonelbre: It should just return `nil, ok`

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+    - Execution intervals and requiredMetainfoLoops should be configurable
+  - Add CheckStatus loop
+    - Queries `exit_pieceinfo` grouping by node ID and counts completed, and incomplete.  If all records are complete and `nodes.exit_loop_count` <= `requiredMetainfoLoops`
+      - Set `nodes.exit_completed` to current time.
+      - Remove all `exit_pieceinfos` for the completed node ID
+    - Execution intervals should be configurable
+- Create GracefulExit endpoint
+  - Endpoints should be secured using the peer Identity provided in context
+  - InitiateExit initiates the exit process by setting `nodes.exit_initiated_at` to current time
+  - GetPutOrders
```
egonelbre: Is this a streaming endpoint? If yes, then the design should reflect this.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+      - Else it checks whether there are any records in `exit_pieceinfo` where `exit_pieceinfo.requested_at` is not null and (`exit_pieceinfo.failed_at is not null and exit_pieceinfo.failed_status_code <> NOT_FOUND`) or `exit_pieceinfo.completed_at` is null
+        - If records are found, it should create new order limits for the found records, and return for reprocessing
+        - Else it returns a list of order limits for pieces to be moved to new nodes from  `exit_pieceinfo` ordered by `durability_ratio`. durability_ratio == num pieces / optimal number of pieces. lower values take presedence.
+  - ProcessPutOrders is called by the storagenode to send processed order limits. ProcessPutOrders verifies the hashes of successful orders limits, updates `exit_pieceinfo.completed_at`, and updates metainfo with the new piece location. Failures update `exit_pieceinfo.failed_at`  `exit_pieceinfo.failed_status_code`, for reprocessing. 
+    - If the failed status code is NOT_FOUND, the exiting storage node no longer stores that piece. In this case the node/piece should be removed from the pointer so that a repair is triggered.
+  - ``` 
+	service GracefulExit {
```
egonelbre: This should be streaming.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+		// Called by the storagenode to initiate an exit request
+		rpc InitiateExit(InitiateExitRequest) returns (InitiateExitResponse) {}
```
egonelbre: Having initiate exit and get put orders separate would be more complicated to write. It'll probably easier to have single streaming rpc.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+
+	message InitiateExitRequest {
+	}
+
+	message InitiateExitResponse {
+	}
+
+
+	message PutOrder {
+		bytes hashing_key
```
egonelbre: Why does it need a hashing key. And it's missing private_key.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+	}
+
+	message InitiateExitRequest {
+	}
+
+	message InitiateExitResponse {
+	}
+
+
+	message PutOrder {
```
egonelbre: Use similar naming as metainfo protocol.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+      - On success...
+        - Removes the successful order limits from `exit_orders`
+        - Updates `bytes_deleted` with the number of bytes deleted
+        - Deletes the pieces that were successfully moved
+      - On failure (ex. satellite is unavailable), successful orders limits stored in `exit_orders` should be reprocessed on the next iteration
+  - Execution intervals and batch sizes should be configurable
 
 ## Open Questions
 
 - What happens if we are doing a repair and a graceful exit on the same segment?
```
egonelbre: There are still a lot of critical open questions.

docs/design/storagenode-graceful-exit.md
```diff
@@ -91,10 +92,169 @@ This process including the Storage nodes transferring their pieces to other node
...
+
+		field satellite_id              blob not null
+		field initiated_at              timestamp ( autoinsert ) not null
+		field completed_at              timestamp ( updateable )
+		field starting_disk_usage       int64 not null
+		field bytes_deleted             int64
+		field completed_exit_signature  blob
+	)
+
+	model exit_orders {
```
egonelbre: Why do we need this table?

ethanadams:
Reworked the document into multiple documents. Closing in favor of a new pull request

PR #2742, opened 2019-08-08
storagenode/console: structure service dependencies

rikysya:
What: Structure console service dependencies

Why: Because constructor became too massive

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/console/service.go
+++ b/storagenode/console/service.go
@@ -29,6 +29,26 @@ var (
 	mon = monkit.Package()
 )
 
+// Storage encapsulates DBs used by console service
+type Storage struct {
+	Console      DB
+	Bandwidth    bandwidth.DB
+	Reputation   reputation.DB
+	StorageUsage storageusage.DB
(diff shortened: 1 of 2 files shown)


rikysya:
I am don't quite like `console.Storage` as name for struct encapsulating all `DBs` and sources... May be anyone has some better suggestions?

egonelbre:
@rikysya I think the issue with a lot of dependencies could be that a single service tries to do everything. I think there should be a better solution for them.

rikysya:
@egonelbre the only way I can think of to split them up is to differentiate between `dashboard` data, which is not related to concrete `satellite`, and `satellite` specific data into 2 separate services `console/dashboard.Service` and `console/satellite.Service`. Then using them as separate services in `server`, or combine under `console.Service` which is what `server` will use. But I am not sure that it is significantly better approach...

rikysya:
outdated

PR #2744, opened 2019-08-09
satellite/audit: implement reservoir struct and RemoteSegment observer method

navillasa:
What:
- Implement reservoir struct and RemoteSegment observer method.
- Needs a test and it's not used by anything right now.

Why:
-  Just thought early feedback and test suggestions could be helpful.
- "Create an audit observer that connects to the metainfo loop" - https://storjlabs.atlassian.net/browse/V3-2361
- "Create a reservoir sampling struct" - https://storjlabs.atlassian.net/browse/V3-2360

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,49 @@
...
+
+type reservoirConfig struct {
+	numVettedSlots   int
+	numUnvettedSlots int
+}
+
+// NewReservoir instantiates a Reservoir
+func NewReservoir(isVetted bool, config reservoirConfig) *Reservoir {
+	var numSlots int
+	if isVetted {
```
mobyvb: imo, we should take the number of slots in as an argument to `NewReservoir`, and the config for vetted vs. unvetted should be in the observer logic. That way this can be a generic reservoir sampler that doesn't have to be specific to audits

satellite/audit/service.go
```diff
@@ -72,6 +78,31 @@ func (service *Service) Run(ctx context.Context) (err error) {
 	})
 }
 
+// RemoteSegment takes a remote segment found in metainfo and creates a reservoir for it if it doesn't exist already
+func (service *Service) RemoteSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
```
mobyvb: Do you think it would be better to make the audit service implement the `Observer` interface, or to create a separate struct for this? Like `reservoirObserver` or something.

satellite/audit/service.go
```diff
@@ -72,6 +78,31 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+	defer mon.Task()(&ctx, path)(&err)
+
+	remote := pointer.GetRemote()
+	pieces := remote.GetRemotePieces()
+
+	for i, piece := range pieces {
+		if _, ok := service.Reservoirs[piece.NodeId]; !ok {
+			reputable, err := service.overlay.IsVetted(ctx, piece.NodeId)
```
mobyvb: Since we are already adding logic here to determine whether the node is reputable, I think it makes sense to also pick the vetted vs. unvetted reservoir count here, and pass it into `NewReservoir`. Mentioned in my other comment.

satellite/audit/service.go
```diff
@@ -72,6 +78,31 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+			if err != nil {
+				service.log.Error("error finding if node is vetted", zap.Error(err))
+				return nil
+			}
+			reservoir := NewReservoir(reputable, service.reservoirConfig)
+			reservoir.Segments[0] = remote
+			service.Reservoirs[piece.NodeId] = reservoir
+		} else {
+			service.Reservoirs[piece.NodeId].sample(remote, i)
```
mobyvb: don't do this inside an `else`. We want to sample the piece whether we are creating a reservoir for the first time or using an already created one.

mobyvb: This `i` is also going to be the index of the piece, not the total number of pieces seen by the reservoir so far. `sample` should take in the remote segment, but the reservoir itself should determine the count of the items seen so far. I will add a comment in the relevant section of the reservoir code.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+// sample makes sure that for every segment in metainfo from index i=numSlots..n-1,
+// pick a random number r = rand(0..i), and if r < numSlots, replace reservoir.Segments[r] with segment
+func (reservoir *Reservoir) sample(segment *pb.RemoteSegment, i int) {
+	if len(reservoir.Segments) < reservoir.NumSlots {
+		reservoir.Segments = append(reservoir.Segments, segment)
+	} else {
+		random := rand.Intn(i)
```
mobyvb: Rather than using an `i` that is passed in, add a variable on the struct called `NumSeen` or something, and increment it every time `sample()` is called. Start it at 0 and increment it before the random number generation.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,49 @@
...
+package audit
+
+import (
+	"math/rand"
+	"storj.io/storj/pkg/pb"
+)
+
+// Reservoir holds a certain number of segments to reflect a random sample
+type Reservoir struct {
+	Segments []*pb.RemoteSegment
```
kaloyan-raev: We should be keeping segment paths (type `storj.Path`) instead of `pb.RemoteSegment`. When the audit service selects a segment path from the reservoir, it must look up its up-to-date pointer from the metainfo db.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,49 @@
...
+	IsVetted bool
+	NumSlots int
+}
+
+type reservoirConfig struct {
+	numVettedSlots   int
+	numUnvettedSlots int
```
kaloyan-raev: `slotsForVetted` and `slotsForUnvetted` sound clearer to me.

kaloyan-raev: Otherwise, it is easy to read as "number of (un)vetted slots", which would be wrong.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"math/rand"
+	"storj.io/storj/pkg/pb"
+)
+
+// Reservoir holds a certain number of segments to reflect a random sample
+type Reservoir struct {
+	Segments []*pb.RemoteSegment
+	IsVetted bool
+	NumSlots int
```
kaloyan-raev: How about just `Slots`?

When referenced it would look like `Reservoir.Slots` which reads well.

kaloyan-raev: Or maybe we should call it `Size`.

satellite/audit/reservoir.go
kaloyan-raev: If we want to be truly generic, then instead of `Segments` field we should have a more generic field `Slots` of type `interface{}`.

satellite/audit/service.go
navillasa: yeah that would probably be cleaner

satellite/audit/service.go
navillasa: oh oops, thanks

satellite/audit/reservoir.go
navillasa: okay I think I also prefer Size because Slots to me sounds like it would be a map or array

satellite/audit/reservoir.go
navillasa: it's actually a little convoluted to make the struct accept the slice of segments as a field (https://golang.org/doc/faq#convert_slice_of_interface) and later append to it, so I think I'm just going to leave it as []*pb.RemoteSegment for now.

satellite/audit/service.go
navillasa: wait actually- re: your first comment, I think it still needs to happen in the `else`, because the slots need to fill up to the reservoir size in the order that they appear in the stream and THEN be randomly switched out.

satellite/audit/reservoir.go
navillasa: I can change it, but also do I need to keep this? it didn't seen like we wanted to distinguish between slots for vetted and unvetted nodes after all in today's architecture meeting

satellite/audit/service.go
mobyvb: Well `sample()` already deals with that logic, doesn't it? First it checks to see if more slots need to be filled, and if not, it does the random switching.

navillasa: okay I see- yeah will switch things around a bit

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,40 @@
...
+	"math/rand"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Reservoir holds a certain number of segments to reflect a random sample
+type Reservoir struct {
+	Paths   []storj.Path
+	Size    int
+	NumSeen int
```
kaloyan-raev: I don't immediately associate `NumSeen` with the index of the reservoir. How about using `index` instead of `NumSeen`. Also, I think it does not need to be exported as it a rather internal thing for the reservoir, i.e. `index` instead of `Index`.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,40 @@
...
+
+// NewReservoir instantiates a Reservoir
+func NewReservoir(size int) *Reservoir {
+	return &Reservoir{
+		Size:    size,
+		NumSeen: 0,
+	}
+}
+
+// sample makes sure that for every segment in metainfo from index i=numSlots..n-1,
```
kaloyan-raev: Use `size` instead of `numSlots` in this comment, so it matches the reservoir's field.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,40 @@
...
+// pick a random number r = rand(0..i), and if r < numSlots, replace reservoir.Segments[r] with segment
+func (reservoir *Reservoir) sample(path storj.Path) {
```
kaloyan-raev: Don't we need this method exported for callers?

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,40 @@
...
+	reservoir.NumSeen++
+	if len(reservoir.Paths) < reservoir.Size {
+		reservoir.Paths = append(reservoir.Paths, path)
+	} else {
+		random := rand.Intn(reservoir.NumSeen)
+		if random < reservoir.Size {
+			reservoir.Paths[random] = path
+		}
+	}
+	return
```
kaloyan-raev: This return statement is unnecessary.

satellite/audit/reservoir_test.go
```diff
@@ -0,0 +1,29 @@
...
+	path3 := storj.Path("test/path3")
+
+	reservoir.sample(path1)
+	require.Equal(t, 1, len(reservoir.Paths))
+
+	reservoir.sample(path2)
+	reservoir.sample(path3)
+	require.Equal(t, 1, len(reservoir.Paths))
+
+	// todo: ... not really sure how/what to test at this level
```
kaloyan-raev: We should also have a test for a reservoir with size > 1. At the end, we can check that there are no duplicated items in the reservoir.

satellite/audit/service.go
```diff
@@ -41,20 +50,32 @@ type Service struct {
...
 		Verifier: NewVerifier(log.Named("audit:verifier"), metainfo, transport, overlay, containment, orders, identity, config.MinBytesPerSecond, config.MinDownloadTimeout),
 		Reporter: NewReporter(log.Named("audit:reporter"), overlay, containment, config.MaxRetriesStatDB, int32(config.MaxReverifyCount)),
 
-		Loop: *sync2.NewCycle(config.Interval),
+		// for audit 2.0
+		overlay:         overlay,
+		reservoirConfig: config.ReservoirConfig,
+		MetainfoLoop:    loop,
+
+		Loop:          *sync2.NewCycle(config.Interval),
```
kaloyan-raev: This is the old "per segment" auditing loop, right? We are going to remove it in favor of the new reservoir loop, right?

satellite/audit/reservoir.go
kaloyan-raev: Yeah, I think we can simplify it to a single "slots" config.

satellite/audit/reservoir.go
navillasa: right now this is in the same package as audits, so it's unexported, but I can go ahead and export it.

satellite/audit/service.go
navillasa: yes, probably not in this pr though

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"math/rand"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Reservoir holds a certain number of segments to reflect a random sample
+type Reservoir struct {
+	Paths []storj.Path
```
kaloyan-raev: I remember a comment from @mobyvb that we may want to have this type generic, so it may be used by other services in the future.

Do we still want to do this? In such case, instead of `Paths []storj.Path`, we should have `Items []interface{}`. And, perhaps, move the type to a separate package under /pkg/reservoirs.

egonelbre: Do not make things unless we have a use-case, because it will end up using more memory and adds more pressure to the GC. I would even go as far as hardcoding the maximum number of paths this reservoir can hold rather than using a slice.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,39 @@
...
+	Size  int
+	index int
```
egonelbre: These can be bytes. And, hardcode the max size as an array.

satellite/audit/service.go
```diff
@@ -63,13 +84,33 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+	//		defer mon.Task()(&ctx)(&err)
+	//		err := service.process(ctx)
+	//		if err != nil {
+	//			service.log.Error("process", zap.Error(err))
+	//		}
+	//		return nil
+	//	})
+	//})
+
+	group.Go(func() error {
```
egonelbre: Is the group necessary here?

Also sync.Cycle has a `Start` that takes group as an argument.

satellite/audit/reservoir.go
kaloyan-raev: Oh, somehow I thought the compiler will optimize the code and the machine code would be more or less the same. If there is a penalty, let's forget about the generalization.

satellite/audit/reservoir.go
kaloyan-raev: The `Size` can be `byte`, but the `index` will take significantly larger values - up to the total number of pieces per node. I can even argue it should be `int64`, if we want storage nodes to store PB of data.

satellite/audit/reservoir.go
egonelbre: Both `interface{}` and a slice create an indirection.

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"storj.io/storj/internal/memory"
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/internal/testrand"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/satellite/audit"
+	"storj.io/storj/storage"
+)
+
+// For every node in testplanet:
```
mobyvb: The setup steps are missed in this comment block (upload 5 files, create new audit observer, etc...)

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,102 @@
...
+			err := uplink.Upload(ctx, satellite, "bucket", path, testData)
+			require.NoError(t, err)
+		}
+
+		observer := audit.NewObserver(zaptest.NewLogger(t), satellite.Overlay.Service, audit.ReservoirConfig{3, 3})
+
+		allPieces := make(map[storj.PieceID]storj.NodeID)
+
+		srv := satellite.Metainfo.Service
+		fmt.Println("srv", srv)
```
mobyvb: `fmt.Println` should be removed here

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,102 @@
...
+				// iterate over every segment in metainfo
+				for it.Next(ctx, &item) {
+					pointer := &pb.Pointer{}
+
+					err = proto.Unmarshal(item.Value, pointer)
+					require.NoError(t, err)
+
+					// todo: why is piece.GetHash() nil?
+					for _, piece := range pointer.GetRemote().GetRemotePieces() {
+						fmt.Println("piece hash", piece.GetHash())
```
mobyvb: get rid of printlns

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+				return nil
+			})
+		require.NoError(t, err)
+
+		err = audits.MetainfoLoop.Join(ctx, observer)
+		require.NoError(t, err)
+
+		for _, node := range planet.StorageNodes {
+			// expect a reservoir for every node
```
mobyvb: unless we upload with an RS config that makes sure we hit every node, we can't guarantee there will be a reservoir for every node. We could either `UploadWithConfig` with an RS config picked based on the number of testplanet nodes, or change our verification logic here to account for nodes that might not have reservoirs

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,39 @@
...
+	return &Reservoir{
+		Size:  size,
+		index: 0,
+	}
+}
+
+// Sample makes sure that for every segment in metainfo from index i=size..n-1,
+// pick a random number r = rand(0..i), and if r < size, replace reservoir.Segments[r] with segment
+func (reservoir *Reservoir) Sample(path storj.Path) {
+	reservoir.index++
```
mobyvb: nit - I think `numSeen` or `seenCount` or something makes more sense than `index` for this variable.

navillasa: you will have to take that up with @kaloyan-raev :) https://github.com/storj/storj/pull/2744#discussion_r312808764

satellite/audit/observer_test.go
navillasa: ahh okay, thank you. I'll use a config based on testplanet nodes I think.

satellite/audit/observer_test.go
navillasa: this was wip to figure out why the panic was happening, but agree

satellite/audit/reservoir.go
mobyvb: <.<

satellite/audit/reservoir.go
navillasa: if we hardcode the max size of the array, then the reservoir size won't be configurable unfortunately!

satellite/audit/service.go
navillasa: the group should be necessary since now we have two different audit loops running even though the new one using the metainfoloop is only creating reservoirs and not using them for selection yet.

satellite/audit/service.go
```diff
@@ -30,6 +31,13 @@ type Config struct {
 	MinBytesPerSecond  memory.Size   `help:"the minimum acceptable bytes that storage nodes can transfer per second to the satellite" default:"128B"`
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"25s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
+
+	ReservoirConfig
+}
+
+// ReservoirConfig contains configurable values for audit reservoirs
+type ReservoirConfig struct {
+	Slots int `help:"number of reservoir slots allotted for nodes" default:"1"`
```
mobyvb: Why not just have `ReservoirSlots` in the main `Config` struct above? Is this extra struct necessary?

satellite/audit/service.go
```diff
@@ -63,13 +83,33 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+			}
+			return nil
+		})
+	})
+
+	group.Go(func() error {
+		return service.ReservoirLoop.Run(ctx, func(ctx context.Context) (err error) {
+			defer mon.Task()(&ctx)(&err)
+			observer := NewObserver(service.log.Named("audit observer"), service.overlay, service.reservoirConfig)
+			err = service.MetainfoLoop.Join(ctx, observer)
```
mobyvb: the reservoirs from the observer should be saved somewhere on the audit service so that they can be accessed from other areas

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/audit"
+	"storj.io/storj/storage"
+	"storj.io/storj/uplink"
+)
+
+// TestAuditObserver does the following:
+// - start testplanet with 5 nodes and a reservoir size of 3
+// - upload like 5 files
```
mobyvb: nit - `upload 5 files`

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,106 @@
...
+// - iterate over all the segments in satellite.Metainfo and store them in allPieces map
+// - create a audit observer and call metaloop.Join(auditObs)
+//
+// Then for every node in testplanet:
+//    - expect that there is a reservoir for that node
```
mobyvb: nit - `expect that there is a reservoir for that node on the audit observer`

satellite/audit/service.go
navillasa: that's true! I'll move it

satellite/audit/observer_test.go
navillasa: oops

satellite/audit/observer.go
```diff
@@ -0,0 +1,56 @@
...
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/overlay"
+)
+
+// Observer observes on the metainfo loop and adds segments to node reservoirs
+type Observer struct {
+	log *zap.Logger
+
+	overlay        *overlay.Service
```
mobyvb: why do you need the overlay to be here? It doesn't look like it is being used

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,106 @@
...
+			err := ul.UploadWithConfig(ctx, satellite, &uplink.RSConfig{
+				MinThreshold:     3,
+				RepairThreshold:  4,
+				SuccessThreshold: 5,
+				MaxThreshold:     5,
+			}, "testbucket", path, testData)
+			require.NoError(t, err)
+		}
+
+		observer := audit.NewObserver(zaptest.NewLogger(t), satellite.Overlay.Service, 3)
```
mobyvb: nit- create this observer right before it is joined to the metainfo loop

satellite/audit/service.go
```diff
@@ -63,13 +79,36 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+	group.Go(func() error {
+		return service.ReservoirLoop.Run(ctx, func(ctx context.Context) (err error) {
+			defer mon.Task()(&ctx)(&err)
+			observer := NewObserver(service.log.Named("audit observer"), service.overlay, service.reservoirSlots)
+			err = service.MetainfoLoop.Join(ctx, observer)
+			if err != nil {
+				service.log.Error("error joining metainfoloop", zap.Error(err))
+				return nil
+			}
+			for nodeID, res := range observer.Reservoirs {
```
mobyvb: I think you should just be able to say `service.Reservoirs = observer.Reservoirs` instead of iterating over every item

navillasa: note for other people thinking this- the reason I looped through is because `Reservoirs` is a map, which is a reference type, so iterating through is necessary to properly copy the values and not just create a reference to the observer.Reservoirs map.

satellite/audit/observer.go
navillasa: oh shoot- took it out when I no longer needed to check if nodes are vetted. I'll remove it.

satellite/audit/observer.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"go.uber.org/zap"
+
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+// Observer observes on the metainfo loop and adds segments to node reservoirs
+type Observer struct {
+	log *zap.Logger
```
mobyvb: nit - doesn't look like this logger is being used. Should we keep it here?

satellite/audit/observer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		require.NoError(t, err)
+
+		observer := audit.NewObserver(zaptest.NewLogger(t), 3)
+		err = audits.MetainfoLoop.Join(ctx, observer)
+		require.NoError(t, err)
+
+		for _, node := range planet.StorageNodes {
+			// expect a reservoir for every node
+			require.NotNil(t, observer.Reservoirs[node.ID()])
+			require.True(t, len(observer.Reservoirs[node.ID()].Paths) <= 3)
```
mobyvb: We should also check that `len(Reservoirs[nodeID].Paths) > 1` since every node should have at least one piece.

satellite/audit/service.go
egonelbre: What I mean is that the loops have the necessary function you don't need to use `group.Go`.

satellite/audit/reservoir.go
egonelbre: It can be configurable up to the specified size.

satellite/audit/service.go
navillasa: so I should be doing it like this with both loops?
```
		service.Loop.Start(ctx, group, func(ctx context.Context) (err error) {
			defer mon.Task()(&ctx)(&err)
			err = service.process(ctx)
			if err != nil {
				service.log.Error("process", zap.Error(err))
			}
			return nil
		})
```

egonelbre: Yeah.

navillasa: thanks!

satellite/audit/reservoir.go
navillasa: ^I'm not totally sure how the config should look/what to implement following this comment.

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,39 @@
...
+	if reservoir.index <= int64(reservoir.Size) {
+		reservoir.Paths = append(reservoir.Paths, path)
+	} else {
+		random := rand.Int63n(reservoir.index)
```
egonelbre: I feel like this might end up hitting the global random lock a lot.

satellite/audit/service.go
```diff
@@ -41,12 +44,18 @@ type Service struct {
 	Reporter reporter
 
 	Loop sync2.Cycle
+
+	// for audit 2.0 using metainfoloop
+	ReservoirLoop  sync2.Cycle
```
egonelbre: Is there a reason it's not implementing a new service instead?

satellite/audit/service.go
```diff
@@ -63,13 +78,32 @@ func (service *Service) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 	service.log.Info("Audit cron is starting up")
 
-	return service.Loop.Run(ctx, func(ctx context.Context) error {
-		err := service.process(ctx)
+	group, ctx := errgroup.WithContext(ctx)
+
+	service.Loop.Start(ctx, group, func(ctx context.Context) (err error) {
```
wthorp: Do we need the entire old service loop?  We probably need reverify, but wasn't the old cursor loop going away?

satellite/audit/reservoir.go
wthorp: https://blog.sgmansfield.com/2016/01/the-hidden-dangers-of-default-rand/

satellite/audit/service.go
navillasa: now it's been separated into audit.Service and audit.ReservoirService

satellite/audit/reservoir.go
navillasa: changed to account for this

satellite/audit/pathcollector.go
```diff
@@ -0,0 +1,51 @@
...
+
+// RemoteSegment takes a remote segment found in metainfo and creates a reservoir for it if it doesn't exist already
+func (pathCollector *PathCollector) RemoteSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx, path)(&err)
+
+	for _, piece := range pointer.GetRemote().GetRemotePieces() {
+		if _, ok := pathCollector.Reservoirs[piece.NodeId]; !ok {
+			pathCollector.Reservoirs[piece.NodeId] = NewReservoir(pathCollector.reservoirSlots)
+		}
+		pathCollector.Reservoirs[piece.NodeId].Sample(pathCollector.rand, path)
```
mobyvb: nit - we could pass `pathCollecter.rand` into `NewReservoir`, then we would only need to pass `path` into `Sample()`.

satellite/audit/pathcollector_test.go
```diff
@@ -0,0 +1,109 @@
...
+		r := rand.New(rand.NewSource(time.Now().Unix()))
+		observer := audit.NewPathCollector(3, r)
+		err = audits.ReservoirService.MetainfoLoop.Join(ctx, observer)
+		require.NoError(t, err)
+
+		for _, node := range planet.StorageNodes {
+			// expect a reservoir for every node
+			require.NotNil(t, observer.Reservoirs[node.ID()])
+			require.True(t, len(observer.Reservoirs[node.ID()].Paths) > 1)
+			require.True(t, len(observer.Reservoirs[node.ID()].Paths) <= 2)
```
mobyvb: Add a comment explaining why you check `<=2` even though you created the path collector with size `3`

satellite/audit/service.go
```diff
@@ -30,6 +32,8 @@ type Config struct {
 	MinBytesPerSecond  memory.Size   `help:"the minimum acceptable bytes that storage nodes can transfer per second to the satellite" default:"128B"`
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"25s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
+
+	Slots int `help:"number of reservoir slots allotted for nodes" default:"1"`
```
mobyvb: Having this `Slots` config seems misleading if we use `2` for all values larger than `2`. I think we should specify that it will be capped at 2 in the help message.

satellite/audit/pathcollector.go
egonelbre: This would add more pointers and make reservoirs types larger.

satellite/audit/pathcollector_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+// TestAuditPathCollector does the following:
+// - start testplanet with 5 nodes and a reservoir size of 3
+// - upload 5 files
+// - iterate over all the segments in satellite.Metainfo and store them in allPieces map
+// - create a audit observer and call metaloop.Join(auditObs)
+//
+// Then for every node in testplanet:
+//    - expect that there is a reservoir for that node on the audit observer
+//    - that the reservoir size is <= 3
```
bryanchriswhite: Is this supposed to match the assertions on lines 100-101?

satellite/audit/reservoir.go
```diff
@@ -0,0 +1,44 @@
...
+	Paths [maxReservoirSize]storj.Path
+	size  int8
+	index int64
+}
+
+// NewReservoir instantiates a Reservoir
+func NewReservoir(size int) *Reservoir {
+	if size > maxReservoirSize {
+		size = maxReservoirSize
+	}
```
kaloyan-raev: What if `size < 1`?

satellite/audit/service.go
```diff
@@ -46,7 +50,7 @@ type Service struct {
 // NewService instantiates a Service with access to a Cursor and Verifier
 func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
 	orders *orders.Service, transport transport.Client, overlay *overlay.Service,
-	containment Containment, identity *identity.FullIdentity) (*Service, error) {
+	containment Containment, identity *identity.FullIdentity, metaLoop *metainfo.Loop) (*Service, error) {
```
kaloyan-raev: It seems we don't use this new `metaLoop` param anywhere.

satellite/audit/service.go
```diff
@@ -58,18 +62,74 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+		defer mon.Task()(&ctx)(&err)
+		pathCollector := NewPathCollector(service.reservoirSlots, service.rand)
+		err = service.MetainfoLoop.Join(ctx, pathCollector)
+		if err != nil {
+			service.log.Error("error joining metainfoloop", zap.Error(err))
+			return nil
+		}
+		for nodeID, res := range pathCollector.Reservoirs {
+			service.Reservoirs[nodeID] = res
+		}
```
kaloyan-raev: What happens if `pathCollector.Reservoirs` does not contain node id that is already present `service.Reservoir`? Will this node id leak in `service.Reservoir`?

Why don't we just make `service.Reservoirs = pathCollector.Reservoirs`?

satellite/audit/service.go
mobyvb: Why is that?

satellite/audit/service.go
navillasa: I thought it wasn't good to do `service.Reservoirs = pathCollector.Reservoirs` because the map is a reference type, so service.Reservoirs would just be pointing to the map at pathCollector.Reservoirs. But if you think that's not a big deal then I'll change it.

satellite/audit/service.go
navillasa: oops, when separating the two services I must've forgotten to delete this param

satellite/audit/pathcollector_test.go
navillasa: yeah thanks for catching

satellite/audit/reservoir.go
navillasa: haha honestly I'm not sure how to handle. I guess I could just set it to maxReservoirSize too? so then the size will always be either 1 or 2 and the configurer will have the illusion of choice

kaloyan-raev: If below 1 then it would be better to set to 1 than to the max.

satellite/audit/service.go
kaloyan-raev: This makes sense. In such case, start with empty `service.Reservoirs` before the for-loop, i.e. `service.Reservoirs = make(map[storj.NodeID]*Reservoir, len(pathCollector.Reservoirs))`. This way we'll avoid leaking reservoirs for nodes that are not present in the latest path collector.

satellite/audit/pathcollector_test.go
```diff
@@ -0,0 +1,112 @@
...
+// - start testplanet with 5 nodes and a reservoir size of 3
+// - upload 5 files
+// - iterate over all the segments in satellite.Metainfo and store them in allPieces map
+// - create a audit observer and call metaloop.Join(auditObs)
+//
+// Then for every node in testplanet:
+//    - expect that there is a reservoir for that node on the audit observer
+//    - that the reservoir size is <= 2 (the maxReservoirSize)
+//    - that every item in the reservoir is unique
+//    - that looking up each pieceID in allPieces results in the same node ID
```
kaloyan-raev: Where do we do this check in the code?

kaloyan-raev: And why do we want to test for this?

navillasa: oops, we didn't end up checking it. we thought we wanted to check in order to make sure that each pieceID was associated with the correct storage node, but on second thought, I'm not sure we're able to do that. so I'm taking this commented line out and the `allPieces` map.

kaloyan-raev: Do we still nee that `satellite.Metainfo.Service.Iterate` block?

satellite/audit/pathcollector.go
```diff
@@ -0,0 +1,51 @@
...
+func NewPathCollector(reservoirSlots int, r *rand.Rand) *PathCollector {
+	return &PathCollector{
+		Reservoirs:     make(map[storj.NodeID]*Reservoir),
+		reservoirSlots: reservoirSlots,
+		rand:           r,
+	}
+}
+
+// RemoteSegment takes a remote segment found in metainfo and creates a reservoir for it if it doesn't exist already
+func (pathCollector *PathCollector) RemoteSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
```
egonelbre: you can name the receiver `collector` I think it's sufficiently descriptive in this context

satellite/audit/pathcollector.go
```diff
@@ -0,0 +1,51 @@
...
+	"math/rand"
+
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+)
+
+// PathCollector uses the metainfo loop to add paths to node reservoirs
+type PathCollector struct {
+	Reservoirs     map[storj.NodeID]*Reservoir
+	reservoirSlots int
```
egonelbre: `slotCount` might be fine

satellite/audit/service.go
```diff
@@ -58,13 +61,64 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+	Reservoirs     map[storj.NodeID]*Reservoir
+	rand           *rand.Rand
+
+	MetainfoLoop  *metainfo.Loop
+	ReservoirLoop sync2.Cycle
+}
+
+// NewReservoirService instantiates Service2
+func NewReservoirService(log *zap.Logger, metaLoop *metainfo.Loop, config Config) (*ReservoirService, error) {
+	r := rand.New(rand.NewSource(time.Now().Unix()))
```
egonelbre: you can inline this

satellite/audit/service.go
```diff
@@ -58,13 +61,64 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+// Once V3-2363 and V3-2364 are implemented, Service2 will replace the existing Service struct.
+type ReservoirService struct {
+	log *zap.Logger
+
+	reservoirSlots int
+	Reservoirs     map[storj.NodeID]*Reservoir
+	rand           *rand.Rand
+
+	MetainfoLoop  *metainfo.Loop
+	ReservoirLoop sync2.Cycle
```
egonelbre: Just Loop is fine here

satellite/audit/service.go
```diff
@@ -58,13 +61,64 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+	return service.ReservoirLoop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+		pathCollector := NewPathCollector(service.reservoirSlots, service.rand)
+		err = service.MetainfoLoop.Join(ctx, pathCollector)
+		if err != nil {
+			service.log.Error("error joining metainfoloop", zap.Error(err))
+			return nil
+		}
+		// We iterate over path collector reservoirs to deep copy them because map is a reference type.
+		service.Reservoirs = make(map[storj.NodeID]*Reservoir, len(pathCollector.Reservoirs))
```
egonelbre: maybe instead `service.Reservoirs = pathCollector.Reservoirs`

satellite/audit/service.go
```diff
@@ -58,13 +61,64 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+}
+
+// NewReservoirService instantiates Service2
+func NewReservoirService(log *zap.Logger, metaLoop *metainfo.Loop, config Config) (*ReservoirService, error) {
+	r := rand.New(rand.NewSource(time.Now().Unix()))
+	return &ReservoirService{
+		log: log,
+
+		reservoirSlots: config.Slots,
+		Reservoirs:     make(map[storj.NodeID]*Reservoir),
```
egonelbre: you can leave this as `nil`

satellite/audit/service.go
egonelbre: Since `pathCollector` is ephemeral, there's no problem in moving the map from one type to another. If the collector was reused then it could be a problem.

satellite/audit/service.go
navillasa: We iterate over path collector reservoirs to deep copy them because map is a reference type.

egonelbre: I know it's a reference type.

navillasa: sorry- deleted comment after I saw your other comment explaining why lol

satellite/audit/pathcollector_test.go
navillasa: no, I got rid of it

satellite/audit/pathcollector.go
```diff
@@ -0,0 +1,51 @@
...
+	return &PathCollector{
+		Reservoirs: make(map[storj.NodeID]*Reservoir),
+		slotCount:  reservoirSlots,
+		rand:       r,
+	}
+}
+
+// RemoteSegment takes a remote segment found in metainfo and creates a reservoir for it if it doesn't exist already
+func (collector *PathCollector) RemoteSegment(ctx context.Context, path storj.Path, pointer *pb.Pointer) (err error) {
+	defer mon.Task()(&ctx, path)(&err)
```
egonelbre: Also slightly worried about this monitoring ending up using a lot of resources.

navillasa: okay I'll take it out

satellite/audit/service.go
```diff
@@ -58,13 +61,59 @@ func NewService(log *zap.Logger, config Config, metainfo *metainfo.Service,
...
+	Loop         sync2.Cycle
+}
+
+// NewReservoirService instantiates Service2
+func NewReservoirService(log *zap.Logger, metaLoop *metainfo.Loop, config Config) (*ReservoirService, error) {
+	return &ReservoirService{
+		log: log,
+
+		reservoirSlots: config.Slots,
+		Reservoirs:     nil,
```
mobyvb: this doesn't need to explicitly be set to `nil`. It will default to that if it is unspecified

PR #2745, opened 2019-08-09
satellite/rewards: update current reward to be finished once redemption cap has reached

VinozzZ:
What: 
- when valid reward constraint fails, we should update current reward to be done

Why:
If current active reward has reached it's maximum redemption capacity, we need to update it to be done and the active offer will fall back to use the default offer

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/satellitedb/usercredits.go
```diff
@@ -70,37 +68,58 @@ func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit)
...
+		shouldCreate = false
+	default:
+		shouldCreate = userCredit.OfferInfo.Status.IsDefault()
+	}
+
+	var dbExec interface {
+		ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
+	}
+
+	if c.tx != nil {
```
rikysya: What will happen with other `usercredit` db calls, if there is a transaction in progress?

VinozzZ: We will use the existing transaction which is what this if statement used for

rikysya: I am talking about other methods like here:
https://github.com/storj/storj/blob/d33c4102567ed451beabcf21307575ab4cc9b86a/satellite/satellitedb/usercredits.go#L29

rikysya: or here: 
https://github.com/storj/storj/blob/d33c4102567ed451beabcf21307575ab4cc9b86a/satellite/satellitedb/usercredits.go#L177

satellite/console/usercredits.go
```diff
@@ -53,3 +54,39 @@ type UserCreditUsage struct {
...
+	CreditsEarned currency.USD
+	ExpiresAt     time.Time
+}
+
+// NewCredit returns a new credit data
+func NewCredit(currentReward *rewards.Offer, creditType CreditType, userID uuid.UUID, referrerID *uuid.UUID) (*CreateCredit, error) {
+	var creditEarned currency.USD
+	switch creditType {
+	case Invitee:
+		creditEarned = currency.Cents(0)
```
coyle: we want to do 0 ?

satellite/satellitedb/usercredits.go
VinozzZ: I can change it to also check if there's existing transaction

satellite/satellitedb/usercredits.go
```diff
@@ -70,37 +68,58 @@ func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit)
 		referrerID = userCredit.ReferredBy[:]
 	}
 
-	var result sql.Result
-	var err error
+	var shouldCreate bool
```
coyle: this seems like it would be simple to do:
```
shouldCreate := false
if !rewards.Partner {
  shouldCreate = userCredit.OfferInfo.Status.IsDefault()
}
```

satellite/satellitedb/usercredits.go
```diff
@@ -70,37 +68,58 @@ func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit)
...
-		}
+		result, err = dbExec.ExecContext(ctx, c.db.Rebind(statement), userCredit.UserID[:], userCredit.OfferID, userCredit.CreditsEarned.Cents(), userCredit.ExpiresAt, referrerID, new([]byte), userCredit.Type, userCredit.OfferID, userCredit.OfferInfo.RedeemableCap, shouldCreate)
 	default:
-		return errs.New("Unsupported database %t", t)
+		return errs.New("unsupported database: %t", t)
 	}
 
 	if err != nil {
+		// check to see if there's a constraint error
+		if driverErr, ok := err.(*pq.Error); (ok && driverErr.Code.Class() == "23") || err == sqlite3.ErrConstraint {
```
coyle: might be good to make 23 a const

satellite/console/service.go
```diff
@@ -176,15 +175,11 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
 
 		if currentReward != nil {
 			// User can only earn credits after activating their account. Therefore, we set the credits to 0 on registration
-			newCredit := UserCredit{
-				UserID:        u.ID,
-				OfferID:       currentReward.ID,
-				ReferredBy:    nil,
-				CreditsEarned: currency.Cents(0),
-				ExpiresAt:     time.Now().UTC().AddDate(0, 0, currentReward.InviteeCreditDurationDays),
+			newCredit, err := NewCredit(currentReward, Invitee, u.ID, nil)
```
coyle: should that always be nil ?

satellite/satellitedb/usercredits.go
VinozzZ: I was thinking that since the logic for giving credit is based on different reward type, should we just create a separate method to handle. I imagine logic may change in the future for the partner once we don't have to bypass the token logic

satellite/console/usercredits.go
VinozzZ: I was debating whether we should have a parameter to pass in the number to make it more flexible. It's 0 here because we can only give invitee credits when they activate their account

satellite/console/service.go
VinozzZ: No, good catch

satellite/satellitedb/usercredits.go
VinozzZ: I found a utility function to check the constraint error

satellite/console/usercredits.go
VinozzZ: Added a comment

satellite/console/service.go
```diff
@@ -175,16 +174,18 @@ func (s *Service) CreateUser(ctx context.Context, user CreateUser, tokenSecret R
...
-				UserID:        u.ID,
-				OfferID:       currentReward.ID,
-				ReferredBy:    nil,
-				CreditsEarned: currency.Cents(0),
-				ExpiresAt:     time.Now().UTC().AddDate(0, 0, currentReward.InviteeCreditDurationDays),
+			var refID *uuid.UUID
+			if refUserID != "" {
+				refID, err = uuid.Parse(refUserID)
+				if err != nil {
+					return errs.New(internalErrMsg)
```
coyle: you don't need to change but typically I prefer to just return `http.StatusInternalServerError`

satellite/satellitedb/usercredits.go
```diff
@@ -110,7 +130,7 @@ func (c *usercredits) Create(ctx context.Context, userCredit console.UserCredit)
 	}
 
 	if rows != 1 {
-		return rewards.MaxRedemptionErr.New("create credit failed")
+		return rewards.MaxRedemptionErr.New("rows not equal to 1")
```
coyle: I think if I were a consumer of this method, I wouldn't understand what `rows not equal to 1` meant as an error.

satellite/satellitedb/usercredits_test.go
```diff
@@ -28,31 +28,16 @@ func TestUsercredits(t *testing.T) {
...
-			{
-				UserID:        user.ID,
-				OfferID:       activeOffer.ID,
-				ReferredBy:    &randomID,
-				CreditsEarned: currency.Cents(100),
-				ExpiresAt:     time.Now().UTC().AddDate(0, 1, 0),
-			},
-		}
+		var invalidUserCredits []console.CreateCredit
+		invalid1, _ := console.NewCredit(activeOffer, console.Invitee, randomID, &referrer.ID)
```
coyle: should the errors be asserted to be nil ?

satellite/satellitedb/usercredits.go
rikysya: Sure, just wanted to make sure that we are aware of the issue if it exists

PR #2746, opened 2019-08-09
uplink/ecclient: performance - do not call Downloader.Close()

mobyvb:
What: 
Instead of closing the grpc ClientStream for a download, and waiting for that to finish, close the underlying grpc ClientConn so we don't have to wait. This has the following performance benefit:

Downloading a 10mb file is on average 3 seconds faster with the change.

`time uplink cp sj://test/10mb.txt ./10mb.txt`

10 downloads with `client.Downloader.Close()` and `client.client.Close()`:
```
real    0m11.046s	0m12.344s 	0m10.895s	0m9.467s	0m11.216s 	0m9.912s	0m11.586s 	0m11.112s 	0m11.634s	0m10.171s
user    0m1.239s	0m1.439s	0m1.541s	0m1.410s	0m1.629s  	0m1.397s	0m1.517s	0m1.439s 	0m1.314s 	0m1.456s
sys     0m0.502s	0m0.566s 	0m0.618s	0m0.645s 	0m0.544s 	0m0.536s 	0m0.586s	0m0.509s	0m0.654s 	0m0.576s

min
real	0m9.467s	
user	0m1.410s	
sys	0m0.544s

max
real	0m12.344s
user	0m1.439s
sys	0m0.566s

avg
real	0m10.938s
user	0m1.438s
sys	0m0.574s
```

10 downloads with only `client.client.Close()`:
```
real    0m7.022s	0m8.467s	0m10.365s	0m9.284s	0m6.775s	0m7.959s	0m4.815s 	0m8.179s	0m7.671s	0m7.756s
user    0m1.501s	0m1.428s 	0m1.350s	0m1.255s	0m1.450s	0m1.260s	0m1.441s	0m1.396s	0m1.531s	0m1.508s
sys     0m0.549s	0m0.438s	0m0.459s	0m0.544s	0m0.523s	0m0.472s	0m0.508s	0m0.470s	0m0.560s	0m0.461s


min
real	0m4.815s
user	0m1.441s
sys	0m0.508s

max
real	0m10.365s
user	0m1.350s
sys	0m0.459s

avg
real	0m7.829s
user	0m1.412s
sys	0m0.498s
```

Why: https://storjlabs.atlassian.net/browse/V3-2245

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/uplink/ecclient/client.go
+++ b/uplink/ecclient/client.go
@@ -494,10 +494,7 @@ type clientCloser struct {
 }
 
 func (client *clientCloser) Close() error {
-	return errs.Combine(
-		client.Downloader.Close(),
-		client.client.Close(),
-	)
+	return client.client.Close()
 }
 
 func nonNilCount(limits []*pb.AddressedOrderLimit) int {


uplink/ecclient/client.go
```diff
@@ -494,10 +494,7 @@ type clientCloser struct {
 }
 
 func (client *clientCloser) Close() error {
-	return errs.Combine(
-		client.Downloader.Close(),
-		client.client.Close(),
-	)
+	return client.client.Close()
```
wthorp: Maybe close Downloader via go routine to log the error and maybe prevent leaks?

mobyvb: From the grpc go docs:
"Close tears down the ClientConn and all underlying connections. "

NewStream creates a new Stream for the client side. This is typically called by generated code. ctx is used for the lifetime of the stream.

To ensure resources are not leaked due to the stream returned, one of the following actions must be performed:

1. Call Close on the ClientConn.
2. Cancel the context provided.
3. Call RecvMsg until a non-nil error is returned. A protobuf-generated
   client-streaming RPC, for instance, might use the helper function
   CloseAndRecv (note that CloseSend does not Recv, therefore is not
   guaranteed to release all resources).
4. Receive a non-nil, non-io.EOF error from Header or SendMsg.


The main thing `Downloader.Close()` does is:
```
closeErr := client.stream.CloseSend()
_, recvErr := client.stream.Recv()
```

Which according to the documentation is not necessary if `ClientConn.Close()` is called

wthorp: well spoken.  :D

egonelbre: We must still close the downloader, otherwise, we end up eventually leaking something, when we change Downloader. There's also a lot of error handling logic this is skipping.

https://github.com/storj/storj/blob/master/uplink/piecestore/download.go#L178

kaloyan-raev: @egonelbre What is your suggestion on this? Call `Downloader.Close()` in a new goroutine?

mniewrzal: I think if something was downloaded without interruption then we can skip closing `Downloader` but if something goes wrong before closing then with `Close` error we can get additional info what happened.

egonelbre: Yes, skip `Recv` when we don't have an error pending.

egonelbre: If we had proper sockets instead of streams, one option would be to close the underlying connection and then try to call Recv... or call Recv in a non-blocking manner... however grpc doesn't support either of them.

mobyvb:
After PR #2711, this change no longer has the performance benefit mentioned. I will continue looking into how closing can be done more quickly

PR #2750, opened 2019-08-09
satellite/console/server: parse html files as templates

crawter:
What: All html files now processed as templates. Also little fix for gzip compression added

Why: 

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact: performance should be improved because templates now will ba cahed and we wont process them with each new request


satellite/console/consoleweb/server.go
```diff
@@ -59,8 +61,22 @@ type Config struct {
...
 	PasswordCost int `internal:"true" help:"password hashing cost (0=automatic)" default:"0"`
+
+	CustomerSupportLink string `help:"link that leads to some resource for customer support" devDefault:"https://storjlabs.atlassian.net/servicedesk/customer/portals" releaseDefault:""`
 }
 
+// consoleTemplate - is used to enumerate template list used in server
+type consoleTemplate int
+
+const (
+	notFound             consoleTemplate = 0
```
rikysya: you can use `iota` to lock values

satellite/console/consoleweb/server.go
```diff
@@ -444,3 +472,66 @@ func (server *Server) gzipHandler(fn http.Handler) http.Handler {
 		fn.ServeHTTP(w, newRequest)
 	})
 }
+
+// initializeTemplates is used to initialize all templates
+func (server *Server) initializeTemplates() (err error) {
+	server.templates = make(map[consoleTemplate]*template.Template)
+
+	notFoundErr := server.parseTemplate(notFound, path.Join(server.config.StaticDir, "static", "errors", "404.html"))
```
rikysya: There is `parseGlob` method for parsing group of templates at once, maybe you can use it here

satellite/console/consoleweb/server.go
```diff
@@ -444,3 +472,66 @@ func (server *Server) gzipHandler(fn http.Handler) http.Handler {
...
+}
+
+// parseTemplate is used to parse template and combine errors while parsing multiple
+func (server *Server) parseTemplate(templateIndex consoleTemplate, templatePath string) error {
+	template, err := template.ParseFiles(templatePath)
+	if err != nil {
+		return err
+	}
+
+	server.templates[templateIndex] = template
```
rikysya: templates can be referenced by name, if I am not mistaken

satellite/console/consoleweb/server.go
```diff
@@ -72,7 +87,8 @@ type Server struct {
 	listener net.Listener
 	server   http.Server
 
-	schema graphql.Schema
+	schema    graphql.Schema
+	templates map[consoleTemplate]*template.Template
```
rikysya: templates can be referenced by name, if I am not mistaken

satellite/console/consoleweb/server.go
```diff
@@ -59,8 +60,22 @@ type Config struct {
 	AuthTokenSecret string `help:"secret used to sign auth tokens" releaseDefault:"" devDefault:"my-suppa-secret-key"`
 
 	PasswordCost int `internal:"true" help:"password hashing cost (0=automatic)" default:"0"`
+
+	CustomerSupportLink string `help:"link that leads to some resource for customer support" devDefault:"https://storjlabs.atlassian.net/servicedesk/customer/portals" releaseDefault:""`
 }
 
+// consoleTemplate - is used to enumerate template list used in server
+type consoleTemplate int
```
egonelbre: Why is this enumeration needed?

satellite/console/consoleweb/server.go
```diff
@@ -72,7 +75,8 @@ type Server struct {
 	listener net.Listener
 	server   http.Server
 
-	schema graphql.Schema
+	schema    graphql.Schema
+	templates map[string]*template.Template
```
egonelbre: Take a look at https://github.com/storj/storj/blob/master/satellite/marketingweb/server.go#L39

egonelbre: In other words you can make a struct with named members and invoke those.

satellite/console/consoleweb/server.go
```diff
@@ -131,7 +173,9 @@ func (server *Server) appHandler(w http.ResponseWriter, req *http.Request) {
 	header.Set("Content-Type", "text/html; charset=UTF-8")
 	header.Set("Content-Security-Policy", strings.Join(cspValues, "; "))
 
-	http.ServeFile(w, req, filepath.Join(server.config.StaticDir, "dist", "index.html"))
+	if server.templates.index == nil || server.templates.index.Execute(w, nil) != nil {
```
rikysya: how template can be nil?

satellite/console/consoleweb/server.go
```diff
@@ -117,8 +125,42 @@ func NewServer(logger *zap.Logger, config Config, service *console.Service, mail
...
+	defer mon.Task()(&ctx)(&err)
+
+	server.schema, err = consoleql.CreateSchema(server.log, server.service, server.mailService)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	err = server.initializeTemplates()
+	if err != nil {
+		// TODO: should it return error if some template can not be initialized or just log about it?
```
egonelbre: Yes, failing to start-up is usually a safer approach.

satellite/console/consoleweb/server.go
egonelbre: Yeah, this should never be nil.

satellite/console/consoleweb/server.go
crawter: but in this case tests will fail, because they didn't generate dist folder

satellite/console/consoleweb/server.go
```diff
@@ -156,8 +200,8 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 	if err != nil {
 		server.log.Error("bucket usage report error", zap.Error(err))
 
-		w.WriteHeader(http.StatusUnauthorized)
-		http.ServeFile(w, r, filepath.Join(server.config.StaticDir, "static", "errors", "404.html"))
+		//TODO: when new error pages will be created - change http.StatusNotFound on http.StatusUnauthorized
+		server.serveError(w, r, http.StatusNotFound)
```
egonelbre: note there's above `http.ServeFile` which also should probably use `server.serveError`

egonelbre: PS: there shouldn't be a `http.ServeFile` anywhere here... this ensures we don't depend on FS directly at the server level.

satellite/console/consoleweb/server.go
```diff
@@ -198,12 +242,9 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	report, err := template.ParseFiles(path.Join(server.config.StaticDir, "static", "reports", "UsageReport.html"))
-	if err != nil {
-		return
+	if err = server.templates.usageReport.Execute(w, bucketRollups); err != nil {
+		server.serveError(w, r, http.StatusNotFound)
```
egonelbre: We probably should log these template errors.

satellite/console/consoleweb/server.go
egonelbre: The templates aren't generated so there should be a template path they can use.

satellite/console/consoleweb/server.go
crawter: server.templates.index will be nil when index.html file is missing. This file is generated by `npm run build` command, which could be not called in some cases. For example in our integration tests

satellite/console/consoleweb/server.go
```diff
@@ -260,58 +301,56 @@ func (server *Server) accountActivationHandler(w http.ResponseWriter, r *http.Re
...
 		}
 
-		http.ServeFile(w, r, filepath.Join(server.config.StaticDir, "static", "resetPassword", "success.html"))
-	case http.MethodGet:
-		t, err := template.ParseFiles(filepath.Join(server.config.StaticDir, "static", "resetPassword", "resetPassword.html"))
-		if err != nil {
-			server.serveError(w, r)
-			return
+		if err := server.templates.success.Execute(w, nil); err != nil {
+			server.serveError(w, r, http.StatusNotFound)
```
egonelbre: For safety return immediately after `serveError`, because often people add some code at the end of the function, and if you don't return from the error handling it could end up executing that code.

satellite/console/consoleweb/server.go
```diff
@@ -260,58 +301,56 @@ func (server *Server) accountActivationHandler(w http.ResponseWriter, r *http.Re
...
 		}
-
-		err = t.Execute(w, nil)
-		if err != nil {
-			server.serveError(w, r)
-			return
+	case http.MethodGet:
+		if err := server.templates.resetPassword.Execute(w, nil); err != nil {
+			server.serveError(w, r, http.StatusNotFound)
```
egonelbre: return after `serveError`

satellite/console/consoleweb/server.go
```diff
@@ -320,19 +359,24 @@ func (server *Server) cancelPasswordRecoveryHandler(w http.ResponseWriter, r *ht
...
 	recoveryToken := r.URL.Query().Get("token")
-	if len(recoveryToken) == 0 {
-		http.Redirect(w, r, "https://storjlabs.atlassian.net/servicedesk/customer/portals", http.StatusSeeOther)
-	}
 
 	// No need to check error as we anyway redirect user to support page
 	_ = server.service.RevokeResetPasswordToken(ctx, recoveryToken)
 
+	// TODO: Should place this link to config
 	http.Redirect(w, r, "https://storjlabs.atlassian.net/servicedesk/customer/portals", http.StatusSeeOther)
```
egonelbre: Yes, use a config flag.

satellite/console/consoleweb/server.go
egonelbre: Can we use some non-generated index.html that will build the javascript things?

satellite/console/consoleweb/server.go
crawter: want to do it in separate pr, because it will change our deployment flow and this pr could ne not approved some time

satellite/console/consoleweb/server.go
crawter: whole dist folder is generated by npm. And we don't have such folder in github sources. To fix it we can add `web/satellite/dist` folder and `web/satellite/dist/index.html` file to git. But i'm not sure that it is a good idea

egonelbre: Let's add a `index.html` that basically contains a text with something like "run npm with XYZ to generate the necessary files"

crawter: @egonelbre may it be done in separate pr also?

egonelbre: Yeah, sure.

satellite/console/consoleweb/server.go
```diff
@@ -320,19 +369,24 @@ func (server *Server) cancelPasswordRecoveryHandler(w http.ResponseWriter, r *ht
...
-	http.ServeFile(w, r, filepath.Join(server.config.StaticDir, "static", "errors", "404.html"))
+func (server *Server) serveError(w http.ResponseWriter, r *http.Request, status int) {
+	// TODO: show different error pages depend on status
+	// F.e. switch(status)
+	//      case http.StatusNotFound: server.executeTemplate(w, r, notFound, nil)
+	//      case http.StatusInternalServerError: server.executeTemplate(w, r, internalError, nil)
+	w.WriteHeader(status)
+
+	if err := server.templates.pageNotFound.Execute(w, nil); err != nil {
+		server.log.Error("error occurred in console/server", zap.Error(err))
```
egonelbre: You don't need to mention the package where the error occurs, since the `server.log` already takes care of it. Here the appropriate message would be `"unable to execute pageNotFound", zap.Error(err)`

PR #2751, opened 2019-08-09
storagenode/console: refactor service and api

rikysya:
What: Refactor service and API endpoints adding missing functionallity from caching service

Why: To expand dashboard with satellite related data and to make use-case of `storagenode/console.Service` scoped to `SNODashboard` needs. Added missing dependencies to `storagenode/console.Service` constructor.

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/console/service.go
```diff
@@ -97,117 +100,132 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
-
-// GetUsedStorageBySatellite returns all info about storagenode disk space usage
-func (s *Service) GetUsedStorageBySatellite(ctx context.Context, satelliteID storj.NodeID) (_ *DiskSpaceInfo, err error) {
-	defer mon.Task()(&ctx)(&err)
-
-	spaceUsed, err := s.pieceStore.SpaceUsedBySatellite(ctx, satelliteID)
-	if err != nil {
-		return nil, err
+	data.DiskSpace = DiskSpaceInfo{
+		Used:      memory.Size(spaceUsege).GB(),
```
crawter: imho bytes better

storagenode/console/service.go
```diff
@@ -97,117 +100,132 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+	StorageDaily   []storageusage.Stamp `json:"storageDaily"`
+	BandwidthDaily []BandwidthUsed      `json:"bandwidthDaily"`
+	Audit          reputation.Metric    `json:"audit"`
+	Uptime         reputation.Metric    `json:"uptime"`
 }
 
-// GetStatsFromSatellite returns storagenode stats from the satellite
-func (s *Service) GetStatsFromSatellite(ctx context.Context, satelliteID storj.NodeID) (_ *reputation.Stats, err error) {
+// GetSatelliteData returns satellite related data
+func (s *Service) GetSatelliteData(ctx context.Context, satellite storj.NodeID) (_ *Satellite, err error) {
```
crawter: would it be better to name this parameter satelliteID?

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +104,71 @@ func (server *Server) Close() error {
...
 }
 
-func (server *Server) checkSatelliteID(satelliteIDs storj.NodeIDList, satelliteID storj.NodeID) error {
-	for _, id := range satelliteIDs {
-		if satelliteID == id {
-			return nil
-		}
+// getSatelliteData returns Satellite data based on provided string representation
+// returns consolidated Satellite data if satelliteID is empty string
+func (server *Server) getSatelliteData(ctx context.Context, satelliteID string) (*console.Satellite, error) {
```
egonelbre: Let's make separate funcs depending on the satelliteID rather than overloading the same call with string.

rikysya: I have 2 separate func `service.GetSatellite` and `service.GetAllSatellitesData`, and invoking one them depending on `satelliteID` param. Do you mean that we should create one more endpoint for consolidated satellite data?

egonelbre: I just mean remove this func and maybe handle the param one level above.

rikysya: Did like that in the first place

But then I have ugly `if/else` block, smth like:
`	satelliteID := strings.TrimLeft(req.URL.Path, "/api/satellite/")
	if satelliteID == "" {
		data.Data, data.Error = server.service.GetAllSatellitesData(ctx)
	} else {
		var id storj.NodeID
		id, data.Error = storj.NodeIDFromString(satelliteID)
		if data.Error == nil {
			data.Data, data.Error = server.service.GetSatelliteData(ctx, id)
		}
	}
`

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +104,71 @@ func (server *Server) Close() error {
...
-	from, to := date.MonthBoundary(time.Now().UTC())
-
-	if satelliteID != nil {
-		return server.service.GetDailyBandwidthUsed(ctx, *satelliteID, from, to)
-	}
+	satelliteID := strings.TrimLeft(req.URL.Path, "/api/satellite/")
 
-	return server.service.GetDailyTotalBandwidthUsed(ctx, from, to)
-}
+	data.Data, data.Error = server.getSatelliteData(ctx, satelliteID)
```
egonelbre: It seems this error is not set in the status code

storagenode/console/consoleserver/server.go
egonelbre:
```
mux.Handle("/api/satellite", http.HandlerFunc(server.satellitesHandler))
mux.Handle("/api/satellite/", http.HandlerFunc(server.satelliteHandler))
```

or 

```
mux.Handle("/api/satellites", http.HandlerFunc(server.satellitesHandler))
mux.Handle("/api/satellite/", http.HandlerFunc(server.satelliteHandler))
```

should also work.

egonelbre: Also since you are using mux package you can use variables

egonelbre: You could also use gorilla mux, which would allow better parsing of path arguments.

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+	Audit          reputation.Metric    `json:"audit"`
+	Uptime         reputation.Metric    `json:"uptime"`
 }
 
-// GetUsedStorageTotal returns all info about storagenode disk space usage
-func (s *Service) GetUsedStorageTotal(ctx context.Context) (_ *DiskSpaceInfo, err error) {
+// GetSatelliteData returns satellite related data
+func (s *Service) GetSatelliteData(ctx context.Context, satelliteID storj.NodeID) (_ *Satellite, err error) {
 	defer mon.Task()(&ctx)(&err)
+	// get current month edges
```
ifraixedes: I don't think that this comment is needed as the `date.MonthBoundary` function documentation already mentions what does this.

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+		BandwidthDaily: bandwidthDaily,
+		Audit:          rep.Audit,
+		Uptime:         rep.Uptime,
+	}, nil
 }
 
-// GetWalletAddress return wallet address of node operator
-func (s *Service) GetWalletAddress(ctx context.Context) string {
+// GetAllSatellitesData returns bandwidth and storage daily usage consolidate
+// among all satellites from node's trust pool
```
ifraixedes:
```suggestion
// among all satellites from the node's trust pool.
```

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+func (s *Service) GetAllSatellitesData(ctx context.Context) (_ *Satellite, err error) {
 	defer mon.Task()(&ctx)(nil)
-	return s.walletAddress
-}
+	// get current month edges
```
ifraixedes: I don't think that this comment is needed as the `date.MonthBoundary` function documentation already mentions what does this.

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
-	return s.reputationDB.Get(ctx, satelliteID)
+	return &Satellite{
+		StorageDaily:   storageDaily,
+		BandwidthDaily: bandwidthDaily,
+	}, nil
 }
 
-// GetDailyStorageUsedForSatellite returns daily SpaceUsageStamps for a particular satellite
-func (s *Service) GetDailyStorageUsedForSatellite(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) (_ []storageusage.Stamp, err error) {
+// VerifySatelliteID verifies if satellite belongs to the trust pool
```
ifraixedes:
```suggestion
// VerifySatelliteID verifies if the satellite belongs to the trust pool.
```

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+func (s *Service) VerifySatelliteID(ctx context.Context, satelliteID storj.NodeID) (err error) {
```
ifraixedes: I think that `VerifySatellite` is a better name.

storagenode/console/service.go
```diff
@@ -97,117 +102,139 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
-// CheckVersion checks to make sure the version is still okay, returning an error if not
-func (s *Service) CheckVersion(ctx context.Context) (err error) {
-	defer mon.Task()(&ctx)(&err)
-	return s.version.CheckVersion(ctx)
+	return nil
 }
 
-// GetSatellites used to retrieve satellites list
-func (s *Service) GetSatellites(ctx context.Context) (_ storj.NodeIDList) {
+// CheckVersion checks if node version >= minAllowedVersion
```
ifraixedes:
```suggestion
// CheckVersion checks if node version >= minAllowedVersion.
```

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,104 @@ func (server *Server) Close() error {
...
+		server.writeError(wr, http.StatusNotFound, err)
+		return
+	}
 
-func (server *Server) getBandwidth(ctx context.Context, satelliteID *storj.NodeID) (_ *console.BandwidthInfo, err error) {
-	if satelliteID != nil {
-		return server.service.GetBandwidthBySatellite(ctx, *satelliteID)
+	data, err := server.service.GetSatelliteData(ctx, satelliteID)
+	if err != nil {
+		server.writeError(wr, http.StatusInternalServerError, err)
```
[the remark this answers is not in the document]
rikysya: Doesn't error has a stack trace?

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,104 @@ func (server *Server) Close() error {
...
+		return
 	}
 
-	return server.service.GetUsedBandwidthTotal(ctx)
+	server.writeData(wr, data)
 }
 
-func (server *Server) getBandwidthChartData(ctx context.Context, satelliteID *storj.NodeID) (_ []console.BandwidthUsed, err error) {
-	from, to := date.MonthBoundary(time.Now().UTC())
+// jsonOutput defines json structure of api response data
```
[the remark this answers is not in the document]
rikysya: Do we have a convention to put dots at the end of the comments?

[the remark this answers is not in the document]
egonelbre: Yes. See https://github.com/storj/docs/blob/master/code/Style.md#comments, https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences and https://golang.org/doc/effective_go.html#commentary.

storagenode/console/consoleserver/server.go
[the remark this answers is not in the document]
ifraixedes: Yes, it does but the stack trace is generated when the error is created by zeebo/errs package and now I see that it will be enough to find out that it's coming from the specific handler function, yesterday I thought that it won't, hence options 2 should be preferred.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,107 @@ func (server *Server) Close() error {
...
-
-	satellites := server.service.GetSatellites(ctx)
-
-	// checks if current satellite id is related to current storage node
-	if satelliteID != nil {
-		if err := server.checkSatelliteID(satellites, *satelliteID); err != nil {
-			return response, err
-		}
-	}
+// satelliteHandler handles satellites api request
```
ifraixedes:
```suggestion
// satelliteHandler handles satellites API request.
```

storagenode/console/consoleserver/server.go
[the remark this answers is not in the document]
rikysya: But do we need to log when client requested untrusted satellite or provide bad satellite name that can't be parsed. Also that is storagenode private server, and the only client is most likely the storagenode owner himself.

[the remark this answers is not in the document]
rikysya: Anyway, moved it to `server.writeError`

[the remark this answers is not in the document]
ifraixedes: I wouldn't log any error which we know the reason.
For example, I wouldn't log `bad satellite name that can't be parsed` the server is acting well if it detects that's bad and it rejects the client because it is sending bad data, so I wouldn't want to see that error in the logs from the server point of view.

The problem with `internal errors` is that we don't know what could be wrong and all the information as possible is welcome to figure it out; ultimately if we have many of them in a short period of time we are sure that something is not behaving well and it's leading the system to a breakdown.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,106 @@ func (server *Server) Close() error {
...
 
-func (server *Server) checkSatelliteID(satelliteIDs storj.NodeIDList, satelliteID storj.NodeID) error {
-	for _, id := range satelliteIDs {
-		if satelliteID == id {
-			return nil
-		}
-	}
+// writeError writes a JSON error payload to http.ResponseWriter log encoding error.
+func (server *Server) writeError(wr http.ResponseWriter, status int, err error) {
+	server.log.Error("api handler error", zap.Int("status code", status), zap.Error(err))
```
ifraixedes: We don't want to log as an error every HTTP status error, for example, `421`, `401` are client error, not a server error.

rikysya: missed the second part somehow:
`Log it inside of server.writeError method doing an if status == http.StatusInternalServerError { log...... }`

storagenode/console/bandwidth.go
```diff
@@ -40,8 +37,8 @@ type Ingress struct {
 type BandwidthInfo struct {
 	Egress    Egress  `json:"egress"`
 	Ingress   Ingress `json:"ingress"`
-	Used      int64   `json:"used"`
-	Remaining int64   `json:"remaining"`
+	Used      float64 `json:"used"`
+	Available float64 `json:"remaining"`
```
stefanbenten: Either change this back to Remaining or adjust the json name 👍

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
...
-	http.ServeFile(writer, request, filepath.Join(server.config.StaticDir, "dist", "index.html"))
+func (server *Server) appHandler(wr http.ResponseWriter, req *http.Request) {
+	http.ServeFile(wr, req, filepath.Join(server.config.StaticDir, "dist", "index.html"))
 }
 
-// appHandler is web app http handler function
-func (server *Server) dashboardHandler(writer http.ResponseWriter, request *http.Request) {
-	ctx := request.Context()
+// dashboardHandler handles dashboard API requests.
+func (server *Server) dashboardHandler(wr http.ResponseWriter, req *http.Request) {
```
stefanbenten: convention is w and r here.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
 }
 
 // appHandler is web app http handler function
-func (server *Server) appHandler(writer http.ResponseWriter, request *http.Request) {
-	http.ServeFile(writer, request, filepath.Join(server.config.StaticDir, "dist", "index.html"))
+func (server *Server) appHandler(wr http.ResponseWriter, req *http.Request) {
```
stefanbenten: convention is w and r here.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
...
-	satellites := server.service.GetSatellites(ctx)
-
-	// checks if current satellite id is related to current storage node
-	if satelliteID != nil {
-		if err := server.checkSatelliteID(satellites, *satelliteID); err != nil {
-			return response, err
-		}
-	}
+// satelliteHandler handles satellites API request.
+func (server *Server) satellitesHandler(wr http.ResponseWriter, req *http.Request) {
```
stefanbenten: convention is w and r here.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
...
+		return
 	}
 
-	walletAddress := server.service.GetWalletAddress(ctx)
+	server.writeData(wr, data)
+}
 
-	versionInfo := server.service.GetVersion(ctx)
+// satelliteHandler handles satellite API requests.
+func (server *Server) satelliteHandler(wr http.ResponseWriter, req *http.Request) {
```
stefanbenten: convention is w and r here.

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
...
+type jsonOutput struct {
+	Data  interface{} `json:"data"`
+	Error string      `json:"error"`
+}
 
-	if satelliteID != nil {
-		return server.service.GetDailyBandwidthUsed(ctx, *satelliteID, from, to)
-	}
+// writeData is helper method to write JSON to http.ResponseWriter and log encoding error.
+func (server *Server) writeData(wr http.ResponseWriter, data interface{}) {
```
stefanbenten: convention is w and r here as well

storagenode/console/consoleserver/server.go
```diff
@@ -127,163 +105,108 @@ func (server *Server) Close() error {
...
-	return server.service.GetUsedStorageTotal(ctx)
 }
 
-func (server *Server) checkSatelliteID(satelliteIDs storj.NodeIDList, satelliteID storj.NodeID) error {
-	for _, id := range satelliteIDs {
-		if satelliteID == id {
-			return nil
-		}
+// writeError writes a JSON error payload to http.ResponseWriter log encoding error.
+func (server *Server) writeError(wr http.ResponseWriter, status int, err error) {
```
stefanbenten: convention is w and r here.

storagenode/console/service.go
```diff
@@ -84,8 +86,11 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
 
 	return &Service{
 		log:                log,
+		trust:              trust,
 		consoleDB:          consoleDB,
 		bandwidthDB:        bandwidth,
+		reputationDB:       reputationDB,
+		storageUsageDB:     storageUsageDB,
```
stefanbenten: I would like to see that we aggregate those two new DBs into a single group/struct

storagenode/console/service.go
```diff
@@ -97,117 +102,143 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
+	data.NodeID = s.kademlia.Local().Id
+	data.Wallet = s.walletAddress
+	data.Version = s.versionInfo.Version
+	data.UpToDate = s.version.IsAllowed()
+	data.Satellites = s.trust.GetSatellites(ctx)
 
-// GetBandwidthBySatellite returns all info about storage node bandwidth usage by satellite
-func (s *Service) GetBandwidthBySatellite(ctx context.Context, satelliteID storj.NodeID) (_ *BandwidthInfo, err error) {
-	defer mon.Task()(&ctx)(&err)
+	spaceUsege, err := s.pieceStore.SpaceUsedForPieces(ctx)
```
stefanbenten: spaceUsage

storagenode/console/service.go
stefanbenten: Typo: spaceUsage and i find the GB value good as it is!

storagenode/console/service.go
stefanbenten: I disagree with that.

storagenode/console/service.go
```diff
@@ -97,117 +102,143 @@ func NewService(log *zap.Logger, consoleDB DB, bandwidth bandwidth.DB, pieceStor
...
-func (s *Service) CheckVersion(ctx context.Context) (err error) {
-	defer mon.Task()(&ctx)(&err)
-	return s.version.CheckVersion(ctx)
+	return nil
 }
 
-// GetSatellites used to retrieve satellites list
-func (s *Service) GetSatellites(ctx context.Context) (_ storj.NodeIDList) {
+// CheckVersion checks if node version >= minAllowedVersion.
+func (s *Service) CheckVersion(ctx context.Context) bool {
```
stefanbenten: I feel like this function is not necessary. Why not asking that below function directly

storagenode/console/service.go
ifraixedes: I was thinking more `VerifySatelliteByID` but it is longer.

However, paying more attention to the implementation of the function, yes your are right about the name, but what's wrong is the doc comment of the function which was it confused to me to suggest the other name.

storagenode/console/bandwidth.go
rikysya: good catch

storagenode/console/service.go
rikysya: https://github.com/storj/storj/pull/2742

storagenode/console/consoleserver/server.go
rikysya: this is not handler

storagenode/console/service.go
rikysya: Because version dependency is encapsulated so I need to expose it(for outer usage), for inner usage I am using it directly. But you are right that right now it is not used.

storagenode/console/consoleserver/server.go
```diff
@@ -121,169 +99,114 @@ func (server *Server) Run(ctx context.Context) (err error) {
...
-	return server.service.GetUsedBandwidthTotal(ctx)
+	server.writeData(w, data)
 }
 
-func (server *Server) getBandwidthChartData(ctx context.Context, satelliteID *storj.NodeID) (_ []console.BandwidthUsed, err error) {
-	from, to := date.MonthBoundary(time.Now().UTC())
+// jsonOutput defines json structure of api response data.
+type jsonOutput struct {
+	Data  interface{} `json:"data"`
+	Error string      `json:"error"`
```
egonelbre: Is this error needed here?

storagenode/console/consoleserver/server.go
egonelbre: It's still managing the same method argument, so changing `w` to `wr` could be confusing and make me think they are different things.

storagenode/console/service.go
egonelbre: I'm actually not quite following why it's here in the first place when the endpoint could use `trust.VerifySatelliteID` directly.

rikysya: `console.Service` dependent on `trust.Pool`, `console.Server` depends on `console.Service`. If `console.Service` is responsible for encapsulating all of the logic needed by Dashboard, then `console.Server` using that dependencies directly makes `console.Service` unreasonable.

rikysya: Either way it is the `server` or `service` that has tons of dependencies

egonelbre: It depends who is responsible for handling authentication.

If `console.Server` is responsible for the authentication, then using trust there would be completely fine.

If `console.Service` is responsible for the authentication, then the call for `VerifySatellite` shouldn't be probably exposed. Since server shouldn't be worried about it.

rikysya: The only reason I moved it out from `service.GetSatellite` method is to return `http,BadRequest` status code in handler, but in my head `console.Service` is responsible for verifying `satellite`

PR #2753, opened 2019-08-09
storagenode: add space used cache for pieces

JessicaGreben:
**What/Why:**

When a piece is uploaded to a storagenode, it first calculates how much free space it has available by iteratively walking through all piecefiles (pieces are stored as files on the storagenode, so we call them piecefiles) stored on disk, stats each piecefile to get size, then sums up those sizes. This is very slow. 

The goal of this PR is to improve the performance by storing live used-space totals in-memory. To accomplish that this PR does makes the following changes:
- adds a data structure to the storagenode that contains a cache for used-space totals
- adds a sqlite table `pieceSpaceUsedDB` that stores the most recent space used cache values
- adds a CacheUpdate service that runs on an interval and recalculates the cache values to make sure the cache doesn't become inaccurate
- update the storagenode services that previously used the iterative approach to calculate space used totals to use the spaceUsedCache values instead. This includes `storagenode/monitor`, `storagenode/inspector`, and `storagenode/console`.

Lifecycle of the spaceUsedCache:
- when the storagenode starts, the spaceUsedCache is initialized with values from the `pieceSpaceUsedDB` 
- when pieces are created or deleted, the spaceUsedCache is updated
- when the CacheUpdate service runs, the spaceUsedCache is updated with those new values (and those values are saved to pieceSpaceUsedDB)

Reference: see this old closed PR that is an initial implementation for this same code for other comments https://github.com/storj/storj/pull/2707

**Describe the tests:**
WIP
 
**Describe the performance impact:**
The performance will be improved now that we are using a value in memory instead of iterating over every piecefile on disk.


storagenode/pieces/blobcache.go
```diff
@@ -0,0 +1,272 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package pieces
+
+import (
+	"context"
+	"storj.io/storj/pkg/storj"
```
zeebo: import in the wrong spot

storagenode/pieces/blobcache.go
```diff
@@ -0,0 +1,272 @@
...
+	blobs.cache.totalBySatellite[satelliteID] += pieceContentSize
+}
+
+// Close satisfies the pieces interface
+func (blobs *BlobsUsageCache) Close() error {
+	return nil
+}
+
+func (blobs *BlobsUsageCache) recalculate(ctx context.Context, newTotalSpaceUsed int64, newTotalSpaceUsedByNamespace map[storj.NodeID]int64) error {
+	spaceUsedWhenIterationStarted := blobs.cache
```
zeebo: isn't there a small data race here where an update call can come in while recalculate is happening? and by small i mean "formally incorrect so possibly triggered by the race detector but would never actually happen in practice."

also, this isn't really the space used when the iteration started, this is also the space used when the iteration ended. the iteration happens before we even call into recalculate.

storagenode/pieces/blobcache.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	err = blobs.Blobs.Delete(ctx, blobRef)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+
+	satelliteID := storj.NodeID{}
+	satelliteID.Scan(blobRef.Namespace)
+	blobs.update(ctx, satelliteID, pieceContentSize)
```
zeebo: shouldn't this be `-pieceContentSize`?

storagenode/pieces/blobcache.go
```diff
@@ -0,0 +1,272 @@
...
+	// get the size written we commit that way this
+	// value will only include the piece content size and not
+	// the header bytes
+	size, err := blob.BlobWriter.Size()
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	satelliteID := storj.NodeID{}
+	satelliteID.Scan(blob.namespace)
+	blob.usageCache.update(ctx, satelliteID, size)
```
zeebo: since the usage cache update can't fail, shouldn't we do it after the commit?

storagenode/pieces/blobcache.go
JessicaGreben: oh yeah these are all messed up. _fixing_

storagenode/pieces/blobcache.go
JessicaGreben: yeah thanks! i missed that

storagenode/pieces/blobcache.go
JessicaGreben: i was debating doing if after or before commit. all the commit does is write the header bytes, the piece content is already written. that is why i updated before

storagenode/pieces/blobcache.go
JessicaGreben: yeah this variable name is wrong, it should be "Ended" not "Started". Also I need to fix the code, because I'm not even tracking the "Started" case. Oops. 

About the data race situation... currently we recalculate and reset the in-memory used space values on startup, but while that is occurring, some updates could come in. I asked an open question about that and here is a possible solution that Paul came up with https://github.com/storj/storj/pull/2707#issuecomment-517965671.  I decided to implement that here, but I actually did it wrong, need to fix that now.

zeebo: That comment/calculation is useful to handle the **race condition**, but I’m talking about a **data race**: two concurrent processes accessing the same memory location where at least one is a write. In other words, the update call protects it’s write with a mutex, and since an update can be concurrent with this read, the read must also be protected by a mutex.

JessicaGreben: i can add a mu.lock here as well, i just assumed since the result of `recalculate` is an estimation anyways that the race condition here didn't actually matter.

storagenode/pieces/blobcache.go
zeebo: sure, but if it’s after, you don’t have to know what commit does to agree the behavior is right. you’re sure that update only happens on correctly committed pieces.

JessicaGreben: good point. i will move the update to after the commit

storagenode/pieces/blobcache.go
zeebo: i think it’s always important to have no data races, even benign ones. this way, there’s no false positives with the race detector, and there are good arguments that there’s no such thing as a benign data race since optimizing compilers assume your program is free of them. another great article: https://software.intel.com/en-us/blogs/2013/01/06/benign-data-races-what-could-possibly-go-wrong

JessicaGreben: that all makes sense. and yeah seems better to avoid race conditions all together. thanks. i will fix.

JessicaGreben: and wow that intel blog really drives home the point. ok ok i will add a lock. haha

storagenode/storagenodedb/piecespaceused.go
```diff
@@ -0,0 +1,161 @@
...
+
+type pieceSpaceUsedDB struct {
+	*InfoDB
+}
+
+// PieceSpaceUsedDB returns database for storing piece expiration data
+func (db *DB) PieceSpaceUsedDB() pieces.PieceSpaceUsedDB { return db.info.PieceSpaceUsedDB() }
+
+// PieceSpaceUsedDB returns database for storing piece expiration data
+func (db *InfoDB) PieceSpaceUsedDB() pieces.PieceSpaceUsedDB { return &db.pieceSpaceUsedDB }
```
jenlij: why wrap twice?

JessicaGreben: i updated the comments for clarity. One is exposing the DB from infoDb, and the other is exposing via the storagenode master db

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,295 @@
...
+	return missed
+}
+
+// Close satisfies the pieces interface
+func (blobs *BlobsUsageCache) Close() error {
+	return nil
+}
+
+// TestCreateV0 creates a new V0 blob that can be written. This is only appropriate in test situations.
+func (blobs *BlobsUsageCache) TestCreateV0(ctx context.Context, ref storage.BlobRef) (_ storage.BlobWriter, err error) {
```
jenlij: is there any way to move this to a test specific file?

storagenode/pieces/cache.go
```diff
@@ -0,0 +1,295 @@
...
+
+	"storj.io/storj/internal/sync2"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storage"
+)
+
+// CacheService updates the space used cache
+type CacheService struct {
+	log        *zap.Logger
+	usageCache *BlobsUsageCache
```
jenlij: It's a little unclear to me why we need both a cache and a blob cache. Could you explain this to me again?

JessicaGreben: the cacheservice runs on a loop and syncs the cache totals to the database.  The BlobsUsageCache wraps the storage.Blobs with the actual cache

storagenode/pieces/cache.go
JessicaGreben: i could move, but then i would need to import this package. i would prefer to leave here.

PR #2757, opened 2019-08-09
uplink/ecclient: performance - close connections faster

mobyvb:
What: 
Create a new context and cancelation function from inside `decodedRanger.Range()`, and use these within `DecodeReaders()` rather than creating them inside `DecodeReaders()`.

It solves an issue related to read calls blocking close calls when we are finished with a download.

20 downloads of a 10mb file (average):
master: 9.64085s
green/close-grpc: 7.51985s
difference: 2.121s

Why: https://storjlabs.atlassian.net/browse/V3-2245

Please describe the tests:
 - Test 1:
 - Test 2:
 

mobyvb:
@egonelbre do you know in what case we should decide to close the downloader as well? I am not sure how to determine whether the reader is being closed because the download is done or because it is ending early.

egonelbre:
You should always close the downloader. However, you shouldn't always try to read the error. In downloader close you can skip recv when unread doesn't contain an error.

uplink/ecclient/client.go
```diff
@@ -541,10 +541,10 @@ func (lr *lazyPieceReader) Close() (err error) {
 	}
 	lr.isClosed = true
 
-	for _, c := range lr.closers {
-		err = errs.Combine(err, c.Close())
-	}
-	return err
+	return errs.Combine(
+		lr.Downloader.Close(),
```
egonelbre: Downloader and client might be nil here.

uplink/piecestore/download.go
```diff
@@ -189,21 +189,25 @@ func (client *Download) Close() (err error) {
 
 	// close our sending end
 	closeErr := client.stream.CloseSend()
-	// try to read any pending error message
-	_, recvErr := client.stream.Recv()
 
 	if alldone {
 		// if we are all done, then we expecte io.EOF, but don't care about them
```
wthorp: 'expecte' spelling

uplink/eestream/decode.go
```diff
@@ -192,6 +192,8 @@ func (dr *decodedRanger) Size() int64 {
 
 func (dr *decodedRanger) Range(ctx context.Context, offset, length int64) (_ io.ReadCloser, err error) {
 	defer mon.Task()(&ctx)(&err)
+
+	ctx, cancel := context.WithCancel(ctx)
```
wthorp: Maybe add a comment on line 204 saying what concrete type the 'rr' is so that the context cancellation stuff makes more sense.

satellite/orders/orders_test.go
```diff
@@ -131,6 +131,8 @@ func TestUploadDownloadBandwidth(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, expectedData, data)
 
+		// We need enough time to pass after the download ends for storagenodes to save orders
+		time.Sleep(200 * time.Millisecond)
```
mobyvb: note for reviewers - the `Download()` call finishing faster means that there is not always enough time for storagenodes to save orders before we calculate totals in this test. I do not know how to sync this test without a time.Sleep

mobyvb:
Wait for an approval from a principal engineer before merging

satellite/orders/orders_test.go
kaloyan-raev: Yeah, I also don't see an alternative way.

uplink/eestream/decode.go
kaloyan-raev: As far as I understand, with this change, closing the DecodeReader will cancel the context for all piece readers. This happens right before calling the Close() method of the piece readers.

Will it still be possible for `Download.Close()` to read the read via `client.stream.Recv()` if the context is already canceled?

egonelbre: Ah, right.. probably not.

mobyvb: `client.stream.Recv()` will return a context canceled error in some cases, but we are ignoring these errors in `Download.Close()`

kaloyan-raev: The question is - will it return the actual error that happened instead of "context canceled"?

mobyvb: No, I added a function `ignoredCanceled(err)` that returns nil if `errs2.IsCanceled(err)` is true. Which means if there is some other error, it will be ignored. I do not know what circumstances could cause a context canceled error + some other error wrapped together.

uplink/piecestore/download.go
```diff
@@ -189,21 +190,25 @@ func (client *Download) Close() (err error) {
...
 	if alldone {
-		// if we are all done, then we expecte io.EOF, but don't care about them
-		return errs.Combine(ignoreEOF(closeErr), ignoreEOF(recvErr))
+		// if we are all done, then we expect io.EOF, but don't care about them
+		return ignoreEOF(closeErr)
 	}
 
 	if client.unread.Errored() {
+		// try to read any pending error message
+		_, recvErr := client.stream.Recv()
```
mobyvb: note from meeting (@zeebo) - do not call `Recv()` here to reduce latency

uplink/eestream/decode.go
mobyvb: Note from meeting - create ticket for making sure we can properly read errors in download.Close() in the case of a canceled context + some other error we do not want the context cancelation to obscure.

egonelbre: I created https://github.com/storj/storj/pull/2771 which should handle the errors better.

uplink/piecestore/download.go
```diff
@@ -242,7 +242,7 @@ func (client *Download) Close() (err error) {
 	}()
 
 	client.closeWithError(nil)
-	return client.closingError
+	return errs2.IgnoreCanceled(client.closingError)
```
egonelbre: I think this cancellation should be handled a level higher.

PR #2754, opened 2019-08-09
Updates the pkg/identity support encode and decode functionality of Peer Identity information

aligeti:
What: Updates the Identity packet to support encoding/decoding of the peer Identity to/From bytes, which can be used to save the whole chain of peer identities in database that stores these information. 
Why:
This PR  is a sub task of the a big feature as described in V3-1992 epic. The V3-1992 will be divided into 6 PRs. This is the first of the 6 PRs.

**The 6 PRs would be as shown below:**
1.  make the encode /decode changes  into the pkg/identities ( PR #2754 )
2. remove certdb (because the design changed quite drastically)
3. implement new certdb (dbx and interface and tests)
4. fix those bootstrap error outputting debug info
5. fix the segment debug outputing info (edited) 
6. wire up the new certdb (edited) 

Please describe the tests:
 - Test 1:adds test function to encode
 - Test 2:adds test function to decode
 - Test 3: compares the peer identities by first encodes, then decodes,  and then compares decoded data with the original information used in encoding. 
 
Please describe the performance impact:

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?
 - [ ] Do the database access patterns make sense?

PR #2760, opened 2019-08-10
satellite/certdb: remove certdb that was used to store uplink certificates

aligeti:
What: 
**satellitedb/certDB: refactors of the node certificate storage DB table**

   The existing implementation doesnt allow to store the complete certificate chain of uplinkIDs or storagenodeIDs, so the current table is dropped and new table will be added which addresses the storage and retrieval of certificates

**pkg/identity: fixes spelling mistakes that I missed on PR #2754**

    Fixes V3-1992/V3-2388

Why: This is a small PR of the below epic
https://storjlabs.atlassian.net/browse/V3-1992

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/orders/endpoint.go
```diff
@@ -152,17 +149,21 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
-					if err != nil {
-						log.Warn("unable to find uplink public key", zap.Error(err))
-						return status.Errorf(codes.Internal, "unable to find uplink public key")
-					}
-					uplinkSignee = &signing.PublicKey{
-						Self: *orderLimit.DeprecatedUplinkId,
-						Key:  uplinkPubKey,
-					}
 				}
+				// @TODO: KISHORE this portion of verifying uplink's public key is disabled until certdb is refactored
```
egonelbre: We can completely remove the old logic, this shouldn't be needed anymore. Order limits using this logic should already have been sent to the satellite.

aligeti: updated. Removed the logic and added a comment to address the implementation in next PR.

satellite/orders/endpoint.go
```diff
@@ -142,32 +139,15 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
-					return Error.New("unable to verify order")
-				}
+			// satellite verifies that it signed the order limit
+			if err := signing.VerifyUplinkOrderSignature(ctx, orderLimit.UplinkPublicKey, order); err != nil {
+				return Error.New("unable to verify order")
 			}
 
+			// TODO: add here .. satellite verifies that the order signature matches pub key in order limit
+			// TODO: implementation logic would be read back the stored public key from DB and verifies the
+			// public key stored in the order limit. Will be implemented in next PR.  [KISHORE]
```
egonelbre: It's already doing it.

aligeti: where is it doing? am bit lost ... I see **VerifyOrderLimitSignature()** & **VerifyUplinkOrderSignature()** are doing it..but it is verifying based on the public key passed in the orderlimit, which a rouge node can change it. dont we need to access the corresponding uplink id's public key from a DB and verify?

egonelbre: A bad storage node cannot change the order limit, since it's signed by the satellite.

egonelbre:
@littleskunk after this PR, the satellite will reject orders that are signed with uplink public certificate. I think there shouldn't be any left at this point.

satellite/orders/endpoint.go
aligeti: OOO ... my bad..... got it there is a satellite signature... yep got it thnx... is it looks good? Also in that case we never need to store uplink's public key. am I right?

egonelbre: We wouldn't have enough storage to keep them. So, yes, we shouldn't need them anymore.

aligeti: now it made sense... I cleaned up the code ... please let me know what you think.

satellite/orders/endpoint.go
```diff
@@ -138,36 +135,11 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
 		}
 
 		rejectErr := func() error {
+			// satellite verifies that it signed the order limit
 			if err := signing.VerifyOrderLimitSignature(ctx, endpoint.satelliteSignee, orderLimit); err != nil {
 				return Error.New("unable to verify order limit")
 			}
 
-			if orderLimit.DeprecatedUplinkId == nil { // new signature handling
-				if err := signing.VerifyUplinkOrderSignature(ctx, orderLimit.UplinkPublicKey, order); err != nil {
```
egonelbre: Why did you remove order signature checking?

aligeti: Got it.. got it.. .got it.. i was under the impression that order limit has all the info of actual uplink's order... I looked at it completely wrong. there were two orders sent back to satellite, 1. the original order limit and 2. the uplink order. Thnx... it is clear... we need to have VerifyUplinkOrderSignature() to verify the order limit that uplink sent to Storagenodes, with the public key that satellite sent to uplink as part of the order limit. Got it.. thnx .. hope I got it this time...  👍

PR #2762, opened 2019-08-12
Update satellite settlement endpoint to batch order processing into transactions.

ethanadams:
What: 
Update satellite settlement endpoint to batch order processing into transactions.

Why:
Minimize # of transactions created and lower traffic to the DB while processing orders.

Please describe the tests:
 - Test 1: Existing tests
 - Test 2:
 

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+		} else {
+			err = errs.Combine(err, tx.Rollback())
+		}
+	}()
+
+	now := time.Now().UTC()
+	intervalStart := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), 0, 0, 0, now.Location())
+
+	rejectedRequests := make(map[storj.SerialNumber]bool)
+	reject := func(serialNumber storj.SerialNumber) {
```
jenlij: Could you add a comment to explain what this is doing?

jenlij: or maybe why the helper function is defined inside the method?

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+				reject(request.OrderLimit.SerialNumber)
+			}
+		}
+	}
+
+	_, err = tx.Exec(updateRollupStatement)
+	if err != nil {
+		return []*pb.SettlementResponse{}, errs.Wrap(err)
+	}
+	for _, request := range requests {
```
jenlij: Is there any way we could add this logic inside the for loop on line 297?

satellite/orders/endpoint.go
```diff
@@ -110,8 +121,19 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
 
 	log := endpoint.log.Named(peer.ID.String())
 	log.Debug("Settlement")
+
+	requests := make([]*ProcessOrderRequest, 0, endpoint.settlementBatchSize)
+
+	defer func() {
+		if len(requests) > 0 {
+			err = endpoint.processOrders(ctx, stream, requests)
```
zeebo: We should use `errs.Combine` to not toss out any other error that may have happened.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+		}
+	}
+
+	// call to get all the bucket IDs
+	query := db.buildGetBucketIdsQuery(len(requests))
+	statement := db.db.Rebind(query)
+
+	args := make([]interface{}, 0, len(requests))
+	for _, request := range requests {
+		args = append(args, request.OrderLimit.SerialNumber.Bytes())
```
jenlij: could we add this to the for loop on line 257?

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
 	_, err = db.db.ExecContext(ctx, db.db.Rebind(statement), storageNodeID.Bytes(), serialNumber.Bytes())
 	return err
 }
+
+// ProcessOrders take a list of order requests and "settles" them in one transaction
+func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.ProcessOrderRequest) (responses []*pb.SettlementResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if len(requests) == 0 {
+		return []*pb.SettlementResponse{}, err
```
zeebo: We should just return `nil, err` instead of `[]*pb.SettlementResponse{}, err` everywhere.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+	}
+	return fmt.Sprintf("SELECT serial_number, bucket_id FROM serial_numbers WHERE serial_number IN (%s);\n", strings.Join(args, ","))
+}
+
+func (db *ordersDB) buildUpdateBucketBandwidthRollupStatements(orderLimit *pb.OrderLimit, order *pb.Order, projectID []byte, bucketName []byte, intervalStart time.Time) string {
+	return fmt.Sprintf(`INSERT INTO bucket_bandwidth_rollups (bucket_name, project_id, interval_start, interval_seconds, action, inline, allocated, settled)
+		VALUES (%s, %s, '%s', %d, %d, %d, %d, %d)
+		ON CONFLICT(bucket_name, project_id, interval_start, action)
+		DO UPDATE SET settled = bucket_bandwidth_rollups.settled + %d;
+`, db.toHex(bucketName), db.toHex(projectID), intervalStart.Format("2006-01-02 15:04:05+00:00"), defaultIntervalSeconds, orderLimit.Action, 0, 0, uint64(order.Amount), uint64(order.Amount))
```
jenlij: is there an extra `uint64(order.Amount))`?

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+}
+
+func (db *ordersDB) buildUpdateStorageNodeBandwidthRollupStatements(orderLimit *pb.OrderLimit, order *pb.Order, intervalStart time.Time) string {
+	return fmt.Sprintf(`INSERT INTO storagenode_bandwidth_rollups (storagenode_id, interval_start, interval_seconds, action, allocated, settled)
+		VALUES (%s, '%s', %d, %d, %d, %d)
+		ON CONFLICT(storagenode_id, interval_start, action)
+		DO UPDATE SET settled = storagenode_bandwidth_rollups.settled + %d;
+`, db.toHex(orderLimit.StorageNodeId.Bytes()), intervalStart.Format("2006-01-02 15:04:05+00:00"), defaultIntervalSeconds, orderLimit.Action, 0, uint64(order.Amount), uint64(order.Amount))
```
jenlij: same here?

satellite/orders/endpoint.go
```diff
@@ -167,55 +189,31 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
-		if err != nil {
-			return err
-		}
-		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, *projectID, bucketName, orderLimit.Action, order.Amount, intervalStart); err != nil {
-			// TODO: we should use the serial number in the same transaction we settle the bandwidth? that way we don't need this undo in an error case?
-			if err := endpoint.DB.UnuseSerialNumber(ctx, orderLimit.SerialNumber, orderLimit.StorageNodeId); err != nil {
-				log.Error("unable to unuse serial number", zap.Error(err))
+		if len(requests) == endpoint.settlementBatchSize {
+			err = endpoint.processOrders(ctx, stream, requests)
+			requests = requests[:0]
```
jenlij: what's going on here with requests[:0]

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,155 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+
+func (db *ordersDB) toHex(value []byte) string {
+	hexValue := hex.EncodeToString(value)
+	switch db.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		return fmt.Sprintf("X'%v'", hexValue)
+	case *pq.Driver:
+		return fmt.Sprintf("decode('%v', 'hex')", hexValue)
+	default:
+		return ""
```
zeebo: This seems like a dangerous thing to me. I know that we use the string as a value in places, so it would be a lot of extra code checking errors, but I'd feel much more comfortable if adding a new database backend (which is expected to happen) didn't have to include finding this code silently corrupting the data (even in the best case of unit test failures).

satellite/orders/endpoint.go
isaachess: This re-slicing makes `requests` an empty slice with length of 0, but maintains its capacity.

satellite/orders/endpoint.go
```diff
@@ -167,55 +189,31 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
 
-		projectID, bucketName, err := SplitBucketID(bucketID)
-		if err != nil {
-			return err
-		}
-		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, *projectID, bucketName, orderLimit.Action, order.Amount, intervalStart); err != nil {
-			// TODO: we should use the serial number in the same transaction we settle the bandwidth? that way we don't need this undo in an error case?
-			if err := endpoint.DB.UnuseSerialNumber(ctx, orderLimit.SerialNumber, orderLimit.StorageNodeId); err != nil {
-				log.Error("unable to unuse serial number", zap.Error(err))
+		if len(requests) == endpoint.settlementBatchSize {
```
isaachess: This may seem overly paranoid, but I'd prefer we change to `>=` instead of `==`, just in case something weird occurs in the future where we skip the exact batch size.

satellite/orders/endpoint.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/orders/endpoint.go
ethanadams: Updated

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+		r := &pb.SettlementResponse{
+			SerialNumber: serialNumber,
+			Status:       pb.SettlementResponse_REJECTED,
+		}
+		rejectedRequests[serialNumber] = true
+		responses = append(responses, r)
+	}
+
+	// processes the insert to used serials table individually so we can handle
+	// the case where the order has already been process.  Duplicates and previously
```
isaachess: `process` should be `processed`

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+				return nil, Error.Wrap(err)
+			}
+		}
+	}
+
+	// call to get all the bucket IDs
+	query := db.buildGetBucketIdsQuery(len(requests))
+	statement := db.db.Rebind(query)
+
+	args := make([]interface{}, 0, len(requests))
```
isaachess: I believe it isn't typically helpful to set the capacity but not the length. It is usually sufficient (and faster) to simply do:

`args := make([]inteface{}, len(requests))`

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+			return nil, errs.Wrap(err)
+		}
+		bucketMap[sn] = bucketID
+	}
+
+	// build all the bandwidth updates into one sql statement
+	var updateRollupStatement string
+	for _, request := range requests {
+		_, rejected := rejectedRequests[request.OrderLimit.SerialNumber]
+		if !rejected {
```
isaachess: I would change this to be:

```
if rejected {
    continue
}
```

This is allows you to not indent below, makes the code more readable.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+			bucketID, ok := bucketMap[request.OrderLimit.SerialNumber]
+			if ok {
```
isaachess: Similarly here, if you do:

```
if !ok {
    reject(request.OrderLimit.SerialNumber)
    continue
}
```

this allows you to save an indentation level, by handling the unusual case first. (I admit this one is more of a nitpick, but just a suggestion.)

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+		}
+	}
+
+	_, err = tx.Exec(updateRollupStatement)
+	if err != nil {
+		return nil, errs.Wrap(err)
+	}
+	for _, request := range requests {
+		_, rejected := rejectedRequests[request.OrderLimit.SerialNumber]
+		if !rejected {
```
isaachess: Similar stylistic comment here, I really like:

```
if rejected {
    continue
}
```

then you can do without the extra indentation and closure.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,191 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+	hexNodeID, err := db.toHex(orderLimit.StorageNodeId.Bytes())
+	if err != nil {
+		return "", err
+	}
+	hexSN, err := db.toHex(orderLimit.SerialNumber.Bytes())
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("INSERT INTO  used_serials (serial_number_id, storage_node_id) SELECT id, %v FROM serial_numbers WHERE serial_number = %s;\n",
```
isaachess: In general, it's best to avoid using `fmt.Sprintf` when preparing sql statements, because you risk sql injection and other problems. If you use the `db.Prepare` method, and then use `*sql.Stmt` structs to apply args to the query, you avoid these problems.

Maybe that's not possible given the way you're doing it here? But in general I believe it is best practice to avoid string formatting on sql queries.

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated

satellite/satellitedb/orders.go
ethanadams: Updated. Left the others per our conversation

satellite/satellitedb/orders.go
isaachess: I think you missed this one. Otherwise, I'm approving.

satellite/orders/service.go
```diff
@@ -22,7 +22,8 @@ import (
 
 // Config is a configuration struct for orders Service.
 type Config struct {
-	Expiration time.Duration `help:"how long until an order expires" default:"168h"` // 7 days
+	Expiration         time.Duration `help:"how long until an order expires" default:"168h"` // 7 days
+	SettlmentBatchSize int           `help:"how many orders to batch per transaction" default:"250"`
```
egonelbre: Typo

satellite/orders/endpoint.go
```diff
@@ -46,6 +46,9 @@ type DB interface {
 	GetBucketBandwidth(ctx context.Context, projectID uuid.UUID, bucketName []byte, from, to time.Time) (int64, error)
 	// GetStorageNodeBandwidth gets total storage node bandwidth from period of time
 	GetStorageNodeBandwidth(ctx context.Context, nodeID storj.NodeID, from, to time.Time) (int64, error)
+
+	// ProcessOrders takes a list of order requests and processes them in a batch
+	ProcessOrders(ctx context.Context, requests []*ProcessOrderRequest) (responses []*pb.SettlementResponse, err error)
```
egonelbre: Try to avoid protobuf types in database code.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,178 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+func (db *ordersDB) buildGetBucketIdsQuery(argCount int) string {
+	args := make([]string, argCount)
+	for i := 0; i < argCount; i++ {
+		args[i] = "?"
+	}
+	return fmt.Sprintf("SELECT serial_number, bucket_id FROM serial_numbers WHERE serial_number IN (%s);\n", strings.Join(args, ","))
+}
+
+func (db *ordersDB) buildUpdateBucketBandwidthRollupStatements(orderLimit *pb.OrderLimit, order *pb.Order, projectID []byte, bucketName []byte, intervalStart time.Time) (string, error) {
+	hexName, err := db.toHex(bucketName)
```
egonelbre: Why do they need to be converted to hex?

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,178 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+	}
+	return responses, nil
+}
+
+func (db *ordersDB) buildGetBucketIdsQuery(argCount int) string {
+	args := make([]string, argCount)
+	for i := 0; i < argCount; i++ {
+		args[i] = "?"
+	}
+	return fmt.Sprintf("SELECT serial_number, bucket_id FROM serial_numbers WHERE serial_number IN (%s);\n", strings.Join(args, ","))
```
egonelbre: there is `strings.Repeat`

satellite/orders/endpoint.go
```diff
@@ -167,55 +189,31 @@ func (endpoint *Endpoint) Settlement(stream pb.Orders_SettlementServer) (err err
...
 
-		projectID, bucketName, err := SplitBucketID(bucketID)
-		if err != nil {
-			return err
-		}
-		if err := endpoint.DB.UpdateBucketBandwidthSettle(ctx, *projectID, bucketName, orderLimit.Action, order.Amount, intervalStart); err != nil {
-			// TODO: we should use the serial number in the same transaction we settle the bandwidth? that way we don't need this undo in an error case?
-			if err := endpoint.DB.UnuseSerialNumber(ctx, orderLimit.SerialNumber, orderLimit.StorageNodeId); err != nil {
-				log.Error("unable to unuse serial number", zap.Error(err))
+		if len(requests) >= endpoint.settlementBatchSize {
```
egonelbre: This would end up deadlocking with the client if it sends orders one at a time.

satellite/satellitedb/orders.go
```diff
@@ -214,3 +218,178 @@ func (db *ordersDB) UnuseSerialNumber(ctx context.Context, serialNumber storj.Se
...
+		}
+		sn, err := storj.SerialNumberFromBytes(serialNumber)
+		if err != nil {
+			return nil, errs.Wrap(err)
+		}
+		bucketMap[sn] = bucketID
+	}
+
+	// build all the bandwidth updates into one sql statement
+	var updateRollupStatement string
```
egonelbre: Is there a sql statement size limit?

satellite/satellitedb/orders.go
ethanadams: I had to create raw SQL for these because the postgres driver doesn't like multiple parametrized statements in the same exec.

satellite/orders/endpoint.go
ethanadams: Should the defer at 127 cover this?
```
	defer func() {
		if len(requests) > 0 {
			err = errs.Combine(err, endpoint.processOrders(ctx, stream, requests))
			if err != nil {
				err = formatError(err)
			}
		}
	}()
```

satellite/satellitedb/orders.go
ethanadams: I ran benchmarks with batch sizes up to 5000.
sqlite fails at 1000, but that is because of the WHERE IN clause limits.  Not the size of sql statement.
I saw no issues with postgres.

satellite/orders/endpoint.go
egonelbre: Not really the main thing is let's say the storage node implements the settlement as:

```
for _, order := range orders {
    send order
    wait for response
}
```

Over time we would expect completely new implementations of storage nodes... in which case the simplest implementation of writing that settlement is that code. However, we can require the async implementation to be the protocol.

ethanadams: so no change for now?

egonelbre: I would either:
1. fix the system such that it supports that specific implementation
2. document that this implementation requires async implementation

satellite/orders/service.go
ethanadams: Creating a new PR for this

satellite/orders/endpoint.go
ethanadams: Creating a new PR for this

satellite/satellitedb/orders.go
ethanadams: I need as array so i can join with ",".  I couldn't find a helper to do something like that

satellite/orders/endpoint.go
ethanadams: Creating a new PR for #2

PR #2764, opened 2019-08-12
satellitedb/overlay: add database for storing peer identities

aligeti:
For verifying storage node signatures we need a database for storing peer identities.

This database implements Set, Get and BatchGet. Get method supports access of one peer_identity info ordered by lastest updated time. BatchGet method supports access of group of peer_identity based on group of nodeIDs at a time.


satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	chain := identity.EncodePeerIdentity(pi)
+
+	var id []byte
+	query := `SELECT node_id FROM peer_identities WHERE serial_number = ?;`
+	err = tx.QueryRow(certs.db.Rebind(query), pi.Leaf.SerialNumber.Bytes()).Scan(&id)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			// create a new entry
+			_, err = tx.Exec(certs.db.Rebind(`INSERT INTO peer_identities ( serial_number, peer_identity, node_id, update_at ) VALUES ( ?, ?, ?, ? );`), pi.Leaf.SerialNumber.Bytes(), chain, nodeID.Bytes(), time.Now())
```
cam-a: If your objective is to insert only if the entry does not exist, you could simplify with
```
    INSERT INTO ... 
    ON CONFLICT (...)
    DO NOTHING
```

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// already public key exists, just return
+	return nil
+}
+
+// Get gets the public key based on the certificate's serial number
+func (certs *certDB) Get(ctx context.Context, nodeID storj.NodeID) (_ *identity.PeerIdentity, err error) {
+	defer mon.Task()(&ctx)(&err)
+	dbxInfo, err := certs.db.Get_PeerIdentity_By_NodeId_OrderBy_Desc_UpdateAt(ctx, dbx.PeerIdentity_NodeId(nodeID.Bytes()))
```
cam-a: dbxInfo is pretty vague. Could you call this something more descriptive like `peerID`?

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if dbxInfo == nil {
+		return nil, Error.New("unknown nodeID :%+v: %+v", nodeID.Bytes(), err)
+	}
+
+	peer, err := identity.DecodePeerIdentity(ctx, dbxInfo.PeerIdentity)
+	return peer, Error.Wrap(err)
+}
+
+// BatchGet gets the public key based on the certificate's serial number
```
cam-a: This says 'based on the serial number', but it queries by the node ID

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,112 @@
...
+func (certs *certDB) BatchGet(ctx context.Context, nodeIDs []storj.NodeID) (peers []*identity.PeerIdentity, err error) {
+	defer mon.Task()(&ctx)(&err)
+	if len(nodeIDs) == 0 {
+		return nil, nil
+	}
+	args := make([]interface{}, 0, len(nodeIDs))
```
cam-a: I think this could be `[]byte`

satellite/satellitedb/database.go
```diff
@@ -98,6 +99,11 @@ func (db *DB) DropSchema(schema string) error {
 	return nil
 }
 
+// CertDB is a getter for uplink's specific info like public key, id, etc...
```
cam-a: Storage node's; not uplink's, right?

satellite/satellitedb/certdb.go
cam-a: I'm not sure what the purpose of the serial number primary key is. It looks like we're doing all of the querying by the node ID

satellite/certdb/certdb.go
```diff
@@ -0,0 +1,21 @@
...
+package certdb
+
+import (
+	"context"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
+)
+
+// DB stores uplink public keys.
```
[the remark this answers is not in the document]
aligeti: yes. Updated the comment.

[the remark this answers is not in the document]
aligeti: yes . Updated the code comment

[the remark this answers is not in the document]
aligeti: updated the code comment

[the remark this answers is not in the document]
aligeti: TEst test 123

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,111 @@
...
+func testDatabase(ctx context.Context, t *testing.T, sndb certdb.DB) {
+	{ //testing variables
+		snID, err := testidentity.NewTestIdentity(ctx)
+		require.NoError(t, err)
+		pi := snID.PeerIdentity()
+		snIDpubbytes, err := pkcrypto.PublicKeyToPEM(pi.CA.PublicKey)
+		require.NoError(t, err)
+
+		{ // New entry
+			err := sndb.Set(ctx, snID.ID, pi)
```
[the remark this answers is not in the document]
aligeti: updated

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,111 @@
...
+			assert.NoError(t, err)
+		}
+
+		{ // Get the corresponding Public key for the serialnum
+			snpi, err := sndb.Get(ctx, snID.ID)
+			assert.NoError(t, err)
+			pubbytes, err := pkcrypto.PublicKeyToPEM(snpi.CA.PublicKey)
```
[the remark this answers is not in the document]
aligeti: updated

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,111 @@
...
+			err := sndb.Set(ctx, sn1PI.ID, sn1PI)
+			assert.NoError(t, err)
+		}
+		sn2FI, err := testidentity.NewTestIdentity(ctx)
+		require.NoError(t, err)
+		sn2PI := sn2FI.PeerIdentity()
+		sn2PIpubbytes, err := pkcrypto.PublicKeyToPEM(sn2PI.CA.PublicKey)
+		require.NoError(t, err)
+
+		{ // adding two different pubkeys for same storagnode
```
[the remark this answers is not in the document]
aligeti: updated

satellite/satellitedb/certdb.go
aligeti: @egonelbre  My understanding is, if Storagenode down the road changes its peer_identity certificate( and keep the same nodeID) then the new certification will be add as a chain to the previous ones and now the certificate serial number is different. If my understanding is, when it comes to make an entry in to the peer_identities table ( previously known as certDB) and assuming if previous entry exists, then the same row cannot be updated and intact a new entry is made and thus causing two entries for the same nodeID but different Certificate's serial number are created. If my understanding is correct, we should always maintain only one entry per node and just keep updating the nodeID's peer_id certificate if it changes, and changed peer id check is made using the certificate's serial number. Please advise. Currently implementation makes multiple entries.

satellite/satellitedb/certdb.go
aligeti: updated

aligeti: I am working on this .. needed some more clarification

satellite/satellitedb/certdb.go
aligeti: empty interface can hold any type

satellite/satellitedb/database.go
aligeti: updated

satellite/satellitedb/certdb.go
aligeti: when storagenode can get a new leaf certificate, which should be store as a new entry. ie same
			storagenode can have multiple entries.But currently for a storagenode to get a new leaf identity, it is controlled in our alpha network. So the logic handles accordingly to make an entry if same serial-num exists.

satellite/satellitedb/certdb.go
aligeti: updated the comment

satellite/satellitedb/certdb.go
bryanchriswhite: nit -- can we split this one really long line?

satellite/satellitedb/certdb.go
bryanchriswhite: maybe use a [`storj.NodeIDList`](https://github.com/storj/storj/blob/master/pkg/storj/node.go#L33) instead and then you can call [`#Bytes()`](https://github.com/storj/storj/blob/master/pkg/storj/node.go#L246).

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,114 @@
...
+	if dbxPeerID == nil {
+		return nil, Error.New("unknown nodeID :%+v: %+v", nodeID.Bytes(), err)
+	}
+
+	peer, err := identity.DecodePeerIdentity(ctx, dbxPeerID.PeerIdentity)
+	return peer, Error.Wrap(err)
+}
+
+// BatchGet gets the public key based on the certificate's serial number
+func (certs *certDB) BatchGet(ctx context.Context, nodeIDs []storj.NodeID) (peers []*identity.PeerIdentity, err error) {
```
bryanchriswhite: maybe `peerIdents` or something instead of `peers` as `peer` means something else and is used in other, unrelated parts of the code.

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,114 @@
...
+		err = errs.Combine(err, rows.Close())
+	}()
+
+	for rows.Next() {
+		r := &dbx.PeerIdentity{}
+		err := rows.Scan(&r.SerialNumber, &r.PeerIdentity, &r.NodeId, &r.UpdateAt)
+		if err != nil {
+			return peers, Error.Wrap(err)
+		}
+		peer, err := identity.DecodePeerIdentity(ctx, r.PeerIdentity)
```
bryanchriswhite: same comment about the word `peer`

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
...
+
+model peer_identity (
+    key serial_number 
+
+    index (
+        name serial_number_update_at
+        fields serial_number update_at 
+    )
+    
+    field serial_number  blob
```
bryanchriswhite: Just curious, why we care about the serial number?

satellite/satellitedb/certdb.go
aligeti: it is expecting interface{} type

satellite/satellitedb/certdb.go
aligeti: updated with peerIdent(s)

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: serial number is unique and had to use to make multiple entries for same nodeID

satellite/satellitedb/certdb.go
bryanchriswhite: I think this can still be simplified, maybe something like this?:
```
func (certs *certDB) BatchGet(ctx context.Context, nodeIDs storj.NodeIDList) ...
...
args := make([]interface{}, 0, len(nodeIDs)) // or `nodeIDs.Len()`, doesn't really matter
for _, nodeID := range nodeIDs {
  args = append(args, nodeID)
}
```

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,111 @@
...
+				require.NoError(t, err)
+				PIDs = append(PIDs, fid.PeerIdentity())
+				NIDs = append(NIDs, fid.PeerIdentity().ID)
+				err = snCerts.Set(ctx, fid.PeerIdentity().ID, fid.PeerIdentity())
+				assert.NoError(t, err)
+			}
+			pkey, err := snCerts.BatchGet(ctx, NIDs)
+			assert.NoError(t, err)
+			assert.NotNil(t, pkey)
+			assert.Equal(t, 10, len(pkey))
```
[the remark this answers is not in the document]
bryanchriswhite: Shouldn't it be simple to do though?

What about something like this?:
```
peerIdents, err := snCerts.BatchGet(ctx, NIDs) // line 110
...
for i, ident := range gotIdents {
	peerIdentBytes, err := identity.EncodePeerIdentity(PIDs[i])
	require.NoError(err)
	gotIdentBytes, err := identity.EncodePeerIdentity(ident)
	require.NoError(err)

assert.EqualValues(t, peerIdentBytes, gotIdentBytes)
}
```

satellite/satellitedb/certdb.go
bryanchriswhite: It looks the same still to me, am i looking at it wrong?

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,116 @@
...
+			assert.NoError(t, err)
+		}
+		sn2FI, err := testidentity.NewTestIdentity(ctx)
+		require.NoError(t, err)
+		sn2PI := sn2FI.PeerIdentity()
+		sn2PIpubBytes, err := pkcrypto.PublicKeyToPEM(sn2PI.CA.PublicKey)
+		require.NoError(t, err)
+
+		{ // adding two different pubkeys for same storagenode
+			{ // add a key for to storagenode ID
```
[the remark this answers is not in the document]
aligeti: hahah nope ...

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,116 @@
...
+			assert.NoError(t, err)
+			pbytes, err := pkcrypto.PublicKeyToPEM(pkey.CA.PublicKey)
+			require.NoError(t, err)
+			assert.EqualValues(t, sn2PIpubBytes, pbytes)
+		}
+		// This describes a scenario that shouldn't happen: the node ID for which
+		// we're storing the identity, is from a completely different identity.
+		// This scenario does ensure that the peer identity will be overwritten
+		// for a given node ID but maybe it's worth a comment or something that
+		// this is for testing convenience
```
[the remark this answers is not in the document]
aligeti: yep... I missed understood as I was doing the same at line 94 generating about 10 ids and adding and reading back to match the entires... I will move it to line 76.

satellite/certdb/certdb_test.go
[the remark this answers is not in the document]
aligeti: updated thnx

satellite/satellitedb/certdb.go
aligeti: updated the the Get function... at line 64  instead of 80. actually at both place needed

satellite/satellitedb/certdb.go
aligeti: updated as suggested

satellite/certdb/certdb.go
```diff
@@ -0,0 +1,21 @@
...
+package certdb
+
+import (
+	"context"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
+)
+
+// DB stores storagenode public keys.
```
bryanchriswhite: I think it would be more accurate if this said something like "DB stores storagenode peer identities."

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,122 @@
...
+		testDatabase(ctx, t, db.CertDB())
+	})
+}
+
+func testDatabase(ctx context.Context, t *testing.T, snCerts certdb.DB) {
+	{ //testing variables
+		snID, err := testidentity.NewTestIdentity(ctx)
+		require.NoError(t, err)
+		peerIdent := snID.PeerIdentity()
+		expectedpubBytes, err := pkcrypto.PublicKeyToPEM(peerIdent.CA.PublicKey)
```
bryanchriswhite: `expectedPubBytes`

satellite/certdb/certdb_test.go
```diff
@@ -0,0 +1,122 @@
...
+		sn2PI := sn2FI.PeerIdentity()
+		sn2PIpubBytes, err := pkcrypto.PublicKeyToPEM(sn2PI.CA.PublicKey)
+		require.NoError(t, err)
+
+		// This describes a scenario that shouldn't happen: the node ID for which
+		// we're storing the identity, is from a completely different identity.
+		// This scenario does ensure that the peer identity will be overwritten
+		// for a given node ID but maybe it's worth a comment or something that
+		// this is for testing convenience
+		{ // adding two different peer ids for same storagenode
```
bryanchriswhite: Mabye something like "adding two different peer identities for the same node ID" would be a bit more accurate/clear.

satellite/satellitedb/certdb.go
bryanchriswhite: Does "updated", meaning you split the line? Because it still looks like one line to me: https://github.com/storj/storj/pull/2764/files#diff-d2811cdba7bcdde3b1ebb1319da813a2R51

satellite/satellitedb/database.go
```diff
@@ -98,6 +99,11 @@ func (db *DB) DropSchema(schema string) error {
 	return nil
 }
 
+// CertDB is a getter for storagenode's specific info like public key, id, etc...
```
bryanchriswhite: What about something like "CertDB is a getter for the peer identity cache".

satellite/satellitedb/locked.go
```diff
@@ -122,6 +124,40 @@ func (m *lockedBuckets) UpdateBucket(ctx context.Context, bucket storj.Bucket) (
 	return m.db.UpdateBucket(ctx, bucket)
 }
 
+// CertDB returns database for storing uplink's public key & ID
```
bryanchriswhite: Can we make this comment consistent with the one at database.go:102?

bryanchriswhite: (that probably applies to all these methods)

satellite/satellitedb/certdb.go
```diff
@@ -0,0 +1,113 @@
...
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
+	dbx "storj.io/storj/satellite/satellitedb/dbx"
+)
+
+type certDB struct {
+	db *dbx.DB
+}
+
+func (certs *certDB) Set(ctx context.Context, nodeID storj.NodeID, peerIdent *identity.PeerIdentity) (err error) {
```
bryanchriswhite: Can we add a comment for `certDB#Set`?

bryanchriswhite:
Is it crazy to suggest renaming the package to something like `IdentDB` or `IdentityDB`? The things that we're actually storing are peer identities so I feel like it would make more sense if the name reflected that.

satellite/satellitedb/certdb.go
```diff
@@ -77,15 +77,15 @@ func (certs *certDB) Get(ctx context.Context, nodeID storj.NodeID) (_ *identity.
...
+// BatchGet gets the public key based on the certificate's nodeID
+func (certs *certDB) BatchGet(ctx context.Context, nodeIDs storj.NodeIDList) (peerIdents []*identity.PeerIdentity, err error) {
 	defer mon.Task()(&ctx)(&err)
 	if len(nodeIDs) == 0 {
 		return nil, nil
 	}
-	args := make([]interface{}, 0, len(nodeIDs))
-	for i := range nodeIDs {
-		args = append(args, nodeIDs[i].Bytes())
+	args := make([]interface{}, 0, len(nodeIDs)) // or `nodeIDs.Len()`, doesn't really matter
```
bryanchriswhite: haha - you can delete the inline comment

satellite/certdb/certdb.go
aligeti: updated

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
 	where storagenode_storage_tally.interval_end_time >= ?
 )
 
+//--- peer_identity ---//
+
+model peer_identity (
+    key serial_number
```
egonelbre: The key needs to be node_id and serial_number. The serial_number is unique per CA.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
...
+    key serial_number 
+
+    index (
+        name serial_number_update_at
+        fields serial_number update_at 
+    )
+    
+    field serial_number  blob
+    field peer_identity  blob
+    field node_id        blob
```
egonelbre: Let's reorder these as:
```
node_id
serial_number
peer_identity
```

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
...
+model peer_identity (
+    key serial_number 
+
+    index (
+        name serial_number_update_at
+        fields serial_number update_at 
+    )
+    
+    field serial_number  blob
+    field peer_identity  blob
```
egonelbre: Let's rename this `der_chain` to be more descriptive how it's encoded.

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,113 @@
...
+	}()
+
+	if peerIdent == nil {
+		return Error.New("Peer Identity cannot be nil")
+	}
+	chain := identity.EncodePeerIdentity(peerIdent)
+
+	var id []byte
+	query := `SELECT node_id FROM peer_identities WHERE serial_number = ?;`
+	err = tx.QueryRow(certs.db.Rebind(query), peerIdent.Leaf.SerialNumber.Bytes()).Scan(&id)
```
egonelbre: You need to select on both the `node_id` and `serial_number`.

satellite/satellitedb/dbx/satellitedb.dbx
bryanchriswhite: Afaik there is nothing that verifies or validates certificate serial numbers at all. They should all be random if they were generated with our identity tool but if someone changed a serial number nobody would notice as long as they re-signed the certificate.

bryanchriswhite:
> The `serial_number` is unique only per `node_id`... different nodes could share the same `serial_number`.

Under what circumstances would that happen?

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: Yeah, that's fine, but that would be a violation of the spec... so when they do that, it would be fine if we reject their signatures.

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: u meant `der_chain` or `peer_chain`

egonelbre: I think `peer_` part is already implicit in the table name... however I guess `chain` would be fine as well. Except in that case add a comment about the encoding.

egonelbre:
@bryanchriswhite CA1 and CA2 generate certificates serially, so give out as 0, 1, 2 etc...

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: Ah, sorry... the key should just be `node_id`.

bryanchriswhite:
@egonelbre afaik the only place serial numbers are being set for certificates is in pkg/peertls/templates.go, using [`newSerialNumber()`](https://github.com/storj/storj/blob/master/pkg/peertls/utils.go#L88).

egonelbre:
@bryanchriswhite that is something that the end-user can customize... e.g. they could use OpenSSL to generate the certificates.

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: in that case how can I make multiple entries for same node_id with different certificates?

egonelbre: I thought @jtolds mentioned that it would be fine to just keep the latest identity rather than all of them.

aligeti: got it..

satellite/satellitedb/identdb.go
aligeti: That might create a problem, as lets say there exists an entry for a nodeID and serial#, not a request to update the same node id, and with a different serial#. The query with latest nodeID and new serial# would return no sql rows, and when I try to `INSERT` it will cause an error as nodeID is set as `UNIQUE`. The query has to be made with nodeID and get serial number. Compare the serial number with the serial number you try to update. If they match just return, otherwise update.

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: updated the changes to update an existing entry

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: changed to peer_chain and added a comment ` // x509 ASN.1 DER content `

satellite/satellitedb/dbx/satellitedb.dbx
aligeti: updated as suggested

satellite/identdb/identdb_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"storj.io/storj/satellite/identdb"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func TestCertDB(t *testing.T) {
+	satellitedbtest.Run(t, func(t *testing.T, db satellite.DB) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+
+		testDatabase(ctx, t, db.IdentDB())
```
egonelbre: No need for the separate func

satellite/identdb/identdb_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"storj.io/storj/internal/testidentity"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pkcrypto"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/identdb"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func TestCertDB(t *testing.T) {
```
egonelbre: IdentDB

satellite/peer.go
```diff
@@ -76,6 +77,8 @@ type DB interface {
 	// DropSchema drops the schema
 	DropSchema(schema string) error
 
+	// IdentDB is a getter for the peer identity cache
```
egonelbre: Why is it called a cache?

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
...
+    key node_id
+
+    index (
+        name serial_number_update_at
+        fields serial_number update_at 
+    )
+    
+    field node_id        blob
+    field serial_number  blob
+    field peer_chain     blob      // x509 ASN.1 DER content
```
egonelbre: `chain` is sufficient

satellite/identdb/identdb.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
+)
+
+// DB stores storagenode peer identities
+type DB interface {
```
egonelbre: The package is called `identdb` and the full name ends up being `identdb.DB` which is weird.

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,127 @@
...
+	"time"
+
+	"github.com/zeebo/errs"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/storj"
+	dbx "storj.io/storj/satellite/satellitedb/dbx"
+)
+
+type certDB struct {
```
egonelbre: This doesn't match the interface naming.

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,127 @@
...
+	}()
+
+	if peerIdent == nil {
+		return Error.New("Peer Identity cannot be nil")
+	}
+	chain := identity.EncodePeerIdentity(peerIdent)
+
+	var serialNum []byte
+	query := `SELECT serial_number FROM peer_identities WHERE node_id = ?;`
+	err = tx.QueryRow(certs.db.Rebind(query), nodeID.Bytes()).Scan(&serialNum)
```
egonelbre: @zeebo is there a way to use transactions such that the generated methods still work?

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,127 @@
...
+			err = tx.Commit()
+		} else {
+			err = errs.Combine(err, tx.Rollback())
+		}
+	}()
+
+	if peerIdent == nil {
+		return Error.New("Peer Identity cannot be nil")
+	}
+	chain := identity.EncodePeerIdentity(peerIdent)
```
egonelbre: Don't encode unless you need to.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -615,6 +615,31 @@ read all (
...
+
+    index (
+        name serial_number_update_at
+        fields serial_number update_at 
+    )
+    
+    field node_id        blob
+    field serial_number  blob
+    field peer_chain     blob      // x509 ASN.1 DER content 
+    field update_at timestamp ( autoinsert, autoupdate )
```
egonelbre: `updated_at`

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,127 @@
...
+	defer mon.Task()(&ctx)(&err)
+	if len(nodeIDs) == 0 {
+		return nil, nil
+	}
+	args := make([]interface{}, 0, nodeIDs.Len())
+	for _, nodeID := range nodeIDs {
+		args = append(args, nodeID)
+	}
+
+	rows, err := certs.db.Query(certs.db.Rebind(`
```
egonelbre: Let's optimize this for postgres similarly how other implementations handle it.

satellite/satellitedb/identdb.go
zeebo: Yes, the dbx wrapped DB type has a dbx wrapped Tx type. I believe you just call Open.

satellite/satellitedb/identdb.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	defer func() {
+		err = errs.Combine(err, rows.Close())
+	}()
+
+	for rows.Next() {
+		var peerChain []byte
+		err := rows.Scan(&peerChain)
+		if err != nil {
+			return peerIdents, Error.Wrap(err)
```
egonelbre: `return nil, Error.Wrap(err)`

satellite/satellitedb/dbx/satellitedb.dbx.go
```diff
@@ -731,6 +739,13 @@ CREATE TABLE offers (
 	type INTEGER NOT NULL,
 	PRIMARY KEY ( id )
 );
+CREATE TABLE peer_identities (
```
aleitner: Also initialized on line 392?

aleitner: generated code so this comment can be ignored

satellite/identdb/identdb.go
aligeti: there is but followed existing name certdb-> identdb in Storj.io/storj/satellite/identdb

egonelbre: The previous `certdb` wasn't great either. I spent some time thinking and it seems the most appropriate place for the interface would be in overlay package. It's storing after all critical storage node information and overlay is responsible for managing that information.

Hence let's move make the fully qualified named `overlay.Identities`, so the interface name would be `Identities`.

egonelbre: PS: I'm not yet completely satisfied with the name `Identities`, but I think it should be good enough for this PR. `PeerIdentities` would be fine as well.

satellite/satellitedb/identdb.go
egonelbre: Yes, but we only store when it has changed or it doesn't exist yet.

satellite/identdb/identdb_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func TestIdentDB(t *testing.T) {
+	satellitedbtest.Run(t, func(t *testing.T, db satellite.DB) {
+		ctx := testcontext.New(t)
+		defer ctx.Cleanup()
+		snCerts := db.IdentDB()
+
+		{ //testing variables
```
navillasa: super nit- space after `//` for consistency

satellite/identdb/identdb_test.go
```diff
@@ -0,0 +1,112 @@
...
+			{ // Get the corresponding Public key for the nodeID
+				gotPeerIdent, err := snCerts.Get(ctx, snID.ID)
+				assert.NoError(t, err)
+				pubBytes, err := pkcrypto.PublicKeyToPEM(gotPeerIdent.CA.PublicKey)
+				assert.NoError(t, err)
+				assert.EqualValues(t, expectedPubBytes, pubBytes)
+			}
+		}
+
+		{ //storagenode testing variables
```
navillasa: same super nit- space after `//`

Jenkinsfile.public
```diff
@@ -91,17 +91,17 @@ pipeline {
                     }
                 }
 
-                stage('Backwards Compatibility') {
-                    environment {
-                        STORJ_NETWORK_HOST4 = '127.0.0.3'
-                        STORJ_NETWORK_HOST6 = '127.0.0.3'
-                    }
+                // stage('Backwards Compatibility') {
```
egonelbre: Why are these commented out?

egonelbre:
@simongui there should be only one peer identity stored per node id.

satellite/satellitedb/dbx/satellitedb.dbx
```diff
@@ -616,6 +616,31 @@ read all (
 	where storagenode_storage_tally.interval_end_time >= ?
 )
 
+//--- peer_identity ---//
+
+model peer_identity (
+    key node_id
+
+    field node_id        blob
+    field serial_number  blob (updatable)
```
bryanchriswhite: Can we add a comment here or somewhere else that makes it clear that the serial number is from the leaf?

satellite/overlay/peeridentities.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// PeerIdentities stores storagenode peer identities
+type PeerIdentities interface {
+	// Set adds a peer identity entry for a node
+	Set(context.Context, storj.NodeID, *identity.PeerIdentity) error
+	// Get gets peer identity
+	Get(context.Context, storj.NodeID) (*identity.PeerIdentity, error)
+	// BatchGet gets all nodes peer identities in a transaction
+	BatchGet(context.Context, storj.NodeIDList) (_ []*identity.PeerIdentity, err error)
```
kaloyan-raev: We don't need named return values here.

egonelbre: removed

satellite/satellitedb/dbx/satellitedb.dbx
egonelbre: renamed the field to `leaf_serial_number`

PR #2771, opened 2019-08-13
uplink/piecestore: handle Download errors better

egonelbre:
What: Handle errors better such that we can call close without context cancellation problems. This should also get rid of the potential double closing of download.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/uplink/piecestore/download.go
+++ b/uplink/piecestore/download.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/zeebo/errs"
 
+	"storj.io/storj/internal/errs2"
 	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/signing"
@@ -41,6 +42,9 @@ type Download struct {
 	allocationStep int64
 
 	unread ReadBuffer
+
+	closed       bool
+	closingError error
 }
 
 // Download starts a new download using the specified order limit at the specified offset and size.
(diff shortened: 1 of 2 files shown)


uplink/piecestore/download.go
```diff
@@ -100,17 +104,25 @@ func (client *Client) Download(ctx context.Context, limit *pb.OrderLimit, pieceP
...
-		// if we have an error or are pending for an error, avoid further communication
-		// however we should still finish reading the unread data.
-		if err != nil || client.unread.Errored() {
+		// if we have an error return the error
+		if err != nil {
 			return read, err
 		}
+		// if we are pending for an error, avoid further requests, but try to finish what's in unread buffer.
+		if client.unread.Errored() {
+			return read, nil
```
mobyvb: why don't you return an error in this case?

egonelbre: Because `unread` still contains valid data, that can be read until it hits the error.

mobyvb: Maybe I am just misunderstanding then. I thought if `client.unread.Errored()` is true, the error was already hit. I also thought that `Read` should return an error, even when the number of bytes read is larger than 0.

egonelbre:
```
Let's say you have this stream of data buffered, which ends in error

|------|----------| <- `unread` contains this
. . . . . . . . . . . . . . E
|------| <-- `read` contains this

Even though unread contains an error, the reading doesn't end in the error...
So we shouldn't return the error until we have read all the valid data.
```

mobyvb: Ok, that makes sense. It looks like in some cases, errors are added to `client.unread` which are never returned or added to `client.closingError`. And this block seems to be the only place where `client.unread.Errored()` is used except for calls to `client.unread.IncludeError()`. So I am assuming it is okay if some unread errors are not propagated outwards?

egonelbre:
@zeebo ideally no, however it is called concurrently... but, there's a `piecestore.LockingDownload` that wraps it and handles the locking.

uplink/piecestore/download.go
egonelbre: `unread.Read` returns the error when it reads it.

mobyvb: !!
makes sense

mobyvb:
@zeebo @egonelbre `lazyPieceReader` also has a mutex that locks Read/Close

egonelbre:
@zeebo yes.

mobyvb:
I think the `TestUploadDownloadBandwidth` test is failing for a similar reason as it was failing in #2757 - not enough time passes after the test calls `Download()` for all the storagenodes to save their orders before we calculate totals. So not all the orders are being counted and the assertion fails.

PR #2772, opened 2019-08-13
storagenode: Add more test assertions

ifraixedes:
What: Add more test cases for empty piece ID.

Why:
Because we have a reported issue from one Storage Node Operator which cannot list the unsent order limits and it looks that the issue is that there is one order limit in the DB which has an empty piece ID.

The new tests pass without any change in the implementation but it's good to have them.

The goal was trying to find out the root of the problem of the ticket [V3-2400](https://storjlabs.atlassian.net/browse/V3-2400) but I ended up also adding a bit more assertions not related to it.

Please describe the tests: Added more assertions to current tests, there is no change in the implementation so there are no new tests.
 
Please describe the performance impact: none.

### DIFF
--- a/storagenode/piecestore/endpoint_test.go
+++ b/storagenode/piecestore/endpoint_test.go
@@ -130,11 +130,6 @@ func TestUpload(t *testing.T) {
 			err:           "",
 		},
 		{ // should err with piece ID not specified
-			contentLength: 1 * memory.KiB,
-			action:        pb.PieceAction_PUT,
-			err:           "missing piece id",
-		},
-		{ // should err with piece ID not specified
 			pieceID:       storj.PieceID{},
 			contentLength: 1 * memory.KiB,
 			action:        pb.PieceAction_PUT,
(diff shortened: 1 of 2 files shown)


storagenode/piecestore/endpoint_test.go
```diff
@@ -233,6 +238,15 @@ func TestDownload(t *testing.T) {
 			pieceID: orderLimit.PieceId,
 			action:  pb.PieceAction_GET,
 		},
+		{ // should err with piece ID not specified
```
mobyvb: If you do not define `pieceID`, won't it default to `storj.PieceID{}` as the 0 value? Is there a reason to have separate cases that set `pieceID: storj.PieceID{}` and that don't set `pieceID` at all?

ifraixedes: No there isn't is about me being stupidly thinking that they would be `nil`.

I'll remove them

mobyvb: I think it is fine if you want to keep them to be safe. This kind of behavior confuses me still so I am happy with the extra case

ifraixedes: I'm not fine with them, they have duplicated test cases.
It was a mistake.

PR #2774, opened 2019-08-13
jenkins: add backwards compatibility test

JessicaGreben:
What/Why:

Have Jenkins run a backwards compatibility test on the master branch to so that we can check we don't break it.

Please describe the tests:
 - Test 1: backward compatibility test where we upload a file on the most recent release tag, then download the same file on current master.

This will make running tests slower, so we will only run this test on the master branch instead of on every branch.


Jenkinsfile.public
```diff
@@ -90,6 +90,23 @@ pipeline {
                         sh 'make test-sim'
                     }
                 }
+
+                stage('Backwards Compatibility') {
+                    when {
```
egonelbre: You can use:
```
when {
    branch 'master'
}
```

Jenkinsfile.public
```diff
@@ -90,6 +90,23 @@ pipeline {
...
+                    expression {
+                        return env.BRANCH_NAME == 'master';
+                        }
```
egonelbre: Indentation messed up.

Jenkinsfile.public
JessicaGreben: i tried that here https://github.com/storj/storj/pull/2774/commits/4cab79887a3da97ed08f2199b432d201c16cec36, but it didn't run the tests. https://build.ops.storj.io/blue/organizations/jenkins/storj/detail/PR-2774/3/pipeline

scripts/test-sim-backwards.sh
```diff
@@ -38,6 +38,8 @@ storj-sim -x --host $STORJ_NETWORK_HOST4 network test bash "$SCRIPTDIR"/test-bac
 
 make -C "$BRANCH_DIR" install-sim
 
+sed -i -e 's#/release/#/branch/#g' $STORJ_NETWORK_DIR/satellite/0/config.yaml
```
egonelbre: This replacement looks weird. I'm wondering what lines is it replacing?

JessicaGreben: its replacing anywhere that have "/release/" in the config file

Jenkinsfile.public
egonelbre: Hmm, weird. I wonder if it's handling PR-s with different branch names.

scripts/test-sim-backwards.sh
JessicaGreben:
```
$ cat /var/folders/s_/n4czty1x4tj297495m66bcsc0000gp/T/tmp.XXXXXXXXXX.NBgZoieO/satellite/0/config.yaml | grep "/release/"
console.static-dir: /private/var/folders/s_/n4czty1x4tj297495m66bcsc0000gp/T/tmp.XXXXXXXXXX.NBgZoieO/release/web/satellite
mail.template-path: /private/var/folders/s_/n4czty1x4tj297495m66bcsc0000gp/T/tmp.XXXXXXXXXX.NBgZoieO/release/web/satellite/static/emails
marketing.static-dir: /private/var/folders/s_/n4czty1x4tj297495m66bcsc0000gp/T/tmp.XXXXXXXXXX.NBgZoieO/release/web/marketing```

egonelbre: Let's add a comment that this renames the static dir paths.

Jenkinsfile.public
JessicaGreben: i don't have a preference to run it for master or all PRs.

egonelbre: Let's run it for all, until it starts causing problems.

egonelbre: One thing we could test is to run it after `make test-sim` in Integration. That would mean it'll use less concurrency and might be faster.

PR #2785, opened 2019-08-14
storagenode/piecestore: Storagenode should not block garbage collection on satellite

mobyvb:
What: 
The storagenode should return from a Retain() call as soon as it gets the bloom filter from the satellite. The bulk of the work is then done in a separate goroutine on the storagenode. This allows the satellite to send retain requests faster since we (the satellite) will not be waiting for nodes to delete pieces before continuing.

This pull request adds `RetainService` to the `piecestore` package. `RetainService` is created and started when the storagenode is started (`peer.go`). When the satellite hits the storagenode's `Retain` endpoint, a retain job is queued onto the retain service and the endpoint returns.

The retain service processes retain jobs as they are queued, allowing a configurable number of jobs to be completed concurrently (default 5). Note that this will only happen in the situation that two different satellites call `Retain` at roughly the same time.

Note - the majority of these changes are moving code from the `piecestore` to the `retain` package. Most of the code relating to repair (besides the repair service run loop and queue) is the same as it was before.

Why: https://storjlabs.atlassian.net/browse/V3-2337

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/gc/gc_test.go
```diff
@@ -92,15 +93,30 @@ func TestGarbageCollection(t *testing.T) {
...
+		for {
+			currTime := time.Now()
+			dt := currTime.Sub(startTime)
+
+			// Check that piece of the deleted object is not on the storagenode
+			pieceAccess, err = targetNode.DB.Pieces().Stat(ctx, storage.BlobRef{
+				Namespace: satellite.ID().Bytes(),
+				Key:       deletedPieceID.Bytes(),
+			})
+			if err != nil && pieceAccess == nil {
```
navillasa: if err != nil is it okay that that err gets reassignd by the Stat call on line 122?

storagenode/piecestore/endpoint.go
```diff
@@ -647,17 +648,48 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
+			endpoint.log.Error("No retain context. Stopping garbage collection.")
+		}
+
+	}(peer.ID, createdBefore, filter)
+
+	return &pb.RetainResponse{}, nil
+}
+
+// RetainPieces is called in a goroutine by Retain
+func (endpoint *Endpoint) RetainPieces(ctx context.Context, satelliteID storj.NodeID, createdBefore time.Time, filter *bloomfilter.Filter) (err error) {
```
ifraixedes: Can we make this function unexported?

storagenode/piecestore/endpoint.go
```diff
@@ -647,17 +648,48 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: It could be worth it to add some parameters to `mon.Task()`

E.g `mon.Task()(&ctx, satelliteID, ceatedBefore, filter.Size())(&err)`

storagenode/piecestore/endpoint.go
```diff
@@ -647,17 +648,48 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
 		return nil, status.Error(codes.InvalidArgument, Error.Wrap(err).Error())
 	}
 
-	numDeleted := 0
-
 	// subtract some time to leave room for clock difference between the satellite and storage node
 	createdBefore := retainReq.GetCreationDate().Add(-endpoint.config.RetainTimeBuffer)
 
+	go func(satelliteID storj.NodeID, createdBefore time.Time, filter *bloomfilter.Filter) {
```
egonelbre: Nothing is waiting for this function to finish.

storagenode/piecestore/endpoint.go
```diff
@@ -647,17 +648,48 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
+		if endpoint.retainCtx != nil {
+			select {
+			case <-endpoint.retainCtx.Done():
+				endpoint.log.Error("retain error", zap.Error(endpoint.retainCtx.Err()))
+				return
+			default:
+			}
+
+			err := endpoint.RetainPieces(endpoint.retainCtx, satelliteID, createdBefore, filter)
```
egonelbre: This should be limited on how many we run at the same time.

satellite/gc/gc_test.go
mobyvb: yeah, it's fine because this if statement basically takes place of an assertion like `assert.Error(t, err)`. If the error exists, it means garbage collection did what it was supposed to. Do you think there is a way I can make this clearer?

storagenode/piecestore/endpoint.go
```diff
@@ -647,57 +650,14 @@ func (endpoint *Endpoint) Retain(ctx context.Context, retainReq *pb.RetainReques
...
-						zap.String("satellite", peer.ID.String()),
-						zap.String("pieceID", pieceID.String()),
-						zap.Error(err))
-					return nil
-				}
-			}
-			numDeleted++
-		}
-		return nil
+	endpoint.retain.QueueRetain(RetainRequest{
```
egonelbre: Stutter in naming.

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"go.uber.org/zap"
+
+	"storj.io/storj/pkg/bloomfilter"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storagenode/pieces"
+)
+
+// RetainService queues and processes retain requests from satellites
+type RetainService struct {
```
egonelbre: Is there a reason it's not a separate package?

mobyvb: No. I will move it to another package.

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+	if s.retainStatus == RetainDisabled {
+		return nil
+	}
+
+	numDeleted := 0
+	satelliteID := req.SatelliteID
+	filter := req.Filter
+	createdBefore := req.CreatedBefore
+
+	s.log.Info("Prepared to run a Retain request.",
```
ifraixedes: I don't know if this is worth it to log as `INFO`, I would do it in `DEBUG` level. What do you think?

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		zap.Time("createdBefore", createdBefore),
+		zap.Int64("filterSize", filter.Size()),
+		zap.String("satellite", satelliteID.String()))
+
+	err = s.store.WalkSatellitePieces(ctx, satelliteID, func(access pieces.StoredPieceAccess) error {
+		// We call Gosched() when done because the GC process is expected to be long and we want to keep it at low priority,
+		// so other goroutines can continue serving requests.
+		defer runtime.Gosched()
+		// See the comment above the Retain() function for a discussion on the correctness
```
ifraixedes: Is this comment still apply?

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		// of using ModTime in place of the more precise CreationTime.
+		mTime, err := access.ModTime(ctx)
+		if err != nil {
+			s.log.Error("failed to determine mtime of blob", zap.Error(err))
```
ifraixedes: I would change this level to a `WARN`.
I've seen that SNO gets alarmed on any ERROR message, hence it would avoid having them to worry a lot if this happens, considering that the process carries on with the next one.

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		if !filter.Contains(pieceID) {
+			s.log.Debug("About to delete piece id",
+				zap.String("satellite", satelliteID.String()),
+				zap.String("pieceID", pieceID.String()),
+				zap.String("retainStatus", s.retainStatus.String()))
+
+			// if retain status is enabled, delete pieceid
+			if s.retainStatus == RetainEnabled {
+				if err = s.store.Delete(ctx, satelliteID, pieceID); err != nil {
+					s.log.Error("failed to delete piece",
```
ifraixedes: This could be considered as a `WARN` level too.

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		go func(ctx context.Context, req RetainRequest) {
+			err := s.retainPieces(ctx, req)
+			if err != nil {
+				s.log.Error("retain error", zap.Error(err))
+			}
+			s.mu.Lock()
+			delete(s.queued, req.SatelliteID)
+			s.mu.Unlock()
+
+			if len(s.queued) == 0 {
```
wthorp: If we don't protect len(s.queued) with the mutex, do we risk missing an incoming Retain() during a Wait()... it doesn't seem like that's an issue or not

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		default:
+		}
+
+		return nil
+	})
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	mon.IntVal("garbage_collection_pieces_deleted").Observe(int64(numDeleted))
+	s.log.Sugar().Debugf("Deleted %d pieces during retain. RetainStatus: %s", numDeleted, s.retainStatus.String())
```
ifraixedes: Use the normal logger `s.log.Debug...`

storagenode/piecestore/endpoint.go
mobyvb: fixed

storagenode/piecestore/retain.go
mobyvb: done

storagenode/piecestore/retain.go
mobyvb: I'm not sure. I can protect this part with the mutex as well.

storagenode/piecestore/retain.go
```diff
@@ -0,0 +1,186 @@
...
+		return
+	}
+	select {
+	case <-s.emptyTrigger:
+	case <-ctx.Done():
+	}
+}
+
+func (s *RetainService) retainPieces(ctx context.Context, req RetainRequest) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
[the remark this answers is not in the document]
mobyvb: We shouldn't have to worry about anything being queued if retain status is disabled (there is another check in `endpoint.Retain` before queue is called). If you still think I should make the change, I can.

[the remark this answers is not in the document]
mobyvb: I did update the monkit call to have the parameters you mentioned.

storagenode/piecestore/retain.go
mobyvb: Updated

storagenode/piecestore/retain.go
mobyvb: Yes. I moved the comment it refers to

storagenode/piecestore/retain.go
mobyvb: Fixed

storagenode/piecestore/retain.go
mobyvb: Done

storagenode/piecestore/retain.go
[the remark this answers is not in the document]
ifraixedes: > We shouldn't have to worry about anything being queued if retain status is disabled (there is another check in endpoint.Retain before queue is called). If you still think I should make the change, I can

I don't fully understand this comment because I think that I didn't explain myself clearly.

My worries are the following:

1. If we call `retainPieces` when enabled or debug then this function call will be shown in monkit as a function that it took some time to performs.
1. If we call `retainPieces` when disabled then this function call will be shown in monkit as a function that ti took very minimal time to performs.

If at some point we would like to analyze some data monkit data collected from several Storage Nodes and we want to see the time that they spend calling `retainPieces` we are going to have some mixed data which between the ones which are enabled/debug and disabled but we are not going to have a sway to exclude the disabled ones because the status isn't informed in the parameters.

Because I think that it's worthless to track calls when disabled, I was suggesting not to perform the call rather than adding the status parameter.

I'm expecting you change it if my point doesn't make sense considering that I'm not sure how we analyze the monkit data and the exposed case isn't real.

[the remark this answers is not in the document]
mobyvb: What I am saying is that in `piecestore.endpoint.Retain`, we do not call `retain.Queue()` if retain status is set to disabled. Which means that in production, `retainPieces` will never be called if retain status is disabled, even if we do not explicitly check for it here. But that might not always be true in the future so I can add the condition just in case.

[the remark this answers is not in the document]
ifraixedes: Ah, OK, I do understand now.
I thought that it was called due to the condition that there is just in the lines below this one.

If we actually check for it just in case, shouldn't we move this call after that conditional?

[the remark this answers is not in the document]
mobyvb: yes, I updated it

PR #2789, opened 2019-08-14
docs/design: automatic updater

aleitner:


docs/design/storage-node-automatic-updater.md
```diff
@@ -0,0 +1,95 @@
+# Storage Node Automatic Updater
+
+## Abstract
+
+Automatic Updater is a process separate from the Storage Node that automatically downloads the latest Storage Node binary to replace the current Storage Node binary.
+
+## Background
+
+As more Storage Node Operators join the network we must ensure these nodes have a mechanism to automatically keep their nodes software up to date.
+If a Storage Node Operator fails to keep their node up to date with the minimum version required by satellites they will no longer be selected for upload or download requests.s
```
kaloyan-raev: Typo at the end of this line.

docs/design/storage-node-automatic-updater.md
```diff
@@ -0,0 +1,95 @@
...
+Currently we are using Docker for automatic updates but we are migrating away from docker so we need to write out own automatic update utility for Storage Nodes.
+
+## Design
+
+* Contacts Version Server to determine if Storage Node needs to be updated.
+* Downloads latest Storage Node binary.
+* Validate binary signature.
+* Stop the Windows Service for Storage Node
```
kaloyan-raev: We are designing a platform-independent autoupdater, right? In such case, let's just say "Stop the Storage Node process". On Windows, this would indeed mean to stop the Windows Service.

docs/design/storage-node-automatic-updater.md
```diff
@@ -0,0 +1,95 @@
...
+* Replaces current Storage Node binary with latest Storage Node binary.
+* Kills current Storage Node binary process.
```
kaloyan-raev: This bullet should be removed. The process should be already stopped at this moment.

egonelbre: No, the order should as is. First, you swap the binaries and then restart the process.

kaloyan-raev: You may not be able to do this on Windows. Very often (if not every time) I cannot overwrite the binary because it is locked by the running process.

Also, if we download and extract the binary to a temp directory, then swapping the binary from the temp directory would be a quick operation and the delay will be insignificant.

egonelbre: You rename `current.exe -> old.exe` then `new.exe -> current.exe`. This usually works. Yes, you cannot delete files in use, however, you can rename them.

kaloyan-raev: Oh, I didn't know about this trick. OK, I agree with you.

egonelbre: But, I guess, since we are using a separate binary, then this wouldn't be problematic. The renaming approach might have been due to needing to update the same binary, where the updater is running.

kaloyan-raev: However, can we avoid the word "kill" here? This should be a graceful shutdown.

egonelbre: stop and start is usually used for services

egonelbre: Updated the writing significantly in multiple parts.

docs/design/storage-node-automatic-updater.md
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+  }
+}
+```
+
+When there is a newer version is available it needs to calculate whether it needs to update. To check whether rollout has reached this node it needs to calculate `hash(rollout_seed, node_id) < rollout_cursor`. This exact behavior may differ for canary nodes, which always get the latest version.
+
+* The update check must verify that it is a trusted server.
+* The update check should have a jitter to avoid a stampeding herd. See http://highscalability.com/blog/2012/4/17/youtube-strategy-adding-jitter-isnt-a-bug.html for more information.
```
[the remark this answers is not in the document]
kaloyan-raev: Often, after update storage nodes to DB migration (which may bring the offline for long minutes) or initiate some communication with satellites. If too many nodes update at the same time, this may bring a large part of the network offline, or the satellites may be overloaded with too many requests. We want to avoid this with adding a jitter.

docs/design/storage-node-automatic-updater.md
```diff
@@ -0,0 +1,143 @@
...
+* Downloaded file may be quarantined by the anti-virus or blocked by the firewall.
+
+### Updating the binaries
+
+To update the binaries we can take two approaches. 
+
+1. Rename `storagenode.exe` into `storagenode.old.<release>.exe`.
+1. Rename `storagenode.<release>.exe` into `storagenode.exe`.
+1. Restart the service using Windows API.
+1. Delete `storagenode.old.<release>.exe`.
```
mniewrzal: Maybe we should give users possibility to revert update to last working version and not delete `storagenode.old.*`?

mniewrzal: oh ok, I just read below `We won't support rolling back to the previous version.` :)

PR #2791, opened 2019-08-14
lib/uplinkc: fix flaky download test

zeebo:
The download driver code loops through the downloaded_data buffer
but doesn't ensure that it always passes a valid pointer to the
Go side. In particular, if the malloc'd memory ends against an
unmapped page, and the test passes a pointer one past the end
of the memory region, and since the Go side always dereferences
the pointer when creating a slice, it will attempt to read
unmapped memory, causing a segfault.

This bug doesn't always present. Indeed, it depends on the details
of your system's memory allocator. I validated that this could be
a cause of observed crashes on OS X by using mmap and mprotect
to do the allocations ensuring that the page directly after the
memory we use was unmapped/protected. The crash happened exactly
as seen, and was fixed by changing this condition in the while
loop.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/lib/uplinkc/testdata/object_test.c
+++ b/lib/uplinkc/testdata/object_test.c
@@ -168,7 +168,7 @@ void handle_project(ProjectRef project) {
             size_t downloaded_total = 0;
 
             size_t size_to_read = 256 + i;
-            while (true) {
+            while (downloaded_total < data_len) {
                 size_t read_size = download_read(downloader, &downloaded_data[downloaded_total], size_to_read, err);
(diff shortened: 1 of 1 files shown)


egonelbre:
@zeebo, @aleitner the reason it was trying to read past was that it needs to test whether it's reading more data than is available and whether it returns the correct error when the stream finishes.

zeebo:
Ok, we should add another check. That check should still pass a valid buffer.

PR #2796, opened 2019-08-15
storagenode/orders: add ordersDB method CleanArchive

cam-a:
What: A new method on the ordersDB to delete items in the archive that are older than a specific time.Duration

Why: Currently there is no deletion from the order_archive_ table. It will continue to grow indefinitely unless the user manually deletes items 

https://storjlabs.atlassian.net/browse/V3-1979

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/orders/db_test.go
+++ b/storagenode/orders/db_test.go
@@ -126,6 +126,15 @@ func TestDB(t *testing.T) {
 			},
 		}, archived, cmp.Comparer(pb.Equal)))
 
+		// with 1 hour ttl, archived order should not be deleted
+		n, err := db.Orders().CleanArchive(ctx, time.Hour)
+		require.NoError(t, err)
+		require.Equal(t, 0, n)
+
+		// with 1 nanosecond ttl, archived order should be deleted
+		n, err = db.Orders().CleanArchive(ctx, time.Nanosecond)
+		require.NoError(t, err)
+		require.Equal(t, 1, n)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


storagenode/orders/db_test.go
```diff
@@ -126,6 +126,15 @@ func TestDB(t *testing.T) {
 			},
 		}, archived, cmp.Comparer(pb.Equal)))
 
+		// with 1 hour ttl, archived order should not be deleted
```
VinozzZ: Could you help me understand why with 1 hour ttl, the order shouldn't be deleted?

cam-a: 1 hour ttl means we want to delete all items that are older than 1 hour. Since the item was just inserted, it hasn't reached the 1 hour limit yet

storagenode/storagenodedb/orders.go
```diff
@@ -259,3 +259,25 @@ func (db *ordersdb) ListArchived(ctx context.Context, limit int) (_ []*orders.Ar
 
 	return infos, ErrInfo.Wrap(rows.Err())
 }
+
+// CleanArchive deletes all entries older than ttl
+func (db *ordersdb) CleanArchive(ctx context.Context, ttl time.Duration) (_ int, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	deleteBefore := time.Now().UTC().Add(-1 * ttl)
```
jenlij: i think you could do .Sub rather than .Add(-1)

PR #2798, opened 2019-08-15
docs/design: Watchdog

aleitner:


docs/design/storj-node-watchdog.md
```diff
@@ -0,0 +1,34 @@
...
+We are migrating away from Docker and we need to implement a process for monitoring Storage Node health.
+
+## Design
+
+Storage Node
+* Health endpoint on storage node which responds that it is running.
+
+Watchdog Process
+* Process runs on a loop periodically checking a Storage Node's Health.
+* Must be able to kill an already existing Storage Node process.
```
littleskunk: Gracefull shutdown and if that doesn't work we should kill it.

docs/design/storj-node-watchdog.md
```diff
@@ -0,0 +1,34 @@
...
+## Implementation
+
+1) Update grpc proto to have a Health endpoint on Storage Nodes.
+2) Write function for receiving Health check requests on Storage Node.
+3) Write Health check service
+    * Loop for periodically checking Health of Storage Node
+    * If the Storage Node can't be contacted we should try restarting the Storage Node process.
+    * If the Storage Node doesn't respond for (30?) seconds we should assume the Storage Node is locked and restart the Storage Node process.
+
+## Open issues (if applicable)
```
littleskunk: I am starting my storage node with `storagenode run --config-dir` How do we make sure we copy all the additional params? If the watchdog executes `storagenode run` that will just fail.

How about panic errors? An invalid config file will prevent the storage node from starting. Later we might add one for a corrupted infoDB. Do we want to restart the storage node in this case?

In a world without kademlia it might be good to detect problems with the internet connection as well. The storage node is still running but with a restart it would contact all satellites and report that it is back online. Should the watchdog detect connection problems?

aleitner: Starting the storage node with config values:
1) We could build the watchdog process and the storage node together.
2) The automatic updater process will have the same issue when restarting the storage node process so we need to talk about best course of action.

Errors:
If the logger is not accessible we should we consider push notifications are relying on displaying the errors in the web SNO dashboard or the command line history. Is it expected that we restart the storage node until the user sees the errors and manually shuts it down?

Connection Problems: 
It is definitely worth discussing if detecting connection problems is within the scope of the watchdog process

PR #2799, opened 2019-08-15
docs/design: Rotating logger

aleitner:


docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
+# Rotating Logger for Storage Node
+
+## Abstract
+
+Storage Node needs a logger that can log to disk, rotate files, compress old files, delete really old files, and ensure a limit on the size of logs.
+
+## Background
+
+Currently log files are agglomerating and are completely unmanaged. Storage nodes need to be able to keep the log files under control to prevent logs from using an exorbitant amount of the storage node's hard drive space.
```
egonelbre: Replace fancy words `agglomerating` and `exorbitant`.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+## Design
+
+Logger Service that runs at storage node startup
+* Write files to disk when log function is called.
+* Rotate the current file being used for logging by current date and log file size.
+* Compress log files older than 3(?) days.
+* Delete compressed files older than 30(?) days.
+* Limit Log files to 10Kb(?).
+
+We can potentially use [natefinch/lumberjack](https://github.com/natefinch/lumberjack) to maintain/rotate/compress/delete logs
```
egonelbre: Add link to https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation to show how to integrate.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+Currently log files are agglomerating and are completely unmanaged. Storage nodes need to be able to keep the log files under control to prevent logs from using an exorbitant amount of the storage node's hard drive space.
+
+## Design
+
+Logger Service that runs at storage node startup
+* Write files to disk when log function is called.
+* Rotate the current file being used for logging by current date and log file size.
+* Compress log files older than 3(?) days.
+* Delete compressed files older than 30(?) days.
+* Limit Log files to 10Kb(?).
```
egonelbre: We can limit them to 1MB or 8MB. 10kb is too little.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+
+Currently log files are agglomerating and are completely unmanaged. Storage nodes need to be able to keep the log files under control to prevent logs from using an exorbitant amount of the storage node's hard drive space.
+
+## Design
+
+Logger Service that runs at storage node startup
+* Write files to disk when log function is called.
+* Rotate the current file being used for logging by current date and log file size.
+* Compress log files older than 3(?) days.
+* Delete compressed files older than 30(?) days.
```
egonelbre: Yeah, 30 sounds good.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Background
+
+Currently log files are agglomerating and are completely unmanaged. Storage nodes need to be able to keep the log files under control to prevent logs from using an exorbitant amount of the storage node's hard drive space.
+
+## Design
+
+Logger Service that runs at storage node startup
+* Write files to disk when log function is called.
+* Rotate the current file being used for logging by current date and log file size.
```
egonelbre: Is it rotating by day?

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+* Compress log files older than 3(?) days.
```
egonelbre: We can immediately compress them after rotating to a new day at least initially, since the separate compression logic would require more code.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+Storage Node needs a logger that can log to disk, rotate files, compress old files, delete really old files, and ensure a limit on the size of logs.
+
+## Background
+
+Currently log files are agglomerating and are completely unmanaged. Storage nodes need to be able to keep the log files under control to prevent logs from using an exorbitant amount of the storage node's hard drive space.
+
+## Design
+
+Logger Service that runs at storage node startup
+* Write files to disk when log function is called.
```
egonelbre: This needs to be buffered if we don't want to hit limits.

docs/design/storage-node-logger.md
```diff
@@ -0,0 +1,30 @@
...
+* Rotate the current file being used for logging by current date and log file size.
+* Compress log files older than 3(?) days.
+* Delete compressed files older than 30(?) days.
+* Limit Log files to 10Kb(?).
+
+We can potentially use [natefinch/lumberjack](https://github.com/natefinch/lumberjack) to maintain/rotate/compress/delete logs
+
+## Implementation
+
+* Create a logger service that runs on storage node startup.
```
egonelbre: If we use lumberjack we don't need a separate service.

PR #2801, opened 2019-08-15
storagenode: Don't exit with fatal error when unsent orders aren't found

ifraixedes:
What: The Storage Node shouldn't stop if some unsent orders aren't found for being archived.

Why: Because it can cause a Storage Node to stop without having a grave problem.

This is a real problem for one of our Storage Node operators, please see https://storjlabs.atlassian.net/browse/V3-2434 for more information.

I haven't been able to find the cause why some unsent orders aren't found at the moment of archiving them so I bring up this workaround.

I also applied some changes that I found while trying to find out the commented cause. Please read the commit messages for knowing more about them.

Please describe the tests: An existing test has been updated to check that some unsent order not found when archiving don't stop the rest of being archived.

Please describe the performance impact: if there is some impact on the performance, it shouldn't be that much.


storagenode/orders/sender.go
```diff
@@ -222,33 +228,47 @@ func (sender *Sender) settle(ctx context.Context, log *zap.Logger, satelliteID s
 		return OrderError.New("failed to start settlement: %v", err)
 	}
 
+	var errList errs.Group
+	errHandle := func(logErrMsg string, err error, fields ...zap.Field) {
+		log.Error(logErrMsg, append([]zap.Field{zap.Error(err)}, fields...)...)
```
navillasa: nit - I'm possibly wrong but it doesn't seem like much more lines to just have the `log.Err` calls and `errList.Add`s in the err check instead of using the `errHandle` function, but this is fine too

ifraixedes: I agree. I conserved because it already existed and it also had a couple of lines.

I'm fine to remove it and directly do what each call to this should do; it's clearer to me than having this handler `errHandle` function

ifraixedes: By the way, this suggestion was applied.

navillasa: thanks for addressing!

PR #2803, opened 2019-08-16
satellite/satellitedb: persist piece counts to/from db

bryanchriswhite:
What: Load last known piece counts from the nodes table when the gc service starts and save piece counts to the db when they're updated during gc.

Why: So that first iteration of gc bloom filters are optimal after a satellite restart. See [V3-2157](https://storjlabs.atlassian.net/browse/V3-2157)

Please describe the tests:
+ `TestOverlaycache_AllPieceCounts`:
  - creates 10 test, node db records
  - sets the piece counts of each to exponentially increasing values
  - calls `AllPieceCounts`
  - asserts that the expected and actual piece counts match
+ `TestOverlaycache_UpdatePieceCounts`:
  - creates 10 test, node db records
  - builds expected piece counts map by assigning exponentially increasing piece counts to each test node
  - calls `UpdatePieceCounts`, setting nodes' piece count fields equal to the expected piece counts map
  - asserts that the expected and actual piece counts match

 

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,48 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
+	for nodeID := range pieceCounts {
+		nodeIDs = append(nodeIDs, nodeID)
+	}
+
+	var (
+		sqlQuery string
+		args     []interface{}
+	)
+	// TODO: optimize this
+	updateSQL := "UPDATE nodes SET ( piece_count ) = ( %d ) WHERE id == ?;"
```
ethanadams: This will fail on postgres due to: "pq: cannot insert multiple commands into a prepared statement".  You'll need to explicitly pass the `id` value.

satellite/gc/piecetracker.go
```diff
@@ -20,7 +20,10 @@ type PieceTracker struct {
 	log          *zap.Logger
 	config       Config
 	creationDate time.Time
-	pieceCounts  map[storj.NodeID]int
+	// TODO: does it make sense to name this type?
```
mobyvb: I don't think the type needs to be named

satellite/gc/piecetracker.go
```diff
@@ -20,7 +20,10 @@ type PieceTracker struct {
...
+	//  (If so, where to put it that doesn't create import cycle?)
+	// TODO: should we use int or int64 consistently for piece count (db type is int64)?
```
mobyvb: let's use int64

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,161 @@
...
+		copy(nodeID[:], node.Id)
+		expectedPieceCounts[nodeID] = pieceCount
+	}
+
+	// update piece count fields on test nodes; set exponentially
+	err = overlay.UpdatePieceCounts(ctx, expectedPieceCounts)
+	require.NoError(t, err)
+
+	// expected and actual piece count maps should match
+	actualPieceCounts, err := overlay.AllPieceCounts(ctx)
```
mobyvb: should we directly query for each node rather than using `AllPieceCounts`?

satellite/satellitedb/overlaycache.go
thepaul: Ethan is right, and also this will also complain on postgres due to the lack of a `==` operator. In most (all?) SQL dialects you want a single `=` to check for equality.

thepaul: And, heh, I don't think these parentheses will work- I think postgres will interpret the first set as a multi-column update, which would be fine, but then the second set will just be plain expression-grouping, and then pg will complain because the source for a multi-column update has to be something that can have multiple columns, like a sub-select.

tl;dr this should look something like `UPDATE nodes SET piece_count = %d WHERE id = E'\\x%x';`

but i'm going to try to figure out a better way to combine all these updates, because it kinda looks like _all_ the nodes are supposed to go in here. Maybe a temp table filled via `COPY FROM`, and then join the `UPDATE` with that?

thepaul: OK I think inline `VALUES` will be best, since the temp table approach would add more db round-trips. It would look something like this:

```go
if len(pieceCounts) == 0 {
    return nil
}
sqlQuery := `UPDATE nodes SET piece_count = newvals.piece_count FROM (VALUES `
args := make([]interface{}, 0, len(pieceCounts) * 2)
for nodeID, pieceCount := range pieceCounts {
    sqlQuery += `(?, ?), `
    args = append(args, nodeID, pieceCount)
}
sqlQuery = sqlQuery[:len(sqlQuery)-2] // trim off the last comma+space
sqlQuery += `) newvals(nodeid, piece_count) WHERE nodes.id = newvals.nodeid`

_, err = cache.db.DB.ExecContext(ctx, cache.db.Rebind(sqlQuery), args...)
```

Of course `UPDATE ... FROM` won't work in sqlite3, so you'll have to switch on the database driver type 😢

satellite/satellitedb/overlaycache_test.go
bryanchriswhite: Yes, I had the same thought in bed last night 😛

satellite/satellitedb/overlaycache.go
bryanchriswhite: Thanks @thepaul! We only need to be using postgres right now anyways.

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,50 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	pieceCounts := make(map[storj.NodeID]int)
+
+	var nodeID storj.NodeID
+	for _, row := range rows {
+		// NB: assumes node IDs in nodes table are always valid.
+		copy(nodeID[:], row.Id)
```
wthorp: I read the comment, but I'm on the fence.  Isn't `NodeIDFromBytes` clearer and more robust if we change implementation?

satellite/satellitedb/overlaycache.go
thepaul: Oh, ok, that's excellent. Most satellite code has to work with both because our tests use sqlite.

satellite/satellitedb/overlaycache.go
bryanchriswhite: Ah! Nice, I forgot about that, thanks!

satellite/gc/service.go
```diff
@@ -75,8 +75,11 @@ func (service *Service) Run(ctx context.Context) (err error) {
 		return nil
 	}
 
-	// TODO retrieve piece counts from overlay (when there is a column for them)
-	lastPieceCounts := make(map[storj.NodeID]int)
+	// load last piece counts from overlay db
+	lastPieceCounts, err := service.overlay.AllPieceCounts(ctx)
+	if err != nil {
+		return Error.Wrap(err)
```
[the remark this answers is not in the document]
mobyvb: I feel like if `AllPieceCounts` fails, we have bigger issues to worry about than the service returning an error.
But as long as we log out that there was an issue, it probably wouldn't be too bad to just use the default piece count for every node in this iteration.

satellite/gc/service.go
```diff
@@ -90,14 +93,17 @@ func (service *Service) Run(ctx context.Context) (err error) {
...
+		// save piece counts in memory for next iteration
 		for id := range lastPieceCounts {
 			delete(lastPieceCounts, id)
 		}
 		for id, info := range pieceTracker.retainInfos {
 			lastPieceCounts[id] = info.Count
 		}
 
+		// save piece counts to db for next satellite restart
+		err = service.overlay.UpdatePieceCounts(ctx, lastPieceCounts)
```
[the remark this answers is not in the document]
mobyvb: We should log the error, but not fail.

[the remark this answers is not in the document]
bryanchriswhite: Ha, :sweat_smile: I guess I forgot about these errors. Oops.

satellite/gc/piecetracker.go
bryanchriswhite: as discussed we'll do this in another PR

satellite/satellitedb/overlaycache.go
ethanadams: Don't we need sqlite support for the storj-sim? Or are you saying this will get resolved in another PR?

I get this when I run storj-sim
```
satellite/0      1256FwPPzP 20:37:56.126 | ERROR	garbage collection	gc/service.go:107	error updating piece counts	{"error": "satellitedb: near \"FROM\": syntax error", "errorVerbose": "satellitedb: near \"FROM\": syntax error\n\tstorj.io/storj/satellite/satellitedb.(*overlaycache).UpdatePieceCounts:951\n\tstorj.io/storj/satellite/satellitedb.(*lockedOverlayCache).UpdatePieceCounts:940\n\tstorj.io/storj/satellite/gc.(*Service).Run.func1:105\n\tstorj.io/storj/internal/sync2.(*Cycle).Run:87\n\tstorj.io/storj/satellite/gc.(*Service).Run:84\n\tstorj.io/storj/satellite.(*Peer).Run.func12:706\n\tgolang.org/x/sync/errgroup.(*Group).Go.func1:57"}
                                         | storj.io/storj/satellite/gc.(*Service).Run.func1
                                         | 	/Users/ethan/workspace/github/storj/satellite/gc/service.go:107
                                         | storj.io/storj/internal/sync2.(*Cycle).Run
                                         | 	/Users/ethan/workspace/github/storj/internal/sync2/cycle.go:87
                                         | storj.io/storj/satellite/gc.(*Service).Run
                                         | 	/Users/ethan/workspace/github/storj/satellite/gc/service.go:84
                                         | storj.io/storj/satellite.(*Peer).Run.func12
                                         | 	/Users/ethan/workspace/github/storj/satellite/peer.go:706
                                         | golang.org/x/sync/errgroup.(*Group).Go.func1
                                         | 	/Users/ethan/go/pkg/mod/golang.org/x/sync@v0.0.0-20190423024810-112230192c58/errgroup/errgroup.go:57
```

bryanchriswhite: @ethanadams yeah, storj-sim won't be able to update the piece counts in the DB but I don't think that's a big deal. Assuming you even needed storj-sim to gc, it will still build the in-memory piece counts.

egonelbre: We don't need storj-sim updating the piece counts, but we should still ensure that all database actions work on both sqlite and postgres.

bryanchriswhite: That's fair

wthorp: In other parts of the codebase we use `switch t := db.db.Driver().(type) {` for this

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,55 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
 
+// AllPieceCounts returns a map of node IDs to piece counts from the db.
+func (cache *overlaycache) AllPieceCounts(ctx context.Context) (_ map[storj.NodeID]int, err error) {
+	defer mon.Task()(&ctx, err)
+
+	// NB: `All_Node_Id_Node_PieceCount_By_PieceCount_Not_Number` selects node
+	// ID and piece count from the nodes table where piece count is not zero.
+	rows, err := cache.db.All_Node_Id_Node_PieceCount_By_PieceCount_Not_Number(ctx)
+	if err != nil {
+		return make(map[storj.NodeID]int), Error.Wrap(err)
```
ifraixedes: I would say that the most common convention in Go is when an error is returned the other return parameters should be ignored; when it isn't respected the documentation of the function should have a comment for it.

I personally prefer that this function return nil when an error and it's the caller which instantiates a map if it requires it. However, I'm happy about leaving as it's and commenting it in the function documentation.

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,55 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
 	return getNodeStats(dbNode), Error.Wrap(tx.Commit())
 }
 
+// AllPieceCounts returns a map of node IDs to piece counts from the db.
+func (cache *overlaycache) AllPieceCounts(ctx context.Context) (_ map[storj.NodeID]int, err error) {
+	defer mon.Task()(&ctx, err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx)(&err)
```

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,55 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
+			nodeIEErrs.Add(err)
+			continue
+		}
+		pieceCounts[nodeID] = int(row.PieceCount)
+	}
+	return pieceCounts, nodeIEErrs.Err()
+}
+
+func (cache *overlaycache) UpdatePieceCounts(ctx context.Context, pieceCounts map[storj.NodeID]int) (err error) {
+	defer mon.Task()(&ctx, err)
```
ifraixedes: Same problem than before.

But for this case, my suggestion is a nit, if you don't apply it, then fix the issue as the one commented before.
```suggestion
	defer mon.Task()(&ctx, len(pieceCounts)(&err)
```

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+
+	"storj.io/storj/internal/dbutil/pgutil"
+	"storj.io/storj/internal/dbutil/pgutil/pgtest"
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	dbx "storj.io/storj/satellite/satellitedb/dbx"
```
ifraixedes:
```suggestion
	"storj.io/storj/satellite/satellitedb/dbx"
```

bryanchriswhite: The 28 other places where this package is imported, it has the `dbx` alias.

bryanchriswhite: so this actually is an alias because the package of `storj.io/storj/satellite/satellitedb/dbx` is `satellitedb`.

satellite/satellitedb/export_test.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package satellitedb
```
mobyvb: won't this file export `Overlaycache` and `LockedOverlaycache` for anything that imports `satellitedb` now? I would assume we only want to access these items inside tests and don't want to expose them for other use cases.

mobyvb: It doesn't actually look like you are using these anywhere. Do we need this file?

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+		dbAccess := _db.TestDBAccess()
+		require.NotNil(t, dbAccess)
+
+		// create test nodes in overlay db
+		testNodes := createTestNodes(ctx, 10, t, dbAccess)
+
+		// set expected piece counts
+		var err error
+		expectedPieceCounts := make(map[storj.NodeID]int)
```
mobyvb: shouldn't you just use the `testPieceCounts()` function to do this part? I don't think this is doing anything different.

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,180 @@
...
+		require.NotNil(t, dbAccess)
+
+		// create test nodes in overlay db
+		testNodes := createTestNodes(ctx, 10, t, dbAccess)
+
+		// set expected piece counts
+		expectedPieceCounts := testPieceCounts(t, testNodes)
+
+		// update piece count fields on test nodes; set exponentially
+		updateTestPieceCounts(t, dbAccess, expectedPieceCounts)
```
mobyvb: You should be calling `overlayService.UpdatePieceCounts` in this test instead of `updateTestPieceCounts` I think

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,180 @@
...
+		_, err := dbAccess.Exec(dbAccess.Rebind(sqlQuery), args...)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+// testPieceCounts builds a piece count map from the node ids of `testNodes`,
+// incrementing the piece count exponentially from one node to the next.
+func testPieceCounts(t *testing.T, testNodes []*dbx.Node) map[storj.NodeID]int {
```
mobyvb: Maybe `generateTestPieceCounts` is a better name

satellite/satellitedb/export_test.go
bryanchriswhite: `Overlaycache` and `LockedOverlaycache` would be exported but `export_test` shouldn't be included in any non-test build. You're right though, we didn't need this in the end.

satellite/satellitedb/overlaycache_test.go
ifraixedes: I see.
I checked it but I looked to the wrong package, so I thought that the package was already named `dbx`

satellite/satellitedb/overlaycache.go
```diff
@@ -902,6 +902,82 @@ func (cache *overlaycache) UpdateUptime(ctx context.Context, nodeID storj.NodeID
...
+			nodeIDErrs.Add(err)
+			continue
+		}
+		pieceCounts[nodeID] = int(row.PieceCount)
+	}
+	return pieceCounts, nodeIDErrs.Err()
+}
+
+func (cache *overlaycache) UpdatePieceCounts(ctx context.Context, pieceCounts map[storj.NodeID]int) (err error) {
+	defer mon.Task()(&ctx, &pieceCounts)(&err)
```
ifraixedes: Ideally, we should pass `pieceCounts`, however, I think that's too much to register for the call considering that the maybe quite big.

My suggestion was to pass `len(pieceCounts)`.

By the way, the first parameter `ctx` must be a pointer, but the following ones should be what they are passed to the function so in this case, it should be directly `pieceCounts`, there is no need to pass a pointer.

bryanchriswhite: Thanks for the explanation. I considered the large size and was planning on having a conversation with @benjaminsirb today about what we want to persist.

ifraixedes: @benjaminsirb will tell you the requirements from a data science point of view.

In my case, I assess the tracing data for having more introspection about how the system behaves.

For this specific case, having the length of the map give me a clue of the time that the function can take to return.

For example, if the function takes sometimes quite a few time and sometimes returns quickly, knowing the length gives a clue of the reason why; if it takes so long and the length is big, then it makes sense, but it the function takes so long but the length is small, then we would know that we may have a bottleneck in the database.

bryanchriswhite: Yea, that makes sense. I think that metric might make more sense to track in a separate monkit call as it seems the convention is to optionally pass the function's arguments to `mon.Task(...)`.

My thought about why it might be useful to track the entire piece counts is to see how they change over time. We can get an instantaneous view of what the piece counts look like at any time because it's in the DB, but I thought it might be interesting to see how this changes over time.  I wasn't sure though if 1) this is that actually useful or 2) are we already doing this somewhere else and/or should we be.

satellite/satellitedb/overlaycache.go
bryanchriswhite: discussion moved to https://github.com/storj/storj/pull/2803#discussion_r317518146

satellite/satellitedb/overlaycache.go
bryanchriswhite: I just removed the second argument from the `mon.Task` call for now and added a todo for now.

bryanchriswhite: I just removed the second argument from the `mon.Task` call and added a todo for now.

ifraixedes: I cannot say about monkit because I haven't still used to analyze the data that it collects.

In other systems/libraries that I used for tracing, I traced function parameters and return values because I wanted to have as much information of each call to the functions in order to compare their execution times and in case of an unexpected result (e.g. error) know the input which caused it.

In our code, I  have barely seen that we pass parameters to `monk.Task(..)` and I don't know if that's intentionally or unintentionally because the people were not aware of it.

> I think that metric might make more sense to track in a separate monkit call as it seems the convention is to optionally pass the function's arguments to mon.Task(...).

I'm not sure about that. As far as I understand that call is for tracking the time that the function takes and some functions execution time should vary depending on the input.

Regarding the optional parameters, I don't think that the variadic parameters of `mon.Task` is because they are optional. They need to be an empty interface variadic parameter because it's the way that it can generally work for any function call.

ifraixedes: I'm not a big fan of this `TODOs` but at the same time, I don't like to stop things of moving forward for a long time.

Because I don't know the reason why we don't pass parameters to `mon.Task` and this is actually a nit on the purpose of this branch, I'm good to approve the PR.

bryanchriswhite: I could absolutely be mistaken about this convention, and I agree that we don't seem to be doing it regularly afaik; however, this is what lead me to think that it's a thing: https://godoc.org/gopkg.in/spacemonkeygo/monkit.v2#Scope.Task.

satellite/satellitedb/overlaycache_test.go
```diff
@@ -0,0 +1,166 @@
...
+func setTestNodePieceCounts(t *testing.T, dbAccess *dbx.DB, pieceCounts map[storj.NodeID]int) {
+	for nodeID, pieceCount := range pieceCounts {
+		var args []interface{}
+		var sqlQuery string
+		args = append(args, pieceCount, nodeID)
+		switch dbAccess.Driver().(type) {
+		case *pq.Driver:
+			sqlQuery = "UPDATE nodes SET piece_count = ( ?::BIGINT ) WHERE id = ?::BYTEA;"
+		default:
+			sqlQuery = "UPDATE nodes SET piece_count = ( ? ) WHERE id = ?;"
```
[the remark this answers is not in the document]
bryanchriswhite: Unfortunately not, we needed the type casting.

PR #2805, opened 2019-08-16
satellite/console: prevent bruteforce

crawter:
What:  I have added Limiter struct. It could be used to store and manage possible attackers entities which could be differentiated by key.

possible usage:
```
maxAttempts := 3
attemptsPeriod := time.Second
banDuration := time.Second
clearPeriod := time.Second * 3

var limiter = NewLimiter(maxAttempts, attemptsPeriod, banDuration, clearPeriod)

// Limit attacker attempts for 'banDuration' if more than 'maxAttempts' fail attempts were made
// during 'attemptsPeriod' duration
if attacker, ok := limiter.Find(someKey); ok && attacker.IsBanned {
        return errs.New("you are banned")
}
if err := someStuff(); err != nil {
         limiter.Attack(someKey)
}
```
Why: For now it is possible to brute force our password in case if we know email.
https://storjlabs.atlassian.net/browse/V3-2308

Please describe the tests:
 

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+			attackers = append(attackers, attacker)
+		}
+	}
+
+	return attackers
+}
+
+// Find can be used to find an attacker by specified key
+func (limiter *Limiter) Find(key string) (Attacker, bool) {
+	attacker, ok := limiter.attackers[key]
```
rikysya: not synchronized

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+	return *attacker, ok
+}
+
+// CleanUp is used to clean all attackers whose ban is expired
+func (limiter *Limiter) CleanUp() {
+	limiter.cleanUpTicker = time.NewTicker(limiter.AttemptsPeriod)
+
+	go func() {
```
rikysya: Use `sync2.Cycle`

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Find can be used to find an attacker by specified key
+func (limiter *Limiter) Find(key string) (Attacker, bool) {
+	attacker, ok := limiter.attackers[key]
+	return *attacker, ok
+}
+
+// CleanUp is used to clean all attackers whose ban is expired
+func (limiter *Limiter) CleanUp() {
```
rikysya: Add `context`

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+	limiter.cleanUpTicker = time.NewTicker(limiter.AttemptsPeriod)
+
+	go func() {
+		for range limiter.cleanUpTicker.C {
```
rikysya: No `context` cancellation handling

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+			func() {
```
rikysya: What is this function wrapper for?

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Find can be used to find an attacker by specified key
+func (limiter *Limiter) Find(key string) (Attacker, bool) {
+	attacker, ok := limiter.attackers[key]
+	return *attacker, ok
+}
+
+// CleanUp is used to clean all attackers whose ban is expired
+func (limiter *Limiter) CleanUp() {
+	limiter.cleanUpTicker = time.NewTicker(limiter.AttemptsPeriod)
```
rikysya: `ticker` leaking?

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+	for key, limit := range limiter.attackers {
+		if time.Now().After(limit.Expire) {
+			delete(limiter.attackers, key)
+		}
+	}
+}
+
+// Close should be used when limiter is no longer needed
+func (limiter *Limiter) Close() {
+	limiter.Lock()
```
rikysya: Same `sync` primitive used for everything, iteration over `atackers` can prevent ticker from stopping.
`Close` doesn't stop `cleanup` loop

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+	// is not cleared by Clear method of Limiter
+	if attacker.IsBanned && attemptTime.After(attacker.Expire) {
+		attacker.IsBanned = false
+	}
+
+	if attacker.IsBanned {
+		return true
+	}
+
+	//attacker.expire = attemptTime.Add(banDuration)
```
rikysya: remove comment

internal/defence/limiter.go
```diff
@@ -0,0 +1,143 @@
...
+// Attack is use to add new fail attack
+func (limiter *Limiter) Attack(key string) bool {
+	limiter.Lock()
+
+	defer limiter.Unlock()
+	now := time.Now()
+
+	// Try to retrieve the
+	if _, found := limiter.attackers[key]; !found {
+		limiter.attackers[key] = &Attacker{
```
rikysya: `atackers` can possible be not initialized here,

go.mod
```diff
@@ -115,6 +115,7 @@ require (
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
 	golang.org/x/sys v0.0.0-20190730183949-1393eb018365
 	golang.org/x/text v0.3.2 // indirect
+	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
```
rikysya: why is it changed?

crawter: because i'm using rate.Limiter from "golang.org/x/time/rate" package

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// Attacker stores information about attacker entity
+type Attacker struct {
+	limiter *rate.Limiter
+	Expire  time.Time
+	// TODO: could be interface{}
+	key string
+
+	IsBanned bool
```
egonelbre: It sounds like `rate.Limiter` might have facilities to do this already.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	"time"
+
+	"golang.org/x/time/rate"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// Attacker stores information about attacker entity
+type Attacker struct {
+	limiter *rate.Limiter
```
egonelbre: Why is this a pointer?

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+		attackers:      map[string]*Attacker{},
+		Attempts:       attempts,
+		AttemptsPeriod: attemptsPeriod,
+		BanDuration:    banDuration,
+		loop:           sync2.NewCycle(clearPeriod),
+	}
+}
+
+// Attack is use to add new fail attack
+func (limiter *Limiter) Attack(key string) bool {
```
egonelbre: `Attack` sounds like a bad name for it.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return attackers
+}
+
+// Find can be used to find an attacker by specified key
+func (limiter *Limiter) Find(key string) (Attacker, bool) {
+	limiter.mu.Lock()
+
```
egonelbre: nit: unneeded whitespace

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	})
+}
+
+func (limiter *Limiter) cleanUpCallback() {
+	limiter.mu.Lock()
+
+	defer limiter.mu.Unlock()
+
+	for key, limit := range limiter.attackers {
+		if time.Now().After(limit.Expire) {
```
egonelbre: avoid calling `time.Now()` multiple times, but instead take it as an argument to `cleanup`

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+// CleanUp is used to clean all attackers whose ban is expired
+func (limiter *Limiter) CleanUp(ctx context.Context) error {
+	return limiter.loop.Run(ctx, func(ctx context.Context) error {
+		limiter.cleanUpCallback()
+		return nil
+	})
+}
+
+func (limiter *Limiter) cleanUpCallback() {
```
egonelbre: this is not a callback, call this `CleanUp`

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	limiter.mu.Lock()
+
+	defer limiter.mu.Unlock()
+
+	attacker, ok := limiter.attackers[key]
+	return *attacker, ok
+}
+
+// CleanUp is used to clean all attackers whose ban is expired
+func (limiter *Limiter) CleanUp(ctx context.Context) error {
```
egonelbre: call it `Run` to make it similar to other things that require running

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	limit := limiter.attackers[key]
+
+	return limit.attack(now, limiter.BanDuration)
+}
+
+// Banned returns the list of banned attackers
+func (limiter *Limiter) Banned() []*Attacker {
+	limiter.mu.Lock()
+
```
egonelbre: nit: unneeded whitespace

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	if isBanned {
+		attacker.IsBanned = true
+		attacker.Expire = attemptTime.Add(banDuration)
+	}
+
+	return isBanned
+}
+
+// Limiter is used to store and manage list of banned entities
+type Limiter struct {
```
egonelbre: The `banned` and `attacker` seem bad word choices. e.g. `RateLimited` would sound nicer

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	attackers map[string]*Attacker
+
+	// Attempts defines how many times attacker could perform an operation
+	Attempts int
+	// AttemptsPeriod defines period in which attempts will count. For example, 5 attempts per minute.
+	AttemptsPeriod time.Duration
+	BanDuration    time.Duration
```
egonelbre: any reason these are public?

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+		limiter.attackers[key] = &Attacker{
+			key:     key,
+			limiter: rate.NewLimiter(rate.Every(limiter.BanDuration), limiter.Attempts),
+			Expire:  now.Add(limiter.BanDuration),
+		}
+	}
+
+	limit := limiter.attackers[key]
+
+	return limit.attack(now, limiter.BanDuration)
```
egonelbre: This sounds like it might end up as a bottleneck.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"storj.io/storj/internal/testcontext"
+)
+
```
egonelbre: Also let's add an example for usage.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+		if attacker.IsBanned {
+			attackers = append(attackers, attacker)
+		}
+	}
+
+	return attackers
+}
+
+// Find can be used to find an attacker by specified key
+func (limiter *Limiter) Find(key string) (Attacker, bool) {
```
egonelbre: Returning the attacker from here sounds dangerous due to potential sharing.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestLimiter(t *testing.T) {
+	ctx := testcontext.New(t)
+	key1 := "email1@example.com"
+	key2 := "email2@example.com"
```
egonelbre: Using email as the key is a bad idea, without other checks. Also let's use a test domain for the tests.

internal/defence/limiter.go
ifraixedes: `rate.NewLimiter` returns a pointer and all the methods are defined for a pointer, so I guess that's for that despite a value would work too.

internal/defence/limiter_test.go
crawter: example of usage is added in description of PR. Should i make more descriptive readme file?

internal/defence/limiter.go
crawter: it will return struct with 2 public fields - IsBanned and Expired. Moreover i returns attacker by value, not by ref

internal/defence/limiter_test.go
egonelbre: https://blog.golang.org/examples

internal/defence/limiter_test.go
crawter: key could be anything. I could possible make a key not string but interface{}

internal/defence/limiter.go
egonelbre: The type contains pointers, hence you are sharing them.

internal/defence/limiter.go
crawter: i tried to make this struct private, but linter return warnings:
`exported method Banned returns unexported type []*defence.attacker, which can be annoying to use`

internal/defence/limiter_test.go
egonelbre: Using string is completely fine, it helps with memory usage.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	for key, limit := range limiter.attackers {
+		if time.Now().After(limit.Expire) {
+			delete(limiter.attackers, key)
+		}
+	}
+}
+
+// Close should be used when limiter is no longer needed
+func (limiter *Limiter) Close() {
+	limiter.mu.Lock()
```
rikysya: Why do locking needed here?

internal/defence/limiter.go
egonelbre: Yeah, the issue is that it probably shouldn't be returned.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// Attacker stores information about attacker entity
+type Attacker struct {
```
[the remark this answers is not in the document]
crawter: i tried to make this struct private, but linter return warnings:
`exported method Banned returns unexported type []*defence.attacker, which can be annoying to use`

internal/defence/limiter.go
crawter: yeah, this fields should be private

internal/defence/limiter.go
crawter: ok, and what if i will return new attacker without this pointers?
f.e. 
```
return Attacker {
    Expire:      a.Expire,
    IsBanned: a.IsBanned,
}, true
```

egonelbre: Why do you need to return it at all?

crawter: because we could have not 'rate limited' attacker, who has a few attempt before ban

internal/defence/limiter.go
ifraixedes: `BanDuration` doesn't look like a good name for what's used.

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Attack is use to add new fail attack
+func (limiter *Limiter) Attack(key string) bool {
+	limiter.mu.Lock()
+
+	defer limiter.mu.Unlock()
+	now := time.Now()
+
+	// Try to retrieve the
+	if _, found := limiter.attackers[key]; !found {
```
ifraixedes: you can rewrite this logic for not having to access to the map twice (second access is in line 87).

For example:

```golang
	limit, found := limiter.attacker[key]
    if !found { ...}

    ...
```

internal/defence/limiter.go
egonelbre: I would expect the whole API be something like:

```
func (limiter *Limiter) Limit(ctx context.Context, now time.Time, key string) bool {}

func Example() {
    limiter := NewLimiter(...)
    if limiter.Limit(ctx, "hello") {
        return // respond with error
    }
}
```

egonelbre: And if really need to then `func (limiter *Limiter) CleanExpired(ctx context.Context, now time.Time){}`

internal/defence/limiter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"golang.org/x/time/rate"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// Attacker stores information about attacker entity
+type Attacker struct {
+	limiter *rate.Limiter
+	Expire  time.Time
```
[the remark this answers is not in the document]
crawter: to make calling code able to check when attacker will be locked out

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+			limiter: rate.NewLimiter(rate.Every(limiter.lockDuration), limiter.attempts),
+			expire:  now.Add(limiter.lockDuration),
+		}
+		limiter.attackers[key] = attacker
+	}
+
+	return attacker.limiter.AllowN(now, 1)
+}
+
+// Run is used to clean all attackers whose ban is expired
```
ifraixedes:
```suggestion
// Run is used to clean all attackers whose ban is expired.
```

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	return &Limiter{
+		attackers:      map[string]*limited{},
+		attempts:       attempts,
+		attemptsPeriod: attemptsPeriod,
+		lockDuration:   lockDuration,
+		loop:           sync2.NewCycle(clearPeriod),
+	}
+}
+
+// Limit is use to add new fail attack
```
ifraixedes:
```suggestion
// Limit is use to add new fail attack.
```

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	attempts int
+	// AttemptsPeriod defines period in which attempts will count. For example, 5 attempts per minute.
+	attemptsPeriod time.Duration
+	lockDuration   time.Duration
+
+	mu   sync.Mutex
+	loop *sync2.Cycle
+}
+
+// NewLimiter is a constructor for Limiter
```
ifraixedes:
```suggestion
// NewLimiter is a constructor for Limiter.
```

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	"storj.io/storj/internal/sync2"
+)
+
+// limited stores information about attacker entity
+type limited struct {
+	limiter *rate.Limiter
+	expire  time.Time
+}
+
+// Limiter is used to store and manage list of banned entities
```
ifraixedes:
```suggestion
// Limiter is used to store and manage a list of banned entities.
```

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// limited stores information about attacker entity
```
ifraixedes:
```suggestion
// limited stores information about attacker entity.
```

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"storj.io/storj/internal/testcontext"
+)
+
+const (
+	key1           = "email1@example.com"
```
ifraixedes: You forgot (commented by Egon) to update these emails to use a `.test` domains
https://tools.ietf.org/html/rfc2606#page-2

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		result := false
+
+		for i := 0; i <= maxAttempts; i++ {
+			result = limiter.Limit(key2)
+		}
+
+		assert.Equal(t, result, false)
+	})
+
+	limiter.Close()
```
ifraixedes: Let's move this to line 29 and call it with a `defer`

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestLimiterConcurrent(t *testing.T) {
+	ctx, cancel := context.WithCancel(testcontext.New(t))
+	var wg sync.WaitGroup
+	limiter := NewLimiter(maxAttempts, attemptsPeriod, banDuration, clearPeriod)
+
```
ifraixedes: Add a `defer limiter.Close()` here

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+	wg.Add(1)
+	go processSecondAttacker(t, &wg, limiter)
+
+	wg.Wait()
+	cancel()
+}
+
+// first attacker performs 3 operation ( with 3 max attempts ) per ~4 seconds ( with attempt duration 1 sec)
+// so he should not be locked
```
ifraixedes:
```suggestion
// so it should not be locked.
```

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		result = limiter.Limit(key1)
+
+		i++
+	}
+
+	assert.True(t, result)
+}
+
+// second attacker performs 4 operation ( with 3 max attempts ) per ~0.8 second ( with attempt duration 1 sec)
+// so he should be locked
```
ifraixedes:
```suggestion
// so it should be locked.
```

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"storj.io/storj/internal/testcontext"
+)
+
+const (
+	key1           = "email1@example.com"
+	key2           = "email2@example.com"
+	maxAttempts    = 3
+	attemptsPeriod = time.Second
+	banDuration    = time.Second
```
ifraixedes: Rename this to `lockInterval`

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+	// AttemptsPeriod defines period in which attempts will count. For example, 5 attempts per minute.
+	attemptsPeriod time.Duration
+	lockDuration   time.Duration
+
+	mu   sync.Mutex
+	loop *sync2.Cycle
+}
+
+// NewLimiter is a constructor for Limiter
+func NewLimiter(attempts int, attemptsPeriod, lockDuration, clearPeriod time.Duration) *Limiter {
```
ifraixedes: `lockDuration` stutters with its type, let's name it `lockInterval`

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Limiter is used to store and manage list of banned entities
+type Limiter struct {
+	attackers map[string]*limited
+
+	// Attempts defines how many times attacker could perform an operation
+	attempts int
+	// AttemptsPeriod defines period in which attempts will count. For example, 5 attempts per minute.
+	attemptsPeriod time.Duration
+	lockDuration   time.Duration
```
ifraixedes: `lockDuration` stutters with its type, let's name it `lockInterval`

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		assert.Equal(t, result, false)
+	})
+
+	limiter.Close()
+}
+
+func TestLimiterConcurrent(t *testing.T) {
+	ctx, cancel := context.WithCancel(testcontext.New(t))
```
zeebo: I think you need a `defer ctx.Cleanup()` on the `ctx` value returned by `testcontext.New`.

internal/defence/limiter.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Limiter is used to store and manage list of banned entities
+type Limiter struct {
+	attackers map[string]*limited
+
+	// Attempts defines how many times attacker could perform an operation
+	attempts int
+	// AttemptsPeriod defines period in which attempts will count. For example, 5 attempts per minute.
+	attemptsPeriod time.Duration
```
ifraixedes: Remove this attribute isn't used for anything with the last changes.

internal/defence/limiter_test.go
ifraixedes: Reduce this value for reducing the time that the test needs for running.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+		result = limiter.Limit(key1)
+
+		i++
+	}
+
+	assert.False(t, result)
+}
+
+// ExampleLimit shows how to use and close Limiter
```
ifraixedes:
```suggestion
// ExampleLimit shows how to use and close Limiter.
```

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		if err := limiter.Run(ctx); err != nil {
+			fmt.Print(err)
+		}
+	}()
+
+	if !limiter.Limit("someKey") {
+		return
+	}
+
+	limiter.Close()
```
ifraixedes: Move this call to just after it's initialization and defer it.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		if err := limiter.Run(ctx); err != nil {
+			fmt.Print(err)
+		}
+	}()
+
+	if !limiter.Limit("someKey") {
+		fmt.Print("you are banned")
+	}
+
+	close()
```
ifraixedes: As before, defer this call.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		result = limiter.Limit(key1)
+
+		i++
+	}
+
+	assert.False(t, result)
+}
+
+// ExampleLimit shows how to use and close Limiter
+func ExampleLimit() {
```
ifraixedes: Example functions aren't executed if they don't have `// Output:`.

Sometimes they are not meant to run, but these ones could run as part of the tests.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,161 @@
...
+		return
+	}
+
+	limiter.Close()
+}
+
+// ExampleLimit shows how to use and close (with context) Limiter
+func ExampleLimit_second() {
+	ctx, close := context.WithCancel(context.Background())
+	limiter := NewLimiter(maxAttempts, attemptsPeriod, banDuration, clearPeriod)
```
ifraixedes: `limiter` isn't closed.

ifraixedes:
I've forgotten to mention that all the `mutex` could be gone using [`sync.Map`](https://golang.org/pkg/sync/#Map) because the mutex only protects the `map` access.

egonelbre:
@ifraixedes sync.Map is intended for very specific scenarios and often can make things slower and use more memory.

ifraixedes:
> @ifraixedes sync.Map is intended for very specific scenarios and often can make things slower and use more memory.

Without profiling, we'll never know. I mentioned from the point of view of having less code and not having the manual sync.

I'm OK leaving at it's, it was just a comment.

internal/defence/limiter_test.go
ifraixedes: This has been missed

internal/defence/limiter_test.go
ifraixedes: Have you tried to reduce these durations for running the test quicker?

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+
+// limited stores information about attacker entity.
+type limited struct {
+	limiter *rate.Limiter
+	expire  time.Time
+}
+
+// Limiter is used to store and manage a list of banned entities.
+type Limiter struct {
+	attackers map[string]*limited
```
egonelbre: this still contains `attackers`

internal/defence/limiter.go
egonelbre: there still are things naming attackers/banning

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"golang.org/x/time/rate"
+
+	"storj.io/storj/internal/sync2"
+)
+
+// limited stores information about attacker entity.
+type limited struct {
+	limiter *rate.Limiter
+	expire  time.Time
```
egonelbre: `expiresAt` maybe?

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+// Limiter is used to store and manage a list of banned entities.
+type Limiter struct {
+	attackers map[string]*limited
+
+	// Attempts defines how many times attacker could perform an operation.
+	attempts     int
+	lockInterval time.Duration
+
+	mu   sync.Mutex
+	loop *sync2.Cycle
```
egonelbre: don't use a pointer

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	expire  time.Time
+}
+
+// Limiter is used to store and manage a list of banned entities.
+type Limiter struct {
+	attackers map[string]*limited
+
+	// Attempts defines how many times attacker could perform an operation.
+	attempts     int
+	lockInterval time.Duration
```
egonelbre: Is this `lockDuration` or `rateInterval`?

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (limiter *Limiter) cleanUp(ctx context.Context, cleanUpTime time.Time) error {
+	limiter.mu.Lock()
+	defer limiter.mu.Unlock()
+
+	for key, limit := range limiter.attackers {
+		select {
+		case <-ctx.Done():
+			limiter.loop.Close()
```
egonelbre: It shouldn't close the loop, this is handled by loop itself.

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	// Attempts defines how many times attacker could perform an operation.
+	attempts     int
+	lockInterval time.Duration
+
+	mu   sync.Mutex
+	loop *sync2.Cycle
+}
+
+// NewLimiter is a constructor for Limiter.
+func NewLimiter(attempts int, lockInterval, clearPeriod time.Duration) *Limiter {
```
egonelbre: `clearPeriod` but the other place is using `clean`

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,166 @@
...
+func processSecondAttacker(t *testing.T, wg *sync.WaitGroup, limiter *Limiter) {
+	defer wg.Done()
+
+	ticker := time.NewTicker(time.Second / 5)
+	i := 0
+	result := false
+
+	for range ticker.C {
+		if i == 4 {
+			ticker.Stop()
```
egonelbre: use defer near time.NewTicker rather than here, it makes it more obvious that it's properly quitting.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,166 @@
...
+// second attacker performs 4 operation ( with 3 max attempts ) per ~0.8 second ( with attempt duration 1 sec)
+// so it should be locked.
+func processSecondAttacker(t *testing.T, wg *sync.WaitGroup, limiter *Limiter) {
+	defer wg.Done()
+
+	ticker := time.NewTicker(time.Second / 5)
+	i := 0
+	result := false
+
+	for range ticker.C {
```
egonelbre:
```
for i := 0; i < 4; i++ {
    if err := sync2.Sleep(ctx, time.Second / 5); err != nil {
        return err
    }
    ...
}
```

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	return &Limiter{
+		attackers:    map[string]*limited{},
+		attempts:     attempts,
+		lockInterval: lockInterval,
+		loop:         sync2.NewCycle(clearPeriod),
+	}
+}
+
+// Limit is use to add new fail attack.
+func (limiter *Limiter) Limit(key string) bool {
```
egonelbre: If you pass in `now` as argument you don't need sleeps in your tests.

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Run is used to clean all attackers whose ban is expired.
+func (limiter *Limiter) Run(ctx context.Context) error {
+	return limiter.loop.Run(ctx, func(ctx context.Context) error {
+		return limiter.cleanUp(ctx, time.Now())
+	})
+}
+
+func (limiter *Limiter) cleanUp(ctx context.Context, cleanUpTime time.Time) error {
```
egonelbre: If you expose this, you don't need sleeps in your tests.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,166 @@
...
+	limiter := NewLimiter(maxAttempts, lockInterval, clearPeriod)
+
+	go func() {
+		err := limiter.Run(ctx)
+		assert.Error(t, err)
+		return
+	}()
+
+	wg.Add(1)
+	go processFirstAttacker(t, &wg, limiter)
```
egonelbre: For handling goroutines `errgroup.Group` is useful, when you need to handle errors.

internal/defence/limiter_test.go
```diff
@@ -0,0 +1,166 @@
...
+			ticker.Stop()
+			break
+		}
+
+		result = limiter.Limit(key1)
+
+		i++
+	}
+
+	assert.True(t, result)
```
egonelbre: This is a race, because `t` can only be used from the goroutine running the test.

internal/defence/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+func NewLimiter(attempts int, lockInterval, clearPeriod time.Duration) *Limiter {
+	return &Limiter{
+		attackers:    map[string]*limited{},
+		attempts:     attempts,
+		lockInterval: lockInterval,
+		loop:         sync2.NewCycle(clearPeriod),
+	}
+}
+
+// Limit is use to add new fail attack.
```
egonelbre: Let's add a comment how key should be chosen to avoid infinite memory growth.

crawter:
it's on hold because there are other priorities, but I will get back to it as soon as I will finish other priorities

PR #2812, opened 2019-08-16
lib/uplink: remove redis and bolt dependencies

isaachess:
What: Make it so that lib/uplink does not contain redis or boltdb in its important path. This was used by the revocation db in the TLS options.

The revocation DB was moved to its own package, so that it is no longer directly mixed with the identity package.

The TLS options were modified to take the revocation DB as an argument, instead of instantiating it within the package. This allows packages that do not need it (such as lib/uplink) to pass in nil, avoiding unnecessary imports.

Why: We are trying to reduce the size of the lib/uplinkc shared library. This saves us about 2MB.

Please describe the tests:
 - No new tests, this should all be covered by existing tests.

### DIFF
--- a/bootstrap/peer.go
+++ b/bootstrap/peer.go
@@ -19,6 +19,7 @@ import (
 	"storj.io/storj/pkg/kademlia"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/peertls/tlsopts"
+	"storj.io/storj/pkg/revocation"
 	"storj.io/storj/pkg/server"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/pkg/transport"
@@ -104,7 +105,13 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, ver
 
 	{ // setup listener and server
 		sc := config.Server
-		options, err := tlsopts.NewOptions(peer.Identity, sc.Config)
+
+		revDB, err := revocation.NewDBFromCfg(sc.Config)
+		if err != nil {
+			return nil, errs.Combine(err, peer.Close())
+		}
+
+		options, err := tlsopts.NewOptions(peer.Identity, sc.Config, revDB)
 		if err != nil {
 			return nil, errs.Combine(err, peer.Close())
 		}
--- a/bootstrap/peer.go
+++ b/bootstrap/peer.go
@@ -18,8 +18,8 @@ import (
 	"storj.io/storj/pkg/identity"
 	"storj.io/storj/pkg/kademlia"
 	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/peertls/extensions"
 	"storj.io/storj/pkg/peertls/tlsopts"
-	"storj.io/storj/pkg/revocation"
 	"storj.io/storj/pkg/server"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/pkg/transport"
@@ -85,7 +85,7 @@ type Peer struct {
 }
 
 // New creates a new Bootstrap Node.
-func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, versionInfo version.Info) (*Peer, error) {
+func New(log *zap.Logger, full *identity.FullIdentity, db DB, revDB extensions.RevocationDB, config Config, versionInfo version.Info) (*Peer, error) {
 	peer := &Peer{
 		Log:      log,
 		Identity: full,
@@ -105,11 +105,6 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, ver
 
 	{ // setup listener and server
 		sc := config.Server
-
-		revDB, err := revocation.NewDBFromCfg(sc.Config)
-		if err != nil {
-			return nil, errs.Combine(err, peer.Close())
-		}
 
 		options, err := tlsopts.NewOptions(peer.Identity, sc.Config, revDB)
 		if err != nil {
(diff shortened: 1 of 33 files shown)


bootstrap/peer.go
```diff
@@ -19,6 +19,7 @@ import (
 	"storj.io/storj/pkg/kademlia"
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/peertls/tlsopts"
+	"storj.io/storj/pkg/revocation"
```
egonelbre: Wondering whether there's a better place than the `pkg` dir.

Maybe `pkg/identity/revocation`, `pkg/peertls/revocation` or `pkg/peertls/tlsopts/revocation`.

pkg/revocation/revocation_db.go
```diff
@@ -1,44 +1,66 @@
...
+	Error = errs.Class("revocation error")
+)
+
+// DB stores the most recently seen revocation for each nodeID
 // (i.e. nodeID [CA certificate's public key hash] is the key, values is
 // the most recently seen revocation).
-type RevocationDB struct {
-	DB storage.KeyValueStore
+type DB struct {
+	KVStore storage.KeyValueStore
```
egonelbre: why is this kvstore public?

pkg/revocation/revocation_db.go
```diff
@@ -1,44 +1,66 @@
...
+		return nil, nil
+	}
+	return NewDB(cfg.RevocationDBURL)
 }
 
-// NewRevocationDB returns a new revocation database given the URL
-func NewRevocationDB(revocationDBURL string) (*RevocationDB, error) {
-	driver, source, err := dbutil.SplitConnstr(revocationDBURL)
+// NewDB returns a new revocation database given the URL
+func NewDB(dbURL string) (*DB, error) {
```
egonelbre: I would also split up bolt and redis implementations of revocation databases, so you can use one without the other.

bootstrap/peer.go
egonelbre: It seems peertls/extensions contains a bunch of additional revocation logic... so maybe

1. keep interface for database in `peertls/revocation`,
2. move revocation logic handling from `peertls/extensions` into `peertls/revocation`
3. keep implementation in `peertls/revocation/revocationkv`, and let the caller pass in the right interface

But, these can be done in separate PR-s.

satellite/peer.go
```diff
@@ -249,7 +250,13 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, config *Config, ve
 	{ // setup listener and server
 		log.Debug("Starting listener and server")
 		sc := config.Server
-		options, err := tlsopts.NewOptions(peer.Identity, sc.Config)
+
+		revDB, err := revocation.NewDBFromCfg(sc.Config)
```
egonelbre: databases are passed in as an argument to avoid logic from depending on specific implementations... revocation was one of those exceptions that we hadn't yet fixed

isaachess: Can you clarify what you mean here? Is there a change you'd like to see made?

pkg/revocation/revocation_db.go
isaachess: Because the old name `DB` was public, and it is referenced in the `testrevocation` package. This was simply a rename to avoid confusing duplicate `DB` variable names.

pkg/revocation/revocation_db.go
isaachess: Are you saying so that, if the caller already knows that they are using bolt or redis, they can avoid importing the other? Because right now the constructor functions just take a string, so it seems any function with the signature `func NewDB(dbURL string) (*DB, error)` will need to import both.

Or am I misunderstanding?

pkg/revocation/revocation_db.go
```diff
@@ -49,38 +71,38 @@ func NewRevocationDB(revocationDBURL string) (*RevocationDB, error) {
 	return db, nil
 }
 
-// newRevocationDBBolt creates a bolt-backed RevocationDB
-func newRevocationDBBolt(path string) (*RevocationDB, error) {
+// newRevocationDBBolt creates a bolt-backed DB
+func newRevocationDBBolt(path string) (*DB, error) {
```
navillasa: I noticed `NewRevocationDB` was changed to `NewDB` so why not make these `newBoltDB` and `newRedisDB`? very nit though

pkg/revocation/revocation_db.go
```diff
@@ -49,38 +71,38 @@ func NewRevocationDB(revocationDBURL string) (*RevocationDB, error) {
...
-		DB: client,
+	return &DB{
+		KVStore: client,
 	}, nil
 }
 
 // Get attempts to retrieve the most recent revocation for the given cert chain
 // (the  key used in the underlying database is the nodeID of the certificate chain).
-func (r RevocationDB) Get(ctx context.Context, chain []*x509.Certificate) (_ *extensions.Revocation, err error) {
+func (r DB) Get(ctx context.Context, chain []*x509.Certificate) (_ *extensions.Revocation, err error) {
```
navillasa: also nit - maybe change the receiver name from `r` to `db`. not a dealbreaker though.

pkg/revocation/revocation_db.go
isaachess: Good call. Fixed.

pkg/revocation/revocation_db.go
isaachess: Good call: fixed.

PR #2818, opened 2019-08-19
storagenode/storagenodedb: initialize used space table with sum over pieceinfo

littleskunk:
What: Initialize used space table with the sum over the old pieceinfo table.

Why: A full node would start with 0 used space and need several minutes to run the new used space calulation. Doing that time the node will accept data and might crash because it is running out of space. We can avoid that issue by using the old used space calucation (sum over pieceinfo) as first value.

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/infodb.go
```diff
@@ -463,6 +463,7 @@ func (db *InfoDB) Migration() *migrate.Migration {
 						satellite_id BLOB
 					)`,
 					`CREATE UNIQUE INDEX idx_piece_space_used_satellite_id ON piece_space_used(satellite_id)`,
+					`INSERT INTO piece_space_used (total) select ifnull(sum(piece_size), 0) from pieceinfo_`,
```
JessicaGreben: wow this is an easy change! yay good idea to use the current size.

JessicaGreben: i remember doing this before where I inserted a record during the migration, but it made tests fail. Are you experiencing that?

PR #2821, opened 2019-08-19
storagenode/orders: add archive cleanup to orders service

cam-a:
What: This PR introduces functionality for routine deletion of archived orders. 

- The user may specify an interval at which to run archive cleanup and a TTL for archived items. During each cleanup, all items that have reached the TTL are deleted

- This archive cleanup job is combined with the order sender into a new combined `orders service`

Why: Currently there is no method of deletion from the orders archive. If a storage node operator wishes to delete items, they must do it manually. 

https://storjlabs.atlassian.net/browse/V3-1979

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/orders/orders_test.go
```diff
@@ -22,17 +22,64 @@ import (
...
+
+		service.Sender.TriggerWait()
+
+		// let archived order age for 100 milliseconds
+		time.Sleep(100 * time.Millisecond)
+
+		err = planet.Uplinks[0].UploadWithConfig(ctx, planet.Satellites[0], &redundancy, "testbucket", "test/path", expectedData)
+		require.NoError(t, err)
+
+		service.Sender.TriggerWait()
```
isaachess: Do you need to sleep again here to prevent flappiness? In general sleeps in tests make me nervous, but I want to make sure we reduce the risk of a flappy test as much as possible.

satellite/orders/orders_test.go
```diff
@@ -22,17 +22,64 @@ import (
...
+
+		oldOrder, err := node.DB.Orders().ListArchived(ctx, 1000)
+		require.NoError(t, err)
+		require.Len(t, oldOrder, 2)
+
+		// trigger cleanup to delete archived orders older than 100 milliseconds
+		service.Cleanup.TriggerWait()
```
isaachess: Are we relying on this one not being deleted simply because we haven't waited 100 milliseconds yet? If so, this seems a high risk of being flappy, and I think we need to redesign this test somehow. Maybe a way to manually trigger orders being ready to archive.

satellite/orders/orders_test.go
```diff
@@ -22,17 +22,64 @@ import (
...
+
+		// assert old order was deleted and new order remains
+		newOrder, err := node.DB.Orders().ListArchived(ctx, 1000)
+		require.NoError(t, err)
+		require.Len(t, newOrder, 1)
+		require.True(t, newOrder[0].ArchivedAt.After(oldOrder[0].ArchivedAt))
```
isaachess: Is the point of this check simply to confirm we ahve the new order, and not the old one? If so, is there some type of ID we can compare instead? This only implies what we actually want to know.

satellite/orders/orders_test.go
cam-a: I also was nervous about adding sleep, but you can't manually set an order's `archived_at`: it's set to time.Now() inside the Archive function. So, I'm not sure how else to test that deletion based on the time of archival works without sleeping for some period

storagenode/orders/service.go
```diff
@@ -74,99 +74,132 @@ type DB interface {
...
 }
 
-// SenderConfig defines configuration for sending orders.
-type SenderConfig struct {
-	Interval time.Duration `help:"duration between sending" default:"1h0m0s"`
-	Timeout  time.Duration `help:"timeout for sending" default:"1h0m0s"`
+// Config defines configuration for sending orders.
+type Config struct {
+	SenderInterval  time.Duration `help:"duration between sending" default:"1h0m0s"`
+	CleanupInterval time.Duration `help:"duration between archive cleanups. Each cleanup deletes all orders archived before the previous iteration, leaving the last period available for reference" default:"744h0m0s"`
```
isaachess: After looking at the code, I do think we should add another config interval, that triggers how often we should run the cleanup. I would name them something like this:

`CleanupInterval` <-- this controls how _often_ clean-up is triggered
`CleanupTTL` <-- this controls which orders are deleted

Alternatively, I would be fine if the _interval_ was fixed, like it runs once when the process starts and then once per day. Each time it runs it deletes orders older than `CleanupTTL`.

cam-a: I was thinking this too. A fixed interval of once per day sounds good to me

satellite/orders/orders_test.go
thepaul: It's usually better to make a way to pass the desired time into the function instead of doing time.Now(), and if that also avoids a sleep call in a test, that makes it even better!

storagenode/storagenodedb/orders.go
```diff
@@ -302,11 +302,11 @@ func (db *ordersdb) ListArchived(ctx context.Context, limit int) (_ []*orders.Ar
...
 }
 
-// CleanArchive deletes all entries older than ttl
-func (db *ordersdb) CleanArchive(ctx context.Context, ttl time.Duration) (_ int, err error) {
+// CleanArchive deletes all entries older than interval
+func (db *ordersdb) CleanArchive(ctx context.Context, interval time.Duration) (_ int, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	deleteBefore := time.Now().UTC().Add(-1 * ttl)
+	deleteBefore := time.Now().UTC().Add(-1 * interval)
```
isaachess: I would change this var back to `ttl`, I think that is slightly more clear.

storagenode/orders/service_test.go
```diff
@@ -0,0 +1,86 @@
...
+		require.NoError(t, err)
+
+		// trigger cleanup
+		service.Cleanup.TriggerWait()
+
+		archived, err := node.DB.Orders().ListArchived(ctx, 10)
+		require.NoError(t, err)
+		require.Len(t, archived, 1)
+		require.Equal(t, archived[0].Limit.SerialNumber, serialNumber1)
+	})
```
isaachess: Nice test!

storagenode/peer.go
```diff
@@ -297,12 +297,12 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, revocationDB exten
 		}
 		pb.RegisterPiecestoreServer(peer.Server.GRPC(), peer.Storage2.Endpoint)
 
-		peer.Storage2.Sender = orders.NewSender(
-			log.Named("piecestore:orderssender"),
+		peer.Storage2.Orders = orders.NewService(
+			log.Named("piecestore:orders"),
```
egonelbre: I think we can call it just `"orders"`

storagenode/orders/service.go
```diff
@@ -71,102 +71,135 @@ type DB interface {
...
 
-// SenderConfig defines configuration for sending orders.
-type SenderConfig struct {
-	Interval time.Duration `help:"duration between sending" default:"1h0m0s"`
-	Timeout  time.Duration `help:"timeout for sending" default:"1h0m0s"`
+// Config defines configuration for sending orders.
+type Config struct {
+	SenderInterval time.Duration `help:"duration between sending" default:"1h0m0s"`
+	ArchiveTTL     time.Duration `help:"length of time to archive orders before deletion" default:"1080h0m0s"` // 45 days
+	Timeout        time.Duration `help:"timeout for sending" default:"1h0m0s"`
```
egonelbre: `SenderTimeout`

storagenode/orders/service.go
```diff
@@ -305,7 +338,7 @@ func (sender *Sender) settle(ctx context.Context, log *zap.Logger, satelliteID s
 }
 
 // Close stops the sending service.
-func (sender *Sender) Close() error {
-	sender.Loop.Close()
+func (service *Service) Close() error {
+	service.Sender.Close()
```
egonelbre: This also needs to close cleanup

storagenode/orders/service.go
```diff
@@ -71,102 +71,135 @@ type DB interface {
...
+	return &Service{
 		log:       log,
 		transport: transport,
 		orders:    orders,
 		config:    config,
 		trust:     trust,
 
-		Loop: *sync2.NewCycle(config.Interval),
+		Sender:  *sync2.NewCycle(config.SenderInterval),
+		Cleanup: *sync2.NewCycle(24 * time.Hour),
```
egonelbre: Add this as a config parameter

storagenode/orders/service.go
```diff
@@ -71,102 +71,135 @@ type DB interface {
...
+
+	var group errgroup.Group
+	service.Sender.Start(ctx, &group, func(ctx context.Context) error {
+		err := service.sendOrders(ctx)
+		if err != nil {
+			service.log.Error("error sending orders: ", zap.Error(err))
+		}
+		return nil
+	})
+	service.Cleanup.Start(ctx, &group, func(ctx context.Context) error {
```
egonelbre: You could pass `service.cleanArchive` as argument and handle error logging in `cleanArchive`. Same for `sendOrders`

storagenode/orders/service.go
```diff
@@ -71,102 +71,125 @@ type DB interface {
...
 		}
 
 		_ = group.Wait() // doesn't return errors
 	} else {
-		sender.log.Debug("no orders to send")
+		service.log.Debug("no orders to send")
 	}
 
 	close(requests)
 	return batchGroup.Wait()
```
egonelbre: Ensure that this func doesn't return an error and log it instead, otherwise the error will shutdown the node.

PR #2824, opened 2019-08-20
pkg/kademlia: clearer error message

egonelbre:
What: "info requested from untrusted peer" message makes it look like the storage node is untrusted. Use "untrusted peer", which suits both the server and the client.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

pkg/kademlia/endpoint.go
```diff
@@ -106,7 +106,7 @@ func (endpoint *Endpoint) RequestInfo(ctx context.Context, req *pb.InfoRequest)
 
 	if self.Type == pb.NodeType_STORAGE {
 		if endpoint.trust == nil {
-			return nil, status.Error(codes.Internal, "missing trust verifier")
+			return nil, status.Error(codes.Internal, "missing trust")
```
ifraixedes: I would say to add the log Error message here, but that work isn't part of this branch and anyway this package is going away soon.

egonelbre: Yeah, I think this check shouldn't be even needed and the constructor should panic/fail.

PR #2825, opened 2019-08-20
pkg/revocation: ensure we close revocation databases

egonelbre:
What: 
* Ensure that we close all databases that we open.
* Proper initialization order for satellite database.
* Clarify namings due to larger contexts.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

cmd/identity/main.go
```diff
@@ -186,12 +186,15 @@ func cmdAuthorize(cmd *cobra.Command, args []string) error {
...
+	revocationDB, err := revocation.NewDBFromCfg(config.Signer.TLS)
 	if err != nil {
 		return errs.New("Error creating revocation database: %+v", err)
 	}
+	defer func() {
+		err = errs.Combine(err, revocationDB.Close())
+	}()
 
-	signedChainBytes, err := config.Signer.Sign(ctx, ident, authToken, revDB)
+	signedChainBytes, err := config.Signer.Sign(ctx, ident, authToken, revocationDB)
```
bryanchriswhite: I think we could pass nil here. This would only be relevant if the certificates server needed to issue a revocation but I don't think that will ever happen because it's grpc identity is disposable.

egonelbre: Let's do a separate PR for that.

PR #2826, opened 2019-08-20
satellite/nodestats: use proper error codes

egonelbre:
What: Endpoints should use `status.Error` for errors.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/nodestats/endpoint.go
```diff
@@ -45,12 +47,11 @@ func (e *Endpoint) GetStats(ctx context.Context, req *pb.GetStatsRequest) (_ *pb
...
 	peer, err := identity.PeerIdentityFromContext(ctx)
 	if err != nil {
-		return nil, NodeStatsEndpointErr.Wrap(err)
+		return nil, status.Error(codes.Unauthenticated, err.Error())
 	}
-
 	node, err := e.overlay.Get(ctx, peer.ID)
 	if err != nil {
-		return nil, NodeStatsEndpointErr.Wrap(err)
+		return nil, status.Error(codes.Unauthenticated, err.Error())
```
ifraixedes: This error could be originated because of an error with the database (connection failure, etc.), in such case, the error code should be `Internal`, for the rest of the cases this code is good.

satellite/nodestats/endpoint.go
```diff
@@ -85,17 +86,16 @@ func (e *Endpoint) DailyStorageUsage(ctx context.Context, req *pb.DailyStorageUs
...
```
ifraixedes: Same comment than above.

satellite/nodestats/endpoint.go
egonelbre: Fixed. Also realized that the error should be `PermissionDenied` as well.

satellite/nodestats/endpoint.go
```diff
@@ -45,12 +47,14 @@ func (e *Endpoint) GetStats(ctx context.Context, req *pb.GetStatsRequest) (_ *pb
...
 	}
-
 	node, err := e.overlay.Get(ctx, peer.ID)
 	if err != nil {
-		return nil, NodeStatsEndpointErr.Wrap(err)
+		if overlay.ErrNodeNotFound.Has(err) {
+			return nil, status.Error(codes.PermissionDenied, err.Error())
+		}
+		return nil, status.Error(codes.Internal, err.Error())
```
ifraixedes: On Internal errors, we want to log for getting as much information as we can on what was wrong.

egonelbre: Good point.

satellite/nodestats/endpoint.go
```diff
@@ -17,8 +19,8 @@ import (
 )
 
 var (
-	// NodeStatsEndpointErr is endpoint error class
-	NodeStatsEndpointErr = errs.Class("node stats endpoint error")
+	// Error is endpoint error class
+	Error = errs.Class("nodestats endpoint error")
```
ifraixedes: It isn't used anymore, shouldn't we remove it?

egonelbre: fixed

PR #2828, opened 2019-08-20
storagenode/retain: fix concurrency issues

egonelbre:
What: 

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/storagenode/retain/retain.go
+++ b/storagenode/retain/retain.go
@@ -27,9 +27,9 @@ var (
 
 // Config defines parameters for the retain service.
 type Config struct {
-	RetainTimeBuffer    time.Duration `help:"allows for small differences in the satellite and storagenode clocks" default:"1h0m0s"`
-	RetainStatus        Status        `help:"allows configuration to enable, disable, or test retain requests from the satellite. Options: (disabled/enabled/debug)" default:"disabled"`
-	MaxConcurrentRetain int           `help:"how many concurrent retain requests can be processed at the same time." default:"5"`
+	MaxTimeSkew time.Duration `help:"allows for small differences in the satellite and storagenode clocks" default:"1h0m0s"`
+	Status      Status        `help:"allows configuration to enable, disable, or test retain requests from the satellite. Options: (disabled/enabled/debug)" default:"disabled"`
+	Concurrency int           `help:"how many concurrent retain requests can be processed at the same time." default:"5"`
 }
 
 // Request contains all the info necessary to process a retain request.
@@ -61,13 +61,13 @@ func (v *Status) Set(s string) error {
 	case "debug":
 		*v = Debug
 	default:
-		return Error.New("invalid RetainStatus %q", s)
+		return Error.New("invalid status %q", s)
 	}
 	return nil
 }
 
 // Type implements pflag.Value.
-func (*Status) Type() string { return "storj.RetainStatus" }
+func (*Status) Type() string { return "storj.Status" }
 
 // String implements pflag.Value.
 func (v *Status) String() string {
@@ -105,7 +105,7 @@ func NewService(log *zap.Logger, store *pieces.Store, config Config) *Service {
 		config:       config,
 		queued:       make(map[storj.NodeID]Request),
 		reqChan:      make(chan Request),
-		semaphore:    make(chan struct{}, config.MaxConcurrentRetain),
+		semaphore:    make(chan struct{}, config.Concurrency),
 		emptyTrigger: make(chan struct{}),
 		store:        store,
 	}
@@ -119,7 +119,7 @@ func (s *Service) Queue(req Request) bool {
 	defer s.mu.Unlock()
 
 	// subtract some time to leave room for clock difference between the satellite and storage node
-	req.CreatedBefore = req.CreatedBefore.Add(-s.config.RetainTimeBuffer)
+	req.CreatedBefore = req.CreatedBefore.Add(-s.config.MaxTimeSkew)
 
 	// only queue retain request if we do not already have one for this satellite
 	if _, ok := s.queued[req.SatelliteID]; !ok {
@@ -186,7 +186,7 @@ func (s *Service) Wait(ctx context.Context) {
 
 // Status returns the retain status.
 func (s *Service) Status() Status {
-	return s.config.RetainStatus
+	return s.config.Status
 }
 
 // ------------------------------------------------------------------------------------------------
@@ -205,13 +205,13 @@ func (s *Service) Status() Status {
 // written to the filesystem _after_ the CreationTime. We make the assumption already that storage
 // nodes and satellites and uplinks have system clocks that are very roughly in sync (that is, they
 // are out of sync with each other by less than an hour of real time, or whatever is configured as
-// RetainTimeBuffer). So if an uplink is not lying about CreationTime and it uploads a piece that
+// MaxTimeSkew). So if an uplink is not lying about CreationTime and it uploads a piece that
 // makes it to a storagenode's disk as quickly as possible, even in the worst-synchronized-clocks
-// case we can assume that `ModTime > (CreationTime - RetainTimeBuffer)`. We also allow for storage
+// case we can assume that `ModTime > (CreationTime - MaxTimeSkew)`. We also allow for storage
 // node operators doing file system manipulations after a piece has been written. If piece files
 // are copied between volumes and their attributes are not preserved, it will be possible for their
 // modification times to be changed to something later in time. This still preserves the inequality
-// relationship mentioned above, `ModTime > (CreationTime - RetainTimeBuffer)`. We only stipulate
+// relationship mentioned above, `ModTime > (CreationTime - MaxTimeSkew)`. We only stipulate
 // that storage node operators must not artificially change blob file modification times to be in
 // the past.
 //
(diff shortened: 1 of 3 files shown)


storagenode/retain/retain.go
```diff
@@ -61,13 +62,13 @@ func (v *Status) Set(s string) error {
...
 	default:
-		return Error.New("invalid RetainStatus %q", s)
+		return Error.New("invalid status %q", s)
 	}
 	return nil
 }
 
 // Type implements pflag.Value.
-func (*Status) Type() string { return "storj.RetainStatus" }
+func (*Status) Type() string { return "storj.Status" }
```
mobyvb: I think the type string should still have "retain" somewhere in it.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
-		}
+	// Hold the lock while we spawn the workers because a concurrent Close call
+	// can race and wait for them.
+	s.cond.L.Lock()
+	defer s.cond.L.Unlock()
+
+	// Ensure Run is only ever called once. If not, there's many subtle
+	// bugs with concurrency.
+	if s.started {
+		return errs.New("service already started")
```
mobyvb: There is an `Error` class that can be used

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
+	}
+	s.started = true
+
+	// Ensure Run doesn't start after it's closed. Then we may leak some
+	// workers.
+	select {
+	case <-s.closed:
+		return errs.New("service Closed")
```
mobyvb: `Error` can be used

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
-		// exit if context has been canceled. Otherwise, block until an item can be added to the semaphore
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case s.semaphore <- struct{}{}:
-		}
+	// Hold the lock while we spawn the workers because a concurrent Close call
+	// can race and wait for them.
+	s.cond.L.Lock()
+	defer s.cond.L.Unlock()
```
mobyvb: This will lock for the entire duration of `Run` which I think will cause issues

zeebo: It doesn't. There's an `Unlock` before we `Wait` on the error group.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
-	if _, ok := s.queued[req.SatelliteID]; !ok {
-		s.queued[req.SatelliteID] = req
-		go func() { s.reqChan <- req }()
+	// subtract some time to leave room for clock difference between the satellite and storage node
+	req.CreatedBefore = req.CreatedBefore.Add(-s.config.MaxTimeSkew)
 
-		return true
-	}
+	s.queued[req.SatelliteID] = req
+	s.cond.Broadcast()
```
egonelbre: Is there a reason it's now waking up all workers?

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
+			// Grab lock to check things.
+			s.cond.L.Lock()
+			defer s.cond.L.Unlock()
+
+			for {
+				// If we have closed, exit.
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-s.closed:
```
egonelbre: Only checking s.closed would suffice, but there's no harm in checking both. The context cancellation error can be returned outside.

storagenode/retain/retain.go
zeebo: Personally, I find that `Signal` is much harder to reason about and harder to maintain correctness of. It depends on the fact that ANYONE in a `(*Cond).Wait` call is available to handle the change in state, but that's not always the case. 

I started this comment by saying "Hmm, I guess not.", but then I realized that if there's a pending `(*Service).Wait()` call, and `Signal` only wakes up that one, no worker will wake up to handle the queued request, leading to deadlock.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,209 @@ type Service struct {
...
+	s.closedOnce.Do(func() { close(s.closed) })
+	s.cond.L.Unlock()
+
+	s.cond.Broadcast()
+	_ = s.group.Wait()
+	return nil
+}
+
+// Wait blocks until the queue and working is empty. When Run exits, it empties the queue.
+func (s *Service) Wait() {
```
mobyvb: We should still call this `TestWaitUntilEmpty` since it is only used for tests @zeebo

zeebo: Ah, I was unaware of that. I'll change it back.

storagenode/retain/retain.go
```diff
@@ -205,13 +310,13 @@ func (s *Service) Status() Status {
 // written to the filesystem _after_ the CreationTime. We make the assumption already that storage
 // nodes and satellites and uplinks have system clocks that are very roughly in sync (that is, they
 // are out of sync with each other by less than an hour of real time, or whatever is configured as
-// RetainTimeBuffer). So if an uplink is not lying about CreationTime and it uploads a piece that
+// MaxTimeSkew). So if an uplink is not lying about CreationTime and it uploads a piece that
```
mobyvb: Would it be better to adjust by `MaxTimeSkew` inside `retainPieces` rather than inside `Queue`? It would probably make more sense in the context of this comment.

storagenode/retain/retain.go
mobyvb: The body is also different:

```
// TestWaitUntilEmpty blocks until the context is canceled or until the queue is empty.
func (s *Service) TestWaitUntilEmpty(ctx context.Context, pollInterval time.Duration) {
	ticker := time.NewTicker(pollInterval)
	defer ticker.Stop()

	for {
		select {
		case <-ticker.C:
			s.cond.L.Lock()
			count := len(s.queued)
			s.cond.L.Unlock()


			if count == 0 {
				return
			}
		case <-ctx.Done():
			return
		}
	}
}
```

mobyvb: And in the tests we passed in `30 * time.Millisecond` as `pollInterval`

storagenode/retain/retain.go
egonelbre: Yeah, that's why used polling, since `Wait` is only used for testing... and `Signal` can avoid spurious wake ups. But, I do agree that it's harder to reason about. Since this isn't called too often, I'm also fine with Broadcast.

storagenode/retain/retain.go
egonelbre: Also `TestWaitUntilEmpty` still needs to handle `ctx`.

zeebo: Why? When the context is canceled, Run exits, which empties the queue, which unblocks Wait.

mobyvb:
Is this ready to be turned into a normal PR instead of a draft?

storagenode/retain/retain.go
```diff
@@ -88,105 +89,210 @@ type Service struct {
 	log    *zap.Logger
 	config Config
 
-	mu     sync.Mutex
-	queued map[storj.NodeID]Request
+	cond    sync.Cond
+	queued  map[storj.NodeID]Request
+	working map[storj.NodeID]struct{}
+	group   errgroup.Group
```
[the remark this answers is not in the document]
egonelbre: no need

storagenode/retain/retain.go
```diff
@@ -88,105 +89,210 @@ type Service struct {
...
+	case <-s.closed:
+		return Error.New("service Closed")
+	default:
+	}
+
+	// Create a sub-context that we can cancel.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Start a goroutine that does most of the work of Close in the case the
```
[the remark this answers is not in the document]
egonelbre: The full sentence is "in the case the context is canceled", so "the" is seems valid.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,210 @@ type Service struct {
...
+		return false
+	default:
+	}
 
-	// only queue retain request if we do not already have one for this satellite
-	if _, ok := s.queued[req.SatelliteID]; !ok {
-		s.queued[req.SatelliteID] = req
-		go func() { s.reqChan <- req }()
+	// subtract some time to leave room for clock difference between the satellite and storage node
+	req.CreatedBefore = req.CreatedBefore.Add(-s.config.MaxTimeSkew)
```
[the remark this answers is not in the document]
egonelbre: fixed

storagenode/retain/retain_test.go
```diff
@@ -27,6 +27,8 @@ import (
 )
 
 func TestRetainPieces(t *testing.T) {
+	const pollInterval = 30 * time.Millisecond
```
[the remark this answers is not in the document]
egonelbre: fixed

storagenode/retain/retain.go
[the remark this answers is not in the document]
egonelbre: I moved "the" to the next line to make it more obvious.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
+}
+
+// finish marks the request as finished, requires mutex to be held
+func (s *Service) finish(request Request) {
+	delete(s.working, request.SatelliteID)
+}
+
+// Close causes any pending Run to exit and waits for any retain requests to
+// clean up.
+func (s *Service) Close() error {
```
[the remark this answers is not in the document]
egonelbre: In this case it wouldn't. Returning an error from `Close` ensures consistency across services/chores.

storagenode/retain/retain.go
```diff
@@ -88,105 +89,207 @@ type Service struct {
...
+	s.cond.L.Lock()
+	s.closedOnce.Do(func() { close(s.closed) })
+	s.cond.L.Unlock()
+
+	s.cond.Broadcast()
+	_ = s.group.Wait()
```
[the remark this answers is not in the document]
egonelbre: https://github.com/storj/storj/pull/2910

PR #2831, opened 2019-08-20
pkg/peertls: extension handling refactor

bryanchriswhite:
What: 
  + convert `extensions.AllHandlers` to `extensions.DefaultHandlers`
  + register configurable handlers explicitly, based on respective config struct
  
Why: To cover the nil revocation db case and clean up a bit.

Please describe the tests: `TestWithOptions_NilRevocationDB` asserts that if the revocation db is nil, extension handling does not error.
 

bootstrap/peer.go
```diff
@@ -85,7 +84,7 @@ type Peer struct {
 }
 
 // New creates a new Bootstrap Node.
-func New(log *zap.Logger, full *identity.FullIdentity, db DB, revDB extensions.RevocationDB, config Config, versionInfo version.Info) (*Peer, error) {
+func New(log *zap.Logger, full *identity.FullIdentity, db DB, config Config, versionInfo version.Info) (*Peer, error) {
```
egonelbre: This PR removes change to ensure that databases are constructed outside of peers / uplink.

bryanchriswhite: :+1:

bryanchriswhite:
> LGTM, but the PR could stand to have a little more context of what the "nil revocation db case" is. The code is clear enough but the intention is not.

Fair, this test was intended to cover the fact that the revocation db can now be nil where it couldn't before. That change was made before this.

PR #2815, opened 2019-08-17
design doc: satellite service separation

JessicaGreben:
What/Why:

Add a design doc that describes process to pull apart the Satellite into many services so that it is more scalable.

Please describe the tests:
n/a
 
Please describe the performance impact:
n/a

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?
 - [ ] Do the database access patterns make sense?

PR #2836, opened 2019-08-20
prototype: separate repair service from satellite

JessicaGreben:
What/Why:
We are currently working on a design doc to separate the satellite into many services. See that doc [here](https://github.com/storj/storj/pull/2815).  This PR serves as a prototype to walk through an example of how to do this separation. 

This PR does the following things to separate out the repair service:
1. Create a `RepairProcess` struct and move repair setup from satellite/peer.go over to satellite/repair.go.
2. Add a subcommand to the satellite binary so now this command can be executed:
`satellite run repair`
3. Update testplanet with these changes

Please describe the tests:
Add a new struct to testplanet called `SatelliteSystem` that contains all the satellite services as we separate them out and updated existing tests to use this new struct. The unit/integration tests should all still pass.
 
wip

### DIFF
--- a/internal/testplanet/satellite.go
+++ b/internal/testplanet/satellite.go
@@ -228,7 +228,9 @@ func (planet *Planet) newSatellites(count int) ([]*SatelliteSystem, error) {
 		log.Debug("id=" + peer.ID().String() + " addr=" + peer.Addr())
 
 		system := SatelliteSystem{Peer: *peer}
-		system.RepairProcess, err = planet.newRepairProcess(i, storageDir)
+		repairProcess, err := planet.newRepairProcess(i, storageDir)
+		system.Repair.Checker = repairProcess.Repair.Checker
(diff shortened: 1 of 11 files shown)


internal/testplanet/satellite.go
```diff
@@ -225,9 +225,125 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
...
+			MaxBufferMem:                  4 * memory.MiB,
+			MaxExcessRateOptimalThreshold: 0.05,
+		},
+	}
+	revDB, err := revocation.NewDBFromCfg(config.Server.Config)
+	if err != nil {
+		return &satellite.RepairProcess{}, errs.New("Error creating revocation database: %+v", err)
+	}
+
+	repairProcess, err := satellite.NewRepairProcess(log, identity,
```
egonelbre: We also need to shut down the repair process when testplanet is closed.

PR #2839, opened 2019-08-21
web/satellite: API keys paged backend

Barterio:
What: 

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/satellite/console/service.go
+++ b/satellite/console/service.go
@@ -1049,6 +1049,32 @@ func (s *Service) DeleteAPIKeys(ctx context.Context, ids []uuid.UUID) (err error
 	return nil
 }
 
+// GetAPIKeys returns paged api key list for given Project
+func (s *Service) GetAPIKeys(ctx context.Context, projectID uuid.UUID, cursor APIKeyCursor) (akp *APIKeyPage, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	auth, err := GetAuth(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = s.isProjectMember(ctx, auth.User.ID, projectID)
+	if err != nil {
+		return nil, ErrUnauthorized.Wrap(err)
+	}
+
+	if cursor.Limit > maxLimit {
+		cursor.Limit = maxLimit
+	}
+
+	akp, err = s.store.APIKeys().GetPagedByProjectID(ctx, projectID, cursor)
+	if err != nil {
+		return nil, errs.New(internalErrMsg)
+	}
+
+	return
+}
+
 // GetAPIKeysInfoByProjectID retrieves all api keys for a given project
 func (s *Service) GetAPIKeysInfoByProjectID(ctx context.Context, projectID uuid.UUID) (info []APIKeyInfo, err error) {
 	defer mon.Task()(&ctx)(&err)
--- a/satellite/satellitedb/apikeys.go
+++ b/satellite/satellitedb/apikeys.go
@@ -5,6 +5,7 @@ package satellitedb
 
 import (
 	"context"
+	"strings"
 
 	"github.com/skyrings/skyring-common/tools/uuid"
 	"github.com/zeebo/errs"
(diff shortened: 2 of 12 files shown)


satellite/console/service.go
```diff
@@ -1049,9 +1049,10 @@ func (s *Service) DeleteAPIKeys(ctx context.Context, ids []uuid.UUID) (err error
 	return nil
 }
 
-// GetAPIKeysInfoByProjectID retrieves all api keys for a given project
-func (s *Service) GetAPIKeysInfoByProjectID(ctx context.Context, projectID uuid.UUID) (info []APIKeyInfo, err error) {
+// GetAPIKeys returns paged api key list for given Project
+func (s *Service) GetAPIKeys(ctx context.Context, projectID uuid.UUID, cursor APIKeyCursor) (akp *APIKeyPage, err error) {
```
egonelbre: check for invalid cursor orders

satellite/satellitedb/apikeys.go
```diff
@@ -15,41 +16,130 @@ import (
...
 	}
 
 	var apiKeys []console.APIKeyInfo
-	var parseErr errs.Group
+	for rows.Next() {
+		ak := console.APIKeyInfo{}
+		var keyIDBytes, projectIDBytes, partnerIDBytes []uint8
+		var keyID, projectID, partnerID uuid.UUID
+
+		err = rows.Scan(&keyIDBytes, &projectIDBytes, &ak.Name, &partnerIDBytes, &ak.CreatedAt)
```
egonelbre: You can implement wrappers for types to handle the unmarshaling, simplifying this code.

satellite/console/service.go
Barterio: Can you make it more clear?

crawter: akp is hard to understand. Would `page` be better?

egonelbre: What I meant is that what if somebody specifies `cursor.OrderDirection = -2`. It seems that `sanitizeOrderDirectionName` defaults to "ASC", but I think the right approach probably is to give an error to the caller.

Although now thinking of it, I might have missed the check - maybe graphql does it already. Alternatively, maybe it's possible to specify what are the allowed values for graphql.

Note this can be done in a separate PR, since it seems there are other fields that might have similar issues.

Barterio: I think that we can move it to separate PR

satellite/satellitedb/apikeys.go
Barterio: I looked through the project and didn't  find any wrapping, can you explain more what do mean by this?

Barterio:
@ethanadams Can you show me the query you are submitting to console?
Did you reinstall console after applying this PR?

Barterio:
@ethanadams On frontend this query used:
```
            `query($projectId: String!, $limit: Int!, $search: String!, $page: Int!, $order: Int!, $orderDirection: Int!) {
                project (
                    id: $projectId,
                ) {
                    apiKeys (
                        cursor: {
                            limit: $limit,
                            search: $search,
                            page: $page,
                            order: $order,
                            orderDirection: $orderDirection
                        }
                    ) {
                        apiKeys {
                            id,
                            name,
                            createdAt
                        }
                        search, 
                        limit, 
                        order,
                        pageCount, 
                        currentPage,
                        totalCount
                    }
                }
            }`;
```

Barterio:
@ethanadams Seems that you use new backend with outdated frontend.
Frontend changes(API and other) will be implemented(PR opened) when this PR will be approved to merge it together.
This made in that way to make it easier to review small PR's

satellite/satellitedb/apikeys.go
```diff
@@ -15,41 +16,130 @@ import (
...
+			}
 		}
 
-		apiKeys = append(apiKeys, *info)
+		ak.ID = keyID
+		ak.ProjectID = projectID
+		ak.PartnerID = partnerID
+
+		apiKeys = append(apiKeys, ak)
 	}
```
egonelbre: This code is missing a check for `rows.Err()`. See example in https://golang.org/pkg/database/sql/#Rows

Barterio: Should I double check for `rows.Err()` after all `rows.Next()` even if I perform all err check inside for loop?

egonelbre: Yes. The `rows.Err()` is about `rows.Next()` returning an error, not about scan itself.

PR #2843, opened 2019-08-21
Experimental refactoring of storage node migrations to support multi-db.

simongui:
**What:** This PR adds the ability for [storage node migration tests](https://github.com/storj/storj/blob/fac9be2130fdb3b04cf2c5e564addbae860e5c94/storagenode/storagenodedb/migrations_test.go#L62) to use multiple database connections in the same way our storage nodes do in production. 

**Why:** Another WIP PR is splitting the storage node SQLite tables into individual databases and there have been PR's preparing the code base for the transition. The business objects in the `master` branch now support multiple connections but temporarily point to a single connection while we finish this transition. To make `TestMigrate` pass during this transition it executes all the SQL statements against `db.VersionsMigration()` just to get access to the 1 SQLite connection. This was temporary.

It's important our migration tests use our databases in the same way our storage nodes will so that we detect errors in our migrations correctly so we can't rely on executing SQL statements only against `db.VersionsMigration()` so this PR proposes one approach to refactoring the snapshots to support detecting which connection a snapshot is related to.

**WIP:** This PR is currently just in experimental form and below are some notes on concerns or thoughts.

Currently we execute the SQL scripts against a single database [here](https://github.com/storj/storj/blob/sg/add-multidbs-to-snmigrations/storagenode/storagenodedb/migrations_test.go#L87). A single migration version may touch multiple database connections in the future so we need a way to define what database a migrations and test snapshots uses. I experimented with doing it [like this](https://github.com/storj/storj/blob/sg/add-multidbs-to-snmigrations/storagenode/storagenodedb/testdata/snapshot_v0.go). This may not be the best way. Thoughts?

**TODO:** There's no actual mapping of database connections to database connection names yet. Thoughts on where that should go? Maybe a map of `map[string]SQLDB` in `storagenode/storagenodedb/database.go`?

This may not be the best approach, this is just what I started experimenting with in the first hour of looking at it.

I started to convert 2 snapshots to the new codified format in `testdata/snapshot_v0.go` and `testdata/snapshot_v1.go`.

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/testdata/snapshot_v0.go
```diff
@@ -0,0 +1,150 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package testdata
+
+var Snapshot_v0 = Snapshot{
```
egonelbre: Maybe something like:

```
var _ = Snapshots.Add(Snapshot{
    Version: 1,
    Databases: Steps{
        "versions": {
            Schema: `
                -- table for keeping serials that need to be verified against
                CREATE TABLE used_serial (
                    satellite_id  BLOB NOT NULL,
                    serial_number BLOB NOT NULL,
                    expiration    TIMESTAMP NOT NULL
                );
                -- primary key on satellite id and serial number
                CREATE UNIQUE INDEX pk_used_serial ON used_serial(satellite_id, serial_number);
                -- expiration index to allow fast deletion
                CREATE INDEX idx_used_serial ON used_serial(expiration);
            `,
            Data: `
                INSERT INTO used_serial VALUES(X'0693a8529105f5ff763e30b6f58ead3fe7a4f93f32b4b298073c01b2b39fa76e',X'18283dd3cec0a5abf6112e903549bdff','2019-04-01 18:58:53.3169599+00:00');
                INSERT INTO used_serial VALUES(X'976a6bbcfcec9d96d847f8642c377d5f23c118187fb0ca21e9e1c5a9fbafa5f7',X'18283dd3cec0a5abf6112e903549bdff','2019-04-01 18:58:53.3169599+00:00');
            `,
            NewData: `
                INSERT INTO used_serial VALUES(X'0693a8529105f5ff763e30b6f58ead3fe7a4f93f32b4b298073c01b2b39fa76e',X'18283dd3cec0a5abf6112e903549bdff','2019-04-01 18:58:53.3169599+00:00');
                INSERT INTO used_serial VALUES(X'976a6bbcfcec9d96d847f8642c377d5f23c118187fb0ca21e9e1c5a9fbafa5f7',X'18283dd3cec0a5abf6112e903549bdff','2019-04-01 18:58:53.3169599+00:00');
            `,
        },
        "orders": {
            Schema: `
                ...
            `,
            ...
        }
    },
})
```

PR #2852, opened 2019-08-22
docs/design: Deletion Performance

egonelbre:
What: Draft notes on deletion performance.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

jtolds:
The more I've thought about this, the more I've become persuaded that we should hoist the 80-way delete from the Uplink to the Satellite. Instead of the Uplink issuing 80 deletes directly to storage nodes, the Uplink should issue a single request to the Satellite, and the Satellite should issue 80 deletes in parallel, while making the Uplink wait. The Satellite can then tune what percent it wants to wait for a response from before replying to the Uplink that the operation was a success. We would not need Orders anymore for this, but it would still provide appropriate delete operation backpressure. We could tune delete performance on the Satellite without worrying that an Uplink is trying to cheat.

calebcase:
Would we also be able to do long tail removal by issuing 80 deletes, but bailing after 52 return?

docs/blueprints/deletion-performance.md
```diff
@@ -0,0 +1,39 @@
...
+
+[A discussion of alternate approaches and the trade offs, advantages, and disadvantages of the specified approach.]
+
+## Implementation
+
+[A description of the steps in the implementation.]
+
+## Open issues (if applicable)
+
+We could probablistically skip deleting pieces. This would minimize the requests that the uplink has to make, at the same time not leaking too much pieces.
```
ifraixedes: Could you elaborate this a little bit in order to understand how a probabilistic approach would work and help?

docs/blueprints/deletion-performance.md
```diff
@@ -0,0 +1,39 @@
...
+- We need to ensure that storage nodes aren't storing too much garbage, since it reduces overall capacity of the network.
+- We need to ensure that satellites aren't storing too much garbage, since that increases running cost of the satellite.
+
+TODO: where exactly are deletes spending time.
+
+## Design
+
+First we can do reduce timeouts for delete requests. Undeleted pieces will eventually get garbage collected, so we can allow some of them to get lost.
+
+The uplink should be able to issue the request without having to wait for the deletion to happen. Currently deletions is implemented as an RPC, instead, use a call without waiting for a response.
```
ifraixedes: I don't get this point. What "a call" means here? Doesn't an RPC response time base on what the server is executing rather than the protocol response itself?

docs/blueprints/deletion-performance.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Background
+
+Current design requires uplinks to send deletion information to all the storage nodes. This ends up taking a considerable amount of time.
+
+There are few goals with regards to deletions:
+
+- We need to ensure that uplinks are responsive.
+- We need to ensure that storage nodes aren't storing too much garbage, since it reduces overall capacity of the network.
+- We need to ensure that satellites aren't storing too much garbage, since that increases running cost of the satellite.
```
ifraixedes: How satellite could store garbage in a delete?
I understand that the garbage data in the storage node side, but I don't get it in the satellite.

docs/blueprints/deletion-performance.md
egonelbre: For example we could only send the deletion requests to only 50%, leaking half of the data, but making half the requests.

docs/blueprints/deletion-performance.md
egonelbre: For example, node could internally delete things async; alternatively we could use UDP for delete requests.

docs/blueprints/deletion-performance.md
egonelbre: For example, we could use tombstones on the pointers, marking them as deleted and then sending delete requests as a batch to the satellite. Similarly we could keep a list of pieces to delete per storage node. Effectively, this would mean that satellite is holding on to garbage so that it could be collected later.

docs/blueprints/deletion-performance.md
```diff
@@ -0,0 +1,39 @@
...
+
+[A discussion of alternate approaches and the trade offs, advantages, and disadvantages of the specified approach.]
+
+## Implementation
+
+[A description of the steps in the implementation.]
+
+## Open issues (if applicable)
+
+We could probablistically skip deleting pieces. This would minimize the requests that the uplink has to make, at the same time not leaking too much pieces. For example we could only send the deletion requests to only 50%, leaking half of the data, but making half the requests.
```
mniewrzal: Maybe we can request to delete from storage nodes that are currently online (according to our knowledge)? Make no sense try to delete piece from node that most probably is offline.

egonelbre:
Superseded by https://github.com/storj/storj/pull/3520

PR #2853, opened 2019-08-22
satellite/marketingweb: fix broken pipe error

VinozzZ:
What: 

If clients drop the connection, it's expected that we will be getting an EPIPE error. Therefore, we will ignore the error in that situation

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

egonelbre:
The question is. Where are the broken pipes coming from? The server shouldn't shutdown before finishing the tests and checking the responses.

VinozzZ:
> The tests shouldn't cause this error.

AFAIK, The reason is because we have redirect in the POST request handler, so the server would shutdown after the response from POST request, but not wait for the redirect to finish

ifraixedes:
For the same reason than @egonelbre it looks strange to me that we need to do this.

> AFAIK, The reason is because we have redirect in the POST request handler, so the server would shutdown after the response from POST request, but not wait for the redirect to finish

I'm not convinced about this; do you have the test output log when that error false positive error happens? it may help me to understand it.

egonelbre:
@ifraixedes, I think the current solution is the right one... the comment doesn't sound right, but the issue is the server starts writing to the connection, nothing is reading it... eventually the thing gets shut down. But also I think there should be some additional checks to ensure that the requests got the right response, not just that they succeeded.

VinozzZ:
@egonelbre I added checks for response status code

PR #2855, opened 2019-08-22
uplink/paths/encryption: support commandline argument to override path cipher to be urlsafe base64 for lists and deletes

jtolds:
What: allow listing and removal on base64-encoded, but encrypted paths
Why: let's say you have api key access to a full bucket. someone uploaded things with the wrong encryption key and now you can't see them. you don't have the encryption key but you'd like to remove them. this lets you find and remove things without encryption keys.

Please describe the tests:
 - pkg/encryption/path_test.go
    - `TestEncryptPath_EncryptionBypass`:
  
      Exercises path encryption to ensure that the encrypted path matches the encryption-bypass encrypted path.
    - `TestDecryptPath_EncryptionBypass`:
  
      Exercises path encryption to ensure that the encryption-bypass "decrypted" path matches the encrypted+url-base64-encoded path.
- uplink/metainfo/kvmetainfo/object_test.go
    - `TestListObjects_EncryptionBypass`:

      Exercises object listing to ensure that when listing with encryption bypass enabled, each path can be decoded and decrypted to match an uploaded object path (only one match per path is allowed and number of objects must match).
    - `TestDeleteObject` (updated):

      Exercises object deletion to ensure no error when deleting with and without encryption bypass.
      _TODO?: maybe we should do a list objects or something to ensure the objects were actually deleted._

 

pkg/encryption/store.go
```diff
@@ -34,6 +34,16 @@ import (
 type Store struct {
 	roots      map[string]*node
 	defaultKey *storj.Key
+
+	// TODO: The below field makes it so we can list and remove files without
+	// trying to decrypt their paths. However, this is the wrong long-term solution.
+	// The right long-term solution is to move the PathCipher out of the bucket
```
jenlij: oh yeah this business again heh

cmd/uplink/cmd/ls.go
```diff
@@ -26,7 +27,8 @@ func init() {
 		Short: "List objects and prefixes or all buckets",
 		RunE:  list,
 	}, RootCmd)
-	recursiveFlag = lsCmd.Flags().Bool("recursive", false, "if true, list recursively")
+	lsRecursiveFlag = lsCmd.Flags().Bool("recursive", false, "if true, list recursively")
```
jenlij: so if I call ls with recursive and I don't have the encryption key, I won't be able to see my files? But If I call it with encrypted, I will be able to see them, but encoded? What if I call encrypted with the right encryption key?

pkg/encryption/encryption.go
```diff
@@ -46,6 +46,8 @@ func Encrypt(data []byte, cipher storj.CipherSuite, key *storj.Key, nonce *storj
 		return EncryptAESGCM(data, key, ToAESGCMNonce(nonce))
 	case storj.EncSecretBox:
 		return EncryptSecretBox(data, key, nonce)
+	case storj.EncURLSafeBase64:
```
jenlij: why doesn't this work?

jenlij:
In regards to documentation, it might be helpful to add this flag/feature to any uplink documentation we have.

jtolds:
Rebased and moved to https://review.dev.storj.io/c/storj/storj/+/102

cmd/uplink/cmd/ls.go
bryanchriswhite: You will get all `url-base64-encoded(encrypted)` paths when using `--encrypted`. No path decryption is attempted.

pkg/encryption/encryption.go
bryanchriswhite: Afaik this is for encrypting data as opposed to paths. We don't want anyone accidentally url-base64-encoding data they intend to be encrypted, instead of actually encrypting it.

pkg/encryption/path.go
```diff
@@ -241,6 +242,15 @@ func derivePathKeyComponent(key *storj.Key, component string) (*storj.Key, error
 
 // encryptPathComponent encrypts a single path component with the provided cipher and key.
 func encryptPathComponent(comp string, cipher storj.CipherSuite, key *storj.Key) (string, error) {
+
+	if cipher == storj.EncURLSafeBase64 {
+		decoded, err := base64.URLEncoding.DecodeString(comp)
+		if err != nil {
+			return "", Error.New("invalid base64 data: %v", err)
+		}
+		return string(decoded), nil
```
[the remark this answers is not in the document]
bryanchriswhite: The output of this function should always be **encrypted**, in the special case where the cipher is `storj.EncURLSafeBase64` the path component is expected to be the `url-base64-encode(encrypted)` path. This allows us to get the encrypted path just by url-base64-decoding the path component.

I don't think there is enough functionality here (especially not in `decryptPathComponent`) to pull out into a separate method(s).

pkg/encryption/path.go
```diff
@@ -279,6 +289,10 @@ func decryptPathComponent(comp string, cipher storj.CipherSuite, key *storj.Key)
 		return "", nil
 	}
 
+	if cipher == storj.EncURLSafeBase64 {
+		return base64.URLEncoding.EncodeToString([]byte(comp)), nil
```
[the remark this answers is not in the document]
bryanchriswhite: ^

uplink/storage/streams/store.go
```diff
@@ -428,23 +428,44 @@ func (s *streamStore) Delete(ctx context.Context, path Path, pathCipher storj.Ci
...
+				if streamMeta.NumberOfSegments > 0 {
+					// the segment is missing but this is a new pointer. let's keep going
+					// until the counter runs out
+					continue
+				} else {
+					// the segment is missing and this is an old pointer. there might not
+					// be any more segments after this. stop.
+					break
+				}
+			}
```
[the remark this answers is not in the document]
bryanchriswhite: tests added

cmd/uplink/cmd/ls.go
bryanchriswhite: From a usability standpoint it might be nice if we always try to decrypt but fail-over to url-base64-encoding. When there is an object in a bucket that was encrypted with a different key it can be difficult to determine which object that is just by using `uplink ls --encrypted`. Currently you won't be able to `uplink ls` a prefix that contains a path component that was encrypted with that other key.

egonelbre: Automatic fail-over with tools that are used in scripts can be problematic.

bryanchriswhite: Maybe instead a subcommand to convert between unencrypted and encrypted-encoded paths or something then.

bryanchriswhite: Or we could add a flag to enable that behavior.

pkg/encryption/store.go
```diff
@@ -34,6 +34,11 @@ import (
 type Store struct {
 	roots      map[string]*node
 	defaultKey *storj.Key
+
+	// TODO: The below field makes it so we can interoperate with
+	// the network without having encryption keys. paths will be encrypted but
+	// base64-encoded, and certain metadata will be unable to be retrieved.
+	EncryptionBypass bool
```
kaloyan-raev: Why is this a TODO?

bryanchriswhite: I'm not sure. I could make it an `NB:` or just remove the `TODO:`.

kaloyan-raev: Yeah, perhaps just rephrase it so it becomes the description for this field, starting with "EncryptionBypass ... "

bryanchriswhite: :+1:

uplink/metainfo/kvmetainfo/buckets_test.go
```diff
@@ -213,17 +213,34 @@ func getBucketNames(bucketList storj.BucketList) []string {
...
+	key := new(storj.Key)
+	copy(key[:], keyStr)
+
+	store := encryption.NewStore()
+	store.SetDefaultKey(key)
+
+	return store
+}
+
+func runPlanet(t *testing.T, test func(*testing.T, context.Context, *testplanet.Planet)) {
```
egonelbre: Why this helper? It doesn't look like it's adding anything much compared to using `testplanet.Run` directly.

bryanchriswhite: It used have more in it.. I kept it so that I wouldn't have to duplicate the testplanet config. I can remove it if you like - I don't feel strongly about it.

cmd/uplink/cmd/ls.go
```diff
@@ -26,7 +27,8 @@ func init() {
...
+	lsEncryptedFlag = lsCmd.Flags().Bool("encrypted", false, "if true, show paths as base64-encoded encrypted paths")
```
egonelbre: Why do we need an extra flag when we already have `enc.path-type`?

pkg/storj/encryption.go
```diff
@@ -44,6 +44,9 @@ const (
 	// EncSecretBox indicates use of XSalsa20-Poly1305 encryption, as provided
 	// by the NaCl cryptography library under the name "Secretbox".
 	EncSecretBox
+	// EncURLSafeBase64 is like EncNull but Base64 encodes/decodes the
+	// binary path data (URL-safe)
+	EncURLSafeBase64
```
egonelbre: Wondering whether it should be named `EncNullBase64URL` the encoding is referred to as "base64url" https://tools.ietf.org/html/rfc4648#section-5.

Alternatively `NullBase64URL` or something.

uplink/metainfo/kvmetainfo/buckets_test.go
egonelbre: Yeah, I would probably avoid it because it only saves two lines, without much benefit.

cmd/uplink/cmd/ls.go
bryanchriswhite: Using `enc.path-type` won't have any effect on a list or delete operation.

pkg/storj/encryption.go
bryanchriswhite: :+1:

cmd/uplink/cmd/ls.go
kaloyan-raev: Isn't `enc.path-type` used only when creating a new bucket?

kaloyan-raev: Perhaps, we can rename `--encrypted` to `--encrypted-paths` to be clearer that is related to paths only.

egonelbre: One option is to have a separate command entirely as well `ls-encrypted` or something.

bryanchriswhite: I believe that's correct. The only place I see it used in uplink is in `GetPathCipherSuite` (uplink/config.go:149) and that's used in `makeBucket` (uplink/cmd/mb.go:58).

PR #2837, opened 2019-08-21
design/docs: add successful pingback to kademlia removal document

egonelbre:
What: Add https://github.com/storj/storj/pull/2686#discussion_r315990346 to kademlia removal document and simplify language.

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact:

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?
 - [ ] Do the database access patterns make sense?

PR #2857, opened 2019-08-22
docs/design: Storage Node downtime tracking

ifraixedes:
What: Add the initial draft of the Storage Node downtime tracking design
document without the implementation steps.

Why: Because first the design should be reviewed and once it is approved I'll add the implementation steps.


docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,202 @@
...
+    uptime_success_count = uptime_success_count + 1,
+    total_uptime_count = total_uptime_count +1
+  WHERE
+    id = ?
+  ```
+* On failure:
+
+  Calculate the number of seconds offline from now and the last contact failure.
+  ```
+  num_seconds_offline =  seconds(from: last_contact_failure, to: now())
```
mobyvb: won't this only work properly if the last contact failure also updated the `nodes_offline_time` table? We will have to make sure we do not update `last_contact_success` or `last_contact_failure` without also updating the new table.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,202 @@
...
+            id = ?
+    ```
+* On failure:
+
+  It calculates the number of offline seconds. We consider that the Storage Node must contact the Satellite every hour.
+
+  Because we don't know exactly at what point it has got offline, we dismiss the hour between the last contact success and the supposed next one.
+
+  ```
+  num_seconds_offline =  seconds(from: last_contact_success, to: now() - 1h)
```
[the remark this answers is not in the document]
littleskunk: Not possible because of the SQL query `last_contact_success < (now()  - 1h)`

docs/design/storagenode-downtime-tracking.md
ifraixedes: That's a very good point @mobyvb 
It's a case that I didn't contemplate.

I've been thinking in which case we wouldn't track some offline period of time and I drew a diagram for being clear about it and, hopefully, to show the cases to others easily.

offline tracking-Offline tracking with other audit updates

We have 2 choices:

1. Assume that we will lose some amount of offline tracking applying the premise commented at the beginning of the design section "Roughly estimate how long they are offline".
2. Do something about it.

### Choice 1 

M understanding that there won't be that many time on what we would lose the tracking of that offline time.

If I'm correct, _audit_ is mostly the procedure which updates the _uptime failure_ in an unregular and "indeterministic" way.

Audit service picks a random stripe and gets the pointer to the segment for downloading erasure shares. During that process, it can detect the Storage Nodes which are offline and update the `last_contact_failure`.

Because of its randomness I assuming that there won't be that many cases where `last_contact_failure` is updated before the _estimate offline time process_ select the node to verify if it's still offline.

What I dislike of this solution is that there is no clear bounding what it's in charge of updating and accessing to a certain part of the database, however, that isn't unusual in a monolith system.

### Choice 2

I've spent some time thinking about how we could track offline time.

Any solution, using the database (adding more columns, tables, queries, etc.) that came to the top of my head didn't work, while the time could be tracked, I found that there were some cases that we would track for some SN more offline track that the one that we know for sure that the SN is offline. That's unadmissible.

I've found a solution. `last_contact_failure` is only updated by the _SN downtime tracking service_. Any other process/service has to notify this service when it finds an SN which is offline.

An alternate solution is to rearhcitecture the system. Have an isolated service which will be in charge of calculating the node reputation, so it would execute the audits, uptimes checks, etc. It isn't a solution itself, mostly a way to rethink a part of the system which is related. Obviously, this requires much more work and it may be part of work being done to split the satellite in different independent services.


I'll present this on today's architecture meeting.

docs/design/storagenode-downtime-tracking.md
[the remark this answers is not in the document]
jtolds: assuming this happens in a transaction, sure, otherwise there's a race condition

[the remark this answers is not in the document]
ifraixedes: I'm not sure if I understand correctly the possible race condition.

I haven't thought about the transaction between we get the node and we update its `last_contact_failure` (the same applies for the success).

We could have a transaction, but I don't know if it's very ideal to block the DB during all the SN ping process.

If we don't have a transaction what can happen?

1. We override `last_contact_success` without reading its last value
2. We override `last_contact_failure` without reading its last value.
3. We update `last_contact_failure` while another process updates `last_contact_success` 
4. We update `last_contact_success` while another process updates `last_contact_failure`

I don't find any problem with _1_ and _2_ except that the 2 processes of _downtime tracking system_ will pick up the SN again a bit earlier or later.

I don't either find a problem with _3_ and _4_ despite the _downtime tracking system_ could run one iteration more depending on if their values are greater or lesser.

Changing when then next _uptime check_ will be done by the _downtime tracking system_ outside on when it should be according to its algorithm doesn't seem problematic.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,207 @@
...
+- `last_contact_success`
+- `last_contact_failure`
+- `total_uptime_count`
+- `uptime_success_count`
+
+### Detecting offline nodes
+
+An independent process runs, in a configurable interval, of time the following query:
+
+```sql
```
mobyvb: so this process is querying for nodes that have been contacted over an hour ago, but where `last_contact_success > last_contact_failure`? Won't that give us mostly online nodes? I'd think we would want to check `last_contact_success < last_contact_failure` in this query in order to get offline nodes that have not been seen in the past hour.

docs/design/storagenode-downtime-tracking.md
mobyvb: > I've found a solution. last_contact_failure is only updated by the SN downtime tracking service. Any other process/service has to notify this service when it finds an SN which is offline.

I think the issue with this is it could have an unexpected performance impact. And part of the reason for redefining how we do uptime tracking is to try not to do as much db work for every little thing that might indirectly affect uptime. The idea of interconnecting any service that might detect that a node is offline with this downtime tracker seems intimidating to me (for a first implementation, at least).

For that reason and the other reasons you provided, I think it is better to go with choice 1, and disregard uptime data that is outside the scope of the processes you have defined in this document. It should still be enough data for us to provide nodes with sufficient proof for disqualification, and will be less complicated than the alternative.

ifraixedes: > I think the issue with this is it could have an unexpected performance impact.

Yes, it should be slower because currently, the parts update the DB directly, after this, they will have to contact to this new service/process and it will update the DB, hence it adds an indirection.

>  And part of the reason for redefining how we do uptime tracking is to try not to do as much db work for every little thing that might indirectly affect uptime.

I don't think that the DB will be more used, the number of updates that it should receive will be the same, but rather than being done directly by some different parts of the code, they will be "queued" to the new service/process.

> For that reason and the other reasons you provided, I think it is better to go with choice 1, and disregard uptime data that is outside the scope of the processes you have defined in this document. It should still be enough data for us to provide nodes with sufficient proof for disqualification, and will be less complicated than the alternative.

I fully agree with that.

docs/design/storagenode-downtime-tracking.md
ifraixedes: > so this process is querying for nodes that have been contacted over an hour ago, but where last_contact_success > last_contact_failure? Won't that give us mostly online nodes?

No, per Kademlia removal design document, the SNs have to ping the satellite every hour, hence if they don't do so, then we have to find out what happens:

1. SN did the ping but due "some network failure" the ping didn't reach the Satellite.
1. SN did the ping but the Satellite wasn't available.
1. SN didn't do the ping.

All of those 3 scenarios cause that the `last_contact_success` doesn't be updated in the 1-hour interval.

From the Satellite point of view, we don't know that of those 3 general cases could be so the way is doing a specific check; that's what this part of the algorithm does.

The Kademlia removal design document doesn't specify any mechanism for ensuring that SNs have contacted the Satellite every hour; this is the part that does such verification.

If the Satellite contacts before the 1-hour the SN (this happens for example by the audit process), then we have:

1. A proof that's online before the 1-hour interval. We can wait for one hour to hear from it again.
1. A proof that's offline. We have to start to measure its offline time, that's handled by the second part of the algorithm (section _Estimating offline time_).

There is another moment when SNs must ping the satellite when they startup (first time, reboot, etc.), but those SNs will be handled as the first case above.

> I'd think we would want to check last_contact_success < last_contact_failure in this query in order to get offline nodes that have not been seen in the past hour.

No, that's handled by the second part of the algorithm (section _Estimating offline time_).

That situation happens:

1. An SN hasn't contacted the Satellite in the 1-hour interval. The Satellite check the SN and it's offline and it updates the `last_contact_failure`. This is done by this part of the algorithm.
1. Another process of the satellite (e.g. Audit) has tried to contact the SN and it was offline. It updates the `last_contact_failure.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Design
+
+For tracking Storage Nodes offline time we need:
+
+- A new SQL database table to store offline time.
+- Detecting which ones are offline.
+- Roughly estimate how long they are offline.
+
+An _uptime check_ referenced in this section is the connection initiated by the Satellite to any Storage Node in the same way that's described in [Network refreshing section of the Kademlia removal design document](kademia-removal.md#network-refreshing).
```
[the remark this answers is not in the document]
ifraixedes: This was an anticipation of the future https://github.com/storj/storj/pull/2837/files#diff-a7e10556bf95db26fb42323ea62d3e68R42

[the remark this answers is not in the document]
bryanchriswhite: oh, nice! :D my mistake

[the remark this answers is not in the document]
ifraixedes: no prob!

I had to reverify which I took or if I did a mistake, I couldn't remember.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,207 @@
...
+The current Satellite database has the table `nodes`. For the offline time calculation we use the following columns:
+
+- `last_contact_success`
+- `last_contact_failure`
+- `total_uptime_count`
+- `uptime_success_count`
+
+### Detecting offline nodes
+
+An independent process runs, in a configurable interval, of time the following query:
```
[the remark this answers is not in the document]
ifraixedes: I'll add a clarification mentioning the kademlia removal requirement

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,207 @@
...
+  ```sql
+  UPDATE nodes
+  SET
+    last_contact_failure = now(),
+    total_uptime_count = total_uptime_count +1
+  WHERE
+    id = ?
+  ```
+
+## Rationale
```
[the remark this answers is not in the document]
ifraixedes: I'll also add the diagram

docs/design/storagenode-downtime-tracking.md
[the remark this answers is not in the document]
ifraixedes: I added it, please review it.

docs/design/storagenode-downtime-tracking.md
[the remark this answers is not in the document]
ifraixedes: I added it, please review it.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,221 @@
...
+### Database
+
+The following new SQL database table will store Storage Nodes offline time.
+
+```sql
+CREATE TABLE nodes_offline_time(
+    node_id BYTEA NOT NULL,
+    tracked_at timestamp with time zone NOT NULL,
+    seconds integer,                                                    -- Measured number of seconds being offline
+    CONSTRAINT nodes_offline_time__pk PRIMARY KEY(node_id, tracked_at)
```
egonelbre: do we need the constraint on `tracked_at`? It seems insufficient for detecting overlaps. At the same time, how likely is it to get the same timestamp.

Maybe this should be just an index.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,221 @@
...
+
+The process runs the query repeatable until it doesn't return records then it ends and will run again in the next interval. The query only gets one record at the time for not holding a record set during the process.
+
+For each node, the Satellite performs an _uptime check_.
+
+* On success
+    ```sql
+        UPDATE nodes
+        SET
+            last_contact_success = now(),
```
egonelbre: Wondering whether this should be `last_contact_success = max(last_contact_success, now()),` which would play more nicely with batching.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,221 @@
...
+    id = ?
+  ```
+
+## Rationale
+
+The designed approach has the drawback that `last_contact_failure` of the `nodes` table may get updated by other satellite services before the _estimating offline time_ subprocess reads the last value and calculates the number of offline seconds.
+
+The following diagram shows one of these scenarios:
+
+![missing tracking offline seconds](images/storagenode-downtime-tracking-missing-tracking-offline-seconds.png)
```
egonelbre: Image is not working.

ifraixedes: Fixed!

docs/design/storagenode-downtime-tracking.md
ifraixedes: I'm thinking that we don't even need an index for the purpose of this document.

The document that should specify how to read this data it's the one that should specify which indexes are needed.

docs/design/storagenode-downtime-tracking.md
ifraixedes: yes, that's much better.
And the same applies for the _on success_  of the _Estimating offline time_ section

ifraixedes:
I'll rename the document to `storage-node-downtime-tracking.md` (just add a `-` between `storage` and `node`), but holding until https://github.com/storj/storj/pull/2884 gets merged into this branch.

docs/design/storagenode-downtime-tracking.md
ifraixedes: Applied.

docs/design/storagenode-downtime-tracking.md
```diff
@@ -0,0 +1,207 @@
...
+```sql
+SELECT
+FROM nodes
+WHERE
+    last_contact_success < (now()  - 1h) AND
+    last_contact_success > last_contact_failure AND
+    disqualified IS NULL
+ORDER BY
+    last_contact_success ASC
+LIMIT 1
```
littleskunk: The limit 1 in combination with the order is a build in killswitch. Lets say this query runs in an interval of 30s. Per day the satellite can track about 3000 storage node. As cheater I don't want that the satellite notice my downtime. All I have to do is creating 3000 fake nodes. The job of the fake nodes is to get selected first. They will ignore all other calls except that one important ping message. If they answer that ping they will get a last_contact_success renew and 0 downtime. At the end the satellite will cycle through the same 3000 fake nodes all the time. With that cycle the satellite my cheater node can stay offline for 1 day without any penalty. By increasing the number of fake nodes I can increase the timespan my cheater node can stay offline.

The issue here is scaling. I used an example with fake nodes but in production this might happen without any malicious intention. The big question is why do we want to limit the query to 1? Why don't we ping them all?

docs/design/storagenode-downtime-tracking.md
littleskunk: > Audit service picks a random stripe and gets the pointer to the segment for downloading erasure shares. During that process, it can detect the Storage Nodes which are offline and update the last_contact_failure.

That is not correct. There are 2 edge cases we would count a storage node as offline even if it is not.
1.) The download takes more than 25 seconds. As far as I know the audit job counts that as offline and will update the nodes table.
2.) The storage node wasn't pinged for a long time. In that case the audit job will assume that the node is offline and don't even try to contact it.

In general the audit job is the reason the old uptime tracking failed. The storage nodes are getting an uneven number of uptime checks and there is no ping involved. The audit job might flag you as offline even if you are available. I don't like the idea of running into the same trap again. Somewhere in the overlay service we have a job that is contacting the storage nodes every 2 hours. That sounds like a better job to identify offline nodes.

ifraixedes: The audit service only updates the `last_contact_failure`; there is no plan to reactivate the current uptime disqualification system, so the SN won't get disqualified.

The SN is only pinged by this new chore, audit or whatever else which currently updates the `last_contact_success` and `last_contact_failure` isn't going to be changed to do any ping.

ifraixedes: My plan is to add to the implementation section to remove the current uptime disqualification system because anyway, it's disabled.

docs/design/storagenode-downtime-tracking.md
ifraixedes: We can ping them all as we can.
There is a limit of 1 for avoiding to hold the result set for a long time, but if that's not an issue then we can run without the limit.

I commented that in https://github.com/storj/storj/pull/2884#discussion_r318168100

ifraixedes: The limit will be removed because there is no DB performance issue.

ifraixedes: It has been removed.

docs/design/storagenode-downtime-tracking.md
ifraixedes: I've removed the constraint.

docs/design/storage-node-downtime-tracking.md
```diff
@@ -0,0 +1,209 @@
...
+- `last_contact_success`
+- `last_contact_failure`
+- `total_uptime_count`
+- `uptime_success_count`
+
+### Detecting offline nodes
+
+Per [Kademlia removal design document](https://github.com/storj/storj/blob/master/docs/design/kademlia-removal.md#network-refreshing), any storage node has to ping the satellite every hour. For storage nodes that have not pinged, we need to contact them directly.
+
+For finding the potentially offline storage nodes, we run a chore, with the following query:
```
egonelbre: For finding the storage nodes gone offline, we run a chore, with the following query:

docs/design/storage-node-downtime-tracking.md
```diff
@@ -0,0 +1,209 @@
...
+
+```sql
+SELECT
+FROM nodes
+WHERE
+    last_contact_success < (now()  - 1h) AND
+    last_contact_success > last_contact_failure AND
```
egonelbre: add a comment `-- only select nodes that were last known to be online`

docs/design/storage-node-downtime-tracking.md
```diff
@@ -0,0 +1,209 @@
...
+- Interval,
+- Number of nodes to check.
+
+The process loops all failed nodes with query:
+
+```sql
+SELECT
+FROM nodes
+WHERE
+    last_contact_success < last_contact_failure AND
```
egonelbre: add a comment `-- only select nodes that were last known to be offline`

docs/design/storage-node-downtime-tracking.md
```diff
@@ -0,0 +1,209 @@
...
+ORDER BY
+    last_contact_success ASC
+```
+
+For each node, the satellite performs an _uptime check_.
+
+* On success, it updates the nodes table with the last contact information:
+
+    ```sql
+        UPDATE nodes
```
egonelbre: remove extra ident from code blocks

ifraixedes:
@zeebo @littleskunk @egonelbre 
I need that any of you approve the PR again for being able to merge it.

Since your last review, the only that changed has been some minor formatting,  rephrasing changes and __adding the implementation section__ and __a couple of items in the issues section__.

Thanks again for your time and sorry for all these pings.

PR #2858, opened 2019-08-22
satellite/nodestats: return storage usage in Byte*hours

rikysya:
What: Make satellite nodestats endpoint return storage usage in Byte*hours instead of trying to query raw Byte values.

Why: Storage usage intended to give user information about expected payout, as we form our payouts table using Byte*hours from `accounting_rollups` table it makes more sense to show it this way to SNO users. Also, calculating raw Byte values is tricky and error prone, so removing it simplifies the flow.

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -223,35 +245,26 @@ func (db *StoragenodeAccounting) QueryStorageNodeUsage(ctx context.Context, node
 		err = errs.Combine(err, rows.Close())
 	}()
 
+	fmt.Println("NodeID: ", nodeID.String())
+	fmt.Println("NodeID unformat", string(nodeID[:]))
```
navillasa: do these print lines need to be here?

rikysya: oops

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +201,41 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
-		WHERE id IN (
-			SELECT MAX(id)
-			FROM accounting_rollups
-			WHERE node_id = ?
-			AND ? <= start_time AND start_time <= ?
-			GROUP BY start_time
-			ORDER BY start_time ASC
-		)
-		WINDOW win AS (ORDER BY start_time)`
+	lastRollup, err := db.db.Find_AccountingTimestamps_Value_By_Name(ctx, dbx.AccountingTimestamps_Name(accounting.LastRollup))
```
rikysya: Why am I not able to get last rollup timestamp in test?

jenlij: which test? what is it giving you instead?

rikysya: `Db_test.go` `TestStorageUsage` I am receiving nil with no error

internal/dbutil/nulltime.go
```diff
@@ -48,3 +49,18 @@ func (nt NullTime) Value() (driver.Value, error) {
 	}
 	return nt.Time, nil
 }
+
+// parseSQLITETimeString parses sqlite times string.
```
jenlij: nit pick, but it may be easier to read if you name it parseSqliteTimeString

internal/dbutil/nulltime.go
```diff
@@ -48,3 +49,18 @@ func (nt NullTime) Value() (driver.Value, error) {
...
+func parseSqliteTimeString(val string) (time.Time, error) {
+	var times time.Time
+	var err error
+
+	times, err = time.Parse(sqliteTimeLayout, val)
+	if err == nil {
+		return times, nil
+	}
+
+	return time.Parse(sqliteTimeLayoutDate, val)
```
jtolds: this makes me really worried. the date in what timezone?

rikysya: If time zone is not provided to `time.Parse` then it is treated as `utc`. As SQLite has no timezone support and we ensure that it we insert `utc` timezone in our code then I guess it should be fine

rikysya: I am little worried about query itself, if we are not deleting `tallies` then I can possibly double count some values. Or maybe this query some performance issue, I wonder if I should lock timestamp and storageusage query it timestamp can possibly be outdated on second connection

jenlij: Is there a way we can always make it UTC in the code that is providing the time?

internal/dbutil/nulltime.go
```diff
@@ -32,7 +33,7 @@ func (nt *NullTime) Scan(value interface{}) error {
 		return nil
 	}
 
-	times, err := time.Parse(sqliteTimeLayout, string(date))
+	times, err := parseSqliteTimeString(string(date))
```
jenlij: Where is the Scan method called?

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,41 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+	if lastRollup == nil {
+		return nil, Error.New("unable to find last accounting rollup timestamp")
+	}
+
+	start, end = start.UTC(), end.UTC()
+
+	query := `SELECT SUM(at_rest_total), start_time
```
jenlij: This looks like it does what it's supposed to, but I'm not a SQL expert. I'd ask someone on data science to double check.

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,41 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
+				FROM (
+					SELECT r.at_rest_total, r.start_time
+					FROM (
+						SELECT MAX(id) as id
+						FROM accounting_rollups
+						WHERE node_id = ?
+						AND ? <= start_time AND start_time <= ?
+						GROUP BY start_time
```
ifraixedes: If I'm not mistaken this subquery only returns one row, one `id`.
If that's the case I would avoid the `GROUP BY`

```sql
SELECT id
FROM accounting_rollups
WHERE node_id = ?
AND ? <= start_time AND start_time <= ?
LIMIT 1
```

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,41 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
+					INNER JOIN accounting_rollups r ON r.id = ids.id
+					UNION
+					SELECT SUM(data_total) as at_rest_total, DATE(interval_end_time) as start_time
+					FROM storagenode_storage_tallies
+					WHERE node_id = ?
+					AND ? < interval_end_time AND interval_end_time <= ?
+					GROUP BY start_time
+				) stamps 
+				GROUP BY start_time
+				ORDER BY start_time ASC`
```
egonelbre: What's the performance of this query over a large database?

internal/dbutil/nulltime.go
egonelbre: Our tests are checking that queries against sqlite are using UTC. However I think we need an exact description how to handle timezones.

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,41 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
-			GROUP BY start_time
-			ORDER BY start_time ASC
-		)
-		WINDOW win AS (ORDER BY start_time)`
+	lastRollup, err := db.db.Find_AccountingTimestamps_Value_By_Name(ctx, dbx.AccountingTimestamps_Name(accounting.LastRollup))
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+	if lastRollup == nil {
+		return nil, Error.New("unable to find last accounting rollup timestamp")
```
egonelbre: Shouldn't this `return nil, nil`, because when there's no data then the totals should be `0`.

satellite/satellitedb/storagenodeaccounting.go
ifraixedes: I didn't know that `start_time` was unique, but it's filtering by `node_id` so having a maximum of a single value is that same value.

rikysya: @ifraixedes, @cam-a  `MAX` is leftover from debugging when I removed code that deletes today`s `rollups`. I left it because in our DB schema for `accounting` we allow duplicated values for the same timestamp

satellite/satellitedb/storagenodeaccounting.go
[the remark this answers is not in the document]
rikysya: @cam-a yeah, they should be unique, But I am not sure if  date from `DATE` function is the same as date inserted by our code

satellite/satellitedb/storagenodeaccounting.go
rikysya: @egonelbre so for DB with 365 * 10000 rollups and 366 * 10000 tallies:
- 0.396 seconds for SQlite
- 0.332 seconds for Postgres

egonelbre: Add it to the pr description as well.

egonelbre:
Note, waiting query tests by @stefanbenten

rikysya:
And query review by @pgerbes1

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,49 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
+				UNION
+					SELECT SUM(data_total) AS at_rest_total, DATE(interval_end_time) AS start_time
+					FROM storagenode_storage_tallies
+					WHERE node_id = ?
+					AND NOT EXISTS (SELECT 1 FROM r WHERE DATE(r.start_time) = DATE(interval_end_time))
+					AND (SELECT value FROM accounting_timestamps WHERE name = ?) < interval_end_time AND interval_end_time <= ?
+					GROUP BY start_time
+			ORDER BY start_time`
+
+	//query := `SELECT at_rest_total, start_time
```
ifraixedes: Please make sure to delete this before merging

satellite/satellitedb/storagenodeaccounting.go
```diff
@@ -200,21 +200,49 @@ func (db *StoragenodeAccounting) QueryPaymentInfo(ctx context.Context, start tim
...
-		WHERE id IN (
-			SELECT MAX(id)
-			FROM accounting_rollups
-			WHERE node_id = ?
-			AND ? <= start_time AND start_time <= ?
-			GROUP BY start_time
-			ORDER BY start_time ASC
-		)
-		WINDOW win AS (ORDER BY start_time)`
+	lastRollup, err := db.db.Find_AccountingTimestamps_Value_By_Name(ctx, dbx.AccountingTimestamps_Name(accounting.LastRollup))
```
ifraixedes: Is this query needed with the last changes?

Before the `lastRollup` value was used in the function, now it's only used for making this function to return earlier. If that's wanted, then fine, otherwise, we can remove it.

rikysya: Yes, it used to check whether we have rollup run at least once and `LastRollup` timestamp is not nil. Then we querying this value acquiring the same reading lock as the main query. This ensures that we are not blocking dB for other reads, and allows `rollup` and other services that writes to dB to acquire pending lock if needed so we are not blocking their work

satellite/satellitedb/storagenodeaccounting.go
rikysya: Yes, one of this query will be deleted as we need to leave one with best performance

satellite/satellitedb/storagenodeaccounting.go
ifraixedes: I think that I'm not at the level to understand that lock acquiring.
If you know that's needed it's fine to me unless that anybody else has something to say about.

rikysya: Well I just try to say that new query can avoid locking 2 db calls into TX which is blocking dB for read and writes

ifraixedes: so it can be removed, isn't it?

rikysya: We don’t need to execute second query if LastRollup is null

ifraixedes: How often `latRollup` is null?

rikysya: Before very first rollup

satellite/satellitedb/storagenodeaccounting.go
egonelbre: Why is this code commented out? If it's not used can we remove it.

satellite/accounting/db_test.go
```diff
@@ -51,39 +48,88 @@ func TestStorageNodeUsage(t *testing.T) {
 		ctx := testcontext.New(t)
 		defer ctx.Cleanup()
 
+		const (
+			days = 30
```
jenlij: why is this a constant rather than a configurable value?

internal/dbutil/nulltime.go
```diff
@@ -48,3 +49,18 @@ func (nt NullTime) Value() (driver.Value, error) {
 	}
 	return nt.Time, nil
 }
+
+// parseSqliteTimeString parses sqlite times string.
+// It tries to process value as string with timezone first,
+// then fallback to parsing as string without timezone.
```
navillasa: it looks like it's falling back to parsing as a date instead of as a timestamp without timezone- maybe the comment can be updated? (but maybe just make a tiny pr to fix or something since this has been open for a while)

satellite/accounting/db_test.go
```diff
@@ -113,37 +159,42 @@ func createBucketStorageTallies(projectID uuid.UUID) (map[string]*accounting.Buc
 	return bucketTallies, expectedTallies, nil
 }
 
-func createRollups(nodes storj.NodeIDList) (accounting.RollupStats, []time.Time) {
-	var dates []time.Time
+// make rollups and tallies for specified nodes and date range
```
ifraixedes:
```suggestion
// make rollups and tallies for specified nodes and date range.
```

PR #2859, opened 2019-08-22
streams: don't encrypt segment count

jtolds:
What: this change makes sure the count of segments is not encrypted.

Why: having the segment count encrypted just makes things hard for no reason - a satellite operator can figure out how many segments an object has by looking at the other segments in the database. but if a user has access but has lost their encryption key, they now can't clean up or delete old segments because they can't know how many there are without just guessing until they get errors. :(

Backwards compatibility: clients will still understand old pointers and will still write old pointers. at some point in the future perhaps we can do a migration for remaining old pointers so we can delete the old code. 

Please describe the tests: covered by existing tests
 

uplink/storage/streams/store.go
```diff
@@ -223,10 +230,10 @@ func (s *streamStore) upload(ctx context.Context, path Path, pathCipher storj.Ci
 			}
 
 			streamInfo, err := proto.Marshal(&pb.StreamInfo{
-				NumberOfSegments: currentSegment + 1,
-				SegmentsSize:     s.segmentSize,
-				LastSegmentSize:  sizeReader.Size(),
-				Metadata:         metadata,
+				DeprecatedNumberOfSegments: currentSegment + 1,
```
navillasa: why do we add 1 to currentSegment to get DeprecatedNumberOfSegments?

jtolds: currentSegment is the index of the current (the last) segment, so the index + 1 is the number of segments from zero indexing (first segment is 0). so this is just the number of segments and we're filling the deprecated field

jtolds:
test fixed @navillasa!

PR #2861, opened 2019-08-23
Green/random selection

navillasa:
https://storjlabs.atlassian.net/browse/V3-2363

What: 
- adds Select method to ReservoirService to select random paths for auditing

Why:
- random selection is not yet implemented for auditing 2.0

Please describe the tests:
 - TestRandomPaths: 
Hopefully appropriately recycled from original cursor test that tested pointer randomness. It should ensure randomness of paths selected for auditing.
 
### DIFF
--- a/satellite/audit/reservoirservice.go
+++ b/satellite/audit/reservoirservice.go
@@ -1,0 +1,117 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit
+
+import (
+	"context"
+	"math/rand"
+	"time"
+
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/sync2"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/metainfo"
+)
+
+// ReservoirService is a temp name for the service struct during the audit 2.0 refactor.
+// Once V3-2363 and V3-2364 are implemented, Service2 will replace the existing Service struct.
+type ReservoirService struct {
+	log *zap.Logger
+
+	nodesToSelect  int
+	PathsToAudit   []storj.Path
+	reservoirSlots int
+	Reservoirs     map[storj.NodeID]*Reservoir
+	rand           *rand.Rand
+
+	MetainfoLoop *metainfo.Loop
+	Loop         sync2.Cycle
+}
+
+// NewReservoirService instantiates Service2
+func NewReservoirService(log *zap.Logger, metaLoop *metainfo.Loop, config Config) (*ReservoirService, error) {
+	return &ReservoirService{
+		log: log,
+
+		nodesToSelect:  config.NodesToSelect,
+		reservoirSlots: config.Slots,
+		rand:           rand.New(rand.NewSource(time.Now().Unix())),
+
+		MetainfoLoop: metaLoop,
+		Loop:         *sync2.NewCycle(config.Interval),
+	}, nil
+}
+
+// Run runs auditing service 2.0
+func (service *ReservoirService) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	service.log.Info("audit 2.0 is starting up")
+
+	return service.Loop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+		pathCollector := NewPathCollector(service.reservoirSlots, service.rand)
+		err = service.MetainfoLoop.Join(ctx, pathCollector)
+		if err != nil {
+			service.log.Error("error joining metainfoloop", zap.Error(err))
+			return nil
+		}
+		service.Reservoirs = pathCollector.Reservoirs
+		return nil
+	})
+}
+
+// Select randomly selects segments to audit
+func (service *ReservoirService) Select(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+
+		service.PathsToAudit = []storj.Path{}
+		if service.Reservoirs != nil {
+			// todo: is it okay that pathsToAudit could end up being less than nodesToSelect?
+			for i := 0; i < service.nodesToSelect; i++ {
+				randomReservoir, err := service.GetRandomReservoir()
+				if err != nil {
+					return err
+				}
+				if randomReservoir == nil {
+					continue
+				}
+				randomPath := GetRandomPath(randomReservoir)
+				service.PathsToAudit = append(service.PathsToAudit, randomPath)
+			}
+		}
+		return nil
+	})
+}
+
+// GetRandomReservoir returns a random reservoir
+func (service *ReservoirService) GetRandomReservoir() (reservoir *Reservoir, err error) {
+	var src cryptoSource
+	rnd := rand.New(src)
+
+	if len(service.Reservoirs) == 0 {
+		return nil, Error.New("no reservoirs available")
+	}
+
+	randomIndex := rnd.Intn(len(service.Reservoirs))
+	for nodeID := range service.Reservoirs {
+		if randomIndex == 0 {
+			return service.Reservoirs[nodeID], nil
+		}
+		randomIndex--
+	}
+	return nil, nil
+}
+
+// GetRandomPath returns a random path
+func GetRandomPath(reservoir *Reservoir) (path storj.Path) {
+	var src cryptoSource
+	rnd := rand.New(src)
+
+	randomIndex := rnd.Intn(len(reservoir.Paths))
+	return reservoir.Paths[randomIndex]
+}
--- a/satellite/audit/selection_test.go
+++ b/satellite/audit/selection_test.go
@@ -1,0 +1,93 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit_test
+
+import (
+	"math"
+	"reflect"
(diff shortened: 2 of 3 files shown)


satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+		service.Reservoirs = pathCollector.Reservoirs
+		return nil
+	})
+}
+
+// Select randomly selects segments to audit
+func (service *ReservoirService) Select(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	return service.Loop.Run(ctx, func(ctx context.Context) (err error) {
```
egonelbre: Why is it running a loop?

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+		defer mon.Task()(&ctx)(&err)
+
+		service.PathsToAudit = []storj.Path{}
+		if service.Reservoirs != nil {
+			// todo: is it okay that pathsToAudit could end up being less than nodesToSelect?
+			for i := 0; i < service.nodesToSelect; i++ {
+				randomReservoir, err := service.GetRandomReservoir()
```
egonelbre: Why does it need to select randomly if we are going to audit all of the reservoirs anyways?

satellite/audit/reservoirservice.go
navillasa: Ahh.. I guess it doesn't need to be a loop. It just needs to happen when reservoirs have been created already. Do you know a better way to do this?

satellite/audit/reservoirservice.go
navillasa: I think according to the design doc we had decided to randomly select reservoirs to choose paths to audit, instead of auditing all reservoirs: https://github.com/storj/storj/blob/master/docs/design/audit-v2.md#design. I thought we were sticking with the idea of auditing as "spot checking" and not thoroughly checking every reservoir. Maybe @jtolds or @wthorp might have more thoughts on this?

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/sync2"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/metainfo"
+)
+
+// ReservoirService is a temp name for the service struct during the audit 2.0 refactor.
+// Once V3-2363 and V3-2364 are implemented, Service2 will replace the existing Service struct.
+type ReservoirService struct {
```
mobyvb: Why not go ahead and call this `Service2`? Or change the comment to say `ReservoirService` instead of `Service2`?

satellite/audit/service.go
```diff
@@ -32,7 +31,8 @@ type Config struct {
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"25s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
 
-	Slots int `help:"number of reservoir slots allotted for nodes, currently capped at 2" default:"1"`
+	Slots         int `help:"number of reservoir slots allotted for nodes, currently capped at 2" default:"1"`
+	NodesToSelect int `help:"number of nodes' reservoirs to randomly select segments from" default:"100"`
```
mobyvb: I'd call this `NumNodesToSelect` or `NodeSelectCount` or something like that. The name `NodesToSelect` does not imply that it is a number to me.

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+		reservoirSlots: config.Slots,
+		rand:           rand.New(rand.NewSource(time.Now().Unix())),
+
+		MetainfoLoop: metaLoop,
+		Loop:         *sync2.NewCycle(config.Interval),
+	}, nil
+}
+
+// Run runs auditing service 2.0
+func (service *ReservoirService) Run(ctx context.Context) (err error) {
```
mobyvb: Rather than `Run` and `Select` I think you could call them `RunPopulateReservoirs` and `RunReservoirAudits` or something. I think `Select` does not accurately explain what is being done inside it.

satellite/audit/reservoirservice.go
mobyvb: If you keep it in a loop, it should be a separate loop than the one being used to hook into the metainfo and create the path collector - I think a single sync2 cycle is not designed to handle two separate calls to `Run` like this.

Maybe not the best way to solve it, but you could probably get rid of the loop by doing the following:
* Create a channel `gotReservoirs chan struct{}`. At the very beginning of this function (`Select`), wait with `<-gotReservoirs`. After that line, start a for loop that iterates and runs audits. At the end of each loop iteration in `Run` above, after `service.Reservoirs = ...` do `gotReservoirs <- struct{}{}` to unblock `Select`.

There may be a better way to design it, but that seems like an easy solution.

satellite/audit/reservoirservice.go
mobyvb: Yeah, if I recall correctly, we had a meeting where this question came up. The general idea was that random selection is easier to implement and should give more reliable behavior. In the long term, I do not think there is a guarantee that we will try to audit all of the reservoirs, and auditing them in a specific order could result in us being unexpectedly biased. This bias is not present with random selection.

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+func (service *ReservoirService) GetRandomReservoir() (reservoir *Reservoir, err error) {
+	var src cryptoSource
+	rnd := rand.New(src)
+
+	if len(service.Reservoirs) == 0 {
+		return nil, Error.New("no reservoirs available")
+	}
+
+	randomIndex := rnd.Intn(len(service.Reservoirs))
+	for nodeID := range service.Reservoirs {
```
mobyvb: you could do `for _, reservoir := range service.Reservoirs {` then you could just `return reservoir` instead of `return service.Reservoir[nodeID]`

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+		if randomIndex == 0 {
+			return service.Reservoirs[nodeID], nil
+		}
+		randomIndex--
+	}
+	// todo: is it okay to return nil, nil here?
```
mobyvb: there should be a way to write this function so that `nil, nil` is never returned. It is actually never going to return `nil, nil` as it is written. The for loop should always return something since `randomIndex` is always less than `len(service.Reservoirs)`. 

One way to rewrite could be something like:
```
var nodeIDList storj.NodeIDList
for nodeID := range service.Reservoirs {
    nodeIDList = append(nodeIDList, nodeID)
}
randomIndex := rnd.Intn(len(nodeIDList))
selectedID := nodeIDList[randomIndex]
return service.Reservoirs[selectedID], nil
```

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil, nil
+}
+
+// GetRandomPath returns a random path
+func GetRandomPath(reservoir *Reservoir) (path storj.Path) {
```
mobyvb: Maybe it makes sense to put this function on the reservoir itself so you can call `reservoir.GetRandomPath()`?

satellite/audit/selection_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/internal/testplanet"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/audit"
+)
+
+func TestRandomPaths(t *testing.T) {
```
mobyvb: Could you add a comment explaining what this test is doing?

satellite/audit/reservoirservice.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return nil
+	})
+}
+
+// Select randomly selects segments to audit
+func (service *ReservoirService) Select(ctx context.Context, reservoirs map[storj.NodeID]*Reservoir) (pathsToAudit []storj.Path, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if len(reservoirs) != 0 {
```
egonelbre: Using `for nodeID, reservoir := range reservoirs {` would roughly achieve the same goal, however that's somewhat implementation specific.

The other would be to put all of them into a `[]*Reservoir` and then shuffle the list. No need for the random selection one-by-one.

egonelbre: rand.Shuffle should be fine

egonelbre: I mean, the point of having multiple paths is to audit all of them, otherwise wouldn't we just select one path to begin with. My idea was to shuffle the reservoirs themselves, not the paths.

egonelbre:
I think the discussion has gotten off-track.

Reservoir sampling has a goal for ensuring that every node gets selected for audits. Building reservoirs ensures that we have at least one audit per node.

We also want to have a queue for audits for nodes since we may finish audits faster than we can create the reservoirs. Hence, we build non-zero reservoirs.

Ideally, we would audit all the reservoirs collected and then create new ones. Due to async nature of things, it one might finish before the other.

Also, we would want to have multiple audit workers, where they run in separate processes. This implies we need to communicate this information.

One way to easily achieve all of these goals is to:
1. create a queue for the segments that will be audited
2. build reservoirs
3. insert reservoirs row major order and in random node id order, into the queue
4. worker regularly pulls work from the queue

PR #2862, opened 2019-08-23
storagenode/reputation: add disqualified flag

rikysya:
What: Add disqualified flag to `storagenode` `reputation` db 

Why:

Please describe the tests:
 - Test 1: Tests `insert` and `get` operations as well check that they used right column order
 - Test 2: Tests `GetAll` operation with column order check
 

storagenode/storagenodedb/database.go
```diff
@@ -653,6 +653,29 @@ func (db *DB) Migration() *migrate.Migration {
 					`INSERT INTO piece_space_used (total) select ifnull(sum(piece_size), 0) from pieceinfo_`,
 				},
 			},
+			{
+				Description: "Add disqualified field to reputation",
+				Version:     18,
+				Action: migrate.SQL{
+					`DROP TABLE reputation;`,
```
mobyvb: Why drop and recreate the table rather than adding a single column?

rikysya: I want `updated_at` to be the last column) May be that’s overkill. But this table is not being used at the moment, so recreating it should not be a problem

storagenode/storagenodedb/reputation.go
```diff
@@ -88,11 +96,40 @@ func (db *reputationDB) Get(ctx context.Context, satelliteID storj.NodeID) (_ *r
...
 	)
+
+	return &stats, ErrReputation.Wrap(err)
+}
+
+// All retrieves all stats from DB
+func (db *reputationDB) All(ctx context.Context) (_ []reputation.Stats, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	rows, err := db.QueryContext(ctx, `SELECT * FROM reputation`)
```
mobyvb: I would feel more comfortable with something like `SELECT satellite_id, uptime_success_count, ... FROM reputation` rather than `SELECT * FROM reputation` because what if the order that you put arguments into `rows.Scan` below does not match up with the column order returned by using `*`?

I think either you should have a test to verify that this function works as expected, or update the query to be specific about the columns and order of columns returned.

rikysya: Yeah, agree, good point

storagenode/reputation/reputation.go
```diff
@@ -16,6 +16,8 @@ type DB interface {
 	Store(ctx context.Context, stats Stats) error
 	// Get retrieves stats for specific satellite
 	Get(ctx context.Context, satelliteID storj.NodeID) (*Stats, error)
+	// All retrieves all stats from DB
+	All(ctx context.Context) ([]Stats, error)
```
egonelbre: This sounds like a bad idea to query the full database.

rikysya: @egonelbre num of rows for this query should be equal to `len(trustpool)`, so instead of querying each `reputation` in `for loop range trustpool` I created this query.

storagenode/storagenodedb/testdata/sqlite.v19.sql
```diff
@@ -0,0 +1,201 @@
...
+    PRIMARY KEY (satellite_id, timestamp)
+);
+
+CREATE TABLE piece_space_used (
+    total INTEGER NOT NULL,
+	satellite_id BLOB
+);
+CREATE UNIQUE INDEX idx_piece_space_used_satellite_id ON piece_space_used(satellite_id);
+
+DROP TABLE reputation;
```
ifraixedes: Is this needed here?

egonelbre: yeah, this shouldn't be needed

rikysya: @ifraixedes  there are 2 `create table` statements for `reputation` table: one without `disqualified` flag, and the second one recreating table with this flag after table drop

rikysya: @egonelbre so your point is that I can delete `table drop` and second `table creation` from v.19 testdata file and just directly modify the first `table creation`?

egonelbre: Yes, the testdata should setup the schema as it is and not do any migrations.

PR #2866, opened 2019-08-23
docs/design: Graceful Exit

ethanadams:
What: When a Storage Node wants to leave the network but does not want to lose their escrow we need to have a mechanism for them to exit the network “gracefully”.

Replaces https://github.com/storj/storj/pull/2734

Why: Give Storage Nodes a mechanism to leave the network while receiving their escrows, and reduce repair caused by node churn on satellites.

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/design/storagenode-graceful-exit-overview.md
```diff
@@ -0,0 +1,164 @@
...
+
+    field initiated_at         timestamp ( updateable )
+    field finished_at          timestamp ( updateable )
+    field starting_disk_usage  int64 not null
+    field bytes_deleted        int64 not null
+    field completion_receipt   blob
+)
+```
+
+### Rationale
```
egonelbre: Rationale should be second level heading and let's move it after satellite database changes

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+The satellite will send a `TransferPiece`, which contains sufficient information to upload the piece to a replacement node.
+
+The storage node will start a new piece upload to the replacement node similar to uplink. It will use `uplink/piecestore.Upload`. Once uploaded, the storage node will verify the piece hash sent by the replacement node corresponds to the piece hash stored in the database.
+
+#### Verifying transfer
+
+The storage node sends the piece hash, order limit, original "uploader" signature, and piece hash signed by the replacement node to the satellite for confirmation.
+The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limits signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send `ExitFailed`.
+
+![Transfer Sequence](./images/storagenode-graceful-exit-transfer-sequence.svg)
```
bryanchriswhite: where is this image?

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Updating the Segment / Pointer
+
+When the upload has been verified the satellite will use `CompareAndSwap` to only switch the appropriate node without changing others. During the upload there may have been other things that happened:
```
bryanchriswhite: I find this section a bit difficult to follow. Am I understanding correctly that these list items are describing how we use compare-and-swap and that it's the node ID in the segment/pointer that we're comparing-and-swapping?

>Segment / pointer was reuploaded. We'll ignore the newly uploaded piece.

Is this the scenario you're describing and/or is this how re-upload would work?
```
client     "re-upload" storagenode    "transfer" storagenode    "original" storagenode           satellite
   |                 |                          |                         |                          |
   |                 |                          |     transfer piece      |                          |
   |                 |                          |            /------------A                          |
   |                 |                          B-----------/             |                          |
   |                 |                          |                         |                          |
   |                 |                          |      piece hash         |                          |
   |                 |                          |-----------\             |                          |
   |                 |                          |            \------------|                          |
   |                 |                          |                         |                          |
   |                 |                          |                         | begin piece verification |
   |                 |                          |                         |------------\             |
   |                 |                          |                         |             \------------C
   |                 |                          |                         |                          |
   |                 |                          |                         |                          |
 . |.  .  .  .  .  . |.  .  .  .  .  .  .  .  . |.  .  .  .  .  .  .  .  .| .  .  .  .  .  .  .  .  .| .
   |                 |                          |                         |    begin re-upload       |
 . |-----------------------------------------------------------------------------------\             | .
   |                 |                          |                         |             \------------D
 . |                 |                          |                         |                          | .
   |/\/\/\/\/\/\/\/\/|\/\/\/\/\/\/\/\/\/\/\/\/\/|\/\/\/\/\/\/\/\/\/\/\/\/\|/\/\/\/\/\/\/\/\/\/\/\/\/\|
 . |                 |                          |                         |                          | .
   | re-upload piece |                          |                         |                          |
 . |--------\        |                          |                         |                          | .
   |         \-------E                          |                         |                          |
 . |                 |                          |                         |                          | .
   |/\/\/\/\/\/\/\/\/|\/\/\/\/\/\/\/\/\/\/\/\/\/|\/\/\/\/\/\/\/\/\/\/\/\/\|/\/\/\/\/\/\/\/\/\/\/\/\/\|
 . |.  .  .  .  .  . |.  .  .  .  .  .  .  .  . |.  .  .  .  .  .  .  .  .| .  .  .  .  .  .  .  .  .| .
   |                 |                          |                         |                          |
   |                 |                          |                         |                          |
   |                 |                          |                         | verification complete    |
   |                 |                          |                         |             /------------|
   |                 |                          |                         F------------/             |
```

Is "newly uploaded piece" the one that was uploaded to the "transfer" node at `B` or the "re-upload" node at `E`?

> Segment / pointer pieces were changed, in that case we load the new one, check whether the exiting storage node ID still needs to be replaced. If it does, then we retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the newly uploaded piece, because one of them needs to be discarded.

Again, "new one" is a bit ambiguous in my opinion. I'm assuming that since is's possible for the exiting node to still need replacement but then I wasn't able to follow the rest of the logic.

>Segment / pointer was deleted or is being deleted. We'll ignore the newly uploaded piece.
Same question.

Maybe it would be easier to put these in terms of whether we will or won't swap with the transfer node.

docs/design/storagenode-graceful-exit-ui.md
```diff
@@ -0,0 +1,53 @@
...
+The storage node operator needs:
+- a way to initiate graceful exit, 
+- a way to monitor graceful exit progress, and
+- terminate graceful exit without escrow.
+
+## Design
+
+Add a command `storagenode exit-satellite` to initiate a Graceful Exit.
+
+The command should present a list of satellites to exit. The user needs to type the satellite domain name to start exiting. Only satellites that we haven't exited will be displayed.
```
bryanchriswhite: ... Only satellites that ~~we~~ haven't exited will be displayed.

docs/design/storagenode-graceful-exit-ui.md
```diff
@@ -0,0 +1,53 @@
...
+
+A storage node operator needs the ability to request a Graceful Exit on a per satellite basis.
+This document describes how the storage node operator manages the Graceful Exit process.
+
+## Background
+
+The storage node operator needs:
+- a way to initiate graceful exit, 
+- a way to monitor graceful exit progress, and
+- terminate graceful exit without escrow.
```
bryanchriswhite: Is this an all-or-nothing deal? Do we have plans and/or can we pay nodes back the difference between the amount in escrow and the cost of repairing an ungraceful exit?

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -52,9 +52,9 @@ The satellite verifies that the original piece hash matches the replacement piec
...
 
-- Segment / pointer was reuploaded. We'll ignore the newly uploaded piece.
-- Segment / pointer pieces were changed, in that case we load the new one, check whether the exiting storage node ID still needs to be replaced. If it does, then we retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the newly uploaded piece, because one of them needs to be discarded.
-- Segment / pointer was deleted or is being deleted. We'll ignore the newly uploaded piece.
+- Segment / pointer was reuploaded. We'll ignore piece transferred for the exit.
+- Segment / pointer pieces were changed, in that case we load the pointer, check whether the exiting storage node ID still needs to be replaced. If it does, then we retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
+- Segment / pointer was deleted or is being deleted. We'll ignore the piece transferred for the exit.
 
 After changing the segment / pointer, we'll delete the piece from transfer queue, update progress, and send a `DeletePiece` to the exiting node.
```
bryanchriswhite: In the cases above where we're ignoring the transferred piece, are we sending delete requests for those pieces?

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limits signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send `ExitFailed`.
+
+![Transfer Sequence](./images/storagenode-graceful-exit-transfer-sequence.svg)
+
+#### Updating the Segment / Pointer
+
+When the upload has been verified the satellite will use `CompareAndSwap` to only switch the appropriate node without changing others. During the upload there may have been other things that happened:
+
+- Segment / pointer was reuploaded. We'll ignore piece transferred for the exit.
+- Segment / pointer pieces were changed, in that case we load the pointer, check whether the exiting storage node ID still needs to be replaced. If it does, then we retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
```
bryanchriswhite:
```suggestion
- Segment / pointer pieces were changed, in that case we load the pointer and retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
```

bryanchriswhite: Isn't that the same thing?

docs/design/storagenode-graceful-exit-ui.md
bryanchriswhite: :+1:

docs/design/storagenode-graceful-exit-overview.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Background
+
+Storage nodes may want to leave the network. Taking a storage node directly offline would mean that the satellite needs to repair the lost pieces. To avoid this we need a process to transfer the pieces to other nodes before taking the exiting node offline. We call this process Graceful Exit. Since this process can take time we need an overview of the nodes exiting and the progress.
+
+A storage node may also want to exit a single satellite rather than the whole network. For example, the storage node doesn't have as much bandwidth available or a satellite's payment structure has changed.
+
+The process must also consider that the storage nodes may have limited bandwidth available or time to exit. This means the process should prioritize pieces with lower durability.
+
+A storage node will no longer participate in repairs or uploads once a Graceful Exit has been initiated. In addition, bandwidth used for piece transfers should not count against bandwidth used.
```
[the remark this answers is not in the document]
kaloyan-raev: Yes, this should be documented here. The exiting node should still respond to downloads and audits as usually for the pieces that have not been transferred yet. Failing to do so may lead to DQ.

docs/design/storagenode-graceful-exit-overview.md
```diff
@@ -0,0 +1,162 @@
...
+    - starts gathering of pieces that need to be transferred.
+3. Satellite finishes gathering pieces that need to be transferred.
+4. Storage node keeps polling satellite for pieces to transfer.
+5. When the satellite doesn't have any more pieces to transfer, it will respond with a completion receipt.
+6. Storage node stores completion information in the database.
+7. Satellite Operator creates a report for exited storage nodes in order to release escrows.
+
+For all of these steps we need to ensure that we have sufficient monitoring.
+
+When a Graceful Exit has been started, it must either succeed or fail. The escrow, held from storage node, will be released only on success. We will call the failure scenario an ungraceful exit.
```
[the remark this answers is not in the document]
kaloyan-raev: IMHO the escrow system is out of scope for this design doc.

docs/design/storagenode-graceful-exit-pieces.md
```diff
@@ -0,0 +1,42 @@
...
+## Implementation
+
+1. Add method for finding exiting nodes to overlay.
+2. Implement transfer queue for pieces.
+3. Implement gexit.Service.
+4. Update satellite to ignore exiting storage nodes for repairs and uploads.
+
+## Open issues (if applicable)
+
+- Can pieces with really high durability can be ignored?
```
kaloyan-raev: I think this is against the economics of the graceful exit. We want the exiting nodes to pay the price for future repairs due to leaving the network. Even if a piece has high durability, it still makes sense to transfer it to a new node. Otherwise, the durability may degrade over time due to gracefully exiting nodes, which will trigger the repair sooner, which will increase the cost on the satellite. A successful graceful exit should not have any negative impact on durability.

docs/design/storagenode-graceful-exit-pieces.md
```diff
@@ -0,0 +1,42 @@
...
+The metainfo loop `Join` guarantees the observer will only receive events at the beginning of the next loop. Hence, one complete metainfo loop is sufficient to collect all the pieces for a given node. 
+
+## Implementation
+
+1. Add method for finding exiting nodes to overlay.
+2. Implement transfer queue for pieces.
+3. Implement gexit.Service.
+4. Update satellite to ignore exiting storage nodes for repairs and uploads.
+
+## Open issues (if applicable)
```
kaloyan-raev: Can we clarify what happens if the satellite crashes or is restarted in the middle of the metainfo loop collecting the pieces to be transferred? We have to ensure that the loop will run again and won't duplicate pieces in the transfer queue.

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+The `worker` should continue to poll the satellite at a configurable interval until it returns pieces to transfer.  
+
+The storage node should concurrently transfer pieces returned by the satellite. The storage node should send a `TransferSucceeded` message as pieces are successfuly transfered. The Storage node should send a `TransferFailed`, with reason, on failure.
+
+The satellites should set the `finished_at` on success, and respond with a `DeletePiece` message. Otherwise set the `failed_at` and `failure_status_code` for reprocessing.
+
+The satellite should respond with a `ExitCompleted` message when all pieces have finished processing. 
+The storage node should store the completion receipt, stop transfer processing, and remove all remaining pieces for the satellite.
+
+When the storage node has failed too many transfers or has sent incorrect data, the satellite will send `ExitFailed`. This indicates that the process has ended ungracefully.
```
kaloyan-raev: I think the satellite should send a completion receipt in this case too, with failed status and reason for the failure.

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+
+- Segment / pointer was reuploaded. We'll ignore piece transferred for the exit.
+- Segment / pointer pieces were changed, in that case we load the pointer and retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
+- Segment / pointer was deleted or is being deleted. We'll ignore the piece transferred for the exit.
+
+After changing the segment / pointer, we'll delete the piece from transfer queue, update progress, and send a `DeletePiece` to the exiting node.
+
+#### Handling transfer failure
+
+If a storage node report a failure to the satellite if it fails to upload to the replacement node, or the replacement node returns a bad piece hash. The satellite will track the failures and retry the transfer to another node.
```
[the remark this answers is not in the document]
kaloyan-raev: Hmm, not really... My understanding is that the satellite will send `ExitFailed` if the exiting nodes send the satellite info that looks like cheating. Here we are talking the transfer may fail because the replacement node returns a hash that the exiting node does not agree to.

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+            bytes private_key;
+            // addressed_order_limit contains the new piece id.
+            AddressedOrderLimit addressed_order_limit;              
+        }
+
+        message DeletePiece {
+            bytes piece_id;
+        }
+
+        message Completed {
```
kaloyan-raev: Should we have `Failed` message too?

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+            // when everything is completed
+            bytes exit_complete_signature;
+        }
+    }
+}
+```
+
+
+## Open issues (if applicable)
```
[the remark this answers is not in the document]
kaloyan-raev: My opinion is that all piece must be transferred regarding the durability level. The durability level should matter only to prioritize which pieces should be transferred first.

docs/design/storagenode-graceful-exit-protocol.md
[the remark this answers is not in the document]
egonelbre: Here, the exiting node should report it as "transfer failed". The other part is about exiting node sending a piece hash which is invalid.

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+
+if !pieces collected {
+    send NotReady
+    return
+}
+
+inprogress pieces
+more pieces := true
+
+when storage node prematurely exits
```
kaloyan-raev: How does the satellite know if the storage node exited prematurely?

docs/design/storagenode-graceful-exit-ui.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Design
+
+Add a command `storagenode exit-satellite` to initiate a Graceful Exit.
+
+The command should present a list of satellites to exit. The user needs to type the satellite domain name to start exiting. Only satellites that haven't been exited will be displayed.
```
kaloyan-raev: The command should display a detailed explanation about the consequences of starting the graceful exit.

docs/design/storagenode-graceful-exit-pieces.md
kaloyan-raev: If this is the case then it makes sense. However, I'd expect this to be quite a rare case. Does the monitoring currently report any segment with such high durability?

kaloyan-raev: OK. I agree then.

docs/design/storagenode-graceful-exit-protocol.md
ethanadams: Agreed, Updated.

docs/design/storagenode-graceful-exit-overview.md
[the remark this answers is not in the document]
ethanadams: Updated

docs/design/storagenode-graceful-exit-pieces.md
```diff
@@ -0,0 +1,42 @@
+# Storage Node Graceful Exit - Gathering Pieces
+
+## Abstract
+
+During [Graceful Exit](storagenode-graceful-exit-overview.md), a satellite needs to find pieces to be transferred from the exiting storage node.
```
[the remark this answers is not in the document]
ethanadams: Updated

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+The satellite picks a piece from the transfer queue and verifies whether it still needs to be transferred. During gathering or the Graceful Exit process the segment may have been changed or deleted.
+
+The satellite will send a `TransferPiece`, which contains sufficient information to upload the piece to a replacement node.
+
+The storage node will start a new piece upload to the replacement node similar to uplink. It will use `uplink/piecestore.Upload`. Once uploaded, the storage node will verify the piece hash sent by the replacement node corresponds to the piece hash stored in the database.
+
+#### Verifying transfer
+
+The storage node sends the piece hash, order limit, original "uploader" signature, and piece hash signed by the replacement node to the satellite for confirmation.
+The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limits signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send `ExitFailed`.
```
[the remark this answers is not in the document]
ethanadams: Updated.  On failed exit, the repairer will correct the pointers.

docs/design/storagenode-graceful-exit-protocol.md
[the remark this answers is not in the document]
ethanadams: Updated

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Updating the Segment / Pointer
+
+When the upload has been verified the satellite will use `CompareAndSwap` to only switch the appropriate node without changing others. During the upload there may have been other things that happened:
+
+- Segment / pointer was reuploaded. We'll ignore piece transferred for the exit.
+- Segment / pointer pieces were changed, in that case we load the pointer and retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
+- Segment / pointer was deleted or is being deleted. We'll ignore the piece transferred for the exit.
+
+After changing the segment / pointer, we'll delete the piece from transfer queue, update progress, and send a `DeletePiece` to the exiting node.
```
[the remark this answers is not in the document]
ethanadams: As discussed, leaving as is.

docs/design/storagenode-graceful-exit-pieces.md
ethanadams: Updated to remove the exiting node when durability > optimal.

docs/design/storagenode-graceful-exit-protocol.md
ethanadams: Updated

docs/design/storagenode-graceful-exit-ui.md
ethanadams: Updated

docs/design/storagenode-graceful-exit-protocol.md
ethanadams: Updated

docs/design/storagenode-graceful-exit-protocol.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Design
+
+The storage node has a Graceful Exit service, which ensures that the process is completed. It queries `graceful_exit_status` to find unfinished exits and starts a per satellite `worker` if needed.
+
+The `worker` polls the satellite, requesting pieces to transfer. The satellite will initiate a Graceful Exit if not already initiated. When first initiated, the satellite will start gathering pieces for the exiting node and return `NotReady`. The satellite will continue to return `NotReady` until the piece gathering process has completed.
+
+The `worker` should continue to poll the satellite at a configurable interval until it returns pieces to transfer.  
+
+The storage node should concurrently transfer pieces returned by the satellite. The storage node should send a `TransferSucceeded` message as pieces are successfuly transfered. The Storage node should send a `TransferFailed`, with reason, on failure.
```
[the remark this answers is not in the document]
ethanadams: Depending on the failure, it will get reprocessed when the satellite iterates over the transfer queue.

docs/design/storagenode-graceful-exit-protocol.md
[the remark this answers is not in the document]
egonelbre: The issue is that when we take too much time to transfer pieces people might start hard-disconnecting, because it's not worth the escrow. This could end up storage nodes not bothering with graceful exit at all.

Ideally, yes, we would transfer all pieces.

docs/design/storagenode-graceful-exit-protocol.md
egonelbre: It gets io.EOF when trying to read the next message.

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+The storage node will start a new piece upload to the replacement node similar to uplink. It will use `uplink/piecestore.Upload`. Once uploaded, the storage node will verify the piece hash sent by the replacement node corresponds to the piece hash stored in the database.
+
+#### Verifying transfer
+
+The storage node sends the piece hash, order limit, original "uploader" signature, and piece hash signed by the replacement node to the satellite for confirmation.
+
+The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limit's signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send an `ExitFailed` message.
+
+![Transfer Sequence](./images/transfer-sequence.svg)
```
bryanchriswhite: I still don't see the `Transfer Sequence` image. Is this to be expected?

image

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Updating the Segment / Pointer
+
+When the upload has been verified the satellite will use `CompareAndSwap` to only switch the appropriate node without changing others. During the upload there may have been other things that happened:
+
+- Segment / pointer was reuploaded. We'll ignore piece transferred for the exit.
+- Segment / pointer pieces were changed, in that case we load the pointer and retry `CompareAndSwap`. When the storage node at that piece number has changed, we'll ignore the piece transferred for the exit, because one of them needs to be discarded.
+- Segment / pointer was deleted or is being deleted. We'll ignore the piece transferred for the exit.
+
+After processing the segment / pointer, we'll delete the piece from transfer queue, update progress, and send a `DeletePiece` to the exiting node.
```
bryanchriswhite: Do we want to add something about deleting any unused transferred pieces?

docs/design/storagenode-graceful-exit/protocol.md
bryanchriswhite: see /storj/storj/raw/master/docs/design/storagenode-graceful-exit/images/transfer-sequence.svg?sanitize=true

ethanadams: Fixed.

docs/design/storagenode-graceful-exit/protocol.md
ethanadams: This should be a fairly rare occurrence.  The thought was to let GC clean it up.

docs/design/storagenode-graceful-exit-protocol.md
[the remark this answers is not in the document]
mobyvb: In this case we should make sure to include the process in the design doc. We could have a configurable value `--graceful-exit.max-durability` or something that determines the cutoff where we decide we no longer care about transferring pieces on segments above a certain durability. Then we just configure it to `rs.Total` so that we transfer everything and we can lower it if we decide that takes too much time.

docs/design/storagenode-graceful-exit-pieces.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/overview.md
```diff
@@ -0,0 +1,164 @@
+# Storage Mode Graceful Exit Overview
```
VinozzZ: nit: typo `Mode` -> `Node`

docs/design/storagenode-graceful-exit-protocol.md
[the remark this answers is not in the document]
ethanadams: Added as a future item in the overview.

docs/design/storagenode-graceful-exit/overview.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/overview.md
```diff
@@ -0,0 +1,169 @@
...
+
+Storage nodes may want to leave the network. Taking a storage node directly offline would mean that the satellite needs to repair the lost pieces. To avoid this we need a process to transfer the pieces to other nodes before taking the exiting node offline. We call this process Graceful Exit. Since this process can take time we need an overview of the nodes exiting and the progress.
+
+A storage node may also want to exit a single satellite rather than the whole network. For example, the storage node doesn't have as much bandwidth available or a satellite's payment structure has changed.
+
+The process must also consider that the storage nodes may have limited bandwidth available or time to exit. This means the process should prioritize pieces with lower durability.
+
+A storage node will no longer participate in repairs or uploads once a Graceful Exit has been initiated. In addition, bandwidth used for piece transfers should not count against bandwidth used.
+
+The exiting node must continue to respond to audit and download requests.
```
mobyvb: And retain requests - we do not want it to keep pieces that might have already been deleted from the satellite it is exiting from.

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+The satellite should respond with an `ExitCompleted` message when all pieces have finished processing. 
+
+If the storage node has failed too many transfers overall, failed the same piece over a certain threshold, or has sent incorrect data, the satellite will send an `ExitFailed` message. This indicates that the process has ended ungracefully.
+
+In either case, the storage node should store the completion / failure receipt, stop transfer processing, and remove all remaining pieces for the satellite.
+
+### Transferring a Piece
+
+The satellite picks a piece from the transfer queue and verifies whether it still needs to be transferred. During gathering or the Graceful Exit process the segment may have been changed or deleted.
```
navillasa: how does it verify whether it still needs to be transferred?

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+The satellite will send a `TransferPiece`, which contains sufficient information to upload the piece to a replacement node.
+
+The storage node will start a new piece upload to the replacement node similar to uplink. It will use `uplink/piecestore.Upload`. Once uploaded, the storage node will verify the piece hash sent by the replacement node corresponds to the piece hash stored in the database.
+
+#### Verifying transfer
+
+The storage node sends the piece hash, order limit, original "uploader" signature, and piece hash signed by the replacement node to the satellite for confirmation.
```
navillasa: would it add too much complexity or not make sense for the replacement storage node to send the piece hash back to the satellite too for confirmation? I guess I'm wondering if it would be possible for the original storage node to tell the satellite that the transfer occurred successfully when it actually didn't. maybe the piece hash signed by the replacement node solves this though?

docs/design/storagenode-graceful-exit/protocol.md
mobyvb: I think it would check the metainfo, and make sure that the segment still exists and the piece is still in the segment (in case that changed between when it was added to the transfer queue and this point)

docs/design/storagenode-graceful-exit/protocol.md
mobyvb: If the replacement node signs the piece hash and the satellite can verify that, it should be safe to assume the transfer was successful even without direct communication with the transfer node.

docs/design/storagenode-graceful-exit/overview.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/ui.md
```diff
@@ -0,0 +1,57 @@
...
+## Implementation
+
+- Add `storagenode exit-satellite` command to storagenode CLI, which calls `gexit.Service.InitiateExit`.
+	- Once initiated [protocol for transferring pieces](protocol.md) should start.
+- Add `storagenode exit-status` command to storagenode CLI. This returns completion status as described above.
+- TODO: terminating graceful exit?
+
+## Open issues (if applicable)
+
+- Should we be able to terminate graceful exit? If we do not provide the feature, storage node operator might try to do this manually, breaking the whole node.
```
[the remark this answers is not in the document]
ethanadams: I think this feature is still up in the air, but If we allow the terminate, it would be an ungraceful exit.

docs/design/storagenode-graceful-exit/protocol.md
navillasa: ok awesome

docs/design/storagenode-graceful-exit/protocol.md
navillasa: I guess it just sounded a little redundant to me to first find the pieces that need transferring via the metainfoloop, add the transfer segment to the queue, then verify that they're in the metainfo again? but maybe like a larger chunk of time is expected to pass between getting added to the transfer queue and the transfer actually happening.

docs/design/storagenode-graceful-exit/pieces.md
```diff
@@ -0,0 +1,45 @@
...
+
+Graceful Exit contains a process that moves existing pieces from one storage node to another. To accomplish this we need a list of pieces that need to be transferred.
+
+Pieces with lower durability have higher importance to be transferred.
+
+## Design
+
+We need a service on the satellite that finds pieces in the metainfo database that need to be transferred. We'll call this service `gexit.Service` or Graceful Exit service.
+
+The service starts by asking overlay for all exiting nodes where `nodes.exit_loop_completed` is false.
```
kaloyan-raev: `nodes.exit_loop_completed` is of type `timestamp`. So the check should be if it is `NULL`, not `false`.

docs/design/storagenode-graceful-exit/pieces.md
```diff
@@ -0,0 +1,45 @@
...
+
+Then joins a metainfo loop to iterate over all segments. For any segment that contains nodes that are exiting it will add an entry to a queue (if durability <= optimal). We call this the transfer queue. If durability > optimal, we remove the exiting node from the segment / pointer.
+
+
+The transfer queue is stored in database. We will need batching when inserting to database to avoid excessive load.
+
+Once metainfo loop has completed successfully it updates `nodes.exit_loop_completed` to indicate the storage nodes is ready for transferring.
+
+In the event that the satellite does not complete the metainfo loop (e.g. satellite is shutodwn), the service will re-enter the metainfo loop for all exiting nodes where `nodes.exit_loop_completed` is false. Pieces that already exist in the queue should not get duplicated.
```
kaloyan-raev: typo: "shutodwn"

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+In either case, the storage node should store the completion / failure receipt, stop transfer processing, and remove all remaining pieces for the satellite.
+
+### Transferring a Piece
+
+The satellite picks a piece from the transfer queue and verifies whether it still needs to be transferred. During gathering or the Graceful Exit process the segment may have been changed or deleted.
+
+The satellite will send a `TransferPiece`, which contains sufficient information to upload the piece to a replacement node.
+
+The storage node will start a new piece upload to the replacement node similar to uplink. It will use `uplink/piecestore.Upload`. Once uploaded, the storage node will verify the piece hash sent by the replacement node corresponds to the piece hash stored in the database.
```
littleskunk: > The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limit's signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send an `ExitFailed` message.

The storage node has to verify the signature as well otherwise my bad storage node will return a valid hash but a bad signature. If the storage node doesn't notice it will fail graceful exit for not doing anything wrong.

docs/design/storagenode-graceful-exit/pieces.md
kaloyan-raev: Here too: `nodes.exit_loop_completed` should be checked for `NULL`, not `false`.

docs/design/storagenode-graceful-exit/reports.md
```diff
@@ -0,0 +1,45 @@
...
+This report should contain:
+- NodeID
+- Wallet address
+- The date the node joined the network
+- The date the node initiated the Graceful Exit
+- GB Transferred (amount of data the node transferred during exiting)
+
+## Design
+
+Add satellite CLI command to list gracefully exited nodes between two dates. The report should query `nodes` table where `exit_finished_at >= start AND exit_finished_at <= end date`.
```
littleskunk: The query is wrong. It should be `exit_finished_at >= start AND exit_finished_at < end date`. The end date is the first of a month 0:00 and if a storage node hits that timestamp it would get a place in 2 reports.

docs/design/storagenode-graceful-exit/reports.md
```diff
@@ -0,0 +1,45 @@
...
+- Wallet address
+- The date the node joined the network
+- The date the node initiated the Graceful Exit
+- GB Transferred (amount of data the node transferred during exiting)
+
+## Design
+
+Add satellite CLI command to list gracefully exited nodes between two dates. The report should query `nodes` table where `exit_finished_at >= start AND exit_finished_at <= end date`. 
+
+Add satellite CLI command to list gracefully exiting nodes between two dates. The report should query `nodes` table where `exit_started_at >= start AND exit_started_at <= end date AND exit_finished_at IS NULL`.
```
littleskunk: Same: `exit_started_at >= start AND exit_started_at < end date AND exit_finished_at IS NULL`

docs/design/storagenode-graceful-exit/protocol.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Verifying transfer
+
+The storage node sends the piece hash, order limit, original "uploader" signature, and piece hash signed by the replacement node to the satellite for confirmation.
+
+The satellite verifies that the original piece hash matches the replacement piece hash, and verifies the order limit's signature. On success, the satellite will update segment / pointer information. If verification fails, the satellite will send an `ExitFailed` message.
```
littleskunk: Please change it into `and verifies both order limit's signature`. If the satellite is missing to verify the uploader signature than I will upload garbage to the second storage node and manipulate my signature. The hash will match and one signature is valid.

docs/design/storagenode-graceful-exit/ui.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Rationale
+
+We could use a design based on writing a number and then asking for confirmation. However, by requesting to type the domain name, it acts as a confirmation.
+
+For `exit-satellite` command it could stay up and show exiting progress. However, exit could take several days and the storage node may even restart during the process.
+
+## Implementation
+
+- Add `storagenode exit-satellite` command to storagenode CLI, which calls `gexit.Service.InitiateExit`.
```
kaloyan-raev: We don't have the `InitiateExit` message in the protocol anymore, right? So this line needs to be updated.

docs/design/storagenode-graceful-exit/ui.md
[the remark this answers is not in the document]
kaloyan-raev: If we decide to implement this, it would require a change in the protocol. Perhaps, we can add a `TerminateExit` message from the storage node. The satellite should respond with `ExitFailed` message and "terminated by storage node" as the Reason.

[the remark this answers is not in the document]
ethanadams: This feature should be considered in a future release.

docs/design/storagenode-graceful-exit/pieces.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/pieces.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/protocol.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/pieces.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/reports.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/reports.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/protocol.md
ethanadams: Updated

docs/design/storagenode-graceful-exit/ui.md
ethanadams: Updated

egonelbre:
@bryanchriswhite only pressing "Update branch" or empty commits preserve it. If you do it on command-line, I think github won't try to figure out whether it was a fast-forward/merge etc.

bryanchriswhite:
> @bryanchriswhite only pressing "Update branch" or empty commits preserve it. If you do it on command-line, I think github won't try to figure out whether it was a fast-forward/merge etc.

I used the "update branch" button on this PR.

egonelbre:
@bryanchriswhite No clue then.

kaloyan-raev:
It seems that the last commit https://github.com/storj/storj/pull/2866/commits/bc203fc94b68831156cb361f84b83c7e9e5eebd3 messed up the branch.

kaloyan-raev:
Or maybe not... It seems I was looking at a strange diff in Github.

PR #2872, opened 2019-08-26
web/marketing: static asset path

VinozzZ:
What: 

change absolute path to relative path for all assets

Why:
marketing web portal is not served at the root so we need to use relative path to access assets

Please describe the tests:
 - Test 1:
 - Test 2:
 

web/marketing/pages/index.html
```diff
@@ -4,7 +4,7 @@
 {{define "head"}}
     <title>Referral Admin Portal</title>
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
-    <link rel="stylesheet" type="text/css" href="/static/css/style.css">
+    <link rel="stylesheet" type="text/css" href="static/css/style.css">
```
egonelbre: Let's add a `{{.BaseURL}}` which you can supply via a config. I think it still should use absolute URL-s because otherwise it's easy to mess up paths. e.g. `/hello` works, but `/hello/world` doesn't.

stefanbenten: I support that idea!

egonelbre: You can supply the base url via template funcs. e.g. register a func called `Site`, which returns a type that contains field `BaseURL`, then you can use it as: `{{ Site.BaseURL }}`

littleskunk:
Looks good on the QA satellite: https://satellite.qa.storj.io/marketing

PR #2877, opened 2019-08-27
storagenode/contact: create chore for nodes to ping satellites

jenlij:
What: Creates a chore for nodes to announce themselves to their trusted satellites. Runs on startup and every hour thereafter

Why: Kademlia Removal

Please describe the tests:
 - Test 1: basic tests of gRPC service
 - Test 2: assert that lastPing time is updated when node contact endpoint is hit
 

storagenode/outreach/chore.go
```diff
@@ -0,0 +1,108 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package outreach
```
egonelbre: Wouldn't this be `contact` system?

storagenode/contact/chore.go
```diff
@@ -0,0 +1,127 @@
...
+	"storj.io/storj/storagenode/trust"
+)
+
+var mon = monkit.Package()
+
+// Config contains configurable parameters for contact chore
+// TODO: Hide MaxSleep from CLI
+type Config struct {
+	Interval time.Duration `help:"how frequently the node contact chore should run" releaseDefault:"1h" devDefault:"30s"`
+	MaxSleep time.Duration `help:"maximum duration to wait before pinging satellites" releaseDefault:"45m" devDefault:"0s"`
```
jenlij: Maybe along with hiding MaxSleep from the CLI, we can write a note in the code that encourages the SNO to keep the default to increase the likelihood of a successful connection to the satellite. To prevent people from messing around in the code.

storagenode/contact/chore.go
```diff
@@ -0,0 +1,130 @@
...
+	satellites := chore.trust.GetSatellites(ctx)
+	for _, satellite := range satellites {
+		satellite := satellite
+		addr, err := chore.trust.GetAddress(ctx, satellite)
+		if err != nil {
+			chore.log.Error("getting satellite address", zap.Error(err))
+			continue
+		}
+		group.Go(func() error {
+			conn, err := chore.transport.DialNode(ctx, &pb.Node{
```
jenlij: since we have endpoints on the satellite the node can reach already, we could try testing this part here and make sure it doesn't error. Then write a todo to check the connection

storagenode/contact/chore.go
```diff
@@ -0,0 +1,130 @@
...
+				Id: satellite,
+				Address: &pb.NodeAddress{
+					Transport: pb.NodeTransport_TCP_TLS_GRPC,
+					Address:   addr,
+				},
+			})
+			if err != nil {
+				return err
+			}
+			_, err = pb.NewNodeClient(conn).Checkin(ctx, &pb.CheckinRequest{
```
jenlij: same testing suggestion as above

storagenode/contact/chore.go
```diff
@@ -0,0 +1,130 @@
...
+			return err
+		})
+	}
+
+	return group.Wait()
+}
+
+// sleep for random interval in [0;maxSleep)
+// returns error if context was cancelled
+func (chore *Chore) sleep(ctx context.Context) error {
```
jenlij: you could test this method just by making sure it returns nil

storagenode/contact/endpoint.go
```diff
@@ -14,19 +15,27 @@ import (
...
-	service *Service
+	log       *zap.Logger
+	PingStats *PingStats
+}
+
+// PingStats contains information regarding who and when the node was last pinged
+type PingStats struct {
+	mu               sync.Mutex
+	lastPinged       time.Time
+	whoPingedNodeID  storj.NodeID
```
jenlij: great

storagenode/peer.go
```diff
@@ -244,8 +246,8 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, revocationDB exten
 	}
 
 	{ // setup contact service
-		peer.Contact.Service = contact.NewService(peer.Log.Named("contact"), peer.Kademlia.RoutingTable.Local(), peer.Transport)
-		peer.Contact.Endpoint = contact.NewEndpoint(peer.Log.Named("contact:endpoint"), peer.Contact.Service)
+		peer.Contact.Endpoint = contact.NewEndpoint(peer.Log.Named("contact:endpoint"))
+		peer.Contact.Chore = contact.NewChore(peer.Log.Named("contact"), config.Contact.Interval, config.Contact.MaxSleep, peer.Storage2.Trust, peer.Transport, peer.Kademlia.RoutingTable.Local())
```
jenlij: maybe update the log name from 'contact' to 'chore'

storagenode/peer.go
```diff
@@ -493,6 +502,7 @@ func (peer *Peer) Close() error {
 		errlist.Add(peer.NodeStats.Cache.Close())
 	}
 
+	// TODO: make sure to close any contact connections
```
jenlij: I think you could remove this TODO since you're calling Chore.Close()

jenlij:
Nice work @cam-a !

storagenode/peer.go
jenlij: or rather "contact: chore"

storagenode/contact/chore.go
```diff
@@ -0,0 +1,129 @@
...
+// Run the contact chore on a regular interval with jitter
+func (chore *Chore) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	chore.log.Info("Storagenode contact chore starting up")
+
+	return chore.Loop.Run(ctx, func(ctx context.Context) error {
+		if err := chore.sleep(ctx); err != nil {
+			return err
+		}
+		if err = chore.pingSatellites(ctx); err != nil {
```
thepaul: This assignment should probably be `err := ` instead of `err = ` to make it more obvious that this error value does not need to live past this block. Written as it is, this `err` becomes a capture of the one declared as a named return on line 57, which is weird because `Chore.Run()` will probably return before most invocations of this anonymous function are even called. All invocations will use the same memory location for this variable, and it won't necessarily be initialized to `nil` each time, and oh man if two of these get invoked at the same time this would probably even cause a race condition. None of this ought to be a problem with the way the code is written right now, but if the code changes this could be a source of some really impressively subtle bugs 😀

storagenode/contact/chore.go
```diff
@@ -0,0 +1,129 @@
...
+				Id: satellite,
+				Address: &pb.NodeAddress{
+					Transport: pb.NodeTransport_TCP_TLS_GRPC,
+					Address:   addr,
+				},
+			})
+			if err != nil {
+				return err
+			}
+			_, err = pb.NewNodeClient(conn).Checkin(ctx, &pb.CheckinRequest{
```
thepaul: I think we need a `defer` before this line to close `conn`, don't we?

storagenode/contact/chore.go
```diff
@@ -0,0 +1,129 @@
...
+// Chore is the contact chore for nodes announcing themselves to their trusted satellites
+type Chore struct {
+	log       *zap.Logger
+	self      overlay.NodeDossier
+	transport transport.Client
+
+	trust *trust.Pool
+
+	maxSleep time.Duration
+	Loop     *sync2.Cycle
```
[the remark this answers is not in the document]
egonelbre: `sync2.Cycle` are exposed for testing, such that tests can pause and trigger them on will, rather than using `Sleeping`. Ideally, yes, they would be hidden, but for testing, it's easier to expose them rather than build a different mechanism for exposing them only for tests.

storagenode/contact/chore.go
cam-a: can't test this since it's a local function and the test file needs to be in a separate package to prevent circular dependencies

jenlij: ah ok

storagenode/contact/endpoint.go
```diff
@@ -14,24 +15,33 @@ import (
...
 // Endpoint implements the contact service Endpoints
 type Endpoint struct {
-	log     *zap.Logger
-	service *Service
+	log       *zap.Logger
+	pingStats *PingStats
+}
+
+// PingStats contains information regarding who and when the node was last pinged
+type PingStats struct {
```
JessicaGreben: why do we need to track when/who the node was last pinged?

jenlij: just for the node dashboard

cam-a: I believe the dashboard indicates the last time you were contacted to let you know you are still externally accessible

storagenode/contact/endpoint.go
```diff
@@ -42,6 +52,22 @@ func (endpoint *Endpoint) Ping(ctx context.Context, req *pb.ContactPingRequest)
...
-	endpoint.service.wasPinged(time.Now(), peerID.ID, p.Addr.String())
+	endpoint.pingStats.wasPinged(time.Now(), peerID.ID, p.Addr.String())
 	return &pb.ContactPingResponse{}, nil
 }
+
+// WhenLastPinged returns last time someone pinged this node.
+func (stats *PingStats) WhenLastPinged() (when time.Time, who storj.NodeID, addr string) {
+	stats.mu.Lock()
+	defer stats.mu.Unlock()
+	return stats.lastPinged, stats.whoPingedNodeID, stats.whoPingedAddress
```
JessicaGreben: it looks like this method `WhenLastPinged` isn't currently being used anywhere.  Is there a use case for this method to be used in the future? Im just thinking that if the node restarts for any reason, pingStats will get reset and these fields will be empty.  Not sure if that will affect anything

cam-a: It will be used by the dashboard to get the last contacted time. I'm also using it in the test

storagenode/contact/chore.go
```diff
@@ -0,0 +1,136 @@
...
+		transport: transport,
+
+		trust: trust,
+
+		maxSleep: maxSleep,
+		Loop:     sync2.NewCycle(interval),
+	}
+}
+
+// Run the contact chore on a regular interval with jitter
```
JessicaGreben: not sure if we should expand on this comment a little bit and explain what is meant by jitter

PR #2888, opened 2019-08-27
satellite/audit: create the audit queue

navillasa:
What:
- adds an audit queue containing paths to audit
- adds worker method to ReservoirService to do audit jobs (that is, auditing paths on the queue)

Why:
- To create clearer organization of the audit system--
    - `reservoirChore.Run` adds reservoir paths to queue in pseudorandom order
    - audit workers will take a path from the queue and run an audit on that path
    - no duplicate paths will be added to the queue
    - paths will be deleted from the queue as soon as an audit worker picks up the path
    - the random selection from the reservoirs (in pr https://github.com/storj/storj/pull/2861) was a bit convoluted

Please describe the tests:
 - TestChoreAndWorkerIntegration:
This test ensures that after uploading 2 files (1 remote segment each), the paths for those files appear appropriately in the queue.

 - TestQueue:
This tests the expected behavior of the Swap and Next methods of the queue struct.
 

satellite/audit/reservoirservice.go
```diff
@@ -31,10 +35,12 @@ type ReservoirService struct {
 // NewReservoirService instantiates ReservoirService
```
ifraixedes:
```suggestion
// NewReservoirService instantiates ReservoirService.
```

satellite/audit/reservoirservice.go
```diff
@@ -46,21 +52,95 @@ func (service *ReservoirService) Run(ctx context.Context) (err error) {
...
+		pathCollector := NewPathCollector(service.config.Slots, service.rand)
 		err = service.MetainfoLoop.Join(ctx, pathCollector)
 		if err != nil {
 			service.log.Error("error joining metainfoloop", zap.Error(err))
 			return nil
 		}
-		service.Reservoirs = pathCollector.Reservoirs
+
+		var queue []storj.Path
+		queuePaths := make(map[storj.Path]bool)
```
ifraixedes: nit: I would use `map[storj.Path]struct{}` to express that the value is meaningless

satellite/audit/reservoirservice.go
```diff
@@ -18,11 +20,13 @@ import (
...
-	log *zap.Logger
+	log    *zap.Logger
+	config Config
+	rand   *rand.Rand
 
-	reservoirSlots int
-	Reservoirs     map[storj.NodeID]*Reservoir
-	rand           *rand.Rand
+	cond   sync.Cond
+	queue  []storj.Path
```
egonelbre: Separate this into a struct.

satellite/audit/reservoirservice.go
```diff
@@ -46,21 +52,95 @@ func (service *ReservoirService) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
 	service.log.Info("audit 2.0 is starting up")
 
+	var group errgroup.Group
```
egonelbre: The worker shouldn't be part of `ReservoirService`

satellite/audit/reservoirservice.go
```diff
@@ -18,11 +20,13 @@ import (
 // ReservoirService is a temp name for the service struct during the audit 2.0 refactor.
 // Once V3-2363 and V3-2364 are implemented, ReservoirService will replace the existing Service struct.
 type ReservoirService struct {
```
egonelbre: `ReservoirService` -> `ReservoirChore`

satellite/audit/service.go
```diff
@@ -31,7 +31,8 @@ type Config struct {
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"25s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
 
-	Slots int `help:"number of reservoir slots allotted for nodes, currently capped at 2" default:"1"`
+	Slots       int `help:"number of reservoir slots allotted for nodes, currently capped at 2" default:"1"`
+	WorkerCount int `help:"number of workers to run audits on paths" default:"3"`
```
egonelbre: Set default to `1`, but for testing use 2

satellite/audit/reservoirservice.go
```diff
@@ -46,21 +52,95 @@ func (service *ReservoirService) Run(ctx context.Context) (err error) {
...
+}
+
+// next gets the next item in the queue.
+func (service *ReservoirService) next() storj.Path {
+	service.cond.L.Lock()
+	defer service.cond.L.Unlock()
+
+	if len(service.queue) == 0 {
+		// This waits until the queue is repopulated.
+		service.cond.Wait()
```
egonelbre: When this is signaled, the queue might not be full

egonelbre: Sorry, I meant that the queue might be empty.

satellite/audit/service2.go
```diff
@@ -0,0 +1,89 @@
...
+func NewService2(log *zap.Logger, config Config, metaloop *metainfo.Loop) (*Service2, error) {
+	queue := newQueue()
+	var workers []*worker
+	for i := 0; i < config.WorkerCount; i++ {
+		workers = append(workers, newWorker(queue))
+	}
+	return &Service2{
+		log: log,
+
+		reservoirChore: NewReservoirChore(log.Named("reservoir chore"), queue, metaloop, config),
```
egonelbre: Services should not import chores. https://github.com/storj/docs/blob/master/code/Architecture.md#interactions-between-layers

satellite/audit/queue.go
```diff
@@ -0,0 +1,62 @@
...
+func (queue *queue) swap(newQueue []storj.Path) {
+	queue.cond.L.Lock()
+	queue.queue = newQueue
+	// Notify workers that queue has been repopulated.
+	queue.cond.Broadcast()
+	queue.cond.L.Unlock()
+}
+
+// next gets the next item in the queue.
+func (queue *queue) next(ctx context.Context) (storj.Path, error) {
```
egonelbre: Make it polling, since it eventually has to be.

egonelbre: The polling happens usually outside of the queue, because it allows for callers to have more control and run other things while there's nothing in it.

See how repair queue is used.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,88 @@
...
+		Interval:     config.ChoreInterval,
+	}
+}
+
+func (chore *ReservoirChore) Run(ctx context.Context) error {
+	return chore.Loop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+
+		select {
+		case <-chore.service.queue.closed:
```
egonelbre: ReservoirChore shouldn't have to care about queue being closed. Queue is a persistence concern so it shouldn't be closed until all services/chores and peer has stopped.

See how repair handles it.

satellite/audit/queue.go
mobyvb: Updated to poll outside of the queue

satellite/audit/reservoirchore.go
mobyvb: I removed the concept of closing from the queue, and the channel is now on the service struct. I do not think the check is necessary since the peer closes the service before the chore, but it's probably safer to have it I think.

satellite/audit/queue.go
```diff
@@ -0,0 +1,44 @@
...
+func (q *queue) swap(newQueue []storj.Path) {
+	q.mu.Lock()
+	q.queue = newQueue
+	q.mu.Unlock()
+}
+
+// next gets the next item in the queue.
+func (q *queue) next() (storj.Path, error) {
+	// return error if queue is empty
+	if len(q.queue) == 0 {
```
egonelbre: You need to lock before checking the length.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+// Run starts the reservoir chore
+func (chore *ReservoirChore) Run(ctx context.Context) error {
+	return chore.Loop.Run(ctx, func(ctx context.Context) (err error) {
+		defer mon.Task()(&ctx)(&err)
+
+		select {
+		case <-ctx.Done():
```
egonelbre: There's no need to check `ctx.Done` here because `chore.Loop.Run` already handles it.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,87 @@
...
+	return &ReservoirChore{
+		log:    log,
+		config: config,
+		rand:   rand.New(rand.NewSource(time.Now().Unix())),
+
+		service: service,
+
+		MetainfoLoop: metaLoop,
+		Loop:         *sync2.NewCycle(config.QueueInterval),
+		Interval:     config.ChoreInterval,
```
egonelbre: The intervals have been messed up. This chore should only have one interval.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,87 @@
...
+			return nil
+		}
+
+		var queue []storj.Path
+		queuePaths := make(map[storj.Path]struct{})
+
+		// Add reservoir paths to queue in pseudorandom order.
+		for i := 0; i < chore.config.Slots; i++ {
+			for _, res := range pathCollector.Reservoirs {
+				path := res.Paths[i]
```
egonelbre: Some of these paths maybe empty, so ignore them.

satellite/audit/service2.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+// Service2 contains information for populating audit queue and processing audits.
+type Service2 struct {
```
egonelbre: Let's call this a `Worker`.

satellite/audit/service2.go
```diff
@@ -0,0 +1,102 @@
...
+		queue:   queue,
+	}, nil
+}
+
+// Run runs audit service 2.0.
+func (service *Service2) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	service.log.Info("audit 2.0 is starting up")
+
+	var group errgroup.Group
```
egonelbre: You can use here a cycle and a limiter to handle these. It would simplify the code.

satellite/audit/service2.go
```diff
@@ -0,0 +1,102 @@
...
+		})
+	}
+
+	return group.Wait()
+
+}
+
+// Close halts the reservoir chore and audit workers.
+func (service *Service2) Close() error {
+	close(service.closed)
```
egonelbre: I do not understand why there is this `closed` here?

mobyvb: Oh, I don't think it got captured in the latest refactor, but we plan to check `service.closed` inside the loop inside `func (*worker) run(ctx context.Context)`. Alternatively, we could create a new ctx with cancel that we pass into that function, and cancel it here. Do you have a preference for how we handle this?

satellite/audit/service2.go
mobyvb: Do you still want us to name the individual processes `worker`s if we rename the main service to `Worker`?

satellite/audit/service2.go
egonelbre: https://github.com/storj/storj/blob/master/satellite/repair/repairer/repairer.go#L70

egonelbre: We can do a similar approach as in repairer.

satellite/audit/service2.go
egonelbre: Let's make the main thing `Worker`, because I think it makes sense to have it top-level. It sits somewhere between a Service and a Chore -- so having a distinct name for it probably will help.

satellite/audit/reservoirchore.go
mobyvb: Fixed

satellite/audit/service2.go
mobyvb: Done

satellite/audit/reservoirchore.go
mobyvb: Fixed

satellite/audit/queue.go
mobyvb: Fixed

satellite/audit/worker.go
```diff
@@ -0,0 +1,83 @@
...
+// Worker contains information for populating audit queue and processing audits.
+type Worker struct {
+	log    *zap.Logger
+	config Config
+	closed chan struct{}
+
+	queue *queue
+}
+
+// NewWorker instantiates Worker and workers.
```
VinozzZ: What's the difference between `Worker` and `workers` in the comment?

satellite/audit/service2.go
mobyvb: I added a limiter, but what is the need for the cycle? `process()` already iterates and does polling when the queue is empty.

satellite/audit/service2.go
mobyvb: I tried to make it similar to repairer, but not sure if it is what you had in mind

satellite/audit/worker.go
mobyvb: Fixed

satellite/audit/service2.go
egonelbre: Cycle allows controlling when the worker goes to check the queue. Otherwise, tests will end up sleeping to figure out whether workers have finished or not -- the reason we use cycle in the first place.

Also, you don't want to keep all the workers in a single process to be alive and pinging the queue.

egonelbre: i.e. https://github.com/storj/storj/pull/2924/files

satellite/audit/service2.go
egonelbre: https://github.com/storj/storj/pull/2924/files

satellite/audit/worker.go
```diff
@@ -0,0 +1,80 @@
...
+// Run runs audit service 2.0.
+func (worker *Worker) Run(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	worker.log.Info("starting")
+
+	// wait for all audits to run
+	defer worker.Limiter.Wait()
+
+	return worker.Loop.Run(ctx, func(ctx context.Context) error {
+		err := worker.process(ctx)
```
mobyvb: @egonelbre , if we queue all the work in the queue to the limiter, then the queue gets updated by the chore, then we call `worker.process(ctx)` again here after the queue interval, I think the following could happen:
* The queue is updated by the chore to have 1000 unique items in it
* 1000 items of work are added to the limiter, and the queue is now empty
* The queue is updated by the chore again, and it has 1000 items again. However, the limiter has only managed to process 500 of the originally queued items. 
* The worker loop runs again, and adds the 1000 items from the queue onto the limiter. Now we have 1500 items to process, but we don't really care about the first 500 anymore because we got a new queue.
I have some ideas in mind, but I don't know if they are the best way to solve the issue.

egonelbre: The limiter ensures that only up to `N` workers can run at a time. If the limit is reached it waits. So if there are 1000 unique items, it will wait after 3 have been added.

mobyvb: Oh, I understand better now. For some reason I was under the impression that `worker.Limiter.Go` would still return if `N` workers were in progress, and that `Go` would still process that work at a later point. After taking another look at the limiter code, I see that I was mistaken.

satellite/audit/worker.go
```diff
@@ -0,0 +1,80 @@
...
+			err := worker.work(ctx, path)
+			if err != nil {
+				worker.log.Error("audit failed", zap.Error(err))
+			}
+		})
+	}
+}
+
+func (worker *Worker) work(ctx context.Context, path storj.Path) error {
+	// handle work
```
ifraixedes: Better to add a `TODO:` prefix

satellite/peer.go
```diff
@@ -489,7 +491,18 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, revocationDB exten
 		}
 
 		// setup audit 2.0
-		peer.Audit.ReservoirService = audit.NewReservoirService(peer.Log.Named("reservoir service"),
+		peer.Audit.Queue = &audit.Queue{}
+
+		peer.Audit.Worker, err = audit.NewWorker(peer.Log.Named("audit service 2"),
```
ifraixedes: Is there any plan to remove the `2` from the logger at some point in the future?

satellite/audit/worker.go
```diff
@@ -34,13 +34,14 @@ func NewWorker(log *zap.Logger, queue *Queue, config Config) (*Worker, error) {
 // Run runs audit service 2.0.
 func (worker *Worker) Run(ctx context.Context) (err error) {
 	defer mon.Task()(&ctx)(&err)
-	worker.log.Info("starting")
+	worker.log.Debug("starting")
 
-	// wait for all audits to run
+	// Wait for all audits to run.
```
ifraixedes: Just to let you know, I haven't commented on these ones because they are purely source code comments.

The ones that we ask for ends with a period are the ones which are used by godoc to generate the HTML documentation and for following the Go conventions.

For the internal ones, at least for, in my opinion, we can be more flexible. Obviously adding a period doesn't hurt.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+		satellite := planet.Satellites[0]
+		satellite.Audit.Worker.Loop.Pause()
+		satellite.Audit.Chore.Loop.Pause()
+
+		ul := planet.Uplinks[0]
+
+		// Upload 2 remote files with 1 segment.
+		for i := 0; i < 2; i++ {
+			testData := testrand.Bytes(8 * memory.KiB)
+			path := "/some/remote/path/" + string(i)
```
ifraixedes: nit:
`string(i)` translates to the UTF-8 character indexed by the `i` rather than string representation of its value. While it isn't a problem if at some point we want to print those paths for debugging or whatever else, it will be harder to read them

https://play.golang.org/p/1XIzSk8NSYH

I would prefer to use something like `path := "/some/remote/path/" + strconv.Itoa(i)`

satellite/audit/queue_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"storj.io/storj/pkg/storj"
+)
+
+func TestQueue(t *testing.T) {
+	queue := &Queue{
+		Queue: []storj.Path{},
+	}
+	_, err := queue.Next()
+	require.True(t, ErrEmptyQueue.Has(err))
```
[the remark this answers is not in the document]
egonelbre: I wouldn't bother with these since it's often more informative to look at the code. There are cases where this extra tag is useful, e.g. loops, where you cannot easily deduce which test is failing.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,85 @@
...
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/metainfo"
+)
+
+// ReservoirChore populates reservoirs and the audit queue.
+type ReservoirChore struct {
+	log    *zap.Logger
+	config Config
+	rand   *rand.Rand
+	Queue  *Queue
```
egonelbre: This queue should be mostly internal, so maybe make it private.

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,85 @@
...
+	Loop   sync2.Cycle
+
+	MetainfoLoop *metainfo.Loop
```
egonelbre: This metainfo loop should be mostly internal, so maybe make it private.

satellite/audit/worker.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// NewWorker instantiates Worker.
+func NewWorker(log *zap.Logger, queue *Queue, config Config) (*Worker, error) {
+	return &Worker{
+		log: log,
+
+		queue:   queue,
+		Loop:    *sync2.NewCycle(config.QueueInterval),
+		Limiter: *sync2.NewLimiter(config.WorkerConcurrency),
```
egonelbre: This limiter should be mostly internal, so maybe make it private.

satellite/audit/worker.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Close halts the worker.
+func (worker *Worker) Close() error {
+	return nil
+}
+
+// process repeatedly removes an item from the queue and runs an audit.
+func (worker *Worker) process(ctx context.Context) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	for {
```
egonelbre: Wait for the workers to complete before running again. This makes writing tests somewhat easier.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+			err := ul.Upload(ctx, satellite, "testbucket", path, testData)
+			require.NoError(t, err)
+		}
+
+		satellite.Audit.Chore.Loop.Restart()
```
egonelbre: You can call `TriggerWait`, without restarting the loop.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+		for _, path := range satellite.Audit.Queue.Queue {
+			_, ok := uniquePaths[path]
+			require.False(t, ok, "expected unique path in chore queue")
+
+			uniquePaths[path] = struct{}{}
+		}
+		require.Len(t, uniquePaths, 2)
+
+		satellite.Audit.Worker.Loop.Restart()
```
egonelbre: You can use `TriggerWait` without restarting the loop.

satellite/audit/queue.go
```diff
@@ -0,0 +1,45 @@
...
+	"storj.io/storj/pkg/storj"
+)
+
+// ErrEmptyQueue is used to indicate that the queue is empty.
+var ErrEmptyQueue = errs.Class("empty audit queue")
+
+// Queue is a list of paths to audit, shared between the reservoir chore and audit workers.
+type Queue struct {
+	mu    sync.Mutex
+	Queue []storj.Path
```
egonelbre: Try to avoid exposing the internal state of the queue. It's used concurrently, so it probably shouldn't be accessed outside the implementation.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+			path := "/some/remote/path/" + string(i)
+			err := ul.Upload(ctx, satellite, "testbucket", path, testData)
+			require.NoError(t, err)
+		}
+
+		satellite.Audit.Chore.Loop.Restart()
+		satellite.Audit.Chore.Loop.TriggerWait()
+
+		uniquePaths := make(map[storj.Path]struct{})
+		assert.Len(t, satellite.Audit.Queue.Queue, 2)
```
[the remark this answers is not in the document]
mobyvb: It's an assert because we also want to check that the queue is unique even if the length check fails.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+			require.False(t, ok, "expected unique path in chore queue")
+
+			uniquePaths[path] = struct{}{}
+		}
+		require.Len(t, uniquePaths, 2)
+
+		satellite.Audit.Worker.Loop.Restart()
+		satellite.Audit.Worker.Loop.TriggerWait()
+
+		require.Len(t, satellite.Audit.Queue.Queue, 0)
```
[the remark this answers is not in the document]
mobyvb: The test error should say something like "should have 0 items but has x". Do you think that is not informative enough?

satellite/audit/integration_test.go
mobyvb: It is paused it at the beginning of the test, so it should be restarted. We can't have it running until this point because otherwise there might be a data race with checking the queue after the chore.

satellite/audit/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+		uniquePaths := make(map[storj.Path]struct{})
+		assert.Len(t, satellite.Audit.Queue.Queue, 2)
+
+		for _, path := range satellite.Audit.Queue.Queue {
+			_, ok := uniquePaths[path]
+			require.False(t, ok, "expected unique path in chore queue")
+
+			uniquePaths[path] = struct{}{}
+		}
+		require.Len(t, uniquePaths, 2)
```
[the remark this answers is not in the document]
mobyvb: Line 41 doesn't check the uniqueness, it just checks that the size of the queue is 2.

satellite/audit/integration_test.go
egonelbre: What I mean, is that you can completely remove `Worker.Loop.Restart` and it should still work.

satellite/audit/integration_test.go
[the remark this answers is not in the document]
ifraixedes: The uniqueness is checked in line 45

satellite/audit/integration_test.go
mobyvb: Oh, sweet.

satellite/audit/integration_test.go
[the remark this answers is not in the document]
ifraixedes: I wanted to read what has the X items.

But as Egon mention, jumping to the line is fine.

I found it more useful when they are `assert` rather than `require` and more useful when the errors happen in the CI or on tests which don't look related to what you're modifying.

satellite/audit/integration_test.go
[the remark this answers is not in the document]
mobyvb: oh, you're right

satellite/audit/reservoirchore.go
navillasa: should `Loop sync2.Cycle` also be unexported?

egonelbre: Ideally yes, however, we need all the loops for controlling tests. There's no other reasonable way to expose them either. So, `sync2.Cycle` is the exception.

navillasa: Ah okay that makes sense.

satellite/audit/service.go
```diff
@@ -31,7 +31,10 @@ type Config struct {
 	MinDownloadTimeout time.Duration `help:"the minimum duration for downloading a share from storage nodes before timing out" default:"25s"`
 	MaxReverifyCount   int           `help:"limit above which we consider an audit is failed" default:"3"`
 
-	Slots int `help:"number of reservoir slots allotted for nodes, currently capped at 2" default:"1"`
+	ChoreInterval     time.Duration `help:"how often to run the reservoir chore" default:"24h"`
+	QueueInterval     time.Duration `help:"how often to recheck an empty audit queue" default:"30s"`
```
mobyvb: if the audit queue is empty, we will need to wait for the chore to run again for it to fill up. Since the chore interval is 24h, I feel like 30s is too frequent for the queue interval. I think it should be at least 1h by default.

satellite/audit/queue_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit
```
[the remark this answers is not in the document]
navillasa: the reason for package audit is because it's a unit test, and the the unexported `queue` field on `Queue` can't be instantiated if this package uses `audit_test`.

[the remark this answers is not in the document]
mobyvb: Oh, it doesn't look like the unexported queue field is used except for 
```
q := &Queue{
	queue: []storj.Path{},
}
```
But that can be replaced with `q := &Queue{}`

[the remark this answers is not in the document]
navillasa: that didn't seem to work because the unexported queue field is still nil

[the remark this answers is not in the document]
navillasa: oop never mind you were right

internal/testplanet/satellite.go
```diff
@@ -162,10 +162,14 @@ func (planet *Planet) newSatellites(count int) ([]*satellite.Peer, error) {
 				MaxExcessRateOptimalThreshold: 0.05,
 			},
 			Audit: audit.Config{
+				Slots:              3,
+				ChoreInterval:      30 * time.Second,
+				QueueInterval:      1 * time.Hour,
```
egonelbre: nit: Let's keep the config organised the same way as the definition.

satellite/audit/queue_test.go
```diff
@@ -0,0 +1,48 @@
...
+	require.EqualValues(t, testQueue1[0], path)
+
+	path, err = q.Next()
+	require.NoError(t, err)
+	require.EqualValues(t, testQueue1[1], path)
+
+	testQueue2 := []storj.Path{"0", "1", "2"}
+	q.Swap(testQueue2)
+
+	path, err = q.Next()
```
egonelbre: This can be written as:
```
for _, expected := range testQueue2 {
    path, err := q.Next()
    require.NoError(t, err)
    require.EqualValues(t, expected, path)
}
```

satellite/audit/reservoirchore.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"go.uber.org/zap"
+
+	"storj.io/storj/internal/sync2"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/metainfo"
+)
+
+// ReservoirChore populates reservoirs and the audit queue.
+type ReservoirChore struct {
```
egonelbre: If the other audit thing is going away, this could be named just `Chore`, because there wouldn't be a naming conflict.

PR #2897, opened 2019-08-28
cmd/certificates: auth export improvements & certificates test script

bryanchriswhite:
What: 
  + Add claimed column to `certificates auth export` output.
  + Replace old certificates test script.
  + Fix `--ident-dir` global flag on `certificates` command.

Why: To make the CSV more useful.

Please describe the tests: 

`./scripts/test-certificates.sh`:
1. Creates an identity for the certificates server.
1. Creates 5 identities, each with `i` number of authorizations. `i` starts at 1 and is incremented by 1 for each created identity.
1. Exports authorizations as csv.
1. Starts the certificates server.
1. Authorizes each identity using the first authorization token for each.
1. Stops the certificates server (to prevent locking on the bolt db).
1. Exports authorizations as csv.
1. Asserts that each identity has `i` authorizations, 1 claimed and `i` - 1 unclaimed.


cmd/certificates/auth.go
```diff
@@ -257,7 +257,12 @@ func writeAuthExport(ctx context.Context, authDB *certificates.AuthorizationDB,
 
 	var authErrs errs.Group
 	for _, auth := range auths {
-		if err := w.Write([]string{email, auth.Token.String()}); err != nil {
+		isClaimed := "false"
+		if auth.Claim != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not sure I follow the question. This helper function is used in the `certificates auth export` subcommand which generates a CSV of the specified authorizations/tokens (all, or by email address/list).

In this helper, `auths` is a slice of `Authorizations` which have a `*Claim` field. If the auth that comes out of the database has a nil claim then it has not been claimed, otherwise, it has and the claim has the details.

Does that answer your question?

stefanbenten:
Might need a test to actually see if the output is right and stays like that longterm.

scripts/test-certificates.sh
```diff
@@ -0,0 +1,123 @@
...
+
+  _identity_create $ident_name
+
+  if [[ i -gt 0 ]]; then
+    _certificates auth create "$i" "$email"
+  fi
+done
+
+exported_auths=$(_certificates auth export)
+_certificates run --signer.min-difficulty 0 &
```
navillasa: you can get the pid of this server by using `PID=$!` and then you can use that to kill it later on instead of `killall "certificates"` on line 14.

scripts/test-certificates.sh
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env bash
+set -ueo pipefail
+source $(dirname $0)/utils.sh
+
+TMPDIR=$(mktemp -d -t tmp.XXXXXXXXXX)
+IDENTS_DIR=$TMPDIR/identities
+CERTS_DIR=$TMPDIR/certificates
+# TODO: make port numbers more dynamic
```
navillasa: I'm leaving the "follows testing guide" box unchecked because I think we're supposed to choose ports dynamically but I'm also not totally sure how to do this in bash....

scripts/test-certificates.sh
bryanchriswhite: Yeah, I tried that but `$!` was returning a pid that was different that the pid of the process that's running when you look in `ps`.

scripts/test-certificates.sh
bryanchriswhite: Yea, I don't think we have any examples of ports being dynamically chosen in bash but I chose a port range above the storj-sim defaults and it works in jenkins.

ifraixedes: I've been trying to find a solution but what I tested didn't work.

thepaul: Here's a dumb but portable (macOS and linux) way to get an unused tcp port:

```sh
    port=$(netstat -ant -p tcp 2>/dev/null | awk '$0~/tcp/ {sub(/.*[:.]/,"",$4); a[$4]=1} END {for(i=1024;i<65000;i++) if(a[i]==0) {print i; exit}}')
    [ -n "$port" ] || { echo "No ports available" >&2; exit 1; }
```

Obviously there's a race between choosing the port and using it, but in many systems that is acceptable. To really fix it, the best thing would be to allow passing `:0` to the application so that the tcp stack itself chooses an unused port when opening the socket, and then you make a way for the application to report when it has chosen a port and when it's ready to accept connections. Sometimes this takes the form of an option like `--report-fd=3` meaning the shell will open a pipe on fd 3 and the application is expected to write its status (including the listening port number) to fd 3 when it's ready to go. Or, sometimes, you make the shell do something like `tail -n +1 -F application.logfile | grep -m1 "Listening on port:"` to wait until the application reports its listening port to the log. But both of those are probably more work than they're worth.

scripts/test-certificates.sh
thepaul: Simplest fix to me looks like:

1. Change `_certificates` so that it invokes the `certificates` binary using `exec`, so it replaces the current process. Then `$!` here will be what you want.
2. The downside is that you'll also have to change lines 67 and 76 to call `_certificates` in the background with `&`. Call `wait` with no arguments afterward to wait for all background processes to finish.

scripts/test-certificates.sh
egonelbre: Use a different localhost to avoid conflicts. These will surely end-up causing flakiness in tests. Go tests use random ports on `127.0.0.1`, other tests use `127.0.0.2` etc.

scripts/test-certificates.sh
egonelbre: Alternatively use go tests which have more control over the concurrency and tests.

scripts/test-certificates.sh
```diff
@@ -0,0 +1,123 @@
...
+
+  if [[ i -gt 0 ]]; then
+    _certificates auth create "$i" "$email"
+  fi
+done
+
+exported_auths=$(_certificates auth export)
+_certificates run --signer.min-difficulty 0 &
+
+sleep 1
```
egonelbre: Would this cause flakiness, if things don't complete in 1s?

scripts/test-certificates.sh
thepaul: Using different interfaces is a nice and easy shortcut for now, but it's really just a punting of the problem. We will still need, eventually, to keep track of what pieces of the system are using which localhost interfaces to avoid collisions. And if we can do that, presumably we can also keep track of which pieces of the system are using which port numbers..

egonelbre: Yeah, I tend to agree. As much as possible tests are using `127.0.0.1:0` and then pick up the actual port being used. This works nicely as long as you can access the listener easily. For processes we do it in some places, e.g. for redis-server. However that's a annoying to work with and requires magic strings to work. Additionally you need to then handle cases that satellite exposes several ports.

Integration tests are pretty isolated and often require a bunch of ports to be set up. e.g. storj-sim integration with default setup ends up using over 35 ports, I think. The port layout of storj-sim is defined to be easier to remember, so they do not map to a sequence of ports. When we run multiple storj-sim-s side-by-side, as we already do, we would need to remap and reconfigure all those 35 ports.

Basically, it's easier to separate interfaces. It's possible, but also annoying. The conflicts also often end up as flakiness.

scripts/test-certificates.sh
thepaul: Using Go would be great also, but: we'd want the Go program to invoke `certificates` as a separate process in order to be sure that the binary behaves as expected. For something intended to run as a daemon, that's pretty important to test. But to do that would take away a lot of the motivation for using Go. Still, I will almost always be in favor of rewriting things away from Bash :)

egonelbre: We had prototypes for Go based tests in https://github.com/storj/storj/pull/1090/files#diff-e4d02494218358e9b7ffa5f637789a58R16

scripts/test-certificates.sh
thepaul: I find Docker to be hugely overused in the industry right now, but for better setup and teardown of isolated test networks that can't step on each other and can't leave surprising background processes lying around, it's really hard to beat containers!

scripts/test-certificates.sh
stefanbenten: Yeah it should..

scripts/test-certificates.sh
```diff
@@ -0,0 +1,123 @@
...
+  token=$(echo "$exported_auths" | grep "$email" | head -n 1 | awk -F , '{print $2}')
+  _identity authorize --signer.address "$CERTS_ADDR" "$ident_name" "$token" > /dev/null
+done
+
+# NB: Certificates server uses bolt by default so it must be shut down before we can export.
+kill_certificates_server
+
+# Expect 10 authorizations total.
+auths=$(_certificates auth export)
+require_lines 10 "$auths" $LINENO
```
stefanbenten: I cant fine the definition of `LINENO`. But maybe i am just blind :)

thepaul: `$LINENO` is one of the many bash magic variables. It gives you the current 1-indexed line number within a script or function, unless you explicitly set it to something else (whereupon it is no longer magic).

thepaul: `$LINENO` is one of the many Bash magic variables. It gives you the current 1-indexed line number within a script or function, unless you explicitly set it to something else (whereupon it is no longer magic).

scripts/test-certificates.sh
bryanchriswhite: I forgot that we can test cli code in go.

scripts/test-certificates.sh
bryanchriswhite: This is just waiting for the certificates server to come up.

bryanchriswhite:
I will be refactoring the certificates service in subsequent PRs. Can we please leave the test script as bash and I will convert it to go tests in those forthcoming PRs? As I attempt to write the go tests on this branch I already see how I'm going to have to turn right around and rewrite them.

This PR was just to add a true/false column to the csv output of the `certificates auth export` subcommand and I feel like it's gotten out of control.

My suggestion to move forward here would be to go with @thepaul's `exec`/`wait` fix and leave the ports for now (I tried @thepaul's suggestion for this but I wasn't able to get it working).

PR #2900, opened 2019-08-28
satellite/orders: Fix for V3-2529: Release v0.19.0 storage nodes can't submit orders, duplicate key value violates unique constraint

ethanadams:
What: 
Add a DB transaction savepoint so we can rollback to the previous "good" point if the used_serial insert fails due to key constraint.  Otherwise the transaction is terminated in postgres, and no other updates can be made.

Why:
If a node submits orders without completing the settlement, the satellite fails on the next settlement b/c of duplicate key violation.

This only happens with postgres, but the fix works for both.

https://storjlabs.atlassian.net/browse/V3-2529

Please describe the tests:
 - Test 1: Added TestSendingReceivingDuplicateOrders
 - Test 2:
 

satellite/satellitedb/orders.go
```diff
@@ -265,15 +265,27 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
 	// processes the insert to used serials table individually so we can handle
 	// the case where the order has already been processed.  Duplicates and previously
 	// processed orders are rejected
-	for _, request := range requests {
+	for i, request := range requests {
+		// create a savepoint so we can rollback if an order has already been used
+		_, err = tx.Exec(fmt.Sprintf("savepoint sp%d;", i))
```
egonelbre: What's the performance implication of using a savepoint?

egonelbre: Could we release the previous savepoint here?

egonelbre: Which benchmark code did you use?

satellite/satellitedb/orders.go
```diff
@@ -266,18 +266,31 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
...
 		}
+
+		insert := "INSERT INTO used_serials (serial_number_id, storage_node_id) SELECT id, ? FROM serial_numbers WHERE serial_number = ?"
+
 		_, err = tx.Exec(db.db.Rebind(insert), request.OrderLimit.StorageNodeId.Bytes(), request.OrderLimit.SerialNumber.Bytes())
 		if err != nil {
 			if pgutil.IsConstraintError(err) || sqliteutil.IsConstraintError(err) {
 				reject(request.OrderLimit.SerialNumber)
+				// rollback to the savepoint before the insert failed
+				_, err = tx.Exec("rollback to savepoint sp;")
```
ifraixedes: ~I think that what it solves the issue isn't the `savepoint` is that in this conditional the `err` should have been assigned to `nil` before the next iteration.~

~Probably what's happening now is that `err` isn't checked anymore in the next iteration because of the added call to `_, err = tx.Exec("savepoint sp;")`.~

~Anyway, it looks to me that the `if err != nil` conditional of line 270 (previous changes) wasn't needed and here a `continue` statement should have been called.~

satellite/satellitedb/orders.go
```diff
@@ -266,18 +266,31 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
 	// the case where the order has already been processed.  Duplicates and previously
 	// processed orders are rejected
 	for _, request := range requests {
-		insert := "INSERT INTO used_serials (serial_number_id, storage_node_id) SELECT id, ? FROM serial_numbers WHERE serial_number = ?"
+		// create a savepoint so we can rollback if an order has already been used
```
ifraixedes: nit: I would add to the comment something like

`Avoid the PG error "current transaction is aborted, commands ignored until end of transaction block" if the below insert fails due any constraint. See https://www.postgresql.org/message-id/13131805-BCBB-42DF-953B-27EE36AAF213%40yahoo.com`

I had to figure it out why the solution that I wrote below and after I strikethrough didn't work

satellite/satellitedb/orders.go
```diff
@@ -266,18 +266,31 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
...
+		_, err = tx.Exec("savepoint sp;")
```
ifraixedes: nit:
```suggestion
		_, err = tx.Exec("savepoint sp")
```

satellite/satellitedb/orders.go
```diff
@@ -266,18 +266,31 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
...
+				// rollback to the savepoint before the insert failed
+				_, err = tx.Exec("rollback to savepoint sp;")
+				if err != nil {
+					return nil, Error.Wrap(err)
+				}
 			} else {
 				return nil, Error.Wrap(err)
 			}
 		}
+		_, err = tx.Exec("release savepoint sp;")
```
ifraixedes: nit:
```suggestion
		_, err = tx.Exec("release savepoint sp")
```

satellite/satellitedb/orders.go
ifraixedes: nit:

```suggestion
				_, err = tx.Exec("rollback to savepoint sp")
```

satellite/satellitedb/orders.go
```diff
@@ -266,18 +266,32 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
 	// the case where the order has already been processed.  Duplicates and previously
 	// processed orders are rejected
 	for _, request := range requests {
-		insert := "INSERT INTO used_serials (serial_number_id, storage_node_id) SELECT id, ? FROM serial_numbers WHERE serial_number = ?"
+		// avoid the PG error "current transaction is aborted, commands ignored until end of transaction block" if the below insert fails due any constraint.
+		// aee https://www.postgresql.org/message-id/13131805-BCBB-42DF-953B-27EE36AAF213%40yahoo.com
```
ifraixedes:
```suggestion
		// see https://www.postgresql.org/message-id/13131805-BCBB-42DF-953B-27EE36AAF213%40yahoo.com
```

PR #2870, opened 2019-08-26
satellite/metainfo: Fix downloading non encrypted segments

mniewrzal:
What: Fix setting encrypted key/nonce during `DownloadSegment` in a case when segment was not encrypted.
This is part of adjustments for #2640 

Why: When a segment is not encrypted then `streamMeta.LastSegmentMeta` can be nil and can cause panic.

Please describe the tests:
 - Test 1: existing tests should cover this case
 
Please describe the performance impact: none

## Code Review Checklist (to be filled out by reviewer)
 - [ ] Does the PR describe what changes are being made?
 - [ ] Does the PR describe why the changes are being made?
 - [ ] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [ ] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [ ] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [ ] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [ ] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [ ] Does any documentation need updating?
 - [ ] Do the database access patterns make sense?

PR #2902, opened 2019-08-28
Storage node auto-updater skeleton

mniewrzal:
What: Simple skeleton for storage node auto-updater. Only initial cmd implementation with a simple loop. 

Why: First part for https://storjlabs.atlassian.net/browse/V3-2481

Please describe the tests:
 - Test 1: not yet
 

cmd/auto-updater/main.go
```diff
@@ -0,0 +1,73 @@
...
+func init() {
+	rootCmd.AddCommand(runCmd)
+
+	runCmd.Flags().StringVar(&interval, "interval", "05s", "interval for checking the new version")
+	runCmd.Flags().StringVar(&versionURL, "version-url", "https://version.storj.io/", "version server URL")
+	runCmd.Flags().StringVar(&binaryPath, "binary-path", "", "the storage node installation location")
+}
+
+func cmdRun(cmd *cobra.Command, args []string) (err error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
ifraixedes: Using `context.Background()` is a temporary solution?

mniewrzal:
@ifraixedes each reviewer should add own checkboxes and check them. See https://github.com/storj/storj/pull/2870#issuecomment-526070437

cmd/auto-updater/main.go
mniewrzal: No, I don't think so. Whats the problem with `context.Background()` here?

egonelbre: You can use `cmd.Context()` here.

ifraixedes: Nothing :sweat_smile: I think that I was too asleep when I did the review and I thought that using background we didn't need to call `context.WithCancel`

ifraixedes:

cmd/auto-updater/main.go
mniewrzal: > You can use cmd.Context() here.

I don't see `Context()` in `cobra.Command` methods.

egonelbre: Ah, nevermind it was `process.Ctx(cmd)`

cmd/auto-updater/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	interval   string
+	versionURL string
+	binaryPath string
+)
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+
+	runCmd.Flags().StringVar(&interval, "interval", "05s", "interval for checking the new version")
```
kaloyan-raev: Let's set the default to `6h` so it matches our current recommendation for watchtower.

cmd/auto-updater/main.go
```diff
@@ -0,0 +1,80 @@
...
+	runCmd.Flags().StringVar(&versionURL, "version-url", "https://version.storj.io/", "version server URL")
```
kaloyan-raev: The default should be `https://version.storj.io/release/`. For some reason, I forgot to save this change in the Jira ticket. I updated it now.

cmd/auto-updater/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/spf13/cobra"
+
+	"storj.io/storj/internal/sync2"
+)
+
+var (
+	rootCmd = &cobra.Command{
+		Use:   "auto-updater",
+		Short: "auto-updater for storage node",
```
kaloyan-raev: Should this start with a capital letter? i.e. `Auto-updater for storage node`

cmd/auto-updater/main.go
kaloyan-raev: @mniewrzal Are you going to use `process.Ctx(cmd)` instead of `context.Background()` as @egonelbre  suggested?

mniewrzal: No, executing cmd with our `process.Exec()` is adding a lot of stuff that we don't need. For now I plan to stay as simple as possible. I can use only `process.Ctx(cmd)` to avoid few lines of code but I'm not sure if it's worth it.

cmd/auto-updater/main.go
mniewrzal: Yes, it looks that in other commands we have a capital letter.

PR #2911, opened 2019-08-29
storagenode/storageusage: add summary, rename timestamp to interval_start

rikysya:
What: storagenode/storageusage:  rename timestamp to interval_start, as we now store GB*hours instead of plain GB's `interval_start` is more descriptive in a way that it start point of calcualtions.

Added `Summary`, `SatelliteSummary` queries to receive summaries for all satellites and for particular one respectively.

Simplified queries as more values for the same date will indicate and of previous interval and beginning of the new one instead of `fresh` snapshot of storagenode storage state

Deleted all values from `storage_usage` table as they are not valid anymore

NOTE: should be merged after https://github.com/storj/storj/pull/2858, to ensure only correct data is stored after table vanishing

Please describe the tests:
 - Test 1: ensures calculations works correctly, and we parse table in a expected manner
 - Test 2: ensure we return 0, or empty result when there are no data without any errors
 

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
-			Timestamp:   timeStamp,
+			AtRestTotal:   atRestTotal,
+			IntervalStart: intervalStart,
 		})
 	}
 
 	return stamps, nil
 }
 
+// Summary returns aggregated storage usage across all satellites
```
ifraixedes:
```suggestion
// Summary returns aggregated storage usage across all satellites.
```

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+func (db *storageusageDB) Summary(ctx context.Context, from, to time.Time) (_ float64, err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: nit:

```suggestion
	defer mon.Task()(&ctx, from, to)(&err)
```

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+	var summary sql.NullFloat64
```
ifraixedes: cannot it just be `float64`?

According the query it doesn't seem that it can be nil.

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+
+	query := `SELECT SUM(at_rest_total) 
+				FROM storage_usage
+				WHERE ? <= interval_start AND interval_start <= ?`
+
+	err = db.QueryRowContext(ctx, query, from.UTC(), to.UTC()).Scan(&summary)
```
ifraixedes: We must return this error

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+	return summary.Float64, nil
+}
+
+// SatelliteSummary returns aggregated storage usage for particular satellite
```
ifraixedes:
```suggestion
// SatelliteSummary returns aggregated storage usage for a particular satellite.
```

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+func (db *storageusageDB) SatelliteSummary(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) (_ float64, err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes: nit:
```suggestion
	defer mon.Task()(&ctx)(&err, satelliteID, from, to)
```

storagenode/storagenodedb/storageusage.go
```diff
@@ -123,22 +119,49 @@ func (db *storageusageDB) GetDailyTotal(ctx context.Context, from, to time.Time)
...
+	var summary sql.NullFloat64
```
ifraixedes: cannot it just be `float64`?

According the query it doesn't seem that it can be nil.

storagenode/storageusage/storageusage.go
```diff
@@ -20,13 +20,15 @@ type DB interface {
 	// GetDailyTotal returns daily storage usage stamps summed across all known satellites
 	// for provided time range
 	GetDailyTotal(ctx context.Context, from, to time.Time) ([]Stamp, error)
+	// Summary returns aggregated storage usage across all satellites
```
ifraixedes:
```suggestion
	// Summary returns aggregated storage usage across all satellites.
```

storagenode/storageusage/storageusage.go
```diff
@@ -20,13 +20,15 @@ type DB interface {
...
+	Summary(ctx context.Context, from, to time.Time) (float64, error)
+	// SatelliteSummary returns aggregated storage usage for particular satellite
```
ifraixedes:
```suggestion
	// SatelliteSummary returns aggregated storage usage for a particular satellite.
```

storagenode/storageusage/storageusage.go
```diff
@@ -20,13 +20,15 @@ type DB interface {
...
+	SatelliteSummary(ctx context.Context, satelliteID storj.NodeID, from, to time.Time) (float64, error)
 }
 
-// Stamp is storage usage stamp for satellite at some point in time
+// Stamp is storage usage stamp for satellite from interval start till next interval
```
ifraixedes:
```suggestion
// Stamp is storage usage stamp for satellite from interval start till next interval.
```

storagenode/storagenodedb/storageusage.go
rikysya: It was float before, but I got an error at some point

storagenode/storagenodedb/database.go
```diff
@@ -16,7 +16,7 @@ import (
 	_ "github.com/mattn/go-sqlite3" // used indirectly
 	"github.com/zeebo/errs"
 	"go.uber.org/zap"
-	monkit "gopkg.in/spacemonkeygo/monkit.v2"
+	"gopkg.in/spacemonkeygo/monkit.v2"
```
jenlij: @egonelbre are we still trying to use the monkit package name alias? I know when I autosave + gofmt in Goland it removes the alias because it says it's redundant.

storagenode/storagenodedb/storageusage.go
```diff
@@ -36,12 +36,12 @@ func (db *storageusageDB) Store(ctx context.Context, stamps []storageusage.Stamp
 		return nil
 	}
 
-	query := `INSERT OR REPLACE INTO storage_usage(satellite_id, at_rest_total, timestamp) 
+	query := `INSERT OR REPLACE INTO storage_usage(satellite_id, at_rest_total, interval_start) 
 			VALUES(?,?,?)`
 
 	return db.withTx(ctx, func(tx *sql.Tx) error {
 		for _, stamp := range stamps {
-			_, err = db.ExecContext(ctx, query, stamp.SatelliteID, stamp.AtRestTotal, stamp.Timestamp.UTC())
```
jenlij: was this an error to use db.ExecContext originally?

storagenode/storageusage/storageusage.go
jenlij: you could also say "until next interval"

storagenode/storagenodedb/database.go
egonelbre: The issue is that the path contains "monkit.v2", module name is `monkit.v2` and package name `monkit`.

Goland seems to take the easy approach by ignoring versions at the end of the import. However, goimports uses the rule that when import path (ignoring module versioning) and package name differ, then it uses an alias to keep things clear.

If `goimports -w -local storj.io` formats the code the same way, then I'm fine with either way. If it's not, then let's prefer `goimports`, because that's the default for VS Code.

storagenode/storagenodedb/storageusage.go
rikysya: yes, it was typo it seems

storagenode/storagenodedb/storageusage.go
ifraixedes: it sounds weird

egonelbre: It could be a `var summary *float64` and then check for the `nil`.

But, yes, when there is no result it will fail to unmarshal into a `float64`.

storagenode/storageusage/storageusage_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"storj.io/storj/storagenode/storageusage"
+)
+
+func TestStorageUsage(t *testing.T) {
+	const (
+		satelliteNum = 10
+		days         = 30
+	)
+
+	now := time.Now().UTC()
```
egonelbre: For tests don't use `UTC()` the databases should handle the conversion internally. It's clearer to enforce this in the database implementation rather than outside.

storagenode/storagenodedb/storageusage.go
egonelbre: `SUM(at_rest_total)` returns `NULL` when the table is empty.

ifraixedes: Thanks, that's what I was missing, I thought that when no rows it will return nothing (i.e. `ErrNoRows`)

storagenode/storagenodedb/storageusage.go
rikysya: @egonelbre if we know that on empty result we won't to return `0` and no error, then `sql.NullTime` let us avoid extra check. Also it seems that it is preferred approach since we have this type provided by `sql` pkg

egonelbre: @rikysya, sounds good.

PR #2913, opened 2019-08-29
use a nested module to keep track of tool dependencies

zeebo:
**What**:
- use go.mod to track dependencies
- clean up some little warnings with xunit
- make the gerrit jenkinsfile match the public one better
- make the dockerfile also use modules to install deps

**Why**: So that we have a single way to manage the dependencies of all our stuff

Please describe the performance impact: Probably none


Jenkinsfile.gerrit
```diff
@@ -39,17 +40,17 @@ pipeline {
...
-                        sh 'go run ./scripts/check-copyright.go'
-                        sh 'go run ./scripts/check-large-files.go'
-                        sh 'go run ./scripts/check-imports.go -race ./...'
-                        sh 'go run ./scripts/check-peer-constraints.go -race -fail=false'
-                        sh 'go run ./scripts/protobuf.go --protoc=$HOME/protoc/bin/protoc lint'
-                        sh 'go run ./scripts/protobuf.go --protoc=$HOME/protoc/bin/protoc check-lock'
-                        sh 'go run ./scripts/atomicalign.go ./...'
-                        sh 'go run ./scripts/check-errs.go ./...'
+                        sh 'cd ./scripts; go run ./check-copyright'
+                        sh 'cd ./scripts; go run ./check-large-files'
```
egonelbre: These both above start at work dir, so by cd-ing into the folder it messes up things.

But `GOMOD=scripts/go.mod` could work, to avoid needing to be in scripts dir.

scripts/go.mod
```diff
@@ -0,0 +1,46 @@
...
+	github.com/kylelemons/godebug v1.1.0
+	github.com/loov/leakcheck v0.0.3
+	github.com/mfridman/tparse v0.7.4
+	github.com/nilslice/protolock v0.14.0
+	github.com/zeebo/errs v1.2.2
+	golang.org/x/tools v0.0.0-20190829051458-42f498d34c4d
+	gopkg.in/spacemonkeygo/dbx.v1 v1.0.0-20190212172312-3af5e1fc2659
+)
+
+// golangci-lint has some dependencies on incorrectly published modules so we
```
egonelbre: I wonder whether it would be easier to just depend on the exact binary version rther than do the replacements?

egonelbre: Yeah. I mean `curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b ${GOPATH}/bin v1.17.1` downloaded the exact version without having to lock in other repository versions. The replacement thing feels fragile.

egonelbre: Ok, let's leave it. But it still feels fragile :)

Jenkinsfile.gerrit
egonelbre: Hmm, I got it to work somehow without needing to cd into the directory... I mean without having to change the tools. I'll try to remember what I did.

egonelbre: Anyways, let's make the scripts take the path as an argument rather than hardcoding the "." in that case. Maybe it makes it more obvious what's happening.

I'll see on Monday whether I had a better solution.

pkg/pb/gen.go
```diff
@@ -3,4 +3,4 @@
 
 package pb
 
-//go:generate go run ../../scripts/protobuf.go generate
+//go:generate sh -c "cd ../../scripts; go run ./protobuf generate"
```
egonelbre: This will fail on windows :(

zeebo: Dang, because no shell. That's a problem.

egonelbre: We can go with this for now.

Jenkinsfile.gerrit
egonelbre: Apparently, no. My final version used `-mod=readonly` avoid modifying `go.mod`.

pkg/pb/gen.go
egonelbre: The best I can think of is having a go script that doesn't have any external dependencies and handles `go.mod` related compilation and things. e.g.

```
//go:generate modrun ../../scripts/protobuf generate
```

zeebo:
I don't think I want to do this anymore.

PR #2917, opened 2019-08-29
uplink/ecclient: change delete logs from err to debug level

navillasa:
What: 
- change delete logs from err to debug level
- remove the sugared logger to speed things up

Why:
- the delete errors are normal and don't need to be exposed to clients in minio

Please describe the tests:
 - Test 1:
 - Test 2:
 

uplink/ecclient/client.go
```diff
@@ -401,14 +401,14 @@ func (ec *ecClient) Delete(ctx context.Context, limits []*pb.AddressedOrderLimit
 				Address: addressedLimit.GetStorageNodeAddress(),
 			})
 			if err != nil {
-				ec.log.Sugar().Errorf("Failed dialing for deleting piece %s from node %s: %v", limit.PieceId, limit.StorageNodeId, err)
+				ec.log.Sugar().Debugf("Failed dialing for deleting piece %s from node %s: %v", limit.PieceId, limit.StorageNodeId, err)
```
mobyvb: Since you're updating these lines anyway, can you remove the sugar logger? You could do something like this instead:
`ec.log.Debug("Failed dialing for deleting piece from node", zap.String("piece id", limit.PieceId.String()), zap.String("node id", limit.StorageNodeId.String()), zap.Error(err))`

mobyvb: There are a lot of other logs in this file that should eventually be updated in the same way, but that doesn't need to be a part of this PR unless you want.

navillasa: okay I went ahead and removed the sugared logger from the file

navillasa:
> It looks good to me however I would like to know
> 
> > the delete errors are normal and don't need to be exposed to clients in minio
> 
> Why are those errors considered normal?

@ifraixedes okay actually looks like they're not normal errors, so that part of what I said is wrong, but I talked to @stefanbenten and it doesn't seem like they're errors that should be exposed to users.

PR #2925, opened 2019-08-30
Repair with hashes

mobyvb:
What: 
Create a specialized ecclient for the repairer that does additional verification:
* Gets uplink-signed piece hashes and signed original order limits from storagenodes during download
* Verifies piece hash and order limit signature
* Calculates piece hash of downloaded data from storagenode, and verifies that it matches the uplink-signed one
* (optional?) Verifies that piece public key on original order limit matches the one used for the other pieces
* Reconstructs segment with downloaded pieces, and repairs to new nodes in the same way we currently do (`ecRepairer.Repair` does the same thing as `ecClient.Repair`, which can be removed now)

Why:
Repair is expensive so we want to download from the smallest possible number of storagenodes to reconstruct the segment. However, if we use the minimum RS number of pieces, we have no way of automatically detecting errors in pieces, and the reconstructed segment could be incorrect without our knowledge. By using hashes to check the validity of pieces before we decode the segment, we can be sure of the authenticity of a piece, and use the minimum number of pieces to get the segment.

Unfortunately, this requires us to bring each piece entirely into memory since we need to download the entire piece before we can verify the hash.

https://storjlabs.atlassian.net/browse/V3-2396
https://storjlabs.atlassian.net/browse/V3-2395
https://storjlabs.atlassian.net/browse/V3-2394

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	if int64(len(pieceBytes)) != pieceSize {
+		return nil, Error.New("didn't download the correct amount of data, want %d, got %d", pieceSize, len(pieceBytes))
+	}
+
+	// get signed piece hash and original order limit
+	hash, originalLimit := downloader.GetHashAndLimit()
+	if hash == nil {
+		return nil, Error.New("Hash was not sent from storagenode.")
```
mniewrzal: error should start from lower case, and should not contain dot at the end

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+	if originalLimit == nil {
+		return nil, Error.New("Original order limit was not sent from storagenode.")
```
mniewrzal: error should start from lower case, and should not contain dot at the end

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,468 @@
...
+
+	// Ensure timer is stopped
+	_ = timer.Stop()
+
+	// TODO: clean up the partially uploaded segment's pieces
+	defer func() {
+		select {
+		case <-ctx.Done():
+			err = Error.New("repair cancelled")
+			// ec.Delete(context.Background(), nodes, pieceID, pba.SatelliteId), //TODO
```
mniewrzal: can you remove this old comment?

satellite/repair/repairer/repairer.go
```diff
@@ -49,16 +49,15 @@ type Service struct {
 }
 
 // NewService creates repairing service
-func NewService(log *zap.Logger, queue queue.RepairQueue, config *Config, interval time.Duration, concurrency int, transport transport.Client, metainfo *metainfo.Service, orders *orders.Service, cache *overlay.Service) *Service {
-	client := ecclient.NewClient(log.Named("ecclient"), transport, config.MaxBufferMem.Int())
-	repairer := NewSegmentRepairer(log.Named("repairer"), metainfo, orders, cache, client, config.Timeout, config.MaxExcessRateOptimalThreshold)
+func NewService(log *zap.Logger, queue queue.RepairQueue, config *Config, transport transport.Client, metainfo *metainfo.Service, orders *orders.Service, cache *overlay.Service, satelliteSignee signing.Signee) *Service {
+	repairer := NewSegmentRepairer(log.Named("repairer"), metainfo, orders, cache, transport, config.Timeout, config.MaxExcessRateOptimalThreshold, satelliteSignee)
```
mniewrzal: I would make `SegmentRepairer` an argument to `NewService`. It will reduce number of parameters and will decouple repairer from service a little bit.

uplink/eestream/unsafe_rs.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *unsafeRSScheme) Encode(input []byte, output func(num int, data []byte)) (
+	err error) {
+	return s.fc.Encode(input, func(s infectious.Share) {
+		output(s.Number, s.Data)
+	})
+}
+
+func (s *unsafeRSScheme) Decode(out []byte, in map[int][]byte) ([]byte, error) {
+	shares := make([]infectious.Share, 0, len(in))
```
mniewrzal: I think using lenght not capacity should be more optimal:
```
	shares := make([]infectious.Share, len(in))
	for num, data := range in {
		shares[num] = infectious.Share{Number: num, Data: data}
	}
```

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,468 @@
...
+	var calculatedHash []byte
+	pieceBytes := make([]byte, 0, pieceSize)
+	buffer := make([]byte, bufferSize)
+	newHash := pkcrypto.NewHash()
+
+	for {
+		// download full piece
+		n, readErr := downloader.Read(buffer)
+		if readErr == io.EOF {
+			calculatedHash = newHash.Sum(buffer[:n])
```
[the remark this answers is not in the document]
mobyvb: We were originally doing something like that, but we thought maybe it would be preferable to hash the piece as it was being read in rather than after `ReadAll`. Now, I don't know if there is any advantage to how we are currently doing it.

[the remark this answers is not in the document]
mniewrzal: only thing I can think off is performance and readability. Maybe you can compare if there is a difference between both ways with simple benchmark?

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,468 @@
...
+		return nil, Error.New("Original order limit was not sent from storagenode.")
+	}
+
+	// verify the hashes from storage node
+	if err := verifyPieceHash(ctx, originalLimit, hash, calculatedHash); err != nil {
+		return nil, err
+	}
+
+	// verify order limit from storage node is signed by the satellite
+	if err := verifyOrderLimitSignature(ctx, ec.satelliteSignee, originalLimit); err != nil {
```
mniewrzal: `verifyOrderLimitSignature` should be before `verifyPieceHash` for cases when `limit.UplinkPublicKey` was manipulated

satellite/repair/repairer/ec.go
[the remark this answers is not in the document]
VinozzZ:
```
// Sum appends the current hash to b and returns the resulting slice.
// It does not change the underlying hash state.
    Sum(b []byte) []byte
```
so the `b` here should already be a hash and `Sum` only appends that to the current hash not actually hashing it and then append? Just want to make sure I understand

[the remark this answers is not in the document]
mniewrzal: `Sum` just appends `b` to current hash (it doesn't have to be hash also, might be just some bytes) without hash processing and returns `hash + b`. In our code `Sum` has mainly `nil` param because we want just pure hash without additions.

uplink/eestream/unsafe_rs.go
VinozzZ: We can't do that since `num` in the map isn't the index but the piece number

mniewrzal: ah yes, my mistake. It still can be done with additional index but I think it won't give so much improvement.

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,450 @@
...
+		return nil, err
+	}
+
+	downloader, err := ps.Download(ctx, limit.GetLimit(), privateKey, 0, pieceSize)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { err = errs.Combine(err, downloader.Close()) }()
+
+	var calculatedHash []byte
```
mniewrzal: this is not needed anymore

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,450 @@
...
+	pieceBytes, err := ioutil.ReadAll(downloader)
+	if err != nil {
+		return nil, err
+	}
+	calculatedHash = pkcrypto.SHA256Hash(pieceBytes)
```
mniewrzal: this should be moved just before place its used. Calculating hash is CPU consuming and if we can avoid it with simple validation we need to do this.

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,450 @@
...
+
+	if int64(len(pieceBytes)) != pieceSize {
+		return nil, Error.New("didn't download the correct amount of data, want %d, got %d", pieceSize, len(pieceBytes))
+	}
+
+	// get signed piece hash and original order limit
+	hash, originalLimit := downloader.GetHashAndLimit()
+	if hash == nil {
+		return nil, Error.New("hash was not sent from storagenode.")
```
mniewrzal: errors shouldn't end with dot

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	if originalLimit == nil {
+		return nil, Error.New("original order limit was not sent from storagenode.")
```
mniewrzal: errors shouldn't end with dot

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,449 @@
...
+	if !unique(limits) {
+		return nil, nil, Error.New("duplicated nodes are not allowed")
+	}
+
+	readers, err := eestream.EncodeReader(ctx, ec.log, ioutil.NopCloser(data), rs)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	type info struct {
```
navillasa: I think this is a carry over from the ec Repair function, but I think a more descriptive name for this struct and the `infos` channel could make this part easier to understand.

satellite/repair/repairer/ec.go
```diff
@@ -0,0 +1,449 @@
...
+
+	mon.IntVal("repair_segment_pieces_total").Observe(int64(pieceCount))
+	mon.IntVal("repair_segment_pieces_successful").Observe(int64(successfulCount))
+	mon.IntVal("repair_segment_pieces_failed").Observe(int64(failureCount))
+	mon.IntVal("repair_segment_pieces_canceled").Observe(int64(cancellationCount))
+
+	return successfulNodes, successfulHashes, nil
+}
+
+// copied from ecclient
```
navillasa: are we keeping these comments long-term? I'm not sure if they're necessary

VinozzZ:
> looks good, I would just spend a moment to think if we can somehow test that e.g. hash doesn't match

test added

satellite/repair/repairer/segments.go
```diff
@@ -101,7 +102,7 @@ func (repairer *SegmentRepairer) Repair(ctx context.Context, path storj.Path) (s
 
 	numHealthy := len(pieces) - len(missingPieces)
 	// irreparable piece, we need k+1 to detect corrupted pieces
-	if int32(numHealthy) < pointer.Remote.Redundancy.MinReq+1 {
+	if int32(numHealthy) < pointer.Remote.Redundancy.MinReq {
```
mniewrzal: is comment up to date when `+1` is removed?

satellite/repair/checker/checker.go
```diff
@@ -163,7 +163,7 @@ func (checker *Checker) updateIrreparableSegmentStatus(ctx context.Context, poin
 
 	// we repair when the number of healthy pieces is less than or equal to the repair threshold
 	// except for the case when the repair and success thresholds are the same (a case usually seen during testing)
-	if numHealthy > redundancy.MinReq && numHealthy <= redundancy.RepairThreshold && numHealthy < redundancy.SuccessThreshold {
+	if numHealthy >= redundancy.MinReq && numHealthy <= redundancy.RepairThreshold && numHealthy < redundancy.SuccessThreshold {
```
mniewrzal: why we need this change? and rest in this file?

mobyvb: Before repair with hashes, a segment was considered irreparable if it had `MinReq` or fewer pieces. This is because we needed at least `MinReq+1` pieces for the reed solomon error correction to work correctly. Now that we are verifying pieces with hashes before decoding into a segment, the minimum number exactly is sufficient to repair correctly.
The changes in this file ensure that we
1. Add to repair if we have exactly the minimum healthy number of pieces
2. Do not add to the irreparable db if we have exactly the minimum healthy number of pieces

satellite/repair/repairer/segments.go
mobyvb: it is updated now

satellite/repair/repairer/ec.go
mobyvb: Hmm I am not sure what a better name for the struct and channel would be, but maybe adding a comment could help?

satellite/repair/repairer/ec.go
mobyvb: I think they can be removed. I was hoping we could get rid of duplicate code here and in ecclient, but I don't know of a good way.

mobyvb: Comments removed

satellite/repair/repairer/ec.go
mobyvb: Added some comments describing the purpose of the struct + channel

satellite/repair/repairer/ec.go
navillasa: maybe someone else has thoughts on this? maybe could make a ec_helpers.go file or something?

satellite/repair/repairer/ec.go
navillasa: thank you, I find those helpful!

PR #2927, opened 2019-08-30
design/blueprints

jtolds:
we've decided to rename design docs to blueprints to indicate that we *don't* intend to keep blueprints up to date once completed. instead, the final step of blueprint implementation is to archive the blueprint and update actual documentation elsewhere (potentially copying and pasting most of the blueprint). this PR updates the template and readme


docs/blueprints/README.md
```diff
@@ -37,22 +46,24 @@ However, it is expected that there should be feedback from the engineering team,
...
+* **Implementation** describes the steps to complete this blueprint. It should contain a rough outline of tasks. If necessary, there can be additional details about the changes to the codebase and process.
+
+* **Wrap up** describes the steps needed
 
 * **Open Issues** contains open questions that the author did not know how to solve.
 
 ## Format
 
-The design document is intended to be read by outside of Storj, hence avoid acronyms that aren't common in the general developer community. Prefer simple sentences. Active voice is usually easier to read. Overall, be clear.
+The blueprint is intended to be read by developer contributors outside of Storj, but not the wider public. It is okay to assume some familiarity with the Storj project. Try to avoid acronyms that aren't common in the general developer community. Prefer simple sentences. Active voice is usually easier to read. Overall, be clear.
```
egonelbre: I wonder should we clarify expected level of quality? e.g. run the document through http://www.hemingwayapp.com (or Grammarly Pro) and simplify sentences when possible.

docs/blueprints/README.md
```diff
@@ -37,22 +46,24 @@ However, it is expected that there should be feedback from the engineering team,
...
+* **Background** gives an overview of the background why this blueprint exists. It describes the problems the blueprint tries to solve. It describes the goals of the design.
 
 * **Design** contains the solution and its parts. The level of detail should correspond to the level of risk. The more problems a wrong solution would cause, the more detailed should be the description. The design should describe the solution to the degree it is usable by the end-user.
 
 * **Rationale** section describes the alternate approaches and trade-offs. It should be clear why the proposed design was chosen among the alternate solutions.
 
-* **Implementation** describes the steps to complete this design document. It should contain a rough outline of tasks. If necessary, there can be additional details about the changes to the codebase and process.
+* **Implementation** describes the steps to complete this blueprint. It should contain a rough outline of tasks. If necessary, there can be additional details about the changes to the codebase and process.
+
+* **Wrap up** describes the steps needed
```
ifraixedes: it's 2 words or one word?

Here is 2 word however in the template is one word https://github.com/storj/storj/pull/2927/files#diff-25e11c5b04ae10c94492b0191cc65769R23

docs/blueprints/README.md
jtolds: i'm reluctant to specify beyond what the reviewers approve

docs/blueprints/README.md
jtolds: i'm not sure it matters? it's fine to have both

jtolds:
trivial rebase

docs/blueprints/README.md
ifraixedes: fine to me, I wanted to know if there was a typo in any of them.

ppenguin:
@jtolio I saw the article about `storj-mount` but didn't find any source code. If possible, I'd like to try my hand at contributing/improving it, would that be possible? Thanks in advance!

ifraixedes:
@ppenguin the `mount` command is deprecated. You should use rclone.

By the way, why have you decided to ask this question, which I don't see how is related to this PR that was closed a long time ago?

ppenguin:
@ifraixedes Thanks for your feedback. I merely used this PR as a point of contact since it's the only link referred to from [this document](https://github.com/storj/storj/blob/master/docs/blueprints/mountable-drive.md).

As I understand (I have used `rclone` long ago), `rclone` offers syncing functionality of exisitng local directories to "the cloud" and vice-versa. What I'm looking for is an actual filesystem interface, to access "the cloud" (a `storj` bucket in this case) as a local (mounted) directory.

If I understand correctly, one current possibility would be to use `s3fs-fuse` in combination with the storj-S3 gateway, where the latter basically converts "S3 protocol" to "storj protocol" on-the-fly. Ideally I'd like to avoid the need for this extra service though, so I would consider to implement a workalike/extension of `s3fs-fuse` or `goofys` if this functionality doesn't exist yet.  (And for that, my intention would then be to use `uplink` and one of the `fuse` programs as starting point).

ifraixedes:
@ppenguin thanks for the explanation.

Now, I can understand why you posted here, but I would suggest moving this proposal to a more visible place. Yes, that document was moved to another directory in this PR but with all the rest so it doesn't keep a good context for your message.

Better options that I personally see to post your considerations are:

* A GH issue
* A post in the forum

I discard our [ideas portal](https://ideas.storj.io/) as you're considering to implement it rather than proposed to Storj for being implemented

If you post in both places, please keep a link on each other.

I appreciate your interest in thinking in implementing the extension.

ppenguin:
@ifraixedes Thanks, appreciate for bearing with me ;)
I'll add an issue on GH, both as "a call for action" and as an "intention to contribute", we might get to the goal quicker then... (And maybe still for good measure to the ideas portal).
Cheers!

egonelbre:
Note rclone does have mount. See https://rclone.org/commands/rclone_mount/.

ppenguin:
@egonelbre Thanks for the info (how could I have missed that...). I'll check out that first then, and continue this endevour if it doesn't satisfy my case. Thanks again for the heads up!

JessicaGreben:
@ppenguin just a heads up, I'm currently working on a fuse fs for storj. I will be adding the code for that here soon https://github.com/storj-thirdparty

ppenguin:
@JessicaGreben Thanks for the heads up, that's great to hear! I already did a rudimentary test of `rclone` in this context (setup and mounting as such was trivial), but I guess even if it covers the sought after functionality, it could be a bit overkill (since it is a huge package deal in terms of functionality) for some use cases. (And I have not yet done performance testing, which I need to and will do in any case, that might also make the picture more clear).

May I ask whether this (functionality/performance) was your reason to work on a dedicated solution, or are there other reasons?

JessicaGreben:
@ppenguin My reason for writing a fuse fs for Storj is because its fun! Also I'm going to use it in a Kubernetes CSI driver that I'm building...this driver will provide native support to mount Storj buckets into Kubernetes pods.

ppenguin:
@JessicaGreben Nice! Consider me willing as an alpha-tester ;). Since likely I'll be interested in a lean solution for mounting `storj` also on constrained devices, that might be a great solution.

JessicaGreben:
@ppenguin sounds good, thanks for offering to help. I will make sure to ping you when I push the code up. I will aim to do that in the next few days.

ppenguin:
@JessicaGreben 
Just checking in to ask whether you had any luck yet with your `storj` fuse fs...
Cheers!

JessicaGreben:
@ppenguin things are going a little slower than expected, as usual in software. im just writing tests now. will have it up in a few. i will ping you once it up.

PR #2934, opened 2019-09-02
storagenode/buildscript: Overwrite existing docker manifest file

littleskunk:
What: Overwrite existing docker manifest file.

Why: Because the script will fail on the second run otherwise.

Please describe the tests:
 - Test 1:
 - Test 2:
 

scripts/deploy-storagenode.sh
```diff
@@ -6,7 +6,7 @@ set -euo pipefail
 : "${TAG:?Must be set to the gitish version of the release without architecture}"
 
 for v in alpha arm beta; do
-	docker manifest create storjlabs/storagenode:$v \
+	docker manifest create --amend storjlabs/storagenode:$v \
 	storjlabs/storagenode:${TAG}-amd64 \
```
ifraixedes: As far as I understood the name of the images in `docker manifest` is this one and the rest listed below.

These images names don't change, so I'm wondering if we are using the same images for `alpha` and `beta`?

On the other hand, what `arm` means in a list of versions, when this is building images for different architectures?

littleskunk: I don't understand the magic myself. Don't ask me^^

PR #2936, opened 2019-09-03
satellite/satellitedb: always release savepoint in process orders

ifraixedes:
What: Always release savepoints and wrap errors.

Why: Not releasing a savepoint can cause nasty deadlocks and not wrapping errors make difficult to identify from where they are generated.

Please describe the tests: No new tests. I haven't found a way how to have a test which could reproduce the problem of not releasing the savepoint. 

Please describe the performance impact: Wrapping errors always produce some performance impact, but it's necessary and the impact should be negligible.

### DIFF
--- a/satellite/satellitedb/orders.go
+++ b/satellite/satellitedb/orders.go
@@ -285,7 +285,8 @@ func (db *ordersDB) ProcessOrders(ctx context.Context, requests []*orders.Proces
 					return nil, Error.Wrap(err)
 				}
 			} else {
-				return nil, Error.Wrap(err)
+				_, rerr := tx.Exec("rollback to savepoint sp")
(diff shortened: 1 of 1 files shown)


thepaul:
What is the evidence that not releasing savepoints can cause nasty deadlocks? That sounds like a major bug in PostgreSQL if true. Committing or rolling back the enclosing transaction should always release any savepoints.

ifraixedes:
> What is the evidence that not releasing savepoints can cause nasty deadlocks?

I searched on the topic and found people replying that not releasing them can cause those.

> That sounds like a major bug in PostgreSQL if true. Committing or rolling back the enclosing transaction should always release any savepoints.

Yes, probably, but I missed the rollback in the `defer` statement, so I didn't realize it.

I can rollback that commit and leave the other one which wrap the errors.

PR #2932, opened 2019-09-02
cmd/certificates: separate subcommand configs

bryanchriswhite:
What: Separate the config structs for the certificates CLI.

Why: To prepare for working on the certificates service in [V3-2382](https://storjlabs.atlassian.net/browse/V3-2382).

Please describe the tests: [`./scripts/test-certificates.sh`](https://github.com/storj/storj/blob/217dee231250ccd90cd5587671db869acadb71c3/scripts/test-certificates.sh)(from #2897, this PR's base) should cover this.
 
Please describe the performance impact:

## Code Review Checklist (to be filled out by reviewer)
 - [x] Does the PR describe what changes are being made?
 - [x] Does the PR describe why the changes are being made?
 - [x] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [x] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [x] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [x] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [x] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [x] Does any documentation need updating?
 - [x] Do the database access patterns make sense?

PR #2938, opened 2019-09-03
{cmd,pkg}/certificates: service refactor

bryanchriswhite:
What: 
- Refactor the certificates service to follow our architecture document.
- Move certificate authorization stuff to it's own package.
- Refactor CLI configs.

Why:

The certificates service was never refactored to match the rest of the services/subsystems after the architecture design was updated and it's more difficult to work on as a result. This refactor will make it easier to finish [V3-2382](https://storjlabs.atlassian.net/browse/V3-2382).

Please describe the tests:

So [I talked about moving the tests in `scripts/test-certificates.sh` to go tests](https://github.com/storj/storj/pull/2897#issuecomment-527434331) but I would do that in a future PR still as this one's pretty big already since we're moving some large files.
 

bryanchriswhite:
based on #2932

cmd/certificates/claims.go
```diff
@@ -33,23 +33,15 @@ var (
...
 	}
-
-	claimsExportCfg struct {
-		Signer certificates.CertServerConfig
-		Raw    bool `default:"false" help:"if true, the raw data structures will be printed"`
-	}
-
-	claimsDeleteCfg struct {
-		Signer certificates.CertServerConfig
-	}
```
bryanchriswhite: moved

cmd/certificates/main.go
```diff
@@ -10,21 +10,12 @@ import (
...
 )
 
-// CertificatesServerFlags defines certificate server configuration
-type CertificatesServerFlags struct {
-	Identity identity.Config
-	Server   server.Config
-
-	Signer certificates.CertServerConfig
-}
-
```
bryanchriswhite: redundant

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package authorizations
```
bryanchriswhite: moved from `certificates` package

pkg/certificates/authorizations/authorizations_test.go
```diff
@@ -567,122 +564,6 @@ func TestToken_Equal(t *testing.T) {
 	assert.False(t, t1.Equal(&t2))
 }
 
-// TODO: test sad path
-func TestCertificateSigner_Sign_E2E(t *testing.T) {
```
bryanchriswhite: moved to peer_test.go

pkg/certificates/authorizations/authorizations_test.go
```diff
@@ -733,91 +614,15 @@ func TestNewClient(t *testing.T) {
...
-		client, err := NewClientFrom(pbClient)
+		client, err := certificatesclient.NewClientFrom(pbClient)
 		assert.NoError(t, err)
 		assert.NotNil(t, client)
 
 		defer ctx.Check(client.Close)
 	})
 }
 
-func TestCertificateSigner_Sign(t *testing.T) {
```
bryanchriswhite: moved to peer_test.go

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package certificatesclient
```
bryanchriswhite: moved from `certificates` package

cmd/certificates/auth.go
```diff
@@ -170,8 +170,8 @@ func writeAuthInfo(ctx context.Context, authDB *certificates.AuthorizationDB, em
 	return nil
 }
 
-func writeTokenInfo(claimed, open certificates.Authorizations, w io.Writer) error {
-	groups := map[string]certificates.Authorizations{
+func writeTokenInfo(claimed, open authorizations.Authorizations, w io.Writer) error {
+	groups := map[string]authorizations.Authorizations{
```
mniewrzal: `authorizations.Authorizations` look a little bit weird :)

cmd/certificates/main.go
```diff
@@ -63,7 +61,17 @@ func cmdRun(cmd *cobra.Command, args []string) error {
...
+	}
+
+	signer, err := runCfg.Signer.Load()
+	if err != nil {
+		return err
+	}
+
+	authorizationDB, err := authorizations.NewDBFromCfg(runCfg.Authorizations)
+	if err != nil {
+		return errs.New("Error opening authorizations database: %+v", err)
```
mniewrzal: errors should start lower case

cmd/certificates/main.go
```diff
@@ -74,7 +82,8 @@ func cmdRun(cmd *cobra.Command, args []string) error {
 		err = errs.Combine(err, revocationDB.Close())
 	}()
 
-	return runCfg.Server.Run(ctx, zap.L(), identity, revocationDB, nil, runCfg.Signer)
+	peer, err := certificates.New(zap.L(), identity, signer, authorizationDB, revocationDB, &runCfg)
```
mniewrzal: I think this `err` is not handled

cmd/identity/main.go
```diff
@@ -57,7 +57,7 @@ var (
 		Concurrency    uint   `default:"4" help:"number of concurrent workers for certificate authority generation"`
 		ParentCertPath string `help:"path to the parent authority's certificate chain"`
 		ParentKeyPath  string `help:"path to the parent authority's private key"`
-		Signer         certificates.CertClientConfig
+		Signer         certificatesclient.ClientConfig
```
mniewrzal: Why separate package? why not just `certificates.ClientConfig`?

cmd/certificates/auth.go
bryanchriswhite: What do you suggest?

cmd/identity/main.go
bryanchriswhite: https://github.com/storj/docs/blob/master/code/Endpoint.md#client

cmd/certificates/main.go
bryanchriswhite: oops

cmd/certificates/main.go
bryanchriswhite: Will change, but this was borrowed from the revocation db error messages that we have in all the commands, so those have the same issue.

cmd/certificates/auth.go
mniewrzal: thats main problem, no idea at the moment.

cmd/identity/main.go
mniewrzal: I see. Most probably `peer` related things should go to `certificates` in root dir (like `versioncontrol`). Then client can stay in `pkg/certificates`. @egonelbre

pkg/certificates/authorizations/db.go
```diff
@@ -0,0 +1,271 @@
...
+	)
+	for i := 0; i < count; i++ {
+		auth, err := NewAuthorization(userID)
+		if err != nil {
+			authErrs.Add(err)
+			continue
+		}
+		newAuths = append(newAuths, auth)
+	}
+	if err := authErrs.Err(); err != nil {
```
mniewrzal: I think this can be just `if authErrs.Err() != nil {`

pkg/certificates/service.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/grpc/peer"
+
+	"storj.io/storj/pkg/certificates/authorizations"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+)
+
+// Certificates implements pb.CertificatesServer
+type Certificates struct {
```
mniewrzal: It looks that this file should be more `endpoint.go` and this struct `Endpoint`.

bryanchriswhite: Yea, you're right.

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Token is a userID and a random byte array, when serialized, can be used like
+// a pre-shared key for claiming certificate signatures.
+type Token struct {
+	// NB: currently email address for convenience
+	UserID string
+	Data   [tokenDataLength]byte
+}
+
+// ClaimOpts hold parameters for claiming an authorization
```
ifraixedes:
```suggestion
// ClaimOpts hold parameters for claiming an authorization.
```

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+	SignedChainBytes [][]byte
+}
+
+func init() {
+	gob.Register(&ecdsa.PublicKey{})
+	gob.Register(&rsa.PublicKey{})
+	gob.Register(elliptic.P256())
+}
+
+// NewAuthorization creates a new, unclaimed authorization with a random token value
```
ifraixedes:
```suggestion
// NewAuthorization creates a new, unclaimed authorization with a random token value.
```

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, ErrInvalidToken.New("data size mismatch")
+	}
+	t := &Token{
+		UserID: userID,
+	}
+	copy(t.Data[:], data)
+	return t, nil
+}
+
+// Unmarshal deserializes a set of authorizations
```
ifraixedes:
```suggestion
// Unmarshal deserializes a set of authorizations.
```

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+func (a *Authorizations) Unmarshal(data []byte) error {
+	decoder := gob.NewDecoder(bytes.NewBuffer(data))
+	if err := decoder.Decode(a); err != nil {
+		return ErrAuthorization.Wrap(err)
+	}
+	return nil
+}
+
+// Marshal serializes a set of authorizations
```
ifraixedes:
```suggestion
// Marshal serializes a set of authorizations.
```

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// String implements the stringer interface and prevents authorization data
+// from completely leaking into logs and errors.
+func (a Authorization) String() string {
+	fmtLen := strconv.Itoa(len(a.Token.UserID) + 7)
+	return fmt.Sprintf("%."+fmtLen+"s..", a.Token.String())
+}
+
+// Equal checks if two tokens have equal user IDs and data
```
ifraixedes:
```suggestion
// Equal checks if two tokens have equal user IDs and data.
```

pkg/certificates/authorizations/db.go
ifraixedes: I think that's used in this way for not calling `authErrs.Err()` twice but the outcome is the same.

pkg/certificates/authorizations/db.go
```diff
@@ -0,0 +1,271 @@
...
+	if err := authDB.add(ctx, userID, newAuths); err != nil {
+		return nil, err
+	}
+
+	return newAuths, nil
+}
+
+// Get retrieves authorizations by user ID.
+func (authDB *DB) Get(ctx context.Context, userID string) (_ Authorizations, err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx, userID)(&err)
```

pkg/certificates/authorizations/db.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// Close closes the authorization database's underlying store.
+func (authDB *DB) Close() error {
+	return ErrAuthorizationDB.Wrap(authDB.db.Close())
+}
+
+// Create creates a new authorization and adds it to the authorization database.
+func (authDB *DB) Create(ctx context.Context, userID string, count int) (_ Authorizations, err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx, userID, count)(&err)
```

pkg/certificates/authorizations/db.go
```diff
@@ -0,0 +1,271 @@
...
+		if auth.Token.Equal(token) {
+			auths[i].Claim = nil
+			return authDB.put(ctx, token.UserID, auths)
+		}
+	}
+	return errs.New("token not found in authorizations DB")
+}
+
+func (authDB *DB) add(ctx context.Context, userID string, newAuths Authorizations) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx, userID)(&err)
```

pkg/certificates/authorizations/db.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return err
+	}
+
+	auths = append(auths, newAuths...)
+	return authDB.put(ctx, userID, auths)
+}
+
+func (authDB *DB) put(ctx context.Context, userID string, auths Authorizations) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx, userID)(&err)
```

pkg/certificates/certificatesclient/client.go
ifraixedes: I would call this package just `client` the problem that I see is that if you need to import several clients then you have to use package aliases.

Do we always use `{parent package name}client`?

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/peertls/tlsopts"
+	"storj.io/storj/pkg/transport"
+)
+
+var mon = monkit.Package()
+
+// ClientConfig is a config struct for use with a certificate signing service client
```
ifraixedes:
```suggestion
// ClientConfig is a config struct for use with a certificate signing service client.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+type ClientConfig struct {
+	Address string `help:"address of the certificate signing rpc service"`
+	TLS     tlsopts.Config
+}
+
+// Client implements pb.CertificateClient
```
ifraixedes:
```suggestion
// Client implements pb.CertificateClient.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (C) 2019 Storj Labs, Inc.
```
ifraixedes: Please can you add a `doc.go` file in this package with a brief description of the package?

cmd/identity/main.go
bryanchriswhite: I hadn't considered that.

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+	TLS     tlsopts.Config
+}
+
+// Client implements pb.CertificateClient
+type Client struct {
+	conn   *grpc.ClientConn
+	client pb.CertificatesClient
+}
+
+// NewClient creates a new certificate signing grpc client
```
ifraixedes:
```suggestion
// NewClient creates a new certificate signing gRPC client.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+func NewClient(ctx context.Context, tc transport.Client, address string) (_ *Client, err error) {
+	defer mon.Task()(&ctx)(&err)
```
ifraixedes:
```suggestion
	defer mon.Task()(&ctx, address)(&err)
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return &Client{
+		conn:   conn,
+		client: pb.NewCertificatesClient(conn),
+	}, nil
+}
+
+// NewClientFrom creates a new certificate signing grpc client from an existing
+// grpc cert signing client
```
ifraixedes:
```suggestion
// gRPC cert signing client.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+func NewClientFrom(client pb.CertificatesClient) (*Client, error) {
+	return &Client{
+		client: client,
+	}, nil
+}
+
+// Sign submits a certificate signing request given the config
```
ifraixedes:
```suggestion
// Sign submits a certificate signing request given the config.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+	client, err := NewClient(ctx, transport.NewClient(tlsOpts), c.Address)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.Sign(ctx, authToken)
+}
+
+// Sign claims an authorization using the token string and returns a signed
+// copy of the client's CA certificate
```
ifraixedes:
```suggestion
// copy of the client's CA certificate.
```

pkg/certificates/certificatesclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+		Timestamp: time.Now().Unix(),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return res.Chain, nil
+}
+
+// Close closes the client
```
ifraixedes:
```suggestion
// Close closes the client.
```

pkg/certificates/peer.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var (
+	mon = monkit.Package()
+
+	// Error is the default error class for the certificates peer.
+	Error = errs.Class("certificates peer error")
+)
+
+// Config is the global certificates config
```
ifraixedes:
```suggestion
// Config is the global certificates config.
```

pkg/certificates/peer.go
```diff
@@ -0,0 +1,104 @@
...
+	Identity identity.Config
+	Server   server.Config
+
+	Signer         identity.FullCAConfig
+	Authorizations authorizations.Config
+
+	MinDifficulty uint `default:"30" help:"minimum difficulty of the requester's identity required to claim an authorization"`
+}
+
+// Peer is the certificates server
```
ifraixedes:
```suggestion
// Peer is the certificates server.
```

pkg/certificates/service.go
```diff
@@ -0,0 +1,86 @@
...
+	log             *zap.Logger
+	ca              *identity.FullCertificateAuthority
+	authorizationDB *authorizations.DB
+	minDifficulty   uint16
+}
+
+// NewCertificatesServer creates a new certificate signing grpc server
```
ifraixedes:
```suggestion
// NewCertificatesServer creates a new certificate signing gRPC server.
```

pkg/certificates/endpoint.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Endpoint implements pb.CertificatesServer
+type Endpoint struct {
+	log             *zap.Logger
+	ca              *identity.FullCertificateAuthority
+	authorizationDB *authorizations.DB
+	minDifficulty   uint16
+}
+
+// NewCertificatesServer creates a new certificate signing grpc server
```
ifraixedes:
```suggestion
// NewCertificatesServer creates a new certificate signing gRPC server.
```

pkg/certificates/endpoint.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc/peer"
+
+	"storj.io/storj/pkg/certificates/authorizations"
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+)
+
+// Endpoint implements pb.CertificatesServer
```
ifraixedes:
```suggestion
// Endpoint implements pb.CertificatesServer.
```

pkg/certificates/endpoint.go
```diff
@@ -0,0 +1,86 @@
...
+type Endpoint struct {
+	log             *zap.Logger
+	ca              *identity.FullCertificateAuthority
+	authorizationDB *authorizations.DB
+	minDifficulty   uint16
+}
+
+// NewCertificatesServer creates a new certificate signing grpc server
+func NewCertificatesServer(log *zap.Logger, ident *identity.FullIdentity, ca *identity.FullCertificateAuthority, authorizationDB *authorizations.DB, minDifficulty uint16) *Endpoint {
```
mniewrzal: This should be just `NewEndpoint`

pkg/certificates/endpoint.go
```diff
@@ -0,0 +1,86 @@
...
+		log:             log,
+		ca:              ca,
+		authorizationDB: authorizationDB,
+		minDifficulty:   minDifficulty,
+	}
+}
+
+// Sign signs the CA certificate of the remote peer's identity with the `certs.ca` certificate.
+// Returns a certificate chain consisting of the remote peer's CA followed by the CA chain.
+func (endpoint Endpoint) Sign(ctx context.Context, req *pb.SigningRequest) (_ *pb.SigningResponse, err error) {
```
ifraixedes: The error returned by this function must be gRPC error  https://godoc.org/google.golang.org/grpc#Errorf

Codes: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

pkg/certificates/endpoint.go
bryanchriswhite: indeed.. i'm taking another look myself now that I understand the architecture doc a little better.

pkg/certificates/endpoint.go
ifraixedes: I'm realizing that we have to update the documentation https://github.com/storj/docs/blob/master/code/Endpoint.md#grpc-implementation

pkg/certificates/authorizations/authorizations.go
bryanchriswhite: Nice, I hadn't seen that yet. :+1:

pkg/certificates/authorizations/db.go
mniewrzal: oh you might be right

pkg/certificates/authorizations/authorizations.go
bryanchriswhite: What about `certificates`? `certificate` just seems weird imo. Mabye we need a better package name in general.

ifraixedes: but then it will be `pkg/certificates/certificates`?

pkg/certificates/authorizations/authorizations.go
```diff
@@ -0,0 +1,183 @@
...
+	"google.golang.org/grpc/peer"
+	"gopkg.in/spacemonkeygo/monkit.v2"
+
+	"storj.io/storj/pkg/identity"
+	"storj.io/storj/pkg/pb"
+)
+
+const (
+	// AuthorizationsBucket is the bucket used with a bolt-backed authorizations DB.
+	AuthorizationsBucket = "authorizations"
```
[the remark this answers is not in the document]
bryanchriswhite: :+1:

pkg/certificates/certificatesclient/client.go
bryanchriswhite: I'm not sure.

pkg/certificates/authorizations/authorizations.go
bryanchriswhite: Sorry, I wasn't clear. I was asking about the existing `certificates` package, it is also a plural.

ifraixedes: ah sorry, I thought that you were talking about this package.

I use the singular, "seems weird" isn't a good reason for making it plural.

I've found sometimes some packages to sound better in the plural, however, I've chosen to be stuck on the convention.

Reasons, why a package name is in the plural, is for example that it clashes with a built-in type name, for example [`bytes`](https://golang.org/pkg/bytes/) and [`errors`](https://golang.org/pkg/errors/).

pkg/certificates/authorizations/db.go
bryanchriswhite: I think we may not want to call `mon.Task` on these unexported methods. What do you think?

pkg/certificates/authorizations/db.go
bryanchriswhite: (same)

pkg/certificates/authorizations/db.go
ifraixedes: Not sure, but I don't think that hurt and the last conversation that I had with @jtolds regarding `mon.Task` calls,  I understood that's good to call almost everywhere.

bryanchriswhite: ok

pkg/certificate/authorization/authorizations.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	t := &Token{
+		UserID: userID,
+	}
+	copy(t.Data[:], data)
+	return t, nil
+}
+
+// Unmarshal deserializes a set of authorizations.
+func (a *Group) Unmarshal(data []byte) error {
```
mniewrzal: `a` -> `g` or `group`

pkg/certificate/authorization/db.go
```diff
@@ -0,0 +1,276 @@
...
+	for i := 0; i < count; i++ {
+		auth, err := NewAuthorization(userID)
+		if err != nil {
+			authErrs.Add(err)
+			continue
+		}
+		newAuths = append(newAuths, auth)
+	}
+	if authErrs.Err() != nil {
+		return nil, ErrAuthorizationDB.Wrap(err)
```
mniewrzal: `ErrAuthorizationDB.Wrap(authErrs.Err() )`

bryanchriswhite: :man_facepalming:

pkg/certificate/authorization/authorizations.go
bryanchriswhite: :+1:

pkg/certificate/doc.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+/*
+Package certificate is responsible for managing certificate signing operations
+on peer identities' certificate chains.
```
mniewrzal: only one `'`

bryanchriswhite: It's a possessive plural apostrophe

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,92 @@
...
+		ChainBytes:    signedChainBytes,
+		MinDifficulty: endpoint.minDifficulty,
+	})
+	if err != nil {
+		return nil, internalErr(err)
+	}
+
+	difficulty, err := peerIdent.ID.Difficulty()
+	if err != nil {
+		endpoint.log.Error("error checking difficulty", zap.Error(err))
```
mniewrzal: why there is no also return here?

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	token, err := authorization.ParseToken(req.AuthToken)
+	if err != nil {
+		endpoint.log.Error("error parsing auth token", zap.Error(err))
```
mniewrzal: why there is no also return here?

pkg/certificate/doc.go
mniewrzal: oh, good to know :)

pkg/certificate/certificateclient/client.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return &Client{
+		conn:   conn,
+		client: pb.NewCertificatesClient(conn),
+	}, nil
+}
+
+// NewClientFrom creates a new certificate signing gRPC client from an existing
+// grpc cert signing client.
+func NewClientFrom(client pb.CertificatesClient) (*Client, error) {
```
mniewrzal: `error` is not used at all

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,92 @@
...
+		zap.Stringer("truncated token", tokenFormatter),
+	)
+
+	return &pb.SigningResponse{
+		Chain: signedChainBytes,
+	}, nil
+}
+
+func internalErr(err error) error {
+	return status.Error(codes.Internal, Error.Wrap(err).Error())
```
ifraixedes: The problem with this is that sometimes we leak internal information to the client.

One example is this problem that we've had with Posgres and the deadlock due to some constraint information; while that information is good to retain in our logs isn't good to leak it to the client (in that case the Storage Node).

The quick solution that I see is to pass the error log message and the client message to the function and in the function log the error and only return  the code and the message something like 

`
func (endpoint *Endpoint) logAndSendinternalErr(logMsg string, err error, clientMsg String) error {
    endpoint.log.Error(logMsg, zap.Error(err))
	return status.Error(codes.Internal, clientMsg)
`

Or get rid of this function and do that logic each time that's needed.

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,92 @@
...
+func (endpoint Endpoint) Sign(ctx context.Context, req *pb.SigningRequest) (_ *pb.SigningResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+	grpcPeer, ok := peer.FromContext(ctx)
+	if !ok {
+		return nil, internalErr(Error.New("unable to get peer from context"))
+	}
+
+	peerIdent, err := identity.PeerIdentityFromPeer(grpcPeer)
+	if err != nil {
+		return nil, internalErr(err)
```
ifraixedes: We have to log the error otherwise we lose information on what's going on.
See my comment in the `internalErr` function.

This comment applies to every `return nil, internalErr(err)` in this function.

pkg/certificate/peer.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var (
+	mon = monkit.Package()
+
+	// Error is the default error class for the certificates peer.
+	Error = errs.Class("certificates peer error")
+)
+
+// Config is the global certificates config
```
ifraixedes:
```suggestion
// Config is the global certificates config.
```

pkg/certificate/endpoint.go
bryanchriswhite: updates, how's that?

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,104 @@
...
+		zap.Stringer("truncated token", tokenFormatter),
+	)
+
+	return &pb.SigningResponse{
+		Chain: signedChainBytes,
+	}, nil
+}
+
+func internalErr(msg string) error {
+	return status.Error(codes.Internal, Error.New(msg).Error())
```
ifraixedes:
```suggestion
	return status.Error(codes.Internal, msg)
```

bryanchriswhite: `Error.new` will effectively wrap the error in the package error class, wouldn't we want this?

ifraixedes: Rith, my bad, I misread `Error.New` I bore in mind `errors.New`

pkg/certificate/endpoint.go
```diff
@@ -0,0 +1,104 @@
...
+	token, err := authorization.ParseToken(req.AuthToken)
+	if err != nil {
+		msg := "error parsing auth token"
+		endpoint.log.Error(msg, zap.Error(err))
+		return nil, internalErr(msg)
+	}
+	tokenFormatter := authorization.Authorization{
+		Token: *token,
+	}
+	endpoint.log.Info("certificate successfully signed",
```
ifraixedes:
```suggestion
	endpoint.log.Warn("certificate successfully signed",
```

bryanchriswhite: why warn instead of info?

ifraixedes: Because chatty logs make harder to find the important information (relevant errors, etc.).

In the end, it doesn't provide any information unless that you are specifically observing the service running, which usually is when you are trying to find some explanation to some bug/error or you're releasing a new release with important changes and you want to keep the service on eye for a while. In those situations, you change the configuration to run the service in `WARN` level, in normal conditions you don't.

ifraixedes: It's most seen from this point of view

* https://peter.bourgon.org/blog/2016/02/07/logging-v-instrumentation.html
* https://peter.bourgon.org/blog/2017/02/21/metrics-tracing-and-logging.html

PR #2941, opened 2019-09-03
pkg/contact: bare-bones service and endpoint

thepaul:
What: 

Creates a bare-bones service and endpoint, not actually used anywhere yet.

Why:

To start getting rid of kademlia. We're going to leave kademlia functionality in place until we can replace it entirely.

Please describe the tests:
 - No tests to test the behavior yet, as it's pretty much no-op at this point. As functionality is added, tests will be added too.
 
None.


pkg/contact/service.go
```diff
@@ -0,0 +1,71 @@
...
+	"gopkg.in/spacemonkeygo/monkit.v2"
+
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/transport"
+	"storj.io/storj/satellite/overlay"
+)
+
+type Config struct {
+	ExternalAddress string `user:"true" help:"the public address of the node, useful for nodes behind NAT" default:""`
+	DialerLimit     int    `help:"Semaphore size" Default:"32"`
```
mniewrzal: nit: `help` always starts from lower case in our code

pkg/contact/service.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/zeebo/errs"
+	"go.uber.org/zap"
+	"gopkg.in/spacemonkeygo/monkit.v2"
+
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/pkg/transport"
+	"storj.io/storj/satellite/overlay"
+)
+
+type Config struct {
```
JessicaGreben: lint error

pkg/contact/endpoint.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package contact
```
egonelbre: There probably should be two packages `satellite/contact` and `storagenode/contact` since their behavior and dependencies are different.

pkg/contact/endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+	if endpoint.service.Local().Type == pb.NodeType_STORAGE {
+		p, ok := peer.FromContext(ctx)
+		if !ok {
+			return nil, status.Error(codes.Internal, "unable to get grpc peer from context")
+		}
+		peerID, err := identity.PeerIdentityFromPeer(p)
+		if err != nil {
+			return nil, status.Error(codes.Unauthenticated, err.Error())
+		}
+		log.Debug("pinged", zap.Stringer("by", peerID.ID), zap.Stringer("srcAddr", p.Addr))
```
egonelbre: `endpoint.log.Debug`

pkg/contact/endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+// NewEndpoint returns a new contact service endpoint
+func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
+	return &Endpoint{
+		log:     log,
+		service: service,
+	}
+}
+
+// Ping provides an easy way to verify a node is online and accepting requests
+func (endpoint *Endpoint) Ping(ctx context.Context, req *pb.PingRequest) (_ *pb.PingResponse, err error) {
```
cam-a: So if this were the satellite's endpoint, inside this function there would be a client that sends a `Ping` to the SN's endpoint right?

thepaul: I clarified things somewhat when I separated the contact package into storagenode and satellite. Names of things are still up in the air- I just used placeholders for now.

storagenode/contact/service.go
```diff
@@ -0,0 +1,60 @@
...
+// Service is the contact service between storage nodes and satellites
+type Service struct {
+	log       *zap.Logger
+	self      *overlay.NodeDossier
+	transport transport.Client
+
+	mu               sync.Mutex
+	lastPinged       time.Time
+	whoPingedNodeID  storj.NodeID
+	whoPingedAddress string
```
egonelbre: I wonder whether we should keep this information per source?

That way dashboard could show status per satellite.

thepaul: That would be good, but I think it would be far lower in priority than finishing kademlia removal.

PR #2943, opened 2019-09-03
uplink: don't stop deleting segments on first error

mniewrzal:
What: Don't stop deleting segments on first error. With current code if we are not able to delete `s0` (e.g. when nodes are not available) then we are not continuing work on next segments and just stop.

Why: Potential fast fix for https://storjlabs.atlassian.net/browse/V3-2550

Please describe the tests:
 - Test 1: automatic test added
 

uplink/storage/streams/store.go
```diff
@@ -437,19 +437,24 @@ func (s *streamStore) Delete(ctx context.Context, path Path, pathCipher storj.Ci
 		return err
 	}
 
+	var errlist errs.Group
 	for i := 0; i < int(numberOfSegments(&stream, &streamMeta)-1); i++ {
 		currentPath, err := createSegmentPath(ctx, int64(i), path.Bucket(), encPath)
 		if err != nil {
-			return err
+			errlist.Add(err)
```
ifraixedes: I took a look at `createSegmentPath` and I'm wondering if these errors should be skipped.

uplink/storage/streams/store.go
```diff
@@ -437,19 +437,24 @@ func (s *streamStore) Delete(ctx context.Context, path Path, pathCipher storj.Ci
...
+			continue
 		}
 
 		err = s.segments.Delete(ctx, currentPath)
 		if err != nil {
-			return err
+			errlist.Add(err)
```
ifraixedes: We are not distinguishing the errors, so the error can be returned in the metainfo or the node but I guess that's worth to carry on.

uplink/storage/streams/store.go
```diff
@@ -437,19 +437,24 @@ func (s *streamStore) Delete(ctx context.Context, path Path, pathCipher storj.Ci
...
+			continue
 		}
 	}
 
-	return s.segments.Delete(ctx, lastSegmentPath)
+	errlist.Add(s.segments.Delete(ctx, lastSegmentPath))
+
+	return errlist.Err()
```
ifraixedes: We'll return errors to the client, however, the segments may be deleted despite the errors.
I'm wondering how we can inform of that to the callers.

uplink/storage/streams/store.go
mniewrzal: this logic will be changed by #2640 so I don't think we need to care too much about it

uplink/storage/streams/store.go
mniewrzal: we plan to reorganize deleting process so I believe then we can address this

uplink/storage/streams/store.go
mniewrzal: yeah, this is problematic. maybe we can make different classes for errors and distinguish metainfo errors from SN errors but as I mentioned earlier we should improve the whole process in general.

PR #2944, opened 2019-09-03
jenkins: work around git.apache.org failures

thepaul:
What: 

Adds a `replace` directive to `go.mod` which causes Go to use the Apache Thrift module from github instead of the one at git.apache.org (which is no longer reachable).

Why:

I got really tired of the errors with Go being unable to fetch git.apache.org/thrift.git. Apparently, Thrift development officially moved to Github quite some time ago anyway. See:

    https://github.com/apache/thrift/commit/9ee29516c419b7eaa95ed89a93b135ea1c683576

So this `replace` directive causes opencensus.io/foo to use the correct url for the thrift module. Until opencensus.io fixes its dependencies:

    https://github.com/census-instrumentation/opencensus-go/issues/1158

this might be the only recourse.

Please describe the tests:
 - If Jenkins can build at all, this functionality is working.
 
### DIFF
--- a/Jenkinsfile.public
+++ b/Jenkinsfile.public
@@ -102,7 +102,7 @@ pipeline {
 
                     steps {
                         sh 'psql -U postgres -c \'create database teststorj3;\''
-                        sh '# make test-sim-backwards-compatible'
+                        sh 'make test-sim-backwards-compatible'
                     }
                 }
             }
--- a/Jenkinsfile.public
+++ b/Jenkinsfile.public
@@ -102,7 +102,7 @@ pipeline {
 
                     steps {
                         sh 'psql -U postgres -c \'create database teststorj3;\''
-                        sh 'make test-sim-backwards-compatible'
+                        sh '# make test-sim-backwards-compatible'
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


thepaul:
I tried to keep the same hash-version in place, but Go insisted on ignoring the directive entirely until I put an actual version there. I don't know why.

At least I didn't make up `v0.12.0` myself; it's what other Go users are recommending to each other in a few different places in order to fix this unresolvable go.opencensus.io dependency. See for example https://github.com/willnorris/imageproxy/issues/190 .

So yeah maybe it's not perfect, but since go.opencensus.io only comes in to the picture because of golang-migrate, and we only use golang-migrate in a very limited way, I feel very comfortable in saying we don't actually use any thrift code anywhere in our product, so... yeah. Whatever.

Go 1.13 will be nice

thepaul:
Wait what did you just say Go 1.13 is released now??

zeebo:
Indeed: https://blog.golang.org/go1.13

littleskunk:
Awesome. After updating my storage node to go1.13 it is finally running again. Yea!

Jenkinsfile.public
```diff
@@ -102,7 +102,7 @@ pipeline {
 
                     steps {
                         sh 'psql -U postgres -c \'create database teststorj3;\''
-                        sh 'make test-sim-backwards-compatible'
+                        sh '# make test-sim-backwards-compatible'
```
egonelbre: When commenting out tests, lets keep a comment, why it was commented out.

egonelbre: But, I modified the test script to do the same replacement.

PR #2948, opened 2019-09-04
pk/pb: add initial proto for the new contact endpoints

JessicaGreben:
What/Why:

As part of the kademlia removal, we need a way for the storagenodes to still check in with the satellites. This PR adds the grpc endpoints for that, but doesn't use them yet.

Please describe the tests:
 n/a
 
n/a


pkg/pb/contact.proto
```diff
@@ -0,0 +1,29 @@
...
+syntax = "proto3";
+option go_package = "pb";
+
+package contact;
+
+import "gogo.proto";
+
+service Contact {
+    rpc Ping(PingRequest) returns (PingResponse);
+    rpc PingBack(PingBackRequest) returns (PingBackResponse);
```
egonelbre: Shouldn't these RPC-s be on different servers. Storage Node would only need `Pingback`, and Satellite would only need `Ping`.

nit: I think pingback a single word, so `Pingback`

JessicaGreben: yeah good point. let me fix.

JessicaGreben: @egonelbre I added that, but what do you think about the names of the 2 services?

pkg/pb/contact.proto
```diff
@@ -0,0 +1,32 @@
...
+// See LICENSE for copying information.
+
+syntax = "proto3";
+option go_package = "pb";
+
+package contact;
+
+import "gogo.proto";
+
+service ContactNode {
```
egonelbre: The package is already called `contact` so the full name would end up being `contact.ContactNode`.

pkg/pb/contact.proto
egonelbre: The package is already called `contact` so the full name would end up being `contact.ContactNode`.

JessicaGreben: do you have any recommendations for names here?

egonelbre: At the moment no good ones.

JessicaGreben: how about ?
```
service TargetNode {
    rpc Pingback(PingbackRequest) returns (PingbackResponse);
}

service SenderNode {
    rpc Ping(PingRequest) returns (PingResponse);
}
```

egonelbre: `Pinger`, `Pingee` :D... might make sense to do the naming exercise.

egonelbre: Maybe names `Satellite` and `StorageNode` would make the most sense... however then I would like to avoid them attaching to the protobuf package.

pkg/pb/contact.proto
```diff
@@ -0,0 +1,32 @@
...
+    rpc Pingback(PingbackRequest) returns (PingbackResponse);
+}
+
+service ContactSatellite {
+    rpc Ping(PingRequest) returns (PingResponse);
+}
+
+message PingRequest {
+    node.Node sender = 1;
```
cam-a: I think there might be some unnecessary fields in `Node` and maybe we should pull them out. Just a thought, it's your call

JessicaGreben: true. we only need the nodeAddress and nodeID right?

cam-a: I think so

pkg/pb/contact.proto
```diff
@@ -0,0 +1,36 @@
...
+    rpc PingNode(ContactPingRequest) returns (ContactPingResponse);
+}
+
+service Node {
+    rpc Checkin(CheckinRequest) returns (CheckinResponse);
+}
+
+message CheckinRequest {
+    node.NodeAddress sender = 1;
+    bytes sender_nodeid = 2 [(gogoproto.customtype) = "NodeID", (gogoproto.nullable) = false];
```
cam-a: nit - I think field names `address` and `id` would be sufficient. Feel free to ignore

pkg/pb/contact.proto
```diff
@@ -0,0 +1,36 @@
...
+    node.NodeCapacity capacity = 3;
+    node.NodeOperator operator = 4;
+}
+
+message CheckinResponse {
+    bool ping_node_success = 1;
+    string error_message = 2;
```
cam-a: Do we need this? If there's an error, it will be returned separately from the response

pkg/pb/contact.proto
egonelbre: sender_nodeid shouldn't be here, it should be captured from the handshake instead.

pkg/pb/contact.proto
JessicaGreben: the error_message is details about ping_node_success if it is false

JessicaGreben: i think i will change it to `ping_error_message`

PR #2951, opened 2019-09-04
satellite,storagenode,bootstrap: add contact service to peer

thepaul:
What: 

Adds contact services and endpoints as necessary to satellite and storagenode.

Why:

We will need these once we remove kademlia.

Please describe the tests:
 - There is not really any new behavior to test right now; if anything works at all once we start using these facilities, then this change works.
 
None

### DIFF
--- a/bootstrap/peer.go
+++ b/bootstrap/peer.go
@@ -23,7 +23,6 @@ import (
 	"storj.io/storj/pkg/server"
 	"storj.io/storj/pkg/storj"
 	"storj.io/storj/pkg/transport"
-	"storj.io/storj/satellite/contact"
 	"storj.io/storj/satellite/overlay"
 	"storj.io/storj/storage"
 )
@@ -75,11 +74,6 @@ type Peer struct {
 		Service      *kademlia.Kademlia
 		Endpoint     *kademlia.Endpoint
 		Inspector    *kademlia.Inspector
-	}
-
-	Contact struct {
-		Service  *contact.Service
-		Endpoint *contact.Endpoint
 	}
 
 	// Web server with web UI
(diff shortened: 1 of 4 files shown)


satellite/peer.go
```diff
@@ -352,6 +357,11 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, revocationDB exten
 		pb.RegisterKadInspectorServer(peer.Server.PrivateGRPC(), peer.Kademlia.Inspector)
 	}
 
+	{ // setup contact service
+		log.Debug("Setting up contact service")
+		peer.Contact.Service = contact.NewService(peer.Log.Named("contact"), peer.Overlay.Service, peer.Transport)
```
JessicaGreben: do we need an endpoint for the SA peer as well that has the contact.PingNode() rpc?

thepaul: I don't think so. I'm worried if we did that, that it would be confusing for PingNode to be serviced by contact.Service on the nodes and contact.Service on the satellites, but where those two contact.Service's are totally different things :/

JessicaGreben: lol. im confused. ok well lets merge this and sort it out in my next ticket.

bootstrap/peer.go
```diff
@@ -76,6 +77,11 @@ type Peer struct {
 		Inspector    *kademlia.Inspector
 	}
 
+	Contact struct {
```
JessicaGreben: do we need a comment here?

storagenode/contact/service.go
```diff
@@ -20,7 +20,7 @@ var mon = monkit.Package()
 // Service is the contact service between storage nodes and satellites
 type Service struct {
 	log       *zap.Logger
-	self      *overlay.NodeDossier
+	self      overlay.NodeDossier
```
JessicaGreben: just out of curiosity why make this a value instead of a pointer?

satellite/peer.go
thepaul: Heh. I mean, cause, storagenode/contact.Service (or Endpoint) and satellite/contact.Service (or Endpoint) are totally different things

bootstrap/peer.go
thepaul: Not according to the linter, and not according to current practice (see the other nearby structs). But we could sure add one if it is needed for readability

JessicaGreben: no need to add one, i just thought the linter would complain

storagenode/contact/service.go
thepaul: Solely for the sake of being able to pass the result of `peer.Kademlia.RoutingTable.Local()` into the `NewService` constructor. But also, it looks like that's how we generally pass these around. We don't _need_ it to be a pointer (it doesn't need to be changed from elsewhere), and the overhead of copying the extra bytes for copy-by-value is probably outweighed by the overhead of gc stuff when we use a pointer.

jenlij: I noticed that too... I bet it's so this service doesn't accidentally overwrite any of the nodeDossier data

PR #2952, opened 2019-09-04
implement contact.checkin method

JessicaGreben:
What/why:

This PR implements the `contact.Checkin` method on the Satellite so that we can process uptime pings from the storagenodes and continue to remove kademlia.

When a storage node makes a request to the `contact.Checkin`, the following things occur:
- the satellite pings the storage node back to confirm a successful connect.
- as part of the ping back, the satellite confirms the storage node ID matches the original checkin request.
- the satellite updates its `nodes` table with new info on uptime, operator, and capacity

Please describe the tests:
No tests added for now, since most methods are network requests.  I will add tests once this is hooked up with the storage node contact endpoints and also once we add the updates to the queries (see performance impact below).

About every hour each storage node will checkin with the satellite to inform the satellite that its still up.  The satellite updates the `nodes` table with the new uptime info along with any updates to operator info and capacity info.  In the PR, this is implemented in 2 separate queries to the database, which is less than ideal and not performant enough.  The plan is to merge this less performant PR, then start right away on another PR that will either combine into an `upsert` method or possibly a stored procedure. https://github.com/storj/storj/pull/2952/files#diff-e7563cda018a8319698a88ccb8cd19b0R48


satellite/contact/endpoint.go
```diff
@@ -28,10 +32,67 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+	}, nil
+}
+
+func pingBack(ctx context.Context, endpoint *Endpoint, req *pb.CheckinRequest) (bool, string, error) {
+	client, err := newClient(ctx,
+		endpoint.log,
+		endpoint.service.transport,
+		req.GetAddress(),
+	)
+	defer func() {
```
thepaul: I believe this `defer` should come after the following err check. If `err != nil` and we return on line 70, we don't need to close anything because we shouldn't have an open stream.

satellite/peer.go
```diff
@@ -360,6 +361,7 @@ func New(log *zap.Logger, full *identity.FullIdentity, db DB, revocationDB exten
 	{ // setup contact service
 		log.Debug("Setting up contact service")
 		peer.Contact.Service = contact.NewService(peer.Log.Named("contact"), peer.Overlay.Service, peer.Transport)
+		peer.Contact.Endpoint = contact.NewEndpoint(peer.Log.Named("contact"), peer.Contact.Service)
```
thepaul: in other places, we differentiate the logs for the service and the endpoint, to be able to tell which one logged a thing. like `s/contact/contact:endpoint/` in this case.

satellite/contact/service.go
```diff
@@ -4,27 +4,76 @@
...
+	c.updateRequestsCache = append(c.updateRequestsCache, updateRequest)
+	c.mu.Unlock()
+	return c.SaveUpdatesToDB(ctx)
+}
+
+// SaveUpdatesToDB saves the values in the uptime update cache to the overlay
+// database once the batch size has been reached
+func (c *Service) SaveUpdatesToDB(ctx context.Context) error {
+	c.mu.Lock()
+	if len(c.updateRequestsCache) < c.batchSize {
```
jenlij: Would it be better to do this check within AddUpdateToCache before calling c.SaveUpdatesToDB?

jenlij:
we don't need batching any longer?

JessicaGreben:
oh no, we def do still need batching.  Im going to update the PR description with the detail, but the long and short of it is this comment https://github.com/storj/storj/pull/2952/files#diff-e7563cda018a8319698a88ccb8cd19b0R48

Turns out the way i implemented batching it was still making many transactions. Paul recommended writing a postgres stored procedure, that that will be the fastest.  I want to get this change in first, then make another PR for the stored procedure since that will be more involved.

satellite/contact/endpoint.go
```diff
@@ -28,10 +34,87 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+		return nil, Error.Wrap(err)
+	}
+
+	return &pb.CheckinResponse{
+		PingNodeSuccess:  pingNodeSuccess,
+		PingErrorMessage: pingErrorMessage,
+	}, nil
+}
+
+func pingBack(ctx context.Context, endpoint *Endpoint, req *pb.CheckinRequest, peerIDFromContext storj.NodeID) (bool, string, error) {
```
cam-a: nit - if we made this a method on endpoint, that would be one less argument to pass in

satellite/contact/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+// newClient dials the target contact endpoint
+func newClient(ctx context.Context, log *zap.Logger, transport transport.Client, target *pb.NodeAddress) (*client, error) {
+	conn, err := transport.DialAddress(ctx, target.String())
+	if err != nil {
+		return nil, Error.Wrap(err)
+	}
+
+	return &client{
+		log:    log,
```
cam-a: does this logger get used?

satellite/contact/endpoint.go
```diff
@@ -37,46 +39,69 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
 	)
-	defer func() {
-		err = errs.Combine(err, client.close())
-	}()
 	if err != nil {
-		return false, "", err
+		// if this is a network error, then return the error otherwise just report internal error
+		_, ok := err.(net.Error)
+		if ok {
+			return false, "", Error.New("couldn't connect to client at addr %s due to network error: %v.", req.GetAddress().String(), err)
```
thepaul: The "client" term might be confusing here, since normally we refer to the destination of a connect attempt as a server. I mean, I get it, the originating node is a client and we're trying to connect back to it, but people might see this error out of context. How about something more concise like `"failed to connect to %s: %v"`?

satellite/contact/endpoint.go
```diff
@@ -28,10 +34,87 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+		return false, "", Error.New("couldn't connect to client at addr: %s due to internal error.", req.GetAddress().String())
+	}
+
+	pingNodeSuccess := true
+	var pingErrorMessage string
 
-	return &pb.CheckinResponse{}, nil
+	p := &peer.Peer{}
+	_, err = client.pingNode(ctx, &pb.ContactPingRequest{}, grpc.Peer(p))
+	if err != nil {
```
cam-a: If there is an error, should we return it here rather than continuing down and checking the identity?

satellite/contact/endpoint.go
```diff
@@ -37,46 +39,69 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+		}
+	}
+
+	// Confirm that the node ID from the initial checkin request
+	// matches that from this pingNode request
+	identityFromPeer, err := identity.PeerIdentityFromPeer(p)
+	if err != nil {
+		return false, "", Error.New("couldn't get identity from peer: %v", err)
+	}
+	if identityFromPeer.ID != peerIDFromContext {
```
thepaul: I don't think this works- this is comparing two slices, and not their contents. Maybe we need a `(*storj.NodeID).Equal()` method that does `bytes.Equal()`. (Or just do `bytes.Equal()` here)

satellite/contact/client.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type client struct {
+	log    *zap.Logger
+	conn   *grpc.ClientConn
+	client pb.ContactClient
+}
+
+// newClient dials the target contact endpoint
+func newClient(ctx context.Context, log *zap.Logger, transport transport.Client, target *pb.NodeAddress) (*client, error) {
```
thepaul: It's occurred to me that we could do the node ID check with less code, using existing functionality. This function could take a new argument `expectedNodeID` and then make a new `dialOption` like

```go
    dialOption, err := transport.tlsOpts.DialOption(expectedNodeID)
```

and then pass that dialOption to `transport.DialAddress()` and that would tell grpc to add a certificate verifier function that checks for the remote cert matching an expected nodeID.

I'm not 100% sure it would work, because `DialAddress()` adds a conflicting `grpc.WithTransportCredentials`, but it adds our additional options to the end of the list and I would guess that later options override earlier ones.

satellite/contact/endpoint.go
```diff
@@ -37,46 +39,69 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
 		endpoint.service.transport,
 		req.GetAddress(),
 	)
-	defer func() {
-		err = errs.Combine(err, client.close())
-	}()
 	if err != nil {
-		return false, "", err
+		// if this is a network error, then return the error otherwise just report internal error
+		_, ok := err.(net.Error)
```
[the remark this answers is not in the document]
JessicaGreben: idk either. I was looking around for the best way, but didn't come up with much. agree this seems off, not sure whats better other than not having the separate check at all

satellite/contact/endpoint.go
```diff
@@ -37,46 +39,69 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+	// the other to update the capacity and operator info. We should combine these into
+	// one to reduce db connections. Consider adding batching and using a stored procedure.
 	_, err = endpoint.service.overlay.UpdateUptime(ctx, peerID, pingNodeSuccess)
 	if err != nil {
 		return nil, Error.Wrap(err)
 	}
 
-	// Should we update the req.NodeCapacity here as well?
+	nodeInfo := pb.InfoResponse{Operator: req.GetOperator(), Capacity: req.GetCapacity()}
+	_, err = endpoint.service.overlay.UpdateNodeInfo(ctx, peerID, &nodeInfo)
```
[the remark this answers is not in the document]
JessicaGreben: sounds good, i will add the version field in my next PR, will add it to jira ticket 2571

satellite/contact/endpoint.go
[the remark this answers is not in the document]
thepaul: Ok, yeah, this is probably as good as it'll get

satellite/contact/endpoint.go
```diff
@@ -28,10 +34,87 @@ func NewEndpoint(log *zap.Logger, service *Service) *Endpoint {
...
+		endpoint.service.transport,
+		req.GetAddress(),
+	)
+	if err != nil {
+		// if this is a network error, then return the error otherwise just report internal error
+		_, ok := err.(net.Error)
+		if ok {
+			return false, "", Error.New("couldn't connect to client at addr %s due to network error: %v.", req.GetAddress().String(), err)
+		}
+		return false, "", Error.New("couldn't connect to client at addr: %s due to internal error.", req.GetAddress().String())
```
thepaul: Need to log the internal error before we throw it away.

satellite/contact/endpoint.go
JessicaGreben: it looks like [storj.NodeID](https://github.com/storj/storj/blob/master/pkg/storj/node.go#L30) is an array and per [go docs](https://golang.org/ref/spec#Comparison_operators) "Array values are comparable if values of the array element type are comparable. Two array values are equal if their corresponding elements are equal."  So I think this comparison works, also a unit test I have just failed here so seems like its working.

satellite/contact/endpoint.go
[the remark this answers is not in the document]
cam-a: Oh, if we're updating the protos, I have a few small changes to request.
Can we change 'Checkin' to 'CheckIn' since it's two words?
`
service Node {
    rpc Checkin(CheckinRequest) returns (CheckinResponse);
}
`
Could we also change the address field to a string in
`
message CheckinRequest {
    node.NodeAddress address = 1;
    node.NodeCapacity capacity = 3;
    node.NodeOperator operator = 4;
}
`
The node.NodeAddress has an extra Transport field that we don't actually need, but I still have to add it in the request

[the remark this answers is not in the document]
JessicaGreben: yeah i will add those changes @cam-a

satellite/contact/endpoint.go
thepaul: Apologies! I did not realize these were arrays, and I did not know that arrays were comparable. Carry on

satellite/contact/client.go
JessicaGreben: nopers, let me remove.

satellite/contact/client.go
JessicaGreben: hm. interesting. let me give this a shot.

JessicaGreben: Alright I added the dialOption, see commit: https://github.com/storj/storj/pull/2952/commits/0ed3cfdea57c5d97c5909cc510872a8684ba6de7

Does it look right?

thepaul: It does! Yes. Looks like you did what I meant, instead of the broken thing that I said :D

JessicaGreben: lol, no no you explained it perfectly

satellite/repair/repair_test.go
```diff
@@ -63,7 +63,7 @@ func TestDataRepair(t *testing.T) {
 		var (
 			testData         = testrand.Bytes(8 * memory.KiB)
 			minThreshold     = 3
-			successThreshold = 7
+			successThreshold = 6
```
jenlij: why was this updated?

thepaul: I don't much like changing this without knowing why it fixes the test. Can we at least run it by @mobyvb - looks like he made some related changes earlier today

satellite/contact/contact_test.go
```diff
@@ -17,19 +22,27 @@ func TestSatelliteContactEndpoint(t *testing.T) {
...
+				Port: 5,
+			},
+			AuthInfo: credentials.TLSInfo{
+				State: tls.ConnectionState{
+					PeerCertificates: []*x509.Certificate{ident.Leaf, ident.CA},
+				},
+			},
+		}
+		peerCtx := peer.NewContext(ctx, &grpcPeer)
+		resp, err := planet.Satellites[0].Contact.Endpoint.Checkin(peerCtx, &pb.CheckinRequest{
```
cam-a: Would be awesome to see that the node info in the overlay is updated when this endpoint is hit

satellite/repair/repair_test.go
mobyvb: It should pass with either 6 or 7 - this change should be inconsequential, though I am still curious why it was made

satellite/contact/contact_test.go
JessicaGreben: for sure, good idea. yeah im going to add more tests next as part of ticket 2571.

satellite/repair/repair_test.go
JessicaGreben: the test was failing, this fixed it. asking around for more info.

JessicaGreben: @mobyvb this test was failing on my branch with the following info: 
```
	Error Trace:	repair_test.go:94
            	            				run.go:70
            	Error:      	Should be false
            	Test:       	TestDataRepair/Sqlite
            	Messages:   	there is not enough available nodes for repairing: need= 8, have= 7
```
changing to 6 fixed it for some reason

mobyvb: I know what it is! We changed the test to set `toKill := numPieces - toDisqualify - int(minReq)` where previously it was `toKill := numPieces - toDisqualify - int(minReq+1)`. This results in the edge case where the segment gets uploaded to 10 pieces, and we want to kill 6 nodes (10 pieces - 1 disqualified - 3 min = 6). Then, since we have 14 nodes total, if we kill 6, we have 7 nodes remaining. 4 of those will be holding pieces, and 1 of those 4 will be disqualified. This means we have only 3 remaining nodes to repair to, which is not sufficient.

Long story short, it's my (and @VinozzZ 's) fault, and the change @JessicaGreben is what we should do. Otherwise we can increase the total nodes to 15 or decrease max threshold to 9. I think I will make a quickfix PR for this since it will affect other people's stuff.

JessicaGreben: found the issue with the test, fixed with this PR: https://github.com/storj/storj/pull/2973

thepaul: So do you still want to make this change, or leave it so you don’t get blamed when it breaks something 2 years from now haha

JessicaGreben: lol! I was going to revert this change but then i would need to get approvals again. 😢 so getting blamed forever seems easier lol

mobyvb: For what it's worth, I think 7 is the better value (since it means there are 2 nodes between repair/success threshold). But the getting approvals again thing is a real problem that I understand.

JessicaGreben: ok ok i will change it back.

JessicaGreben: _done_

PR #2964, opened 2019-09-05
all: add archview annotations

egonelbre:
Make codebase archview compatible. This adds annotations for satellite and storage node.

[ArchView](https://github.com/storj/archview) allows to visualize code architecture. However, it requires annotations to identify, which parts are important.

Makefile adds two targets `make diagrams` and `make diagrams-graphml`.

`make diagrams` requires [GraphViz](https://www.graphviz.org/). As a result it outputs `.svg` files, this is the quick way to get a visualization.

`make diagrams-graphml` requires [yEd](https://yed.yworks.com) or a similar tool. This will create `.graphml` files that can be opened in a graph editor and layout there.

The best configuration for yEd has been to use Hierarchical Layout, orientation left-to-right, edges octilinear, and layouting groups. Also you can manually group in yEd based on class or some other criteria.

Please describe the tests:
 - Test 1:
 - Test 2:
 

egonelbre:
Currently there still might be some things that are mis-classified.

mobyvb:
Does the linter need to be updated around service/worker/chore/etc... comments?

egonelbre:
@mobyvb the linter doesn't seem to be complaining.

mobyvb:
> @mobyvb the linter doesn't seem to be complaining.

@egonelbre I guess I meant does the linter need to be updated so that it fails in the future if someone forgets to add an annotation to something that should have it?

egonelbre:
@mobyvb the issue is that the annotation should only be for relevant architecture types, not all of them. So it would be really hard to create an automatic decision. Otherwise I would have used that algorithm instead of annotations.

satellite/audit/pathcollector.go
```diff
@@ -9,9 +9,14 @@ import (
 
 	"storj.io/storj/pkg/pb"
 	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite/metainfo"
 )
 
+var _ metainfo.Observer = (*PathCollector)(nil)
```
ifraixedes: These things should have a comment because it will be difficult for people to grasp for what they are.

satellite/metainfo/loop.go
```diff
@@ -58,6 +60,8 @@ type LoopConfig struct {
 }
 
 // Loop is a metainfo loop service.
+//
+// architecture: Service
```
ifraixedes: I'm not clear that this is a service, can you explain to me why it's?

storagenode/monitor/monitor.go
```diff
@@ -34,6 +34,8 @@ type Config struct {
 }
 
 // Service which monitors disk usage and updates kademlia network as necessary.
+//
+// architecture: Service
```
ifraixedes: I'm not clear why this is a service, could you explain to me, please?

satellite/audit/pathcollector.go
egonelbre: This is the idiomatic way to explicitly state that `*PathCollector` implements the interface `metainfo.Observer`. I can add a comment, but I'm wondering whether the better approach would be to make sure all people know how to use it instead.

ifraixedes: I wanted to use in the past to guarantee that a type always implements a specific interface when such an interface isn't used for the same package. But with some people who I was working didn't like it, but anyway, that's another story.

For me it's OK, maybe the question is if we should always add them otherwise, when?

I thought that you added because the tool that you have created needs a way to identify which interface implements a specific type, but I may have thought wrong.

satellite/metainfo/loop.go
egonelbre: I'm kind of going back and forth with some of the things whether they are chores or services. And I think we need a clarification what a service/chore is.

Effectively we have three properties:

A. concurrent-safety: e.g. it would be dangerous to run multiple tallys at the same time without coordination
B. interval: e.g. some things require to happen in an interval
C. things that can be depended on: e.g. metainfo.Loop

We have different combinations of these:

1. concurrent-safe, non-interval, can be depended: e.g. overlay.Service
2. concurrent-safe, interval based, can be depended: e.g. metainfo.Loop
3. concurrent-safe, interval based, should not be depended on: e.g. audit.Worker
4. non-concurrent-safe, interval based, should not be depended on: e.g. audit.Chore

Currently "Service" means it can be depended on and it is concurrent-safe.

Interval-based and should not be depended on, are considered Chores. Currently there's a distinction for "Workers" to indicate, that they are concurrent-safe and usually do work based on a queue.

I'm not certain this organization makes sense, or whether everything is correct based on these criteria.

satellite/audit/pathcollector.go
egonelbre: Yes, I added this, because the tool needs some way to detect the depndency.

But, usually, yes, you should add them, unless it causes some sort of dependency cycle. Also for really common interfaces, like `io.Writer`, `io.Closer`, `fmt.Stringer`, it probably doesn't make much sense.

satellite/metainfo/loop.go
ifraixedes: I'm a bit worry that this may be hard to understand, however, I'm fine to go ahead with this and tune the definition through the time.

storagenode/monitor/monitor.go
egonelbre: Yeah, this is one of those, where I'm completely unsure what it should be. It might be that this needs to be split into a service and a chore.

satellite/audit/pathcollector.go
ifraixedes: Can we add a short comment in some of our current documentation about this?

satellite/metainfo/loop.go
mniewrzal: I think I agree with the definition of `Service`. For me its something that I can depend on and get some results on demand (concurrent-safe or not). I'm not sure if `concurrent-safe` property fits `Chores` because my understanding is that we are starting them, they can depend on services and they are doing some repetitive work. We don't care what's happening internally. On the other hand, `concurrent-safe` may indicate if it's possible to start one chore several times independently.

egonelbre: Anyways, I added a discussion point to architecture meeting, so we can flesh it out a bit more.

satellite/audit/pathcollector.go
egonelbre: https://github.com/storj/docs/pull/109

Makefile
```diff
@@ -331,6 +331,16 @@ test-docker-clean: ## Clean up Docker environment used in test-docker target
 
 ##@ Tooling
 
+.PHONY: diagrams
+diagrams:
+	archview -skip-class "Peer,Master Database" -trim-prefix storj.io/storj/satellite/   ./satellite/...   | dot -T svg -o satellite.svg
+	archview -skip-class "Peer,Master Database" -trim-prefix storj.io/storj/storagenode/ ./storagenode/... | dot -T svg -o storage-node.svg
+
+.PHONY: diagrams-graphml
+diagrams-graphml:
```
mobyvb: the `diagrams` command works well for me, but this one seems to stack everything on top of each other when I open the files with a `graphml` editor. The connections on the graph seem to be in place, but it doesn't look nice visually. Is this intended?

egonelbre: Yes, that's to be expected. The exported `graphml` doesn't contain any positioning information, you need to run a layouting algorithm in the editor. In yEd it's `Layout -> Hierarchical`

PR #2965, opened 2019-09-05
satellite/audit: worker now verifies and reverifies

navillasa:
What:
- Add functionality to Worker.work function so that it will be responsible for verifying and reverifying, as with audit service 1.0.
- Removes unneeded code (e.g. audit/cursor.go, audit/cursor_test.go, audit/service.go) from the audit 1.0 service. These will no longer be needed since the audit workers take audit jobs from the queue (instead of the cursor selecting segments from the metainfo db).
- Updates tests (mostly reverify_test.go and verify_test.go) to use the worker's verifier instead of the verifier that was on the audit.Service struct.

Why:
This PR completes audit 2.0. The worker doing the verification step makes the audit 2.0 
https://storjlabs.atlassian.net/browse/V3-2364

Please describe the tests:
 - Previously the verify and reverify tests would use a cursor to select a random segment then stripe to audit. Now the tests have a worker use the audit queue to retrieve paths to audit.
 

satellite/audit/disqualification_test.go
```diff
@@ -65,10 +62,10 @@ func TestDisqualificationTooManyFailedAudits(t *testing.T) {
 		prevReputation := calcReputation(dossier)
 
 		// Report the audit failure until the node gets disqualified due to many
-		// failed audits
+		// failed audits.
 		iterations := 1
 		for ; ; iterations++ {
-			_, err := satellitePeer.Audit.Service.Reporter.RecordAudits(ctx, report)
+			_, err := satellitePeer.Audit.Worker.Reporter.RecordAudits(ctx, report)
```
mobyvb: If we make `Reporter` exported on `peer.Audit`, and private on `Worker`, we can just use `satellitePeer.Audit.Reporter.RecordAudits` here

satellite/audit/disqualification_test.go
```diff
@@ -45,9 +45,6 @@ func TestDisqualificationTooManyFailedAudits(t *testing.T) {
 			},
 		},
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		err := planet.Satellites[0].Audit.Service.Close()
```
mobyvb: There are several places in this file where you removed `satellitePeer.Audit.Service.Close()` and did not replace it with `satellitePeer.Audit.Worker.Close()`.
On that note, `worker.Close()` doesn't currently do anything, but it should do `worker.Loop.Close()`

mobyvb: Closing the audit worker is important still because these tests will be affected by passing/failing audits that might run in the background if the worker isn't stopped.

satellite/audit/reporter_test.go
```diff
@@ -21,8 +21,6 @@ func TestReportPendingAudits(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 0,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		err := planet.Satellites[0].Audit.Service.Close()
```
mobyvb: we should still close the audit worker

satellite/audit/reporter_test.go
```diff
@@ -58,8 +56,6 @@ func TestRecordAuditsAtLeastOnce(t *testing.T) {
...
```
mobyvb: we should still close the audit worker. Also, if we make `Reporter` exported on `peer.Audit`, we do not necessarily need to create a new reporter as part of this test (and other tests in this file).

satellite/peer.go
```diff
@@ -196,10 +191,9 @@ type Peer struct {
 		Inspector *irreparable.Inspector
 	}
 	Audit struct {
-		Service *audit.Service
-		Queue   *audit.Queue
-		Worker  *audit.Worker
-		Chore   *audit.Chore
+		Queue  *audit.Queue
+		Worker *audit.Worker
+		Chore  *audit.Chore
```
mobyvb: adding `Reporter` and `Verifier` here could simplify tests

satellite/audit/disqualification_test.go
navillasa: ok cool!

satellite/peer.go
navillasa: okay! they should still be passed into NewWorker too right?

satellite/audit/disqualification_test.go
navillasa: ah you're right, I definitely forgot to add it back.

satellite/audit/reporter_test.go
navillasa: I think at least for the tests in this file, it is still necessary to make the new reporter because we're instantiating it with different max retry values, which affects the outcome of the test.

satellite/audit/disqualification_test.go
navillasa: hm I'm getting a panic: close on closed channel now that I've added `worker.Loop.Close()` but I'm not totally sure why the channel would be closed already.

satellite/audit/verifier.go
```diff
@@ -535,8 +553,8 @@ OUTER:
 	return err
 }
 
-// checkIfSegmentDeleted checks if stripe's pointer has been deleted since stripe was selected.
-func (verifier *Verifier) checkIfSegmentDeleted(ctx context.Context, segmentPath string, oldPointer *pb.Pointer) (newPointer *pb.Pointer, err error) {
+// checkIfSegmentDeleted checks if path's pointer has been deleted since path was selected.
+func (verifier *Verifier) checkIfSegmentDeleted(ctx context.Context, segmentPath string) (newPointer *pb.Pointer, err error) {
```
egonelbre: We need to verify whether it's the same pointer we started with or not otherwise there could be a situation:

```
1. audit starts with pointer X
2. repair repairs pointer Y, and deletes some pieces from storage nodes A and B
3. audit fails for A and B
```

To ensure that this doesn't happen, we need to verify in step: 3. whether `pointer X` is the same we started with, otherwise we have requested nodes A and B to delete the piece and will fail the audit for them.

satellite/audit/verifier.go
```diff
@@ -698,3 +717,41 @@ func rebuildStripe(ctx context.Context, fec *infectious.FEC, corrected []infecti
...
+		return 0, err
+	}
+
+	// the last segment could be smaller than stripe size
+	if pointer.GetSegmentSize() < int64(redundancy.StripeSize()) {
+		return 0, nil
+	}
+
+	var src cryptoSource
+	rnd := rand.New(src)
```
egonelbre: We technically don't need cryptographic randomness, having a good starting seed with really good crypto generator should be sufficient.

satellite/audit/verifier_test.go
```diff
@@ -466,83 +534,97 @@ func TestVerifierDeletedSegment(t *testing.T) {
...
-	testplanet.Run(t, testplanet.Config{
-		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 1,
-	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		audits := planet.Satellites[0].Audit.Service
-		err := audits.Close()
-		require.NoError(t, err)
-
-		ul := planet.Uplinks[0]
-		testData := testrand.Bytes(8 * memory.KiB)
+// TODO: Commenting this test for now because currently when we get paths from the audit queue,
```
egonelbre: We need this for correctness, so it must not be removed.

navillasa: okay. how do you think we should access the old pointer?

egonelbre: Before starting the audit, you load it from metainfo.

navillasa: I'm not sure I totally know where you mean. This `checkIfSegmentDeleted` function is called early on in the `Verify` function:
```
// checkIfSegmentDeleted checks if path's pointer has been deleted since path was selected.
func (verifier *Verifier) checkIfSegmentDeleted(ctx context.Context, segmentPath string) (newPointer *pb.Pointer, err error) {
	defer mon.Task()(&ctx)(&err)

	newPointer, err = verifier.metainfo.Get(ctx, segmentPath)
	if err != nil {
		if storage.ErrKeyNotFound.Has(err) {
			return nil, ErrSegmentDeleted.New("%q", segmentPath)
		}
		return nil, err
	}

	// TODO: commenting because we no longer have access to the old pointer
	//if oldPointer != nil && oldPointer.CreationDate != newPointer.CreationDate {
	//	return nil, ErrSegmentDeleted.New("%q", segmentPath)
	//}
	return newPointer, nil
}
```
Should a metainfo.Get call happen before this call in `Verify` and that will be the old pointer? And then inside this function we call metainfo.Get again, and this time that will be the `newPointer`?

egonelbre: It could happen at the start of verify or before verify, but we need this logic of:

1. load pointer from metainfo
2. audit nodes
3. check whether pointer in metainfo still matches step 1.
4. handle audit failures / successes

But, yes, `checkIfSegmentDeleted` is probably not the best place to load the old pointer.

navillasa: the way audit service 1.0 worked though, it didn't look like it followed those steps either, since this comparison between old and new pointers happened before auditing the nodes occurred right?

egonelbre: In https://github.com/storj/storj/blob/master/satellite/audit/verifier.go#L109 the check happens after downloading the pieces, only then it goes to check the downloaded pieces. The old pointer was already in the initial call.

navillasa: ah okay I see. do you have any opinion then on where to load the old pointer? should it still happen in the top of the verify call?

satellite/audit/verifier.go
```diff
@@ -74,13 +78,22 @@ func NewVerifier(log *zap.Logger, metainfo *metainfo.Service, transport transpor
...
 }
 
-// Verify downloads shares then verifies the data correctness at the given stripe
-func (verifier *Verifier) Verify(ctx context.Context, stripe *Stripe, skip map[storj.NodeID]bool) (report *Report, err error) {
+// Verify downloads shares then verifies the data correctness at a random stripe.
+func (verifier *Verifier) Verify(ctx context.Context, path storj.Path, skip map[storj.NodeID]bool) (report *Report, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	pointer := stripe.Segment
+	pointer, err := verifier.checkIfSegmentDeleted(ctx, path)
```
egonelbre: Replace this with loading the pointer from metainfo rather than calling `checkIfSegmentDeleted`. And keep the argument `oldPointer` for `checkIfSegmentDeleted`

navillasa: okay sounds good, thanks for the help!

navillasa: okay I did the above. this is looking difficult to test though. previously the test looked like this:
```
		err := ul.Upload(ctx, planet.Satellites[0], "testbucket", "test/path", testData)
		require.NoError(t, err)

		satellite.Audit.Chore.Loop.TriggerWait()
		path, err := queue.Next()
		require.NoError(t, err)

		// replace the file
		err = ul.Upload(ctx, planet.Satellites[0], "testbucket", "test/path", testData)
		require.NoError(t, err)

		report, err := audits.Verifier.Verify(ctx, path, nil)
		require.True(t, audit.ErrSegmentDeleted.Has(err))
		assert.Empty(t, report)
```
however, now in order for `audit.ErrSegmentDeleted.Has(err)` to be true, I think we would have to call Verify and the re-upload concurrently where the re-upload happens after Verify has started but before the `checkIfSegmentDeleted` call happens. do you know a way to test this that's not flaky?

satellite/audit/worker.go
```diff
@@ -78,6 +102,45 @@ func (worker *Worker) process(ctx context.Context) (err error) {
 }
 
 func (worker *Worker) work(ctx context.Context, path storj.Path) error {
-	// TODO: handle work.
-	return nil
+	var errlist errs.Group
+	report, err := worker.verifier.Reverify(ctx, path)
```
mobyvb: nit - add a comment: `First, attempt to reverify nodes for this segment that are in containment mode`

satellite/audit/worker.go
```diff
@@ -78,6 +102,45 @@ func (worker *Worker) process(ctx context.Context) (err error) {
...
+		}
+		for _, nodeID := range report.Fails {
+			skip[nodeID] = true
+		}
+		for _, pending := range report.PendingAudits {
+			skip[pending.NodeID] = true
+		}
+	}
+
+	report, err = worker.verifier.Verify(ctx, path, skip)
```
mobyvb: nit - add another comment `Next, audit the the remaining nodes that are not in containment mode`

satellite/audit/verifier.go
navillasa: I think it might have been @kaloyan-raev who originally added the cryptographic randomness? I'm going to just move it to a separate file for now I think.

navillasa: I'm also not totally sure what "having a good starting seed with a really good crypto generator" means in terms of what we should keep and get rid of.

satellite/audit/reverify_test.go
```diff
@@ -366,49 +401,57 @@ func TestReverifyDeletedSegment(t *testing.T) {
...
 		require.NoError(t, err)
 
 		// delete the file
 		err = ul.Delete(ctx, planet.Satellites[0], "testbucket", "test/path")
 		require.NoError(t, err)
 
-		report, err := audits.Verifier.Reverify(ctx, stripe)
-		require.NoError(t, err)
+		report, err := audits.Verifier.Reverify(ctx, path)
+		// TODO: is this desired behavior for if a segment is deleted?
```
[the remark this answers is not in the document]
mobyvb: Looks like the error is only logged in the worker, so it should be fine to treat this error as expected

satellite/audit/verifier.go
thepaul: "A good starting seed with a really good crypto generator" is basically the definition of cryptographic randomness :)

I believe the randomness in this case does need to be crypto-strength, to make things harder for attackers who want to subvert the auditing. I certainly agree with moving it into a separate file, or maybe even a separate package under `internal/`?

egonelbre: moving it to `internal` might be overkill.

satellite/audit/verifier.go
egonelbre: The best I can think of is adding a callback that can be set for testing, e.g. at the beggining of `checkIfSegmentDeleted` and in the callback then modify the metainfo database.

```
// checkIfSegmentDeleted checks if stripe's pointer has been deleted since stripe was selected.
func (verifier *Verifier) checkIfSegmentDeleted(ctx context.Context, segmentPath string, oldPointer *pb.Pointer) (newPointer *pb.Pointer, err error) {
	defer mon.Task()(&ctx)(&err)

	if verifier.OnTestingCheckSegmentDeletedHook != nil {
		verifier.OnTestingCheckSegmentDeletedHook
```

satellite/audit/verifier.go
```diff
@@ -535,7 +559,7 @@ OUTER:
 	return err
 }
 
-// checkIfSegmentDeleted checks if stripe's pointer has been deleted since stripe was selected.
+// checkIfSegmentDeleted checks if path's pointer has been deleted since path was selected.
```
egonelbre: Note let's rename this to `checkSegmentAltered` rather than deleted, because the important part is whether it was changed in any way, not just deletion.

satellite/audit/verifier.go
navillasa: added!

satellite/audit/cryptosource.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package audit
+
+import (
+	crand "crypto/rand"
```
egonelbre: alias is no longer necessary

satellite/audit/verifier_test.go
```diff
@@ -313,23 +358,29 @@ func TestVerifierHappyPath(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 4, UplinkCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
-		audits := planet.Satellites[0].Audit.Service
-		err := audits.Close()
-		require.NoError(t, err)
+		planet.Satellites[0].Audit.Worker.Loop.Pause()
```
mobyvb: nit - this pattern is followed in several tests I've seen, and it can be reordered:
```
satellite := planet.Satellites[0]
audits := satellite.Audit
queue := audits.Queue

audits.Worker.Loop.Pause()
```

PR #2971, opened 2019-09-06
pkg/certificates: add authorization endpoint and refactor

bryanchriswhite:
What: Add authorization http endpoint for creating authorizations within a trusted network environment.

Why: [V3-2382](https://storjlabs.atlassian.net/browse/V3-2382?atlOrigin=eyJpIjoiNGUwYzUwMTlmN2U3NGUwZGIyMTllYTFhYzliYzg0NzUiLCJwIjoiaiJ9)

Please describe the tests:
`authorizations_test.go`
- Add `TestNewDBFromCfg`: exercises `authorization.NewDBFromCfg`
`endpoint_test.go`
- Add `TestEndpoint_Create`: 
  + Creates a new authorization endpoint (without a listener).
  + Calls the endopint's `Create` method with a valid authorization request.
  + Asserts that it gets a parsable token back with the correct user ID.
- Add `TestEndpoint_Run_httpCreate`:
  + Creates a new authorization endpoint and calls `Run`.
  + Sends an HTTP `POST` request to the `v1/authorization/create` path  with a valid user ID.
  + Asserts:
    * The HTTP response status is 201 cretated
    * The HTTP response body contains a parsable token with the correct user ID.
- Add `TestEndpoint_Run_httpErrors`:
  + Creates a new authorization endpoint and calls `Run`.
  + Sends an HTTP request for each entry in the following table, asserting the response status code matches the test case:
    ```
	testCases := []struct {
		name       string
		userID     string
		urlPath    string
		httpMethod string
		statusCode int
	}{
		{
			"missing user ID",
			"",
			"/v1/authorization",
			http.MethodPut,
			http.StatusUnprocessableEntity,
		},
		{
			"unsupported http method (GET)",
			"user@mail.test",
			"/v1/authorization",
			http.MethodGet,
			http.StatusMethodNotAllowed,
		},
		{
			"unsupported http method (PUT)",
			"user@mail.test",
			"/v1/authorization",
			http.MethodPost,
			http.StatusMethodNotAllowed,
		},
		{
			"not found",
			"",
			"/",
			http.MethodPut,
			http.StatusNotFound,
		},
	}
    ```
 

bryanchriswhite:
I'm not sure what the linter wants.

pkg/certificate/authorization/authorizations_test.go
```diff
@@ -43,24 +43,39 @@ var (
 	}
 )
 
-func TestCertSignerConfig_NewAuthDB(t *testing.T) {
+func TestNewDB(t *testing.T) {
 	ctx := testcontext.New(t)
 	defer ctx.Cleanup()
 
-	authDB, err := newTestAuthDB(ctx)
+	dbURL := "bolt://" + ctx.File("authorizations.db")
```
stefanbenten: Not sure if thats the right way to do that? iirc, there are elsewhere temp file being created.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+	if httpReq.Method != http.MethodPost {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
+		http.Error(writer, msg, http.StatusBadRequest)
+		return
+	}
+
+	fmt.Println("no error!")
+	userID, err := ioutil.ReadAll(httpReq.Body)
+	if err != nil || bytes.Equal([]byte{}, userID) {
```
stefanbenten: NIT: we can compare string(userID) == "", to get rid of the bytes dependency

egonelbre:
@bryanchriswhite when you make a request you need to close the body. See the second example in https://golang.org/pkg/net/http/.

pkg/certificate/peer.go
```diff
@@ -31,8 +32,9 @@ type Config struct {
 	Identity identity.Config
 	Server   server.Config
 
-	Signer         identity.FullCAConfig
-	Authorizations authorization.Config
+	Signer            identity.FullCAConfig
+	AuthorizationDB   authorization.DBConfig
+	AuthorizationAddr string
```
stefanbenten: Shouldnt this be an exported config flag?

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+	mux.HandleFunc("/v1/authorization/create", endpoint.httpCreate)
+
+	return endpoint
+}
+
+// Create creates an authorization from the given authorization request.
+func (endpoint *Endpoint) Create(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
+	mon.Task()(&ctx, req.UserId)(&err)
+
+	group, err := endpoint.db.Create(ctx, req.UserId, 1)
```
stefanbenten: We should add a check to see if there are open auth tokens and return them first, instead of blind generating a new one.
this ensures the users setup a new node (or at least identity) before generating them a new token.

pkg/certificate/authorization/endpoint.go
thepaul: I’d strongly prefer staying with `bytes.Equal` (or just testing `len(userID) == 0` if the concern is the extra stdlib import). Conversion to string involves a new heap allocation and linear copy, and since this looks like it’s a payload from the internet, who knows how big it is?

pkg/certificate/authorization/authorizations_test.go
mniewrzal: look ok to me. `ctx.Dir` and `ctx.File` are in temp location that will be removed with cleanup.

pkg/certificate/authorization/authorizations_test.go
```diff
@@ -622,7 +633,11 @@ func TestNewClient(t *testing.T) {
 	})
 }
 
-func newTestAuthDB(ctx *testcontext.Context) (*DB, error) {
+func newTestAuthDB(t *testing.T, ctx *testcontext.Context) *DB {
 	dbURL := "bolt://" + ctx.File("authorizations.db")
-	return NewDB(dbURL, false)
+	db, err := NewDB(dbURL, false)
+	if err != nil {
```
mniewrzal: to have the same result just use `require.NoError`

pkg/certificate/authorization/authorizations_test.go
bryanchriswhite: I'm not sure what you mean.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+	ctx := context.Background()
+
+	if httpReq.Method != http.MethodPost {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
+		http.Error(writer, msg, http.StatusBadRequest)
+		return
+	}
+
+	fmt.Println("no error!")
```
mniewrzal: remove `Println`

pkg/certificate/peer.go
bryanchriswhite: It it, it just needs default and help tags.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+// Create creates an authorization from the given authorization request.
+func (endpoint *Endpoint) Create(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
+	mon.Task()(&ctx, req.UserId)(&err)
+
+	group, err := endpoint.db.Create(ctx, req.UserId, 1)
+	if err != nil {
+		msg := "error creating authorization"
+		err := ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(err))
+		return nil, status.Error(codes.Internal, err.Error())
```
mniewrzal: why not:
```go
endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
return nil, status.Error(codes.Internal, msg)
```

pkg/certificate/peer.go
```diff
@@ -74,9 +81,18 @@ func New(log *zap.Logger, ident *identity.FullIdentity, ca *identity.FullCertifi
...
+	peer.Certificate.Endpoint = NewEndpoint(log.Named("certificate"), ca, authorizationDB, uint16(config.MinDifficulty))
+	pb.RegisterCertificatesServer(peer.Server.GRPC(), peer.Certificate.Endpoint)
+
+	peer.Authorization.Listener, err = net.Listen("tcp", config.AuthorizationAddr)
+	if err != nil {
+		return nil, errs.Combine(err, peer.Close())
+	}
+
+	peer.Authorization.Endpoint = authorization.NewEndpoint(log.Named("authorization"), authorizationDB, peer.Authorization.Listener)
+	pb.RegisterAuthorizationsServer(peer.Server.PrivateGRPC(), peer.Authorization.Endpoint)
```
mniewrzal: can you add a comment why in this case `PrivateGRPC` is used?

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+	return group.Wait()
+}
+
+// Close closes the endpoint HTTP server.
+func (endpoint *Endpoint) Close() error {
+	return endpoint.server.Close()
+}
+
+func (endpoint *Endpoint) httpCreate(writer http.ResponseWriter, httpReq *http.Request) {
+	ctx := context.Background()
```
mniewrzal: I guess it would be good to add monkit here

pkg/certificate/authorization/endpoint.go
bryanchriswhite: We don't get the `ErrEndpoint` wrapping in the grpc error if we do that.

pkg/certificate/authorization/authorizations_test.go
bryanchriswhite: It seems to be a convention that test helper code uses `t.Fatal` inside of a condition rather than using `require`.

egonelbre: `require` would be completely fine. I think some of the testing code was not using `require`, because we didn't know about it. Initially everything was using `assert`, so to stop the test you needed to explicitly call `Fatal`.

bryanchriswhite: :+1:

pkg/certificate/authorization/endpoint.go
mniewrzal: hmm result for endpoint consumer will be `authorization endpoint error: error creating authorization`, ok looks good. One thing that bothers me a little bit is `endpoint` in the default error message. I may give an impression that something goes wrong with endpoint not the service itself but treat that as nit.

bryanchriswhite: What are you calling the "service" here?

mniewrzal: because by definition endpoint shouldn't contain any business logic, its just processing network requests. Here we don't have explicity different service but still, endpoint is just passing a request to `DB` and error is coming from there. 
https://github.com/storj/storj/wiki/Architecture#layers

pkg/certificate/peer.go
```diff
@@ -31,8 +32,9 @@ type Config struct {
 	Identity identity.Config
 	Server   server.Config
 
-	Signer         identity.FullCAConfig
-	Authorizations authorization.Config
+	Signer            identity.FullCAConfig
+	AuthorizationDB   authorization.DBConfig
+	AuthorizationAddr string `default:"127.0.0.1:9000" help:""`
```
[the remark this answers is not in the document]
bryanchriswhite: Yep, will fix.. got distracted by the baby.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Create creates an authorization from the given authorization request.
+func (endpoint *Endpoint) Create(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
+	mon.Task()(&ctx, req.UserId)(&err)
+
+	existingGroup, err := endpoint.db.Get(ctx, req.UserId)
+	if err != nil {
+		msg := "error getting authorizations"
+		err := ErrEndpoint.New(msg)
```
mniewrzal: hiding error

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
+		http.Error(writer, msg, http.StatusBadRequest)
+		return
+	}
+
+	userID, err := ioutil.ReadAll(httpReq.Body)
+	if bytes.Equal([]byte{}, userID) {
+		msg := "missing user ID body"
+		err = ErrEndpoint.Wrap(err)
+		endpoint.log.Error(msg, zap.Error(err))
```
mniewrzal: nit: I don't think this needs to be logged. The message is clear what needs to be added to fix request.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (endpoint *Endpoint) httpCreate(writer http.ResponseWriter, httpReq *http.Request) {
+	var err error
+	ctx := context.Background()
+	mon.Task()(&ctx)(&err)
+
+	if httpReq.Method != http.MethodPost {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
```
mniewrzal: duplicated `ErrEndpoint.New(msg)`

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+		return &pb.AuthorizationResponse{
+			Token: authorization.Token.String(),
+		}, nil
+	}
+
+	createdGroup, err := endpoint.db.Create(ctx, req.UserId, 1)
+	if err != nil {
+		msg := "error creating authorization"
+		err := ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(err))
```
mniewrzal: hiding error

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+	if httpReq.Method != http.MethodPost {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
+		http.Error(writer, msg, http.StatusBadRequest)
+		return
+	}
+
+	userID, err := ioutil.ReadAll(httpReq.Body)
+	if bytes.Equal([]byte{}, userID) {
```
mniewrzal: this needs to be after `if err != nil {` in case with error method will return empty `userID`

pkg/certificate/peer.go
```diff
@@ -74,9 +81,19 @@ func New(log *zap.Logger, ident *identity.FullIdentity, ca *identity.FullCertifi
...
 
-	peer.Certificates.AuthorizationDB = authorizationDB
-	peer.Certificates.Endpoint = NewEndpoint(log.Named("certificates"), ca, authorizationDB, uint16(config.MinDifficulty))
-	pb.RegisterCertificatesServer(peer.Server.GRPC(), peer.Certificates.Endpoint)
+	peer.AuthorizationDB = authorizationDB
+
+	peer.Certificate.Endpoint = NewEndpoint(log.Named("certificate"), ca, authorizationDB, uint16(config.MinDifficulty))
+	pb.RegisterCertificatesServer(peer.Server.GRPC(), peer.Certificate.Endpoint)
+
+	peer.Authorization.Listener, err = net.Listen("tcp", config.AuthorizationAddr)
```
mniewrzal: nit: Maybe we need validation (check if `config.AuthorizationAddr` is localhost) to avoid exposing endpoint by accident

bryanchriswhite: It isn't always necessarily the loopback interface though. It could be correct and desirable to bind to an private network interface.

pkg/certificate/authorization/endpoint.go
bryanchriswhite: You're right.

pkg/certificate/authorization/endpoint.go
ifraixedes: We should only log system errors, user input errors shouldn't be logged because they don't provide us any information about our system isn't behaving correctly.

bryanchriswhite: :+1: that makes sense

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Close closes the endpoint HTTP server.
+func (endpoint *Endpoint) Close() error {
+	return endpoint.server.Close()
+}
+
+func (endpoint *Endpoint) httpCreate(writer http.ResponseWriter, httpReq *http.Request) {
+	var err error
+	ctx := context.Background()
```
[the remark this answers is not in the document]
bryanchriswhite: Oh nice, I didn't know about this - thanks!

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,161 @@
...
+	mon.Task()(&ctx)(&err)
+
+	if httpReq.Method != http.MethodPost {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(ErrEndpoint.New(msg)))
+		http.Error(writer, msg, http.StatusBadRequest)
```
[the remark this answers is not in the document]
bryanchriswhite: :+1:

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,161 @@
...
+		return
+	}
+
+	userID, err := ioutil.ReadAll(httpReq.Body)
+	if err != nil || bytes.Equal([]byte{}, userID) {
+		msg := "error reading body"
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(err))
+		http.Error(writer, msg, http.StatusBadRequest)
```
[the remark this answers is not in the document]
bryanchriswhite: :+1:

pkg/certificate/authorization/endpoint_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	endpoint := NewEndpoint(zaptest.NewLogger(t), newTestAuthDB(t, ctx), listener)
+	require.NotNil(t, endpoint)
+
+	ctx.Go(func() error {
+		return errs2.IgnoreCanceled(endpoint.Run(ctx))
+	})
+	defer ctx.Check(endpoint.Close)
+
+	userID := "user@mail.example"
```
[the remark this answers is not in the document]
bryanchriswhite: FWIW they're both reserved TLDs but I take your point that in a test, it makes sense to use the `.test` domain.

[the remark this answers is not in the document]
ifraixedes: yes, both are reserved, but I would change them for you what you said.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+		http.Error(writer, msg, http.StatusInternalServerError)
+		return
+	}
+
+	writer.WriteHeader(http.StatusCreated)
+	if _, err = writer.Write([]byte(authorizationRes.Token)); err != nil {
+		msg := "error writing response"
+		err = ErrEndpoint.Wrap(err)
+		endpoint.log.Error(msg, zap.Error(err))
+		http.Error(writer, msg, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not sure what the expected error code actually should be in this case because yes, there was an error but the resource was created.

pkg/certificate/authorization/endpoint.go
stefanbenten: True, fair point @thepaul

pkg/certificate/authorization/endpoint.go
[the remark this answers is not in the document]
ifraixedes: we cannot do anything in this case.
Just log and that's it.

The client won't get any response and if get's any, it will be the first status code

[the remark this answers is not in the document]
bryanchriswhite: I checked, the response body is the error message (instead of a token) and the status code is 201. I think this still accurately represents the state of the resource after the request and communicates the error message. So while we may not have any other choice at the moment, I think this is okay.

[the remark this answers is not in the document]
ifraixedes: Yes, that's correct.
It still applies that I've said in writing the header after the first write and that causes an error, however it won't be logged as I said because we use this helper function which ignores any error, hence, we don't get any error, we don't log anything.

Out of curiosity, how did you simulate a test which makes the `writer.Write` to return an error?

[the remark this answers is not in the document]
bryanchriswhite: I just changed the condition from `err != nil` to `err == nil` and ran the tests.

bryanchriswhite:
Can we check the checkboxes plz?

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	createdGroup, err := service.db.Create(ctx, req.UserId, 1)
+	if err != nil {
+		msg := "error creating authorization"
+		err = ErrService.Wrap(err)
+		service.log.Error(msg, zap.Error(err))
+
+		switch err {
+		case ErrCount, ErrEmptyUserID:
```
ifraixedes: These errors aren't directly returned they are wrapped.

I think that the solution is `errs.Unwrap(err)` and then compare the error, however, I was going to verify it and I found:

1. There is no test case for when this function returns an error (it's tested through the endpoint)
2. It turns that these 2 errors should never happen because
	* `ErrCount` is set to 1
	* `ErrEmptyUserID` if it's empty it will be already be detected by the `service.db.Get` call executed above.

	Nevertheless, I think that's good anyway to have these checks here.

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,82 @@
...
+// GetOrCreate will return an authorization for the given user ID
+func (service *Service) GetOrCreate(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	existingGroup, err := service.db.Get(ctx, req.UserId)
+	if err != nil {
+		msg := "error getting authorizations"
+		err = ErrService.Wrap(err)
+		service.log.Error(msg, zap.Error(err))
+		return nil, status.Error(codes.Internal, msg)
```
ifraixedes: See my below comment, but here the error could be `status.Error(codes.InvalidArgument, msg)` because an empty `userID`.

ifraixedes: Let's add a test case in _endpoint_test.go#TestEndpoint_Create` for checking the errors that can be returned because of invalid data.

Example:

```go
	 _, err := endpoint.Create(ctx, &pb.AuthorizationRequest{})                                                          
	require.Error(t, err)                                                                                               
	require.Equal(t, codes.InvalidArgument, status.Code(err), "invalid gRPC stats code")
```

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,82 @@
...
+		return &pb.AuthorizationResponse{
+			Token: authorization.Token.String(),
+		}, nil
+	}
+
+	createdGroup, err := service.db.Create(ctx, req.UserId, 1)
+	if err != nil {
+		msg := "error creating authorization"
+		err = ErrService.Wrap(err)
+		service.log.Error(msg, zap.Error(err))
```
ifraixedes: When the error is due to invalid arguments (i.e. `status.Error(codes.InvalidArgument, msg)`) we shouldn't log the error.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+	authorizationReq := &pb.AuthorizationRequest{
+		UserId: string(userID),
+	}
+
+	authorizationRes, err := endpoint.Create(ctx, authorizationReq)
+	if err != nil {
+		msg := "error creating authorization"
+		err = ErrEndpoint.Wrap(err)
+		endpoint.log.Error(msg, zap.Error(err))
+		http.Error(writer, msg, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
ifraixedes: This change was forgotten.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (endpoint *Endpoint) handleAuthorization(writer http.ResponseWriter, httpReq *http.Request) {
+	var err error
+	ctx := httpReq.Context()
+	mon.Task()(&ctx)(&err)
+
+	if httpReq.Method != http.MethodPut {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(err))
```
ifraixedes: This error shouldn't be logged, it's a client fault.

pkg/certificate/authorization/service.go
bryanchriswhite: Are you suggesting testing the same error cases in the endpoint and the service?

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,82 @@
...
+var ErrService = errs.Class("authorization service error")
+
+// Service is the authorization service.
+type Service struct {
+	log *zap.Logger
+	db  *DB
+}
+
+// NewService creates a new authorization service.
+func NewService(log *zap.Logger, db *DB) *Service {
```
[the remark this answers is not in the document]
bryanchriswhite: The endpoint doesn't need to know about that logic but it makes sense that it handles the protobuf request and response types. I'll make this nicer.

pkg/certificate/authorization/service.go
```diff
@@ -32,37 +30,36 @@ func NewService(log *zap.Logger, db *DB) *Service {
...
+// GetOrCreate will return an authorization for the given user ID.
+func (service *Service) GetOrCreate(ctx context.Context, userID string) (_ *Token, err error) {
 	defer mon.Task()(&ctx)(&err)
 
-	existingGroup, err := service.db.Get(ctx, req.UserId)
+	if userID == "" {
+		msg := "error getting authorizations"
+		err = ErrService.Wrap(err)
+		service.log.Error(msg, zap.Error(err))
+		return nil, status.Error(codes.InvalidArgument, msg)
```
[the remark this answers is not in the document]
ifraixedes: For me also, they shouldn't depend on protocol buffers structures, but in Storj they do, so not claiming, just saying.

pkg/certificate/authorization/service.go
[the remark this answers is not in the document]
ifraixedes: I thought that they don't have to return gRPC status codes because they shouldn't depend on the transport/communication protocol. However, in my opinion, that line is violated accepting protocol buffers structures, but Storj services use them.

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// GetOrCreate will return an authorization for the given user ID.
+func (service *Service) GetOrCreate(ctx context.Context, userID string) (_ *Token, err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if userID == "" {
+		msg := "error getting authorizations"
+		err = ErrService.Wrap(err)
+		service.log.Error(msg, zap.Error(err))
```
ifraixedes: There is no need to log this error is a user input error.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (endpoint *Endpoint) handleAuthorization(writer http.ResponseWriter, httpReq *http.Request) {
+	var err error
+	ctx := httpReq.Context()
+	mon.Task()(&ctx)(&err)
+
+	if httpReq.Method != http.MethodPut {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
```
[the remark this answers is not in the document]
ifraixedes: the `msg` var is used for the _HTTP error_ response too.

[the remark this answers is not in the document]
egonelbre: The issue is that this can cause problems. Let's say someone submits a method `X%d`, then the `ErrEndpoint.New` would interpret that `%d` as a formatting argument causing an error. In some really bad scenarios this can lead to an injection attack.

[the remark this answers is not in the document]
ifraixedes: For me `.New` should be resilient to it when no arguments are passed.

On the other hand, for this case, we are reading an _http_ method, so I see quite difficult to inject anything.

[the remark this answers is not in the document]
egonelbre: Yeah, I don't think there's an injection possibility here.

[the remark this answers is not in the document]
ifraixedes: Nonetheless, it was a good point, which I didn't think on it until you mentioned.

pkg/certificate/authorization/service.go
```diff
@@ -0,0 +1,82 @@
...
+// NewService creates a new authorization service.
+func NewService(log *zap.Logger, db *DB) *Service {
+	return &Service{
+		log: log,
+		db:  db,
+	}
+}
+
+// GetOrCreate will return an authorization for the given user ID
+func (service *Service) GetOrCreate(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
```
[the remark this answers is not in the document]
ifraixedes: Oh, that's what I commented.
For me, the services shouldn't depend on protocol buffers, but I've seen several services that they do, so I thought that it is the way to go in Storj.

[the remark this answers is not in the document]
egonelbre: It's more of we haven't fixed everything, but preference should be to avoid them, since it ties our internal code to protobuf, which makes fixes to protobuf harder.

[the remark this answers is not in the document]
ifraixedes: > It's more of we haven't fixed everything,

Good to know, so I can point to that in new services.

> but preference should be to avoid them, since it ties our internal code to protobuf, which makes fixes to protobuf harder.

Exactly that, it's because I said that "in my opinion, they shouldn't depend on protobuf".
Thanks for the clarification.

I'm going to improve https://github.com/storj/docs/blob/master/code/Service.md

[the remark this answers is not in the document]
egonelbre: Just to clarify there are some messages that need to use protobuf for serialization, like `pb.OrderLimit` since the signature is based on it... and we don't have a good alternative at the moment. But for new services we should prefer either new structs or regular arguments.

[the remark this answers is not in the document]
ifraixedes: That's the example in the document, however we should think in replace the example by one that it doesn't make any exception and make the people raise the point when they encounter one.

Thanks for the clarification.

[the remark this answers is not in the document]
ifraixedes: https://github.com/storj/docs/pull/110

pkg/certificate/authorization/endpoint_test.go
```diff
@@ -15,60 +15,8 @@ import (
 
 	"storj.io/storj/internal/errs2"
 	"storj.io/storj/internal/testcontext"
-	"storj.io/storj/pkg/pb"
 )
 
-func TestEndpoint_Create(t *testing.T) {
```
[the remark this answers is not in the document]
bryanchriswhite: moved to service_test.go

pkg/certificate/authorization/service.go
bryanchriswhite: sorry this keeps happening.. it's a copy/paste/neglect issue on my part.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+	}
+
+	mux.HandleFunc("/v1/authorization", endpoint.handleAuthorization)
+
+	return endpoint
+}
+
+// Create creates an authorization from the given authorization request.
+func (endpoint *Endpoint) Create(ctx context.Context, req *pb.AuthorizationRequest) (_ *pb.AuthorizationResponse, err error) {
```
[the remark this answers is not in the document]
bryanchriswhite: Probably not.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if httpReq.Method != http.MethodPut {
+		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
+		err = ErrEndpoint.New(msg)
+		endpoint.log.Error(msg, zap.Error(err))
+		http.Error(writer, msg, http.StatusMethodNotAllowed)
+		return
+	}
+
+	userID, err := ioutil.ReadAll(httpReq.Body)
```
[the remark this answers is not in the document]
bryanchriswhite: I think it should be part of the path rather than a query string param but I take your point. Originally I was using POST but PUT makes more sense and I didn't think about how that should effect the route.

pkg/certificate/authorization/endpoint.go
```diff
@@ -88,7 +92,6 @@ func (endpoint *Endpoint) handleAuthorization(writer http.ResponseWriter, httpRe
 	if httpReq.Method != http.MethodPut {
 		msg := fmt.Sprintf("unsupported HTTP method: %s", httpReq.Method)
 		err = ErrEndpoint.New(msg)
```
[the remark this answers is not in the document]
bryanchriswhite: err is only here for use with the `mon.Task` call.

pkg/certificate/peer.go
```diff
@@ -43,23 +45,28 @@ type Peer struct {
...
-	Server *server.Server
+	Server          *server.Server
+	AuthorizationDB *authorization.DB
 
 	// services and endpoints
-	Certificates struct {
-		AuthorizationDB *authorization.DB
-		Endpoint        *Endpoint
+	Certificate struct {
+		Endpoint *Endpoint
```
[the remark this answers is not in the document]
bryanchriswhite: Do you mean in addition to in the endpoint struct or instead of?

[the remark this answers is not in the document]
egonelbre: In addition to Endpoint. It somewhat keeps things clearer if the endpoints don't initialize or run services themselves.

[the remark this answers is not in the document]
bryanchriswhite: makes sense

pkg/certificate/peer.go
```diff
@@ -43,23 +45,29 @@ type Peer struct {
...
+	Authorization struct {
+		Listener net.Listener
+		Service  *authorization.Service
+		Endpoint *authorization.Endpoint
 	}
 }
 
 // New creates a new certificates peer.
-func New(log *zap.Logger, ident *identity.FullIdentity, ca *identity.FullCertificateAuthority, authorizationDB *authorization.DB, revocationDB *revocation.DB, config *Config) (*Peer, error) {
+func New(log *zap.Logger, ident *identity.FullIdentity, ca *identity.FullCertificateAuthority, authorizationDB *authorization.DB, revocationDB *revocation.DB, config *Config) (_ *Peer, err error) {
```
ifraixedes: I think that there is no need to declare `err` in the return values.

pkg/certificate/authorization/service.go
ifraixedes: This has been forgotten.

pkg/certificate/authorization/endpoint.go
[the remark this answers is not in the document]
ifraixedes: That's good, let's add a comment because I didn't think on it when I saw @mniewrzal comment and I also thought that it should be removed.

pkg/certificate/peer.go
bryanchriswhite: `err` is required for line 90:
```
	peer.Authorization.Listener, err = net.Listen("tcp", config.AuthorizationAddr)
```

ifraixedes: For those cases, the var should be declared in the function body.
Named return parameters should be only used for at least one of these cases

* documentation purpose.
* Needed because the value must be referenced in a `defer` statement.

They shouldn't be used only for the sake of declaring a variable.

pkg/certificate/authorization/service.go
bryanchriswhite: Ah, thanks... i lost track of it.

pkg/certificate/authorization/endpoint.go
```diff
@@ -0,0 +1,113 @@
...
+	log      *zap.Logger
+	db       *DB
+	service  *Service
+	server   http.Server
+	listener net.Listener
+}
+
+// NewEndpoint creates a authorization endpoint.
+func NewEndpoint(log *zap.Logger, db *DB, listener net.Listener) *Endpoint {
+	service := NewService(log, db)
```
thepaul: Elsewhere in the codebase our endpoints are largely stateless and don't create or own objects like this *Service. Can we make it so the Service is created in peer.go (or wherever needs to call NewEndpoint()) and the Service pointer gets passed in to NewEndpoint() as a parameter?

pkg/certificate/authorization/endpoint.go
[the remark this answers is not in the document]
thepaul: This still seems like an issue to be addressed, even if the worst outcome is that it causes a panic. If nothing else, can it just be `err = ErrEndpoint.New("%s", msg)` to avoid confusion as to whether it's meant to be a format template or not?

bryanchriswhite:
Is it actually, finally over?

PR #2972, opened 2019-09-06
storagenode/pieces: fix race condition in cache service

cam-a:
What: Add a mutex lock in `PersistCacheCacheTotals`

Why: When data is uploaded to a storage node it writes the amount to a map to be periodically synced to the database. If a write occurs simultaneously with a sync, it can result in a race condition

https://storjlabs.atlassian.net/browse/V3-2568

Please describe the tests:
 - Test 1: a test to ensure this situation no longer results in a race
 

storagenode/pieces/cache_test.go
```diff
@@ -332,3 +332,18 @@ func TestCacheCreateMultipleSatellites(t *testing.T) {
...
 }
+
+func TestConcurrency(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 0,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		node := planet.StorageNodes[0]
+		satellite := planet.Satellites[0]
+
+		go func() {
```
egonelbre: All goroutines must finish before exiting the test.

storagenode/pieces/cache_test.go
```diff
@@ -332,3 +332,21 @@ func TestCacheCreateMultipleSatellites(t *testing.T) {
...
 }
+
+func TestConcurrency(t *testing.T) {
+	testplanet.Run(t, testplanet.Config{
+		SatelliteCount: 1, StorageNodeCount: 1, UplinkCount: 0,
+	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		node := planet.StorageNodes[0]
+		satellite := planet.Satellites[0]
+
+		done := make(chan bool)
```
egonelbre: Use  `ctx.Go` (or `errgroup.Group`) rather than channels and `go func`. They are easy to get wrong. While the current code is correct, I'd prefer if we didn't use `go func` directly, unless exceptional cases.

PR #2978, opened 2019-09-09
satellite/server diagnostics and http headers improved

crawter:
What: 
1. added X-Content-Type-Options header to prevent MIME-type sniffing
2. added all headers to index.html file
3. added robots.txt file and appropriate handler to improve seo
(https://en.wikipedia.org/wiki/Robots_exclusion_standard)

Why: to improve some standard website metrics 

Please describe the tests:
 - Test 1:
 - Test 2:
 

web/satellite/robots.txt
```diff
@@ -0,0 +1,3 @@
+User-agent: *
+Disallow:
+Disallow: /cgi-bin/
```
egonelbre: Instead of serving this from a file, I think we can hard-code this.

satellite/console/consoleweb/server.go
```diff
@@ -140,6 +144,9 @@ func (server *Server) Run(ctx context.Context) (err error) {
 		return Error.Wrap(err)
 	}
 
+	mime.AddExtensionType(".ttf", "font/ttf")
+	mime.AddExtensionType(".txt", "text/plain")
```
egonelbre: These should be called in an init func.

satellite/console/consoleweb/server.go
```diff
@@ -437,6 +445,15 @@ func (server *Server) grapqlHandler(w http.ResponseWriter, r *http.Request) {
 	sugar.Debug(result)
 }
 
+func (server *Server) seoHandler(w http.ResponseWriter, req *http.Request) {
+	header := w.Header()
+
+	header.Set(contentType, mime.TypeByExtension(".txt"))
+	header.Set(contentTypeOptions, "nosniff")
```
egonelbre: Might as well avoid the extra variable, it doesn't add much additional clarity.

satellite/console/consoleweb/server.go
crawter: in NewServer or initializeTemplates?

egonelbre: `func init() {` at the global level

web/satellite/robots.txt
rikysya: Why do we need to add production level settings to our public development codebase?

egonelbre: Yeah, I was thinking of reducing possibilities of getting the robots file wrong. But I guess it doesn't matter that much.

This robots won't matter for dev environment.

But, yeah, I think the implementation might end up simpler if we don't have it.

crawter: this is the very base implementation of robots.txt file. The content of robots.txt could be taken from config in future

crawter: we already have a few reasons to modify config of satellite server and we can include this one also. But having "/robots.txt" path is a standard metric for web application

rikysya: I was thinking that this endpoint can be opened by some `proxy`, like we expose a server through `https proxy` to outer world. As `SEO` settings is individual for every `satellite owner` (every satellite is indexed differently) it should be configured as deployments step but not here.

satellite/console/consoleweb/utils.go
```diff
@@ -14,6 +15,18 @@ import (
 	"storj.io/storj/satellite/console/consoleweb/consoleql"
 )
 
+func init() {
+	err := mime.AddExtensionType(".ttf", "font/ttf")
+	if err != nil {
+		panic(err.Error())
```
VinozzZ: Should we just terminate the web process and log the error instead of panic here?

crawter: we don't have access to logger here

PR #2979, opened 2019-09-09
cmd: windows log file workaround

mniewrzal:
What: Workaround for `zap` issue https://github.com/uber-go/zap/pull/624

Why: We need logging to file for Windows

Please describe the tests:
 - Test 1: none
 

pkg/process/logging.go
```diff
@@ -40,6 +41,12 @@ func newLogger() (*zap.Logger, error) {
 		levelEncoder = zapcore.CapitalLevelEncoder
 	}
 
+	winFileSink := func(u *url.URL) (zap.Sink, error) {
+		// Remove leading slash left by url.Parse()
+		return os.OpenFile(u.Path[1:], os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
```
egonelbre: Would this panic when you pass in something like `winfile://`?

mniewrzal: yes, for `winfile://g:\test.log` -> `Error: couldn't open sink "winfile://g:\\test.log": can't parse "winfile://g:\\test.log" as a URL: parse winfile://g:\test.log: invalid port ":\\test.log" after host`

egonelbre: I mean just `winfile://`?

mniewrzal: oh, `Error: couldn't open sink "winfile://": no sink found for scheme "winfile"`

egonelbre: I was more worried that it will panic with an out-of-bounds.

mniewrzal: It looks it stops before entering `winFileSink()`. I pushed it as a draft because I didn't look too much into it, just to show idea.

pkg/process/logging.go
```diff
@@ -40,6 +41,15 @@ func newLogger() (*zap.Logger, error) {
...
+	}
+	err := zap.RegisterSink("winfile", winFileSink)
```
egonelbre: should this happen in a `func init()` rather than `newLogger`?

mniewrzal: yes, it needs to be registered only once so `init` will be better.

mniewrzal: hmm should be throw `fatal` in case of error in `init()`?

egonelbre: `panic` is pretty much the only option unfortunately

PR #2980, opened 2019-09-09
satellite/repair: Use k+1 pieces instead of k for repair

mobyvb:
What: 
Go back to using k+1 minimum pieces for repair so that we can take advantage of RS error detection until we can design a system that supports k pieces.

Why:
Since we can't rely on uplink signing keys being the same on all storagenodes for a segment (https://github.com/storj/storj/pull/2975), even if we are checking signed hashes, there is nothing preventing a storagenode from generating its own "uplink public key", signing a hash of incorrect data with that, updating its original order limit, and sending the incorrect data as a response to `GET_REPAIR`.
This pull request still allows us to take advantage of the hash checking of repair, which will allow us to detect failed audits during repair in the future, and to discard data that does not match the signed hash provided, but it will still decode with error correction in case any pieces are incorrect.

Please describe the tests:
 - Test 1:
 - Test 2:
 
### DIFF
--- a/satellite/repair/checker/checker.go
+++ b/satellite/repair/checker/checker.go
@@ -161,10 +161,10 @@ func (checker *Checker) updateIrreparableSegmentStatus(ctx context.Context, poin
 	numHealthy := int32(len(pieces) - len(missingPieces))
 	redundancy := pointer.Remote.Redundancy
 
-	// we repair when the number of healthy pieces is less than or equal to the repair threshold and is greater or equal to
+	// we repair when the number of healthy pieces is less than or equal to the repair threshold and is greater than
 	// minimum required pieces in redundancy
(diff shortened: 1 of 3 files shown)


mobyvb:
> > there is nothing preventing a storagenode from generating its own "uplink public key", signing a hash of incorrect data with that, updating its original order limit, and sending the incorrect data as a response to GET_REPAIR
> 
> Updating the order limit will not work. The storage node could create a new order limit and sign it but it is trivial for the satellite to notice that.

Ok, that makes sense. But we are not currently checking the satellite signature during repair verification, so we should at least make sure that verification is done before releasing repair in its current state. I'll work on a PR for that now.

mobyvb:
Turns out the satellite signature on the order limit is already verified. False alarm :1st_place_medal:

PR #2983, opened 2019-09-09
Handle migration for split storagenodedbs

isaachess:
What: This extends the migration system for storagenodedb to handle the new split sqlite dbs. Going forward each separate db maintains its own list of migrations which are aggregated into a single group of storagenodedb migrations.

Why: Now that we are splitting the storagenodedb into separate sqlite dbs, we need a way to handle migration properly.

Please describe the tests:
 - Test 1: The migrations_test has been updated to cover new functionality.
 
### DIFF
--- a/storagenode/storagenodedb/pieceinfo.go
+++ b/storagenode/storagenodedb/pieceinfo.go
@@ -21,7 +21,10 @@ import (
 // ErrPieceInfo represents errors from the piece info database.
 var ErrPieceInfo = errs.Class("v0pieceinfodb error")
 
-var v0PieceInfoDatabaseFilename = "pieceinfo.db"
+const (
+	PieceInfoDBName             = "pieceinfo"
+	V0PieceInfoDatabaseFilename = "pieceinfo.db"
+)
 
 type v0PieceInfoDB struct {
 	SQLDB
--- a/storagenode/storagenodedb/bandwidthdb.go
+++ b/storagenode/storagenodedb/bandwidthdb.go
@@ -21,7 +21,10 @@ import (
 // ErrBandwidth represents errors from the bandwidthdb database.
 var ErrBandwidth = errs.Class("bandwidthdb error")
 
-var BandwidthDatabaseFilename = "bandwidth.db"
+const (
+	BandwidthDBName           = "bandwidth"
+	BandwidthDatabaseFilename = "bandwidth.db"
+)
 
 type bandwidthDB struct {
 	// Moved to top of struct to resolve alignment issue with atomic operations on ARM
--- a/storagenode/storagenodedb/database.go
+++ b/storagenode/storagenodedb/database.go
@@ -203,7 +203,7 @@ func (db *DB) openDatabases(databasesPath string) error {
 		db.pieceExpirationDB.SQLDB = pieceExpirationDB
 	}
 
-	v0PieceInfoDB, err := db.openDatabase(db.sqliteDriverInstanceKey, filepath.Join(databasesPath, v0PieceInfoDatabaseFilename))
+	v0PieceInfoDB, err := db.openDatabase(db.sqliteDriverInstanceKey, filepath.Join(databasesPath, V0PieceInfoDatabaseFilename))
 	if err != nil {
 		return err
 	}
@@ -384,7 +384,15 @@ func (db *DB) RoutingTable() (kdb, ndb, adb storage.KeyValueStore) {
 // RawDatabases are required for testing purposes
 func (db *DB) RawDatabases() map[string]SQLDB {
 	return map[string]SQLDB{
-		"versions": db.versionsDB,
+		BandwidthDBName:       db.bandwidthDB,
+		OrdersDBName:          db.ordersDB,
+		PieceExpirationDBName: db.pieceExpirationDB,
+		PieceSpaceUsedDBName:  db.pieceSpaceUsedDB,
+		ReputationDBName:      db.reputationDB,
+		StorageUsageDBName:    db.storageUsageDB,
+		UsedSerialsDBName:     db.usedSerialsDB,
+		PieceInfoDBName:       db.v0PieceInfoDB,
+		VersionsDBName:        db.versionsDB,
 	}
 }
 
@@ -815,10 +823,10 @@ func (db *DB) Migration() *migrate.Migration {
 						return ErrDatabase.Wrap(err)
 					}
 
-					if err := sqliteutil.MigrateToDatabase(ctx, db.sqliteConnections, db.sqliteDriverInstanceKey, VersionsDatabaseFilename, v0PieceInfoDatabaseFilename, "pieceinfo_"); err != nil {
+					if err := sqliteutil.MigrateToDatabase(ctx, db.sqliteConnections, db.sqliteDriverInstanceKey, VersionsDatabaseFilename, V0PieceInfoDatabaseFilename, "pieceinfo_"); err != nil {
 						return ErrDatabase.Wrap(err)
 					}
-					if err := db.closeDatabase(v0PieceInfoDatabaseFilename); err != nil {
+					if err := db.closeDatabase(V0PieceInfoDatabaseFilename); err != nil {
 						return ErrDatabase.Wrap(err)
 					}
 
--- a/storagenode/storagenodedb/migrations_test.go
+++ b/storagenode/storagenodedb/migrations_test.go
@@ -5,6 +5,7 @@ package storagenodedb_test
 
 import (
 	"fmt"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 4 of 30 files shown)


storagenode/storagenodedb/pieceinfo.go
```diff
@@ -212,6 +212,23 @@ func (db *v0PieceInfoDB) GetExpired(ctx context.Context, expiredAt time.Time, li
 	return infos, nil
 }
 
+func (db *v0PieceInfoDB) Migration() *StorageNodeDBMigration {
```
egonelbre: As I talked in Slack there shouldn't be multiple versions since it complicates moving data from one database to another, or splitting and merging them. This would mean that we also need some sort of dependency resolver with regards to database versions.

isaachess: Can you help me understand what you mean here? I see your comment in slack now (https://storj.slack.com/archives/CAQV0AE2Z/p1567662737254000, I scrolled back to see what I missed), but I hadn't seen it until this moment. Can you help me understand your concerns better?

I don't see how it complicates moving data, splitting or merging them, personally. That can still be done via a migration.

isaachess: Do review what we discussed in person: The concern is that, without a global ordering of migration steps, it becomes harder to have migration dependencies and ordering between new databases. For example if I need to run 1 migration on db A, then 1 migration on new db B, then another on A, and these steps depend on each being run before the other. This scenario is very hard (almost impossible) in the way it's done.

We discussed a new way to do this, which I'll work to implement.

jtolds: @egonelbre - if we have multiple databases, there are fundamentally multiple versions. there is no escaping the fact that we lose atomicity across the independent databases. within a single database, you can atomically apply a migration or not. with multiple databases you can only atomically apply a migration or not _within a database_. otherwise, you have to manage the fact that power loss can happen at any time, and only some databases potentially migrated.

the only safe way to maintain the precondition that all databases are running the latest version is to keep track of the independent database versions independently.

before we consider anything besides a separate version per database, we need to hear a solution for how we keep track of partial migration status in the face of power loss across independent databases. in the meantime, a separate version per database is the right approach

jtolds: @egonelbre - perhaps i'm misunderstanding what you're saying.

let's say we have three independent databases, A, B, and C. my point is that a migration can only operate on A or B or C, but not any together. are you saying that we number the migrations such that migration 1 applies only to A, migration 2 applies only to C, and migration 3 applies only to B? so that there is a monotonic complete ordering of all migrations across all databases, even though database B only ever sees its own schema version as 2 or database A only ever sees its schema as version 1?

simongui: At first this code used a global schema version and if I understand our migration process, if the migration returns an error it will not bump up the schema version. That means the next time the storage node runs, it'll attempt the migration again.

This would cause the DB split to happen again. One of the first steps of the DB split is to detect if the split DB's exist and delete them. It should in theory re-run the migration because the tables from the original database haven't been dropped yet.

It may not be atomic but this makes it idempotent doesn't it? So it fails, we replay, no side effects.

I think the way @isaachess has this code working still retains that functionality but I wanted to bring it up to make sure.

isaachess: The db split migration (v 20 in your PR, @simongui) does not use a tx, so it's true that we need to manually handle the atomicity and idempotency of that migration. Sounds like you are doing that by what you've described.

This change does not affect the migration of splitting the databases, but it does ensure that migrations done on each database after the split each occur atomically.

egonelbre: @jtolds, yes, the migration can only work on one database at a time however I think we'll need migrations that cross the databases.

Let's take a simple scenario of migrations:

PR1:
```
Database A:
v1: creates table X with column G

Database B:
v1: creates table Y
```

PR2:
```
Database A:
v1: creates table X with column G

Database B:
v1: creates table Y
v2: adds column Y.N = A.X.G
```

PR:3
```
Database A:
v1: creates table X with column G
v2: renames X.G to X.H

Database B:
v1: creates table Y
v2: adds column Y.N = A.X.G
```

Assuming storage node starts on v0 for both databases, there now isn't a way to run either migration sequentially. This would mean that we need to introduce an additional mechanism to create a global ordering between the migration steps. By having a single migration definition we'll avoid this issue because it's defined as the global ordering.

For a single version table, yes, there would be a potential for a version not being committed even though the migration itself succeeded. However, there's a neat solution for it. Keep the version table replicated in all tables. That way you could still run the migrations, have global ordering and won't have problems with atomicity.

thepaul: I don't think JT is arguing for multiple migration definitions. The migration definitions can and should still have unique numbering. My point (and, I think, JT's) is that the different databases must all have their own versions- that is, their own idea of the most recent migration step they have successfully completed. If migration number 489 has some statements to run on both databases A and B, but the B part fails, then the migration step as a whole has failed. When the operator tries again later, presumably having fixed the reason for the failure, the migration engine sees that database A has already completed migration 489 but database B has not. It can cleanly and safely re-execute the B part of migration 489, and then declare that migration 489 has succeeded. Migration steps needn't be idempotent in general.

For the sake of better precision, I'm picturing something like:

```go
return &migrate.Migration{
    Table: "versions",
    Steps: []*migrate.Step{
        {
            Description: "create table X with column G, and table Y",
            Version: 1,
            DBAction: map[string]migrate.Action{
                "databaseA": migrate.SQL{`CREATE TABLE x (g INT)`},
                "databaseB": migrate.SQL{
                    `CREATE TABLE y (foobar INT)`,
                },
            },
        },
        {
            Description: "adds column Y.N",
            Version: 2,
            DBAction: map[string]migrate.Action{
                "databaseB": migrate.Func(func(log *zap.Logger, db migrate.DB, tx *sql.Tx) error {
                    gValue, err := getGValueFromTableXInDatabaseA()
                    if err != nil {
                        return err
                    }
                    _, err = tx.Exec(`ALTER TABLE y ADD COLUMN n INT DEFAULT $1`, gValue)
                    return err
                },
            },
        },
        {
            Description: "rename X.G to X.H",
            Version: 3,
            DBAction: map[string]migrate.Action{
                "databaseA": migrate.SQL{`ALTER TABLE x RENAME g TO h`},
            },
        },
    },
}
```

where databaseA and databaseB both have `versions` tables and the possibility exists for the values in those `versions` tables to be different when not all steps of a migration have completed. But `ABS(A.versions.version - B.versions.version)` should always be either `0` or `1`; otherwise something drastically wrong has happened and human intervention is required.

> For a single version table, yes, there would be a potential for a version not being committed even though the migration itself succeeded. However, there's a neat solution for it. Keep the version table replicated in all tables. That way you could still run the migrations, have global ordering and won't have problems with atomicity.

I'm not sure how you mean the part about keeping the version table replicated in all tables, but I don't think that the potential for uncommitted version changes has to be a problem. Can't we just make the version change be part of the same transaction as the migration steps that have to run on a given database? It either all gets committed together or none of it does.

egonelbre: Yeah, getting an exact example is a good idea. The first restriction would be that the migration steps are always contrained to modifying one database at a time:

```
return &migrate.Migration{
    Table: "versions",
    Steps: []*migrate.Step{
        {
            Description: "create table X with column G",
            Version: 1,
            DBAction: migrate.SQLWith(
                db.databaseA,
                `CREATE TABLE x (g INT)`
            ),
        },
        {
            Description: "create table Y",
            Version: 2,
            DBAction: migrate.SQLWith(
                db.databaseB,
                `CREATE TABLE y (foobar INT)`
            ),
        },
        {
            Description: "adds column Y.N",
            Version: 3,
            DBAction: migrate.SQLWithFunc(
                db.databaseB,
                func(log *zap.Logger, db migrate.DB, tx *sql.Tx) error {
                    gValue, err := getGValueFromTableXInDatabaseA()
                    if err != nil {
                        return err
                    }
                    _, err = tx.Exec(`ALTER TABLE y ADD COLUMN n INT DEFAULT $1`, gValue)
                    return err
                },
            },
        },
        {
            Description: "rename X.G to X.H",
            Version: 4,
            DBAction:  migrate.SQLWith(
                db.databaseA,
                `ALTER TABLE x RENAME g TO h`,
            ),
        },
    },
}
```

This restriction ensures that we don't accidentally cause migrations that depend on ordering between database. e.g. accidentally trying to do alter table in one database and in the other table the information based on it.

Otherwise, yes, that's what I mean't. Every step in the migration bumps the version in all the database rather than just one. This effectively "replicates" the version table between the databases.

thepaul: Good point about ordering! It sounds like we're on the same page, so :tada:!

isaachess: @egonelbre This all makes sense. Our current migration tools (snapshots and the `Migration` struct) all assume, right now, that updates only go on a single database, and only tracks the version on a single db. For example, the `Migration.Run` method takes a db, runs each step, and for each step updates the version in the single db that is passed in to `Run`.

Anyway, I'm happy to make this update to ensure we support a global ordering for each migration step, but let's meet briefly to ensure that we're on the same page about how to extend our existing code to support this.

egonelbre: I also talked with @jtolds that, we don't need the replicated version table.

isaachess: I don't know what we mean by "replicated." Right now the `Migration` struct handles updating the version table for the db when executing the migration step. But currently it always does this on the same db, so it seems it would be useful to have the `Migration` struct be able to manage the appropriate version table update for each db.

thepaul: > I also talked with @jtolds that, we don't need the replicated version table.

Does this mean without having a version table in each db? How would we replace the benefit of being able to bump the version number in the same transaction as the migration changes? Without that, there will always be a possibility of db changes having already been made but the version number not reflecting the fact—which means all migrations will need to be made idempotent, which can be quite a pain to get right in some cases.

isaachess: We will have a version table in each db.

thepaul: Oh okay. Cool.

isaachess:
Close in favor of https://github.com/storj/storj/pull/3012.

PR #2985, opened 2019-09-09
satellite/metainfo: Verify storagenode signature on satellite upload

mobyvb:
What: 
Verify the storagenode-signed piece hash sent to the satellite before including that piece in the pointer metainfo.

Why:
We do not want storagenodes to be audited for pieces they never received, and the uplink might provide us with piece info that has not been properly signed by the storagenode.
https://storjlabs.atlassian.net/browse/V3-1583

Please describe the tests:
 - Test 1:
 - Test 2:
 

mobyvb:
TODO: add a boolean flag to the pointer in metainfo that keeps track of whether the piece hashes have been verified (so we can distinguish from this change going forward)

satellite/metainfo/metainfo.go
```diff
@@ -475,18 +478,38 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
 	defer mon.Task()(&ctx)(&err)
 
 	if pointer.Type == pb.Pointer_REMOTE {
-		var remotePieces []*pb.RemotePiece
 		remote := pointer.Remote
+
+		nodeIDList := storj.NodeIDList{}
```
thepaul: The section that starts here and ends on line 494 is very nicely self-contained (the only inputs are `endpoint`, `ctx`, and `remote.RemotePieces`, and the only output is `peerIDMap` (and, ok, `err`). This should be broken out to a helper method with a signature something like

```go
func (endpoint *Endpoint) mapNodesFor(ctx context.Context, pieces []*pb.RemotePiece) (map[storj.NodeID]*identity.PeerIdentity, error) {
```

satellite/metainfo/metainfo.go
```diff
@@ -475,18 +479,38 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
+		var remotePieces []*pb.RemotePiece
 		allSizesValid := true
 		lastPieceSize := int64(0)
 		for _, piece := range remote.RemotePieces {
 
-			// TODO enable piece hash signature verification
+			// Verify storagenode signature on piecehash
+			peerID := peerIDMap[piece.NodeId]
+			if peerID == nil {
+				endpoint.log.Warn("Peer ID is nil for node", zap.String("nodeID", piece.NodeId.String()))
```
thepaul: It looks like this could only happen when we don't have the chain stored yet. Perhaps it would help for the error message to indicate this, so the future log reader doesn't have to go back and search. "Identity chain unknown for node" maybe?

pkg/pb/pointerdb.proto
```diff
@@ -56,6 +56,7 @@ message Pointer {
   google.protobuf.Timestamp expiration_date = 7 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
 
   bytes metadata = 8;
+  bool piece_hashes_verified = 9;
```
thepaul: Could we possibly make this an integer, and return the number of piece hashes that were successfully verified? There might not be any great way to let the uplink know whether some of the hashes failed verification otherwise, and I think that "we stored your pointer, but not with as much redundancy as you think" is information that the uplink needs to know.

thepaul: Oh, oops. Not here, this doesn't get returned to the client. Can we add such an integer to SegmentCommitResponse, instead

mobyvb: We can do something like that - the only issue is that we want to be able to distinguish pointers that were created after this change vs. before this change. And `0` is an ambiguous value for an integer in this case. Does it mean the pointer was created before this change or does it mean we successfully verified 0 hashes? Of course the latter is impossible if we end up storing the pointer, but I think it makes the purpose of this value less obvious.

thepaul: You're totally right, it's better to have a bare flag in the pointer. I had my structs mixed up.

thepaul: In case my second comment above got lost because it came seconds before yours, see above. :) I'm hoping for an int field in SegmentCommitResponse instead

mobyvb: We could still add a field in `SegmentCommitResponse` if you think it would be worthwhile.

thepaul: Yes please. Good idea

mobyvb: Done

satellite/metainfo/metainfo.go
mobyvb: Done

satellite/metainfo/metainfo.go
mobyvb: Done

satellite/metainfo/metainfo.go
```diff
@@ -1454,7 +1489,9 @@ func (endpoint *Endpoint) CommitSegment(ctx context.Context, req *pb.SegmentComm
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	return &pb.SegmentCommitResponse{}, nil
+	return &pb.SegmentCommitResponse{
+		SuccessfulPieces: len(pointer.Remote.RemotePieces),
```
thepaul:
```suggestion
		SuccessfulPieces: int32(len(pointer.Remote.RemotePieces)),
```

satellite/metainfo/metainfo.go
```diff
@@ -475,18 +479,30 @@ func (endpoint *Endpoint) filterValidPieces(ctx context.Context, pointer *pb.Poi
...
+		}
+
+		var remotePieces []*pb.RemotePiece
 		allSizesValid := true
 		lastPieceSize := int64(0)
 		for _, piece := range remote.RemotePieces {
 
-			// TODO enable piece hash signature verification
+			// Verify storagenode signature on piecehash
+			peerID := peerIDMap[piece.NodeId]
```
VinozzZ: nit: here we can do `peerID, ok := peerIDMap[piece.NodeId]` and then check `ok` is true or false

mobyvb: Done

PR #2989, opened 2019-09-10
satellite/overlay: don't use transport observers

thepaul:
What: 

Don't use transport observers to get information about node addresses or uptime. Those will come from the contact service (see #2952).

This probably can't pass tests or be merged into master until #2952 is merged first.

Why:

Honestly, I'm not totally sure, except that V3-2305 said to do it.

Please describe the tests:
 - Existing tests exercise node information updates and node uptime updates. If they continue to work, I consider this change to be sufficiently tested.
 
Performance may be mildly impacted, as a result of updating node information every time a node checks in, rather than every time we dial out to a node. On the other hand, I'm not entirely sure how often we dial out to nodes now, so maybe this will actually improve performance.

### DIFF
--- a/satellite/overlay/service.go
+++ b/satellite/overlay/service.go
@@ -402,44 +402,6 @@ func (service *Service) UpdateCheckIn(ctx context.Context, node NodeCheckInInfo)
 	return service.db.UpdateCheckIn(ctx, node, service.config.Node)
 }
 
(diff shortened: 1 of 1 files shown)


egonelbre:
It probably needs to disable some tests like `TestMergePlanets`, since Kademlia pings wouldn't be updating the respective tables.

PR #2992, opened 2019-09-10
satellite/accounting: implement tally as an observer

egonelbre:
What: Make tally use metainfo loop rather than separate iteration.

Why: Fixes https://storjlabs.atlassian.net/browse/V3-2398

Things to do:
- [x] Separate PR to make observers take `metainfo.Path` that has project id, bucket name, bucket id etc. rather than requiring observer to do it. https://github.com/storj/storj/pull/3003
- [x] Remove return argument from ProjectAccounting.SaveTallies, if possible. There shouldn't be a need to return it.
- [x] Fix BucketTally struct https://github.com/storj/storj/pull/3041
- [x] Fix tests.
- [x] Follow-up PR to remove `Iterate` from metainfo service. https://github.com/storj/storj/pull/3196

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/accounting/tally/tally.go
```diff
@@ -38,204 +34,185 @@ type Config struct {
...
-	} else {
-		if len(nodeData) > 0 {
-			err = t.storagenodeAccountingDB.SaveTallies(ctx, latestTally, nodeData)
-			if err != nil {
-				errAtRest = errs.New("Saving storage node data-at-rest failed : %v", err)
-			}
+		return Error.Wrap(err)
+	}
+	if lastTime.IsZero() {
+		lastTime = time.Now()
```
egonelbre: This changes the behavior to use the full duration of the looping, when there's no timestamp in the database.

satellite/accounting/tally/tally.go
```diff
@@ -38,204 +34,185 @@ type Config struct {
...
+	}
+
+	// add up all nodes and buckets
+	observer := NewObserver()
+	err = service.metainfoLoop.Join(ctx, observer)
+	if err != nil {
+		return Error.Wrap(err)
+	}
+	finishTime := time.Now()
```
stefanbenten: If i understand that correctly, we will have problems at some point, if the loop takes longer than 24 hours.
We are unable to account the storage for nodes within that timeframe then...

That has not changed with this PR, but made me aware of it.

egonelbre: Why would it cause problems?

stefanbenten: Accounting is getting more vague in terms of we pay them for files that could be deleted already by the time we account it.

egonelbre: Yes, that is the case.

stefanbenten: Thats an issue imo and another reason to keep the table size small and set ourself a maximum iteration time, possibly log the time it took to do a full iteration?

satellite/accounting/tally/tally.go
```diff
@@ -38,204 +34,185 @@ type Config struct {
...
+	bucket.MetadataSize += int64(len(pointer.Metadata))
+
+	// add node info
+	remote := pointer.GetRemote()
+	redundancy := remote.GetRedundancy()
+	segmentSize := pointer.GetSegmentSize()
+	minimumRequired := redundancy.GetMinReq()
+
+	if remote == nil || redundancy == nil || minimumRequired <= 0 {
+		// TODO: observer.log.Error("failed sanity check")
```
JessicaGreben: why is this todo instead of just adding the log?

satellite/accounting/tally/tally.go
```diff
@@ -38,204 +34,185 @@ type Config struct {
...
+		return nil
 	}
 
-	if last {
-		s.ObjectCount++
+	pieceSize := float64(segmentSize / int64(minimumRequired))
+	for _, piece := range pointer.GetRemote().GetRemotePieces() {
```
JessicaGreben: you could say `range remote.GetRemotePieces()` instead

satellite/accounting/tally/tally.go
egonelbre: Just forgot.

satellite/accounting/tally/tally.go
egonelbre: We probably don't need to check, since there wouldn't be a panic when there aren't pices.

satellite/accounting/projectusage_test.go
```diff
@@ -40,6 +40,8 @@ func TestProjectUsageStorage(t *testing.T) {
 	testplanet.Run(t, testplanet.Config{
 		SatelliteCount: 1, StorageNodeCount: 6, UplinkCount: 1,
 	}, func(t *testing.T, ctx *testcontext.Context, planet *testplanet.Planet) {
+		planet.Satellites[0].Accounting.Tally.Loop.Pause()
```
egonelbre: Had to disable it here, because live accounting has a bug: https://storjlabs.atlassian.net/browse/V3-2852

PR #2999, opened 2019-09-10
remove kademlia: create upsert query to update uptime 

JessicaGreben:
**What/Why:**
When a storage node checkins with the satellite, we need to store the check-in info in the `nodes` database table. Previously to save this info we were making multiple requests to the database, which included a request to retrieve existing node record, then another to update uptime and another to update the capacity and operator info. In this PR we combine those queries into one upsert query.

**Description of the tests:**
- updated existing tests that used UpdateUptime and replaced them with UpdateCheckIn
- add some other specific tests to confirm UpdateCheckin behavior

**Performance impact:**
I wrote a benchmark test to compare the old query for this versus the new upsert query.  The results show the sqlite old vs new is the same and we would expect that since we didn't write an upsert query for sqlite. The Postgres new upsert shows a strong improvement over the old query.
```
$ go test -bench=BenchmarkUpdateCheckIn .
goos: darwin
goarch: amd64
pkg: storj.io/storj/satellite/contact
BenchmarkUpdateCheckIn/Sqlite/old-12     	    3000	    466037 ns/op
BenchmarkUpdateCheckIn/Sqlite/new-12     	    3000	    465539 ns/op
BenchmarkUpdateCheckIn/Postgres/old-12   	     100	  10411219 ns/op
BenchmarkUpdateCheckIn/Postgres/new-12   	    1000	   2051137 ns/op
PASS
```


satellite/contact/benchmark_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/overlay"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func BenchmarkUpdateCheckIn(b *testing.B) {
```
egonelbre: Usually it's better to make an end-to-end test first then run the benchmark with the old code and then with the new code. Then you can compare the results with benchstat or benchcmp.

satellite/contact/benchmark_test.go
```diff
@@ -0,0 +1,105 @@
...
+			value := pb.Node{
+				Id: node.NodeID,
+				Address: &pb.NodeAddress{
+					Transport: pb.NodeTransport_TCP_TLS_GRPC,
+					Address:   node.Address.GetAddress(),
+				},
+			}
+			err := db.OverlayCache().UpdateAddress(ctx, &value, config)
+			if err != nil {
+				fmt.Println("err:", err)
```
egonelbre: `fmt.Println` -> `b.Fatal`

satellite/contact/benchmark_test.go
JessicaGreben: oh neat, good idea. thanks!

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+					THEN current_timestamp
+					ELSE nodes.last_contact_success
+				END,
+				last_contact_failure = CASE WHEN $16 IS FALSE
+					THEN current_timestamp
+					ELSE nodes.last_contact_failure
+				END,
+				-- this disqualified case statement resolves to: 
+				-- when (new.uptime_reputation_alpha /(new.uptime_reputation_alpha + new.uptime_reputation_beta)) <= config.UptimeReputationDQ
+				disqualified = CASE WHEN (($18::numeric*nodes.uptime_reputation_alpha + $19::numeric*(1+$20)/2)/(($18::numeric*nodes.uptime_reputation_alpha + $19::numeric*(1+$20)/2)+($18::numeric*nodes.uptime_reputation_beta + $19::numeric*(1-$20)/2))) <= $17
```
thepaul: I put some algebra on this and here's what came out the other end:

`CASE WHEN (($18::numeric*uptime_reputation_alpha + $19::numeric*$16::bool::int) / ($18::numeric*(uptime_reputation_alpha + uptime_reputation_beta) + $19::numeric)) <= $17`

it's a _little_ bit shorter ¯\\_(ツ)_/¯

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+			DO UPDATE
+			SET
+				total_uptime_count=nodes.total_uptime_count+1,
+				uptime_reputation_alpha=$18::numeric*nodes.uptime_reputation_alpha + $19::numeric*(1+$20)/2,
+				uptime_reputation_beta=$18::numeric*nodes.uptime_reputation_beta + $19::numeric*(1-$20)/2,
+				email=$6,
+				wallet=$7,
+				free_bandwidth=$8,
+				free_disk=$9,
+				address = CASE WHEN $2 = nodes.address
```
thepaul: I still think this is equivalent to just `address = $2`. If that's not the desired behavior, this may need a change.

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+					THEN nodes.address
+					ELSE $2
+				END,
+				uptime_success_count = CASE WHEN $16 IS TRUE
```
thepaul: This could be shortened a little with the same use-$16-as-integer trick:

`uptime_success_count = uptime_success_count + $16::bool::int,`

thepaul:
Oh, I guess the `Version` field and the failure to build are also minor concerns....

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+		_, err = cache.UpdateNodeInfo(ctx, node.NodeID, &pbInfo)
+		if err != nil {
+			return Error.Wrap(err)
+		}
+	case *pq.Driver:
+		// v is a single feedback value that allows us to update both alpha and beta
+		var v float64 = -1
+		uptimeSuccess := 0
+		lastContactSuccess := time.Time{}.UTC()
+		lastContactFailure := time.Time{}.UTC()
```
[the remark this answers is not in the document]
JessicaGreben: oops good catch, yes this should be time.Now() for `lastContactFailure`

satellite/satellitedb/overlaycache.go
JessicaGreben: true. lol

satellite/satellitedb/overlaycache.go
JessicaGreben: neat!

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+				piece_count, major, minor, patch, hash, timestamp, release, latency_90
+			)
+			VALUES (
+				$1, $2, $3, $4, $5,
+				$6, $7, $8, $9,
+				0, 0, $10, 1,
+				current_timestamp, current_timestamp, $11, $12,
+				$21, $22, $13, $14,
+				false, null,
+				0, 0, 0, 0, '', $15, false, 0
```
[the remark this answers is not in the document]
JessicaGreben: this is a great idea. let me give it a shot.

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+				audit_success_count, total_audit_count, uptime_success_count, total_uptime_count,
+				created_at, updated_at, last_contact_success, last_contact_failure,
+				audit_reputation_alpha, audit_reputation_beta, uptime_reputation_alpha, uptime_reputation_beta,
+				contained, disqualified,
+				piece_count, major, minor, patch, hash, timestamp, release, latency_90
+			)
+			VALUES (
+				$1, $2, $3, $4, $5,
+				$6, $7, $8, $9,
+				0, 0, $10, 1,
```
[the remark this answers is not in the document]
JessicaGreben: awesome!

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+				current_timestamp, current_timestamp, $11, $12,
+				$21, $22, $13, $14,
+				false, null,
+				0, 0, 0, 0, '', $15, false, 0
+			)
+			ON CONFLICT (id)
+			DO UPDATE
+			SET
+				total_uptime_count=nodes.total_uptime_count+1,
+				uptime_reputation_alpha=$18::numeric*nodes.uptime_reputation_alpha + $19::numeric*(1+$20)/2,
```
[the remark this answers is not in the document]
JessicaGreben: The way it is now, `v` aka `$20` can be either `1` or `-1`.
Replacing the variable `$20` with `$16::bool::int` changes the option of `1`/`-1` to `1`/`0`.  I'm not sure this would work for the negative case/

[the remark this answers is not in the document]
thepaul: Oh yeah, don't replace `$20` with `$16::bool::int` directly. You'd have to replace the whole expression `(1+$20)/2` with `$16::bool::int` for this to work. 

    (1+1)/2 = 1
    (-1+1)/2 = 0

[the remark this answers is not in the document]
JessicaGreben: ohhhh ok i see what you mean, let me try that out.

satellite/satellitedb/overlaycache.go
JessicaGreben: hmmm i might leave it alone, i feel nervous mucking with the math.

thepaul: Fair enough.

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,126 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+				created_at, updated_at, last_contact_success, last_contact_failure,
+				audit_reputation_alpha, audit_reputation_beta, uptime_reputation_alpha, uptime_reputation_beta,
+				contained, disqualified,
+				piece_count, major, minor, patch, hash, timestamp, release, latency_90
+			)
+			VALUES (
+				$1, $2, $3, $4, $5,
+				$6, $7, $8, $9,
+				0, 0, $10, 1,
+				current_timestamp, current_timestamp, $11, $12,
```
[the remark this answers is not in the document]
JessicaGreben: I think adding a `CASE WHEN` for args `$11` and `$12` is better, let me add that change

satellite/satellitedb/overlaycache.go
```diff
@@ -1395,61 +1389,62 @@ func (cache *overlaycache) UpdateCheckIn(ctx context.Context, node overlay.NodeC
...
+				$11, $12, $13, $14
 			)
 			ON CONFLICT (id)
 			DO UPDATE
 			SET
-				total_uptime_count=nodes.total_uptime_count+1,
-				uptime_reputation_alpha=$18::numeric*nodes.uptime_reputation_alpha + $19::numeric*(1+$20)/2,
-				uptime_reputation_beta=$18::numeric*nodes.uptime_reputation_beta + $19::numeric*(1-$20)/2,
+				address=$2,
+				last_net=$3,
```
thepaul: Good catch. I hadn't noticed that `last_net` wasn't being updated. How about `protocol`; can that ever change?

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,121 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+					THEN current_timestamp
+					ELSE nodes.last_contact_success
+				END,
+				last_contact_failure = CASE WHEN $10 IS FALSE
+					THEN current_timestamp
+					ELSE nodes.last_contact_failure
+				END,
+				-- this disqualified case statement resolves to: 
+				-- when (new.uptime_reputation_alpha /(new.uptime_reputation_alpha + new.uptime_reputation_beta)) <= config.UptimeReputationDQ
+				disqualified = CASE WHEN (($16::numeric*nodes.uptime_reputation_alpha + $17::numeric*$10::bool::int) / (($16::numeric*nodes.uptime_reputation_alpha + $17::numeric*$10::bool::int) + ($16::numeric*nodes.uptime_reputation_beta + $17::numeric*(NOT $10)::bool::int))) <= $15
```
thepaul: Here’s a potential problem- if the node meets the disqualification criteria but it is already disqualified, we are still updating the disqualification timestamp here. It might be difficult then to go back and see when a node was _first_ disqualified. A node operator might even take advantage of this on purpose, so they could tell the support team “but it just got DQ’d today!” when really it has been out for a month.

The solution probably has to be something like `disqualified = CASE WHEN (huge existing expression) <= $15 AND NOT nodes.disqualified`

JessicaGreben: ah great catch! let me fix that

satellite/satellitedb/overlaycache.go
JessicaGreben: yeah i wasn't sure about protocol. let me look into adding it

satellite/contact/endpoint.go
```diff
@@ -47,31 +48,14 @@ func (endpoint *Endpoint) Checkin(ctx context.Context, req *pb.CheckinRequest) (
...
-	nodeInfo := pb.InfoResponse{Operator: req.GetOperator(), Capacity: req.GetCapacity()}
-	_, err = endpoint.service.overlay.UpdateNodeInfo(ctx, peerID, &nodeInfo)
-	if err != nil {
-		return nil, Error.Wrap(err)
+	nodeInfo := overlay.NodeCheckInInfo{
+		NodeID:   peerID,
+		Address:  req.Address,
+		IsUp:     pingNodeSuccess,
+		Capacity: req.Capacity,
+		Operator: req.Operator,
```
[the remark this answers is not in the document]
JessicaGreben: why do we need version here? it doesn't look like the `nodes` table has a version field.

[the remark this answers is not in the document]
JessicaGreben: oops nvm looks like it is `major/minor/patch` is the version

satellite/contact/db_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/pb"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/overlay"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func TestUpdateCheckIn(t *testing.T) {
```
cam-a: Awesome test. Would you consider moving this into `satellite/overlay/service_test.go` since this is testing the overlay method rather than contact endpoint or service?

satellite/overlay/service.go
```diff
@@ -383,6 +396,12 @@ func (service *Service) UpdateUptime(ctx context.Context, nodeID storj.NodeID, i
 	return service.db.UpdateUptime(ctx, nodeID, isUp, lambda, weight, uptimeDQ)
 }
 
+// UpdateCheckIn updates a single storagenode's check-in info.
+func (service *Service) UpdateCheckIn(ctx context.Context, node NodeCheckInInfo) (err error) {
+	defer mon.Task()(&ctx)(&err)
```
cam-a: We need to resolve the IP address of the node here. See `Put`

cam-a: Alternatively, we could resolve the IP from satellite/contact endpoint and pass it in with NodeCheckInInfo

JessicaGreben: see which `Put`? how do you resolve the IP from the endpoint?

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,130 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+				-- this disqualified case statement resolves to: 
+				-- when (new.uptime_reputation_alpha /(new.uptime_reputation_alpha + new.uptime_reputation_beta)) <= config.UptimeReputationDQ
+				disqualified = CASE WHEN (($16::numeric*nodes.uptime_reputation_alpha + $17::numeric*$10::bool::int) / (($16::numeric*nodes.uptime_reputation_alpha + $17::numeric*$10::bool::int) + ($16::numeric*nodes.uptime_reputation_beta + $17::numeric*(NOT $10)::bool::int))) <= $15 AND nodes.disqualified IS NULL
+					THEN current_timestamp
+					ELSE nodes.disqualified
+				END;
+			`
+		_, err = cache.db.ExecContext(ctx, query,
+			// args $1 - $5
+			node.NodeID.Bytes(), node.Address.GetAddress(), node.LastIP, node.Address.GetTransport(), int(pb.NodeType_STORAGE),
```
cam-a: I believe node.Address.GetTransport() will be nil. The contact endpoint is not setting this field before passing NodeCheckInInfo to this method

JessicaGreben: oh yeah i forgot about that. i wonder if i should change it back to send the transport. Or if i should remove the transport here in the query. thoughts?

satellite/overlay/service.go
cam-a: https://github.com/storj/storj/blob/54855ef8a247ec445e66fe218579ed7194cf1448/satellite/overlay/service.go#L317

JessicaGreben: oh neat. great, let me add that here.

satellite/satellitedb/overlaycache.go
cam-a: I'm not sure why the transport is necessary, but I think passing it along is the safer bet

cam-a: Oh wait. That info isn't being sent by the node. But if it's a gRPC endpoint that is calling the overlay method, the transport must be gRPC right? I would add a Transport field to NodeCheckInInfo and set it to `pb.NodeTransport_TCP_TLS_GRPC` when passing it from the endpoint to the overlay

cam-a: Hmmm, then again we're basically hard coding this value everywhere. Hard code it, or pass it, up to you

JessicaGreben: hmm. hard one. maybe i will hard code for now for simplicity.

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,130 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+
+	if node.Address.GetAddress() == "" {
+		return Error.New("error UpdateCheckIn: missing the storage node address")
+	}
+
+	switch t := cache.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		value := pb.Node{
+			Id: node.NodeID,
+			Address: &pb.NodeAddress{
```
cam-a: nit - you can change this to `Address: node.Address` now

JessicaGreben: good catch.

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,127 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+func (cache *overlaycache) UpdateCheckIn(ctx context.Context, node overlay.NodeCheckInInfo, config overlay.NodeSelectionConfig) (err error) {
+	defer mon.Task()(&ctx)(&err)
+
+	if node.Address.GetAddress() == "" {
+		return Error.New("error UpdateCheckIn: missing the storage node address")
+	}
+
+	switch t := cache.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		value := pb.Node{
```
cam-a: add `LastIp: node.LastIP`

JessicaGreben: nice catch!

satellite/satellitedb/overlaycache.go
```diff
@@ -1338,3 +1338,128 @@ func populateUpdateFields(dbNode *dbx.Node, updateReq *overlay.UpdateRequest) db
...
+			config.AuditReputationAlpha0, config.AuditReputationBeta0, uptimeReputationAlpha, uptimeReputationBeta,
+			// args $15 - $17
+			config.UptimeReputationDQ, config.UptimeReputationLambda, config.UptimeReputationWeight,
+			// args $18 - $23
+			semVer.Major, semVer.Minor, semVer.Patch, node.Version.GetCommitHash(), node.Version.Timestamp, node.Version.GetRelease(),
+		)
+		if err != nil {
+			return Error.Wrap(err)
+		}
+		mon.FloatVal("UpdateCheckIn query execution time (seconds)").Observe(time.Since(start).Seconds())
```
egonelbre: Doesn't this give the exact same info as `defer mon.Task()(&ctx)(&err)`?

PR #3001, opened 2019-09-11
satellite/overlay: avoid large statement for piece counts

egonelbre:
What: Use array arguments instead of constructing a large single query.

Why: Concatenated SQL statements are harder to audit and often can end up slower.

```
Benchmark                                      old ns/op      new ns/op     delta
BenchmarkDB_PieceCounts/Sqlite/Update-32       2403696100     136937262     -94.30%
BenchmarkDB_PieceCounts/Sqlite/All-32          16454915       17416226      +5.84%
BenchmarkDB_PieceCounts/Postgres/Update-32     562646350      149697062     -73.39%
BenchmarkDB_PieceCounts/Postgres/All-32        13657257       14630751      +7.13%
```

PS: Each commit is separately reviewable.

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/overlay/piececount_test.go
```diff
@@ -0,0 +1,123 @@
...
+				},
+				LastIp: "0.0.0.0",
+			}, overlay.NodeSelectionConfig{}))
+		}
+
+		// check that they are initialized to zero
+		initialCounts, err := overlaydb.AllPieceCounts(ctx)
+		require.NoError(t, err)
+		require.Empty(t, initialCounts)
+		// TODO: make it actually return everything
```
stefanbenten: Is it fine to leave the TODO here and remove the code below?
Or better fix it directly? 👍

satellite/satellitedb/overlaycache.go
```diff
@@ -926,53 +926,66 @@ func (cache *overlaycache) AllPieceCounts(ctx context.Context) (_ map[storj.Node
...
-func (cache *overlaycache) sqliteUpdatePieceCounts(ctx context.Context, pieceCounts map[storj.NodeID]int) (err error) {
-	defer mon.Task()(&ctx)(&err)
+	switch t := cache.db.Driver().(type) {
+	case *sqlite3.SQLiteDriver:
+		err = cache.db.WithTx(ctx, func(ctx context.Context, tx *dbx.Tx) error {
+			query := tx.Rebind(`UPDATE nodes SET piece_count = ? WHERE id = ?`)
+			for _, count := range counts {
+				_, err := tx.Tx.ExecContext(ctx, query, count.Count, count.ID)
+				if err != nil {
+					return Error.Wrap(err)
```
stefanbenten: Wouldnt this mean that if a single update fails, we return out of the entire function and discard the other inserts.
And the piece count for some nodes could be updated by then, as i dont see that we rollback in an error case?

stefanbenten: I am not entirely up to speed how critical this number is..

satellite/overlay/piececount_test.go
egonelbre: I didn't want to fix it directly at this moment, otherwise the PR would grow too large. But I certainly can remove it.

satellite/satellitedb/overlaycache.go
egonelbre: > Wouldnt this mean that if a single update fails, we return out of the entire function and discard the other inserts.

Yes, but this would also mean that there's something critically wrong already. Either we ended up creating piece count updates for storage nodes that aren't in our table. Or somebody deleted the node from the table... either would be problematic.

There of course could be an issue with database.

Of course, we don't care that much here, since it's for sqlite, which is mainly for testing.

> And the piece count for some nodes could be updated by then, as i dont see that we rollback in an error case?

When `WithTx` returns an error, it will rollback the transaction.

egonelbre:
@stefanbenten this is not for metrics, this number is used for calculating garbage collection bloom filter sizes.

satellite/satellitedb/overlaycache.go
stefanbenten: Thanks, that should be ok then 👍

satellite/satellitedb/overlaycache.go
```diff
@@ -930,49 +931,61 @@ func (cache *overlaycache) AllPieceCounts(ctx context.Context) (_ map[storj.Node
...
-		sqlQuery += `(?::BYTEA, ?::BIGINT), `
-		args = append(args, nodeID, pieceCount)
-	}
-	sqlQuery = sqlQuery[:len(sqlQuery)-2]
-	// trim off the last comma+space
-	sqlQuery += `) newvals(nodeid, piece_count) WHERE nodes.id = newvals.nodeid`
-	_, err = cache.db.DB.ExecContext(ctx, cache.db.Rebind(sqlQuery), args...)
-	return err
-}
+	sort.Slice(counts, func(i, k int) bool {
```
VinozzZ: Just curious why not use `.SliceStable` here?

egonelbre: Because the node ids are guaranteed to be unique, so there wouldn’t be a difference in the result.

PR #3003, opened 2019-09-11
satellite/metainfo.Loop: use a parsed path for observers

egonelbre:
What: 

Why:

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/metainfo/loop.go
```diff
@@ -163,18 +176,36 @@ waitformore:
...
+				rawPath := item.Key.String()
 				pointer := &pb.Pointer{}
 
 				err = proto.Unmarshal(item.Value, pointer)
 				if err != nil {
 					return LoopError.New("unexpected error unmarshalling pointer %s", err)
 				}
 
-				nextObservers := observers[:0]
+				pathElements := storj.SplitPath(rawPath)
```
egonelbre: @mniewrzal did we have a func to do this parsing instead of calling `storj.SplitPath`?

satellite/repair/checker/checker.go
```diff
@@ -271,19 +271,22 @@ func (obs *checkerObserver) RemoteSegment(ctx context.Context, path storj.Path,
...
-		err = obs.irrdb.Delete(ctx, []byte(path))
+		err = obs.irrdb.Delete(ctx, []byte(path.Raw))
 		if err != nil {
 			obs.log.Error("error deleting entry from irreparable db", zap.Error(err))
 			return nil
 		}
 	} else if numHealthy < redundancy.MinReq && numHealthy < redundancy.RepairThreshold {
-		pathElements := storj.SplitPath(path)
+		// TODO: see whether this can be handled with metainfo.ScopedPath
+		pathElements := storj.SplitPath(path.Raw)
```
VinozzZ: I saw that you put `SegmentPath` onto `ScopedPath` in previous commits. Seems like it would be nice to use that here. What's the reason that you decided not to do that?

egonelbre: The reason is that here, it's not really a "segment path", but something else. I also suspect that this code is actually wrong in the first place.

Also, the concept of "segment path" isn't well defined... i.e. should it contain the segment number or not?... so yeah, opted out of implementing it for now.

PR #3005, opened 2019-09-11
storagenode/storagenodedb: prepare for multiple databases

isaachess:
What: This PR prepares the `migrate_test.go` file for the database split happening in storagenodedb. It allows us to get schemas for multiple databases, as well as define a "snapshot" of the multi-database state.

It also moves the snapshot sql into a go struct, for two reasons: (1) with multiple dbs it allows more explicit labeling of DB (rather than relying upon a file-path, for example), and (2) it allows us to re-use a previous state when that makes sense, without having to redefine the SQL each time.

Why: This is prep work to merge in the splitting of the storagenodedbs and will keep that final PR smaller.

Please describe the tests:
 - Test 1: the updated `migrate_test` function covers all use-cases.
 

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
```diff
@@ -0,0 +1,86 @@
...
+	"storj.io/storj/internal/dbutil/dbschema"
+	"storj.io/storj/internal/dbutil/sqliteutil"
+)
+
+// Snapshots is the global variable that stores all the snapshots for testing
+var Snapshots MultiDBSnapshots
+
+// MultiDBSnapshots provides a convenient list of MultiDBSnapshot
+type MultiDBSnapshots struct {
+	List []*MultiDBSnapshot
```
zeebo: Why not a `map[int]*MultiDBSnapshot`? Then we wouldn't need to sort and lookup would be constant time.

storagenode/storagenodedb/database.go
```diff
@@ -296,6 +296,13 @@ func (db *DB) RoutingTable() (kdb, ndb, adb storage.KeyValueStore) {
 	return db.kdb, db.ndb, db.adb
 }
 
+// RawDatabases are required for testing purposes
+func (db *DB) RawDatabases() map[string]SQLDB {
+	return map[string]SQLDB{
+		"versions": db.versionsDB,
+	}
+}
+
```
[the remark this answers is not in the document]
isaachess: Yes, in the next PR when we do the final split, I'll take care of that there.

storagenode/storagenodedb/storagenodedbtest/v14.go
```diff
@@ -0,0 +1,6 @@
+package storagenodedbtest
+
+var v14 = Snapshots.Add(&MultiDBSnapshot{
+	Version:   14,
+	Databases: v13.Databases,
+})
```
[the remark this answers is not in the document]
isaachess: They're not empty, they happen to be the same as the previous versions. This is one of the reasons it's convenient to have it in go, I don't have to duplicate a bunch of SQL in this case. :)

isaachess:
Calling it `versions` is not something that will survive the next PR, so I'm not too worried about it. In the next PR, when we use this for all the dbs, the naming will make sense for every db.

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
```diff
@@ -0,0 +1,107 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package storagenodedbtest
```
egonelbre: Keep the testdata in `testdata`. The `storagenodedbtest` is intended to be imported by other packages for running tests against storage node database.

egonelbre: The reason `storagenodedbtest` is a separate pacakge, is because you cannot have "test-only" exported things in a package, so it was moved to another package.

storagenode/storagenodedb/storagenodedbtest/v0.go
```diff
@@ -0,0 +1,128 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package storagenodedbtest
+
+var v0 = MultiDBSnapshot{
+	Version: 0,
+	Databases: Databases{
+		"versions": &DBSnapshot{
```
egonelbre: I think `&DBSnapshot{` can be replaced with just `{`.

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
isaachess: Ok, I'll move them into `testdata` and just call it `package testdata`.

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
```diff
@@ -0,0 +1,107 @@
...
+type Databases map[string]*DBSnapshot
+
+// DBSnapshot is a snapshot of a single DB. It separates the SQL and NewData,
+// and after LoadSnapshot is called provides the shema and data.
+type DBSnapshot struct {
+	SQL     string
+	NewData string
+
+	// These are populated by the LoadSnapshot method
+	Schema *dbschema.Schema
```
egonelbre: You can embed `*dbschema.Snapshot` here instead of separating them.

I also try to separate the "definition" from the processed thing when performance isn't a concern. This ends up as more code, but it can be clearer, because you don't change the existing entities that could get out of sync.

In this case I was even thinking now, what if we did it when we compare the states, rather than upfront.

egonelbre: The main reason it did the loading upfront previously was that it's slightly easier with files.

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
isaachess: Done

storagenode/storagenodedb/storagenodedbtest/multidbsnapshot.go
isaachess: I have separated the definitions. The one that contains the sql I called `MultiDBState`, and the parsed version that has the schema and data I called `MultiDBSnapshot`.

I chose not to embed the `*dbschema.Snapshot`, as it doesn't quite apply: that struct has a version number, stores the script, etc., and I prefer to explicitly name my variables, which I think makes sense in this instance.

I also changed so it does the conversion at checking time, rather than "load" them all at the beginning, as you suggested.

egonelbre: I was initially thinking of doing the loading in the comparison, i.e. the test that that https://github.com/storj/storj/pull/3005/files#diff-59b42b6304dea128437983b526d7ee58R123.. but the current approach is fine as well.

PR #3012, opened 2019-09-11
Add test for migration 21

isaachess:
What: Create migration test for step that splits into multiple dbs

Why: We need to test the migration into multiple dbs

Please describe the tests:
 - Test 1: This is an updated test, all is well.
 

isaachess:
That would be fine. I didn't want to use the `Filename` one as a generic db identifier, it felt weird to use that variable in that way. And I wanted them to be `const`, so I just made them separate.

If we combine them, I'd like to keep the genera `<name>DBName`, and find some other way to use the filename?

simongui:
Yeah, you know what, I can do that as part, let's just get this merged and tests passing. I'll see what I can do in my code to strip off the file extension and share one key for all our code.

PR #3038, opened 2019-09-12
docs/blueprints: referral manager v1

VinozzZ:
What: 

design doc for referral manager v1

Why:

Due to business requirement changes, we need to pivot from the previous design.

Please describe the tests:
 - Test 1:
 - Test 2:
 

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    results := make(map[satelliteURL]int)
+    for _, satellite := range satelliteURLs {
+        successCount := satellite.AddInvitationTokens(tokens)
+        results[satellite] = successCount
+    }
+    return results
+}
```
ethanadams: I noticed you are sharing state across calls in `referralManager.eligibleUsers`.  Is this necessary? I'm assuming you are doing this so the list of users they agreed to is what gets processed. But, what if 2 people are running these referrals at the same time?  Is it safe enough to just generate the list again in GenerateAndSendInvitationTokens since it's not likely it changed much since GetEligibleUsers was called (unless the another user ran referrals at the same time)? This would also reduce the amount of re-generations.

Maybe `GetEligibleUsers` could return a count and hash of the IDs to be processed and the CLI could pass the hash to `GenerateAndSendInvitationTokens` so that when the list gets generated again, it can be hashed and compared.
If they don't match, `GenerateAndSendInvitationTokens` could return an error.

Or if it is not critical for the user list to be exact, the CLI could just push the user count they agreed to back to `GenerateAndSendInvitationTokens`.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,196 @@
...
+    - if it is a success, the satellite will proceed with the account creation and
+    then send a request to the Referral Manager to save the newly created user ID along with the used token.
+    - if it is an invalid token, the satellite will display a proper message in the UI.
+4. Referral Manager reaches out to the host satellite holding the redeemed token, so that it can be removed from the UI.
+    
+_User Interface For Displaying Referral Link_
+
+1. Users will have a referral tab on the UI. On click, it will try to retrieve referral tokens from the backend:
+    - if the front-end receives tokens, it should display them.
+    - if the front-end receives empty response payload, it should display a message `No available referral link. Try again later.`
```
ethanadams: I'm assuming this is the Satellite UI, and the backend is the satellite DB?

docs/blueprints/referral-manager-v1.md
VinozzZ: Good point! In this case, I don't think it's critical for the user list to be exact. As long as we can make sure that the total user count is the same, we can proceed with the `GenerateAndSendInvitationTokens`. I will update the code to check the user count instead. 
What do you mean by `This would also reduce the amount of re-generations.` ? From what I understand, the generated user list will be saved in memory. If another user ran `GetEligibleUsers` at the same time, as long as we have the same total user count, it should be fine. Is it not a good idea to run it again?

mobyvb: Discussed this with @VinozzZ and we think we could solve this problem the following way:
`GetEligibleUsers` checks `referralManger.eligibleUsers` before starting. If `referralManager.eligibleUsers` is not nil, it means one of two things:
1. Someone else is using the referral manager at the same time and has not generated/sent tokens yet (very unlikely)
2. The CLI was ctrl+Cd between the get users step and the generate/send tokens step (more likely)

So if `GetEligibleUsers` is called and `referralManager.eligibleUsers` is set to something other than nil, we return an error and the above information is conveyed to the CLI user. Then the CLI user has the option to abort or to continue anyway. If they decide to continue, we call a new endpoint like `ClearUsers` which sets `eligibleUsers` to nil and start over  with `GetEligibleUsers`.

We just need to make sure `GenerateAndSendInvitationTokens sets `referralManager.eligibleUsers` to nil when it finishes.

ethanadams: Works for me since this is really an internal tool anyway.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,197 @@
...
+
+Pseudocode
+
+The Token struct represents both a Go struct (on satellite and Referral Manager) as well as the schema for the `tokens` table on the Referral Manager
+
+```
+type Token struct {
+	Secret [32]byte
+	OwnerID uuid.UUID
+	RedeemedBy uuid.UUID
```
mobyvb: Thoughts on changing `RedeemedBy` to `RedeemedID` to make it more consistent with `OwnerID`?

docs/blueprints/referral-manager-v1.md
VinozzZ: i just updated the doc to reflect this

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,229 @@
...
+
+The Token struct represents both a Go struct (on satellite and Referral Manager) as well as the schema for the `tokens` table on the Referral Manager
+
+```
+type Token struct {
+	Secret [32]byte
+	OwnerID uuid.UUID
+	RedeemedID uuid.UUID
+	HostSatelliteURL string
+    RedeemedSatelliteURL string
```
thepaul: the difference in indentation here looks like an easy mistake, but just to make sure- should this be a tab instead of 4 spaces?

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,229 @@
...
+        - if there are unredeemed tokens, Referral Manager respond satellite's request with the tokens.
+        - if no unredeemed token is found for this user ID and such user ID doesn't exist in `users` table, Referral Manager will add an entry for this user ID in `users` table.
+    - Satellite receives the response from Referral Manager.
+        - if the response playload contains tokens, satellite will display them in the UI.
+        - if its an empty response payload, satellite will display a message `No available referral link. Try again later.`
+
+
+_Referral Manager CLI_
+
+1. `referral manager start` initiates the invitation token generation process. This process begins by checking whether there are existing `eligibleUsers` in memory in Referral Manager.
```
thepaul: What is `eligibleUsers` conceptually? Is it just a list of all eligible users that the referral manager knows about? What does it mean for a user to be eligible?

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,229 @@
...
+: A CLI that will allow operators to invoke referral link generation.
+
+Referral Link
+: A one-time-use URL that contains a unique invitation token.
+
+Invitation Token
+: A random number generated by Referral Manager for each user in the referral program.
+
+Host Satellite
+: The satellite that a particular token is stored on before it is redeemed.
```
thepaul: What does it mean for a token to be redeemed? Does that mean someone is signing up for an account using a `referral link` that somebody else gave them? Can tokens be redeemed only once, then? I'm having some trouble picturing the structure of the overall referral program. If I want Fred Rogers to sign up, do I need to go to the satellite web interface to request that a new fresh `invitation token` be generated, so I can give it to Fred? I don't just get a "referral URL" that I can give to everyone I want?

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,229 @@
...
+: A stand-alone process running outside of satellites. It should only talk to tardigrade branded satellites.
+
+Referral Manager CLI
+: A CLI that will allow operators to invoke referral link generation.
+
+Referral Link
+: A one-time-use URL that contains a unique invitation token.
+
+Invitation Token
+: A random number generated by Referral Manager for each user in the referral program.
```
thepaul: So each user gets one, and only one, invitation token? Do we not want to allow people to refer more than once?

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,229 @@
...
+type Token struct {
+	Secret [32]byte
+	OwnerID uuid.UUID
+	RedeemedID uuid.UUID
+	HostSatelliteURL string
+    RedeemedSatelliteURL string
+	Status string
+}
+```
+Statuses: `unsent` (host satellite doesn't know about it yet), `unredeemed` (host satellite knows about it but it has not been used), `redeemed` (someone has used this referral link to register alreadys)
```
thepaul: How can the host satellite not know about a token, when the definition of "host satellite" is "the satellite that a particular token is stored on before it is redeemed"?

docs/blueprints/referral-manager-v1.md
VinozzZ: added the `eligible users` in `Definition` section

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,73 @@
...
+Invitation Token
+: An random number generated by Referral Manager for each user to be able to participate in our referral program.
+
+### How it will work
+
+_Referral Link Generation_
+
+1. Marketing team distributes new referral links through Referral Manager CLI, which asks tardigrade satellites for userIDs.
+2. Satellite receives user infos request, which:
+    - starts gathering of users whose current count of remaining invitation token is 0.
```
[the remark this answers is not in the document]
VinozzZ: business requirement. so everyone will have a chance to get a referral link

docs/blueprints/referral-manager-v1.md
VinozzZ: discussed in our call

docs/blueprints/referral-manager-v1.md
VinozzZ: changed to `owner's satellite`

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,232 @@
...
+
+Redeemed Satellite
+: The satellite where the new user redeeming a token has registered on.
+
+### How it will work
+
+_User Interface_
+
+1. Users will have a referral tab on the UI. On click, it will trigger satellite to send a request to retrieve referral links from Referral Manager:
+    - Referral Manager will query unredeemed tokens in `tokens` table for this user ID.
```
mobyvb: nit - this user ID and satellite

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,232 @@
...
+- Create a private repository for Referral Manager.
+- Create `tokens` and `users` table in Referral Manager database.
+- Create an `inspector` endpoint on Referral Manager.
+    - Implementing a method `GenerateTokens` for generating referral tokens and save them into `tokens` table on Referral Manager.
+    - Implementing a mathod `GetEligibleUsers` for querying users who don't have any referral tokens associated with.
+- Implementing an endpoint `GetTokens` for requesting unredeemed tokens from Referral Manager. 
+- Implementing an endpoint `Redeem` on Referral Manager for verifying invitation tokens.
+- Implementing an endpoint `ClearUsers` on Referral Manager for setting in-memory `eligibleUsers` to be empty from Referral Manager CLI.
+- Replace existing registration token logic.
+
```
mobyvb: Also implement an endpoint `RegisterUser` on the Referral Manager for registering a new user

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,232 @@
...
+```
+type User struct {
+    ID uuid.UUID
+    satelliteURL string
+}
+```
+
+
+**Endpoints for satellites:**
```
mobyvb: We should also add `RegisterUser` here

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,232 @@
...
+    fmt.Printf("Successfully created %d tokens.\n", len(newTokens))
+}
+```
+
+## Wrapup
+
+Team Green will be responsible implementing this blueprint.
+
+## Open issues
+1. The existing token is implemented with 32 random bytes.
```
mobyvb: I don't think we need to keep this as an open issue.

docs/blueprints/referral-manager-v1.md
mobyvb: Actually, should that be a part of `Redeem`? Maybe we don't need a separate `RegisterUser` endpoint.

VinozzZ: Yea, i imagined it to be a part of the `Redeem`

VinozzZ: Added the pseudocode

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+: A CLI that will allow operators to invoke referral link generation.
+
+Referral Link
+: A one-time-use URL that contains a unique invitation token.
+
+Invitation Token
+: A random number generated by Referral Manager for each user in the referral program.
+
+Eligible Users
+: Users who have 0 unredeemed referral link.
```
jtolds: maybe this should be "less than x unredeemed referral links" so we can top people off. let's say everyone gets 3 referral links and someone uses 2. we might want to top them back off to 3

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+        - if there are unredeemed tokens, Referral Manager respond satellite's request with the tokens.
+        - if no unredeemed token is found for this user ID and such user ID doesn't exist in `users` table, Referral Manager will add an entry for this user ID in `users` table.
+    - Satellite receives the response from Referral Manager.
+        - if the response playload contains tokens, satellite will display them in the UI.
+        - if its an empty response payload, satellite will display a message `No available referral link. Try again later.`
+
+
+_Referral Manager CLI_
+
+1. `referral manager start` initiates the invitation token generation process. This process begins by checking whether there are existing `eligibleUsers` in memory in Referral Manager.
```
jtolds: we should just use the database. no sense in having a list in memory and in the database. just query the database for eligible users directly and combine these eligibleUsers and users query steps

jtolds: specifically, the cli should just find eligible users and create them tokens and store them back in the database. if that happens in a transaction it's okay to run multiple copies of the CLI together

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+    - If `eligibleUsers` is empty, the Referral Manager queries `users` table to get users with 0 remaining invites.
+    - If `eligibleUsers` is not empty, it means one of two things:
+        * Someone else is using the Referral Manager at the same time and has not generated/sent tokens yet.
```
jtolds: this isn't something you need to worry about if you update the database in a transaction. 

at a high level:
```
1. cli asks referral manager to start
2. referral manager gets a list of eligible users from the database
3. referral manager loops over list. for each one (or for a batch of them), start a transaction, make sure they're still eligible, generate tokens and insert them into the right table, commit the transaction
```

multiple clis running concurrently will work correctly, and you don't need to keep some separate in-memory database

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+        * The CLI was terminated between the get `eligibleUsers` step and the generate/send tokens step.
+    
+        Either one of the cases, We will return an error and display `Referral token generation already in progress or canceled early. Do you want to continue anyway?`. If they decide to continue, we call `ClearUsers` to sets eligibleUsers to empty and the Referral Manager queries `users` table to get users with 0 remaining invites. 
+2. The Referral Manager returns the number of users to the CLI.
+3. The CLI receives the number of users and displays: `How many invitation do you want to generate for each user?`
```
jtolds: this should have just been a command line argument to the process to kick off generation. no need to do this question/answer after having started.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+4. After the user enters the number of invites to generate, the CLI displays: `Generating X amount of invitation tokens for Y amount of users. Yes/No.`
```
jtolds: unnecessary - if you really want to be careful, have the cli have a --dry-run commandline option that just outputs what it would have done. no need to do input

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+    - if the token exists in `tokens` table:
+        - if it is not redeemed, Referral Manager sends back a success response to the redeemed satellite and mark the token as redeemed in the Referral Manager's database
+        - if it token is already redeemed, Referral Manager sends back a `invalid token` response to the redeemed satellite.
+        
+    - if the token doesn't exist in `tokens` table, Referral Manager sends back a `invalid token` response to the redeemed satellite.
+3. Redeemed satellite receives the response, which:
+    - if it is a success, the satellite will proceed with the account creation and
+    then send a request to the Referral Manager, which: 
+        - updates the `tokens` table to store the newly created user ID in `redeemedID` column for that particular referral token.
+        - updates the `users` table to store the newly created user ID and the satellite URL the user belongs to.
```
jtolds: can we combine this referral manager step with the previous referral manager step? so that redeeming the token also updates the tokens and users table with the new user id? this would mean that the user ids would need to be speculatively generated in advance of trying to use a token, but that's okay. we could have some sort of referral manager user id that can be something like 32 random bytes of data and then there's no risk of collision, but also no cost if we throw it out

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+: The satellite that an owner of a particular token belongs to.
+
+Redeemed Satellite
+: The satellite where the new user redeeming a token has registered on.
+
+### How it will work
+
+_User Interface_
+
+1. Users will have a referral tab on the UI. On click, it will trigger satellite to send a request to retrieve referral links from Referral Manager:
```
egonelbre: Will it do this on every request? Is there a debounce? Is there a cache?

docs/blueprints/referral-manager-v1.md
egonelbre: In addition, having things in memory would horizontal scaling more difficult.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+    - If `eligibleUsers` is not empty, it means one of two things:
+        * Someone else is using the Referral Manager at the same time and has not generated/sent tokens yet.
+        * The CLI was terminated between the get `eligibleUsers` step and the generate/send tokens step.
+    
+        Either one of the cases, We will return an error and display `Referral token generation already in progress or canceled early. Do you want to continue anyway?`. If they decide to continue, we call `ClearUsers` to sets eligibleUsers to empty and the Referral Manager queries `users` table to get users with 0 remaining invites. 
+2. The Referral Manager returns the number of users to the CLI.
+3. The CLI receives the number of users and displays: `How many invitation do you want to generate for each user?`
+4. After the user enters the number of invites to generate, the CLI displays: `Generating X amount of invitation tokens for Y amount of users. Yes/No.`
+   - If the input is `Yes`, it will generate x amount of invitation tokens per user based on the input from CLI and save them into `tokens` table.
```
egonelbre: Is there a reason we aren't generating them on demand rather than upfront?

e.g. flow wise:
1. referral manager says that you have 3 tokens available,
2. user clicks "create link" or "send link to email",
3. satellite contacts referral manager which creates the token.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,236 @@
...
+3. Redeemed satellite receives the response, which:
+    - if it is a success, the satellite will proceed with the account creation and
+    then send a request to the Referral Manager, which: 
+        - updates the `tokens` table to store the newly created user ID in `redeemedID` column for that particular referral token.
+        - updates the `users` table to store the newly created user ID and the satellite URL the user belongs to.
+    - if it is an invalid token, the satellite will display a proper message in the UI.
+    
+## Rationale
+
+We will create a `inspector` port that's only listening on `localhost` so CLI could be only used by Referral Manager operators.
```
egonelbre: Inspector port doesn't sound quite right. Something like `admin`, `cli` might be better.

docs/blueprints/referral-manager-v1.md
VinozzZ: How would the CLI user know how many tokens they want to create per user if they don't know how many total `eligibleUsers` there are? Or do we just want them to input the total number of tokens they want to create? If that's the case, do we care to make the selected `eligibleUsers` to be evenly distributed across all satellites?

docs/blueprints/referral-manager-v1.md
VinozzZ: I think it's because we want to keep a set amount of available tokens we give out to users. If we generate tokens on demand, then we will need to have some sort of global counter to keep track of how many we have generated. Also satellite doesn't know anything about the tokens, so it would have to contact referral manager first to get the token count for a user anyway. If we generate it upfront, then there will be less round trip

docs/blueprints/referral-manager-v1.md
VinozzZ: As if right now, yes, it's happening on every request so the data is up-to-date. We probably should have a debounce. Maybe we can say, it should only make one request per hour?

docs/blueprints/referral-manager-v1.md
egonelbre: Not really, each user could have a "tokens available", then the issuing the tokens would be changing just that number.

egonelbre: > If we generate tokens on demand, then we will need to have some sort of global counter to keep track of how many we have generated.

Does this means we track the unused token indefinitely?

VinozzZ: Yea, the marketing team would like to be able to collect metrics on token usage.

mobyvb: Updated for generating tokens on demand.

docs/blueprints/referral-manager-v1.md
mobyvb: Fixed.

docs/blueprints/referral-manager-v1.md
mobyvb: Done.

docs/blueprints/referral-manager-v1.md
mobyvb: Done.

docs/blueprints/referral-manager-v1.md
mobyvb: Done.

docs/blueprints/referral-manager-v1.md
mobyvb: Fixed

docs/blueprints/referral-manager-v1.md
mobyvb: Done

docs/blueprints/referral-manager-v1.md
```diff
@@ -43,61 +43,47 @@ Redeemed Satellite
...
-        - If there are unredeemed tokens, the Referral Manager will respond to the satellite's request with the tokens.
-        - If no unredeemed tokens are found for this user ID and such user ID doesn't exist in the Referral Manager's `users` table, the Referral Manager will add an entry for this user ID and in the `users` table.
+    - The Referral Manager will attempt to generate and/or fetch unredeemed tokens in the `tokens` table for this user ID and satellite.
+        - If the user ID doesn't exist in the Referral Manager's `users` table, the Referral Manager will add an entry for this user ID and satellite in the `users` table and return an empty response to the satellite.
+        - If the `users` table contains a value larger than `0` for `new_tokens` for this user ID, the Referral Manager will generate that number of new tokens, add the new unredeemed tokens to the `tokens` table, and set the value of `new_tokens` for that user to `0` in the `users` table.
+        - If there are unredeemed tokens in the `tokens` table, the Referral Manager will respond to the satellite's request with the tokens. Otherwise, it will return an empty response.
```
VinozzZ: Shouldn't the RM also respond with the newly generated tokens as well?

docs/blueprints/referral-manager-v1.md
```diff
@@ -43,61 +43,47 @@ Redeemed Satellite
...
-2. The Referral Manager returns the number of users for each satellite to the CLI.
-3. The CLI receives the number of users and displays: `How many invitations do you want to generate for each user?`
-4. After the operator enters the number of invites per user to generate, the CLI displays: `Generating X invitation tokens for Y users on Z satellites. Yes/No.`
-   - If the input is `Yes`, it will generate the amount of invitation tokens per user based on the input from CLI and save them into `tokens` table.
-   - The CLI will call `ClearUsers` endpoint to set `eligibleUser` to be empty if the input is `No`.
+1. `referral-manager start --tokens-per-user=3 <satelliteURL1> <satelliteURL2> ...` initiates the invitation token generation process. The `--dry-run` flag can optionally be added to run the process without actually generating tokens. The dry run allows for estimating how many new tokens would be created by the command.
+2. The Referral Manager queries the `users` table for all users where `new_tokens < flags.tokens_per_user` and where the satellite matches one of the CLI-provided satellite URLs. These are the "eligible users".
```
VinozzZ: I think here we will need to also look into the `tokens` table since we only want to generate tokens for users whose unredeemed token is less than x amount. Maybe it would be something like `new_tokens + unredeemed_tokens < flags.eligible_user_token_count` ? I think `flags.tokens_per_user` is the input for the amount of new tokens we want to create for each `eligibleUser`

docs/blueprints/referral-manager-v1.md
```diff
@@ -160,77 +156,45 @@ func Redeem(ctx, token, userID, satelliteURL) error {
...
-    users := db.GetUsers()
-    referralManager.eligibleUsers = eligibleUsers
-
-    return eligibleUsersCounts, nil
-}
 
 // GenerateTokens generates tokens, saves those in the referral manager db
-GenerateTokens(tokensPerUser int) (int, error) {
-    eligibleUsers := referralManager.eligibleUsers
+GenerateTokens(tokensPerUser int, satelliteURLs []string, dryRun bool) (tokenCount, eligibleUserCount int, error) {
```
VinozzZ: We probably need another argument here for the `unredeemed token amount`

docs/blueprints/referral-manager-v1.md
```diff
@@ -160,77 +156,45 @@ func Redeem(ctx, token, userID, satelliteURL) error {
...
 
 ## Wrapup
 
 Team Green will be responsible implementing this blueprint.
 
-## Open issues
\ No newline at end of file
+## Open issues
+
+Should we have another CLI command for querying the global number of unredeemed tokens + ungenerated tokens?
```
VinozzZ: I think we would be fine without it since the data can be accessed through redash

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+The Token struct represents both a Go struct (on satellite and Referral Manager) as well as the schema for the `tokens` table on the Referral Manager
+
+```
+type Token struct {
+    Secret [32]byte
+    OwnerID uuid.UUID
+    RedeemedID uuid.UUID
+    OwnerSatelliteURL string
+    RedeemedSatelliteURL string
+    Status string
```
egonelbre: This probably should be some sort of enum.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+    fmt.Printf("Successfully created %d tokens for %d users.\n", tokenCount, userCount)
+    if flags.dryRun {
+        fmt.Println("This was a dry run. Run again without the --dry-run flag to actually generate tokens.")
+    }
+}
+```
+
+## Wrapup
+
+Team Green will be responsible implementing this blueprint.
```
egonelbre: Remove the `Team Green` comment since the responsibilities what people do, move around.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+        - If there are unredeemed tokens in the `tokens` table, the Referral Manager will respond to the satellite's request with the tokens after generating new tokens. Otherwise, it will return an empty response.
+    - The satellite receives the response from Referral Manager.
+        - If the response payload contains tokens, the satellite will display them in the UI.
+        - If it's an empty response payload, the satellite will display a message `No available referral links. Try again later.`
+
+
+_Referral Manager CLI_
+
+1. `referral-manager start --tokens-per-user=3 <satelliteURL1> <satelliteURL2> ...` initiates the invitation token generation process. The `--dry-run` flag can optionally be added to run the process without actually generating tokens. The dry run allows for estimating how many new tokens would be created by the command.
+2. The Referral Manager queries both `users` and `tokens` tables for all users where `new_tokens + unredeemed_tokens < flags.tokens_per_user` and where the satellite matches one of the CLI-provided satellite URLs. These are the "eligible users".
```
thepaul: You mentioned that we want to add another flag limiting the maximum number of unredeemed tokens per user, which would also come into play here.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+We will create an admin port that's only listening on `localhost` so CLI could be only used by Referral Manager operators.
+
+We could also set the admin port to only accept requests coming from storj vpn.
+
+## Implementation
+
+- Create a private repository for Referral Manager.
+- Create `tokens` and `users` table in Referral Manager database.
+- Create an admin endpoint on Referral Manager.
+    - Implementing a method `GenerateTokens` for upadting the `users` table on Referral Manager to contain new tokens for eligible users.
```
thepaul: upadting -> updating :D

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+    user, err := db.GetUser(userID, satelliteURL)
+    if err == UserNotFound {
+        db.CreateUser(userID, satelliteURL)
+        return nil
+    }
+    if user.NewTokens > 0 {
+        for i:=0; i<user.NewTokens; i++ {
+            newToken := Token{data: generateRandomToken(), user: user.ID, satellite: user.satelliteURL}
+            db.CreateToken(newToken)
+            tokens = append(tokens, newToken)
```
thepaul: This line can be taken out.

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+        }
+        db.UpdateUserNewTokens(userID, satelliteURL, 0)
```
thepaul: This should be in a transaction with the fetching part, above (line 122)

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+    }
+    tokens := db.GetTokensByUserIDAndSatelliteURL(userID, satelliteURL)
+
+    return tokens
+}
+
+// Redeem marks a token as redeemed and stores user info into database
+func Redeem(ctx, token, userID, satelliteURL) error {
+	// only update the status if the status of a token is unredeems
```
thepaul: unredeems -> Unredeemed, I think

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+	tokenStatus, err := db.RedeemToken(ctx, token, userID)
+	if err != nil {
+		return err
+	}
+
+    // decrease unredeemed token count in users table by 1
+    err := db.UpdateUserInfo(ctx, user)
```
thepaul: This will also need to be in a transaction

docs/blueprints/referral-manager-v1.md
```diff
@@ -0,0 +1,206 @@
...
+    if err != nil {
+        return err
+    }
+
+    // save user info into users table
+    err := db.CreateUser(ctx, userID, satelliteURL)
```
thepaul: If we do this in a transaction with the other parts, then we would get an error if the userID already exists and the above changes would be rolled back. Just in case something goes wrong somewhere.

PR #3040, opened 2019-09-12
uplink/storage/streams: Add test for interrupted deletes

isaachess:
What: Add a test to ensure that when an uplink/storage/streams.Delete call is interrupted, subsequent calls complete successfully. If the first call is interrupted it may delete some segments but not all. In subsequent calls we want to make sure it still works.

Why: I have a ticket for this, but discovered it was already fixed by @mniewrzal. I simply confirmed it locally and wrote a unit-test to ensure it will work in the future.

I also made a simple refactor of the test file to make sure I didn't have to copy/paste code.

Please describe the tests:
 - Test 1: The entire PR is a test.
 

isaachess:
@simongui I think that is unnecessary. Because the test exists within an isolated test environment, and only one item is uploaded, there is no chance for it to return anything _other_ than the one item that was uploaded.

PR #3045, opened 2019-09-13
satellite/console: GraphQL input length limitation.

Barterio:
What: 
Implemented graphql input length limitation 
Why:
To prevent bad people to break satellite by sending huge requests which will break console.

Please describe the tests:
 - Test 1:
 - Test 2:
 

Barterio:
Need advice about better error message

satellite/console/consoleweb/utils.go
```diff
@@ -48,8 +51,13 @@ func getToken(req *http.Request) string {
 	return value[len(authorizationBearer):]
 }
 
+
 // getQuery retrieves graphql query from request
 func getQuery(req *http.Request) (query graphqlJSON, err error) {
+	if req.ContentLength > ContentLengthLimit {
```
egonelbre: ContentLength header cannot be trusted.

Barterio: Good point. Replaced with ``http.MaxBytesReader`` where possible

satellite/console/consoleweb/utils.go
```diff
@@ -49,20 +52,22 @@ func getToken(req *http.Request) string {
...
+		return queryPOST(writer, req)
 	default:
 		return query, errs.New("wrong http request type")
 	}
 }
 
 // queryPOST retrieves graphql query from POST request
-func queryPOST(req *http.Request) (query graphqlJSON, err error) {
+func queryPOST(writer http.ResponseWriter, req *http.Request) (query graphqlJSON, err error) {
+	req.Body = http.MaxBytesReader(writer, req.Body, ContentLengthLimit)
```
rikysya: I guess `http.Request` may be required to be immutable. 
Anyway, if you are modifying `*http.Request` you can skip the propagation and do it in the handler itself. 
We can also modify code to use only `io.Reader` and `URL` object

satellite/console/consoleweb/utils.go
```diff
@@ -15,6 +15,9 @@ import (
 	"storj.io/storj/satellite/console/consoleweb/consoleql"
 )
 
+// ContentLengthLimit describes 4KB limit
+const ContentLengthLimit = 1 << 12
```
egonelbre: `4 * memory.KB` is a little bit easier to read

satellite/console/consoleweb/utils.go
Barterio: https://golang.org/pkg/net/http/#Handler
``Except for reading the body, handlers should not modify the provided Request.``
This is the case when I'm reading the body.
Everything else left unchanged.

egonelbre: `req.Body = ...` changes the request.

rikysya: ok

satellite/console/consoleweb/utils.go
```diff
@@ -63,6 +67,7 @@ func getQuery(req *http.Request) (query graphqlJSON, err error) {
 
 // queryPOST retrieves graphql query from POST request
 func queryPOST(req *http.Request) (query graphqlJSON, err error) {
+
```
rikysya: I guess this extra new line is not needed

Barterio: yes

satellite/console/consoleweb/server.go
```diff
@@ -392,6 +393,8 @@ func (server *Server) grapqlHandler(w http.ResponseWriter, r *http.Request) {
 	defer mon.Task()(&ctx)(nil)
 	w.Header().Set(contentType, applicationJSON)
 
+	r.Body = http.MaxBytesReader(w, r.Body, ContentLengthLimit.Int64())
```
egonelbre: This is changing the request.

satellite/console/consoleweb/utils.go
Barterio: This is refers to reading the body.
According to `http` documentation, prohibited any other request changing except for body reading.

egonelbre:
```
xyz := http.MaxBytesReader(writer, req.Body, ContentLengthLimit), <-- this part is fine, this doesn't change the request
req.Body = xyz <-- this is not, because it's changing the request

File: satellite/console/consoleweb/utils.go [hunk 6]
```diff
@@ -53,23 +53,25 @@ func getToken(req *http.Request) string {
...
 	}
 }
 
 // queryPOST retrieves graphql query from POST request
-func queryPOST(req *http.Request) (query graphqlJSON, err error) {
+func queryPOST(w http.ResponseWriter, req *http.Request) (query graphqlJSON, err error) {
 	switch typ := req.Header.Get(contentType); typ {
 	case applicationGraphql:
-		body, err := ioutil.ReadAll(req.Body)
+		limitedReader := http.MaxBytesReader(w, req.Body, ContentLengthLimit.Int64())
```
egonelbre: This wouldn't limit `json` queries and call close on `limitedReader` rather than `req.Body`.

Barterio: good point. missed that

satellite/console/consoleweb/utils.go
```diff
@@ -67,16 +67,15 @@ func getQuery(w http.ResponseWriter, req *http.Request) (query graphqlJSON, err
 
 // queryPOST retrieves graphql query from POST request
 func queryPOST(w http.ResponseWriter, req *http.Request) (query graphqlJSON, err error) {
+	limitedReader := http.MaxBytesReader(w, req.Body, ContentLengthLimit.Int64())
```
egonelbre: This isn't being closed when the content-type is some different. Use `defer` to close things, it helps to avoid these bugs.

egonelbre: Ah nevermind, actually with HTTP handler, the Server automatically closes the Body. But would be still nice to use defer here.

PR #3047, opened 2019-09-13
satellite/console/server: fix usage rollup report

rikysya:
What: satellite/console/server: fix usage rollup report

Why: to fix it

Please describe the tests:
 - Test 1:
 - Test 2:
 

satellite/console/consoleweb/server.go
```diff
@@ -193,10 +193,16 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 	var projectID *uuid.UUID
 	var since, before time.Time
 
-	tokenCookie, err := r.Cookie("tokenKey")
+	host, _, err := net.SplitHostPort(r.Host)
 	if err != nil {
 		server.log.Error("bucket usage report error", zap.Error(err))
+		server.serveError(w, r, http.StatusNotFound)
```
ifraixedes: This looks more to me more a `StatusInternalServerError`

satellite/console/consoleweb/server.go
```diff
@@ -193,10 +193,16 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
...
+		return
+	}
 
+	tokenCookie, err := r.Cookie(host + "_tokenKey")
+	if err != nil {
+		server.log.Error("bucket usage report error", zap.Error(err))
```
ifraixedes: We don't need to log as error the unauthorized request.
I see others below, but they should not do so.

satellite/console/consoleweb/server.go
```diff
@@ -214,9 +219,7 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 	defer func() {
 		if err != nil {
 			server.log.Error("bucket usage report error", zap.Error(err))
```
ifraixedes: Because you have modified the line below, it's worth to remove this log call. `StatusNotFound` isn't an error in our system so we don't have to log it.

satellite/console/consoleweb/server.go
```diff
@@ -249,8 +252,7 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 	}
 
 	if err = server.templates.usageReport.Execute(w, bucketRollups); err != nil {
-		server.log.Error("satellite/console/server: usage report template could not be executed", zap.Error(err))
-		server.serveError(w, r, http.StatusNotFound)
+		err = errs.New("usage report template could not be executed: %v", err)
```
ifraixedes: We should log this error and return `StatusInternalServerError`

satellite/console/consoleweb/server.go
rikysya: there is a `TODO` regarding the issue

satellite/console/consoleweb/server.go
rikysya: it is logged in defer

satellite/console/consoleweb/server.go
ifraixedes: I've seen it after, however, I've seen for unauthenticated, not for this one, so I'm not sure if that's apply considering that you haven't added the `TODO`  , though you may have forgotten it.

satellite/console/consoleweb/server.go
egonelbre: `server.templates.usageReport.Execute` writes the header already so the status cannot be changed.

ifraixedes: Oh, I see.
It's very confusing with the `StatusNotFound`.
Can you add a `TODO` in the `defer` mentioning what the correct status code should be?

rikysya: `template.Execute` accepts `io.Writer`, so how can it modify the header of the request? Or you are talking about smth else?

egonelbre: See documentation in https://golang.org/pkg/net/http/#ResponseWriter

ifraixedes: You cannot if you pass the response writer, which is good for no buffering.

Mostly, the only thing that we can do here is logging the error and assigned to `nil`.

satellite/console/consoleweb/server.go
```diff
@@ -192,33 +192,32 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
...
+		if err != nil {
+			// TODO: use http.StatusUnauthorized status when appropriate page will be created
+			server.serveError(w, r, http.StatusNotFound)
+		}
+	}()
+
+	host, _, err := net.SplitHostPort(r.Host)
 	if err != nil {
 		server.log.Error("bucket usage report error", zap.Error(err))
+		return
```
ifraixedes: Why this error will turn in to a `StatusUnauthorized`?

satellite/console/consoleweb/server.go
```diff
@@ -192,33 +192,32 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 
 	var projectID *uuid.UUID
 	var since, before time.Time
+	var bucketRollups []console.BucketUsageRollup
```
egonelbre: You can move these where they are loaded/created rather than at the top.

satellite/console/consoleweb/server.go
```diff
@@ -192,33 +192,32 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
...
 
-	tokenCookie, err := r.Cookie("tokenKey")
+	defer func() {
+		if err != nil {
+			// TODO: use http.StatusUnauthorized status when appropriate page will be created
+			server.serveError(w, r, http.StatusNotFound)
```
egonelbre: Note, use the appropriate errors codes. Using defer will make it more complicated.

satellite/console/consoleweb/server.go
```diff
@@ -242,16 +241,14 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 		zap.Stringer("since", since),
 		zap.Stringer("before", before))
 
-	ctx = console.WithAuth(ctx, auth)
-	bucketRollups, err := server.service.GetBucketUsageRollups(ctx, *projectID, since, before)
+	bucketRollups, err = server.service.GetBucketUsageRollups(ctx, *projectID, since, before)
 	if err != nil {
+		server.log.Error("bucket usage report error", zap.Error(err))
```
ifraixedes: I doubt that any error returned by `GetBucketUsageRollups` will be a `StatusUnauthorized`.

satellite/console/consoleweb/server.go
```diff
@@ -242,16 +241,14 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
...
 		return
 	}
 
-	if err = server.templates.usageReport.Execute(w, bucketRollups); err != nil {
-		server.log.Error("satellite/console/server: usage report template could not be executed", zap.Error(err))
-		server.serveError(w, r, http.StatusNotFound)
-		return
+	if terr := server.templates.usageReport.Execute(w, bucketRollups); terr != nil {
+		server.log.Error("bucket usage report error", zap.Error(terr))
```
ifraixedes:
```suggestion
		server.log.Error("bucket usage report template execution", zap.Error(terr))
```

satellite/console/consoleweb/server.go
rikysya: ?

satellite/console/consoleweb/server.go
rikysya: It could be

rikysya: And in `defer` it was `StatusNotFound`, and a note that right now we have only error page

satellite/console/consoleweb/server.go
```diff
@@ -190,68 +190,61 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
...
 	defer mon.Task()(&ctx)(&err)
 
-	var projectID *uuid.UUID
-	var since, before time.Time
-
-	tokenCookie, err := r.Cookie("tokenKey")
+	host, _, err := net.SplitHostPort(r.Host)
 	if err != nil {
 		server.log.Error("bucket usage report error", zap.Error(err))
+		server.serveError(w, r, http.StatusInternalServerError)
```
egonelbre: This is not an internal error, this would be an invalid request.

satellite/console/consoleweb/server.go
```diff
@@ -190,68 +190,61 @@ func (server *Server) bucketUsageReportHandler(w http.ResponseWriter, r *http.Re
 	var err error
 	defer mon.Task()(&ctx)(&err)
 
-	var projectID *uuid.UUID
-	var since, before time.Time
-
-	tokenCookie, err := r.Cookie("tokenKey")
+	host, _, err := net.SplitHostPort(r.Host)
 	if err != nil {
 		server.log.Error("bucket usage report error", zap.Error(err))
```
egonelbre: Lower the level here to `Debug` since this is a caller error, not the server error.

rikysya: Well, if the parsed `request` has invalid `host`, but we reach this handler than I guess server might work incorrectly

egonelbre: I'm not sure whether the http.Server verifies whether the `Host` field is a valid `HostPort` or not.

rikysya: And that is not `request.URL.Host`, which comes from the client

egonelbre:
```
// For server requests, Host specifies the host on which the URL
// is sought. Per RFC 7230, section 5.4, this is either the value
// of the "Host" header or the host name given in the URL itself.
// It may be of the form "host:port". 
```

rikysya: So my idea is that if `host` is incorrect(in come way), than `server` has some errors parsing the `request` so it's `InternalServerError`

If it is not valid `HostPort`, than we parse this field incorrectly, which is `InternalServerError` as well

egonelbre: Ok, yeah, it looks like it ensures that it's a valid host:

```
	// Find the target host. Prefer the Host: header, but if that
	// is not given, use the host from the request URL.
	//
	// Clean the host, in case it arrives with unexpected stuff in it.
	host := cleanHost(r.Host)
	if host == "" {
		if r.URL == nil {
			return errMissingHost
		}
		host = cleanHost(r.URL.Host)
	}
```

satellite/console/consoleweb/server.go
ifraixedes: This handled by the `defer` which currently is `NotFound` but it will be changed by `Unauthorized` when the page is created.

Isn't this an `InternalServerError`.

Although all who create the page and do refactoring may have to go and check each return I think that it's quite easy to forget to do so and only update the `defer` states which is where the `TODO` comment is placed.

satellite/console/consoleweb/server.go
ifraixedes: I know that there is one page, but I think that the logic of distinguishing errors should be implemented and returning now just the `NotFound` because there is only one page.

PR #3049, opened 2019-09-13
satellite/satellitedb: Initial GE Satellite DB Implementation

ethanadams:
What: 
Add basic CRUD operations for graceful_exit_progress and graceful_exit_transfer_queue tables.

Why:
Need for managing graceful exit transfer progress.

https://storjlabs.atlassian.net/browse/V3-2592
https://github.com/storj/storj/blob/master/docs/blueprints/storagenode-graceful-exit/overview.md#satellite-database-changes

Please describe the tests:
 - Test 1: TestProgress: Tests add, update, get all, and delete
 - Test 2: TestTransferQueueItem: Tests add, update, get all by node ID with limit, and delete
 

satellite/satellitedb/migrate.go
```diff
@@ -1211,26 +1211,26 @@ func (db *DB) PostgresMigration() *migrate.Migration {
 				Description: "Add Graceful Exit tables and update nodes table",
 				Version:     56,
 				Action: migrate.SQL{
-					`ALTER TABLE nodes ADD COLUMN exit_loop_completed_at timestamp with time zone;`,
-					`ALTER TABLE nodes ADD COLUMN exit_initiated_at timestamp with time zone;`,
-					`ALTER TABLE nodes ADD COLUMN exit_finished_at timestamp with time zone;`,
+					`ALTER TABLE nodes ADD COLUMN exit_loop_completed_at timestamp;`,
```
egonelbre: This is changing migrations. Has it been rolled out or not?

ethanadams: Great point.  Looks like the original schema commit is in the v0.21.0 branch going out Wednesday.  I'll fix the migrations.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,61 @@
...
+	// DeleteProgress deletes a graceful exit progress entry in the database
+	DeleteProgress(ctx context.Context, nodeID storj.NodeID) error
+	// IncrementProgressBytesTransferred increments bytes transferred value
+	IncrementProgressBytesTransferred(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
+	// GetProgress gets a graceful exit progress entry in the database
+	GetProgress(ctx context.Context, nodeID storj.NodeID) (*Progress, error)
+	// GetAllProgress gets all graceful exit progress entries in the database
+	GetAllProgress(ctx context.Context) ([]*Progress, error)
+
+	// CreateProgress creates a graceful exit progress entry in the database
```
navillasa: I think should be `// CreateTransferQueueItem ...`

satellite/satellitedb/gracefulexit.go
```diff
@@ -0,0 +1,232 @@
...
+		NodeID:           nID,
+		BytesTransferred: dbxProgress.BytesTransferred,
+		UpdatedAt:        dbxProgress.UpdatedAt,
+	}
+
+	return progress, err
+}
+
+// GetAllProgress gets all graceful exit progress entries in the database.
+func (db *gracefulexitDB) GetAllProgress(ctx context.Context) (_ []*gracefulexit.Progress, err error) {
```
navillasa: in what scenarios will we be using GetAllProgress? It seems like this could be an expensive operation.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,141 @@
...
+		item.DurabilityRatio = 1.2
+		item.RequestedAt = time.Now()
+
+		err = geDB.UpdateTransferQueueItem(ctx, *item)
+		require.NoError(t, err)
+
+		latestItem, err := geDB.GetTransferQueueItem(ctx, nodeID, path)
+		require.NoError(t, err)
+		require.Equal(t, item.DurabilityRatio, latestItem.DurabilityRatio)
+		require.True(t, item.RequestedAt.Truncate(time.Millisecond).Equal(latestItem.RequestedAt.Truncate(time.Millisecond)))
```
navillasa: it seems like truncating the requested_at times to millisecond might make all of the rows have the same values, so the comparison could be true even if it's getting the wrong row. does it have to be milliseconds because of SQLite or could we try microseconds?

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	LastFailedAt    time.Time
+	LastFailedCode  int
+	FailedCount     int
+	FinishedAt      time.Time
+}
+
+// DB implements CRUD operations for graceful exit service
+//
+// architecture: Database
+type DB interface {
```
egonelbre: Separate These into two separate databases.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	// CreateProgress creates a graceful exit progress entry in the database.
+	CreateProgress(ctx context.Context, nodeID storj.NodeID) error
+	// UpdateProgress updates a graceful exit progress entry in the database.
+	UpdateProgress(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
+	// DeleteProgress deletes a graceful exit progress entry in the database.
+	DeleteProgress(ctx context.Context, nodeID storj.NodeID) error
```
egonelbre: When would we need to delete progress entries? AFAIU, we'll always need them to list exited nodes.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	// IncrementProgressBytesTransferred increments bytes transferred value.
+	IncrementProgressBytesTransferred(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
```
egonelbre: Why is there a need for both `UpdateProgress` and `IncrementProgressBytesTransferred`?

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	// GetProgress gets a graceful exit progress entry in the database.
+	GetProgress(ctx context.Context, nodeID storj.NodeID) (*Progress, error)
```
egonelbre: What would call this?

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// CreateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	CreateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
```
egonelbre: Rename to `Enqueue`, which should be sufficient, also this should be batched.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	// UpdateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	UpdateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
+	// DeleteTransferQueueItem deletes a graceful exit transfer queue entry in the database.
+	DeleteTransferQueueItem(ctx context.Context, nodeID storj.NodeID, path []byte) error
+	// GetTransferQueueItem gets a graceful exit transfer queue entry in the database.
+	GetTransferQueueItem(ctx context.Context, nodeID storj.NodeID, path []byte) (*TransferQueueItem, error)
```
egonelbre: I'm not following how it would be used.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	// GetIncompleteTransferQueueItemsByNodeIDWithLimits gets incomplete graceful exit transfer queue entries in the database ordered by the queued date ascending.
+	GetIncompleteTransferQueueItemsByNodeIDWithLimits(ctx context.Context, nodeID storj.NodeID, limit int, offset int64) ([]*TransferQueueItem, error)
```
egonelbre: The `TransferQueue` should only contain incomplete entries. So a method `ListIncomplete(ctx context.Context, nodeID storj.NodeID, limit int, offset int64)` would be a sufficient name.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// DB implements CRUD operations for graceful exit service
+//
+// architecture: Database
+type DB interface {
+	// CreateProgress creates a graceful exit progress entry in the database.
+	CreateProgress(ctx context.Context, nodeID storj.NodeID) error
+	// UpdateProgress updates a graceful exit progress entry in the database.
+	UpdateProgress(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
```
egonelbre: Do we want to update anything else besides failure count?

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	NodeID          storj.NodeID
+	Path            []byte
+	PieceNum        int
+	DurabilityRatio float64
+	QueuedAt        time.Time
+	RequestedAt     time.Time
+	LastFailedAt    time.Time
+	LastFailedCode  int
+	FailedCount     int
+	FinishedAt      time.Time
```
egonelbre: I'm not sure whether there's value in keeping the finished items in the queue.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	IncrementProgressBytesTransferred(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
+	// GetProgress gets a graceful exit progress entry in the database.
+	GetProgress(ctx context.Context, nodeID storj.NodeID) (*Progress, error)
+
+	// CreateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	CreateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
+	// UpdateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	UpdateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
+	// DeleteTransferQueueItem deletes a graceful exit transfer queue entry in the database.
+	DeleteTransferQueueItem(ctx context.Context, nodeID storj.NodeID, path []byte) error
```
egonelbre: This should be batched.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	DeleteProgress(ctx context.Context, nodeID storj.NodeID) error
+	// IncrementProgressBytesTransferred increments bytes transferred value.
+	IncrementProgressBytesTransferred(ctx context.Context, nodeID storj.NodeID, bytesTransferred int64) error
+	// GetProgress gets a graceful exit progress entry in the database.
+	GetProgress(ctx context.Context, nodeID storj.NodeID) (*Progress, error)
+
+	// CreateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	CreateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
+	// UpdateTransferQueueItem creates a graceful exit transfer queue entry in the database.
+	UpdateTransferQueueItem(ctx context.Context, item TransferQueueItem) error
```
egonelbre: This probably should be batched as well.

satellite/gracefulexit/db.go
ethanadams: Removed

satellite/gracefulexit/db.go
ethanadams: Removed Update, reworked Increment

satellite/gracefulexit/db.go
ethanadams: Tests

satellite/gracefulexit/db.go
ethanadams: Updated

satellite/gracefulexit/db.go
ethanadams: As discussed, added failed and success counts to the progress table so we can delete the processed queue items.

satellite/gracefulexit/db.go
ethanadams: Removed

satellite/gracefulexit/db.go
ethanadams: Added delete finished

satellite/gracefulexit/db.go
ethanadams: Added delete by nodeID and delete transferred. I'm going to hold off on batch until I see a need.

satellite/gracefulexit/db.go
ethanadams: Will change to batch in a future PR

satellite/satellitedb/utils.go
```diff
@@ -28,29 +28,54 @@ type postgresNodeIDList storj.NodeIDList
...
+		return []byte("{}")
+	}
+
+	var total int
+	for _, path := range values {
+		total += len(path)
+	}
 	var wp, x int
-	out := make([]byte, 2+len(nodes)*(6+storj.NodeIDSize*2)-1)
+	out := make([]byte, 2+(6*len(values))+(total*2)-1)
```
mobyvb: what does this math mean? I don't really understand what it was doing originally or what it is doing now. Could we add a comment explaining it?

satellite/satellitedb/utils_test.go
```diff
@@ -47,3 +47,18 @@ func TestPostgresNodeIDsArray(t *testing.T) {
 
 	assert.Equal(t, expected.(string), string(got.([]byte)))
 }
+
+func TestPostgresPathsArray(t *testing.T) {
+	ids := make([][]byte, 10)
+	for i := range ids {
+		ids[i] = testrand.BytesInt(testrand.Intn(1000))
```
mobyvb: Do you think this is better than using `testrand.NodeID().Bytes()`?

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+	FailedCount     int
+	FinishedAt      time.Time
+}
+
+// DB implements CRUD operations for graceful exit service
+//
+// architecture: Database
+type DB interface {
+	// IncrementProgress increments transfer stats for a node.
+	IncrementProgress(ctx context.Context, nodeID storj.NodeID, bytes int64, transferred int64, failed int64) error
```
mobyvb: nit - I think `UpdateProgress` is a better name for this. I also think `successfulTransfers` and `failedTransfers` are preferable names to `transferred` and `failed`.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,151 @@
...
+		item, err := geDB.GetTransferQueueItem(ctx, nodeID1, path1)
+		require.NoError(t, err)
+
+		item.FinishedAt = time.Now()
+		err = geDB.UpdateTransferQueueItem(ctx, *item)
+		require.NoError(t, err)
+
+		queueItems, err := geDB.GetIncomplete(ctx, nodeID1, 10, 0)
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queueItems))
```
mobyvb: nit - `require.Len(t, queueItems, 1)` should also work

mobyvb: It might also be worth testing `GetIncomplete` before updating with `FinishedAt`, and making sure the length of the items returned is 2.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+			// test delete
+			err := geDB.DeleteTransferQueueItem(ctx, queueItem.NodeID, queueItem.Path)
+			require.NoError(t, err)
+		}
+		// test delete finished
+		err = geDB.DeleteFinishedTransferQueueItems(ctx, nodeID1)
+		require.NoError(t, err)
+
+		_, err = geDB.GetTransferQueueItem(ctx, nodeID1, path1)
```
mobyvb: I would also test this `GetTransferQueueItem` before the call to `DeleteFinishedTransferQueueItems` to make sure that there was in fact an item before the delete function was called.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,151 @@
...
+		require.Error(t, err)
+
+		// test delete all for a node
+		err = geDB.DeleteTransferQueueItems(ctx, nodeID2)
+		require.NoError(t, err)
+
+		queueItems, err = geDB.GetIncomplete(ctx, nodeID1, 10, 0)
```
mobyvb: Check `nodeID2` as well, otherwise we don't know if the last delete works.

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,61 @@
...
+package gracefulexit
+
+import (
+	"context"
+	"time"
+
+	"storj.io/storj/pkg/storj"
+)
+
+// Progress represents the persisted graceful exit progress record
```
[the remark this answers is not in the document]
ethanadams: Updated

satellite/gracefulexit/db.go
```diff
@@ -0,0 +1,59 @@
...
+type Progress struct {
+	NodeID            storj.NodeID
+	BytesTransferred  int64
+	PiecesTransferred int64
+	PiecesFailed      int64
+	UpdatedAt         time.Time
+}
+
+// TransferQueueItem represents the persisted graceful exit queue record
+type TransferQueueItem struct {
```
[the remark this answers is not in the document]
ethanadams: Will leave as is for now

satellite/satellitedb/utils.go
ethanadams: Reverted the util changes. I missed pg.ByteaArray function

satellite/satellitedb/utils_test.go
ethanadams: Reverted

satellite/gracefulexit/db.go
ethanadams: My concern with UpdateProgress is that it sounds like you are setting the values rather than incrementing.
Changed transferred and failed

satellite/satellitedb/gracefulexit_test.go
ethanadams: Updated

satellite/satellitedb/gracefulexit_test.go
ethanadams: Updated

satellite/satellitedb/gracefulexit_test.go
ethanadams: I actually intended for that line to check nodeID2. Updated.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+			_, err = geDB.GetTransferQueueItem(ctx, queueItem.NodeID, queueItem.Path)
+			require.NoError(t, err)
+
+			// test delete
+			err = geDB.DeleteTransferQueueItem(ctx, queueItem.NodeID, queueItem.Path)
+			require.NoError(t, err)
+		}
+		// test delete finished
+		err = geDB.DeleteFinishedTransferQueueItems(ctx, nodeID1)
```
mobyvb: Is this deleting something that was not deleted on line 140 (`geDB.DeleteTransferQueueItem`)? I am worried that it is not actually testing anything if the item in question was already deleted

ethanadams: You are correct. Made some changes and added scoping to be more clear.

satellite/satellitedb/gracefulexit.go
```diff
@@ -0,0 +1,234 @@
...
+		dbx.GracefulExitTransferQueue_Path(item.Path),
+		update,
+	)
+}
+
+// DeleteTransferQueueItem deletes a graceful exit transfer queue entry.
+func (db *gracefulexitDB) DeleteTransferQueueItem(ctx context.Context, nodeID storj.NodeID, path []byte) (err error) {
+	defer mon.Task()(&ctx)(&err)
+	_, err = db.db.Delete_GracefulExitTransferQueue_By_NodeId_And_Path(ctx, dbx.GracefulExitTransferQueue_NodeId(nodeID.Bytes()), dbx.GracefulExitTransferQueue_Path(path))
+	return err
```
egonelbre: Let's wrap all the errors.

satellite/satellitedb/gracefulexit.go
```diff
@@ -0,0 +1,234 @@
...
+		item.RequestedAt = *dbxTransferQueue.RequestedAt
+	}
+	if dbxTransferQueue.LastFailedAt != nil && !dbxTransferQueue.LastFailedAt.IsZero() {
+		item.LastFailedAt = *dbxTransferQueue.LastFailedAt
+	}
+	if dbxTransferQueue.FinishedAt != nil && !dbxTransferQueue.FinishedAt.IsZero() {
+		item.FinishedAt = *dbxTransferQueue.FinishedAt
+	}
+
+	return item, Error.Wrap(err)
```
egonelbre: Here you can `return item, nil`

satellite/satellitedb/gracefulexit.go
```diff
@@ -0,0 +1,234 @@
...
+	var transferQueueItemRows = make([]*gracefulexit.TransferQueueItem, len(dbxTransferQueueItemRows))
+	for i, dbxTransferQueue := range dbxTransferQueueItemRows {
+		transferQueueItem, err := dbxToTransferQueueItem(dbxTransferQueue)
+		if err != nil {
+			return nil, Error.Wrap(err)
+		}
+		transferQueueItemRows[i] = transferQueueItem
+	}
+
+	return transferQueueItemRows, Error.Wrap(err)
```
egonelbre: Here you can `return item, nil`

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,168 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package satellitedb_test
```
egonelbre: The `satellitedbtest` usually test the interface not the particular implementation. For example if we had a completely separate implementation not using "dbx", then these tests wouldn't apply to it. This would mean that it's probably more worthwhile to keep them near the interface rather than the implementation.

satellite/satellitedb/gracefulexit_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/satellite"
+	"storj.io/storj/satellite/gracefulexit"
+	"storj.io/storj/satellite/satellitedb/satellitedbtest"
+)
+
+func TestProgress(t *testing.T) {
+	// test basic graceful exit progress crud
+	satellitedbtest.Run(t, func(t *testing.T, db satellite.DB) {
+		ctx := context.Background()
```
egonelbre: `testcontext` also tries to ensure that the tests don't deadlock or run for too long.

satellite/satellitedb/gracefulexit.go
ethanadams: Updated

satellite/satellitedb/gracefulexit.go
ethanadams: Updated

satellite/satellitedb/gracefulexit.go
ethanadams: Moved

satellite/satellitedb/gracefulexit_test.go
ethanadams: Moved

satellite/satellitedb/gracefulexit_test.go
ethanadams: Updated

PR #3057, opened 2019-09-16
storagenode/storagenodedb: refactor both data access objects and migrations to support multiple DB connections

simongui:
**What:** This PR is stage 2 of a refactoring to position the storage node data access code to accept multiple database connections for a future PR that splits the storage node sqlite3 tables logically split database files.

This PR has 2 major changes.
1. Make modifications to open and close multiple databases but for now only opens and closes 1 (future PR adds more).
1. Adds the support for looking up SQL databases by database name key in a map.
1. Modifies the migration infrastructure to support multiple database connections per migration step to position the code base for future migrations across multiple database connections.
1. Modifies the migration test and snapshot infrastructure to support passing in multiple database connections.
1. Makes storagenode tests use on-disk databases rather than in-memory.

The way the `object.Configure(SQLDB)` is being used may seem a bit odd but there's a reason why it was done this way. There's some outside consumers that are taking a reference to the business object which means even if we change the reference of database.object to a new instance outside consumers are referencing a closed database connection that is now obsolete. This PR is already huge in refactoring so this was a clean approach given the current code. In the future we should probably refactor how the data access objects are designed so it's more of a tell don't ask pattern because ask oriented patterns are brittle and that caused me friction and really strange bugs when I made what should be considered internal changes. The objects appeared to have good encapsulation but didn't.

For now each data access object still uses `info.db` via this code but this now supports passing in multiple.

```
db.legacyInfoDB.Configure(legacyInfoDB)
db.bandwidthDB.Configure(legacyInfoDB)
db.ordersDB.Configure(legacyInfoDB)
db.pieceExpirationDB.Configure(legacyInfoDB)
db.v0PieceInfoDB.Configure(legacyInfoDB)
db.pieceSpaceUsedDB.Configure(legacyInfoDB)
db.reputationDB.Configure(legacyInfoDB)
db.storageUsageDB.Configure(legacyInfoDB)
db.usedSerialsDB.Configure(legacyInfoDB)
db.satellitesDB.Configure(legacyInfoDB)
```

Why: A future PR will include splitting the storage node SQLite databases into multiple database files and all of these changes were required to allow the data access objects to support that.

Please describe the tests:
 - Test 1:
 - Test 2:
 

storagenode/storagenodedb/legacyinfo.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Configure sets the underlining SQLDB connection.
+func (db *legacyInfoDB) Configure(sqlDB SQLDB) {
+	db.SQLDB = sqlDB
+}
+
+// Rebind rebind parameters
+// These are implemented because the migrate.DB interface requires them.
+// Maybe in the future we should untangle those.
+func (db *legacyInfoDB) Rebind(s string) string { return s }
```
egonelbre: If you add `Rebind` and `Schema` to `SQLDB` then you won't need to reimplement it for all of them. You should also be able to handle `Configure` such that there is no repetition.

egonelbre: Alternatively there could be a separate type that implements these methods and all other databases would embed it.

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +116,80 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
 
-	versionsPath := ""
-	versionsDB, err := openTestDatabase()
+// openDatabases opens all the SQLite3 storage node databases and returns if any fails to open successfully.
+func (db *DB) openDatabases() error {
+	// These objeccts have a Configure method to allow setting the underlining SQLDB connection
+	// that each uses internally to do data access to the SQLite3 databases.
+	// The reason it was done this way was because there's some outside consumers that are
+	// taking a reference to the business object.
+	legacyInfoDB, err := db.openDatabase(LegacyInfoDBName)
```
egonelbre: Why is it called `legacy info db`? As far as I understand we are not entirely getting rid of it, so calling it legacy doesn't seem to fit.

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +116,80 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
 	}
-	pieces := filestore.New(log, piecesDir)
+	return db, nil
+}
 
-	versionsPath := ""
-	versionsDB, err := openTestDatabase()
+// openDatabases opens all the SQLite3 storage node databases and returns if any fails to open successfully.
+func (db *DB) openDatabases() error {
+	// These objeccts have a Configure method to allow setting the underlining SQLDB connection
```
egonelbre: typo `objeccts`

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +116,80 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
-	db.SetMaxOpenConns(1)
-	db.SetConnMaxLifetime(-1)
+	db.sqlDatabases[dbName] = sqlDB
+	dbutil.Configure(sqlDB, mon)
 
-	mon.Chain("db_stats", monkit.StatSourceFunc(
-		func(cb func(name string, val float64)) {
-			monkit.StatSourceFromStruct(db.Stats()).Stats(cb)
-		}))
+	db.log.Sugar().Debugf("opened database %s", dbName)
```
egonelbre: prefer not to use sugar since that output cannot be easily parsed by tools

storagenode/storagenodedb/database.go
simongui: I'm pretty sure we have pretty much deprecated it's use? We had a discussion about renaming this because the work @isaachess did makes migrations a per-database concept now and makes this obsolete unless I'm misunderstanding something?

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +205,35 @@ func (db *DB) Close() error {
...
+}
+
+// closeDatabase closes the specified SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabase(dbName string) (err error) {
+	if conn, ok := db.sqlDatabases[dbName]; ok {
+		err = errs.Combine(err, conn.Close())
+		delete(db.sqlDatabases, dbName)
+	}
+	if err == nil {
+		db.log.Sugar().Debugf("closed database %s", dbName)
```
egonelbre: I'm not sure how useful this log message is.

storagenode/storagenodedb/database.go
isaachess: The database will exist, but will not have any data in it once all the migrations are complete. That's per the migration steps in @simongui's other PR.

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +205,35 @@ func (db *DB) Close() error {
...
+	var err error
+
+	for k := range db.sqlDatabases {
+		_ = errs.Combine(err, db.closeDatabase(k))
+	}
+	return err
+}
+
+// closeDatabase closes the specified SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabase(dbName string) (err error) {
```
egonelbre: Why handle them by name rather than by directly calling the method on the specific database implementation. It feels like an unnecessary indirection.

storagenode/storagenodedb/database.go
simongui: This became useful in my troubleshooting when errors occurred.

storagenode/storagenodedb/database.go
egonelbre: Where are versions for file migrations stored?

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +116,80 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
 	if err != nil {
 		return nil, ErrDatabase.Wrap(err)
 	}
 
-	// Set max idle and max open to 1 to control concurrent access to the memory DB
-	// Setting max open higher than 1 results in table locked errors
-	db.SetMaxIdleConns(1)
-	db.SetMaxOpenConns(1)
-	db.SetConnMaxLifetime(-1)
+	db.sqlDatabases[dbName] = sqlDB
```
isaachess: Because this has no mutex (which is fine), I would prefer a comment above the method that states it is not concurrent-safe. And that propagates all the way up to the `New` call.

It's a small thing, but I think it's nice to be explicit about these guarantees (and the lack of them) when possible.

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +205,35 @@ func (db *DB) Close() error {
...
 
-// Versions returns the instance of the versions database.
-func (db *DB) Versions() SQLDB {
-	return db.versionsDB
+// closeDatabases closes all the SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabases() error {
+	var err error
+
+	for k := range db.sqlDatabases {
+		_ = errs.Combine(err, db.closeDatabase(k))
```
isaachess: don't we need `err =` here? I believe this will discard all the errors ... ?

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +205,35 @@ func (db *DB) Close() error {
...
+	}
+	return err
+}
+
+// closeDatabase closes the specified SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabase(dbName string) (err error) {
+	if conn, ok := db.sqlDatabases[dbName]; ok {
```
egonelbre: Why would a database be double closed? This seems like an error prone way to handle life-cycles.

In general for lifecycles: 
1. For every open, there should be exactly one close call.
2. There should be no early closing of shared resources... e.g. which means if there's a resource `X` that is shared by different things only the last `close` should actually close the resource `X`... see `sync2` package with multitee things to ensure this is the case.

storagenode/storagenodedb/database.go
simongui: I think you reviewed the approach of multi-db migrations from another PR? That PR added a versions table to each database and dynamically creates it as needed. So this database is now completely legacy and only exists for the sequence of migration steps.

We did something similar with pieceinfo when we did a big refactor there and named it `V0PieceInfo`, we named it such a way that it was clear it was older code.

If you dislike the name suggest a new one and I'll happily rename 😄

storagenode/storagenodedb/database.go
simongui: Fixed!

storagenode/storagenodedb/database.go
simongui: I introduced this because as this code base is growing the longer the list of databases is becoming. Directly accessing objects in every type of access is starting to become brittle and error prone to copy/paste errors.

storagenode/storagenodedb/database.go
simongui: Done!

storagenode/storagenodedb/database.go
simongui: Fixed!

storagenode/storagenodedb/database.go
simongui: I'm aware of lifecycles and objects scope. The original design of the `database` object and its child objects had poor encapsulation to begin with that caused all kinds of bugs and friction. I tried to improve this incrementally in this PR. I don't think I introduced an early close or double close have I? If I have that wasn't intended 😄

storagenode/storagenodedb/database.go
simongui: Fixed!

storagenode/storagenodedb/database.go
egonelbre: If there's no double close you can replace this code with:

```
conn, ok := db.sqlDatabases[dbName]
if !ok { return Error.New("double close") }
```

storagenode/storagenodedb/storagenodesqldb.go
```diff
@@ -3,25 +3,25 @@
 
 package storagenodedb
 
-// versions represents the database that contains the database schema version history.
-type versionsDB struct {
-	location string
+type storageNodeSQLDB struct {
```
egonelbre: Stutter, the full name of this type is `storagenodedb.storageNodeSQLDB`

storagenode/storagenodedb/database.go
egonelbre: > I think you reviewed the approach of multi-db migrations from another PR?

Not really, I saw it was too big and noticed a few problems, so it had to be split up, hence I didn't spend time doing a proper review.

But for the name. I think "deprecated" might be a slightly better choice than "legacy" because we want to indicate that we shouldn't use it anymore.

Alternatively rather than moving everything out of the main file, we could skip migrating one db out of it and repurpose the main file, so we won't have an extra file.

storagenode/storagenodedb/database.go
egonelbre: Yeah looks, like it's correct and can be replaced with:
```
// closeDatabase closes the specified SQLite database connections and removes them from the associated maps.
func (db *DB) closeDatabase(dbName string) (err error) {
	conn, ok := db.sqlDatabases[dbName]
	if !ok {
		return ErrDatabase.New("double close")
	}
	delete(db.sqlDatabases, dbName)
	return ErrDatabase.Wrap(conn.Close())
}
```

storagenode/storagenodedb/database.go
egonelbre: I'm not quite following. why would `closeDatabase("bandwidth")` is less brittle than `bandwidth.close`. You can also have a map of databases that handle them.

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +209,35 @@ func (db *DB) Close() error {
...
+		db.closeDatabases(),
 	)
 }
 
-// Versions returns the instance of the versions database.
-func (db *DB) Versions() SQLDB {
-	return db.versionsDB
+// closeDatabases closes all the SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabases() error {
+	var err error
```
egonelbre: Use `errs.Group` this way it won't create a large chain of error wrapping.

storagenode/storagenodedb/pieceexpiration.go
```diff
@@ -16,17 +16,18 @@ import (
...
 }
 
-// newPieceExpirationDB returns a new instance of pieceExpirationDB initialized with the specified database.
-func newPieceExpirationDB(db SQLDB, location string) *pieceExpirationDB {
-	return &pieceExpirationDB{
-		location: location,
-		SQLDB:    db,
-	}
+// newPieceExpirationDB returns a new instance of pieceExpirationDB.
+func newPieceExpirationDB() *pieceExpirationDB {
```
egonelbre: I'm not sure whether the constructors are doing anything useful... might as well have `&pieceExpirationDB{}`... it would be just as clear.

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +117,83 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
-		v0PieceInfoDB:     newV0PieceInfoDB(versionsDB, versionsPath),
-		bandwidthDB:       newBandwidthDB(versionsDB, versionsPath),
-		ordersDB:          newOrdersDB(versionsDB, versionsPath),
-		pieceExpirationDB: newPieceExpirationDB(versionsDB, versionsPath),
-		pieceSpaceUsedDB:  newPieceSpaceUsedDB(versionsDB, versionsPath),
-		reputationDB:      newReputationDB(versionsDB, versionsPath),
-		storageUsageDB:    newStorageusageDB(versionsDB, versionsPath),
-		usedSerialsDB:     newUsedSerialsDB(versionsDB, versionsPath),
-		satellitesDB:      newSatellitesDB(versionsDB, versionsPath),
+		dbDirectory: filepath.Dir(config.Info2),
```
egonelbre: I'm wondering whether we should have each database separately configurable or not. There have been at least requests to move some databases to a different drive compared to the blob storage. Or at least let's make a flag that specifies the directory rather than reusing info2.

Also why pass in `Info2` with basename, when it's not being used?

storagenode/storagenodedb/database.go
```diff
@@ -115,108 +117,83 @@ func New(log *zap.Logger, config Config) (*DB, error) {
...
-	db.SetMaxIdleConns(1)
-	db.SetMaxOpenConns(1)
-	db.SetConnMaxLifetime(-1)
+	dbutil.Configure(sqlDB, mon)
 
-	mon.Chain("db_stats", monkit.StatSourceFunc(
-		func(cb func(name string, val float64)) {
-			monkit.StatSourceFromStruct(db.Stats()).Stats(cb)
-		}))
+	db.log.Debug(fmt.Sprintf("opened database %s", dbName))
```
egonelbre: this would add a lot of noise to storj-sim when starting/shutting down

storagenode/storagenodedb/database.go
```diff
@@ -232,22 +209,35 @@ func (db *DB) Close() error {
...
+}
+
+// closeDatabase closes the specified SQLite database connections and removes them from the associated maps.
+func (db *DB) closeDatabase(dbName string) (err error) {
+	if conn, ok := db.sqlDatabases[dbName]; ok {
+		err = errs.Combine(err, conn.Close())
+		delete(db.sqlDatabases, dbName)
+	}
+	if err == nil {
+		db.log.Debug(fmt.Sprintf("closed database %s", dbName))
```
egonelbre: This will end up adding a lot of noise to storj-sim. ~100 lines.

storagenode/storagenodedb/pieceexpiration.go
```diff
@@ -16,17 +16,18 @@ import (
 // ErrPieceExpiration represents errors from the piece expiration database.
 var ErrPieceExpiration = errs.Class("piece expiration error")
 
+const (
```
egonelbre: one const can be defined in one line

storagenode/storagenodedb/legacyinfo.go
simongui: Done!

storagenode/storagenodedb/database.go
simongui: Done.

storagenode/storagenodedb/database.go
simongui: If we stick with no array/map/etc list of databases it means every function that needs to operate across all the databases requires copy/pasting code. For example, this wouldn't be possible:

```
func (db *DB) closeDatabases() error {
	var err error

	for k := range db.sqlDatabases {
		err = errs.Combine(err, db.closeDatabase(k))
	}
	return err
}
```

It's brittle in the sense that when someone adds a new database it's very easy to miss spots in the code and bugs are extremely easy to introduce because you should have copy/pasted something you didn't. This alleviates _some_ of that.

I had considered shifting the concern at a higher scoped object (such as database) but this PR is so much redesigning already I felt that was over designing things for what we need right now. There are a lot of different ways to approach OO design and I felt this was already a big improvement over what we had.

storagenode/storagenodedb/storagenodesqldb.go
simongui: Fixed! Renamed to `migratableDB`.

storagenode/storagenodedb/database.go
simongui: Done!

storagenode/storagenodedb/pieceexpiration.go
simongui: Done!

storagenode/storagenodedb/database.go
simongui: Can this be a future feature request that we prioritize? The main purpose of these PR's is long over due.

storagenode/storagenodedb/database.go
simongui: I feel this is important for debugging both storj-sim and even more important for production. That's what `DEBUG` flagged logs are for, aren't they? Especially since you want the ability in the future to locate databases on a separate drive, that may increase more failure scenarios that we will need to troubleshoot and so will our support team.

storagenode/storagenodedb/pieceexpiration.go
simongui: Done.

storagenode/storagenodedb/database.go
egonelbre: Yup, a separate PR would be fine.

storagenode/storagenodedb/database.go
egonelbre: In that case, you would also want to log the path. For production, debug levels are disabled so they wouldn't be of any significant help.

egonelbre: Also, wouldn't the error come from some other place instead?

storagenode/storagenodedb/migrations_test.go
```diff
@@ -121,6 +125,17 @@ func TestMigrate(t *testing.T) {
 
 		// verify schema and data for each db in the expected snapshot
 		for dbName, dbSnapshot := range multiDBSnapshot.DBSnapshots {
+			// If the tables and indexes of the schema are empty, that's
+			// semantically the same as nil. Set to nil explicitly to help with
+			// comparison to snapshot.
+			schema, ok := schemas[dbName]
+			if ok && len(schema.Tables) == 0 {
```
egonelbre: Tip: this can be also checked with using `go-cmp` and https://godoc.org/github.com/google/go-cmp/cmp/cmpopts#EquateEmpty option, this means you don't have to manually find all the places that need fixing.

isaachess:
Looks great!

PR #3062, opened 2019-09-17
uplink/satellite: fix for case when inline segment is last one

mniewrzal:
What: With migration to new metainfo API I missed a case when the last segment can be inline segment. For example, there is such a case when file size is multiplication of max segment size. 

Why: We had a bug where files uploaded with old uplink were not accessible because new uplink was not able to get redundancy scheme values.

Please describe the tests:
 - Test 1: backward compatibility test and test-sim tests updated
 - Test 2: to test manual try upload file with the previous release and download with master/this branch
 
Please describe the performance impact: overhead should be minimal

### DIFF
--- a/satellite/metainfo/metainfo.go
+++ b/satellite/metainfo/metainfo.go
@@ -1082,6 +1082,11 @@ func (endpoint *Endpoint) CommitObject(ctx context.Context, req *pb.ObjectCommit
 		return nil, status.Errorf(codes.NotFound, "unable to find object: %q/%q", streamID.Bucket, streamID.EncryptedPath)
 	}
 
+	if lastSegmentPointer.Remote == nil {
+		lastSegmentPointer.Remote = &pb.RemoteSegment{}
+	}
+	// RS is set always for last segment to emulate RS per object
+	lastSegmentPointer.Remote.Redundancy = streamID.Redundancy
 	lastSegmentPointer.Metadata = req.EncryptedMetadata
 
 	err = endpoint.metainfo.Delete(ctx, lastSegmentPath)
@@ -1158,6 +1163,33 @@ func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetReques
 
 	if pointer.Remote != nil {
 		object.RedundancyScheme = pointer.Remote.Redundancy
+	} else if streamMeta.NumberOfSegments == 0 || streamMeta.NumberOfSegments > 1 {
+		// workaround
+		// new metainfo API redundancy scheme is on object level (not per segment) because
+		// of that RS is taken always from last segment, old implemetation were saving RS
+		// per segment and in some cases when for remote file last segment is inline segment
+		// then we are missing RS. This part will search for RS in othere segments than last one
+
+		// NumberOfSegments == 0 - pointer with encrypted num of segments
+		// NumberOfSegments > 1 - pointer with unencrypted num of segments and multiple segments
+
+		index := int64(0)
+		for {
+			pointer, _, err = endpoint.getPointer(ctx, keyInfo.ProjectID, index, req.Bucket, req.EncryptedPath)
+			if err != nil {
+				if storage.ErrKeyNotFound.Has(err) {
+					break
+				}
+
+				endpoint.log.Error("unable to get pointer", zap.Error(err))
+				return nil, status.Error(codes.Internal, "unable to get object")
+			}
+			if pointer.Remote != nil {
+				object.RedundancyScheme = pointer.Remote.Redundancy
+				break
+			}
+			index++
+		}
 	}
 
 	return &pb.ObjectGetResponse{
@@ -1541,12 +1573,18 @@ func (endpoint *Endpoint) MakeInlineSegment(ctx context.Context, req *pb.Segment
 		// that will be affected is our per-project bandwidth and storage limits.
 	}
 
+	metadata, err := proto.Marshal(&pb.SegmentMeta{
+		EncryptedKey: req.EncryptedKey,
+		KeyNonce:     req.EncryptedKeyNonce.Bytes(),
+	})
+
 	pointer := &pb.Pointer{
 		Type:           pb.Pointer_INLINE,
 		SegmentSize:    inlineUsed,
 		CreationDate:   streamID.CreationDate,
 		ExpirationDate: streamID.ExpirationDate,
 		InlineSegment:  req.EncryptedInlineData,
+		Metadata:       metadata,
 	}
 
 	err = endpoint.metainfo.Put(ctx, path, pointer)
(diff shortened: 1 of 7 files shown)


satellite/metainfo/metainfo.go
```diff
@@ -1158,6 +1163,33 @@ func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetReques
...
 		object.RedundancyScheme = pointer.Remote.Redundancy
+	} else if streamMeta.NumberOfSegments == 0 || streamMeta.NumberOfSegments > 1 {
+		// workaround
+		// new metainfo API redundancy scheme is on object level (not per segment) because
+		// of that RS is taken always from last segment, old implemetation were saving RS
+		// per segment and in some cases when for remote file last segment is inline segment
+		// then we are missing RS. This part will search for RS in othere segments than last one
+
+		// NumberOfSegments == 0 - pointer with encrypted num of segments
+		// NumberOfSegments > 1 - pointer with unencrypted num of segments and multiple segments
```
navillasa: maybe these two comments can go above the check

uplink/storage/segments/store.go
```diff
@@ -161,7 +161,8 @@ func (s *segmentStore) Get(ctx context.Context, streamID storj.StreamID, segment
 	}
 
 	switch {
-	case len(info.EncryptedInlineData) != 0:
+	// no order limits also means its inline segment
+	case len(info.EncryptedInlineData) != 0 || len(limits) == 0:
```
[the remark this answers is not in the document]
jtolds: i don't think we should have two different fields both overloaded to indicate if something is inline. we should stick with just one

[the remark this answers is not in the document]
jtolds: if we have to use something besides EncryptedInlineData, why not use the pointer.Type instead?

satellite/metainfo/metainfo.go
```diff
@@ -1158,6 +1163,33 @@ func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetReques
 
 	if pointer.Remote != nil {
```
jtolds: pointer.Type?

satellite/metainfo/metainfo.go
```diff
@@ -1158,6 +1163,33 @@ func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetReques
...
 		object.RedundancyScheme = pointer.Remote.Redundancy
+	} else if streamMeta.NumberOfSegments == 0 || streamMeta.NumberOfSegments > 1 {
+		// workaround
```
navillasa: if it's a workaround does that mean we have a plan for handling this in the future?

super nit but there are a few spelling errors in the comment chunk below-- here are some edits that might also help with clarity:
```
	// The new metainfo API redundancy scheme is on object level (not per segment).
                // Because of that, RS is always taken from the last segment.
                // The old implementation saves RS per segment, and in some cases
                // when the remote file's last segment is an inline segment, we end up
		// missing an RS scheme. This loop will search for RS in segments other than the last one.
```

satellite/metainfo/metainfo.go
jtolds: oh ignore me here

satellite/metainfo/metainfo.go
```diff
@@ -1158,6 +1163,33 @@ func (endpoint *Endpoint) GetObject(ctx context.Context, req *pb.ObjectGetReques
...
+		// then we are missing RS. This part will search for RS in othere segments than last one
+
+		// NumberOfSegments == 0 - pointer with encrypted num of segments
+		// NumberOfSegments > 1 - pointer with unencrypted num of segments and multiple segments
+
+		index := int64(0)
+		for {
+			pointer, _, err = endpoint.getPointer(ctx, keyInfo.ProjectID, index, req.Bucket, req.EncryptedPath)
+			if err != nil {
+				if storage.ErrKeyNotFound.Has(err) {
```
navillasa: what happens in this path where we don't find the pointer? Will we still end up missing an RS scheme or will it not matter at that point?

satellite/metainfo/metainfo.go
```diff
@@ -1840,6 +1878,7 @@ func (endpoint *Endpoint) DownloadSegment(ctx context.Context, req *pb.SegmentDo
 			}
 			segmentMeta = streamMeta.LastSegmentMeta
 		} else {
+			segmentMeta = &pb.SegmentMeta{}
```
navillasa: why are we adding this?

uplink/storage/segments/store.go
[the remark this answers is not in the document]
mniewrzal: > if it's inline, len(EncryptedInlineData) must be > 0, right?

no, we can have empty inline segment. I was surprised too. 

> if we have to use something besides EncryptedInlineData, why not use the pointer.Type instead?

I think we missed that during designing new methods for metainfo. At the moment its what we have after `DownloadSegment`. I agree its a little bit ugly. Maybe we can update this request but we shouldn't do this in a hurry.

satellite/metainfo/metainfo.go
mniewrzal: unfortunately no. With new metainfo object is keeping redundancy scheme and we need to put RS into inline and remote pointers (last segment). We are not sure what type of will be last segment.

uplink/storage/segments/store.go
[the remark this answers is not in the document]
jtolds: okay weird. i would have thought for sure that an encrypted empty inline value would still have encryption padding and therefore wouldn't be empty. that might be worth tracking down what's going on there. maybe we have a problem with inline encryption

satellite/metainfo/metainfo.go
mniewrzal: otherwise, its `nil` and unmarshaling will fail

satellite/metainfo/metainfo.go
mniewrzal: at this point we did what we could and we need to go without RS. In theory, it shouldn't happen with uplink/libuplink.

satellite/metainfo/metainfo.go
mniewrzal: > if it's a workaround does that mean we have a plan for handling this in the future?

Maybe, I'm not sure. Maybe we can do migration for files without RS in last segment but it needs to be discussed. 

Thank you for fixing spelling errors :)

satellite/metainfo/metainfo.go
mniewrzal: done

uplink/storage/segments/store.go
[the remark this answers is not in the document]
mniewrzal: yeah maybe, its happening here https://github.com/storj/storj/blob/64c467ffe761882e766c981d0fdcb409882f0093/uplink/storage/streams/store.go#L208

empty `data` is going into `encryption.Encrypt` and as a result we have also empty `cipherData`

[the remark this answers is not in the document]
jtolds: do you know what s.cipher is?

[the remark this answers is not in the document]
jtolds: yeah something's up:

https://play.golang.org/p/DmiQphUiEqv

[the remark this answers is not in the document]
jtolds: oh nvm, false alarm: https://github.com/storj/storj/blob/64c467ffe761882e766c981d0fdcb409882f0093/pkg/encryption/encryption.go#L37-L40

PR #3063, opened 2019-09-17
autoupdater: use blang/semver

mniewrzal:
What: This change updates auto-updater to use https://github.com/blang/semver as a tool to process version strings. 

Why: We want to be compliant with https://semver.org/

Please describe the tests:
 - Test 1: basic automatic test added
 

cmd/auto-updater/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	tmpExec := ctx.File("storagenode")
+	err = ioutil.WriteFile(tmpExec, content, 0755)
+	require.NoError(t, err)
+
+	var mux http.ServeMux
+	content, err = ioutil.ReadFile("testdata/fake-storagenode.zip")
+	require.NoError(t, err)
+	mux.HandleFunc("/download", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write(content)
```
navillasa: maybe redundant comment but looks like jenkins linter is unhappy that this doesn't have an error check

mniewrzal: thanks, I missed that one

mniewrzal:
> I guess I'm just curious what the main differences are between this new package and the previous `version.SemVer` values and methods used

In general our current support for version parsing is simple and in fact, we are not fully following  semantic versioning. This package implements parsing and processing (e.g. comparing) versions in a more comprehensive way. A decision about using it was made during the last arch review meeting.

cmd/auto-updater/main.go
```diff
@@ -167,37 +162,50 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
...
 		if err := update(ctx); err != nil {
 			// don't finish loop in case of error just wait for another execution
 			log.Println(err)
 		}
 		return nil
-	})
+	}
+
+	if loopInterval == 0 {
+		err = loopFunc(ctx)
```
kaloyan-raev: Why do we have this special case? Does it mean that if `loopInterval == 0` then the check or update will be executed only once and the program will exit?

cmd/auto-updater/main.go
```diff
@@ -167,37 +162,50 @@ func cmdRun(cmd *cobra.Command, args []string) (err error) {
...
 	}
 
 	scanner := bufio.NewScanner(bytes.NewReader(out))
 	for scanner.Scan() {
 		line := scanner.Text()
 		prefix := "Version: "
 		if strings.HasPrefix(line, prefix) {
-			return version.NewSemVer(line[len(prefix):])
+			// use version string without 'v'
+			return semver.Make(line[len(prefix)+1:])
```
kaloyan-raev: How are we sure that in this case the version starts with `v`? Should we have an additional `strings.HasPrefix(line, "v")` check?

cmd/auto-updater/main.go
mniewrzal: yes, with loopInterval == 0 program will execute only once. Its a case for automatic test.

cmd/auto-updater/main.go
mniewrzal: yeah, additional check makes sense. I will add it.

cmd/auto-updater/main.go
kaloyan-raev: This sounds like an easter egg.

To me, loopInterval == 0 sound like a valid case for the Cycle to not wait between loop runs.

Can we check for a negative interval instead? If interval < 0 then run only once. And document it in the flag description.

mniewrzal: done

kaloyan-raev: Oh, shoot! I thought `0` is a valid value for `sync2.NewCycle`, but it is the `time.Ticker` that does not accept it. Then maybe check for `loopInterval <= 0` instead. Sorry about that!

cmd/auto-updater/main.go
```diff
@@ -60,7 +61,7 @@ var (
 func init() {
 	rootCmd.AddCommand(runCmd)
 
-	runCmd.Flags().StringVar(&interval, "interval", "06h", "interval for checking the new version")
+	runCmd.Flags().StringVar(&interval, "interval", "06h", "interval for checking the new version, negative value will execute version check only once")
```
kaloyan-raev: This should now say "non-positive value" or "0 or less".

mniewrzal: ah yes :)

PR #3064, opened 2019-09-17
internal/dbutil/sqliteutil: add MigrateTablesToDatabase

isaachess:
What: Add a function `MigrateTablesToDatabase` to migrate from one `*sql.DB` to another, able to specify which tables you want to keep.

Why: To support splitting storagenodedb into multiple sql databases, we need a simple way to migrate from one db to another.

Please describe the tests:
 - Test 1: `TestMigrateTablesToDatabase` ensures we are migrating properly.
 - Test 2: `TestKeepTables` ensures that function is working as expected.
 
### DIFF
--- a/internal/dbutil/sqliteutil/migrator.go
+++ b/internal/dbutil/sqliteutil/migrator.go
@@ -6,6 +6,7 @@ package sqliteutil
 import (
 	"context"
 	"database/sql"
+	"fmt"
 
 	"github.com/mattn/go-sqlite3"
 	"github.com/zeebo/errs"
@@ -58,10 +59,10 @@ func MigrateTablesToDatabase(ctx context.Context, srcDB, destDB *sql.DB, tablesT
 	}
 
 	if err := srcConn.Close(); err != nil {
-		return err
+		return errs.Wrap(err)
 	}
 	if err := destConn.Close(); err != nil {
-		return err
+		return errs.Wrap(err)
 	}
 
 	// Remove tables we don't want to keep from the cloned destination database.
@@ -145,7 +146,7 @@ func KeepTables(ctx context.Context, db *sql.DB, tablesToKeep ...string) error {
 		var tableName string
 		err = rows.Scan(&tableName)
 		if err != nil {
-			return errs.Wrap(err)
+			return errs.Combine(err, rows.Close())
 		}
 		tables = append(tables, tableName)
 	}
@@ -158,14 +159,15 @@ func KeepTables(ctx context.Context, db *sql.DB, tablesToKeep ...string) error {
 	for _, tableName := range tables {
 		if !tableToKeep(tableName, tablesToKeep) {
 			// Drop tables we aren't told to keep in the destination database.
-			_, err = db.Exec("DROP TABLE "+tableName+";", nil)
+			_, err = db.Exec(fmt.Sprintf("DROP TABLE %s;", tableName))
 			if err != nil {
 				return errs.Wrap(err)
 			}
 		}
 	}
 
-	// VACUUM the database to reclaim the space used by the dropped tables.
+	// VACUUM the database to reclaim the space used by the dropped tables. The
+	// data will not actually be reclaimed until the db has been closed.
 	_, err = db.Exec("VACUUM;")
 	if err != nil {
 		return errs.Wrap(err)
--- a/internal/dbutil/sqliteutil/migrator.go
+++ b/internal/dbutil/sqliteutil/migrator.go
@@ -1,0 +1,183 @@
+// Copyright (C) 2019 Storj Labs, Inc.
+// See LICENSE for copying information.
+
+package sqliteutil
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/mattn/go-sqlite3"
+	"github.com/zeebo/errs"
+)
+
+// MigrateTablesToDatabase copies the specified tables from srcDB into destDB.
+// All tables in destDB will be dropped other than those specified in
+// tablesToKeep.
+func MigrateTablesToDatabase(ctx context.Context, srcDB, destDB *sql.DB, tablesToKeep ...string) error {
+	// Retrieve the raw Sqlite3 driver connections for the src and dest so that
+	// we can execute the backup API for a corruption safe clone.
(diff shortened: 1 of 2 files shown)


internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	// Collect a list of the tables. We must do this because we can't do DDL
+	// statements like drop tables while a query result is open.
+	var tables []string
+	for rows.Next() {
+		var tableName string
+		err = rows.Scan(&tableName)
+		if err != nil {
+			return errs.Wrap(err)
```
zeebo: This needs to close `rows`.

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,183 @@
...
+	err = rows.Close()
+	if err != nil {
+		return errs.Wrap(err)
+	}
+
+	// Loop over the list of tables and decide which ones to keep and which to drop.
+	for _, tableName := range tables {
+		if !tableToKeep(tableName, tablesToKeep) {
+			// Drop tables we aren't told to keep in the destination database.
+			_, err = db.Exec("DROP TABLE "+tableName+";", nil)
```
zeebo: why not `db.Exec("DROP TABLE ?", tableName)`? Regardless, passing `nil` here doesn't seem appropriate. That's passing `[]interface{}{nil}` rather than `[]interface{}{}` since the arguments list is variadic.

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,183 @@
...
+			if err != nil {
+				return errs.Wrap(err)
+			}
+		}
+	}
+
+	// VACUUM the database to reclaim the space used by the dropped tables.
```
simongui: We should probably update my out of date comment here to say that the disk space won't be reclaimed until the `Close()` method is called in case someone else uses this code in the future 😄

internal/dbutil/sqliteutil/migrator.go
isaachess: Fixed.

internal/dbutil/sqliteutil/migrator.go
isaachess: Fixed.

isaachess: Sorry, I had to change it to `db.Exec(fmt.Sprintf("DROP TABLE %s;", tableName)`. It appears that sqlite does not allow table names to be used in parameter substitution.

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,185 @@
...
+			return errs.Wrap(err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return errs.Wrap(err)
+	}
+
+	if err := srcConn.Close(); err != nil {
```
egonelbre: srcConn is not closed on errors

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,185 @@
...
+		return errs.Wrap(err)
+	}
+	if err := destConn.Close(); err != nil {
```
egonelbre: destConn is not closed on errors

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,185 @@
...
+				return errs.New("unable to get database driver")
+			}
+
+			err = backup(ctx, srcSqliteConn, destSqliteConn)
+			if err != nil {
+				return errs.New("unable to backup database")
+			}
+			return nil
+		})
+		if err != nil {
```
egonelbre: errs.Wrap automatically ignores nils

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// Collect a list of the tables. We must do this because we can't do DDL
+	// statements like drop tables while a query result is open.
+	var tables []string
+	for rows.Next() {
+		var tableName string
+		err = rows.Scan(&tableName)
+		if err != nil {
+			return errs.Combine(err, rows.Close())
```
egonelbre: `rows` is not closed on error.

internal/dbutil/sqliteutil/migrator_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func TestMigrateTablesToDatabase(t *testing.T) {
+	ctx := context.Background()
+	srcDB := newMemDB(t)
+	destDB := newMemDB(t)
+
+	defer func() {
+		require.NoError(t, srcDB.Close())
+		require.NoError(t, destDB.Close())
```
egonelbre: destDB is not closed when srcDB closing fails

internal/dbutil/sqliteutil/migrator_test.go
```diff
@@ -0,0 +1,94 @@
...
+	snapshot, err := sqliteutil.LoadSnapshotFromSQL(query)
+	require.NoError(t, err)
+
+	require.Equal(t, snapshot.Schema, destSchema)
+	require.Equal(t, snapshot.Data, destData)
+}
+
+func TestKeepTables(t *testing.T) {
+	ctx := context.Background()
+	db := newMemDB(t)
```
egonelbre: this db is not closed on failures

internal/dbutil/sqliteutil/migrator.go
```diff
@@ -0,0 +1,185 @@
...
+	err = rows.Close()
+	if err != nil {
+		return errs.Wrap(err)
+	}
+
+	// Loop over the list of tables and decide which ones to keep and which to drop.
+	for _, tableName := range tables {
+		if !tableToKeep(tableName, tablesToKeep) {
+			// Drop tables we aren't told to keep in the destination database.
+			_, err = db.Exec(fmt.Sprintf("DROP TABLE %s;", tableName))
```
egonelbre: How will database migration recover when the second drop tables fails?

PR #3066, opened 2019-09-17
add satelliteSystem to testplanet and update tests

JessicaGreben:
What/Why:

Now that we are going to be pulling apart the satellite into many subsystems, we need to update testplanet so that it can support these changes. This PR adds a new struct to testplanet called SatelliteSystem that will contains all the satellite services as we separate them out and updated existing tests to use this new struct.  Currently the SatelliteSystem only embeds the Satellite.Peer.

Please describe the tests:
Update any tests that used satellite.Peer and replace with testplanet.SatelliteSystem
 
n/a


internal/testplanet/planet.go
```diff
@@ -79,6 +79,11 @@ type Planet struct {
 	cancel func()
 }
 
+// SatelliteSystem contains all the processes needed to run a full Satellite setup
+type SatelliteSystem struct {
```
egonelbre: I just realized that this could be just named `testplanet.Satellite`, might end up looking nicer in code needing to use it. But `SatelliteSystem` is fine as well.

PR #3008, opened 2019-09-11
storagenode/storagenodedb: add new tables for graceful exit

VinozzZ:
What: 

- add `satellites_exit_progress` and `satellites` table into storagenodedb

Why:

These tables will be used to track graceful exit process on storage node
https://storjlabs.atlassian.net/browse/V3-2590?atlOrigin=eyJpIjoiZThjNzRiMzg1Y2RhNGJiOThjNzk1Y2NmYTJiNTRhNWQiLCJwIjoiaiJ9

Please describe the tests:
 - Test 1:
 - Test 2:
 
Please describe the performance impact:

## Code Review Checklist (to be filled out by reviewer)
 - [x] Does the PR describe what changes are being made?
 - [x] Does the PR describe why the changes are being made?
 - [x] Does the code follow [our style guide](https://github.com/storj/docs/blob/master/code/Style.md)?
 - [x] Does the code follow [our testing guide](https://github.com/storj/docs/blob/master/code/Testing.md)?
 - [x] Is the PR appropriately sized? (If it could be broken into smaller PRs it should be)
 - [x] Does the new code have enough tests? (*every* PR should have tests or justification otherwise. Bug-fix PRs especially)
 - [x] Does the new code have enough documentation that answers "how do I use it?" and "what does it do?"? (both source documentation and [higher level](https://github.com/storj/docs), diagrams?)
 - [x] Does any documentation need updating?
 - [x] Do the database access patterns make sense?

PR #3067, opened 2019-09-17
storagenode/storagenodedb: add SQL receiver logic for graceful exit

wthorp:
What: 
added graceful exit db methods
corrected "NUll" to "NULL" spelling
generalized `withTx()` for use with `SQLDB`
updated/created interfaces as necessary
added tests

Why:
We previously created graceful exit database tables for ticket https://storjlabs.atlassian.net/browse/V3-2590 in PR https://github.com/storj/storj/pull/3008. 
This PR makes those tables do something.

Please describe the tests:
 - Test 1: Each DB method doesn't explode
 
In the future, this may perform at all.


storagenode/satellites/satellites.go
```diff
@@ -0,0 +1,39 @@
...
+	//ExitedOk reflects a graceful exit that succeeded
+	ExitedOk
+	//ExitedFailed reflects a graceful exit that failed
+	ExitedFailed
+)
+
+// DB works with satellite database
+//
+// architecture: Database
+type DB interface {
```
mobyvb: We need a `GetGracefulExit` function or something to allow us to check the status

storagenode/storagenodedb/storagenodedbtest/satellites_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storagenode"
+	"storj.io/storj/storagenode/satellites"
+	"storj.io/storj/storagenode/storagenodedb/storagenodedbtest"
+)
+
+// initiate graceful exit doesn't explode
+func TestInitiateGracefulExitDoesNotExplode(t *testing.T) {
```
mobyvb: After a `Get...` function is added, we should improve all these tests by using it to validate that the data is what we expect.

storagenode/satellites/satellites.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	//Unexpected status should not be used for sanity checking
+	Unexpected Status = iota
+	//Normal status reflects a lack of graceful exit
+	Normal
+	//Exiting reflects an active graceful exit
+	Exiting
+	//ExitedOk reflects a graceful exit that succeeded
+	ExitedOk
```
mobyvb: nit - I think `ExitedSuccess` matches `ExitedFailed` better than `ExitedOk`

storagenode/storagenodedb/satellites.go
```diff
@@ -17,3 +24,41 @@ const SatellitesDBName = "satellites"
 type satellitesDB struct {
 	migratableDB
 }
+
+// initiate graceful exit
```
mobyvb: These comments should probably begin with the name of the function (I think the linter will complain otherwise)

storagenode/satellites/satellites.go
```diff
@@ -0,0 +1,51 @@
...
+	//Exiting reflects an active graceful exit
+	Exiting
+	//ExitSucceeded reflects a graceful exit that succeeded
+	ExitSucceeded
+	//ExitFailed reflects a graceful exit that failed
+	ExitFailed
+)
+
+//ExitProcess contains the status of a graceful exit
+type ExitProcess struct {
```
egonelbre: `ExitProgress` rather than `ExitProcess`

storagenode/storagenodedb/storagenodedbtest/satellites_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"storj.io/storj/internal/testcontext"
+	"storj.io/storj/pkg/storj"
+	"storj.io/storj/storagenode"
+	"storj.io/storj/storagenode/satellites"
+	"storj.io/storj/storagenode/storagenodedb/storagenodedbtest"
+)
+
+// TestGracefulExitDB tests the graceful exit database calls
+func TestGracefulExitDB(t *testing.T) { //satelliteID storj.NodeID, finishedAt time.Time, exitStatus satelliteStatus, completionReceipt []byte) (err error) {
```
mobyvb: Let's test with 2 distinct satellite Node IDs. The test looks good overall besides that.

wthorp: done

storagenode/satellites/satellites.go
```diff
@@ -0,0 +1,51 @@
...
+	Unexpected Status = iota
+	//Normal status reflects a lack of graceful exit
+	Normal
+	//Exiting reflects an active graceful exit
+	Exiting
+	//ExitSucceeded reflects a graceful exit that succeeded
+	ExitSucceeded
+	//ExitFailed reflects a graceful exit that failed
+	ExitFailed
+)
```
ethanadams: Should these values be explicitly set?  I'm concerned someone slips a constant in there and it changes all the values.

storagenode/satellites/satellites.go
ethanadams: Looks like this still needs to be changed to `ExitProgress`.

storagenode/satellites/satellites.go
```diff
@@ -0,0 +1,51 @@
...
+
+//ExitProcess contains the status of a graceful exit
+type ExitProcess struct {
+	SatelliteID       storj.NodeID
+	InitiatedAt       *time.Time
+	FinishedAt        *time.Time
```
ethanadams: nit - I don't think these need to be pointers.

wthorp: They do because they're nullable in the DB... the SQL libs do all the work if they're nullable.

storagenode/satellites/satellites.go
wthorp: weird.  bad merge maybe.  fixed.

storagenode/satellites/satellites.go
wthorp: fixed

